PR #9695, opened 2022-01-25
feat: add sharing column to trackedentityinstancefilter [DHIS2-12373]

ameenhere:
This PR adds a new column "sharing" of type jsonb to the table "trackedentityinstancefilter" table.


dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/trackedentityfilter/hibernate/TrackedEntityInstanceFilter.hbm.xml
```diff
@@ -41,6 +41,9 @@
     <property name="eventFilters" type="jlbEventFilter" />
 
     <property name="translations" type="jblTranslations"/>
+
+    <!-- Sharing -->
```
teleivo, line 45: I would remove the comment as it seems redundant to me. The property is already called sharing.

teleivo (approved):
Tiny change 😄 I am curious, how is sharing then implemented? I did not find any class/interface in TrackedEntityInstanceFilter hierarchy that looked like it had anything to do with sharing. Can you please point me to where the magic is happening 😄

zubaira (approved):

PR #9671, opened 2022-01-21
fix: periodOffset misses some data [DHIS2-12496]

jimgrace:
See [DHIS2-12496](https://jira.dhis2.org/browse/DHIS2-12496), especially the screen shots showing the problems and the fixes.

### Problem 1: periodOffset returns offset data only when there is also un-offset data

**The problem:** `DataHandler#getAggregatedDataValueMap` called `DataHandler#addItemBasedOnPeriodOffset` once for each combination of DimensionalItemObject (`dimItem`) and returned data row within the selected reporting periods. If the dimItem had a periodOffset, then this method called `PeriodOffsetUtils#getPeriodOffsetRow`. This method searched through all the returned grid data to find a row with data that matched the dimItem and the offset period. `addItemBasedOnPeriodOffset` then added to the result map a value using the key (including dimItem and period) from the original row, but the value from the offset row.

The problem was that this worked only if there were values for both the reported period and the offset period. If there was no value in the reported period, then the search through the grid for the offset period would not happen.

**The fix:** `DataHandler#getAggregatedDataValueMap` now calls `DataHandler#addItemBasedOnPeriodOffset` for each combination of dimItem and returned data row from the query (including those outside the selected reporting periods). If the dimItem has a periodOffset, it calls `PeriodOffsetUtils#getPeriodOffsetRow`. However this method, instead of looking through the entire grid for the offset period, now takes the given row with the offset period and shifts the period by the negative offset to find the period that the value should be reported for. It then returns a copy of the row with the offset period replaced by the reporting period. `addItemBasedOnPeriodOffset` then adds to the result map a value using both the key and the value from the adjusted row (but it only adds values if the resulting rows are within the selected reporting periods).

This ensures that that all the data desired is returned, shifted or not. Note that a single query row can provide the data value for multiple periodOffsets of the same data element, for example to a dimItem 3 periods later with a periodOffset of -3, and also to a dimItem 4 periods later with a periodOffset of -4.

As well as fixing the problem, these changes made the code smaller, simpler, and faster.

### Problem 2: periodOffset returns data only within selected reporting years

**The problem:** `DefaultQueryPlanner#groupByPeriodType` called `PeriodOffsetUtils#getPeriodTypePeriodMap`. This, in turn, called `PartitionUtils#getPeriodTypePeriodMap` and subsequently added map entries for all the shifted periods.

[...]

PR #9705, opened 2022-01-26
fix: DataHandler fix for dashboard 500 DHIS2-12518

jimgrace:
Symptom: See [DHIS2-12518](https://jira.dhis2.org/browse/DHIS2-12518). Some dashboard items aren't loading and getting a 500 error from the server.

Cause: In [pull/9671](https://github.com/dhis2/dhis2-core/pull/9671), in `DataHandler#getAggregatedDataValueMap` a test for `isPeriodInPeriods` was erroneously moved from an `if` clause to the `else` clause. It should have been removed altogether.

Testing: With this fix, all dashboard items display on all dashboards without the 500 error.


maikelarabori (approved):
Thanks @jimgrace :+1:

jbee (approved):

PR #9707, opened 2022-01-26
fix: update SMSCode data model

zubaira:
Update SMSCode data model and convert optionId integer field into CategoryOptionCombo reference. 

```
  {
    "compulsory": false,
    "code": "s",
    "dataElement": {
      "id": "...."
    },
    "optionId": {
      "id": "...."
    }
  }
]
```


dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/sms/hibernate/SMSCode.hbm.xml
```diff
@@ -24,7 +24,8 @@
 			class="org.hisp.dhis.trackedentity.TrackedEntityAttribute" column="trackedentityattributeid"
 			foreign-key="fk_trackedentityattribute_trackedentityattributeid" />
 
-		<property name="optionId" type="integer" />
+		<many-to-one name="optionId" class="org.hisp.dhis.category.CategoryOptionCombo"
+					 column="optionid" foreign-key="fk_categoryoptioncombo_categoryoptioncomboid" />
```
enricocolasante, line 28: Where does this foreign key `fk_categoryoptioncombo_categoryoptioncomboid` come from?
Do we need to create it in the flyway script?

enricocolasante (changes requested):

enricocolasante (approved):

PR #9714, opened 2022-01-27
feat: [WIP] attribute category options to dataValue

vietnguyen:
https://github.com/dhis2/notes/discussions/324


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalue/DataValueCategory.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Getter;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * @author viet@dhis2.org
+ */
+@Getter
+@AllArgsConstructor
+public class DataValueCategory
```
larshelge, line 42: I am thinking we only use one of these objects, i.e. just `DataValueCategory` and not also `AttributeCategory`, as these objects are identical. We can differentiate in the property name in `DataValue`.

larshelge:
Closing due to inactivity, please open a new PR later.

PR #9716, opened 2022-01-28
test: adds await to task summary requests

vilkg:
Async data value import test is sometimes flaky due to taskSummaries not being available immediately after the job is completed. Since it sounds like this is well-known behavior, adding a wait for taskSummaries.

### DIFF
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
@@ -70,7 +70,6 @@ public class SystemActions
             .extractList( "completed" ).contains( true );
 
         with()
-            .pollInterval( 1, TimeUnit.SECONDS )
             .atMost( 20, TimeUnit.SECONDS )
             .await().until( () -> taskIsCompleted.call() );
 
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
@@ -27,11 +27,16 @@
  */
 package org.hisp.dhis.actions;
 
-import java.util.List;
-import java.util.logging.Logger;
-
 import org.hisp.dhis.dto.ApiResponse;
 import org.hisp.dhis.dto.ImportSummary;
+
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+import static org.awaitility.Awaitility.await;
+import static org.awaitility.Awaitility.with;
 
 /**
  * @author Gintare Vilkelyte <vilkelyte.gintare@gmail.com>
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
```diff
@@ -46,30 +51,52 @@ public SystemActions()
...
+     * @param taskType
+     * @param taskId
+     * @return
+     */
+    public ApiResponse waitForTaskSummaries( String taskType, String taskId )
     {
-        return get( "/taskSummaries/" + taskType + "/" + taskId );
+        String url = String.format( "/taskSummaries/%s/%s", taskType, taskId );
+
+        await().ignoreExceptions().until( () -> !get( url ).validateStatus( 200 )
```
jbee, line 96: I think this now has a potential to never finish in case you never get a 200. We should add some timeout.

stian-sandvold (approved):

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
vilkg: @jbee Awatility's default timeout is 10 seconds. [[ref](https://github.com/awaitility/awaitility/wiki/Usage#defaults)]. I can set it explicitly if that's clearer.

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
```diff
@@ -46,30 +51,52 @@ public SystemActions()
...
-            completed = response.extractList( "completed" ).contains( true );
-        }
+        Callable<Boolean> taskIsCompleted = () -> getTask( taskType, taskId )
+            .validateStatus( 200 )
+            .extractList( "completed" ).contains( true );
+
+        with()
+            .pollInterval( 1, TimeUnit.SECONDS )
+            .atMost( 20, TimeUnit.SECONDS )
+            .await().until( () -> taskIsCompleted.call() );
```
teleivo, line 75: not sure how this looks in the logs if the test failed. adding a message to `await()` https://github.com/awaitility/awaitility/wiki/Usage#advanced could be helpful

teleivo (commented):
I think we should reduce the pollDelay
other than that I just have a few questions 😄

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
```diff
@@ -46,30 +51,52 @@ public SystemActions()
...
-            response = get( "/tasks/" + taskType + "/" + taskId );
-            response.validate().statusCode( 200 );
-            completed = response.extractList( "completed" ).contains( true );
-        }
+        Callable<Boolean> taskIsCompleted = () -> getTask( taskType, taskId )
+            .validateStatus( 200 )
+            .extractList( "completed" ).contains( true );
+
+        with()
+            .pollInterval( 1, TimeUnit.SECONDS )
```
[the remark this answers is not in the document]
vilkg: > why did you choose
> atMost(20, SECONDS) vs default 10s

I ran the tests many times and 10 seconds weren't enough to import for some of the runs. So just to avoid the flakes. 10 seconds is plenty for an import summary to be generated. 

> not ignoring exceptions vs ignoreExceptions

Ignoring exceptions is needed for the case when the response is null. Rest assured throws an exception then, which was also our original problem. I never saw task progress response being null, so didn't think there's a need to ignore exceptions there.

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
vilkg: I tried that, but it adds way too much noise and no useful output.

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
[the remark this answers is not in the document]
teleivo: Do you agree to explicitly set the `.pollDelay( 100, TimeUnit.MILLISECONDS )` so it will not be 1 second?

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
jbee: Ah, ok - as long as there is a timeout I am fine with this. No need to set it explicitly.

jbee (approved):

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/SystemActions.java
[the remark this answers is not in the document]
vilkg: Done

teleivo (approved):

PR #9724, opened 2022-01-31
fix: refine data entry app locked messages for DHIS2-4674

tomzemp:
This PR refines message for data sets to provide more detail than the current general purpose `Data set is locked`. See https://jira.dhis2.org/browse/DHIS2-4674 for background.

We have used the following messages:

| Situation | Message in UI |
| ------- | ------- |
| Date is past expiry days for period | Data entry has concluded |
| Date is outside of defined data input periods | Data input period has closed |
| Data is approved | Data has been approved and cannot be modified |
| Other* | Data set is locked |

\* General purpose message is left in case locked definition on backend (from DataSetService) is updated to cover locking reasons beyond those above.

GetDataValuesForDataSetAction has been modified to split current locked value into locked and approved values to provide necessary detail to frontend for determining the appropriate message.


tomzemp:
This PR has been closed and replaced by https://github.com/dhis2/dhis2-core/pull/9730 as SonarQube Analysis was not properly executing due to this PR being on a fork.

PR #9726, opened 2022-01-31
feat: extend teifilter model to include attribute value filters [DHIS2-12376]

ameenhere:
This PR extends the trackedentityinstancefilter table to add a jsonb column called "attributevaluefilters". This stores any filters related to tracked entity attribute values.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_30__Add_attributefilter_column_to_trackedentityinstancefilter.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_30__Add_attributefilter_column_to_trackedentityinstancefilter.sql
@@ -1,0 +1,1 @@
+alter table trackedentityinstancefilter add column if not exists attributevaluefilters jsonb default '[]'::jsonb;
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/TrackedEntityInstanceFilterController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/TrackedEntityInstanceFilterController.java
@@ -27,8 +27,14 @@
  */
 package org.hisp.dhis.webapi.controller;
 
+import java.util.List;
+
+import org.hisp.dhis.common.DhisApiVersion;
+import org.hisp.dhis.common.IllegalQueryException;
 import org.hisp.dhis.schema.descriptors.TrackedEntityInstanceFilterSchemaDescriptor;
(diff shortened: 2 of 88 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_30__Add_attributefilter_column_to_trackedentityinstancefilter.sql
```diff
@@ -0,0 +1 @@
+alter table trackedentityinstancefilter add column if not exists attributevaluefilters jsonb default '[]'::jsonb;
```
enricocolasante, line 1: Duplicated script?
Or you just want to be really sure? :D

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/TrackedEntityInstanceFilterController.java
```diff
@@ -38,7 +44,34 @@
  */
 @Controller
 @RequestMapping( value = TrackedEntityInstanceFilterSchemaDescriptor.API_ENDPOINT )
+@ApiVersion( include = { DhisApiVersion.ALL, DhisApiVersion.DEFAULT } )
```
enricocolasante, line 47: Do we really need this annotation here?

enricocolasante (changes requested):
Duplicated flyway script

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_30__Add_attributefilter_column_to_trackedentityinstancefilter.sql
ameenhere: haha. Actually Zubair "reserved" the 2.38.29 flyway version so I renamed my file. but for some reason, it got pushed as a new file. Deleted the 2.38.29. The new script is 2.38.30 :) Good catch!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/TrackedEntityInstanceFilterController.java
ameenhere: If the annotation is missing, api/{number}/rest/Of/the/url will not work. So I just made it to be standard. this means even if the api is accesed with or without the apiVersion, it continues to reach here.

enricocolasante (approved):

zubaira (approved):

PR #9734, opened 2022-02-01
feat: min/maxDate() for Indicator exprs DHIS2-9831

jimgrace:
See [DHIS2-9831](https://jira.dhis2.org/browse/DHIS2-9831). This PR adds two indicator expression functions:
  
  .minDate() – the start of the earliest period to get data values from
  
  .maxDate() – the end of the latest period to get data values from
  
  These functions limit the analytics query periods that will be returned from a data item, to accomodate installations where semantics shift with time so a dataElement/disaggregation until/after some time should be included in an indicator but not otherwise (for example it may cause double counting). PEPFAR needs this, and Scott has told me that other installations have similar shifting semantics and could make good use of these functions as well.
  
  ### QueryModifiers
  
  An `int periodOffset` field was formerly used in `DimensionalItemId` to hold a period offset as parsed from an expression, and then copied to `BaseDimensionalItemObject` so it could be applied in analytics. This has been replaced by a `QueryModifiers` object (common variable name `queryMods`) which holds the values for periodOffset, minDate, and maxDate. Like the former `int periodOffset`, it starts in `DimensionalItemId` to record the modifiers as parsed in the expression, and then is transferred to `BaseDimensionalItemObject` for use in analytics. When no modifiers are parsed for an item, the reference to `queryMods` is `null` in the `DimensionalItemId`, and subsequently `null` in the corresponding `BaseDimensionalItemObject`.
  
  `QueryModifiers` is idempotent; all fields are `final`. It has no constructors which must specify all the fields (outside of the internal one created and used by `@Builder`, which is not used directly). The only place in our code where all the fields must be referenced is in the `QueryModifiers#cloneBuilder` method, which allows a new `QueryModifiers` to be created from an old one while adding or overriding one (or could be multiple) modifiers. This is used in parsing to add a periodOffset, minDate, or maxDate to any any modifiers that may or may not already exist. The code doing the adding need only know about the field it is adding, not the other fields.
  
  Lombok does all the rest of the work for this class.
  
  When other modifiers are added in the future, such as for example an override for the default `AggregationType` of a `BaseDimensionalObject` as we have talked about for some time, they need only be defined in the grammar, defined in `QueryModifiers`, added to `QueryModifiers#cloneBuilder`, assigned a value in the routine that walks the parse tree, and referenced in analytics when they are used.
  
  ### Parsing
  
  `CommonExpressionVisitor` starts with a null `QueryModifiers queryMods` reference, which remains null unless any modifiers are parsed. When any modifier is parsed (by `PeriodOffset`, `FunctionMinDate`, or `FunctionMaxDate`), `CommonExpressionVisitor` gives that method a reference to a new `QueryModifiersBuilder`, if there is no queryMods yet, or to the `cloneBuilder` from the existing queryMods if there is. In this way, these modifier functions can be either chained or nested within parentheses, and a newly-constituted queryMods can add to other modifications and subsequently drop out when it becomes out of scope.
  
  Initially I thought I would need to duplicate the logic of the former `CommonExpressionVisitor#visitWithPeriodOffset` to add similar methods `visitWithMinDate` and `visitWithMaxDate`, but as I kept playing with the design this fell out, and is much nicer. Only one method `visitWithQueryMods` is needed to save the old queryMods, visit the subtree with a new (modified) queryMods, and then restore the old queryMods as the parse tree traversal continues.
  
  ### Transfer from DimensionalItemId to BaseDimensionalItemObject
  
  The transfer of a queryMods reference from `DimensionalItemId` to `BaseDimensionalItemObject` uses basically the same logic as did the transfer of the `int periodOffset`. queryMods (null or nonnull) is transferred from `DimensionalItemId` to `BaseDimensionalItemObject`. If it's nonnull and the `BaseDimensionalItemObject` is an atomic object (like `DataElement` from the database, but not like `ReportingRate` which is a new object constructed each time for a pair of `DataSet` and `ReportingRateMetric`), then the atomic object is cloned using serialization in `DataDimensionExtractor#cloneIfNeeded`, following the same logic that cloned an atomic object if periodOffset was nonzero. This is important to allow the same atomic object such as `DataElement` to be present multiple times in an expression with potentially different `queryMods`. An atomic object is cloned only if there are query modifiers present.
  
  ### Analytics
  
  `DefaultQueryPlanner#groupByMinMaxDate` creates a different subquery for each combination of minDate and maxDate. It follows the pattern of the other groupers in this class: It finds all the params `DataElements`, uses `QueryPlannerUtils#getMinMaxDateDateElementMap` to group them by combination of minDate and maxDate, and creates a different subquery for each combination.
  
  Fortunately the analytics tables already have columns `pestartdate` and `peenddate` to record the start and end of the period providing the data. These are used for minDate/maxDate so for example if the semantics changed at some quarter of the year referenced by minDate and/or maxDate, and if the reporting period requested is for a single yearly period, it can still aggregate the data from some quarters with one set of rules applied, and from other quarters with a different set of rules applied. Originally I thought I was going to be limiting the number of query `periods` and `filterPeriods`, but this is much better.
  
  `DataQueryParams#startDate` and `endDate` were already in place to query `pestartdate` and `peenddate`. All I had to do was decouple them in `JdbcAnalyticsManager` so that they operate independently. One, or the other, or both, or neither of these dates may be present.
  
  ### Testing

  `QueryModifiersTest` tests `QueryModifiers`.

  `ExpressionServiceTest` now tests the parsing of periodOffset, minDate, and maxDate. (It did not test periodOffset before.)
  
  `DimensionServiceTest` now tests periodOffset, minDate, and maxDate. (It did not test periodOffset before.)
  
  `QueryPlannerTest` adds a test that exercises the new `DefaultQueryPlanner#groupByMinMaxDate`.
  
  Four unit tests had to be changed because they depended on the order of data retrieved from a `Set`. Since `BaseDimensionalItemObject` was changed, objects like `DataElement` and `CategoryOptionCombo` are being returned in a different order. These four unit tests are `DataElementGroupResolverTest`, `CategoryOptionResolverTest`, `CategoyOptionGroupResolverTest`, and `CategoryComboMapTest`. All have been refactored so they do not depend on the order of objects retrieved from a `Set`.
  
  A test indicator was created that shows one data element up to an end date, a different data element between that date and a second date, and a third data element after the second date.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * The maximum date (end of any period) for querying this object.
+     */
+    @JsonProperty
+    private final Date maxDate;
+
+    /**
+     * Create a builder starting with values from this object.
+     */
+    public QueryModifiersBuilder cloneBuilder()
```
jbee, line 72: see `@Builder( toBuilder = true )`

jbee (changes requested):
IMHO `queryMods` should not be `null` when nothing in the expression provides details. Instead a NULL object can be used to prevent consumers of this information to handle this as two different cases. As I understand it the dates of that NULL object would be `null` while the offset is `0`. There could be a `QueryModifiers#isModified()` to check for this NULL object. This should remove the need for some `null` checks and risk of NPE.

As a general comment I feel the parsing solution is overly complicated in a way that drags out the code over several files which makes even this simple feature quite hard to follow. This is not a problom of this PR but the tech approch chosen here.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
jimgrace: Thanks for the tip! Done.

jimgrace:
@jbee You understand correctly about the defaults. Here's my twisted journey on the question of whether we should have `queryMods == null`:

1. Initially I had the exact same thought as you about `queryMods` not being null to avoid all the null testing and the NPE risk! I even started coding it that way initially. I allocated a default `QueryModifiers` object in `DimensionalItemId`, `BaseDimensionalItemObject`, and `CommonExpressionVisitor`. But I realized that most of the other objects in these classes initialized to null (or an empty collection) as the default. I figured that the way I had implemented it was not the "dhis2 way" and my reviewers might object. I misjudged!

2. Thinking that `default == null` was more the "dhis2 way", I refactored it to have null as the default, together with all the null checks. This is what you reviewed.

3. Your comment encouraged me to refactor back again to my original non-null code. So I went back to always having a non-null `queryMods` on object creation for `DimensionalItemId`, `BaseDimensionalItemObject`, and `CommonExpressionVisitor`. This however created problem with HTTP PATCH of `BaseDimensionalItemObject` descendents (like `DataElement`): we do the patching by loading the object from the database, converting it to json, patching the json, and persisting it back to the database. This created a problem. The e2e patch tests were failing. If I didn't include `queryMods` in the json, the patching failed. But I didn't want to include `queryMods` in json because then it would be exposed through the WebAPI as part of all the `BaseDimensionalItemObject` database objects. At least this is what I concluded; if you have further insight into this please let me know.

4. Next I tried for a hybrid solution: `BaseDimensionalItemObject` could have `queryMods == null` on object creation, but `DimensionalItemId` and `CommonExpressionVisitor` would always assign a default `queryMods` when created, and it would never be null. So the null tests never have to be done in `DimensionalItemId` and `CommonExpressionVisitor`. When the handoff of `queryMods` is made from `DimensionalItemId` to `BaseDimensionalItemObject`, then it will always be non-null in `BaseDimensionalItemObject` during analytics processing. This means that the null tests in `BaseDimensionalItemObject` never have to be done in the analytics code. Everything was looking good but this wreaked havoc on `ExpressionServiceTest`, where it creates many test objects that are descended from `BaseDimensionalItemObject`, and therefore these test objects had `queryMods == null`, but it was comparing them with `BaseDimensionalItemObjects` that come out of the parsing process and therefore have queryMods not null.

5. I considered the alternatives, including: (a) revert to `queryMods == null `when there are no mods, (b) mess with the `equals` method so that `BaseDimensionalItemObject` with default query mods equals `BaseDimensionalItemObject` with null query mods, or (c) fix `ExpressionServiceTest` so that it would work. I didn't like option (b) since it violates the usual expectation of the `equals` method, and means we couldn't someday use `lombak.EqualsAndHashCode` for this. I started trying to code option (c) in `ExpressionServiceTest`. For example where in `setUpTest` there is `valueMap = new ImmutableMap.Builder...`, I created an initial `valueMapWithoutQueryMods`, and then streamed it to add `queryMods` to each key, resulting in the final `valueMap`. But then I found that a lot of the tests create their own `BaseDimensionalItemObject` descendants (with a null `queryMods`), and test if the parsing code will return them.

6. I decided that option (a) was the least bad of the three choices. I backed up the Git commits so `null` is again the default for no modifiers.

Regarding your comment about the parsing solution being overly complicated, part of the purpose of this PR is to put in place the `QueryModifiers` object so similar mods in the future will touch many fewer files.

If you have suggestions for further simplification beyond this I would love to hear them! I'll take any blame or credit for the way ANTLR parsing was implemented in dhis2. Lars asked me to convert our expression parsing from regexp to ANTLR, and it was first released in indicator parsing in 2.31, Dec 2018. Since then it's been complicated further by putting the grammar and common classes in a separate repo `dhis2-antlr-expression-parser` as well as the program rule logic in `dhis2-rule-engine`, so all this code can be shared between the backend and the Android app. ANTLR allows us to provide a much richer expression syntax in a much more structured and maintainable way than the regexp parsing had done. But I'm always happy to hear suggestions for further improvement if you have them! Any major refactors might have to be coordinated between the backend and the Android app, but I'm all in favor of that if the result is better code.

jbee (approved):
You make avoiding `null` references sound like some kind of black magic :D It really should be the opposite. The pattern should help to make code more straight forward. But it has to be used consistently otherwise it breaks down. Loading an object should not be a problem or any different. When using the NULL object pattern there must never be a code path that can create an instance having a `null` reference instead of the NULL object. That is the pattern so if such a path exists you just have not applied the pattern fully. This simply means the object by default needs to initialize the field with the NULL object also for the code path for database loading.

I do agree that the code base currently leans more towards using `null` but I think this by now is not where we want to be and more an artefact of the past. Anyhow, this has given you enough trouble. We don't have to get religious about this - but we do want to move towards avoiding `null` references. 

My comment on the ANTLR parsing was more a personal note. I personally think using ANTLR here is total overkill and makes things complex for very little gain that in my mind does not justify using it. I can also see how a regex based parsing gets out of hand and I also would not use it for the problem we want to solve. I'd be very much in favour of a "hand written" PEG parser. If you leave all the academic fluff aside it just means you use methods to represent non-terminals. Your "grammar" then are methods calling other methods that are part of it. A grammar sequence is calling the corresponding methods in sequence and so forth. A recursion in grammar is methods calling each other recursivly. Very simple and powerful and most importantly it not only can but must be made and adopted for the problem at hand. Maybe we have a call some time to throw around some ideas and help me understand just how large the expression language is we try to process here.

maikelarabori (approved):

larshelge (approved):

PR #9736, opened 2022-02-01
feat: Add data set metadata endpoint [DHIS2-12522]

larshelge:
Adds a new metadata endpoint optimized for the new data entry web app. The response is based on the data sets for which the current user has data write access to, and includes the associated data elements, indicators and categories/combos/options. Category options are nested within categories. For data set categories, only category options for which the current user has data write access to are included.

Adds field presets `simple` and `identifiable`. Centralizes presets to interface `FieldPreset`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
@@ -29,6 +29,9 @@ package org.hisp.dhis.dxf2.metadata;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+/**
+ * @author Lars Helge Overland
+ */
 public interface DataSetMetadataExportService
 {
     /**
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
@@ -30,6 +30,11 @@ package org.hisp.dhis.dxf2.metadata;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
 /**
+ * Service responsible for retrieving metadata associated with data sets. The
+ * object response is suitable for applications for data capture and form
+ * rendering. This is a specialized service which related to the general
+ * {@link MetadataExportService}.
+ *
  * @author Lars Helge Overland
  */
 public interface DataSetMetadataExportService
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
@@ -41,5 +41,5 @@ public interface DataSetMetadataExportService
      *
      * @return an {@link ObjectNode}.
      */
-    ObjectNode getDataSetMetadata( MetadataExportParams params );
+    ObjectNode getDataSetMetadata();
 }
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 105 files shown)


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPath.java
```diff
@@ -98,6 +98,15 @@ public FieldPath( String name, List<String> path )
         this.transformers = new ArrayList<>();
     }
 
+    public FieldPath( String name, List<String> path, boolean exclude, boolean preset )
+    {
+        this.name = name;
+        this.path = path;
+        this.exclude = exclude;
+        this.preset = preset;
+        this.transformers = new ArrayList<>();
```
jbee, line 107: I just see `this.transformers` from the constructor above in the diff - I think it is best if we can limit ourselves to one constructor that actually sets the fields and the others just calling that one. If that isn't possible usually it is a good indication for some bad practice going on.

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
```diff
@@ -204,18 +208,29 @@ public void applyPresets( List<FieldPath> presets, Map<String, FieldPath> fieldP
             {
                 continue;
             }
-
-            if ( PRESET_ALL.equals( preset.getName() ) )
+            if ( FieldPreset.ALL.equals( preset.getName() ) )
```
jbee, line 211: This starts to look like `switch( preset.getName()) {` is the better approach here.

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
```diff
@@ -43,6 +45,42 @@
...
+     * Performs a flat mapping of the given collection using the given mapping
+     * function.
+     *
+     * @param <T>
+     * @param <U>
+     * @param collection the collection of objects to map.
+     * @param mapper the mapping function.
+     * @return a set of mapped objects.
+     */
+    public static <T, U> Set<U> flatMapToSet( Collection<T> collection,
```
jbee, line 58: Just a suggestion: for mapping it often reads more intuitive if we call the generics `A` and `B` instead of `T` and `U`.

dhis-2/dhis-support/dhis-support-commons/src/test/java/org/hisp/dhis/commons/util/CollectionUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.List;
+import java.util.Set;
+
+import org.hisp.dhis.dataelement.DataElement;
+import org.hisp.dhis.dataset.DataSet;
+import org.junit.jupiter.api.Test;
+
+public class CollectionUtilsTest
```
jbee, line 40: I think you can drop the `public` for test and method now that we use junit5. But I know this is not set in all modules.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/metadata/DataSetMetadataExportController.java
```diff
@@ -0,0 +1,62 @@
...
+public class DataSetMetadataExportController
+{
+    private final DataSetMetadataExportService exportService;
+
+    @GetMapping
+    public ResponseEntity<JsonNode> getMetadata()
+    {
+        ObjectNode jsonNode = exportService.getDataSetMetadata();
+
+        return new ResponseEntity<>( jsonNode, HttpStatus.OK );
```
jbee, line 60: There is `ResponseEntity.ok(body)` to make this nicer.

jbee (commented):
I did not quite understand the business level of this - lots of code seemed related to adding presets as a feature. So cannot really say if this makes sense on the business level. On code level it looks fine. Just some suggestions for improvements.

Another observation: For the most part we did keep the node level out of service layer this far but that moved lots of code into controllers which maybe never was a good thing. Just want to highlight that this now starts to move some of that into the service layer. Generally I think this is the right thing. What I am thinking about is if this maybe can or should accept a transformer of some kind so the most code is node independent and the final transformation to nodes or some other representation is provided by the caller.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DataSetMetadataExportService.java
```diff
@@ -0,0 +1,45 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.dxf2.metadata;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * @author Lars Helge Overland
+ */
+public interface DataSetMetadataExportService
```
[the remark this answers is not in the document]
larshelge: Agreed, added now.

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPath.java
larshelge: Yep, issue is that the properties are final and so calling another constructor is not allowed.

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
larshelge: Could be yes. Just followed the existing code, think I leave it for now.

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
larshelge: Alright, updated it.

dhis-2/dhis-support/dhis-support-commons/src/test/java/org/hisp/dhis/commons/util/CollectionUtilsTest.java
larshelge: Okay, updated it now.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/metadata/DataSetMetadataExportController.java
larshelge: Thanks for the tip, updated it now.

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPath.java
jbee: Hm... any constructor must set final fields and it is possible to call constructors to do so. So this might be a case where we just got into having multiple constructors that set fields because there is no constructor that does this for all fields which we could call. Anyhow, just wanted to point this out in case the code could make use of calling another constructor instead.

jbee (approved):

mortenoh (approved):

PR #9740, opened 2022-02-02
chore: what is failing e2e?

teleivo:


teleivo:
closing since we identified the missing metadata in the e2e tests.

https://github.com/dhis2/dhis2-core/pull/9741 fixes them
we should look into the 2 commits referenced here that introduced the issue. also why the test began to fail weeks after the test data changed.

teleivo:
closing since we identified the missing metadata in the e2e tests.

https://github.com/dhis2/dhis2-core/pull/9741 fixes them
we should look into the 2 commits referenced here that introduced the issue. also why the test began to fail weeks after the test data changed.

PR #9744, opened 2022-02-03
feat: Flush cache on analytics update DHIS2-12072

jimgrace:
See [DHIS2-12072](https://jira.dhis2.org/browse/DHIS2-12072). This PR:
- Always flushes the analytics cache after an analytics update
- Doesn't change how the cache settings affect the browser Cache-Control
- For cache mode `FIXED`, sets the internal analytics cache time to 2 weeks unless the strategy is `NO_CACHE`
- Doesn't change how cache mode `PROGRESSIVE` works

I chose 2 weeks for a fixed cache time instead of infinite, because 2 weeks is likely to be longer than until the next analytics update, and if it isn't (such as on a static demo system) this ensures that all cache entries will eventually age out (including any unpopular ones).

I haven't made more radical changes at this point, like ripping out all the `PROGRESSIVE` caching logic. We could do that later, or we could do that now if you like. (We would also have to take it out of the UI.)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCacheSettings.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCacheSettings.java
@@ -31,6 +31,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.Math.max;
 import static org.hisp.dhis.analytics.AnalyticsCacheTtlMode.FIXED;
 import static org.hisp.dhis.analytics.AnalyticsCacheTtlMode.PROGRESSIVE;
+import static org.hisp.dhis.common.cache.CacheStrategy.CACHE_TWO_WEEKS;
 import static org.hisp.dhis.common.cache.CacheStrategy.NO_CACHE;
 import static org.hisp.dhis.setting.SettingKey.ANALYTICS_CACHE_PROGRESSIVE_TTL_FACTOR;
 import static org.hisp.dhis.setting.SettingKey.ANALYTICS_CACHE_TTL_MODE;
@@ -110,7 +111,13 @@ public class AnalyticsCacheSettings
 
     /**
      * Retrieves the expiration time in seconds based on the system settings
-     * defined by the {@link org.hisp.dhis.setting.SettingKey#CACHE_STRATEGY}
+     * based on the {@link org.hisp.dhis.setting.SettingKey#CACHE_STRATEGY}. If
+     * it says not to cache, return 0 so no caching will take place. Otherwise
+     * return a long time. This is because we flush the analytics cache after on
+     * analytics rebuild. For this purpose, two weeks is considered to be "a
+     * long time". Two weeks is likely to be longer than until the next
+     * analytics rebuild, and if it isn't, this ensures that all cache entries
+     * will eventually be aged out.
      *
      * @see CacheStrategy
      *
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCacheSettings.java
```diff
@@ -121,25 +128,9 @@ public long fixedExpirationTimeOrDefault()
...
-            {
-                return defaultExpirationTime.toSeconds();
-            }
-            else
-            {
-                // Return ZERO (always expire)
-                return NO_CACHE.toSeconds();
-            }
-        }
+        return (NO_CACHE.equals( cacheStrategy ))
```
maikelarabori, line 131: Hi @jimgrace,

With this change, it seems that we can only return the expiration time of `0` or `2 weeks`. It means that the user can only choose between `2 weeks` and `0`.
In this case, do we want to ignore the other possible caching timeouts defined by the users, defined in `CacheStrategy`?

If this is the objective it is alright, but then I think it would be better to remove all other possible timeouts in `CacheStrategy` and also remove them from the UI...

maikelarabori (commented):
I have one minor comment @jimgrace. Actually, this is more like a question that I have regarding this change. Please take a look.

Thanks,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCacheSettings.java
jimgrace: @maikelarabori Thanks for the question. The idea is that the `CacheStrategy` will still determine the setting in the browser Cache-Control requests that come from the user, but effectively the backend data cache will be either NONE or 2 weeks. Allowing the user to select the `CacheStrategy` will still give them the choice of how the Cache-Control will be set in the browser, and hence how stale the data may appear from that perspective. But in the backend cache the data will never be stale because it is being flushed on analytics rebuild.

We will have to document this carefully.

In Slack messaging including PEPFAR, Lars, Phil, and me, we agreed that flushing the data cache in the backend on analytics rebuild was probably the right thing to do. I wrote this PR to see if this was a reasonable approach to implementing this. It looks like Lars wants your opinion on this too.

Under the circumstances, I think it might be reasonable to just delete all the `PROGRESSIVE` cache logic from the backend, and remove that option from the frontend as well. But it looked like it was a lot of work to implement that option in the first place, and I assumed it was done for good reason to support some installations. So at least for the moment I left it in. (It also means that we don't have to change the frontend at all for now.) But I could be persuaded that we should remove it now and change the frontend as well.

maikelarabori: Thanks for the great clarification @jimgrace. That makes sense :+1: 
Regarding the `PROGRESSIVE` thing, that was a requirement I was asked to implement some time ago.
I agree that it introduces extra complexity. If for some reason this can be dropped or is not really used by clients, I would certainly agree to remove it in order to keep things simple. But as you well said, some installations might miss it. This would be the only concern I guess.

Cheers,
Maikel

maikelarabori (approved):

larshelge (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCacheSettings.java
larshelge: Good points. Right I think we probably can remove the progressive caching. I certainly know this was my idea at the time and the thinking was that older data could have longer cache TTL than newer data. Though with the analytics tables and cache being in sync, this may perhaps be outdated now.

PR #9745, opened 2022-02-03
feat: As user I would like to have query time estimation feedback [2.38-DHIS2-TECH829-1]

d-bernat:
After demo there is an accepted suggestion to change the entry point for analytics queries analyze:
There are four new entry points:

analytics/explain
analytics/events/aggregate/{program}/explain
analytics/events/query/{program}/explain
analytics/enrollments/query/{program}/explain

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AnalyticsController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AnalyticsController.java
@@ -66,7 +66,7 @@ public class AnalyticsController
 {
     private static final String RESOURCE_PATH = "/analytics";
 
-    private static final String ANALYZE_PATH = "/analyze";
+    private static final String EXPLAIN_PATH = "/explain";
 
     private static final String DATA_VALUE_SET_PATH = "/dataValueSet";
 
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AnalyticsController.java
```diff
@@ -86,7 +86,7 @@
     // -------------------------------------------------------------------------
 
     @PreAuthorize( "hasRole('F_PERFORM_MAINTENANCE')" )
-    @GetMapping( value = RESOURCE_PATH + ANALYZE_PATH, produces = { APPLICATION_JSON_VALUE, "application/javascript" } )
+    @GetMapping( value = RESOURCE_PATH + EXPLAIN_PATH, produces = { APPLICATION_JSON_VALUE, "application/javascript" } )
     public @ResponseBody Grid getAnalyzeJson( // JSON
```
maikelarabori, line 90: Should we also change the name of the methods and any exiting variable, and use`explain` instead of `analyze`?

maikelarabori (approved):
Looks good @d-bernat, just one minor question for your attention.

Thanks!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AnalyticsController.java
d-bernat: yeah, good idea, let me change it

gnespolino (approved):

PR #9747, opened 2022-02-04
fix: attributeOptionCombo is not required in NTI DHIS2-12460

teleivo:
https://jira.dhis2.org/browse/DHIS2-12460

Main changes are in `PreCheckDataRelationsValidationHook.java`. Try not only looking at its diff, but the file as a whole as well. This should make the validations we are doing clearer.

## Bug Fixes

Fixes 3 bugs in the NTI when importing events:
* import of **valid** event with only attributeCategoryOptions set for a program with non default category combo **failed** (bug A) => works now
* import an **invalid** event where the attributeCategoryOption is not in the programs category combo. Storing the event with the default attributeoptioncomboid (bug B). => now fails with error E1117
* import an **invalid** event to a non-default program with an invalid AttributeOptionCombo id but a valid attributeCategoryOption. The event is stored the DB without an attributeoptioncomboid (bug C). => now fails with error E1115
* ProgramMapper did not map category combo if it was the default one. Due to mapstruct wrapping the set with `hasCategoryCombo` (returning false for default category combo). Renaming `hasCategoryCombo` fixes this.

### Error messages

Error messages where sometimes confusing. We sometimes used `E1115( "Could not find CategoryOptionCombo: `{0}`." )` even though the CategoryOptionCombo did exist. => only use E1115 and E1116 when the CategoryOptionCombo or CategoryOptions are not in the DB.

#### E1117

Use E1117 to signal that the combination of AttributeOptionCombo, CategoryOptions and program category combo is not correct/cannot be resolved.

E1117 missed one argument. It now gets the category combo, category options. 

#### E1053

Added error code E1053.

If payload has AOC and COs which do not match error is:
```json
      {
        "message": "AttributeCategoryOptions `FbLZS3ueWbQ` do not match the AttributeOptionCombo `l5QR5hJ4u44`.",
        "errorCode": "E1053",
        "trackerType": "EVENT",
        "uid": "nQPceuzJ14Q"
      }
```

#### E0154

Added error code E0154.

If payload contains an AOC that is not in the programs category combo.

```json
{
        "message": "AttributeOptionCombo `l5QR5hJ4u44` is not in the event programs category combo `CategoryCombo$HibernateProxy$nqNL79Nw (bjDvmb4bfuf)`.",
        "errorCode": "E1054",
        "trackerType": "EVENT",
        "uid": "WeGL87MzTSH"
      }
```

Previously we returned E1117 which has less details for the user to fix the problem.

#### Aggregate Errors

E1115 and E1116 => AOC, COs not found are aggregated into one report. Report will have one E1116 per CO not found.

### Tests

* made some mock based tests strict, so we only stub whats needed, keeping tests cleaner.
* unit tested the PreCheckDataRelationsValidationHook area that was not tested, where the 3 bugs were.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/TrackerImportValidationContext.java
```diff
@@ -201,16 +200,6 @@ public ProgramStageInstance getProgramStageInstance( String event )
         return bundle.getPreheat().getRelationship( bundle.getIdentifier(), relationship );
     }
 
-    public CategoryOptionCombo getCategoryOptionCombo( String id )
```
enricocolasante, line 204: The idea is to remove this class at some point, right?

I am not very familiar with the optionCombo/categoryCombo code but I thought one of the goal of this fix was to move the caches to the Preheat.
Maybe after this PR you have a much better understanding if this is possible and in case how to do it

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckDataRelationsValidationHook.java
```diff
@@ -116,20 +117,39 @@ public void validateEnrollment( ValidationErrorReporter reporter, Enrollment enr
...
         TrackerImportValidationContext context = reporter.getValidationContext();
-
         ProgramStage programStage = context.getProgramStage( event.getProgramStage() );
         OrganisationUnit organisationUnit = context.getOrganisationUnit( event.getOrgUnit() );
         Program program = context.getProgram( event.getProgram() );
 
+        validateProgramStageInProgram( reporter, event, programStage, program );
+        validateRegistrationProgram( reporter, event, context, program );
+        validateProgramHasOrgUnit( reporter, event, context, organisationUnit, program );
+        validateEventCategoryCombo( reporter, event, program );
```
enricocolasante, line 138: Would make sense to have the validation of category combo in its own hook?

enricocolasante (commented):
As I am not so familiar with the requirements, I just reviewed the code and I didn't checked all the different cases.
Just a couple of questions on the NTI refactor overall

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/TrackerImportValidationContext.java
teleivo: Yes :) I talked about it with Stian and we agreed to to this in a follow up https://jira.dhis2.org/browse/DHIS2-12623
This fix was already quite complex. I think we have a good set of tests now and at least another dev with more understanding of this 😋

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckDataRelationsValidationHook.java
teleivo: Maybe 🤷🏻 There is also EventCategoryOptValidationHook which already interacts with PreCheckDataRelationsValidationHook. PreCheckDataRelationsValidationHook has to come before EventCategoryOptValidationHook. PreCheckDataRelationsValidationHook has removeOnError `true` while `EventCategoryOptValidationHook` does not.

I could create a EventCategoryComboValidationHook with removeOnError `true` and make sure it runs before EventCategoryOptValidationHook. I just don't know if that has any effect on other validation hooks 🤷🏻 
If someone feels strongly about it I can investigate this in this issue. Otherwise I would postpone it, since this is already quite big of a change.

enricocolasante: Let's postpone it, absolutely.
It was just a consideration

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerErrorCode.java
```diff
@@ -71,6 +71,8 @@
     E1048( "Object: `{0}`, uid: `{1}`, has an invalid uid format." ),
     E1049( "Could not find OrganisationUnit: `{0}`, linked to Tracked Entity." ),
     E1050( "Event ScheduledAt date is missing." ),
+    E1053( "AttributeCategoryOptions `{0}` do not match the AttributeOptionCombo `{1}`." ),
```
abyot, line 74: do not match or " ... is not part of " ?

teleivo: It could be both

if

`aoc.getCategoryOptions().containsAll( categoryOptions )
            && aoc.getCategoryOptions().size() == categoryOptions.size()`

is true, than we know that the AOC is the identifier of the given categoryOptions. If this condition is false, we know its not. The reason could be that 1. the COs are missing a CO or 2. that a CO is given that is not within the AOCs COs.

Does the wording then make sense? For the error to clearly represent the 2 error cases we would probably need another error code and more logic.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckDataRelationsValidationHook.java
```diff
@@ -146,177 +166,354 @@ public void validateEvent( ValidationErrorReporter reporter, Event event )
...
+        TrackerImportValidationContext context, OrganisationUnit organisationUnit, Program program )
+    {
+        if ( programDoesNotHaveOrgUnit( program, organisationUnit, context.getProgramWithOrgUnitsMap() ) )
         {
             reporter.addError( event, E1029, organisationUnit, program );
         }
+    }
 
-        validateEventCategoryCombo( reporter, event, program );
+    private void validateEventCategoryCombo( ValidationErrorReporter reporter,
```
abyot, line 180: will `validateEventCategoryOptionCombo` be much more readable than `validateEventCategoryCombo` here?

Events categoryCombo is just program's categoryCombo.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckDataRelationsValidationHook.java
```diff
@@ -146,177 +166,354 @@ public void validateEvent( ValidationErrorReporter reporter, Event event )
...
+        {
+            reporter.addError( event, E1115, event.getAttributeOptionCombo() );
+            return false;
+        }
+        return true;
     }
 
-    private void validateRelationshipReference( ValidationErrorReporter reporter, Relationship relationship,
-        RelationshipItem item )
+    private boolean hasNoAttributeOptionComboSet( Event event )
```
abyot, line 246: should we rename the method to `hasNoAttributeCategoryOptionSet` instead?

teleivo: `hasNoAttributeOptionComboSet` checks `event.getAttributeOptionCombo()`. Since its not concerned with `AttributeCategoryOption` I would stick with its name.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckDataRelationsValidationHook.java
```diff
@@ -146,177 +166,354 @@ public void validateEvent( ValidationErrorReporter reporter, Event event )
...
 
-    // TODO: This method needs some love and care, the logic here is very hard
-    // to read.
-    protected void validateEventCategoryCombo( ValidationErrorReporter reporter,
-        Event event, Program program )
+    private boolean hasAttributeOptionComboSet( Event event )
     {
-        TrackerImportValidationContext context = reporter.getValidationContext();
-        TrackerPreheat preheat = reporter.getValidationContext().getBundle().getPreheat();
+        return !hasNoAttributeOptionComboSet( event );
```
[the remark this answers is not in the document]
teleivo: I find it more readable. Most of the checks use `hasNoAttributeOptionComboSet`. That's why I created it using `hasNo..` instead of `has`. So we don't have to do `!has...` all the time. Reading a "positive" check is easier, than negating in my head. There is a check that uses its negation `hasAttributeOptionComboSet` which to me is easier to read than `!hasNoAttributeOptionComboSet()`(that makes my head spin 😅).

abyot (approved):

PR #9757, opened 2022-02-06
fix: make local copy for object mapper filter provider

mortenoh:
Small fix for `FieldFilterService` that makes sure that the `FilterProvider` is set on a fresh copy of `ObjectMapper` (pre-configured for field filtering).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
+++ b/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
@@ -147,6 +147,9 @@ public class FieldFilterService
         return objectMapper;
     }
 
+    /**
+     * Recursively applies FieldTransformers to a Json node.
+     */
     private void applyTransformers( JsonNode node, JsonNode parent, String path,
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
```diff
@@ -106,7 +106,10 @@ public FieldFilterService( FieldPathHelper fieldPathHelper, ObjectMapper jsonMap
         fieldPathHelper.apply( fieldPaths, HibernateProxyUtils.getRealClass( firstObject ) );
 
         SimpleFilterProvider filterProvider = getSimpleFilterProvider( fieldPaths, params.isSkipSharing() );
-        ObjectMapper objectMapper = jsonMapper.setFilterProvider( filterProvider );
+
+        // only set filter provider on a local copy so that we don't affect
+        // other object mappers (running across other threads)
+        ObjectMapper objectMapper = jsonMapper.copy().setFilterProvider( filterProvider );
```
mortenoh, line 112: This is the only line that actually changes any behavior (the rest is cleanup + docs). This makes sure that a copy of the object mapper is created and so should be safe across threads/requests.

larshelge (approved):

PR #9759, opened 2022-02-07
feat: Add DataElement/TEA support for web_hook notifications[DHIS2-12576]

zubaira:
PR will add support for some additional template variables

```
    ENROLLMENT_ORG_UNIT_NAME( "enrollment_org_unit_name" )
    ENROLLMENT_ORG_UNIT_CODE( "enrollment_org_unit_code" )
```
Besides, that web request payload will now contain data elements and tracked entity attributes depending upon the object the web request was attached to i.e either `ProgramIntance` or `ProgramStageInstance`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultTrackerNotificationWebHookService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultTrackerNotificationWebHookService.java
@@ -31,12 +31,14 @@ import java.net.URI;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
+import org.hisp.dhis.eventdatavalue.EventDataValue;
 import org.hisp.dhis.notification.ProgramNotificationMessageRenderer;
 import org.hisp.dhis.notification.ProgramStageNotificationMessageRenderer;
 import org.hisp.dhis.program.ProgramInstance;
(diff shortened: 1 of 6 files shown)


mortenoh (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultTrackerNotificationWebHookService.java
```diff
@@ -124,10 +128,19 @@ public void handleEvent( String psi )
...
         ProgramStageNotificationMessageRenderer.VARIABLE_RESOLVERS
-            .forEach( ( key, value ) -> payload.put( key.name(), value.apply( instance ) ) );
-        sendPost( templates, renderService.toJsonAsString( payload ) );
+            .forEach( ( key, value ) -> requestPayload.put( key.name(), value.apply( instance ) ) );
+
+        // populate data values
+        instance.getEventDataValues().forEach( dv -> requestPayload.put( dv.getDataElement(), dv.getValue() ) );
+
+        // populate tracked entity attributes
+        instance.getProgramInstance().getEntityInstance().getTrackedEntityAttributeValues()
```
larshelge, line 140: Do we want to add a null check or verify that this is a tracker program instance perhaps to avoid NPE?

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultTrackerNotificationWebHookService.java
zubaira: i think we do have a check 
https://github.com/dhis2/dhis2-core/blob/4a6b022d0246f7dc6ea841f82ffb7659352784d7/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultTrackerNotificationWebHookService.java#L97

PR #9761, opened 2022-02-07
fix: Propagate includeDeleted options to enrollments and events [DHIS2-12285]

enricocolasante:


gnespolino (approved):

zubaira (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/TrackerTest.java
```diff
@@ -274,6 +279,27 @@ public TrackedEntityInstance persistTrackedEntityInstanceWithEnrollmentAndEvents
         return _persistTrackedEntityInstanceWithEnrollmentAndEvents( 5, enrollmentValues );
     }
 
+    public void deleteOneEnrollment(
+        org.hisp.dhis.dxf2.events.trackedentity.TrackedEntityInstance trackedEntityInstance )
+    {
+        List<Enrollment> enrollments = trackedEntityInstance.getEnrollments();
+        if ( !enrollments.isEmpty() )
```
teleivo, line 286: Should these helpers have an assertion in them in case no enrollment/event was deleted?

When I write a test in the future using `deleteOneEnrollment()` I think I'd assume it did delete one.

It is probably also useful to return the deleted enrollment/event. So one can use that in assertions in tests to make sure some API does not return that deleted entity. With that you then also have the choice of not making any assertion in the helper but letting the test do it. Not sure which is the best one. I lean towards making the assertion in the helper.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/trackedentity/store/DefaultEnrollmentStore.java
```diff
@@ -80,7 +80,8 @@ public DefaultEnrollmentStore( @Qualifier( "readOnlyJdbcTemplate" ) JdbcTemplate
     {
         EnrollmentRowCallbackHandler handler = new EnrollmentRowCallbackHandler();
 
-        jdbcTemplate.query( withAclCheck( GET_ENROLLMENT_SQL_BY_TEI, ctx, " pi.programid IN (:programIds)" ),
+        jdbcTemplate.query( withIncludeDeleted(
```
teleivo, line 83: I am not sure of a better name, but when I read this code, I think

-- it will include deleted records

but it will not necessarily since its checking the `ctx.getQueryParams().isIncludeDeleted()`.

Also should that be default behavior in tracker? To only include soft-deleted records if the user wants them? If so this design of having to wrap queries with `withIncludeDeleted()` can lead to someone forgetting to do that. Not sure if there is a way to do that by default to mitigate that risk. Of course only if that is the default behavior we want.

teleivo (commented):
I mostly have questions and some minor suggestions 😄

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/trackedentity/store/AbstractStore.java
```diff
@@ -156,6 +156,22 @@ MapSqlParameterSource createIdsParam( List<Long> ids, Long userId )
 
     abstract String getRelationshipEntityColumn();
 
+    /**
+     *
+     * @param sql an sql statement to which we want to "attach" the include
+     *        delete condition
+     * @param ctx the {@see AAggregateContext} object containing information
```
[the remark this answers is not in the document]
enricocolasante: Done

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/TrackerTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/trackedentity/store/DefaultEnrollmentStore.java
enricocolasante: @stian-sandvold @Bekkalizer I am really not sure what is the default behaviour here, do you have more information?

Anyway, if the default is to exclude soft-deleted values, I can try to reverse the logic here but in general we do not have a good solution about that in the system.

enricocolasante: I made it explicit that includeDeleted is false.

teleivo (approved):
👍🏻 😄

stian-sandvold (approved):

PR #9771, opened 2022-02-09
feat: use new field filter in web layer

mortenoh:
## Summary

* Updates `AbstractFullReadOnlyController` to use the new field filtering for `/api/type` and `/api/type/id` endpoints


dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/EventChartControllerTest.java
```diff
@@ -96,6 +96,8 @@ void testThatGetEventChartsDoesNotContainNewEventVisualizations()
 
         // Then
         final JsonResponse response = GET( "/eventCharts/" + uid ).content();
+        System.err.println( response );
```
jbee, line 99: remove

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -248,17 +249,14 @@ private UserQueryParams makeUserQueryParams( WebOptions options )
...
-        }
+        user.ifPresent( users::add );
 
         return users;
     }
 
     @Override
     @GetMapping( "/{uid}/{property}" )
-    public @ResponseBody RootNode getObjectProperty(
+    public @ResponseBody ResponseEntity<ObjectNode> getObjectProperty(
```
jbee, line 259: `@ResponseBody` can be removed

jbee (approved):
Simple cleanup left to do. 

Also the fix I did works but maybe isn't the best way to implement duplication. Then again the list is so short that this hardly matters.

Given the time-constrains I think we can also address the comments in a later PR.

netroms (approved):

PR #9775, opened 2022-02-09
fix: Exception:OrganisationUnit can't be null - [ DHIS2-12347 ] [ 2.38 ]

lucaCambi77:
https://jira.dhis2.org/browse/DHIS2-12347

When the org unit is null for psi on DB, we do not validate the user authorization and add a warning.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
@@ -43,6 +43,7 @@ import static org.hisp.dhis.tracker.validation.hooks.TrackerImporterAssertErrors
 
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 
 import org.hisp.dhis.category.CategoryOption;
 import org.hisp.dhis.category.CategoryOptionCombo;
@@ -79,6 +80,7 @@ import org.springframework.stereotype.Component;
  */
 @Component
 @RequiredArgsConstructor
+@Slf4j
 public class PreCheckSecurityOwnershipValidationHook
     extends AbstractTrackerDtoValidationHook
 {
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -461,9 +463,14 @@ private void checkEventWriteAccess( ValidationErrorReporter reporter, Event even
         checkNotNull( user, USER_CANT_BE_NULL );
         checkNotNull( programStage, PROGRAM_STAGE_CANT_BE_NULL );
         checkNotNull( programStage.getProgram(), PROGRAM_CANT_BE_NULL );
-        checkNotNull( eventOrgUnit, ORGANISATION_UNIT_CANT_BE_NULL );
 
-        if ( isCreatableInSearchScope ? !organisationUnitService.isInUserSearchHierarchyCached( user, eventOrgUnit )
+        if ( eventOrgUnit == null )
```
enricocolasante, line 467: If this is a temporary fix or a possible temporary fix, it would be better to add a TODO to remember that we need to have another look

enricocolasante (approved):
Maybe a TODO would help in this case

zubaira (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
lucaCambi77: we can add it but the test added should fail when we'll update that part later on

enricocolasante: I was just saying that we should have something to remind us that this is temporary but I just saw that https://jira.dhis2.org/browse/DHIS2-12600 was already created, so no need for a TODO

lucaCambi77: yes we have Jira and acceptance criteria so let's just use those instead of TODO that most of the time is forgetten

PR #9776, opened 2022-02-09
fix: data value export all format with or without compression [DHIS2-11925] (2.36) (#9769)

jbee:
Based on cherry-pick of #9769 

need some manual merging
* imports
* keeping the params object (future) in controller over the individual parameters (past)
* keeping extension of test setup from future


jbee:
This fails integration tests in a way I do not feel comfortable with making this change. At this point it is hard to say what of the changes might be behind this. I'd rather not have it backported.

PR #9778, opened 2022-02-09
refactor: Create ExpressionParams

jimgrace:
`ExpressionService.getExpressionValue` had 3 overloads with 2, 10, and 13 parameters, respectively.

This PR creates the class `ExpressionParams` to build the parameters for `getExpressionValue`, and reduces the three overloads to a single method `getExpressionValue( ExpressionParams exParams )`.

Mostly the callers were just mechanically refactored to use `ExpressionParams.builder()` to make the calls. However in a couple of cases parameters have been omitted in the call where they are now supplied by default.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
@@ -27,10 +27,17 @@
  */
 package org.hisp.dhis.expression;
 
+import static org.hisp.dhis.expression.MissingValueStrategy.NEVER_SKIP;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
+
 import org.hisp.dhis.analytics.DataType;
 import org.hisp.dhis.common.DimensionalItemId;
 import org.hisp.dhis.common.DimensionalItemObject;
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionService.java
```diff
@@ -275,59 +273,7 @@ void getExpressionDimensionalItemMaps( String expression, ParseType parseType, D
...
-     * @param dataType the data type the expression should return.
-     * @return the calculated value.
-     */
-    Object getExpressionValue( String expression, ParseType parseType,
-        Map<DimensionalItemId, DimensionalItemObject> itemMap, Map<DimensionalItemObject, Object> valueMap,
-        Map<String, Constant> constantMap, Map<String, Integer> orgUnitCountMap,
-        Map<String, OrganisationUnitGroup> orgUnitGroupMap, Integer days,
-        MissingValueStrategy missingValueStrategy, OrganisationUnit orgUnit, List<Period> samplePeriods,
-        MapMap<Period, DimensionalItemObject, Object> periodValueMap, DataType dataType );
+    Object getExpressionValue( ExpressionParams exParams );
```
jbee, line 278: Javadoc now misses the `@return` which given we return an `Object` is important to describe even beyond the old text ;)

jbee (commented):
Just two minor points to have a look at.

stian-sandvold (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionService.java
jimgrace: Ooops, done. I was too excited because I'd removed all the other params.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
```diff
@@ -0,0 +1,137 @@
...
+import org.hisp.dhis.period.Period;
+
+/**
+ * Parameters to evaluate an expression in {@see ExpressionService}
+ *
+ * @author Jim Grace
+ */
+@Getter
+@ToString
+@EqualsAndHashCode
```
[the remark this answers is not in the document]
jimgrace: Ah, understood, thanks! I needed `@EqualsAndHashCode` only for mocking `when`. And `@ToString` is just a kindness when debugging or reading a unit test failure.

jbee (approved):

PR #9792, opened 2022-02-11
fix: Missing errorReports when using DELETE strategy and validation fails [ DHIS2-11300 ]

lucaCambi77:
When we delete enrollments, a basic payload with uid is required. All the other info should be retrieved by pre-fetched entities (program, ou, and tei).

For ou, similarly to [12347](https://jira.dhis2.org/browse/DHIS2-12347), we allow ou to be null and log a warning.
Fixed also missing mapper for sharing in program and programinstance

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
@@ -180,7 +180,7 @@ public class PreCheckSecurityOwnershipValidationHook
         checkNotNull( enrollment, ENROLLMENT_CANT_BE_NULL );
         checkNotNull( program, PROGRAM_CANT_BE_NULL );
 
-        checkEnrollmentOrgUnit( enrollment, context, strategy, program, reporter );
+        checkEnrollmentOrgUnit( reporter, context, strategy, enrollment, program );
 
         if ( strategy.isDelete() )
         {
@@ -205,10 +205,10 @@ public class PreCheckSecurityOwnershipValidationHook
             ownerOrgUnit );
     }
 
-    private void checkEnrollmentOrgUnit( Enrollment enrollment, TrackerImportValidationContext context,
-        TrackerImportStrategy strategy, Program program, ValidationErrorReporter reporter )
+    private void checkEnrollmentOrgUnit( ValidationErrorReporter reporter, TrackerImportValidationContext context,
+        TrackerImportStrategy strategy, Enrollment enrollment, Program program )
     {
-        OrganisationUnit enrollmentOrgUnit = null;
+        OrganisationUnit enrollmentOrgUnit;
 
         if ( strategy.isUpdateOrDelete() )
         {
@@ -494,7 +494,7 @@ public class PreCheckSecurityOwnershipValidationHook
         checkNotNull( programStage, PROGRAM_STAGE_CANT_BE_NULL );
         checkNotNull( programStage.getProgram(), PROGRAM_CANT_BE_NULL );
 
-        checkEventOrgUnitWriteAccess( reporter, event, eventOrgUnit, isCreatableInSearchScope, bundle, user );
+        checkEventOrgUnitWriteAccess( reporter, bundle, event, eventOrgUnit, isCreatableInSearchScope, user );
 
         if ( programStage.getProgram().isWithoutRegistration() )
         {
@@ -519,9 +519,9 @@ public class PreCheckSecurityOwnershipValidationHook
         }
     }
 
-    private void checkEventOrgUnitWriteAccess( ValidationErrorReporter reporter, Event event,
+    private void checkEventOrgUnitWriteAccess( ValidationErrorReporter reporter, TrackerBundle bundle, Event event,
         OrganisationUnit eventOrgUnit,
-        boolean isCreatableInSearchScope, TrackerBundle bundle, User user )
+        boolean isCreatableInSearchScope, User user )
     {
         if ( eventOrgUnit == null )
         {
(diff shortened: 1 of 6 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -206,10 +201,42 @@ public void validateEnrollment( ValidationErrorReporter reporter, Enrollment enr
             }
         }
 
-        checkWriteEnrollmentAccess( reporter, enrollment, program, enrollment.getTrackedEntity(),
+        checkWriteEnrollmentAccess( reporter, enrollment, program, context,
             ownerOrgUnit );
     }
 
+    private void checkEnrollmentOrgUnit( Enrollment enrollment, TrackerImportValidationContext context,
```
teleivo, line 208: reporter usually goes first in these checks, I would stick to that.

I would go for this order `reporter, context, strategy, enrollment, program)` (grouping the "entities" like enrollment, program and the other types)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -464,25 +494,7 @@ private void checkEventWriteAccess( ValidationErrorReporter reporter, Event even
...
-            TrackerErrorReport error = TrackerErrorReport.builder()
-                .uid( event.getUid() )
-                .trackerType( TrackerType.EVENT )
-                .errorCode( TrackerErrorCode.E1000 )
-                .addArg( user )
-                .addArg( eventOrgUnit )
-                .build( bundle );
-            reporter.addError( error );
-        }
+        checkEventOrgUnitWriteAccess( reporter, event, eventOrgUnit, isCreatableInSearchScope, bundle, user );
```
teleivo, line 497: I would also group parameters `reporter, bundle` first.

this keeps it familiar since our TrackerValidationHook interface is
`void validate( ValidationErrorReporter report, TrackerImportValidationContext bundle )`

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHookTest.java
```diff
@@ -394,6 +394,39 @@ void verifyValidationFailsForTrackedEntityAndUserWithoutWriteAccess()
...
+        validatorToTest.validateEnrollment( reporter, enrollment );
+
+        assertFalse( reporter.hasErrors() );
+        verify( organisationUnitService, times( 0 ) ).isInUserHierarchyCached( user, organisationUnit );
+
+        when( ctx.getStrategy( enrollment ) ).thenReturn( TrackerImportStrategy.DELETE );
+
+        validatorToTest.validateEnrollment( reporter, enrollment );
+
+        assertFalse( reporter.hasErrors() );
```
teleivo, line 426: if these are two test cases, they should each get their own test so they can succeed/fail independently. and in theory be able to run in parallel

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -206,10 +201,42 @@ public void validateEnrollment( ValidationErrorReporter reporter, Enrollment enr
...
+        OrganisationUnit enrollmentOrgUnit = null;
+
+        if ( strategy.isUpdateOrDelete() )
+        {
+            enrollmentOrgUnit = context.getProgramInstance( enrollment.getEnrollment() )
+                .getOrganisationUnit();
+
+            if ( enrollmentOrgUnit == null )
+            {
+                log.warn( "ProgramInstance " + enrollment.getEnrollment()
```
teleivo, line 220: just to understand, in case we update or delete an enrollment, we allow it if the enrollment has no orgUnit assigned?
would there be another way to get the orgUnit? @enricocolasante 
I mean could someone not in the orgUnit of the enrollment delete it? I know the enrollment has none set in this case but maybe its program?

teleivo (commented):
I have some minor suggestions and a question 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
enricocolasante: I think this is temporary and part of a bigger problem.
OrgUnit shouldn't be null but we don't have this constraint in the DB, so some enrollments could not have and orgUnit.

Said so, I am not sure what are the rules to have the permission to delete. So I have no idea if it would be the same using the program orgUnit instead.
Maybe @stian-sandvold or @Bekkalizer have a better understanding of it

lucaCambi77: sorry, i am missing something here. What do you mean by program orgunit? afaik program has sharing, program instance has an orgunit assigned that, for now, can be null but there is a discussion to make it not null. Does a program itself has an orgunit assigned ?

teleivo: I was just wondering if there is another entity with an orgUnit that is connected to the enrollment where the orgUnit is not null 😅 So we could use that (like a proxy) to check if the user and this orgUnit match. Program has a `Set<OrgUnits>` but it's not mapped in the programs `PreheatMapper`.

This just came to mind, since I also don't know what the risks are of someone outside the "enrollments orgUnit" (it must belong to some orgUnit I assume) deleting the enrollment are. @Bekkalizer and @stian-sandvold probably know this 😁

teleivo: just asked in the standup. @stian-sandvold clarified it for me so I'll resolve this 😄

teleivo (approved):

PR #9793, opened 2022-02-11
fix: ReportTable/Chart migration [DHIS2-12601]

maikelarabori:
Some very old DHIS2 instances (ie.: 2.30, 2.31) are being updated and facing a migration issue related to the Flyway script updated in this PR.

The problem shows up because those old instances contain tables that are not present in Sierra Leone, so we never knew about their existence. So, we could not predict that we needed to take them into consideration as part of the removal of ReportTable and Chart APIs.

This script will remove those tables and force a CASCADE when removing the main tables (`reporttable` and `chart`)

For this script to work, we need to ask the clients/users to use the following Flyway flags in `dhis.conf`:
```
flyway.migrate_out_of_order=true
flyway.repair_before_migration=true
```

Do you think this fix is appropriate for this situation @larshelge / @Philip-Larsen-Donnelly ?

**NOTE**
_This would be also applied to 2.38._

Thanks!

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdSchemes.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdSchemes.java
@@ -44,6 +44,8 @@ public class IdSchemes
 
     private IdScheme dataElementIdScheme;
 
+    private IdScheme dataElementGroupIdScheme;
+
     private IdScheme categoryOptionComboIdScheme;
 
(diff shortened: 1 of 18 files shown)


larshelge:
I think we should not impose out or order on everyone as it is quite involved, and we will have to tell everyone about it, even when upgrading to a patch release.

maikelarabori:
We decided to favour the release notes in order to help the user to fix this issue/error.
It can be seen here: https://github.com/dhis2/dhis2-releases/pull/70

PR #9795, opened 2022-02-11
feat: Add description test for extractDataMatrixValue function [DHIS2-12364]

zubaira:


enricocolasante (changes requested):
Shouldn't you bump rule-engine version in the pom?

zubaira:
> Shouldn't you bump rule-engine version in the pom?

done

gnespolino (approved):

enricocolasante (approved):

PR #9807, opened 2022-02-14
feat: cached data integrity checks [DHIS2-12599]

jbee:
### Summary
On high level this adds caching to the data integrity checks.
But this has a number of consequences in the details.

The `GET` API now reads from the cache but does not trigger the checks.
To trigger the checks a `POST` is send to the same URL as reading cache.

As the computation triggered by a `POST` is now asynchronously performed by an ad-hoc scheduled job, calling `GET` directly after is likely not to find the results in the cache yet. Therefore a `timeout` parameter was added to the cache access API so that you can poll with some server side waiting (if wanted).
The idea here was that we could block wait on the cache and return as soon as the data is provided by the asynchronous computation. However, for now I did not implement block wait as the redis cache would not support it. Therefore I made a sleep wait loop trying to find it in the cache every 10-50ms (based on timeout). Since this is a very check operation I made the loop waiting relatively short to avoid long waits without a need for it.

Also the summary and details report got a `Date` field added for the `finishedTime` so that the user can understand how "old" this result is. The cache will hold results up to an hour. After that they outdate and the check must be rerun.

### Running Checks via Jobs
As now all data integrity checks are run from a job the data integrity job needed to be able to run the set of checks for one of 3 outcomes:
* report (legacy support of the old report)
* summary (new summary reports)
* details (new details reports)

This can be understood as a generally useful feature as soon as the job configuration also lets you chose this parameter in the app this means the users can also opt to run legacy or the new summary or details reports. 

For backwards compatibility the new type parameter is not required and when left out will perform the legacy report.

### Transactions
Doing all checks in  single transaction has shown problematic. While this would be nicest for consistency it has severe problems:
* a failing check can cause the transaction to fail and thereby fail all subsequent checks
* we get long running transactions potentially blocking database in significant ways (even read-only)

To avoid these problems transactions are now opened individually. For the programmatic checks this means they open when we call other services which are transnational. Usually this is the case once in these programmatic checks.
For the YAML based checks the transactions are now opened on the store level so that the service can call a spring proxy that has the `@Transactional` annotation. Therefore a new `DataIntegrityStore` and its implementation needed to be extracted from the service.

### Check Bugfix
The first YAML check had bad SQL for the summary check.
The convention is that a summary should return count and possibly a percentage.
The SQL was copied as still contained some text columns added for readability when used directly.  

### Automatic Testing
Existing tests have been adopted to cover the new asynchronous processing.
Luckily this was already setup for the legacy reports.

### Manual Testing
* Start the server and login
* Trigger summary: `POST` http://localhost:8080/api/dataIntegrity/summary
* Check summary: `GET` http://localhost:8080/api/dataIntegrity/summary?timeout=500
* Trigger details: `POST` http://localhost:8080/api/dataIntegrity/details
* Check details: `GET` http://localhost:8080/api/dataIntegrity/details?timeout=500
 
Note that a timeout of 500ms may not be long enough when running all checks like the examples given will do.
You could also test just single checks like:
* Trigger: `POST` http://localhost:8080/api/dataIntegrity/summary?checks=categories-no-options
* Check: `GET` http://localhost:8080/api/dataIntegrity/summary?checks=categories-no-options?timeout=500

### Documentation
https://github.com/dhis2/dhis2-docs/pull/906

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
@@ -519,8 +519,8 @@ public class DefaultDataIntegrityService
             .severity( DataIntegritySeverity.WARNING )
             .section( "Legacy" )
             .description( name.replace( '_', ' ' ) )
-            .runDetailsCheck( c -> new DataIntegrityDetails( c, new Date(), check.get() ) )
-            .runSummaryCheck( c -> new DataIntegritySummary( c, new Date(), check.get().size(), null ) )
+            .runDetailsCheck( c -> new DataIntegrityDetails( c, new Date(), null, check.get() ) )
+            .runSummaryCheck( c -> new DataIntegritySummary( c, new Date(), null, check.get().size(), null ) )
             .build() );
     }
 
@@ -814,7 +814,7 @@ public class DefaultDataIntegrityService
     {
         runDataIntegrityChecks( "Data Integrity summary checks", expandChecks( checks ), progress, summaryCache,
             check -> check.getRunSummaryCheck().apply( check ),
-            ( check, ex ) -> new DataIntegritySummary( check, new Date(), -1, null ) );
+            ( check, ex ) -> new DataIntegritySummary( check, new Date(), ex.getMessage(), -1, null ) );
     }
 
     @Override
@@ -830,8 +830,7 @@ public class DefaultDataIntegrityService
     {
         runDataIntegrityChecks( "Data Integrity details checks", expandChecks( checks ), progress, detailsCache,
             check -> check.getRunDetailsCheck().apply( check ),
-            ( check, ex ) -> new DataIntegrityDetails( check, new Date(),
-                List.of( new DataIntegrityIssue( "", "", ex.getMessage(), List.of() ) ) ) );
+            ( check, ex ) -> new DataIntegrityDetails( check, new Date(), ex.getMessage(), List.of() ) );
     }
 
     private <T> Map<String, T> getCached( Set<String> checks, long timeout, Cache<T> cache )
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.dataintegrity;
 
+import static java.lang.System.currentTimeMillis;
 import static java.util.Collections.unmodifiableCollection;
 import static java.util.Collections.unmodifiableSet;
 import static java.util.stream.Collectors.groupingBy;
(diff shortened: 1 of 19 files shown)


mediremi (approved):
LGTM

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
```diff
@@ -784,38 +802,95 @@ private String getInvalidExpressionDescription( String expression )
...
+
+    private <T> Map<String, T> getCached( Set<String> checks, long timeout, Cache<T> cache )
+    {
+        Set<String> names = expandChecks( checks );
+        long giveUpTime = currentTimeMillis() + timeout;
+        Map<String, T> resByName = new LinkedHashMap<>();
+        boolean retry = false;
+        do
+        {
+            if ( retry )
```
larshelge, line 844: This part seems a bit complicated. Just trying to understand the reason/logic for the retry?

jbee: The loop tries to get values from the cache. I assume you just triggered a set of checks moments ago so you will not immediately find all results in the cache. Therefore you can give the `timeout` value. When we have extracted results once from the cache we check if we have all results we wanted, if not and if we have not reached the timeout point we do sleep for a while and then try again to find the missing results in the cache. Again, if not complete sleep and retry. Until either we found all results in the cache or the timeout stops us. This makes it more convinient for clients to trigger and fetch the data. And gives us the chance to use block waiting in the future where we can wait exactly until all results are present in the cache (or some timeout occurred).

larshelge (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/hibernate/HibernateDataIntegrityStore.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    @Transactional( readOnly = true )
+    public DataIntegrityDetails queryDetails( DataIntegrityCheck check, String sql )
+    {
+        @SuppressWarnings( "unchecked" )
+        List<Object[]> rows = sessionFactory.getCurrentSession().createNativeQuery( sql ).list();
+        return new DataIntegrityDetails( check, new Date(), null, rows.stream()
+            .map( row -> new DataIntegrityIssue( (String) row[0],
```
netroms, line 76: Maybe a little dense/hard to read? Maybe put the forth param to DataIntegrityDetails on the line above/

netroms (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/hibernate/HibernateDataIntegrityStore.java
jbee: Had to merge this PR into #9814 - will try making it more readable there once this is merged so I can resolve the conflicts with master.

jbee: I introduced a helper method here https://github.com/dhis2/dhis2-core/pull/9814/files#diff-27419f335112488a229aa6db65ac231250233db214802e8e4d951725f565c5d8R79-R81 to make this more readable.

PR #9810, opened 2022-02-14
feat: new role (granted to execute explain) [2.38-DHIS2-12605]

d-bernat:
There is a new role needed. User in the role can perform explain analyze query on analytics queries: 
http://localhost:8080/dhis/api/38/analytics/explain?dimension=dx%3ATcTRasl29wo,pe%3ALAST_12_MONTHS&filter=ou%3AUSER_ORGUNIT&includeNumDen=false&skipMeta=true&skipData=false

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
@@ -371,6 +371,7 @@ F_MOBILE_DELETE_SMS=Delete SMS
 F_SEND_EMAIL=Send Email
 F_VIEW_SETTINGS=View settings
 F_PERFORM_MAINTENANCE=Perform maintenance tasks
+F_PERFORM_ANALYTICS_EXPLAIN=Perform analytics explain query  
 F_PRUNE_ORGANISATION_UNITS=Prune organisation units
 F_ORGANISATION_UNIT_SPLIT=Split organisation unit
 F_ORGANISATION_UNIT_MERGE=Merge organisation units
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v38/V2_38_32__Add_new_user_role_for_analytics_explain_endpoint.java
```diff
@@ -0,0 +1,87 @@
...
+                superUserRoleIds.add( superUserRole.getInt( 1 ) );
+            }
+        }
+        catch ( SQLException e )
+        {
+            log.error( "Flyway java migration error", e );
+            throw new FlywayException( e );
+        }
+
+        if ( !superUserRoleIds.isEmpty() )
```
gnespolino, line 69: why we add F_PERFORM_ANALYTICS_EXPLAIN only when superUser has no roles assigned ?

dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
```diff
@@ -1801,7 +1802,6 @@ oauth2_clients=OAuth2 clients
 
 #-- Authorities ---------------------------------------------------------------#
 
-F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS=Search tracked entity instance across all organisation units
```
gnespolino, line 1804: why we removed this one ?

gnespolino (approved):
Approved with a couple of questions I don't understand! Good one!

dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
d-bernat: it was twice there, I took the latest one written by Lars...

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v38/V2_38_32__Add_new_user_role_for_analytics_explain_endpoint.java
d-bernat: ???

maikelarabori (approved):
Looks good @d-bernat, just a very minor comment.

PR #9813, opened 2022-02-15
fix: program rule notification lazy load exception [DHIS2-12614]

zubaira:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultProgramNotificationService.java
```diff
@@ -309,7 +309,7 @@ public void sendProgramRuleTriggeredEventNotifications( long pnt, long programSt
     public void sendProgramRuleTriggeredEventNotifications( long pnt, ProgramStageInstance programStageInstance )
     {
         MessageBatch messageBatch = createProgramStageInstanceMessageBatch( notificationTemplateService.get( pnt ),
-            Collections.singletonList( programStageInstance ) );
+            Collections.singletonList( programStageInstanceStore.getByUid( programStageInstance.getUid() ) ) );
```
enricocolasante, line 312: If this is happening for PorgramStageInstance, are we sure it is not happening also for ProgramInstance?

enricocolasante (changes requested):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultProgramNotificationService.java
zubaira: tested programInstance notifications and they seem to work fine. Lazy exception is the result of accessing dataElement collection from programStage but in case of programInstance notification we don't have DataElements available to be shown in notification template.

PR #9814, opened 2022-02-15
feat: data integrity checks that only one default category exists [DHIS2-12219]

jbee:
### Summary
Adds 4 YAML based data integrity checks:
* there is only one category named `"default"`
* there is only one category option named `"default"`
* there is only one category combo named `"default"`
* there is only one category option combo named `"default"`

This also makes a small fix for the YAML based summary query allowing both queries that return only 1 value (count) and queries returning 2 values (count and percentage).

### Automatic Testing
New tests were added for each of the 4 details and summary checks.
A base class for data integrity checks was extracted.
The idea is that every section (like categories) uses its own test file to keep the dependencies of services needed specific to a single area. So test cases shown as deleted were only moved to another file. No existing test case was deleted.

### Manual Testing
The easiest way to manually test this is to change the UID of the row with name `default` of the 4 affected tables to some other UID. This then causes the existing rows to be flagged as they do not have the expected UID.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DefaultCategoryService.java
```diff
@@ -566,11 +566,7 @@ public void generateDefaultDimension()
     @Transactional( readOnly = true )
     public CategoryOptionCombo getDefaultCategoryOptionCombo()
     {
-        CategoryCombo categoryCombo = getCategoryComboByName( CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME );
-
-        return categoryCombo != null && categoryCombo.hasOptionCombos()
-            ? categoryCombo.getOptionCombos().iterator().next()
-            : null;
+        return categoryOptionComboStore.getByName( CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME );
```
jbee, line 569: FYI: I think the old code was fine but in a spring test changes did not become visible unless I would try to reload the value from the database.

larshelge (approved):

vietnguyen (approved):

PR #9820, opened 2022-02-15
refactor: Simplify CommonExpressionVisitor

jimgrace:
The main effect of this PR is to make `CommonExpressionVisitor` much smaller and better-organized so that it is easier to understand and maintain. It is reduced from 823 lines to 195 lines.

This PR also lays the foundation for implementing [DHIS2-12082](https://jira.dhis2.org/browse/DHIS2-12082) (OrgUnit dataSet and program expression functions) and [DHIS2-11522](https://jira.dhis2.org/browse/DHIS2-11522) (Indicator sub-expressions). I hope it can get a relatively quick review and approval (after fixing anything serious of course) so I can get these features into 2.38. I believe that the code can be further improved, but hopefully this can be done after the PR is merged rather than delaying its approval.

### CommonExpressionVisitor decomposition

In addition to service and I18n references, `CommonExpressionVisitor` used to have 26 fields, and it was not immediately obvious which were parameters, which were state variables, and which were output values. With the refactor it now has 4 fields plus references to the following four classes:

- `ExpressionParams exParams` was recently introduced to simplify and better structure calls to _ExpressionService.getExpressionValue_, but was also done with this refactor in mind. _exParams_ is now just referenced by _CommonExpressionVisitor_ rather than copying individual fields from _exParams_ into _CommonExpressionVisitor_.

- `ProgramExpressionParams progExParams` is likewise used inside _CommonExpressionVisitor_ for parameters provided by _ProgramIndicatorService_.

- `ExpressionState exState` contains state information that may change as nodes in a parsed expression tree are visited, like the current values of _periodOffset_ and _stageOffset_, whether nulls should be replaced by default values in this part of the parse tree, etc.

- `ExpressionInfo exInfo` contains information that is gathered and returned from the (raw, unevaluated) expression by the new `ExpressionItem::getExpressionInfo` method (see below).

Expression visiting code had been referencing all 26 fields directly from _CommonExpressionVisitor_. Now most of these fields must be referenced through these other classes. Examples:

> visitor.getValueMap() -> visitor.getExParams().getValueMap()
> visitor.getProgramIndicator() -> visitor.getProgExParams().getProgramIndicator()
> visitor.getStageOffset() -> visitor.getExState().getStageOffset()
> visitor.getItemIds() -> visitor.getExInfo().getItemIds()

### ExpressionItem visiting modes

The `ExpressionItem` interface has been refactored to reduce the number of main visiting modes from six to four:

`ExpressionItem::getExpressionInfo` (ITEM_GET_EXPRESSION_INFO) takes the place of both `getItemId` and `getOrgUnitGroup`. The `getItemId` method had collected _DimensionalItemIds_ (for _DataElements_, _ProgramAttributes_, etc.) and `getOrgUnitGroup` had collected _OrgUnitGroup_ ids. This followed the pattern of the pre-ANTLR code where a different regexp search was used to find different types of results. However, the upcoming [DHIS2-12082](https://jira.dhis2.org/browse/DHIS2-12082) will additionally require the collection of _DataSet_ and _Program_ ids from the new parse nodes. Rather than having one pass through the nodes for each different type of information to be collected, it makes sense to collect these in a single pass and put them into the `ExpressionInfo` object. In the future, calls to _ExpressionService_ may make better use of this single pass instead of multiple passes through the parse tree, by referring to a once-collected _ExpressionInfo_ object. The _ExpressionInfo_ object could even be cached. The same expression string will always produce the same _ExpressionInfo_ object.

Note that the `getExpressionInfo` method is still different from `getDescription`, since the latter fetches metadata object names from the database, which is only needed when the user is viewing the expression in the Maintenance app. Because of the database access, _getDescription_ takes substantially longer than _getExpressionInfo_.

Note also that the fromer `getOrgUnitGroup` had been kind of overloaded in that it collected the _orgUnitGroup_ ids from both the `OUG{uid}` org unit group count data item (used only in indicators) and also the `orgUnit.group` function (used only in validation rules and predictors). For clarity, these are now collected in two separate fields in _ExpressionInfo::orgUnitGroupCountIds_, and _ExpressionInfo::orgUnitGroupIds_.  They are retrieved by two different methods in _ExpressionService_.

`ExpressionItem::regenerate `(ITEM_REGENERATE) has been removed, along with the code that implemented it. (TLDR; This had been used to substitute constants and _orgUnitGroup_ counts into indicator expressions that are downloaded into the data entry form and evaluated in JavaScript to provide additional values for reference during data entry. Because ANTLR performed poorly at walking the parse tree so many times for data entry in some implementations, this feature was reverted back to regexp code, and can be found at `DefaultExpressionService::regenerateIndicatorExpression`.)

The remaining four visiting method definitions, used as _CommonExpressionVisitor_ arguments, have been moved from the _ParserUtils_ class to the _ExpressionItem_ interface where the methods they reference are defined. (Note that this code was written before lambda expressions were available. If there is a more elegant way to do this with lambda expressions, perhaps it can be done later. For the upcoming features these were the important changes to make.)

### constantMap

It was once the case that all users of `ExpressionService` had to fetch the constant map (all constants in the system) from `ConstantService` and pass this map to `ExpressionService.` However for getting expression information (_itemIds_, _orgUnitGroups_, and descriptions), the constant map was fetched by `DefaultExpressionService` on behalf of the caller. Profiling showed that this fetching was a substantial part of the time required for these operations, so a cache had been added.

When `ExpressionParams` was recently implemented, this behaviour was not changed, and _constantMap_ was added to _ExpressionParams_. For this PR, however, a simple refactor would have added _constantMap_ to _ProgramExpressionParams_ also. This would have resulted in the _constantMap_ being supplied by two different params classes.

Instead, _constantMap_ has now been removed from _ExpressionParams_ so it is no longer the burden of the _ExpressionService_ caller (indicators, validation rules, and predictors) to fetch. _DefaultExpressionService_ now always fetches the _constantMap_, using the cache, and provides the map to _CommonExpressionVisitor_. It also shares access to this cache for use by `DefaultProgramIndicatorService`, providing a consistent and fast way of supplying the _constantMap_ to all expression parsing.

### IdObjectManager

The `IdentifiableObjectManager` is now being used where possible to look up objects. This allows several class-specific services to be removed from _CommonExpressionVisitor_ (and from several classes that create it).

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionInfo.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionInfo.java
@@ -25,32 +25,47 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package org.hisp.dhis.program.variable;
+package org.hisp.dhis.expression;
 
-import org.hisp.dhis.parser.expression.CommonExpressionVisitor;
-import org.hisp.dhis.parser.expression.ProgramExpressionParams;
-import org.hisp.dhis.program.AnalyticsType;
+import java.util.HashSet;
+import java.util.Set;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import org.hisp.dhis.common.DimensionalItemId;
 
 /**
- * Program indicator variable: event date (also used for execution date)
+ * Information parsed from an expression
+ * <p>
+ * This is only information that is gathered from parsing the expression, and
+ * contains no information from any other source such as the database (either
+ * data or metadata). In other words, the same expression string will always
+ * result in the same information here.
  *
  * @author Jim Grace
  */
-public class vEventDate
-    extends ProgramDateVariable
+@Getter
+@Setter
+public class ExpressionInfo
 {
-    @Override
-    public Object getSql( CommonExpressionVisitor visitor )
-    {
-        ProgramExpressionParams progExParams = visitor.getProgExParams();
+    /**
+     * The dimensional item ids found.
+     */
+    private Set<DimensionalItemId> itemIds = new HashSet<>();
 
-        if ( AnalyticsType.ENROLLMENT == progExParams.getProgramIndicator().getAnalyticsType() )
-        {
-            return visitor.getStatementBuilder().getProgramIndicatorEventColumnSql(
-                null, "executiondate", progExParams.getReportingStartDate(), progExParams.getReportingEndDate(),
-                progExParams.getProgramIndicator() );
-        }
+    /**
+     * The sampled dimensional item ids found (for predictors).
+     */
+    private Set<DimensionalItemId> sampleItemIds = new HashSet<>();
 
-        return "executiondate";
-    }
+    /**
+     * Ids of org unit groups that will need org unit group member counts.
+     */
+    private Set<String> orgUnitGroupCountIds = new HashSet<>();
+
+    /**
+     * Ids of org unit groups found in orgUnits.groups function.
+     */
+    private Set<String> orgUnitGroupIds = new HashSet<>();
 }
--- a/dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/ProgramExpressionParams.java
+++ b/dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/ProgramExpressionParams.java
@@ -36,7 +36,7 @@ import lombok.Getter;
 import org.hisp.dhis.program.ProgramIndicator;
 
 /**
- * Parameters to generate SQL from a program expression
+ * Parameters to generate SQL from a program indicator expression
  *
  * @author Jim Grace
  */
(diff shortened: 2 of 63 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionInfo.java
```diff
@@ -0,0 +1,71 @@
...
+ * This is only information that is gathered from parsing the expression, and
+ * contains no information from any other source such as the database (either
+ * data or metadata). In other words, the same expression string will always
+ * result in the same information here.
+ *
+ * @author Jim Grace
+ */
+@Getter
+@Setter
+public class ExpressionInfo
```
jbee, line 50: I guess all fields can also be `final`?

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/ProgramExpressionParams.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.hisp.dhis.program.ProgramIndicator;
+
+/**
+ * Parameters to generate SQL from a program expression
+ *
+ * @author Jim Grace
+ */
+@Getter
+@Builder
```
jbee, line 44: You could think about `final` fields and using `toBuilder` to create new objects should you want to modify the params after they have been created initially.

jbee (approved):
I can't say I understand what the code does but its good you could simplify and shorten it quite a bit. On the code level this looks good. I hope it also make using this more straight forward but I cannot judge it.

Its good you try using short names. I think you can even drop the `ex`/`Ex` prefix in local fields and method names. In the context they are used this should be clear and if there is any doubt left one can always just look at the type to see that it is about expressions.

Left some comments go push little more in the "immutable"/"unmodifiable" direction. Not sure this is possible here - if it is no big change place think about it.

jbee (approved):

PR #9828, opened 2022-02-16
fix: Add not-null to ProgramStage.hbm to enable validation in API

stian-sandvold:
See title


dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/ProgramStage.hbm.xml
```diff
@@ -15,7 +15,7 @@
     </id>
     &identifiableProperties;
 
-    <property name="name" column="name" length="230" />
+    <property name="name" column="name" not-null="true" length="230" />
```
larshelge, line 18: Do we have/need a migration that verifies that there no program stages with null in the name column, so that this does not crash?

enricocolasante (approved):

teleivo (approved):

PR #9839, opened 2022-02-16
feat: data integrity check to find non-unique category combos [DHIS2-12639]

jbee:
### Summary
Adds a data integrity check that lists all category combos that do not have a combination of categories that is unique. This means there are "duplicates" combos that could be eliminated. 

In this context a combo is understood as a set, not a list. 

The YAML based check will also list the "duplicate" combos. Providing the references from YAML based checks wasn't implemented all the way so this needed some more groundwork. Now YAML based checks have examples in core that use the full range of the feature. 

### Automatic Testing
Only the existence is tested as the SQL used makes use of postgres specific operators and tests so far are H2 based.

### Manual Testing
Basic test workflow:

* start server and login
* `POST` http://localhost:8080/api/dataIntegrity/details?checks=categories_unique_category_combo (no body)
* `GET` http://localhost:8080/api/dataIntegrity/details?checks=categories_unique_category_combo&timeout=500
* check that you got the duplicates identified
* check also that the `refs` refer to the correct combo that is a duplicate

I think there is one combo with a duplicate in the demo database. This is one just having one category in the combo.
To test this more thoroughly goto maintenance app an add a category combination with two or more categories. 
First maybe just duplicate one that does exist with another name. Then test again after you have swapped the order of categories in the duplicate combo. You should still have it identified as duplicate.


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReader.java
```diff
@@ -147,8 +147,7 @@ private static String sanitiseSQL( String sql )
             .replaceAll( "select '[^']+' as [^,]+,", "select " )
             .replaceAll( "'[^']+' as description", "" )
             .replace( "::varchar", "" )
-            .replace( "|| '%'", "" )
-            .replaceAll( " as [^,]+,", "," ) );
```
jbee, line 151: FYI: This was placed to avoid issues with SQL when used with H2 but it was way to aggressive and cut out sections from SQL that were required when used with postgres.

vietnguyen (approved):

netroms (approved):

PR #9844, opened 2022-02-17
fix: system settings JSON as default [DHIS2-12516]

jbee:
### Summary
The problem here was that the endpoint path was mapped twice and depending on what a browser would send by default in its `Accept` header one or the other would win. I think the JSON returning one was given the `produces` with HTML for that reason so that it would win. But it  never actually return HTML - always JSON. 

After discussion in Slack we decided to remove the JSONP endpoint resolving the ambiguity.

### Automatic Testing
Does cover the method, did not need adjustment.

### Manual Testing
* `GET /api/systemSettings` without setting any headers in a browser should get you JSON
* `GET /api/systemSettings` without setting any `Accept` header at all in a REST tool should get you JSON

### Release Notes Update
https://github.com/dhis2/dhis2-releases/pull/72

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/SystemSettingController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/SystemSettingController.java
@@ -266,7 +266,7 @@ public class SystemSettingController
         return Optional.empty();
     }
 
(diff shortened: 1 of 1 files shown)


netroms (approved):

vietnguyen (approved):

jbee:
We are going to remove JSONP

PR #9846, opened 2022-02-17
feat: Mark legacy eventVisualization as NEW after PUT [DHIS2-12647]

maikelarabori:
When an event report created in the old EventReport app (on `/eventReports`) is opened, modified and saved in the new app (it means PUT request to /eventVisualizations) it should be marked as incompatible with the old app and no longer show up on `/eventReports`.

This is achieved by setting the `legacy` flag to false on every PUT operation that is triggered through the new endpoint `/eventVisualizations`.

I also added a migration to remove all old FKs from the legacy/unused tables so they do not conflict with the new set of tables of EventVisualization.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_33__Drop_Legacy_Constraints_EventReport_EventChart.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_33__Drop_Legacy_Constraints_EventReport_EventChart.sql
@@ -1,0 +1,60 @@
+-- This script relates to the issue https://jira.dhis2.org/browse/DHIS2-12647
+-- Removes all foreign keys/constraints from legacy (unused) tables related to "event chart" and "event report".
+-- This is causing collection issues (during updated) on the new set of tables for "event visualization".
+
+-- 1) Removing all FKs from all "eventchart*" tables
+DO
+$$
+    DECLARE
+        t      text;
+        tables text[] := array ['eventchart','eventchart_attributedimensions','eventchart_categorydimensions',
+            'eventchart_categoryoptiongroupsetdimensions','eventchart_columns','eventchart_dataelementdimensions',
+            'eventchart_filters','eventchart_itemorgunitgroups','eventchart_organisationunits',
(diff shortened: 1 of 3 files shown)


d-bernat (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_33__Drop_Legacy_Constraints_EventReport_EventChart.sql
```diff
@@ -0,0 +1,60 @@
...
+            'eventreport_programindicatordimensions','eventreport_rows'];
+        r      record;
+    BEGIN
+        FOREACH t IN ARRAY tables
+            LOOP
+                FOR r IN (
+                    SELECT constraint_name
+                    FROM information_schema.table_constraints
+                    WHERE table_name = t
+                      AND constraint_name LIKE 'fk%'
```
gnespolino, line 52: same here

gnespolino (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_33__Drop_Legacy_Constraints_EventReport_EventChart.sql
```diff
@@ -0,0 +1,60 @@
...
+            'eventchart_programindicatordimensions','eventchart_rows'];
+        r      record;
+    BEGIN
+        FOREACH t IN ARRAY tables
+            LOOP
+                FOR r IN (
+                    SELECT constraint_name
+                    FROM information_schema.table_constraints
+                    WHERE table_name = t
+                      AND constraint_name LIKE 'fk%'
```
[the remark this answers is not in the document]
maikelarabori: Good catch, done

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_33__Drop_Legacy_Constraints_EventReport_EventChart.sql
maikelarabori: Good catch, done

PR #9855, opened 2022-02-20
fix: add assignedUserUsername and DisplayName to event DHIS2-12620

teleivo:
@stian-sandvold the fields where not exposed in `/tracker/events/a6f092d0d44` since we did not have them in our tracker domain models. This change exposes them. An open question I have is: should the fields be exposed on import as well? If yes, this change is ready. If not, I will need to make a few more changes 😄

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/domain/Event.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/domain/Event.java
@@ -127,6 +127,12 @@ public class Event
     private String assignedUser;
 
     @JsonProperty
+    private String assignedUserUsername;
+
+    @JsonProperty
+    private String assignedUserDisplayName;
+
+    @JsonProperty
     private String createdBy;
 
     @JsonProperty


stian-sandvold:
> @stian-sandvold the fields where not exposed in `/tracker/events/a6f092d0d44` since we did not have them in our tracker domain models. This change exposes them. An open question I have is: should the fields be exposed on import as well? If yes, this change is ready. If not, I will need to make a few more changes 😄

Not sure what it would mean to have these exposed on import? But users can not assign a user using either of these two fields. If a user does send in this info, we should just ignore it, since its not properties that would impact the import.

teleivo:
> > @stian-sandvold the fields where not exposed in `/tracker/events/a6f092d0d44` since we did not have them in our tracker domain models. This change exposes them. An open question I have is: should the fields be exposed on import as well? If yes, this change is ready. If not, I will need to make a few more changes smile
> 
> Not sure what it would mean to have these exposed on import? But users can not assign a user using either of these two fields. If a user does send in this info, we should just ignore it, since its not properties that would impact the import.

What I mean by exposing them on import is that we currently use the same model for import and export. So every JSONProperty we have in our models will at least be deserialized on import AFAIK. So users can send it in their JSON payload and it will make its way into tracker. The values in these fields might not be taken into account in our business logic or be "lost" since we use mappers to translate from the tracker domain model to the one used 
in program rules or persisting. Does that make sense?

What I am used to is a separation of models (import/export), so that you never risk user from storing/influencing anything they shouldn't. That's why I brought it up. Maybe our mappers are playing the role of what I am used to with 2 separate models. Just less explicit. There is also still a risk since by default

https://mapstruct.org/documentation/stable/reference/html/#basic-mappings
| When a property has the same name as its target entity counterpart, it will be mapped implicitly.

unless we define `@BeanMapping(ignoreByDefault = true)` which we do not always.

I will check to ensure these fields will not influence the import. I can bring this up in a design meeting later on. I think in https://jira.dhis2.org/browse/DHIS2-12563 I might need to split the models anyway.

teleivo:
closing this one as requirements are being discussed

PR #9857, opened 2022-02-21
fix: consistent and unsurprising query param behavior tracker/relationships DHIS2-12659

teleivo:
# What

* add query parameter `trackedEntity` (keeping parameter `tei` for backwards compatibility) to adhere to the new tracker APIs consistent naming
* fail if more than one of the mutually exclusive parameters (`trackedEntity`, `tei`, `enrollment`, `event`) is given as described in the API docs

No parameters leads to response

```json
{
  "httpStatus": "Bad Request",
  "httpStatusCode": 400,
  "status": "ERROR",
  "message": "Missing required parameter 'trackedEntity', 'enrollment' or 'event'."
}
```

More than one parameter leads to response

```json
{
  "httpStatus": "Bad Request",
  "httpStatusCode": 400,
  "status": "ERROR",
  "message": "Only one of parameters 'trackedEntity', 'enrollment' or 'event' is allowed."
}
```

# How

Since we already have code dealing with `trackedEntity`, `enrollment`, `event` in a generic way, I wanted to provide a single getter for the identifier to be fetched from the DB. No matter from which query param it originated from. 

At first I tried to make sure an invalid (query param count != 1) TrackerRelationshipCriteria could not be created by
* throwing `WebMessageException` in setters when more than one of the mutually exclusive query parameters were set
* trying `javax.validation` annotation on  TrackerRelationshipCriteria with an `@Valid` annotation on the controller method parameter `TrackerRelationshipCriteria`

These approaches did not work since they lead to a HTTP status 500 with stacktraces in the message, instead of the desired response messages shown above.

Would like to get your opinion @jbee on whether implementing it this way makes sense to you or if you would have a different approach in mind 😋

### DIFF
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerRelationshipsExportControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerRelationshipsExportControllerTest.java
@@ -119,6 +119,42 @@ class TrackerRelationshipsExportControllerTest extends DhisControllerConvenience
     }
 
     @Test
+    void getRelationshipsBadRequestWithAllParams()
+    {
+        assertEquals( "Only one of parameters 'tei', 'enrollment' or 'event' is allowed.",
+            GET( "/tracker/relationships?tei=Hq3Kc6HK4OZ&enrollment=Hq3Kc6HK4OZ&event=Hq3Kc6HK4OZ" )
+                .error( HttpStatus.BAD_REQUEST )
+                .getMessage() );
+    }
+
+    @Test
+    void getRelationshipsBadRequestWithTeiAndEnrollment()
+    {
+        assertEquals( "Only one of parameters 'tei', 'enrollment' or 'event' is allowed.",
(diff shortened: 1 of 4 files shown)


jbee (approved):
This certainly is one possible way of doing this. Main benefits I see is quality of the error message  that we can provide and that this constraint is enforced more towards the service level. Which also is my criticism here that this kind of mixes API/controller with something that otherwise would perfectly suit to become service level API thing. 

Just to give you another idea of what you could do: spring does allow to map to different handler methods based on the presence of parameters. You could use this to map your query to 4 different handler methods each just accepting one of the 4 mutual exclusive parameters. Then they build the criteria from that and so on. This moves the validation into the spring layer but the error messages will certainly not be as clear and testing on more service oriented levels will not also check and benefit from this limitation. 

Overall I think complex query parameter validation should be a service level task. We do have this in several places where the service has a dedicated method like
```java
? validate(SomethingQuery query);

List<Something> list(SomethingQuery query);
```
The big question becomes how are errors indicated by the `validate` method. Some throw exceptions, some return reports or lists or errors. Usually in those places this builds on `ErrorMessage` as the item for a single error. This then would need some translation to a `WebMessage` where this becomes the `WebMessageResponse` and you decide the response code based on the validation outcome.

PR #9863, opened 2022-02-21
fix: NPE in item options [23.8-DHIS2-12674]

d-bernat:
just small NPE in item options

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
@@ -318,7 +318,7 @@ public abstract class AbstractAnalyticsService
                     .getOptionSetObject()
                     .getOptions()
                     .stream()
-                    .filter( opt -> grid.getRows().stream().anyMatch( r -> {
+                    .filter( opt -> opt != null && grid.getRows().stream().anyMatch( r -> {
                         Object o = r.get( columnIndex );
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
```diff
@@ -318,7 +318,7 @@ protected void addMetadata( EventQueryParams params, List<DimensionItemKeywords.
                     .getOptionSetObject()
                     .getOptions()
                     .stream()
-                    .filter( opt -> grid.getRows().stream().anyMatch( r -> {
+                    .filter( opt -> opt != null && grid.getRows().stream().anyMatch( r -> {
                         Object o = r.get( columnIndex );
                         if ( o instanceof String )
                         {
```
maikelarabori, line 324: In the expression below `((String) o).equalsIgnoreCase( opt.getCode() )`, should we also check is if `opt.getCode()` is not null? I believe that the `code` is not a mandatory attribute.

maikelarabori (approved):
Thanks @d-bernat, just one minor comment. :+1:

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
d-bernat: cleared

gnespolino (approved):

PR #9864, opened 2022-02-21
refactor: Centralize tracker value type validation

abyot:
- value type validations are now using a common service
- option set value types (option codes) must exact match
- testing FILE_RESOURCE valueType for just valueType doesn't make sense. Tests are now updated to check if the mentioned fileResource exists or not .... then the check becomes fileResource exists or not.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHook.java
```diff
@@ -111,20 +105,21 @@ private void validateMandatoryDataValues( Event event, TrackerImportValidationCo
...
+            dataValue.getValue() );
 
-        if ( status != null )
+        if ( errorMessage != null )
         {
-            reporter.addError( event, TrackerErrorCode.E1302, dataElement.getUid(),
-                status );
+            reporter.addError( event, getTrackerErrorCode( errorMessage ),
+                dataElement != null ? dataElement.getUid() : null,
+                errorMessage.getMessage() );
```
teleivo, line 118: It looks like you are passing the error message as an argument to the report (not only here). This would mean the string value of the error message gets interpolated as an arg in the tracker error message. Can you please check if the error message is rendered as you would expect it.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHook.java
```diff
@@ -172,24 +167,28 @@ private void validateDataValueDataElementIsConnectedToProgramStage( ValidationEr
...
-        {
-            return;
-        }
-
-        boolean isFile = dataElement.getValueType() != null && dataElement.getValueType().isFile();
-        if ( !isFile )
+        switch ( errorMessage.getErrorCode() )
         {
-            return;
+        case E2027:
```
teleivo, line 174: if these are the same mappings we've already declared you could add an EnumMap in a shared place for these hooks to use. If not, just ignore this comment 😋

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/TeTaValidationTest.java
```diff
@@ -230,11 +230,24 @@ void testTeaInvalidImage()
         TrackerImportParams trackerImportParams = createBundleFromJson(
             "tracker/validations/te-program_with_tea_invalid_image_value.json" );
         TrackerImportReport trackerImportReport = trackerImportService.importTracker( trackerImportParams );
-        assertEquals( 2, trackerImportReport.getValidationReport().getErrors().size() );
+        assertEquals( 1, trackerImportReport.getValidationReport().getErrors().size() );
         assertThat( trackerImportReport.getValidationReport().getErrors(),
-            hasItem( hasProperty( "errorCode", equalTo( TrackerErrorCode.E1085 ) ) ) );
```
teleivo, line 235: Does the import payload still have 2 errors? If so (only if so 😋 ) you changed behavior away from what I understood the goal of NTI validation was. To provide all errors we can so the user can fix all errors at once, then try the import again and expect it to work. @stian-sandvold is that the goal of NTI validation 😄 ?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHookTest.java
```diff
@@ -693,15 +703,14 @@ void failValidationWhenDataElementValueTypeIsInvalid()
...
-        runAndAssertValidationForDataValue( ValueType.FILE_RESOURCE, "not_valid_uid" );
     }
 
     @Test
     void successValidationDataElementOptionValueIsValid()
     {
         setUpIdentifiers();
 
-        DataValue validDataValue = dataValue( "code" );
+        DataValue validDataValue = dataValue( "CODE" );
```
teleivo, line 713: why does that matter? is this value always uppercase?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHookTest.java
```diff
@@ -375,7 +384,7 @@ void failValidationWhenDataElementValueTypeIsNull()
         hook.validateEvent( reporter, event );
 
         assertThat( reporter.getReportList(), hasSize( 1 ) );
-        assertEquals( TrackerErrorCode.E1302, reporter.getReportList().get( 0 ).getErrorCode() );
```
teleivo, line 378: was this the wrong error code?

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/validations/events-with_valid_option_value.json
```diff
@@ -22,7 +22,7 @@
           "storedBy": "admin",
           "createdAt": "2019-01-28T12:10:38.113",
           "dataElement": "LjQPcfuEI6S",
-          "value": "male",
```
teleivo, line 25: was this value previously invalid?

dhis-2/dhis-web/dhis-web-apps/.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules/
 node/
+/bin/
```
teleivo, line 3: was this intentional?

teleivo (commented):
Looks well 😄

I would rename, move a few things as commented.

There are 2 things I would like to understand before giving my ✅ 😋 

1. You changed some of the error codes, removed some error codes in tests and changed some of the test data to all uppercase, pascal case. Since this PR has the title refactor I would not expect behavior to change. So no tests, no test data to change. Is it more than a refactoring? Did we not validate some things properly before?

2. AFAIK the NTI validation should be greedy. Return all errors (unless validation mode is FAIL_FAST). The intention being that users can fix all errors in one go, import again and things work. I made a comment on one of the tests and cc'ed  @stian-sandvold since it seems to me that this change might go against that "design goal".

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/AttributeValidationHook.java
```diff
@@ -160,4 +151,29 @@ protected void validateAttributeUniqueness( ValidationErrorReporter errorReporte
...
+            return TrackerErrorCode.E1106;
+        case E2043:
+            return TrackerErrorCode.E1304;
+        case E2026:
+            return TrackerErrorCode.E1009;
+        default:
+            return null;
+        }
+    }
+
```
enricocolasante, line 178: I have some concerns about this part.
`ValueTypeValidationUtils` is validating the value type and returning some error codes if something is wrong.
These error codes only really make sense for the old tracker importer, to be used in the new tracker importer we need to convert them.
First conceptual problem is that we are coupling the new importer to the old one throw these codes.
Second problem is that if one day we remove the old importer, we will have a validator that returns error codes that are not really used by any code in the system, but they still need to be converted to be used in the NTI.

If we plan at any point in the future to remove the old importer we can either duplicate the code for this validator, so when the old importer is gone one of the duplicates is gone or we can abstract the validator to not really know about the codes and then give the old and new importer the responsibility to use the right code based on the error reported by the validator

teleivo: Very good point 👍🏻 

If ValueTypeValidationUtils is only used by new and old tracker than duplicating it with error codes appropriate to each package/domain seems easy and clean.

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/validations/events-with_valid_option_value.json
abyot: no it was considered valid. but last time we decided to not use equalsIgnoreCase.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHookTest.java
teleivo: The title of the PR is refactor, so I would not expect test data or expectations to change since behavior should be preserved in a refactoring. That is why I am asking 😋 So previously `code` was deemed valid which it should not have been since the value has to match the metadata exactly?

PR #9865, opened 2022-02-21
chore: Clean up service core module [TECH-936]

enricocolasante:


dhis-2/dhis-services/dhis-service-core/pom.xml
```diff
@@ -260,7 +259,7 @@
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter</artifactId>
```
teleivo, line 263: My gut feeling is that we should not change this to junit-jupiter-api. https://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven-engines-configure also states dependency on artifact `junit-jupiter`. It says

> To configure support for JUnit Jupiter based tests, configure test scoped dependencies on the JUnit Jupiter API and the JUnit Jupiter TestEngine implementation similar to the following.

AFAIK API is for the things like the `@Test` and assertions, the engine is then the thing running the tests. I suspect that if you only have `junit-jupiter-api` left in DHIS2, tests will not run. See https://maven.apache.org/surefire/maven-surefire-plugin/usage.html `Using Different Testing Providers` Surefire knows we use JUnit 5 also because the `junit-jupiter-engine` dependency is in our project.

I think we should tell maven-dependency-plugin to ignore it being unused instead https://github.com/dhis2/dhis2-core/blob/7187c6e2c5c6089ef6c49c90ae2cdfdbf09f3523/dhis-2/pom.xml#L712

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/resolver/CategoryOptionGroupResolverTest.java
```diff
@@ -31,9 +31,9 @@
 import static org.hisp.dhis.DhisConvenienceTest.createCategoryOptionGroup;
 import static org.hisp.dhis.expression.ParseType.INDICATOR_EXPRESSION;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
-import static org.wildfly.common.Assert.assertTrue;
```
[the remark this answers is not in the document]
enricocolasante: 👍

dhis-2/dhis-services/dhis-service-core/pom.xml
enricocolasante: I agree

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/system/SystemUpdateService.java
```diff
@@ -50,9 +50,9 @@
 import org.hisp.dhis.user.UserCredentials;
 import org.hisp.dhis.user.UserCredentialsStore;
 import org.hisp.dhis.user.UserService;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 
-import com.google.api.client.http.HttpStatusCodes;
```
teleivo, line 55: should this one also be banned?

teleivo (approved):
🎉

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/system/SystemUpdateService.java
jbee: As spring has something like it I think so yes. If you cannot use the spring one this is a good indication that the code is in the wrong module doing HTTP stuff in a non web facing module.

enricocolasante: We should definitely do something about it but should we ban it or exclude it?

I think it make sense to ban the JUnit4 dependency because we cannot really remove it, but it this case (and also the wildfly-common) maybe we can just remove it. In both cases they just come from one dependency in the pom.
What do you think? @teleivo @jbee

teleivo: So my guess is the package comes from dependency https://developers.google.com/api-client-library/java
If we are interacting with google, removing this client library is probably not a good or convenient option. We might still be able to ban these imports. @jbee could there be a module in this repo where using these codes would be ok? If so we shouldn't ban these imports in every module.

jbee: This would need some investigation into how far we use the library and what for. I am not aware of any usage so I'd hope its just a few smaller cases.

enricocolasante: My suggestion was to just remove `com.google.http-client:google-http-client-apache-v2` that is part of the `api-client-library`. We are not using it anywhere.
In the end banning or excluding a specific library has the same effect, the only difference is that banning will appear when compiling with maven while if we exclude the (sub)dependency it would not be possible to add it while coding

teleivo: that sounds good to me then 😄

PR #9867, opened 2022-02-22
fix: Return dates in NTI report with Iso8601NoTz format [DHIS2-11947]

enricocolasante:
[DHIS2-11947](https://jira.dhis2.org/browse/DHIS2-11947) couldn't be reproduce as the date was always present but the format was not the same that it is used in the payload

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerReportUtils.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerReportUtils.java
@@ -40,6 +40,7 @@ import org.hisp.dhis.tracker.bundle.TrackerBundle;
 import org.hisp.dhis.tracker.domain.Enrollment;
 import org.hisp.dhis.tracker.domain.Event;
 import org.hisp.dhis.tracker.domain.TrackedEntity;
+import org.hisp.dhis.util.DateUtils;
 import org.hisp.dhis.util.ObjectUtils;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerReportUtils.java
```diff
@@ -75,7 +76,7 @@ else if ( Date.class.isAssignableFrom( argument.getClass() ) )
         }
         else if ( Instant.class.isAssignableFrom( argument.getClass() ) )
         {
-            return DateFormat.getInstance().format( Date.from( (Instant) argument ) );
+            return DateUtils.getIso8601NoTz( DateUtils.fromInstant( (Instant) argument ) );
```
teleivo, line 79: why did you switch to `DateUtils.fromInstant`? `Date.from` seems to have the same signature

teleivo (approved):
1 minor comment

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerReportUtils.java
enricocolasante: Just because `DateUtils.fromInstant` it is null safe and it is exactly how we do it in the serializer `WriteInstantStdSerializer`

larshelge (approved):

PR #9870, opened 2022-02-22
fix: prevent category combination deletion when predictor is linked [DHIS2-10687]

jbee:
### Summary
Also prevent CC deletion when there is a linked COC linked to a predictor.

### Manual Testing
1. create a new CC
2. create a new DE linked to the CC
3. create a new predictor linked to the new DE and also select a _output category option combo_ after selecting the _output data element_
4. try to delete the CC => should point out it is linked to a predictor
5. delete the predictor
6. try to delete the CC => now should be possible to delete

Rerun the above test but without selecting anything in the predictors _output category option combo_ field - the attempt to delete the CC in step 4 should now succeed directly.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/predictor/PredictorDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/predictor/PredictorDeletionHandler.java
@@ -34,6 +34,7 @@ import java.util.List;
 
 import lombok.AllArgsConstructor;
 
+import org.hisp.dhis.category.CategoryCombo;
 import org.hisp.dhis.category.CategoryOptionCombo;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.expression.Expression;
@@ -60,6 +61,7 @@ public class PredictorDeletionHandler extends DeletionHandler
         whenDeleting( PredictorGroup.class, this::deletePredictorGroup );
         whenVetoing( DataElement.class, this::allowDeleteDataElement );
         whenVetoing( CategoryOptionCombo.class, this::allowDeleteCategoryOptionCombo );
+        whenVetoing( CategoryCombo.class, this::allowDeleteCategoryCombo );
     }
 
     private void deleteExpression( Expression expression )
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/predictor/PredictorDeletionHandler.java
```diff
@@ -111,6 +113,15 @@ private DeletionVeto allowDeleteCategoryOptionCombo( CategoryOptionCombo optionC
         return vetoIfExists( "SELECT COUNT(*) FROM predictor where generatoroutputcombo=" + optionCombo.getId() );
     }
 
+    private DeletionVeto allowDeleteCategoryCombo( CategoryCombo categoryCombo )
+    {
+        return vetoIfExists( "SELECT COUNT(*) FROM predictor p where exists ("
```
larshelge, line 118: I am wondering if we are not fixing the root cause here. It should be enough to check for the category option combo of the predictor. 

The question is why PredictorDeletionHandler#allowDeleteCategoryOptionCombo is not triggered. This check should be triggered when the delete of category option combo is attempted.

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/predictor/PredictorDeletionHandler.java
jbee: Yeah, we talked about this. It should have worked without this extra check but due to order or manner of how this is processed it does not work as we want it to. This could be considered a workaround. In general I think we could need something more centralised to deal with the blocking or unlinking on deletion. While the deletion handler concept is nicely composable there appear to be difficulties arising from the looseness of the concept that for example gives little to no control to order or combinations.

vietnguyen: Just a guess, but i think `PredictorDeletionHandler#allowDeleteCategoryOptionCombo` is not triggered because the deletion of `CategoryOptionCombo` occurs in this `updateNoAcl` method in class `CategoryComboDeletionHandler`

```java
private void deleteCategoryOptionCombo( CategoryOptionCombo categoryOptionCombo )
    {
        for ( CategoryCombo categoryCombo : categoryService.getAllCategoryCombos() )
        {
            categoryCombo.getOptionCombos().remove( categoryOptionCombo );
            idObjectManager.updateNoAcl( categoryCombo );
        }
    }
```

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/predictor/PredictorDeletionHandler.java
larshelge: I think this is not actually where the option combo is deleted @vietnguyen , I think this is where we remove the association from category combo before deleting the option combo.

PR #9871, opened 2022-02-22
New syntax of repeatable stages in headers [2.38-DHIS2-12683]

d-bernat:
There is bug in headers, the separator had t be changed, now it is ~. Comma is not possible to use, cause we are using it for options, headers, etc separator.


gnespolino (approved):

maikelarabori (approved):
Thanks, @d-bernat :+1:
I believe, later, we can revisit the separator thing and adjust all of them (in general) so we can have a more uniform solution.

PR #9877, opened 2022-02-22
feat: make api geoFeatures to support GeoJsonObject

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12654

1. This PR add a new parameter `coordinateField` to the controller `GeoFeatureController`
- If the `coordinateField` is not provided, then coordinates will be retrieved from `geometry` property of the OrganisationUnits.
- If the `coordinateField` parameter is present, then coordinates will be retrieved from GeoJson Attribute value of the OrganisationUnits. If the attribute value is null, then it will fallback to the `geometry` property value.

2. I have moved the business logics code from `GeoFeatureController` to new service class `GeoFeatureService`
3. Also updated the validation for GeoJsonObject import, add new error code for the cases of unsupported GeoJson type.


jbee (approved):
:+1:

mortenoh (approved):

PR #9879, opened 2022-02-23
fix: Fix broken user update

netroms:
**Summary:**

* Fixes broken user update when using the backward compatible user.userCredentials format.
* Also fixes a bug resulting in not being able to remove or update user groups on users.


larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -233,6 +231,10 @@
 
     private FileResource avatar;
 
+    // Backward comp. field, will be removed when front-end has converted to new
+    // User model
+    private transient UserCredWrapperDto userCredentialsRaw;
```
jbee, line 236: Maybe bit off topic but I saw this now: Could we use a less generic term name. Feel like either `Dto` or `Wrapper` would do. Maybe also go back to spelling out `Credentials` fully. I'd find `UserCredentialsDto` or UserCredentialsWrapper` would communicate better.

jbee (approved):

PR #9885, opened 2022-02-23
chore: Clean up service program rule module [TECH-938]

enricocolasante:
I merged a previous PR and I forgot to exclude the dependencies that we talked about so I am doing it in this PR.

I managed to completely exclude `org.wildfly.common:wildfly-common`.
It is a utility library that we shouldn't use, so it make sense that is not present in the build.

I tried to do the same with `com.google.http-client:google-http-client` because we use some utils from this library as well but it turns out that we cannot really exclude it as it is in the Authentication process.
So in this case I just banned `com.google.api.client.util.**` imports

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EnrollmentTrackerConverterService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EnrollmentTrackerConverterService.java
@@ -27,7 +27,7 @@
  */
 package org.hisp.dhis.tracker.converter;
 
-import static com.google.api.client.util.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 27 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EnrollmentTrackerConverterService.java
```diff
@@ -27,7 +27,7 @@
  */
 package org.hisp.dhis.tracker.converter;
 
-import static com.google.api.client.util.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkNotNull;
```
jbee, line 30: just as a note for the future: I think we can use `java.util.Objects` for this - it has `requireNonNull`.

jbee (approved):
Thanks for cleaning this up.

larshelge (approved):

PR #9895, opened 2022-02-24
feat: OrgUnit.dataSet and program exp. DHIS2-12082

jimgrace:
See [DHIS2-12082](https://jira.dhis2.org/browse/DHIS2-12082). This PR implements expression functions `orgUnit.dataSet()` and `orgUnit.program()` that can be used in validation rule and predictor expressions.

It adds the classes `FunctionOrgUnitDataSet` and `FunctionOrgUnitProgram` to process the parsed functions. When called with their method `getExpressionInfo`, these add the parsed UIDs to new `ExpressionInfo` fields `orgUnitDataSetIds` and `orgUnitProgramIds`, respectively.

### Validation rule and predictor refactoring -- Former behavior

Formerly, `DefaultValidationService`:

1. called `expressionService.getExpressionDimensionalItemIds` to get the left-side and right-side `DimensionalItemIds` (to identify `DataElements`, etc.), and collected all the left and right side Ids together before calling `dimensionService.getNoAclDataDimensionalItemObjectMap` to resolve them all at once into `DimensionalItemObjects`. (They are resolved all at once because this performs much better.)

2. called `expressionService.getExpressionOrgUnitGroupIds` for each left and right side expression (causing `DefaultExpressionService` to make a second pass through the parse tree of each expression) to collect all the referenced `orgUnitGroup` Ids (if any), and then called `idObjectManager.getNoAcl` to resolve them all at once to the `OrganisationUnitGroup` objects (again, all at once for better performance).

Formerly, `DefaultPredictionService`:

1. Unlike `DefaultValidationService`, `DefaultPredictionService` called `expressionService.getExpressionDimensionalItemMaps` for the generator expression, causing `DefaultExpressionService` to walk the expression parse tree to gather both the `itemIds` (current period for a predictor) as well as the `sampledItemIds` (inside aggregation functions, sampled past periods for a predictor). Then the `DefaultExpressionService` called `dimensionService.getDataDimensionalItemObjectMap` to generate the id->object maps for both the `itemIds` and `sampleItemIds` (in two separate calls). Next `DefaultPredictionService` called `expressionService.getExpressionDimensionalItemMaps` again for the skip test expression (if any), for which `DefaultExpressionService` would walk the parse tree for the skip test, and call `dimensionService.getDataDimensionalItemObjectMap` again.

2. More like `DefaultValidationService`, `DefaultPredictionService` called `expressionService.getExpressionOrgUnitGroupIds` for the generator expression, and also for the skip test expression if any (causing `DefaultExpressionService` to make another pass through the parse tree of each expression) to collect all the referenced `orgUnitGroup` Ids (if any), and then called `organisationUnitGroupService.getOrganisationUnitGroup` to resolved them all at once to the `OrganisationUnitGroup` objects (again, all at once for performance reasons).

### Validation rule and predictor refactoring -- New behavior

To support the new `orgUnit.dataSet()` and `orgUnit.program()` functions, this requires new id -> object maps to be generated for both `DataSets` and `Programs` if/when these functions are present. If we followed the old way of doing things, then `ValidationService` and `PredictionService` would both have to collect not only `itemIds`, and `orgUnitGroupIds`, but also `dataSetIds` and `programIds` (calling `ExpressionService` for four different passes through the parse tree), and then each of these services would need to generate maps for all four of these object types. This would put additional burden on `ValidationService` and `PredictionService` and cause four passes through the parse tree for every expression (which in ANTLR takes a measurable amount of CPU).

Instead, `DefaultValidationService`, `DefaultPredictorService`, and `DefaultExpressionService` have been refactored in the way they collect these Ids and generate the object maps, putting less burden than before on the validation and predictor services, and allowing the expression service to do most of the work. Much of the groundwork for this refactor was achieved when `ExpressionParameters` was introduced in [pull/9778](https://github.com/dhis2/dhis2-core/pull/9778) and `ExpressionInfo` was introduced in [pull/9820](https://github.com/dhis2/dhis2-core/pull/9820). Both of these PRs were done with the requirements of this PR in mind. Here's how it now works:

1. `ExpressionService.getExpressionInfo` has now been made public, so the validation and predictor services can call it just once to collect all the Ids in a single pass through the parsed tree for each expression: `itemIds` (and `sampleItemIds`), `orgUnitGroupIds`, `dataSetIds`, and `programIds`.

2. `ExpressionService.getExpressionInfo` calls can be "chained" by passing the `ExpressionInfo` object returned from one call as input to the next (by including the `ExpressionInfo` reference in `ExpressionParams`). So the Ids of each type can accumulate through multiple calls in order to get the collection of all Ids in a collection of expressions.

3. A new `ExpressionService` method `ExpressionParams getBaseExpressionParams( ExpressionInfo info )` has been added that can take an `ExpressionInfo` (possibly accumulated to contain the ids from many expressions), and make a single call to `dimensionService.getDataDimensionalItemObjectMap` for all the `itemIds` (both `itemIds` and predictor `sampleItemIds` in one call), and a single call to `idObjectManager.getNoAcl` for each of the other id types (`OrganisationUnitGroup`, `DataSet`, and `Program`, only if present in the expressions). The resulting id -> object maps are packaged in a "base" `ExpressionParams` object, which the validation and predictor services can subsequently use when finding the value of particular expressions with particular data.

This removes the burden from the validation and predictor services of calling `dimensionService` and `idObjectManager` to create these id -> object maps, puts all these calls in one place, and reduces to 1 the number of times the parse tree must be traversed for each expression to get the information required.

`DefaultValidationService` needs to be particular about the `itemIds` it collects. A validation run can include validation rules of various period types, and data to evaluate all the validation rules of each period type (for a set of organisation units) is fetched in a single query. So after each chained call to `expressionService.getExpressionInfo`, `DefaultValidationService` extracts the `itemIds`, saves them according to the `periodType` of the validation rule, and initializes a new, empty set of `itemIds` for parsing the next expression. The other collections in `ExpressionInfo` just accumulate from expression to expression. After all the expressions have been examined, the entire set of `itemIds` is put back into `ExpressionInfo` and the call is made to `getBaseExpressionParams` to get a base `ExpressionParams` with the maps for `DimensionalItemObject`, `OrganisationUnitGroup`, `DataSet`, and `Program`.

`DefaultPredictionService` is simpler, since it only process a predictor generator expression and an optional skip test expression, and both use the same `periodType`. The prediction service just calls `getExpressionInfo` and chains the results from the generator expression with the results from the skip test expression, to generate a base `ExpressionParams` object that can be used for evaluating both.

### Other refactors

`DefaultValidationService`, `ValidationRunContext`, and `DataValidationTask` were all refactored for their revised interactions with `ExpressionService`.

`DefaultPredictionService.predict` was refactored to reduce the complexity as reported by SonarLint. The `processPrediction` and `rememberPredictedValue` methods were added to break up the logic.

### Testing

Unit tests for `ExpressionService`, `ValidationService`, and `PredictionService` were all updated for the new features. The suggested test procedures that are written up in [DHIS2-12082](https://jira.dhis2.org/browse/DHIS2-12082) were all run successfully.


dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/expression/ExpressionService2Test.java
```diff
@@ -198,6 +205,14 @@
...
     private OrganisationUnitGroup groupB;
 
+    DataSet dataSetA;
+
+    DataSet dataSetB;
+
+    Program programA;
+
+    Program programB;
+
```
jbee, line 215: `private`?

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionDataValueFetcher.java
```diff
@@ -171,8 +171,8 @@ public void init(
 
         orgUnitLookup = orgUnits.stream().collect( Collectors.toMap( OrganisationUnit::getPath, ou -> ou ) );
         dataElementLookup = dataElements.stream().collect( Collectors.toMap( DataElement::getId, de -> de ) );
-        dataElementLookup.putAll( dataElementOperands.stream().collect(
-            Collectors.toMap( d -> d.getDataElement().getId(), DataElementOperand::getDataElement ) ) );
+        dataElementLookup.putAll( dataElementOperands.stream().map( DataElementOperand::getDataElement )
+            .distinct().collect( Collectors.toMap( DataElement::getId, d -> d ) ) );
```
jbee, line 175: There is `Function::identity` for `d -> d`

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EnrollmentDateValidationHookTest.java
```diff
@@ -111,7 +111,7 @@ void testDatesMustNotBeInTheFuture()
         hasTrackerError( reporter, E1021, ENROLLMENT, enrollment.getUid() );
     }
 
-    @Test
+    // @Test
```
jbee, line 114: ?

jbee (commented):
Some small comments...

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/expression/ExpressionService2Test.java
jimgrace: Yup, thanks.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionDataValueFetcher.java
jimgrace: I knew there was something, just couldn't remember what. :)

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/expression/ExpressionService2Test.java
jimgrace: Oops, yup, thanks.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionDataValueFetcher.java
jimgrace: I knew there was something, just couldn't remember what. Thanks.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EnrollmentDateValidationHookTest.java
jimgrace: Ha! It's a test that fails when you're a developer testing your code and you're 5 hours west of UTC and it's within 5 hours of midnight, which frequently happens for me. I always have to disable this test to get all the other tests to run, and except for this time, I always remember to put it back. It's on my personal list to fix when my 2.38 features are done, since I am well situated ;) to test the fix.

jbee (approved):

enricocolasante (approved):

PR #9917, opened 2022-02-28
fix: support locales with 4 letter variants

mortenoh:
## Summary

* Adds support for locale files that has variants between 2 and 4 letters (including lower and uppercase)

[DHIS2-10427](https://jira.dhis2.org/browse/DHIS2-10427)


larshelge:
Closing due to inactivity, please re-open later.

PR #9918, opened 2022-02-28
feat: support csv for metadata

mortenoh:
## Summary

* Updates `AbstractFullReadOnlyController` to support CSV for metadata

[DHIS2-12035](https://jira.dhis2.org/browse/DHIS2-12035)


jbee (approved):
First I thought it would be good to move code more to the service layer but this really is the controller layer translation so splitting that would most likely just be more confusing and also inconsistent with how we usually split code.

Long term we might invent a new concept for this type of API related transformation so it does not occur directly in the controllers?

netroms (approved):

mortenoh:
> First I thought it would be good to move code more to the service layer but this really is the controller layer translation so splitting that would most likely just be more confusing and also inconsistent with how we usually split code.
> 
> Long term we might invent a new concept for this type of API related transformation so it does not occur directly in the controllers?

Yes, I think a common way to handle translations would be good here (similar to what Spring provides through message convers). I didn't want to make this too far from what we are already doing (the method is 50% copy-paste of the main method).

PR #9919, opened 2022-02-28
feat: Add entitycriteriaquery to teifilter [DHIS2-12574]

ameenhere:
This PR adds a new jsonb column to trackedentityinstancefilter table.  All filtering criteria will be part of this jsonb column. 3 columns, namely followup, enrollmentcreatedperiod and enrollmentstatus will be moved within the jsonb column. For backward compatibility with Android, these 3 columns will still be added as part of the API response but will not be part of the database.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentityfilter/TrackedEntityInstanceFilter.java
```diff
@@ -177,36 +160,71 @@ public void setSortOrder( int sortOrder )
...
     public void setEnrollmentCreatedPeriod( FilterPeriod enrollmentCreatedPeriod )
     {
-        this.enrollmentCreatedPeriod = enrollmentCreatedPeriod;
+        if ( this.entityQueryCriteria == null )
+        {
+            this.entityQueryCriteria = new EntityQueryCriteria();
+        }
+
+        DateFilterPeriod enrollmentCreatedDate = new DateFilterPeriod();
+        enrollmentCreatedDate.setStartBuffer( enrollmentCreatedPeriod.getPeriodFrom() );
```
enricocolasante, line 223: Possible NPE to check

enricocolasante (approved):
Just double check the warning raised by github
There is one possible NPE that needs to be checked

zubaira (approved):

PR #9923, opened 2022-03-01
feat: Indicator sub-expressions DHIS2-11522

jimgrace:
See [DHIS2-11522](https://jira.dhis2.org/browse/DHIS2-11522). This PR implements the indicator `subExpression()` function. This allows a bit of expression logic, the subexpression, to be translated to SQL and executed by `JdbcAnalyticsManager` as part of an analytics query when fetching the `value` (or `textvalue`) column for an indicator. For example, say you want to count how many orgUnits had more than 100 malaria cases. You can use the following indicator numerator expression:

```
subExpression(if(#{Khoh0Cei7oh}>100,1,0)
```

This will generate logic to use when accessing the "value" column in an analytics query. So instead of:

```
select ax."dx", ... sum(value) from analytics_2022 ...
```

the query will contain:

```
select ax."dx", ... sum( case when value > 100 then 1 else 0 end) from analytics_2022 ...
```

Some key changes (not a complete list):

### FunctionSubExpression

This class visits the parsed subExpression() function. It actually subclasses `DimensionItem`, which is otherwise the parent of classes used to contain the UIDs for `DimensionalItemObjects` such as `DataElements #{dataElementUid}`, `DataElementOperands #{dataElementUid.catOptionComboUid}`, `ProgramIndicators I{programIndicatorUid}`, etc. In fact `FunctionSubExpression` returns a `DimensionItem` that has the UIDs for either a `DataElement` or a `DataElementOperand` (the one referenced within the subexpression), and a `QueryModifiers` object with two fields filled in: (1) the SQL to be used in place of the analytics `value` column (or `textvalue` column for TEXT data elements), and (2) the `ValueType` returned by the subexpression. (The subexpression `ValueType` may be different from the data element's value type. For example, the data element may have a TEXT `ValueType` but the subexpression has an `if` function that returns a numeric value depending on the text value of the data element.)

`FunctionSubExpression` constructs the `DimensionItem` as follows: First, it makes a `getExpressionInfo` pass through the subtree inside the function, to find the UID of the `DataElement` or UIDs of the `DataElementOperand` contained within. It also does validity checking on the subexpression to make sure that one, and only one `DataElement` or `DataElementOperand` is referenced (1 or more times), and that no other types of `DimensionItems` are referenced in the subexpression. This pass also returns an `Object` that tells what `ValueType` is produced by the subexpression, because on this pass the Java code evaluates the subexpression and supplies default values with the correct data types.

Then it makes a second pass through the function argument's subtree, this time with the `getSql` method to translate the subexpression into SQL. Nearly all of this logic is already there, because basic expressions already know how to convert themselves to SQL for use in program indicators. For example, an if function can convert itself to a SQL CASE statement. The only SQL-generating logic that was added is in `DimItemDataElementAndOperand`, where `getSql` returns the column name, either `textvalue` if the `DataElement` has a text or date `ValueType`, or `value` otherwise. `DimItemDataElementAndOperand` also checks to see if it's generating SQL for the subExpression function as otherwise it's not valid to be representing a `DataElement` or `DataElementOperand` as SQL.

### QueryModifiers

`subExpression` and `valueType` were added.

### DataElement

A `getAnalyticsDataType` method was added that returns `DataType.TEXT` if the `ValueType` is text or date, or `DataType.NUMERIC` otherwise. I named the function `getAnalyticsDataType` because these are not all the DataTypes, but they are the two values used in analytics. This logic had been in `QueryPlannerUtils.getDataTypeDataElementMap` but was moved here because we need to know at parsing time which of these `DataTypes` a `DataElement` has and hence which value column will be used in analytics.

A `getValueColumn` method was added. If there is no subexpression, it returns `value` or `textvalue`, depending on `getAnalyticsDataType`. If there is a subexpression (in an attached `QueryModifiers`), it returns the subexpression.

### DataQueryParams

Added `valueColumn`, so this can be passed from `DefaultQueryPlaner` to `JdbcAnalyticsManager`. If no explicit `valueColumn` has been set, `getValueColumn` returns `value` for use in `JdbcAnalyticsManager` with other analytics tables. Also added `queryModsId` (see below).

### JdbcAnalyticsManager

This uses `DataQueryParams.getValueColumn` as the SQL query column name (or expression) for values. It can be `value`, or `textvalue`, or a subexpression. It also uses `queryModsId` (see below).

### DimensionalObject

Defines symbolic names for:

```
    String VALUE_COLUMN_NAME = "value";

    String TEXTVALUE_COLUMN_NAME = "textvalue";
```

I wasn't entirely sure where to put these; they can be moved if there is a better place.

### queryModsId, DefaultQueryPlaner, and QueryPlannerUtils

[Updated 14-Mar-2022] Fixed in PR [pull/10091](https://github.com/dhis2/dhis2-core/pull/10091), see also [DHIS2-12836](https://jira.dhis2.org/browse/DHIS2-12836).

### Testing

`ExpressionServiceTest.testSubExpressions` shows the details of how the `QueryModifiers` are used in `DimensionItem` for subexpressions.

`AnalyticsServiceQueryModifiersTest` is an end-to-end integration test for all `QueryModifiers` such as subexpressions that includes parsing, analytics, and expression evaluation.

Because of the performance overhead of building analytics, `AnalyticsServiceQueryModifiersTest` puts all the test cases into a single `@Test` method. I suggest that in the future we (maybe I) extend `IntegrationTestBase` and/or `BaseSpringTest` to provide `@BeforeAll` and `@AfterAll` methods that can be overridden for use in a test like this so the overhead of building analytics could be done once but the individual test cases could have their own methods. This could also be used to refactor `AnalyticsServiceTest` and `EventAnalyticsServiceTest` to make them much more usable, which I would be willing to do.

In addition to the unit tests, I have tested indicators with subexpressions in a working system. I have compared SL Demo analytics dashboards in modified and unmodified systems and they look the same.

### Performance

I suspected that caching inside `FunctionSubExpression` might be a good idea, and profiling confirmed this. `FunctionSubExpression` makes two passes through the subexpression parse tree each time it is called, first to get information and the data type of the subtree, and second to get the SQL for the subexpression. On the SQL pass, the referenced `DataElement` needs to be looked up from its identifier to know whether it has a `value` or `textvalue` column depending on its `ValueType`. These passes are made not only on the initial calls to get expression info, but on every call to evaluate a single expression value result. This is because we need to find the `DimensionItem` UIDs each time to know which value from the database to put where in the expression. Profiling showed that this was indeed the highest CPU user in a test with subexpressions. A subexpression cache was implemented to help this.

The second-highest performance hit was initializing `ExpressionParams.samplePeriods` to a default list of periods. This was easily solved by creating the list as a static final and assigning it on instance creation. [Updated 14-Mar-2022] Fixed in PR [pull/10091](https://github.com/dhis2/dhis2-core/pull/10091).

After these two speed-ups, a test analytics visualization with subexpressions was reduced from 2.8 minutes to 1.1 minutes on my machine.

The next-highest performance hit is getting a `I18n` from `Defaulti18nManager` -> `UserSettingLocaleManager` -> `DefaultUserSettingService` -> get `UserSettingKey.UI_LOCAL` for the current user. The problem is not finding the current user but getting the locale setting from the DB. We might consider a per-user cache here at some point.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
```diff
@@ -75,4 +75,53 @@
...
+    private final ValueType valueType;
+
+    // -------------------------------------------------------------------------
+    // Logic
+    // -------------------------------------------------------------------------
+
+    /**
+     * Gets the query modifiers that matter for analytics grouping.
+     */
+    public QueryModifiers getQueryModsForAnalyticsGrouping()
```
jbee, line 99: such modifying methods should better be named `with...` instead of `get...`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
```diff
@@ -75,4 +75,53 @@
...
+    // -------------------------------------------------------------------------
+    // Supportive methods
+    // -------------------------------------------------------------------------
+
+    /**
+     * Checks if these query mods have all default values.
+     */
+    private boolean isDefault()
+    {
+        return this.equals( QueryModifiers.builder().build() );
```
jbee, line 125: Maybe there should be a `DEFAULT` constant for `QueryModifiers.builder().build()` then in this class.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElement.java
```diff
@@ -282,6 +285,32 @@ public boolean isFileType()
         return getValueType().isFile();
     }
 
+    /**
+     * Data type for use in analytics. Both text and date types are recognized
+     * as TEXT. Everything else is recognized as NUMERIC.
+     */
+    public DataType getAnalyticsDataType()
+    {
+        return (getValueType().isText() || getValueType().isDate())
```
jbee, line 294: given that `getValueType()` is no longer a simple field access I think it is good practice to extract a local variable if the result is needed multiple times. Otherwise one uses these "getters" and in reality this triggers a cascade of computation each time. Also this helps to see that the same value is used in the return expression.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
```diff
@@ -61,6 +61,8 @@
 @Builder( toBuilder = true )
 public class ExpressionParams
 {
+    private static final List<Period> SAMPLE_PERIODS = List.of( PeriodType.getPeriodFromIsoString( "19990101" ) );
```
jbee, line 64: Maybe a comment should explain why `1999-01-01` was chosen?

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceSubExpressionTest.java
```diff
@@ -0,0 +1,299 @@
...
+        {
+            service.dropTables();
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // Test
+    // -------------------------------------------------------------------------
+
+    @Disabled
```
jbee, line 200: ?

jbee (commented):
As usual for expressions I cannot say much on the semantics of sub-expressions as I don't know much about their usage or function.

A general observation: there are many "getters" that actually have logic. While this is not necessarily a problem (especially if this in the context of immutable data that is transformed or sliced in such getters) it does indicate a high degree of orthogonal features or at least the possibility of variables that have highly complex interaction. This is a problem in the sense that it at some point becomes almost impossible to predict what is or even should happen in such a system. 

Some code style comments.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
jimgrace: Done.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryModifiers.java
jimgrace: Good idea, done.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElement.java
jimgrace: Done.

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceSubExpressionTest.java
jimgrace: See my comments under Testing (and also the several commits "try to fix integration test"). This works fine on my system in the IDE or on the command line, but doesn't work on the integration test server. I didn't want to further hold up the PR after my attempts at fixing this had failed. (Insights from anyone would be most welcome!) I've now put a comment in the code about this.

jimgrace:
Thanks as always for the review, @jbee. I had also been wondering about the logic in the "getters". Good thing to pay attention to.

jbee (approved):

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceSubExpressionTest.java
jbee: Missed that when speed reading the description :D

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElement.java
```diff
@@ -282,6 +285,35 @@ public boolean isFileType()
         return getValueType().isFile();
     }
 
+    /**
+     * Data type for use in analytics. Both text and date types are recognized
+     * as TEXT. Everything else is recognized as NUMERIC. Note that this needs
+     * to be based on the QueryModifiers.valueType if present.
+     */
+    public DataType getAnalyticsDataType()
```
larshelge, line 293: I think this is a bit unclear. I.e. in what analytics context do we think of everything which is non text or date as numeric?

jimgrace: Hi @larshelge, thanks. When populating the analytics `value` and `textvalue` columns [here in JdbcAnalyticsTableManager](https://github.com/dhis2/dhis2-core/blob/master/dhis-2%2Fdhis-services%2Fdhis-service-analytics%2Fsrc%2Fmain%2Fjava%2Forg%2Fhisp%2Fdhis%2Fanalytics%2Ftable%2FJdbcAnalyticsTableManager.java#L252-L257), the text and date values go into `textvalue` and everything else goes into `value`.

I was patterning this code after the logic [here in QueryPlannerUtils](https://github.com/dhis2/dhis2-core/blob/master/dhis-2%2Fdhis-services%2Fdhis-service-analytics%2Fsrc%2Fmain%2Fjava%2Forg%2Fhisp%2Fdhis%2Fanalytics%2Fdata%2FQueryPlannerUtils.java#L104-L108) which uses `DataType.TEXT` to represent text and date values and `DataType.NUMERIC` for everything else. If there's a better way to code this or to comment it, I'm open to suggestions. It seemed a bit odd to be using a subset of `DataType` enum values for this special purpose, which is why I called this `getAnalyticsDataType` instead of just `getDataType` (as mentioned above under **DataElement**).

Other possibilities include:
1. add the analytics column name as a field in the DataType enum.
2. define a new 2-value enum, possibly added as a field in the DataType enum.
3. just use the column name in `DataElement` and `QueryPlannerUtils` instead of a two-valued `DataType`.
4. use a dummy column name for the value when the subexpression is generated, and have `JdbcAnalyticsManager` substitute the real column name (not as elegant for parsing, but keeps the analytics SQL knowledge more localized).

Most of these represent tradeoffs between (a) extending knowledge of analytics SQL columns into other parts of the code vs. (b) defining logical constructs in other parts of the code which are really there because analytics uses two columns for data values.

I'm open to suggestions.

jimgrace:
On other test branches I tried a couple of draft PRs with commits to try and diagnose the problem with `AnalyticsServiceQueryModifiersTest` not working. I isolated it to a problem in the analytics table build. There were only five rows in the `datavalue` table, and the analytics build created the `analytics_2022` table but did not put any rows in it. This was before any expressions with `QueryModifiers` were even created; just a straight analytics build problem. As I was putting more debug statements into the table builds to try and further investigate the problem, the build started working. Then when I changed this PR to enable the test, it also worked. Very strange, but the test is working now.

PR #9934, opened 2022-03-01
feat: add orgUnitField to Map View

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12734

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_39__Add_organisationUnitCoordinateField_column_to_mapView_table.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_39__Add_organisationUnitCoordinateField_column_to_mapView_table.sql
@@ -1,0 +1,1 @@
+alter table mapview add column if not exists orgUnitField varchar(255);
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_39__Add_organisationUnitCoordinateField_column_to_mapView_table.sql
```diff
@@ -0,0 +1 @@
+alter table mapview add column organisationunitcoordinatefield varchar(255);
```
jbee, line 1: `if not exists` is missing

jbee (commented):
Just the SQL needs the `if not exists` - otherwise good to go.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/mapping/MapView.java
```diff
@@ -187,6 +187,8 @@
 
     private String eventCoordinateField;
 
+    private String organisationUnitCoordinateField;
```
larshelge, line 190: Actually we have a field already @vietnguyen , if you look at `BaseAnalyticalObject` we have a field `orgUnitField` already. So you can just map that `orgUnitField` field in the hibernate mapping but not add this field to `MapView`.

vietnguyen: fixed

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.38/V2_38_39__Add_organisationUnitCoordinateField_column_to_mapView_table.sql
vietnguyen: fixed

jbee (approved):

PR #9935, opened 2022-03-01
fix: Sms configuration update [ DHIS2-12590 ] 

lucaCambi77:
[DHIS2-12590](https://jira.dhis2.org/browse/DHIS2-12590)

Default SMS length added as system key. Also added a maxSmsLength property for the gateway that can be set, otherwise default system one is used.
Message column in outbound_sms table it now text

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsGateway.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsGateway.java
@@ -110,36 +110,28 @@ public abstract class SmsGateway
         {
             response = restTemplate.exchange( urlTemplate, httpMethod, request, klass );
 
-            if ( response != null )
-            {
-                statusCode = response.getStatusCode();
-            }
-            else
-            {
-                log.error( "Server response is null" );
-                statusCode = HttpStatus.INTERNAL_SERVER_ERROR;
-            }
+            statusCode = response.getStatusCode();
         }
         catch ( HttpClientErrorException ex )
         {
-            log.error( "Client error", ex );
+            log.error( "Sms request Client error", ex );
 
             statusCode = ex.getStatusCode();
         }
         catch ( HttpServerErrorException ex )
         {
-            log.error( "Server error", ex );
+            log.error( "Sms request Server error", ex );
 
             statusCode = ex.getStatusCode();
         }
         catch ( Exception ex )
         {
-            log.error( "Error", ex );
+            log.error( "Sms request Error", ex );
 
             statusCode = HttpStatus.INTERNAL_SERVER_ERROR;
         }
 
-        log.info( "Response status code: " + statusCode );
+        log.info( "Sms request status code: " + statusCode );
 
         return statusCode;
     }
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsGateway.java
```diff
@@ -110,36 +110,28 @@ public HttpStatus send( String urlTemplate, HttpEntity<?> request, HttpMethod ht
...
-            {
-                log.error( "Server response is null" );
-                statusCode = HttpStatus.INTERNAL_SERVER_ERROR;
-            }
+            statusCode = response.getStatusCode();
         }
         catch ( HttpClientErrorException ex )
         {
-            log.error( "Client error", ex );
+            log.error( "Sms request Client error", ex );
```
teleivo, line 117: Can you please lower case the former first word of the sentence. Same in some of the logs below.

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/sms/SmsMessageSenderTest.java
```diff
@@ -72,13 +75,16 @@
...
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
 import com.google.common.collect.Sets;
 
 /**
  * @author Zubair Asghar.
  */
 @ExtendWith( MockitoExtension.class )
+@MockitoSettings( strictness = Strictness.LENIENT )
```
teleivo, line 87: Please adapt your tests so that we use mockito strict. If you see tests with `@MockitoSettings( strictness = Strictness.LENIENT )` its because previously that was the default mockito setting. When migrating Mockito/JUnit 5 we had to add this setting since a lot of the tests would fail. We should make an effort to make all our tests use the now default mockito strict mode. This helps keep tests easier to read/maintain going forward.

teleivo (changes requested):
looks good to me 😄 expect for the test which should use mockitos strict setting (which is the default). once strict I would see it as ✅

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsGateway.java
```diff
@@ -110,36 +110,28 @@ public HttpStatus send( String urlTemplate, HttpEntity<?> request, HttpMethod ht
         {
             response = restTemplate.exchange( urlTemplate, httpMethod, request, klass );
 
-            if ( response != null )
-            {
-                statusCode = response.getStatusCode();
-            }
-            else
-            {
-                log.error( "Server response is null" );
```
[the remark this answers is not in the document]
lucaCambi77: no it can't be null, i don't know why it was there

enricocolasante (changes requested):
Master is already 2.39 so if you add a flyway script you should rename to 2.39 standard and then use the current name when backport to 2.38 branch

enricocolasante (approved):

PR #9945, opened 2022-03-02
fix: Change date fields from string to Date

netroms:
Changes the date fields from String to Date in order to have consistent and correct date formatting.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/MeController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/MeController.java
@@ -164,7 +164,7 @@ public class MeController
         Sets.newHashSet( UserSettingKey.values() ) );
 
     @GetMapping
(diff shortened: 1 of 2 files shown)


mortenoh (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/MeController.java
```diff
@@ -164,7 +164,7 @@
         Sets.newHashSet( UserSettingKey.values() ) );
 
     @GetMapping
-    public ResponseEntity<JsonNode> getCurrentUser( @CurrentUser( required = true ) User user,
+    public @ResponseBody ResponseEntity<JsonNode> getCurrentUser( @CurrentUser( required = true ) User user,
```
jbee, line 167: If you have the `ResponseEntity` wrapper you do not need the `@ResponseBody` - but for any other return type it is needed.

jbee (approved):

PR #9957, opened 2022-03-03
fix: Fix NPE checking for null org unit [DHIS2-12713]

enricocolasante:


lucaCambi77 (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -235,9 +235,12 @@ public DefaultTrackerAccessManager( AclService aclService, TrackerOwnershipManag
         {
             OrganisationUnit ou = programInstance.getOrganisationUnit();
 
-            if ( !canAccess( user, program, ou ) )
+            if ( ou != null )
             {
-                errors.add( "User has no read access to organisation unit: " + ou.getUid() );
+                if ( !canAccess( user, program, ou ) )
```
teleivo, line 240: I think you can merge these if statements into one. Maybe even better, make canAccess null safe.

teleivo (changes requested):

teleivo (approved):

PR #9980, opened 2022-03-04
feat: Check username not already exists and make it immutable

netroms:
### Summary:

This PR adds 3 validations to the User API, in the UserObjectBundleHook:
* User object must contain a reference to at least one UserRole.
* Username can not be modifed once created.
* Username must be unique.

These validations are only checked in the UI before this PR, except for username which is enforced in the database table via an unique constraint.
This PR adds the same validations to the API endpoint. 

This PR also removes the invitee can chose his/her own username feature.
Since usernames now in practice is immutable, the inviter will have to choose the username when sending the invite.


dhis-2/dhis-e2e-test/src/main/resources/config.properties
```diff
@@ -3,8 +3,8 @@
 instance.url=http://localhost:8080/api
 # properties for user set up on the instance.
 # user must be able to import metadata.
-user.default.username=system
-user.default.password=System123
+user.default.username=admin
+user.default.password=district
```
jbee, line 7: Didn't I just recently change these to the values now replaced because otherwise e2e gives you trouble locally with demo database?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/DefaultSecurityService.java
```diff
@@ -242,7 +243,7 @@ public void prepareUserForInvite( User user )
 
         if ( user.getUsername() == null || user.getUsername().isEmpty() )
         {
-            String username = "invite-" + user.getEmail() + "-" + CodeGenerator.generateUid();
+            String username = "invite-" + user.getEmail() + "-" + CodeGenerator.generateUid().toLowerCase();
```
jbee, line 246: hm.. isn't the email good enough? and should the email not be unique you cannot invite that user as it obviously already is a user? - a user with this generated name is kind of annoying

dhis-2/dhis-web/dhis-web-commons-resources/src/main/webapp/dhis-web-commons/useraccount/account.vm
```diff
@@ -37,6 +37,7 @@
         <td style="width:140px"><label for="code"></label></td>
         <td>
             <input type="hidden" id="inviteUsername" name="inviteUsername" value="$username">
+            <input type="hidden" id="username" name="username" value="$username">
```
mediremi, line 40:
```suggestion
```

dhis-2/dhis-web/dhis-web-commons-resources/src/main/webapp/dhis-web-commons/useraccount/account.vm
```diff
@@ -49,16 +50,22 @@
...
         <td>
     </tr>
+    #else
+        <tr>
+            <td><label id="label_username" for="username">$i18n.getString( "user_name" )</label></td>
+            <td>
+                <input type="text" id="username" name="username" autocomplete="off"></td>
+            <td>
+        </tr>
+    #end
```
mediremi, line 68: I think this part can be simplified to:

```
<tr>
  <td>
    <label id="label_username" for="username">$i18n.getString( "user_name" )</label>
  </td>
  <td>
    <input
        type="text"
        id="username"
        name="username"
        autocomplete="off"
        
        #if( $accountAction == "invited" )
        disabled="disabled"
        value="$username"
        #end
    />
  </td>
</tr>
```

larshelge:
One unit test is failing, which could seem related:

```

Error:  Failures: 
Error:    AccountControllerTest.testCreateAccount:66->DhisMockMvcControllerTest.assertWebMessage:58->DhisMockMvcControllerTest.assertWebMessage:69 unexpected message ==> expected: <User self registration is not allowed> but was: <Password must have at least 8, and at most 40 characters>
[INFO] 
```

mediremi (approved):

jbee (approved):

jbee (approved):

PR #9981, opened 2022-03-04
feat: add orgUnitFieldName to mapView

vietnguyen:
In https://github.com/dhis2/dhis2-core/pull/9934 we added property `orgUnitField` to `MapView` in database level, which stores an ID of an `Attribute`
Now, client also needs the `Attribute`'s name and doesn't want to send an additional request to server.

This PR add a transient property `orgUnitFieldName` to `MapView`, its value will be set in `MapController#postProcessResponseEntity`

Note that the method `attributeService.getAttribute(id)` is already using a `Cache<Attribute> attributeCache`

```json
"orgUnitFieldName": "testgeojson",
"orgUnitField": "jhqTNk3hS2N",
```

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/mapping/MapView.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/mapping/MapView.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Optional;
 
 import org.hisp.dhis.analytics.EventOutputType;
+import org.hisp.dhis.attribute.Attribute;
 import org.hisp.dhis.common.BaseAnalyticalObject;
 import org.hisp.dhis.common.BaseIdentifiableObject;
 import org.hisp.dhis.common.DimensionalItemObject;
@@ -218,6 +219,15 @@ public class MapView
     private transient List<OrganisationUnit> organisationUnitsAtLevel = new ArrayList<>();
 
     private transient List<OrganisationUnit> organisationUnitsInGroups = new ArrayList<>();
+
+    /**
+     * The displayName of the {@link Attribute} which has ID stored by property
+     * {@link MapView#orgUnitField}
+     * <p>
+     * The value of this transient property will be set in
+     * MapController#postProcessResponseEntity
+     */
+    private transient String orgUnitFieldDisplayName;
 
     public MapView()
     {
@@ -876,5 +886,17 @@ public class MapView
     public void setParentLevel( int parentLevel )
     {
         this.parentLevel = parentLevel;
+    }
+
+    @JsonProperty
+    @JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
+    public String getOrgUnitFieldDisplayName()
+    {
+        return orgUnitFieldDisplayName;
+    }
+
+    public void setOrgUnitFieldDisplayName( String orgUnitFieldDisplayName )
+    {
+        this.orgUnitFieldDisplayName = orgUnitFieldDisplayName;
     }
 }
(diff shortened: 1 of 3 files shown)


turban:
Not sure if this is the best solution. If the name of the attribute is changed in the Maintenance app, it should also be reflected in the apps using it. We also need to handle translations. It would be preferable if the attribute name is fetched from the attribute itself. This is similar to what we do with other mapView properties like `legendSet`: 

Screenshot 2022-03-03 at 11 27 54

When we save a map with an orgUnitField (could be named differently if we don't want to reuse an existing prop) id should be sufficient: 

```
"orgUnitField": {
    "id": "ihn1wb9eho8"
} 
```

when a map (mapView) is returned it should also include the attribute name with the current value from db:

```
"orgUnitField": {
    "id": "ihn1wb9eho8",
    "name": "Catchment area"
} 
```

vietnguyen:
> Not sure if this is the best solution. If the name of the attribute is changed in the Maintenance app, it should also be reflected in the apps using it. We also need to handle translations. It would be preferable if the attribute name is fetched from the attribute itself. This is similar to what we do with other mapView properties like `legendSet`:
> 
> Screenshot 2022-03-03 at 11 27 54
> 
> When we save a map with an orgUnitField (could be named differently if we don't want to reuse an existing prop) id should be sufficient ("name" can be included but have no effect when saving the map):
> 
> ```
> "orgUnitField": {
>     "id": "ihn1wb9eho8"
> } 
> ```
> 
> when a map (mapView) is returned it should also include the attribute name with the current value from db:
> 
> ```
> "orgUnitField": {
>     "id": "ihn1wb9eho8",
>     "name": "Catchment area"
> } 
> ```
> 
> This is how we show the attribute name in DHIS2 Maps:
> 
> Screenshot 2022-03-04 at 13 30 40
> 
> If this solution is not doable, I think we can ignore this PR, and just use `orgUnitField": "jhqTNk3hS2N`, and an extra API request to get the attribute name (which I would like to avoid).

@turban The attribute name is getting from the attribute, and also translation is handled by calling `getDisplayName()`. And if the name of the attribute is changed, it will also reflected in the response of `api/maps` endpoint. We use an attributeCache for all endpoint in order to retrieve the attributeName. This is the same for other metadata endpoints which we need to return the attribute name in attributeValues, where only attribute id is saved in database

turban:
I see. I would prefer to have the same pattern as already used for other properties: 

Screenshot 2022-03-07 at 10 54 37

But this solution is also fine. `orgUnitFieldName` will be a read only property?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/mapping/MapView.java
```diff
@@ -219,6 +220,15 @@
...
     private transient List<OrganisationUnit> organisationUnitsInGroups = new ArrayList<>();
 
+    /**
+     * The displayName of the {@link Attribute} which has ID stored by property
+     * {@link MapView#orgUnitField}
+     * <p>
+     * The value of this transient property will be set in
+     * MapController#postProcessResponseEntity
+     */
+    private transient String orgUnitFieldName;
```
larshelge, line 230: Could we call this property `orgUnitFieldDisplayName` to make the name explicit, and allow for a potential `orgUnitFieldName` with the un-translated name in the future?

vietnguyen: Changed to `orgUnitFieldDisplayName` now.

vietnguyen:
> I see. I would prefer to have the same pattern as already used for other properties:
> 
> Screenshot 2022-03-07 at 10 54 37
> 
> But this solution is also fine. `orgUnitFieldName` will be a read only property? @vietnguyen

Yes, it's a readonly property. We only save the Attribute ID to property `orgUnitField` in database.

larshelge (approved):

PR #9983, opened 2022-03-04
fix: /generateAndReserve returns duplicate values when attribute pattern contains a constant [ DHIS2-12182 ]

lucaCambi77:
adds lower case in query and index to reserved value table


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_1__Add_index_to_reserved_value.sql
```diff
@@ -0,0 +1 @@
+create unique index if not exists in_reservedvalue_value_generation on reservedvalue using btree (ownerobject, owneruid, key, lower (value));
```
stian-sandvold, line 1: Since we are basically recreating an index we already have, lets remove the old one.

enricocolasante (approved):

stian-sandvold (approved):

PR #9990, opened 2022-03-04
fix: Analytics export crash with invalid dates [DHIS2-12581]

maikelarabori:
These changes aim to fix an issue/crash during the analytics tables export.

Currently, we have a limit of years allowed by the analytics table when exporting events.
As you can see below (in the PR), we are limited to dates between 1975 and 2024.
(I'm keeping this behaviour to minimize risks. Ideally, at some point in the future, we need to remove this limitation and stop using the periods' table during the analytics export.)

This fix will basically ensure that we only read events and years that belong to that interval (inclusive).
It will guarantee that no partitions with invalid years (out of that interval) will be created, avoiding joins with the table `_dateperiodstructure` on top of years that do not exist in that same table.

When we try to join using years out of the interval 1975-2024 the partition tables will be populated with `null` values in the `year` column.
Later, when the process tries to attach the partition to the main table, we get something like:
```
org.postgresql.util.PSQLException:
ERROR: partition constraint of relation "analytics_event_ywhgjzui7r5_1900" is violated by some row
```
That happens because 1900 is out of the interval and the partition table will contain `null` values in the `year` column with causes the violation.

This was reported by the community and more details can be found at https://jira.dhis2.org/browse/DHIS2-12581


dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTable.java
```diff
@@ -37,6 +37,10 @@
  */
 public abstract class ResourceTable<T>
 {
+    public static final int OLDEST_YEAR_PERIOD_SUPPORTED = 1975;
+
+    public static final int NEWEST_YEAR_PERIOD_SUPPORTED = 2024;
```
d-bernat, line 42: hmm, what about to use (current year) + 2

d-bernat (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTable.java
maikelarabori: Done

gnespolino (approved):

PR #9992, opened 2022-03-06
fix: totalPages param does not work in event analytics [DHIS2-12529]

d-bernat:
Calculating total number and page count is a potential performance problem in event analytics '/query' endpoints. Currently the totalPages param is not working and returns the total.

The details for what should be returned are specified in this bug report: https://jira.dhis2.org/browse/DHIS2-12530

### DIFF
--- a/dhis-2/dhis-api/src/test/java/org/hisp/dhis/common/EventDataQueryRequestTest.java
+++ b/dhis-2/dhis-api/src/test/java/org/hisp/dhis/common/EventDataQueryRequestTest.java
@@ -28,15 +28,23 @@
 package org.hisp.dhis.common;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-api/src/test/java/org/hisp/dhis/common/EventDataQueryRequestTest.java
```diff
@@ -34,10 +34,17 @@
 import java.util.Set;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 public class EventDataQueryRequestTest
 {
 
+    public static boolean[] totalPagesFlags()
```
maikelarabori, line 43: Is this being used?

maikelarabori (approved):
Looks good Dusan, just one minor comment left. :+1:

gnespolino (approved):

dhis-2/dhis-api/src/test/java/org/hisp/dhis/common/EventDataQueryRequestTest.java
maikelarabori: I just realized it's being used at `@MethodSource( value = "totalPagesFlags" )`

PR #9997, opened 2022-03-07
fix: don't let presets overwrite explicit fields

mortenoh:
Fixes a bug where preset expansion would overwrite field transformers.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
+++ b/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
@@ -238,7 +238,12 @@ public class FieldPathHelper
             }
         }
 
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
```diff
@@ -238,7 +238,12 @@ else if ( FieldPreset.SIMPLE.equals( preset.getName() ) )
             }
         }
 
-        fieldPaths.forEach( fp -> fieldPathMap.put( fp.toFullPath(), fp ) );
+        fieldPaths.forEach( fp -> {
+            if ( !fieldPathMap.containsKey( fp.toFullPath() ) )
+            {
+                fieldPathMap.put( fp.toFullPath(), fp );
```
jbee, line 244: I guess you could use `putIfAbsent` to avoid the if and allow to write this in a lambda without the curlies.

jbee (approved):

vietnguyen (approved):

PR #10011, opened 2022-03-08
fix: recursively apply access objects

mortenoh:
## Summary

* Recursively applies `access` object to any depth
* Updates `/api/dataSets/{uid}/categoryCombos` to use new field filter
* Updates `/api/dimensions/constraints` to use new field filter
* Updates `/api/dimensions/recommendations` to use new field filter
* Updates `/api/dimensions/dataSet/{uid}` to use new field filter


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterParams.java
```diff
@@ -51,6 +52,8 @@
     @Builder.Default
     private final Set<String> filters = Collections.singleton( "*" );
 
+    private User user;
```
jbee, line 55: `final`?

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
```diff
@@ -65,6 +70,22 @@
     @Qualifier( "jsonMapper" )
     private final ObjectMapper jsonMapper;
 
+    private final SchemaService schemaService;
+
+    private final AclService aclService;
+
+    private final CurrentUserService currentUserService;
+
+    public FieldFilterService( FieldPathHelper fieldPathHelper, ObjectMapper jsonMapper, SchemaService schemaService,
```
jbee, line 79: could this use `@AllArgsConstructor` when you annotate the `@Qualifier` on the field?

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
```diff
@@ -130,6 +152,45 @@ public FieldFilterService( FieldPathHelper fieldPathHelper, ObjectMapper jsonMap
...
+            {
+                fieldPathHelper.visitFieldPaths( object, List.of( fp ), o -> {
+                    if ( !(o instanceof BaseIdentifiableObject) )
+                    {
+                        return;
+                    }
+
+                    BaseIdentifiableObject baseIdentifiableObject = (BaseIdentifiableObject) o;
+                    baseIdentifiableObject
+                        .setAccess( aclService.getAccess( baseIdentifiableObject, params.getUser() ) );
```
jbee, line 188: isn't it more readable just to do 
```java
if ( (o instanceof BaseIdentifiableObject) )
{
  BaseIdentifiableObject bao = (BaseIdentifiableObject) o;
  bao.setAccess( aclService.getAccess( bao, params.getUser() ) );
}
```

jbee (commented):
Some details

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
mortenoh: But still need to configure the jsonMapper (it does local customization on it)

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterParams.java
mortenoh: It can't be, the initialization of this property can happen later in the pipeline (so we don't require a user property when creating a param object), keeping it non final for now.

larshelge (approved):

vietnguyen (approved):

PR #10027, opened 2022-03-09
fix: Master table population for partial update [DHIS2-12677]

maikelarabori:
This is a tricky bug that happens during the analytics partial export.

When the update is partial the code will target the respective partition and will skip the master table. This is the current behaviour.

The problem arises when the table has no partitions (ie. `enrollments`). In such cases, the table export does not update any table (will also skip the master).

So, to fix it, I'm adding an extra condition (`&& table.getTableType().hasLatestPartition()`).
This condition will tell if the current table actually has partitions. If it does not have partitions we need to update the master table, so the data in the analytics enrollment tables get updated.

The client reported that when running analytics export setting `Number of last years of data to include` to `1`, he could see the new events rows, but not the new enrollments rows.
More details can be found in the issue: https://jira.dhis2.org/browse/DHIS2-12677
That's the reason why this bug is being fixed.

**_Porting it from 2.35_**

**NOTE**
_I also removed a migration labelled for 2.38 that was sitting there by mistake._

### DIFF
--- a/dhis-2/dhis-api/pom.xml
+++ b/dhis-2/dhis-api/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.hisp.dhis</groupId>
     <artifactId>dhis</artifactId>
-    <version>2.36.9-SNAPSHOT</version>
+    <version>2.36.10-SNAPSHOT</version>
   </parent>
 
   <artifactId>dhis-api</artifactId>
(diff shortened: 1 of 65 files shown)


lucaCambi77 (approved):
i can approve for my part, not sure about the rest :)

maikelarabori:
> i can approve for my part, not sure about the rest :)

Thanks @lucaCambi77 :+1:

PR #10035, opened 2022-03-09
fix: bump rule-engine version [DHIS2-12537]

enricocolasante:
I changed the version of rule engine to the latest one that fix a problem in the logs. Now we are logging the evaluation of the rules in debug mode as before it was causing a lot of noise in the logs


jimgrace (approved):
Looks good. I agree with what you're doing with `getNullReplacementValue`, making it like the latest code. Thanks.

PR #10041, opened 2022-03-09
fix: Event update should only check ownership orgunit for tracker programs [DHIS2-12732] (2.39)

ameenhere:
Skip checking event organisation unit in case of event update. Ownership org unit is only required to be checked.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
@@ -330,9 +330,9 @@ public class PreCheckSecurityOwnershipValidationHook
 
         checkEventWriteAccess( reporter, event, programStage, organisationUnit, ownerOrgUnit,
             categoryOptionCombo,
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -330,7 +330,9 @@ private void validateCreateEvent( ValidationErrorReporter reporter, Event event,
 
         checkEventWriteAccess( reporter, event, programStage, organisationUnit, ownerOrgUnit,
             categoryOptionCombo,
-            teiUid, isCreatableInSearchScope ); // TODO: calculate correct
+            teiUid, isCreatableInSearchScope, reporter.getBundle().getStrategy( event ) ); // TODO:
```
teleivo, line 333: reporter is already passed in, so we could save adding yet another param 😅

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
ameenhere: Good point! :D

PR #10044, opened 2022-03-10
fix: bulk sharing update is slow

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12805

### Before the fix ~15 seconds
Screen Shot 2022-03-10 at 1 18 58 AM

### After the fix  ~1 second

Screen Shot 2022-03-12 at 12 00 38 AM

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
@@ -29,6 +29,10 @@ package org.hisp.dhis.jsonpatch;
 
 import static org.hisp.dhis.util.JsonUtils.jsonToObject;
 
+import java.util.Collection;
+
+import org.hisp.dhis.common.BaseIdentifiableObject;
+import org.hisp.dhis.commons.collection.CollectionUtils;
 import org.hisp.dhis.commons.jackson.jsonpatch.JsonPatch;
 import org.hisp.dhis.commons.jackson.jsonpatch.JsonPatchException;
 import org.hisp.dhis.schema.Property;
@@ -40,6 +44,7 @@ import org.springframework.transaction.annotation.Transactional;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
 /**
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
```diff
@@ -104,8 +108,45 @@ public JsonPatchManager(
...
-                node.set( property.getCollectionName(), jsonMapper.valueToTree( data ) );
+
+                Collection collection = (Collection) data;
+
+                if ( CollectionUtils.isEmpty( collection ) )
+                {
+                    continue;
+                }
+
+                if ( BaseIdentifiableObject.class.isAssignableFrom( property.getItemKlass() ) )
```
jbee, line 119: This should be added to the above if so we get
```java
if ( property.isCollection() && BaseIdentifiableObject.class.isAssignableFrom( property.getItemKlass() ) )
```
and the 
```java
else
                {
                    node.set( property.getCollectionName(), jsonMapper.valueToTree( data ) );
                }
```
should then apply both if it is not a collection or if you cannot do this shallow transformation based on `BaseIdentifiableObject` - otherwise you miss non collections I think

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
```diff
@@ -73,7 +78,7 @@ public JsonPatchManager(
      * @param object Jackson Object to apply the patch to.
      * @return New instance of the object with the patch applied.
      */
-    @Transactional
+    @Transactional( readOnly = true )
```
jbee, line 81: Now the big question is: can this cause creation of new periods?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
vietnguyen: We already do the conversion from object to jsonNode before going into the method `handleCollectionUpdates`
```
        Schema schema = schemaService.getSchema( object.getClass() );
        JsonNode node = jsonMapper.valueToTree( object );

        // since valueToTree does not properly handle our deeply nested classes,
        // we need to make another trip to make sure all collections are
        // correctly made into json nodes.
        handleCollectionUpdates( object, schema, (ObjectNode) node );
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jsonpatch/JsonPatchManager.java
vietnguyen: in this patchManager, we only get the data and convert to jsonNode, then patch the jsonNode object. 
No saving data to database, so the Periods will just be copied to the jsonNode, no matter if its existed or not.
The saving/updating will happen  in the importService, where the write transaction should be handled.

jbee (approved):

mortenoh (approved):

PR #10061, opened 2022-03-10
fix: set displayName on UG/UGA if requested by field filter

mortenoh:


jbee (commented):
I guess now you have 3 times the same apply function just that they filter on different path and apply a different function. I think this should be easily extracted to a generic `apply` function having two extra parameters to provide filter property and a callback to set setting.

mortenoh:
> I guess now you have 3 times the same apply function just that they filter on different path and apply a different function. I think this should be easily extracted to a generic `apply` function having two extra parameters to provide filter property and a callback to set setting.

Yes, all this functionality will be refactored soon (just need to get it in today as its breaking apps)

vietnguyen (approved):

netroms (approved):

PR #9996, opened 2022-03-07
fix: return full relationship items DHIS2-12625

teleivo:
NTI `relationship.from/to` will now return the trackedEntity, enrollment and event objects
instead of only their uids. The relationship fields of trackedEntity, enrollment and event will not be returned if they are inside a `relationship.from/to`. (This prevents a StackOverflow due to the recursive structure.)

Imports of relationships into NTI will thus also require nested from/to
relationship items i.e

    "from": { "trackedEntity": { "trackedEntity": "qwerw1243"}}

So exported payloads can directly be imported again.

The following shows sample JSON payloads for import/export in NTI after this PRs changes.

**Note:** unfortunately the GET endpoints are inconsistent in their behavior. Meaning they return a different level of detail by default. The query parameter `fields` is also only partially supported in `GET /api/tracker/trackedEntities`. So the examples you see are mostly what you will get without being able to request more or less details using `fields`. See https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/new-tracker.html

### POST /api/tracker

```json
{
    "relationships": [
        {
            "relationshipType": "mxZDvSZYxlw",
            "from": {
                "event": {
                    "event": "a6f092d0d44"
                }
            },
            "to": {
                "trackedEntity": {
                    "trackedEntity": "mXAzn3hMR5a"
                }
            }
        }
    ]
}
```

### GET /api/tracker/trackedEntities/Ea0rRdBPAIp

Relationships are not returned by default but can be returned using `fields=relationships`. See next response sample

```json
{
  "trackedEntity": "Ea0rRdBPAIp",
  "trackedEntityType": "nEenWmSyUEp",
  "createdAt": "2013-09-08T21:40:28.000",
  "createdAtClient": "2013-09-08T21:40:28.000",
  "updatedAt": "2014-07-23T12:53:31.638",
  "orgUnit": "DiszpKrYNg8",
  "inactive": false,
  "deleted": false,
  "potentialDuplicate": false,
  "relationships": [],
  "attributes": [
    {
      "attribute": "zDhUuAYrxNC",
      "displayName": "Last name",
      "createdAt": "2016-01-12T09:10:26.986",
      "updatedAt": "2016-01-12T09:10:35.884",
      "valueType": "TEXT",
      "value": "Zimmerman"
    },
    {
      "attribute": "w75KJ2mc4zz",
      "code": "MMD_PER_NAM",
      "displayName": "First name",
      "createdAt": "2016-01-12T09:10:26.986",
[...]

PR #10068, opened 2022-03-11
fix: return full relationship items DHIS2-12625 [2.38]

teleivo:
backport of (#9996)

* chore: move export mappers to its usage

since these mappers purpose is to define what the exported payloads look like,
they should live with the controllers using them

* fix: return full relationship items

relationship.from/to will now return the entire tei, enrollment, event
instead of only their uids. their relationship fields will be ignored,
since that would cause a StackOverflow due to the recursive structure.

imports of relationships will thus also require nested from/to
relationship items i.e

    "from": { "trackedEntity": { "trackedEntity": "qwerw1243"}}

* chore: prevent encoding RelationsipItem relationships

by using dedicated classes for TrackedEntity, Enrollment, Event inside of RelationshipItems

### DIFF
--- a/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/databuilder/RelationshipDataBuilder.java
+++ b/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/databuilder/RelationshipDataBuilder.java
@@ -28,8 +28,9 @@
 
 package org.hisp.dhis.tracker.importer.databuilder;
 
-import com.google.gson.JsonObject;
 import org.hisp.dhis.helpers.JsonObjectBuilder;
+
+import com.google.gson.JsonObject;
 
 /**
  * @author Gintare Vilkelyte <vilkelyte.gintare@gmail.com>
@@ -51,9 +52,7 @@ public class RelationshipDataBuilder implements TrackerImporterDataBuilder
 
     public RelationshipDataBuilder setFromEntity( String entityName, String entityId )
     {
-        this.jsonBuilder.addObject( "from", new JsonObjectBuilder()
-            .addProperty( entityName, entityId ) );
-
+        this.jsonBuilder.addObject( "from", relationshipItem(entityName, entityId));
         return this;
     }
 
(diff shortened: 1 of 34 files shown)


teleivo:
@stian-sandvold could you please do the same here for sonaqube ([9996](https://github.com/dhis2/dhis2-core/pull/9996#issuecomment-1063896127)) 🥺 😄

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/relationships/RelationshipsTests.java
```diff
@@ -391,18 +395,14 @@ public void shouldNotImportDuplicateRelationships( String fromTei1, String toTei
 
         JsonObject relationship1 = JsonObjectBuilder.jsonObject()
             .addProperty( "relationshipType", relationshipTypeId )
-            .addObject( "from", JsonObjectBuilder.jsonObject()
-                .addProperty( "trackedEntity", fromTei1 ) )
-            .addObject( "to", JsonObjectBuilder.jsonObject()
-                .addProperty( "trackedEntity", toTei1 ) )
+            .addObject( "from", relationshipItem("trackedEntity", fromTei1 ) )
```
stian-sandvold, line 398: codestyle

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/databuilder/RelationshipDataBuilder.java
```diff
@@ -51,9 +52,7 @@ public RelationshipDataBuilder setRelationshipType( String relationshipType )
 
     public RelationshipDataBuilder setFromEntity( String entityName, String entityId )
     {
-        this.jsonBuilder.addObject( "from", new JsonObjectBuilder()
-            .addProperty( entityName, entityId ) );
-
+        this.jsonBuilder.addObject( "from", relationshipItem(entityName, entityId));
```
stian-sandvold, line 55: codestyle

stian-sandvold (approved):
code looks good, but a couple of files have codestyle issues (missing spaces between paranthesis and params) I added a couple of comments in each file I saw it in :)

teleivo:
> code looks good, but a couple of files have codestyle issues (missing spaces between paranthesis and params) I added a couple of comments in each file I saw it in :)

I will most likely follow up with some more tests and fix these there.

PR #10080, opened 2022-03-11
test: TrackerTrackedEntitiesExportController

teleivo:
* basic behavior
* current behavior of fields query parameter
* behavior when returning TEIs with relationships


teleivo:
pushing tests with field filter fix as they are complementary

PR #10103, opened 2022-03-14
fix: add validations for teiFilter parameters (master)

ameenhere:
This PR adds some validations before persisting a teiFilter.
1. Check if the date filter periods are complete.
2. Check if the assigned users are provided in case of PROVIDED assigned user mode.
3. Check if the organisation unit is provided in case of SELECTED oumode.
4. Check if the order parameters are valid.


enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/webapi/controller/event/mapper/OrderParamsHelper.java
```diff
@@ -54,4 +59,21 @@
                 .build() )
             .collect( Collectors.toList() );
     }
+
+    public static List<String> validateOrderParams( List<OrderParam> orderParams,
+        Map<String, TrackedEntityAttribute> attributes )
+    {
+        List<String> errors = new ArrayList<>();
+        if ( orderParams != null && !orderParams.isEmpty() )
```
teleivo, line 67: I think the case of `orderParams.isEmpty()` is covered by the foreach loop, so you can remove it.

As you might know, I like guard clauses to reduce nesting 😋 You could turn it into

```java
if (orderParams == null) {
return errors;
}

for ( OrderParam orderParam : orderParams ) {
...
```

zubaira (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentityfilter/DefaultTrackedEntityInstanceFilterService.java
```diff
@@ -130,6 +139,57 @@ public TrackedEntityInstanceFilter get( long id )
...
+        validateOrganisationUnits( errors, eqc );
+
+        validateOrderParams( errors, eqc );
+
+        return errors;
+    }
+
+    private void validateOrganisationUnits( List<String> errors, EntityQueryCriteria eqc )
+    {
+        if ( StringUtils.isEmpty( eqc.getOrganisationUnit() )
```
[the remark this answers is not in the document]
ameenhere: Agreed, but havent changed it yet. Going ahead with isEmpty for now though due to lack of time.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/webapi/controller/event/mapper/OrderParamsHelper.java
```diff
@@ -64,16 +64,20 @@
         Map<String, TrackedEntityAttribute> attributes )
     {
         List<String> errors = new ArrayList<>();
-        if ( orderParams != null && !orderParams.isEmpty() )
+
+        if ( orderParams == null || orderParams.isEmpty() )
```
teleivo, line 68: if orderParams.isEmpty() the later loop will not do anything, so this check is redundant

PR #10115, opened 2022-03-15
fix: System update messages are not sent to all group recipients

netroms:


jbee (approved):
I am not quite getting how this changes anything. Does `ImmutableSet` have some odd behaviour we should know about? Or is this change just since we now have `Set.of`? Then how is this PR changing anything?

larshelge:
I think this changes it so that the check for whether a message is created for the specific version is done once at the beginning, so that we make sure the message sending is aborted after the first message.

jbee:
Oh, I missed the added `for` :grimacing:

larshelge (approved):

PR #10117, opened 2022-03-15
fix: nested conditions are now always AND [DHIS2-12839]

gnespolino:
We had
```
Repeatable QueryItem1( dataelement1, GT 5, LT 10)
Repeatable QueryItem2( dataelement2, GT 6, LT 11)
QueryItem3( dataelement3, EQ, "test")
```

it was converted into:
```
(DE1>5 OR DE1<10 OR DE2>6 OR DE2<11) and DE3="TEST"
```

now it is like this:
```
((DE1>5 AND DE1<10) OR (DE2>6 AND DE2<11)) AND DE3="TEST"
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
@@ -28,6 +28,8 @@
 package org.hisp.dhis.analytics.event.data;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.stream.Collectors.joining;
+import static org.apache.commons.collections4.CollectionUtils.emptyIfNull;
 import static org.hisp.dhis.analytics.DataQueryParams.NUMERATOR_DENOMINATOR_PROPERTIES_COUNT;
 import static org.hisp.dhis.analytics.SortOrder.ASC;
 import static org.hisp.dhis.analytics.SortOrder.DESC;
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -821,53 +827,63 @@ else if ( header.getValueType() == ValueType.REFERENCE )
...
+            .collect( Collectors.groupingBy(
+                QueryItem::hasRepeatableStageParams ) );
+
+        Collection<String> repeatableSqlConditions = asSqlCollection( itemsByRepeatableFlag.get( true ), params );
+        Collection<String> nonRepeatableSqlConditions = asSqlCollection( itemsByRepeatableFlag.get( false ), params );
+
+        return (!repeatableSqlConditions.isEmpty() || !nonRepeatableSqlConditions.isEmpty() ? hlp.whereAnd() + " "
+            : "") +
+            Stream.of(
+                joinSql( repeatableSqlConditions, joining( " or ", "(", ")" ) ),
```
larshelge, line 842: I am wondering if this code can be written in a way which is a bit more readable. I had a hard time understanding what is going on here.

gnespolino: Yes you're right.
Basically I need " and " when at least one of the two collections is not empty, then I need to join those collections using " OR " and " AND " respectively. Streams is because if collections are empty I don't need neither "(" and ")". I'll refactor it in a more readable way

maikelarabori (approved):

d-bernat (approved):

PR #10138, opened 2022-03-15
fix: Do not consider the same attribute in TEI and enrollment as duplicate [DHIS2-12486]

enricocolasante:


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
```diff
@@ -0,0 +1,274 @@
...
+        programInstance = createProgramInstance( program, tei, orgUnit );
+        programInstance.setAttributeValues( Collections.singleton( attributeValue ) );
+        trackedEntityAttributeValue = createTrackedEntityAttributeValue( 'A', tei, uniqueAttribute );
+    }
+
+    @Test
+    void verifySupplierWhenNoUniqueAttributeIsPresentInTheSystem()
+    {
+        // given
+        TrackerPreheat preheat = new TrackerPreheat();
```
teleivo, line 118: You could move that line into the setup if every test needs a fresh preheat

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
```diff
@@ -0,0 +1,274 @@
...
+        when( trackedEntityAttributeService.getAllUniqueTrackedEntityAttributes() )
+            .thenReturn( Collections.emptyList() );
+
+        // when
```
teleivo, line 122: This might be a matter of style, I would not add the comments for given, when, then.
For me using newlines and grouping the code into arrange, act, assert is sufficient to see what is being tested and what the expectations are.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
```diff
@@ -0,0 +1,274 @@
...
+        TrackerImportParams importParams = TrackerImportParams.builder()
+            .trackedEntities( Collections.singletonList( trackedEntity() ) )
+            .enrollments( Collections.singletonList( enrollment( TEI_UID ) ) )
+            .build();
+
+        // when
+        this.supplier.preheatAdd( importParams, preheat );
+
+        // then
+        assertThat( preheat.getUniqueAttributeValues(), hasSize( 0 ) );
```
teleivo, line 145: Why does this one return 0 unique attributes, and the test below 2 even both say ...HaveTheSameUniqueAttribute?

teleivo (commented):
minor suggestions, and one question (after which I can give my ✅  😋)

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/UniqueAttributeSupplierTest.java
enricocolasante: I have changed the test names to differentiate the ones that found unique attributes and the ones that do not find any.
It is not easy to explain everything in the name though.

We are going to have a validation error if we are using the same value for the same unique attributes for 2 different TEIs or enrollments in different TEIs, and we are NOT going to have a validation error if we have the same value for the same unique attribute for a TEI and an enrollment in this TEI, or if the TEI in the payload is the same as the TEI in the DB (in this case it is just an update)

teleivo: uf 😅 yes, that is hard to put in a method name 😂 thank you for explaining!

PR #10181, opened 2022-03-17
feat: enhanced conditions for analytics [DHIS2-12838]

gnespolino:
This PR implements the "enhanced" conditions on analytics dimensions/filters.

it's based on a flag to be enabled in the request, so frontend can still use the old behavior and this PR can be merged being backward compatible.

enhancedConditions=false (old behavior):

```
dimension=a:GT:20:LT:40&dimension=b:GT:1:LT:5
```

means

``` 
a>20 and a<40 and b>1 and b<5 
```

enhancedConditions=true (new behavior):

```
dimension=a:GT:20:LT:40_OR_b:GT:1:LT:5&dimension=c:EQ:test
```

means

``` 
((a>20 and a<40) or (b>1 and b<5)) and c = "test"
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
@@ -896,7 +896,6 @@ public abstract class AbstractJdbcEventAnalyticsManager
 
     private String getItemsSqlForEnhancedConditions( EventQueryParams params, SqlHelper hlp )
     {
-
         Map<UUID, String> sqlConditionByGroup = params.getItems()
             .stream()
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/EventDataQueryRequest.java
```diff
@@ -66,9 +66,9 @@
 
     private Date endDate;
 
-    private Set<String> dimension;
+    private Set<Set<String>> dimension;
```
maikelarabori, line 69: Does this support things like `programStageUid.ElementUid`, or only single `uid`? Is it also been taken into consideration during query time?

gnespolino: Yes everything is as it was before, the only difference is that we now split any dimension param on _OR_ and assign the same group to it

d-bernat (approved):

PR #10226, opened 2022-03-21
fix: TEI sort not working in Tracker Capture app [ 2.35 ]

lucaCambi77:
fix registration date ( aka created ) not sorting. It is currently sorting only ascending, so a fix has been done to create a dynamic order condition

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
@@ -136,6 +136,10 @@ public class HibernateTrackedEntityInstanceStore
 
     private static final String GT_EQUAL = " >= ";
 
+    private static final String UID_VALUE_SEPARATOR = ":";
+
+    private static final String UID_VALUE_PAIR_SEPARATOR = ";@//@;";
+
     // -------------------------------------------------------------------------
     // Dependencies
     // -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -1285,10 +1285,6 @@ private void extractOrderField( boolean innerOrder, TrackedEntityInstanceQueryPa
 
         if ( prop.length == 2 && (prop[1].equals( "desc" ) || prop[1].equals( "asc" )) )
         {
-            if ( prop[0].equals( CREATED_ID ) )
```
stian-sandvold, line 1288: Will it now be sorting on created instead of tei id? Does that solve the problem? We made this change as a means to optimize. Would be nice with a short description on how this solves the problem :)

lucaCambi77: it will sort on the registration date as required by the front end, that is the issue. I see in later versions this has been removed as well. Coming to performance, I didn't know about that, but if so we should remove from the front end the possibility to sort by registration date?

stian-sandvold: Since registration date should have the same order as the primary key (trackedentityinstanceid) and the tables are naturally sorted by their primary key, we wanted to translate the registration date/created date to trackedentityinstanceid instead, so avoid the some times extra cost of sorting.

So I'm not sure I understand the problem of the frontend, if this change fixes the issue - I thought it didnt sort at all? It should still change order if you switch between ASC and DESC, correct?

lucaCambi77: ok, I see the problem now. Created is already translated properly but then is not sorted. So I think we should do like in newer versions and add a static translation.

zubaira (approved):

PR #10262, opened 2022-03-23
chore: Clean up service tracker and modules using it [TECH-945]

enricocolasante:


dhis-2/dhis-services/dhis-service-tracker/pom.xml
```diff
@@ -15,53 +15,161 @@
...
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.hisp.dhis</groupId>
             <artifactId>dhis-service-core</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
```
teleivo, line 43: why do we need the dependency but exclude all its dependencies? shouldnt that be resolved since all of the modules have the same parent with the same dependency versions?

enricocolasante: Removed, this is how I really test that anything is coming from other dependencies.
Just sneaky slipped into the PR

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/job/TrackerImportParamsSerdeTest.java
```diff
@@ -64,9 +65,9 @@ protected void initTest()
     {
     }
 
+    @SneakyThrows
```
teleivo, line 68: Even its name does not give me confidence in that I want to use it :joy: I know this is a test but I would not use it. I don't want sneaky code. Even if its annoying, I like it obvious 😋 Also its not in https://github.com/dhis2/wow-backend/blob/master/guides/lombok.md#allowed-annotations

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationServiceTest.java
```diff
@@ -60,7 +60,6 @@
 import org.hisp.dhis.tracker.report.ValidationErrorReporter;
 import org.hisp.dhis.tracker.validation.hooks.AbstractTrackerDtoValidationHook;
 import org.hisp.dhis.user.User;
-import org.jetbrains.annotations.NotNull;
```
teleivo, line 63: should we ban this one? It was confusing to me to get these different NotNull;NonNull annotations suggested by the IDE.

teleivo (approved):
I would not use SneakyThrows, even in tests.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationServiceTest.java
enricocolasante: Done :)

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/job/TrackerImportParamsSerdeTest.java
enricocolasante: I used it just because the exception here is coming from a subdependency and it is messing up what I am doing.
But I agree.
Just using Exception instead

teleivo (approved):

jbee (approved):

PR #10269, opened 2022-03-23
fix: add back csv for metadata export (2.38)

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12825

### DIFF
--- a/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
+++ b/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) 2004-2021, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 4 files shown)


mortenoh (approved):

dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
```diff
@@ -0,0 +1,193 @@
...
+
+        CsvSchema.Builder schemaBuilder = CsvSchema.builder()
+            .setUseHeader( true );
+
+        // build schema
+        for ( Node child : rootNode.getChildren() )
+        {
+            if ( child.isCollection() )
+            {
+                if ( !child.getChildren().isEmpty() )
```
jbee, line 86: could be combined - I think there is an intelliJ inspection that should flag this for you as well

jbee (approved):

dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
vietnguyen: fixed

PR #10270, opened 2022-03-23
fix: Update identifiers of data elements to uid

stian-sandvold:


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/mapper/ProgramStageInstanceMapper.java
```diff
@@ -65,7 +68,41 @@ public ProgramStageInstance map( Event event )
...
+        if ( psi == null )
+        {
+            psi = mapForInsert( event );
+        }
+        else
+        {
+            mapForUpdate( event, psi );
+        }
+
+        // We run the psi through this method to make sure data value
```
larshelge, line 80: Perhaps move the Javadoc to the `mapDataValueIdentifier` method, to avoid large inline Javadoc inside a method.

ameenhere (approved):

lucaCambi77 (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/mapper/ProgramStageInstanceMapper.java
```diff
@@ -65,7 +68,40 @@ public ProgramStageInstance map( Event event )
...
+     */
+    private ProgramStageInstance mapDataValueIdentifier( ProgramStageInstance psi,
+        Map<String, DataElement> dataElementMap )
+    {
+        for ( EventDataValue dv : psi.getEventDataValues() )
+        {
+            DataElement de = dataElementMap.get( dv.getDataElement() );
+            if ( de != null )
+            {
+                dv.setDataElement( de.getUid() );
```
teleivo, line 101: should we do something else in case there is a `dv.getDataElement()` (non-null) but it was not found in the `dataElementMap`?

teleivo (commented):
Does this warrant a test? Since I think this relates to a bug we had/have

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/mapper/ProgramStageInstanceMapper.java
stian-sandvold: If de is null, it means it did not exists in the database, and it should already be caught by the validation, so if we got to this code with that case, we would have to ignore it anyway, we can't really recover from it in that case.

I'll have a look about making a small test for this

teleivo (approved):

larshelge:
@stian-sandvold the integration-h2-tests seem to be stalling.

PR #10274, opened 2022-03-24
feat: add back new field filtering

mortenoh:
## Summary

PR that adds back the new field filtering (which was removed in 2.38)

* [x] Updates all controllers to only use the new field filter
* [x] Streaming JSON (using `JsonGenerator`)
* [x] Update `property.simple` to correctly identify simple types (needed for new CSV)
* [x] Support zip/gzip for `FieldFilterParams` and `MetadataExportParams`using `MessageConverter`


netroms (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/MessageConverterUtils.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @author Morten Olav Hansen
+ */
+public final class MessageConverterUtils
+{
+    private MessageConverterUtils()
+    {
+    }
+
+    public static final ImmutableList<MediaType> XML_SUPPORTED_MEDIA_TYPES = ImmutableList.<MediaType> builder()
```
jbee, line 50: We could just use `List.of` - these are equally effectively immutable.

jbee (approved):
Just a note on use of `ImmutableList`.

PR #10276, opened 2022-03-24
chore: Clean up all support modules to remove all the warnings from analyze [TECH-801]

enricocolasante:
There were some leftovers from the support modules clean up. Now `mvn dependency:analyze` returns no warning for support modules.

I needed to add some ignore to make it work and there is probably some discussion in the future around that as some of those dependencies indicate bad smells and they shouldn't be there at all.

### DIFF
--- a/dhis-2/dhis-api/pom.xml
+++ b/dhis-2/dhis-api/pom.xml
@@ -64,6 +64,10 @@
       <artifactId>spring-context</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-beans</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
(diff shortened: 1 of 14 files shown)


jbee (approved):

teleivo (approved):
you said there are no warnings after these changes 🎉 thanks a lot for your efforts on this!!!

should we try https://maven.apache.org/plugins/maven-dependency-plugin/examples/failing-the-build-on-dependency-analysis-warnings.html
so your work does not get lost. I also forget to add an explicit dependency when adding an import in a module for something we do not have a dependency yet. IDEs could do a better job at handling that, letting us know 😋 

if we experience to many false positives, we can discuss whether we want to keep it or not. I think its worth a try at least to keep our pom.xmls cleaner

enricocolasante:
> you said there are no warnings after these changes 🎉 thanks a lot for your efforts on this!!!
> 
> should we try https://maven.apache.org/plugins/maven-dependency-plugin/examples/failing-the-build-on-dependency-analysis-warnings.html so your work does not get lost. I also forget to add an explicit dependency when adding an import in a module for something we do not have a dependency yet. IDEs could do a better job at handling that, letting us know 😋
> 
> if we experience to many false positives, we can discuss whether we want to keep it or not. I think its worth a try at least to keep our pom.xmls cleaner

Yes, that is the idea.
I initially thought it wasn't going to be possible but know I can see the light at the end of the tunnel.
But I need to do the same thing for the remaining modules before I can switch the plugin on.

And we will need to discuss some false positive as they are controversial 😬

PR #10282, opened 2022-03-24
feat: check DHIS2 is running on JDK 11+ [DHIS2-12931]

jbee:
Intent here is to check at startup if the java version is sufficient to run DHIS2.

The odd thing is that usually a JVM should already reject running `.class` files compiled for a target to does not support.
Not sure how anyone gets past that point but this might be early enough to detect it and end with an error message that is helpful.

Needs testing by those that have the issue. I don't know how they got to have it.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
@@ -52,7 +52,7 @@ public class DisEnvCompatibilityCheck implements WebApplicationInitializer
             || javaVersion.startsWith( "9" )
             || javaVersion.startsWith( "10" ) )
         {
-            log.error( "DHIS2 requires Java 11+" );
+            log.error( "DHIS 2 requires Java 11+. Please upgrade your JDK version. Startup process was aborted." );
             System.exit( 1 );
         }
     }
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public void onStartup( ServletContext context )
+    {
+        String javaVersion = System.getProperty( "java.version" );
+        log.info( "Java Version: " + javaVersion );
+        if ( javaVersion.startsWith( "1." )
+            || javaVersion.startsWith( "9" )
+            || javaVersion.startsWith( "10" ) )
+        {
+            log.error( "DHIS2 requires Java 11+" );
```
larshelge, line 55: Perhaps add something like "Aborting DHIS 2 startup process, Java JDK 11+ is required" so that local sysamdmins understand what's going on?

larshelge (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
jbee: updated now.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.WebApplicationInitializer;
+
+/**
+ * Checks environment compatibility for DHIS2 at startup.
+ *
+ * @author Jan Bernitt
+ */
+@Slf4j
+@Order( 1 )
+public class DisEnvCompatibilityCheck implements WebApplicationInitializer
```
larshelge, line 44: Should this be `Dhis2EventCompatibilityCheck` ?

jbee: it should have been `Dhis2Env(ironment)CompatibilityCheck`

mortenoh (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DisEnvCompatibilityCheck.java
jbee: name is fined now

PR #10285, opened 2022-03-24
fix: Event import fails when data value is an option and sort_order has gaps [ DHIS2-12933 ] [ patch/2.38.0 ]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/10271


vilkg (approved):
approved by RCB as well.

enricocolasante (approved):

PR #10289, opened 2022-03-24
refactor: Create SingleSetupIntegrationTestBase

jimgrace:
This PR creates `SingleSetupIntegrationTestBase`, a base class for integration tests that have a single setup for the entire class instead of a setup per test.

The use case for this is for tests that need a resource-intensive setup (such as building analytics tables) that is sufficient for all the tests.

As part of this PR, `AnalyticsServiceQueryModifiersTest` has been refactored to use this base class. In the future, tests such as `AnalyticsServiceTest` and `EventAnalyticsServiceTest` can also be refactored to use this base. This will greatly simplify them and make their test cases much easier to add and maintain.


jbee (approved):
Nice!

d-bernat (approved):

PR #10291, opened 2022-03-24
fix: Aggregated queries using wrong aliases [DHIS2-12947]

maikelarabori:
Analytics events aggregated queries are failing because of invalid aliases.
Aggregated queries should always be treated as `group by` queries.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
@@ -33,6 +33,7 @@ import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.collections4.CollectionUtils.emptyIfNull;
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -347,9 +348,25 @@ else if ( ValueType.ORGANISATION_UNIT == queryItem.getValueType() )
...
             else if ( queryItem.getValueType() == ValueType.NUMBER && !isGroupByClause )
             {
-                ColumnAndAlias columnAndAlias = getColumnAndAlias( queryItem, false, queryItem.getItemName() );
+                ColumnAndAlias columnAndAlias;
+
+                // For aggregated queries, we do not prefix the columns as they
+                // will be always "group by" queries.
+                if ( isAggregated )
+                {
+                    columnAndAlias = getColumnAndAlias( queryItem, true,
```
d-bernat, line 357: you can use isAggregate directly

maikelarabori: Right, done

gnespolino (approved):

d-bernat (approved):

PR #10292, opened 2022-03-24
perf: bulk load data elements used in indicator expressions [DHIS2-11100]

jbee:
This is the low effort improvement where we load data elements in one go by their UIDs instead of loading them one at a time by UID.

Since the deletion handler calling this also has a loop over the indicators this still causes multiple queries but just as many as there are indicators. Further the loop in the deletion handler just wants to know if the data element is used. Loading data elements from the database is entirely not needed if a dedicated DB query can be used to find out if data elements exist given a set of UIDs. The proper solution should work that way:

1. collect all UIDs of data elements referenced in all indicators
2. make a single DB exists/count query checking for all data elements by UID

This hopefully will be done by another PR - this easy fix gives us a quick solution that should make the issue much smaller already.


jbee:
Closing this in favour of #10302

PR #10299, opened 2022-03-25
chore: Remove all warnings from depdency:analyze [TECH-801]

enricocolasante:
- Remove all warnings when running `mvn depency:analyze`
- Activate failOnWarnings when a new warning is returned
- Restore dhis-service-audit-consumer module

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/pom.xml
+++ b/dhis-2/dhis-services/dhis-service-administration/pom.xml
@@ -108,7 +108,6 @@
     <dependency>
        <groupId>org.apache.logging.log4j</groupId>
        <artifactId>log4j-slf4j-impl</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.springframework.security</groupId>
--- a/dhis-2/dhis-services/dhis-service-administration/pom.xml
+++ b/dhis-2/dhis-services/dhis-service-administration/pom.xml
@@ -102,8 +102,13 @@
       <artifactId>commons-io</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+       <groupId>org.apache.logging.log4j</groupId>
+       <artifactId>log4j-slf4j-impl</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.springframework.security</groupId>
--- a/dhis-2/dhis-services/dhis-service-audit-consumer/pom.xml
+++ b/dhis-2/dhis-services/dhis-service-audit-consumer/pom.xml
@@ -1,0 +1,62 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  
+  <parent>
+    <groupId>org.hisp.dhis</groupId>
+    <artifactId>dhis-services</artifactId>
+    <version>2.39-SNAPSHOT</version>
+  </parent>
+  
+  <artifactId>dhis-service-audit-consumer</artifactId>
+  <packaging>jar</packaging>
+  <name>DHIS Audit consumer service</name>
+  
+  <dependencies>
+    
+    <!-- DHIS -->
+    
+    <dependency>
+      <groupId>org.hisp.dhis</groupId>
+      <artifactId>dhis-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.hisp.dhis</groupId>
+      <artifactId>dhis-support-artemis</artifactId>
+    </dependency>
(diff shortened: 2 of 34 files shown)


dhis-2/dhis-services/dhis-service-administration/pom.xml
```diff
@@ -102,8 +102,13 @@
...
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+       <groupId>org.apache.logging.log4j</groupId>
+       <artifactId>log4j-slf4j-impl</artifactId>
+      <scope>provided</scope>
```
teleivo, line 111: in the dhis-api/pom.xml its the other way around with slf4j-api provided. why is that?

dhis-2/dhis-services/dhis-service-audit-consumer/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  
+  <parent>
+    <groupId>org.hisp.dhis</groupId>
+    <artifactId>dhis-services</artifactId>
+    <version>2.39-SNAPSHOT</version>
+  </parent>
+  
+  <artifactId>dhis-service-audit-consumer</artifactId>
```
teleivo, line 12: why was it deleted before?

dhis-2/dhis-services/dhis-service-dxf2/pom.xml
```diff
@@ -202,8 +202,13 @@
...
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+       <groupId>org.apache.logging.log4j</groupId>
+       <artifactId>log4j-slf4j-impl</artifactId>
+      <scope>provided</scope>
```
teleivo, line 211: if we define the scopes in the root pom, we would not need to repeat that in each pom AFAIK

dhis-2/dhis-web-api-test/pom.xml
```diff
@@ -209,6 +389,27 @@
       <scope>test</scope>
     </dependency>
 
+    <!--    <dependency>-->
```
teleivo, line 392: I would remove it instead of commenting it if we don't need it

dhis-2/pom.xml
```diff
@@ -821,7 +841,8 @@
                                 <RestrictImports>
                                     <reason>Do not use this json library. Use com.google.gson instead</reason>
```
teleivo, line 842: when should one use com.google.gson? for testing we can/should advocate for https://github.com/dhis2/json-tree 😋 @jbee

teleivo (commented):
a few questions before ✅

dhis-2/dhis-services/dhis-service-audit-consumer/pom.xml
enricocolasante: Error 😬 
It is not directly used anywhere so I thought it was just floating there but this is not the case.
@mortenoh knows more about it and how it is used

dhis-2/dhis-web-api-test/pom.xml
enricocolasante: Done

dhis-2/pom.xml
enricocolasante: Absolutely.
We just cannot ban com.google.gson yet as it is widely used in the system

dhis-2/dhis-services/dhis-service-dxf2/pom.xml
enricocolasante: Done

dhis-2/dhis-services/dhis-service-administration/pom.xml
enricocolasante: I am not sure about this specific one.

Anyway, I think that logs in general is something that we will need to review at some point.
At least for now I managed to remove a lot of redundant logging dependencies that were coming from a lot of different dependencies.

dhis-2/dhis-services/dhis-service-audit-consumer/pom.xml
mortenoh: Yes, this was an error and needs to be merged back immediately. This disabled all audits in DHIS2.

mortenoh (approved):

teleivo (approved):

dhis-2/dhis-services/dhis-service-audit-consumer/src/main/java/org/hisp/dhis/audit/AbstractAuditConsumer.java
```diff
@@ -0,0 +1,88 @@
...
+public abstract class AbstractAuditConsumer
+    implements AuditConsumer
+{
+    protected AuditService auditService;
+
+    protected ObjectMapper objectMapper;
+
+    protected boolean isAuditLogEnabled;
+
+    protected boolean isAuditDatabaseEnabled;
```
jbee, line 51: I think it is better to make these `final` and run the initialisation here instead of the in the inheriting ones.

jbee (approved):

PR #10300, opened 2022-03-25
fix: Fix program rule variable migration when a misconfiguration is present [DHIS2-12884]

enricocolasante:
@zubaira Does it make sense to have `TEXT` as the default value type in case of misconfiguration?

@ameenhere I am just changing the flyway script as it is not released yet. Is that ok?


zubaira:
> @zubaira Does it make sense to have `TEXT` as the default value type in case of misconfiguration?
> 
> @ameenhere I am just changing the flyway script as it is not released yet. Is that ok?

@enricocolasante 
valuetype can be null only if prv does not have any dataelement or TEA and its sourceType is NOT calculated value. In that case setting, valuetype=TEXT should be fine.

zubaira (approved):

stian-sandvold (approved):

ameenhere (approved):
Yes. Since it is not released. It is okay.. But this change should be part of the 2.38.0 release as well. And also make sure, the files are exactly the same in all versions (2.38, 2.38.0 and master). Spaces, semicolons, caps etc etc. Because the checksum is created based on the content, and only if the content is exactly the same, the same checksum will be created.

PR #10304, opened 2022-03-28
build: Make analyze depenencies and check formatting steps required by tests

enricocolasante:
- Transform `check-formatting` workflow into a job in `run-tests` workflow
- Do not run test jobs if `check-formatting` and `check-dependencies` fail
- ~~Add a label check to skip `check-dependencies` job~~
- Add cancel-in-progress configuration for `codeql-analysis` workflow


.github/workflows/run-tests.yml
```diff
@@ -8,8 +8,44 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: maven
+
```
teleivo, line 21: I would remove newlines between steps

.github/workflows/run-tests.yml
```diff
@@ -8,8 +8,44 @@ concurrency:
...
+      - name: Check formatting in core
+        run: mvn speedy-spotless:check --batch-mode --no-transfer-progress --update-snapshots -q -f ./dhis-2/pom.xml
+
+      - name: Check formatting in web
+        run:
+          mvn speedy-spotless:check --batch-mode --no-transfer-progress --update-snapshots -q -f ./dhis-2/dhis-web/pom.xml
+  analyze-dependencies:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.pull_request.labels.*.name, 'skip-check-dependencies')"
```
teleivo, line 30: Why would we want to skip this check? We don't allow skipping the formatting check as well 😋

.github/workflows/run-tests.yml
```diff
@@ -8,8 +8,44 @@ concurrency:
...
+          distribution: temurin
+          cache: maven
+
+      - name: Check formatting in core
+        run: mvn speedy-spotless:check --batch-mode --no-transfer-progress --update-snapshots -q -f ./dhis-2/pom.xml
+
+      - name: Check formatting in web
+        run:
+          mvn speedy-spotless:check --batch-mode --no-transfer-progress --update-snapshots -q -f ./dhis-2/dhis-web/pom.xml
+  analyze-dependencies:
```
teleivo, line 28: I would name it check-dependencies (consistent with check-formatting)

.github/workflows/run-tests.yml
```diff
@@ -8,8 +8,44 @@ concurrency:
...
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: maven
+
+      - name: Check formatting in core
+        run: mvn speedy-spotless:check --batch-mode --no-transfer-progress --update-snapshots -q -f ./dhis-2/pom.xml
```
teleivo, line 23: do we actually need update-snapshots for the formatting check? if we don't then we could save a bit of time/bandwidth; at least for formatting

.github/workflows/run-tests.yml
```diff
@@ -8,8 +8,44 @@ concurrency:
     group: ${{ github.workflow}}-${{ github.ref }}
     cancel-in-progress: true
 jobs:
+  check-formatting:
```
teleivo, line 11: Did you choose separate jobs for formatting & dependencies instead of putting them all in 1 job, since formatting does not need the code to be compiled?

dhis-2/pom.xml
```diff
@@ -707,18 +707,8 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-dependency-plugin</artifactId>
                     <version>${maven-dependency-plugin.version}</version>
-                    <executions>
```
teleivo, line 710: Are you removing it since we are now explicitly calling the plugin in the pipeline? It would still be good to automatically run it locally, so devs get notified earlier (locally) of dependency issues. Currently it would run with `mvn verify` or `mvn install` which I think makes sense.

teleivo (commented):
nice 🎉 I have some minor suggestions and some questions

.github/workflows/run-tests.yml
enricocolasante: I am just being conservative for now.
I am not sure how this will work for the whole team and I want a way to make it not a blocker.
I think that after a while we can evaluate if remove this condition

.github/workflows/run-tests.yml
enricocolasante: Same as before, for now I would maintain them separated and if everything works as we are expecting we can merge them in one single job

.github/workflows/run-tests.yml
enricocolasante: Done

.github/workflows/run-tests.yml
enricocolasante: Done

.github/workflows/run-tests.yml
enricocolasante: Done

dhis-2/pom.xml
enricocolasante: You are right, I put it back and I also removed the condition.
It didn't make sense to have one and the other, skipping with the label but then running it during install wouldn't make much sense.
We are now forcing the PR to be dependencies compliant

.github/workflows/run-tests.yml
enricocolasante: Done, I grouped them together

.github/workflows/run-tests.yml
```diff
@@ -18,15 +19,34 @@ jobs:
           java-version: 11
           distribution: temurin
           cache: maven
+      - name: Check formatting in core
+        run: mvn speedy-spotless:check --batch-mode --no-transfer-progress -q -f ./dhis-2/pom.xml
+      - name: Check formatting in web
+        run: mvn speedy-spotless:check --batch-mode --no-transfer-progress -q -f ./dhis-2/dhis-web/pom.xml
+      - name: Check maven dependencies
+        run: mvn dependency:analyze --threads 2C --batch-mode -no-transfer-progress --update-snapshots -f ./dhis-2/pom.xml
```
teleivo, line 27: I remember you saying dhis-web/pom.xml is a separate root pom. So currently we would not analyze its dependencies, right? If so I think we should we also run the dependency:analyze on it.

.github/workflows/run-tests.yml
```diff
@@ -18,15 +19,34 @@ jobs:
...
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: maven
       - name: Test core
-        run: mvn clean install --threads 1C --batch-mode -no-transfer-progress --update-snapshots -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
-
+        run: mvn clean install --threads 2C --batch-mode -no-transfer-progress --update-snapshots -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
```
teleivo, line 42: Since the `mvn clean install` steps are using the default profile, they will run the `analyze-only` checks. You configured it in the default profile and it binds to the verify phase by [default](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html). Not sure how much that time that takes. We could run `mvn package` [which does everything up to verify](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference). Are there any other things running in verify, that we actually want to run and would otherwise lose?

teleivo (commented):
We would add ~4.5min to our build 😬
https://github.com/dhis2/dhis2-core/runs/5737675933?check_suite_focus=true

Post Setup JDK 11 takes 1.5min because of the formatting check. It took as long before
https://github.com/dhis2/dhis2-core/runs/5738575510?check_suite_focus=true
I did not realize that, just saw that formatting takes ~40s
Could you please check why that is, what is it doing that takes so long. In our test jobs Post Setup JDK 11 takes 0s.

Maybe we should not run dependency analysis as standalone. And run it with our tests instead. This way we don't compile twice. verify phase to which analyze:only is bound to by default, seems so late to me. I wouldn't even want to run the tests. https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html Needs test-compile to have run. We could try binding it to `process-test-classes` which is right before test.

.github/workflows/run-tests.yml
enricocolasante: We should, but to do so we need to fix them before.
I am not including it as part of this PR

vilkg (approved):

teleivo:
Nice 😄  you were able to get rid of the 1.5min of Post Set up JDK 11 in check-formatting. We have the same issue in check-dependencies. Why is it that this happens? We are using the maven cache also in our test jobs and I don't see the same Post Set up JDK 11 timings. Does it have something to do with mavens --update-snapshots?

dhis-2/pom.xml
```diff
@@ -407,6 +407,9 @@
 
         <profile>
             <id>sonarqube</id>
+            <properties>
+                <skipTests>true</skipTests>
```
teleivo, line 411: If Sonarqube needs jacoco coverage report, that would prevent it from getting them.

dhis-2/pom.xml
```diff
@@ -463,6 +466,9 @@
 
         <profile>
             <id>dev</id>
+            <properties>
+                <skipTests>true</skipTests>
```
teleivo, line 470: what is the dev mode actually for? not sure if people expect the tests to run

dhis-2/pom.xml
vilkg: it does for coverage, but the coverage has been broken for a while now. @stian-sandvold made an attempt to fix that, but I believe it was reverted. So I think it's fine to skip tests for now, until the coverage is fixed.

enricocolasante: As soon as the tests are not skipped anymore, the analyze will run and everything will still work fine

dhis-2/pom.xml
enricocolasante: Exactly.
The tests are skipped.
This profile is used to build the image for the api tests

vilkg: I think the profile can be removed if that's the only reason it's there. We can always do -DskipTests and not use a profile. I would agree that `dev` might implicitly mean that unit tests are included. But that's not necessary for API tests.

enricocolasante:
> Nice 😄 you were able to get rid of the 1.5min of Post Set up JDK 11 in check-formatting. We have the same issue in check-dependencies. Why is it that this happens? We are using the maven cache also in our test jobs and I don't see the same Post Set up JDK 11 timings. Does it have something to do with mavens --update-snapshots?

The Post Set up JDK 11 is intermittent.
Now it is not there: https://github.com/dhis2/dhis2-core/runs/5750972904?check_suite_focus=true

check-formatting now is not using the cache at all and this is making Set up and Post Set up close to 0s.

I think now things kind of make sense. We are not running `dependency:analyze-only` if the tests are skipped as this could create problems if the test sources were not compiled.
We are running `dependency:analyze-only` in the tests job but it is taking almost no time

.github/workflows/run-tests.yml
```diff
@@ -49,9 +49,9 @@ jobs:
           distribution: temurin
           cache: maven
       - name: Test core
-        run: mvn test --threads 2C --batch-mode -no-transfer-progress --update-snapshots -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
+        run: mvn test --threads 2C --batch-mode -no-transfer-progress -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
```
teleivo, line 52: maybe `-` missing in `-no-transfer-progress` => `--no-transfer-progress`

.github/workflows/run-tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           distribution: temurin
           cache: maven
       - name: Check maven dependencies
-        run: mvn dependency:analyze --threads 2C --batch-mode -no-transfer-progress -f ./dhis-2/pom.xml
+        run: mvn dependency:analyze --threads 2C --batch-mode -no-transfer-progress --update-snapshots -f ./dhis-2/pom.xml
```
teleivo, line 22: is there a `-` missing in `-no-transfer-progress` => `--no-transfer-progress` ?

.github/workflows/run-tests.yml
```diff
@@ -18,15 +18,46 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: maven
+      - name: Test core
+        run: mvn test --threads 2C --batch-mode -no-transfer-progress -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
```
teleivo, line 52: I think we should use `mvn package` at some point. If we don't then we might have a successful Github PR merged, to find that packaging fails on Jenkins which I am unsure how many devs looks at.

Not sure if package should be used in `Test core` and `Test dhis-web` or only in `Test dhis-web` (which I assume produces the dhis.war). Do we distribute any Java artifact (jar, war) from this repo other than the dhis.war?

dhis-2/build-dev.sh
```diff
@@ -40,8 +40,8 @@ print() {
 print "Building dhis2-core..."
 
 export MAVEN_OPTS="-Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.test.skip=true"
-mvn clean install --batch-mode --no-transfer-progress --threads 1C -Pdev -f "$DIR"/pom.xml -pl -dhis-web-embedded-jetty
-mvn clean install --batch-mode --no-transfer-progress --threads 1C -Pdev -f "$DIR"/dhis-web/pom.xml
+mvn clean install --batch-mode --no-transfer-progress --threads 2C -DskipTests -f "$DIR"/pom.xml -pl -dhis-web-embedded-jetty
```
teleivo, line 43: `-DskipTests` only says don't run tests. Adding `-Dmaven.test.skip=true` will make sure we don't compile tests, which we do not need here 😄

dhis-2/build-nocompr.sh
```diff
@@ -3,5 +3,5 @@
 # Requires maven to be on the classpath
 # Invokes the dev profile which skips tests and disables compression of war artifacts for a speedy build
 
-mvn clean install --batch-mode --no-transfer-progress -Pdev
-mvn clean install --batch-mode --no-transfer-progress -Pdev -f dhis-web/pom.xml -U
+mvn clean install --batch-mode --no-transfer-progress -DskipTests
```
teleivo, line 6: Please add `-Dmaven.test.skip=true`

dhis-2/pom.xml
```diff
@@ -452,48 +454,6 @@
...
-                                <artifactId>asm</artifactId>
-                                <version>${ow2.asm.version}</version>
-                            </dependency>
-                        </dependencies>
-                    </plugin>
-
-                    <plugin>
-                        <groupId>org.zeroturnaround</groupId>
-                        <artifactId>jrebel-maven-plugin</artifactId>
-                        <version>${jrebel-x-maven-plugin.version}</version>
```
teleivo, line 480: can we also remove the plugin version var? not sure if this plugin is still needed

dhis-2/pom.xml
```diff
@@ -854,6 +815,24 @@
...
+                    <execution>
+                        <id>analyze</id>
+                        <goals>
+                            <goal>analyze-only</goal>
+                        </goals>
+                        <configuration>
+                            <failOnWarning>true</failOnWarning>
+                            <skip>${skipTests}</skip>
+                        </configuration>
+                        <phase>test-compile</phase>
```
teleivo, line 832: I think it has to come after phase test-compile https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html
that would be process-test-classes https://maven.apache.org/ref/3.8.5/maven-core/lifecycles.html

.github/workflows/run-tests.yml
```diff
@@ -8,7 +8,7 @@ concurrency:
     group: ${{ github.workflow}}-${{ github.ref }}
     cancel-in-progress: true
 jobs:
-  unit-test:
+  check-dependencies:
```
teleivo, line 11: I thought you bound the dependency:analyze-only to a phase after test-compile, so it runs during `mvn test`. In that case we would not need separate jobs. I do like having explicit jobs for this but we worked hard to get the build stable at around 15min. Adding ~5min back on the build doesnt feel good 😬 So if we can run the check before our tests but within the test jobs, that would hopefully reduce that time to its necessary minimum. As we wouldnt setup JDK, cache, compile again https://github.com/dhis2/dhis2-core/pull/10304#pullrequestreview-925471526

.github/workflows/run-tests.yml
enricocolasante: Done

dhis-2/build-dev.sh
enricocolasante: Reverting the remove of dev profile.
We are doing too many things in this PR and the dev profile is also used in the dhis-web projects.
In dev profile tests are already skipped, I will maintain it that way

dhis-2/pom.xml
enricocolasante: Reverting

dhis-2/pom.xml
```diff
@@ -429,6 +429,16 @@
                             </dependency>
                         </dependencies>
                     </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-compiler-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>default-testCompile</id>
+                                <phase>none</phase>
```
teleivo, line 438: why is that necessary?

.github/workflows/run-tests.yml
teleivo: I agree with a separate job check-dependencies clearly highlighting at what stage the build failed.

We are adding 4.5min though:
Screenshot from 2022-03-30 15-31-51

I value time more than the above. I don't think the benefit of the dedicated job warrants the added time. If a test fails for example, I need to check which test failed. Same with, dependency checks failing the build, I need to check which dependencies caused the build to fail. Thus, navigate to the build logs to see the error. If we are on a tie on this one 😋 I would let the backend team meeting decide.

teleivo: Oh, so your argument is we run formatting before test jobs in either case (with separate dependency-check or not). So you calculate check-dependencies duration - check-formatting duration. We still have 3.5min added time. Since you said the actual time it takes to check dependencies is fast, most of it must be spent setting up JDK, cache and compiling. Since we already do that repeatedly throughout our workflows I would rather not do that again with check-dependencies. Thus, include that in the test jobs

enricocolasante: In this whole process most of the time is used to compile source and compile the tests.
We do it again in every job.
Anyway, I agree that we can remove the check-dependencies job as it is included in the other test jobs.
I would think to put check-formatting back to a separate workflow then. Formatting errors are not so frequent. Do we want one more minute in the build over sometimes not running the test jobs?

teleivo: hard to say 😬 

having test jobs depend on the check-formatting 
pro: we save 3 jobs from being taken of the 20 concurrent job per organization quota in case formatting is wrong
con: we add 1min to our build time for all the builds where formatting is ok

If we assume like you said formatting is mostly ok, then keeping it as a separate workflow/job running in parallel to the tests is probably best. So I would also go back to that. Maybe you can add your improvements into it though. You did adapt it a bit to get rid of the longer JDK post setup step 😄

teleivo (approved):
thank you so much for all your work on this 🥇 🎉

PR #10309, opened 2022-03-28
fix: Fix program rule variable migration when a misconfiguration is present [DHIS2-12884]

enricocolasante:


vilkg:
Original PR was approved and this one is a backport. Merging.

PR #10316, opened 2022-03-28
refactor: Overhaul AnalyticsServiceTest

jimgrace:
This refactor makes `AnalyticsServiceTest` easier to maintain and extend. It keeps exactly the same test cases as before. Other tests may be added in the future in a much simpler way.

### Former design

There were formerly four tests. (Each caused all the analytics tables to be rebuilt):

1. `queryValidationResultTable` counted the validation results in analytics.

2. `testMappingAggregation` covered many test cases, so as not to rebuild the analytics tables for each case. It used the _dataQueryParams_ map (keyed by a string test case name) to hold parameters that could be passed to `analyticsService.getAggregatedDataValueMapping( DataQueryParams params )`. The output was compared with the _results_ map using the same key. It also used the _analyticalObjectHashMap_ map for three test cases that also tested `analyticsService.getAggregatedDataValueMapping( AnalyticalObject object )` by comparing with the same _results_ map.

3. `testGridAggregation` iterated over the same _dataQueryParams_ and _results_ maps as `testMappingAggregation`, but tested `analyticsService.getAggregatedDataValues` instead, using `AnalyticsTestUtils.assertResultGrid` to see if the Grid had the same expected _results_.

4. `testSetAggregation` chose two of the test cases from the _dataQueryParams_ map to test with `analyticsService.getAggregatedDataValueSet` for expected results.

### New design

The new design uses `SingleSetupIntegrationTestBase` as introduced in [pull/10289](https://github.com/dhis2/dhis2-core/pull/10289). A single analytics table build is done for all tests.

The `queryValidationResultTable` test is unchanged.

The other three tests have been refactored into one `@Test` per test case. The test method names include the original test case key strings so that the lineage of each test can be traced to the old code (except that two test names have been corrected, see below). Notes:

- Objects that are needed by two or more tests and/or are needed for the analytics build are created during setup. Objects that can be created after the analytics build and are used by only one test are created within that test (`Indicators` and `Visualizations`).

- All these tests call `assertDataValues`, which tests both `analyticsService.getAggregatedDataValueMapping` (formerly by `testMappingAggregation`) and `analyticsService.getAggregatedDataValues` (formerly by `testGridAggregation`).

- The three test cases that had also `analyticalObjectHashMap` entries now call `getDataValueMapping` which calls `analyticsService.getAggregatedDataValueMapping( AnalyticalObject object )`

- The two test cases that were referenced by the old `testSetAggregation` now call `getDataValueSet` which calls `analyticsService.getAggregatedDataValueSet`.

### Other details

- `AnalyticsTestUtils.assertResultMapping` was only used by the old _AnalyticsServiceTest_. It has been removed and `dhis.utils.Assertions.assertMapEquals` is used in its place. A major flaw of _assertResultMapping_ was that it tested only in one direction. It made sure that each result was expected. However it did not check to see that each expected value was in the results. This included the extreme case where a test with no results would be a success!

  The reason for this unfortunate behavior is that the value keys returned by the two overloads of `analyticsService.getAggregatedDataValueMapping` are encoded differently depending on whether the argument is `DataQueryParams` or `AnalyticalObject`. There are three test cases test that call both overloads: `deC_ouB_2017_03`, `deA_ouA_2017_Q01`, and `inC_deB_deC_2017_Q01`. What they had done before is to put two value keys in the _results_ map with the same expected value. _AnalyticsTestUtils.assertResultMapping_ would find one value key for one call, and the other value key for the other call, but would never find both keys in any test. In the refactor, these three test methods now explicitly specify the different value key they expect for each of the two different calls. This allows _AnalyticsTestUtils.assertResultMapping_ to be replaced by _assertMapEquals_ which checks in both directions. This closes a major testing loophole.

- Expected results are now coded as `Map<String, Object>` rather than `Map<String, Double>`, so `assertMapEquals` can be used because analytics returns results as `Map<String, Object>`.

- The parameters for `AnalyticsTestUtils` methods `assertResultGrid` and `assertResultSet` are reordered so the expected value is first, following the usual convention.

- The `scenario` string parameter was removed from `AnalyticsTestUtils.assertResultGrid` since it is no longer needed, and from its caller in `EventAnalyticsServiceTest`. (Each scenario is now a different method, and the stack trace now shows which test had the error.)

- Minor changes were made to `EventAnalyticsServiceTest` to accommodate the minor changes in `AnalyticsTestUtils`. Otherwise `EventAnalyticsServiceTest` has not been changed, although it could also be refactored to use `SingleSetupIntegrationTestBase` in the future as well if that is desired. (Currently `EventAnalyticsServiceTest` only builds analytics once, and doesn't have the test case maps the way `AnalyticsServiceTest` did.)

- Two test names were corrected:
. `inE_deA_reRateA_2017_Q01` was renamed to `inE_deA_reRateB_2017_Q01` because it uses reporting rate B.
. `inF_deA_reRateB_2017_Q01` was renamed to `inF_deA_reRateA_2017_Q01` because it uses reporting rate A.

- `Periods` had been created at different points within `setUpTest`. `y2017_mar` duplicated `peMar` and was merged. `y2017` was renamed to `year` for consistency with the other period names.

- `DataQueryParams` are now built one parameter per line which makes them easier to read. This made it more obvious that the former `deA_deB_deD_ouC_ouE_2017_04_params` was built with two identical calls to `withOutputFormat` in different places. This has been fixed.

### Other benefits

The number of tests has increased from 4 to 19, so a failing test will not block as much subsequent testing.

Since analytics tables are only built once instead of four times, the module runs much faster. On my machine it now takes around 2:45 m:ss instead of 7:25.

Although the line count has increased from 719 to 855 lines, the java file size has shrunk from 40K to 33K. This means less code to look at with a more generous use of white space.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceTest.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceTest.java
@@ -27,17 +27,17 @@
  */
 package org.hisp.dhis.analytics.data;
 
+import static java.util.Collections.emptyList;
 import static org.hisp.dhis.expression.Operator.equal_to;
+import static org.hisp.dhis.utils.Assertions.assertMapEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.hisp.dhis.IntegrationTestBase;
+import org.hisp.dhis.SingleSetupIntegrationTestBase;
 import org.hisp.dhis.analytics.AggregationType;
 import org.hisp.dhis.analytics.AnalyticsAggregationType;
 import org.hisp.dhis.analytics.AnalyticsService;
@@ -56,7 +56,6 @@ import org.hisp.dhis.category.CategoryOptionGroupSet;
 import org.hisp.dhis.category.CategoryService;
 import org.hisp.dhis.common.AnalyticalObject;
 import org.hisp.dhis.common.DataDimensionType;
-import org.hisp.dhis.common.Grid;
 import org.hisp.dhis.common.ReportingRate;
 import org.hisp.dhis.common.ValueType;
 import org.hisp.dhis.dataelement.DataElement;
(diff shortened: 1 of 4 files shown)


d-bernat (approved):

maikelarabori (approved):
Nice work @jimgrace :+1:

PR #10325, opened 2022-03-30
fix: dataValue export file name to include start and end date

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12554
- This is to bring back a part of the old feature in import/export app. Previously in the old struts import/export app we return download file with the file name format `dataValuesStartDate_EndDate_facilityName.json.zip`
- However, when moved to new react app, this function is not included in the new controller.
- This PR added back just a part of the format  which includes start date and end date. The facility name/ orgUnit name is not added back because we allow user to select multiple orgUnits for exporting. Therefore, appending all the orgUnit names to the file name seems not a good solution. 
Example : `dataValues_2022-03-01_2022-03-30.json.zip`

Test
- Added mock test and controller test.

### DIFF
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/DataValueControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/DataValueControllerTest.java
@@ -30,14 +30,18 @@ package org.hisp.dhis.webapi.controller;
 import static org.hisp.dhis.webapi.WebClient.Body;
 import static org.hisp.dhis.webapi.utils.WebClientUtils.assertStatus;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.util.Collections;
 import java.util.List;
 
 import org.hisp.dhis.datavalue.DataValue;
 import org.hisp.dhis.datavalue.DataValueService;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/DataValueControllerTest.java
```diff
@@ -118,6 +122,38 @@ public void testAddDataValueWithBody()
...
+
+        HttpResponse response = POST( "/dataValues", body );
+        assertStatus( HttpStatus.CREATED, response );
+        switchToUserWithOrgUnitDataView( "testUser", orgUnitId );
+        String url = "/dataValueSets?orgUnit=" + orgUnitId + "&startDate=2022-01-01&endDate=2022-01-30&dataSet=" + dsId
+            +
+            "&format=json&compression=zip&attachment=dataValues.json.zip";
+        MvcResult dataValueResponse = webRequestWithMvcResult(
+            buildMockRequest( HttpMethod.GET, url, Collections.emptyList(), null, null ) );
+        assertTrue( dataValueResponse.getResponse().getHeader( "Content-Disposition" )
```
jbee, line 153: Wonder if you added the `webRequestWithMvcResult` so you could read the header? In that case this is not needed. You can read headers from the `HttpResonse` returned by the `GET` method.

jbee (approved):

mortenoh (approved):

PR #10330, opened 2022-03-30
fix: NotificationMap capacity cap does not accidentially clear entries [DHIS2-12992]

jbee:
### Summary
The issue was that when capacity was reached for a `JobType` entry list the `removeLast()` to remove the oldest entry was called on the wrong collection. Instead of removing an entry from the list of a particular `JobConfiguration` for that `JobType` the complete list was removed from the map.

This affected the `Notification`s but not the summary objects.

### Automatic Testing
This seemed an issue best tested using unit tests. 
I added a new test for this. 
To allow better testing I also made the capacity a parameter.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/notification/NotificationMap.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/notification/NotificationMap.java
@@ -41,45 +41,65 @@ import org.hisp.dhis.scheduling.JobConfiguration;
 import org.hisp.dhis.scheduling.JobType;
 
 /**
+ * Keeps an ordered list of {@link Notification}s and/or summary {@link Object}s
+ * per {@link JobType} and {@link JobConfiguration} UID.
+ * <p>
+ * For each {@link JobType} the capacity of entries is capped at a fixed maximum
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/notification/NotificationMap.java
```diff
@@ -29,57 +29,82 @@
 
 import static java.util.Arrays.stream;
 import static java.util.Collections.unmodifiableMap;
+import static java.util.stream.Collectors.toMap;
 
-import java.util.Deque;
-import java.util.EnumMap;
-import java.util.LinkedList;
-import java.util.Map;
+import java.util.*;
```
mortenoh, line 34: import collapse

mortenoh (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/notification/NotificationMap.java
jbee: :+1: forgot to set this after my spring clean IDE reinstall - will do now

PR #10347, opened 2022-03-31
chore: reduce visibility, mutability in tracker DHIS2-12563

teleivo:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/DefaultTrackerImporter.java
```diff
@@ -45,49 +42,46 @@
 @Component
 @Primary
 @RequiredArgsConstructor
-public class TrackerImportStrategyImpl implements TrackerImportStrategyHandler
+public class DefaultTrackerImporter implements TrackerImporter
```
enricocolasante, line 45: Just a reminder on this component, as you renamed it in this PR.
I found that the component `DefaultTrackerImporter` having 2 TrackerImporters as fields is weird. We need to mark it as @Primary there are other TrackerImporter components that are really never used but here.
Do `TrackerSyncImporter` and `TrackerAsyncImporter` need to implement TrackerImporter?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportController.java
```diff
@@ -91,7 +88,7 @@
 {
     static final String TRACKER_JOB_ADDED = "Tracker job added";
 
-    private final TrackerImportStrategyHandler trackerImportStrategy;
+    private final TrackerImporter trackerImportStrategy;
```
enricocolasante, line 91: Should this field also be renamed to `trackerImporter`?

enricocolasante (approved):
Just a couple of comments

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/DefaultTrackerImporter.java
teleivo: I did not find a better name (yet) then `DefaultTrackerImporter`.

So what the `DefaultTrackerImporter` does is dispatch to the `TrackerAsyncImporter` or `TrackerSyncImporter` based on the `async` query parameter. All of them implement the `TrackerImporter` interface. `DefaultTrackerImporter` is what we actually use in the controller (via the `@Primary` mechanism). I thought of merging all of them into one `DefaultTrackerImporter` as the code for async and sync is minimal. We would also not need the `@Primary` which is a bit obscure. On the other hand, having it separate makes testing look cleaner. It allows us to reuse them separately, not sure if we ever need that 🤷🏻 

What would you prefer?
* keep structure as is, with a better name. Please suggest one 😂 
* merge them into a single `TrackerImporter` less naming for us to do, and no `@Primary` 😋

enricocolasante: I would suggest just to make `TrackerAsyncImporter` and `TrackerSyncImporter` not implementing the TrackerImporter interface.
Doing that we can just remove the `@Primary` annotation.
Then we could maybe rename the 2 importer with some other name that will make it clear that they shouldn't implement the `TrackerImporter` interface.

enricocolasante: But anyway, I think is something that we can have on our mind and if we ever find good names we can change them

PR #10349, opened 2022-03-31
fix: Paging issues in event/enrollment analytics [2.39-dhis2-12897-2]

d-bernat:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
```diff
@@ -0,0 +1,78 @@
...
+     * The default is true (always total pages flag activated).
+     */
+    private boolean totalPages = true;
+
+    public void checkAndMaybeModifyPagingCriteria( int analyticsMaxLimit )
+    {
+        if ( isPaging() )
+        {
+            setPageSize( getPageSize() > analyticsMaxLimit ? analyticsMaxLimit : getPageSize() );
+
```
maikelarabori, line 69: Lines 69 to 71 can be removed I think (including the `setPaging( true )` call).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    private boolean paging = true;
+
+    /**
+     * The paging parameter. When set to false we should not count total pages.
+     * The default is true (always total pages flag activated).
+     */
+    private boolean totalPages = true;
+
+    public void checkAndMaybeModifyPagingCriteria( int analyticsMaxLimit )
```
maikelarabori, line 64: Maybe you could rename it to `definePageSize`, because this is what this method is doing.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
```diff
@@ -0,0 +1,78 @@
...
+    {
+        if ( isPaging() )
+        {
+            setPageSize( getPageSize() > analyticsMaxLimit ? analyticsMaxLimit : getPageSize() );
```
maikelarabori, line 68: If `pageSize` is set to null for any reason, this line will throw an NPE. Maybe would be nice to add a null check?

maikelarabori (commented):
Hi @d-bernat, just a few small comments left.

Thanks!
Maikel

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
d-bernat: done

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
d-bernat: done

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticsPagingCriteria.java
d-bernat: done

maikelarabori (approved):
Thanks, looks good! :+1:

gnespolino (approved):

PR #10353, opened 2022-03-31
fix: Upgrade to latest Spring version

netroms:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCache.java
```diff
@@ -64,17 +70,27 @@ public AnalyticsCache( final CacheProvider cacheProvider,
...
-        log.info( String.format( "Analytics server-side cache is enabled with expiration time: %d s",
-            initialExpirationTime ) );
+            log.info( String.format( "Analytics server-side cache is enabled with expiration time: %d s",
+                initialExpirationTime ) );
+        }
     }
 
     public Optional<Grid> get( final String key )
     {
+        initializeCache();
```
maikelarabori, line 92: Do we need to call `initializeCache()` before every method call?
Could we call `initializeCache()` only once inside the `constructor`, and remove all others calls inside each method?

maikelarabori (approved):

maikelarabori:
Regarding the AnalyticsCache I added a small comment, other than that it looks good to me.

Cheers,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/cache/AnalyticsCache.java
netroms: Unfortunately not, this is the reason for the refactor since we can't do database queries in bean constructors.

mortenoh (approved):

PR #10372, opened 2022-04-04
chore: Create separate workflow for integration h2 tests

enricocolasante:
This is one first attempt to reduce the running time of tests in the build.
At the moment we are including tests that are using h2 as a DB in the unit tests.
The idea is to review the h2 tests (understand if they need the DB at all and how we can reduce Spring scanning of the classes to create the Context) and run the h2 tests in a separate workflow so it is easier to see if we are making progress reducing the running time.

Separating the workflow reduces the overall time to run a PR, it should be around 12 minutes now, against 15-16 minutes before but of course we have one more job that is running and consuming resources.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -35,12 +35,6 @@ jobs:
           cache: maven
       - name: Run integration tests
         run: mvn --threads 2C test --batch-mode --no-transfer-progress -Pintegration -DexcludeAnalyticsTests=true -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
-      - name: Archive surefire reports
-        uses: actions/upload-artifact@v2
-        with:
-          name: surefire-reports
-          path: "**/target/surefire-reports/TEST-*.xml"
-          retention-days: 10
 
   integration-h2-test:
     runs-on: ubuntu-latest
@@ -54,12 +48,6 @@ jobs:
           cache: maven
       - name: Run integration h2 tests
         run: mvn --threads 2C test --batch-mode --no-transfer-progress -PintegrationH2 -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
-      - name: Archive surefire reports
-        uses: actions/upload-artifact@v2
-        with:
-          name: surefire-reports
-          path: "**/target/surefire-reports/TEST-*.xml"
-          retention-days: 10
 
   integration-test-analytics:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/IntegrationH2Test.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.Tag;
+
+/**
+ * @author Morten Olav Hansen <mortenoh@gmail.com>
```
enricocolasante, line 38:
```suggestion
 * @author Enrico Colasante
```

.github/workflows/run-tests.yml
```diff
@@ -42,6 +42,25 @@ jobs:
...
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: maven
+      - name: Run integration h2 tests
+        run: mvn --threads 2C test --batch-mode --no-transfer-progress -PintegrationH2 -f ./dhis-2/pom.xml -pl -dhis-web-embedded-jetty
+      - name: Archive surefire reports
+        uses: actions/upload-artifact@v2
+        with:
+          name: surefire-reports
```
teleivo, line 60: We don't need them anymore. I wanted to remove the other surefire report archive steps as well.

teleivo: I forgot to say 😬 : unless you want to use the xml reports for further analysis.

teleivo:
Have you checked test count? Does it match what you expect?

Looking at the overall timings it looks a bit suspicious

image

That if we would add unit + h2 integration tests > 20min. Of course now each of these jobs has some setup cost which maybe amounts to 2min. Did the "unit-test" job previously take 18min sometimes?

enricocolasante:
> Have you checked test count? Does it match what you expect?
> 
> Looking at the overall timings it looks a bit suspicious
> 
> image
> 
> That if we would add unit + h2 integration tests > 20min. Of course now each of these jobs has some setup cost which maybe amounts to 2min. Did the "unit-test" job previously take 18min sometimes?

Yes, the unit-test job lately was the slowest of all and it could take more than 15 minutes.
If you look at the unit-test job now, you can see that the test-core step takes less than 4 minutes. The tests themselves are really fast, the whole compiling overhead that makes the job slower.

In addition, the idea would be to cleanup the h2 tests setup, as for now every module is always scanning all the dependencies to build the context. You can see that the time for the h2 tests are increasing going down the modules, this is because the biggest part of the time is spent to scan to build the context, and modules with more dependencies take more to scan.
Once we clean up the setup, the h2 tests should be much faster and then we can rethink the separation of the jobs

.github/workflows/run-tests.yml
enricocolasante: I don't need them.
Not sure if someone needs them though

teleivo (approved):

jbee (approved):

PR #10377, opened 2022-04-04
fix: Adding back endpoint removed by mistake [DHIS2-13006]

maikelarabori:
**Needs formal approval from Release Control Board.**

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/XmlPathMappingJackson2XmlHttpMessageConverter.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/XmlPathMappingJackson2XmlHttpMessageConverter.java
@@ -52,7 +52,7 @@ public class XmlPathMappingJackson2XmlHttpMessageConverter extends MappingJackso
         Pattern.compile( "/(\\d\\d/)?events(/?$|/.+)" ),
         Pattern.compile( "/(\\d\\d/)?trackedEntityInstances(/?$|/.+)" ),
         Pattern.compile( "/(\\d\\d/)?dataValueSets(/?$|/.+)" ),
-        Pattern.compile( "/analytics/dataValueSet.xml" ),
+        Pattern.compile( "/(\\d\\d/)?analytics(/?$|/.+)" ),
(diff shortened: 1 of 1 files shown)


larshelge:
Should we just allow all analytics requests for XML? Including `/analytics` ? I think that is better, as some might be using the XML there too.

larshelge:
We should also include API versions like for the other.

gnespolino (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/XmlPathMappingJackson2XmlHttpMessageConverter.java
```diff
@@ -52,6 +52,7 @@
         Pattern.compile( "/(\\d\\d/)?events(/?$|/.+)" ),
         Pattern.compile( "/(\\d\\d/)?trackedEntityInstances(/?$|/.+)" ),
         Pattern.compile( "/(\\d\\d/)?dataValueSets(/?$|/.+)" ),
+        Pattern.compile( "/analytics/dataValueSet.xml" ),
```
Philip-Larsen-Donnelly, line 55: Will this handle versioned endpoints without the (\\d\\d/)? pattern.
In general, shouldn't it align with the others?
e.g.
"/(\\d\\d/)?dataValueSet(/?$|/.+)"
?

Philip-Larsen-Donnelly: Sorry - I see Lars already made this comment.

maikelarabori: Done. Versions included.

maikelarabori:
> Should we just allow all analytics requests for XML? Including `/analytics` ? I think that is better, as some might be using the XML there too.

Done, using `/analytics` as regexp.

maikelarabori:
> We should also include API versions like for the other. The `/(\\d\\d/)?` refers to the optional API version @maikelarabori .

Done, versions included.

larshelge (approved):

PR #10379, opened 2022-04-04
chore: split domain and view models DHIS2-12563

teleivo:
## Overview of changes in [DHIS2-12563](https://jira.dhis2.org/browse/DHIS2-12563)

* tracker view (JSON import/export) and domain models are split since users only specify idSchemes in query parameters. The metadata identifier fields in the user-facing import/export JSON are strings containing only the identifier without idScheme.

* Attribute.attribute
* DataValue.dataElement
* Event.programStage
* Event.attributeOptionCombo
* Event.attributeCategoryOptions
* TrackedEntity.trackedEntityType
* Relationship.relationshipType
* {Enrollment :heavy_check_mark:,Event,ProgramOwner}.program
* {Enrollment,Event,TrackedEntity,ProgramOwner}.orgUnit

will become idScheme aware.

## This PR

* split view/domain model
  * `webapi.controller.tracker.imports` maps from `...webapi.controller.tracker.view` (JSON) to `...tracker.domain`
  * `webapi.controller.tracker.exports` maps from `org.hisp.dhis.dxf2.events.` to `...webapi.controller.tracker.view` (JSON)
* `...tracker.domain.Enrollment.program` is now a `MetadataIdentifier` instead of a `String`
* `AttributeValues` and `Attributes` are mapped onto `Program` and `Enrollment.program` during Preheat so we can match `Enrollment.program` using idScheme `ATTRIBUTE`


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/EnrollmentMapper.java
```diff
@@ -0,0 +1,65 @@
...
+    default org.hisp.dhis.tracker.domain.MetadataIdentifier from( String identifier,
+        @Context TrackerIdSchemeParams idSchemeParams )
+    {
+
+        TrackerIdScheme idScheme = idSchemeParams.getProgramIdScheme().getIdScheme();
+        if ( idScheme == TrackerIdScheme.ATTRIBUTE )
+        {
+            return MetadataIdentifier.ofAttribute( idSchemeParams.getProgramIdScheme().getAttributeUid(), identifier );
+        }
+        return MetadataIdentifier.of( idSchemeParams.getProgramIdScheme().getIdScheme(), identifier, null );
```
enricocolasante, line 63: As this if is probably going to appear in a lot places, maybe it will make sense at some point to have a static constructor in `MetadataIdentifier` that takes a TrackerIdSchemeParam and an identifier, and put this if logic in there.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/domain/MetadataIdentifier.java
```diff
@@ -0,0 +1,236 @@
...
+ */
+@Value
+@AllArgsConstructor( access = AccessLevel.PRIVATE )
+public class MetadataIdentifier
+{
+
+    /**
+     * Represents the idScheme the {@link #identifier} is in.
+     */
+    @JsonProperty
```
enricocolasante, line 60: A `MetadaIdentifier` is basically a `TrackerIdSchemeParam` with an identifier associated with it, is this right?
Instead of having attributeValue, we have attributeUid in it, it is just reversed logic

Just wondering if it would make sense then to have `TrackerIdSchemeParam` as a field in this class.
Does it make sense? And would it make the code easier?

enricocolasante (commented):
It is a long PR :)
A couple of question that will help me go through the rest of the code

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/EnrollmentMapper.java
teleivo: Yes, sounds good 👌🏻I wasn't sure about the dependencies (which of the two should know of each other) so postponed it. I will think about it when working on the next field.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/domain/MetadataIdentifier.java
teleivo: TrackerIdSchemeParam represents the query parameters for idSchemes like `idScheme, programIdScheme, ...` https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/new-tracker.html#request-parameters So it will either be UID, CODE, NAME and if its ATTRIBUTE it will be idScheme attribute and the attribute UID. It does not identify metadata.

MetadaIdentifier is identifying metadata using an idScheme and an actual identifier. Identifiers are the UID, code, name values or for idScheme ATTRIBUTE its the attribute UID and value.

So they are distinct. Also the TrackerIdSchemeParam is referring to the entire import. The MetadataIdentifier is referring to one single metadata instance.

Before we had a TrackerIdentifier which was mixing the two things, which I think is confusing.

I also think migrating the next fields, will highlight some patterns/shortcomings, so I can adapt the API later on.

enricocolasante: I agree and understand that they are distinct.
Talking in DHIS2 language, could we say that `TrackerIdentifier` is in fact a `TrackerIdSchemeParamInstance`?

To be honest, it is just me thinking out loud trying to understand how he code works.
Anyway, I agree that adding the next fields will highlight how to proceed

teleivo: no worries, any question is welcome 👍🏻 

I would not say so, but I might not have fully understood the DHIS2 language 😋 

Reason being is that TrackerIdSchemeParam is a configuration of the import. The MetadataIdentifier is an identifier of metadata (encapsulating the identifier idScheme).

You could in theory create a Set<Enrollment> with enrollment.programs using different idSchemes.

enricocolasante (approved):

lucaCambi77 (approved):

PR #10380, opened 2022-04-04
fix: (2.38) make flyway migration idempotent

zubaira:


gnespolino (approved):

ameenhere (changes requested):
2.38, 2.38.0 and master should be exactly the same. In terms of what is the new script version you are changing to.

zubaira:
> 2.38, 2.38.0 and master should be exactly the same. In terms of what is the new script version you are changing to.

raising PR for 2.38.0

ameenhere (approved):

PR #10381, opened 2022-04-04
fix: make smscode flyway migration idempotent

zubaira:


gnespolino (approved):

ameenhere (approved):

ameenhere (changes requested):
2..38.40 seems to be used alreaady in 2.38 branch. And in the wow-backend seems to be reserved already.

ameenhere (approved):

PR #10424, opened 2022-04-06
feat: Add parameter to fetch trigram indexable attributes [DHIS2-12048] (2.39)

ameenhere:
This PR adds a new query parameter **indexableOnly** on the `api/trackedEntityAttributes` that can be used to filter based on trigram indexable trackedEntityAttributes. The new query parameter works like an additional filter if provided with more filter parameters.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
```diff
@@ -396,7 +397,7 @@ else if ( (property.isSimple() || property.itemIs( PropertyType.REFERENCE ))
 
         List<String> fields = Lists.newArrayList( contextService.getParameterValues( "fields" ) );
         List<String> filters = Lists.newArrayList( contextService.getParameterValues( "filter" ) );
-        forceFiltering( filters );
+        forceFiltering( new WebOptions( rpParameters ), filters );
```
teleivo, line 400: `rpParameters` is not a clear var name to me (I assume the prefix rp  = request parameter ; with again parameters as suffix). What do you think about simply calling the var `parameters`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -223,4 +225,29 @@ private Map<String, String> getRequiredValues( TrackedEntityAttribute attr, Map<
         return result;
     }
 
+    // --------------------------------------------------------------------------
+    // Hooks
```
teleivo, line 229: Not sure if we have a "standard"/"guideline" for writing such comments. IMHO these do not add anything useful, other than us having to scroll more 😋

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -223,4 +225,29 @@ private Map<String, String> getRequiredValues( TrackedEntityAttribute attr, Map<
...
+        }
+
+        if ( filters.stream().anyMatch( f -> f.startsWith( "id:" ) ) )
+        {
+            throw new IllegalArgumentException(
+                "indexableOnly parameter cannot be set if a separate filter for id is specified" );
+        }
+
+        StringBuilder sb = new StringBuilder();
+        sb.append( "id:in:" );
```
teleivo, line 246: you should be able to pass that string into the constructor https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#StringBuilder(java.lang.String)

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
```diff
@@ -0,0 +1,231 @@
...
+    @Test
+    void getIndexableOnlyAttributes()
+    {
+
+        JsonObject json = GET( "/trackedEntityAttributes?indexableOnly=true" ).content( HttpStatus.OK );
+
+        assertFalse( json.isEmpty() );
+
+        Set<String> expectedAttributeSet = new HashSet<>();
+        expectedAttributeSet.add( "\"AttributeA\"" );
```
teleivo, line 155: if these tests will only run on java 11 I would use [Set.of](https://docs.oracle.com/javase/9/docs/api/java/util/Set.html#of-E...-)

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    @Test
+    void getIndexableOnlyAttributes()
+    {
+
+        JsonObject json = GET( "/trackedEntityAttributes?indexableOnly=true" ).content( HttpStatus.OK );
+
+        assertFalse( json.isEmpty() );
+
+        Set<String> expectedAttributeSet = new HashSet<>();
```
teleivo, line 154: I think a short var name like `expected` is good enough => `assertAttributeList(json, expected)`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -223,4 +225,29 @@ private Map<String, String> getRequiredValues( TrackedEntityAttribute attr, Map<
...
+        Set<TrackedEntityAttribute> indexableTeas = trackedEntityAttributeService
+            .getAllTrigramIndexableTrackedEntityAttributes();
```
teleivo, line 248: could this return an empty set? and if so would the filter that we are constructing here be valid and its behavior be ok for the user? I haven't seen a test for this case (which does not mean that there is none 😋 ).

teleivo (commented):
super readable tests 😄 my comments are just minor things I would adjust.
The only question I have before giving my ✅ is at https://github.com/dhis2/dhis2-core/pull/10424/files#diff-9a2d90e1c34918011df57b22a350ac53af33a8efb4a8a32a6b8f8fd24772a0e9R248

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    private static void assertAttributeList( JsonObject actualJson, Set<String> expectedAttributeSet )
+    {
+        assertEquals( actualJson.getArray( "trackedEntityAttributes" ).size(), expectedAttributeSet.size() );
+        int i = 0;
+        while ( i < expectedAttributeSet.size() )
+        {
+            assertTrue( expectedAttributeSet.contains(
+                actualJson.getArray( "trackedEntityAttributes" ).get( i ).asObject().get( "displayName" ).asObject()
```
teleivo, line 226: @jbee maybe for this PR and also because I'd like to learn 😋 do you have a suggestion of how we could write this differently now that you added some new capabilities to json-tree?

My understanding is that we have a Set<String> of expected attribute `displayName`s and a JSON `trackedEntityAttributes` list. Each element in the list is a complex object with a field of `displayName`. We want to make sure that the `trackedEntityAttributes` contains all and only the displayNames in the `expectedAttributeSet`.

jbee: Hm.. yes there are several things that are odd here.

First `get( ...).asObject()` should be written as `getObject(...)`. The `asObject()` is only meant to be used where we get a `JsonValue` handed without having control over the type but we expect it to be a `JsonObject` so we "cast" using `asObject()`. If we drill into a tree we can directly use `getObject(...).

Second, I'd assume the `displayName` is **not** an object but a string. So I'd expect to see `getString("displayName")`.

If a loop was needed here, it should be a `for` loop not a `while` - but I think we can just make an assert instead. 

If I read the test correct this could be a single `assertEquals`
```java
assertEquals( expectedAttributeSet, actualJson.getArray( "trackedEntityAttributes" )
  .viewAsList( e -> e.asObject().getString( "displayName" ) )
  .stream().map( JsonString::string )
  .collect( toSet() )
);
```
Key is the `viewAsList` where we to the equivalent of a `stream().map` just that this is purely a view transformation from the array of objects to the array of the object's display name. Then we have a `JsobList<JsonString>` of the display names which we convert to a `Set<String>` using `.stream().map( JsonString::string ).collect( toSet() )`.

This assert has the benefit of clearly telling you how the sets differ should that be the case. If that is not that important you can use `containsAll`. 

```java
assertTrue( actualJson.getArray( "trackedEntityAttributes" )
  .viewAsList( e -> e.asObject().getString( "displayName" ) )
  .containsAll(JsonString::string, expectedAttributeSet )
);
```
If hopefully becomes clear how this is doing the same just the other way around. Only the true/false is less helpful if it fails.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -223,4 +225,29 @@ private Map<String, String> getRequiredValues( TrackedEntityAttribute attr, Map<
...
+        sb.append( indexableTeas.stream().map( tea -> tea.getUid() ).collect( Collectors.joining( ",", "[", "]" ) ) );
```
teleivo, line 249: this line has a smell with an easy fix. see sonarqube analysis.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
ameenhere: Its an existing var name used throughout that class. I did'nt want to rename the var name for the whole class. But we could call it "Request Path Parameters" if that makes more sense for rpParameters 😃

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
ameenhere: Yea agreed. It was part of the copy paste from another Controller class that implemented the hooks. Removed it here.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
ameenhere: Done. This is only from 2.38 onwards, so Java 11 it is.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
ameenhere: Done!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
ameenhere: In theory it shouldnt return an empty set. But to make it cleaner, its better to handle it. Good catch! Also added a test case for that.

teleivo (approved):

zubaira (approved):

PR #10451, opened 2022-04-08
fix: ORG_UNIT_CODE Generation fails [ DHIS2-13032 ]

lucaCambi77:
for segments that do not have a pattern (e.g. ORG_UNIT_CODE), the pattern is serialized as null instead of an empty string
This fix adds a null safe check for those cases

### DIFF
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
@@ -77,8 +77,6 @@ class TrackedEntityAttributeControllerTest extends DhisWebSpringTest
             .content( TestUtils.convertObjectToJsonBytes( trackedEntityAttribute ) ) )
             .andExpect( status().is( HttpStatus.SC_CREATED ) );
 
-        dbmsManager.clearSession();
-
         mvc.perform( get( TrackedEntityAttributeSchemaDescriptor.API_ENDPOINT + "/" + trackedEntityAttribute.getUid()
             + "/generateAndReserve" ).param( "ORG_UNIT_CODE", "A030101" ).session( session ) )
             .andExpect( status().isOk() );
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
@@ -43,15 +43,14 @@ import org.hisp.dhis.webapi.utils.TestUtils;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.mock.web.MockHttpSession;
-import org.springframework.transaction.support.TransactionTemplate;
 
+/**
+ * @author Luca Cambi
+ */
 class TrackedEntityAttributeControllerTest extends DhisWebSpringTest
 {
     @Autowired
     protected DbmsManager dbmsManager;
-
-    @Autowired
-    protected TransactionTemplate transactionTemplate;
 
     @Test
     void shouldGenerateRandomValuesOrgUnitCodeAndRandom()
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
@@ -33,7 +33,6 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 
 import org.apache.http.HttpStatus;
 import org.hisp.dhis.common.Objects;
-import org.hisp.dhis.dbms.DbmsManager;
 import org.hisp.dhis.schema.descriptors.TrackedEntityAttributeSchemaDescriptor;
 import org.hisp.dhis.textpattern.TextPattern;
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -223,4 +189,22 @@
...
+        TrackedEntityAttribute trackedEntityAttribute = trackedEntityAttributeService.getTrackedEntityAttribute( id );
+
+        if ( trackedEntityAttribute == null )
+        {
+            throw new WebMessageException( WebMessageUtils.notFound( TrackedEntityAttribute.class, id ) );
+        }
+
+        if ( trackedEntityAttribute.getTextPattern() == null )
+        {
+            throw new WebMessageException( WebMessageUtils.badRequest( "Attribute does not contain pattern." ) );
```
teleivo, line 204: badRequest and probably notFound are already statically imported, so you can remove `WebMessageUtils.`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -120,35 +113,16 @@
...
     }
 
     @GetMapping( "/{id}/requiredValues" )
     @ApiVersion( { DhisApiVersion.DEFAULT, DhisApiVersion.ALL } )
     public @ResponseBody Map<String, List<String>> getRequiredValues( @PathVariable String id )
-        throws WebMessageException
     {
         TrackedEntityAttribute trackedEntityAttribute = trackedEntityAttributeService.getTrackedEntityAttribute( id );
 
-        if ( trackedEntityAttribute == null )
```
teleivo, line 140: why are you not calling doing

`        TrackedEntityAttribute attribute = getTrackedEntityAttribute( id );`

instead of

`        TrackedEntityAttribute trackedEntityAttribute = trackedEntityAttributeService.getTrackedEntityAttribute( id );`

above? Since you remove the checks here

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -91,13 +91,6 @@
         @PathVariable String id )
         throws WebMessageException
     {
-        TrackedEntityAttribute trackedEntityAttribute = trackedEntityAttributeService.getTrackedEntityAttribute( id );
```
teleivo, line 94: why are is this removed here and in the method below?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
+import org.hisp.dhis.webapi.DhisWebSpringTest;
+import org.hisp.dhis.webapi.utils.TestUtils;
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockHttpSession;
+
+/**
+ * @author Luca Cambi
+ */
+class TrackedEntityAttributeControllerTest extends DhisWebSpringTest
```
teleivo, line 48: You could look into using `DhisControllerConvenienceTest` which might remove some of the setup you have to do here and makes the tests very nice to read IMHO. Check for usages like

https://github.com/dhis2/dhis2-core/blob/63161ceba0ab1bc2dfb77e5452ffc879f9036949/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/SecurityControllerTest.java

teleivo (commented):
I have a few questions on code that has been removed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
lucaCambi77: every call for reserve values is now passing in _reserve_  private method where we have all relevant checks

teleivo: where is that method? did you make that change in this PR?

enricocolasante (approved):

enricocolasante (approved):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/attribute/TrackedEntityAttributeControllerTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        trackedEntityAttribute.setTextPattern( textPattern );
+        trackedEntityAttribute.setPattern( pattern );
+
+        String uid = assertStatus( org.springframework.http.HttpStatus.CREATED,
+            POST( TrackedEntityAttributeSchemaDescriptor.API_ENDPOINT,
+                new String( TestUtils.convertObjectToJsonBytes( trackedEntityAttribute ) ) ) );
+
+        assertStatus( org.springframework.http.HttpStatus.OK, GET(
+            TrackedEntityAttributeSchemaDescriptor.API_ENDPOINT + "/" + uid + "/generateAndReserve"
```
teleivo, line 69: just in case you did not know 😄  you can also use placeholders construct the URL https://github.com/dhis2/dhis2-core/blob/d654e335f20678b18edca3ad1c9787f455b5b82b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/AbstractCrudControllerTest.java#L129

teleivo (approved):

PR #10453, opened 2022-04-08
fix: Add display names for period types [2.36-DHIS2-12967]

d-bernat:
Mismatches between frontend and backend, see attached image
see: [https://jira.dhis2.org/browse/DHIS2-12967](url)


gnespolino (approved):

maikelarabori (approved):
Looks good. I just do not remember if in this case, we should change all `properties` files or only the root one `i18n_global.properties`...

d-bernat:
> Looks good. I just do not remember if in this case, we should change all `properties` files or only the root one `i18n_global.properties`...

This is the outcome of the discussion with Jan. We agreed to localize it

PR #10477, opened 2022-04-12
fix: oulevel column name in JdbcAnalyticsManager

tokland:
Fixes https://jira.dhis2.org/browse/DHIS2-13094


stian-sandvold:
This was fixed in a related jira issue (12989). Confirmed by @d-bernat . Closing

PR #10500, opened 2022-04-13
feat: Include data set org unit associations in metadata response

larshelge:
Includes associations between data sets and organisation units in the `/api/dataSetMetadata` endpoint. Adds a string array property `organisationUnits` to each data set.

Makes the current solution for program org unit associations fully generic by renaming program-specific variables and moving classes to generic package.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/DataSetService.java
```diff
@@ -304,4 +305,15 @@ LockStatus getLockStatus( User user, DataElement dataElement, Period period, Org
...
+    /**
+     * Return a mapping between the given data sets and the associated
+     * organisation units. Only data sets for which the current user has data
+     * write sharing access to are returned.
+     *
+     * @param dataSetUids the data set identifiers.
+     * @return a {@link SetValuedMap} between data sets and organisation unit
+     *         identifiers.
+     */
+    SetValuedMap<String, String> getDataSetOrganisationUnitsAssociations();
```
jbee, line 318: Do we want to keep using the special collections? I'd thought we could use `Map<String, Set<String>>` since the building now can be written nicer using `computeIfAbsent`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/DataSetOrganisationUnitAssociationsQueryBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+import lombok.Getter;
+
+import org.hisp.dhis.user.CurrentUserService;
+import org.springframework.stereotype.Service;
+
+@Service
+public class DataSetOrganisationUnitAssociationsQueryBuilder
+    extends AbstractOrganisationUnitAssociationsQueryBuilder
+{
+    @Getter( AccessLevel.PROTECTED )
```
jbee, line 40: could also annotate the class instead of all fields

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/DefaultDataSetService.java
```diff
@@ -72,6 +75,9 @@
 
     private final DataApprovalService dataApprovalService;
 
+    @Qualifier( "jdbcDataSetOrgUnitAssociationsStore" )
```
jbee, line 78: Do we need the `@Qualifier` when the type should be sufficiently specific?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultDataSetMetadataExportService.java
```diff
@@ -158,6 +166,46 @@ private Set<CategoryOption> getCategoryOptions(
...
+    /**
+     * Returns data sets as a list of {@link ObjectNode}. Includes associations
+     * to organisation units.
+     *
+     * @param dataSets the list of {@link DataSet}.
+     * @param dataSetOrgUnits the associations between data sets and
+     *        organisation units.
+     * @return data sets as a list of {@link ObjectNode}
+     */
+    private List<ObjectNode> getDataSets( List<DataSet> dataSets, SetValuedMap<String, String> dataSetOrgUnits )
```
jbee, line 178: `get` is maybe an overused prefix for a method doing a conversion or creation of a new type of object from another object. I suggest `create` or `to` as prefixes in such cases.

jbee (approved):
Some code style comments, nothing significant.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/DataSetService.java
larshelge: Agreed. I just didn't want to change the underlying JdbcOrgUnitAssociationsStore interface in order to avoid a growing task. We could change this later.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/DataSetOrganisationUnitAssociationsQueryBuilder.java
larshelge: Agreed. Was just following the convention from the other classes like `ProgramOrganisationUnitAssociationsQueryBuilder` for consistency.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultDataSetMetadataExportService.java
larshelge: Agreed. Fixed.

PR #10504, opened 2022-04-14
fix: endpoint api/resources.json return 404 ( 2.36) 

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12591

### Issue
- Request to `api/resources.json` got `404` response.
- `api/resources` and `api/resources.xml` work correctly. 
- This only happens to `2.36`.  It doesn't occur in `2.35` or `2.37+`
- It's also strange that this issue only happens to the endpoint `api/resources`. Other endpoints are not affected.

### Fix
- I couldn't find the different in `WebMvcConfig` between `2.36` and other versions, maybe there is some hidden configs causing this.
- Found https://github.com/spring-projects/spring-framework/issues/23915 which stated that Spring changed the default value of `useSuffixPatternMatch` to false. ( 2.36 is using spring 5.3.2 ) 
- So I turn that setting to `true` and it solves the issue.

Screen Shot 2022-04-14 at 14 00 43


jbee (approved):
As discussed on slack I think since we explicitly set what was the default in older spring versions we should not see a change in behaviour for endpoints that we didn't have prior.

mortenoh (approved):

PR #10516, opened 2022-04-15
fix: V{value_count} in PI filter DHIS2-13099

jimgrace:
See [DHIS2-13099](https://jira.dhis2.org/browse/DHIS2-13099). The Program Indicator variables `V{value_count}` and `V{zero_pos_value_count}` were not working properly when used in a program indicator filter. This is because they were counting the values of data elements or attributes specified in the filter, instead of counting the values of those items specified in the program indicator main expression. This has been fixed.

(The User Manual clearly says that these functions should return the number of values "in the expression part of the event." See [here](https://docs.dhis2.org/en/full/use/user-guides/dhis-core-version-master/dhis2-user-manual.html#configure_program_indicator), under the table _Variables to use in a program indicator expression or filter_.)

A secondary problem is that if no data elements or attributes are present in the expression, the `V{value_count}` and `V{zero_pos_value_count}` functions were generating invalid SQL. This has been fixed so they now generate a "0" in the SQL query, because there are no values that could be found.

The unit tests for these functions have been adjusted accordingly.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
@@ -382,7 +382,9 @@ public class DefaultProgramIndicatorService
     private String _getAnalyticsSql( String expression, ProgramIndicator programIndicator, Date startDate, Date endDate,
         String tableAlias )
     {
-        Set<String> uids = getDataElementAndAttributeIdentifiers( expression, programIndicator.getAnalyticsType() );
+        // Get the uids from the expression even if this is the filter
+        Set<String> uids = getDataElementAndAttributeIdentifiers( programIndicator.getExpression(),
+            programIndicator.getAnalyticsType() );
 
         ProgramExpressionParams params = ProgramExpressionParams.builder()
             .programIndicator( programIndicator )
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/variable/vValueCount.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/variable/vValueCount.java
@@ -27,6 +27,8 @@
  */
 package org.hisp.dhis.program.variable;
 
+import java.util.Set;
+
 import org.hisp.dhis.commons.util.TextUtils;
 import org.hisp.dhis.parser.expression.CommonExpressionVisitor;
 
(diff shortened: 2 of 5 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
```diff
@@ -382,7 +382,9 @@ private String getAnalyticsSqlCacheKey( String expression, ProgramIndicator prog
     private String _getAnalyticsSql( String expression, ProgramIndicator programIndicator, Date startDate, Date endDate,
         String tableAlias )
     {
-        Set<String> uids = getDataElementAndAttributeIdentifiers( expression, programIndicator.getAnalyticsType() );
+        // Get the uids from the expression even if this is the filter
+        Set<String> uids = getDataElementAndAttributeIdentifiers( programIndicator.getExpression(),
```
maikelarabori, line 386: Hey @jimgrace, I'm not really familiar with this feature. So my question might not make sense, but I just want to make sure it's not a typo.

We are now using `programIndicator.getExpression()` instead of `expression`.

In the line number `400`, though -> `String sql = castString( Parser.visit( expression, visitor ) );`, we are using `expression`. Is it expected?

maikelarabori (approved):
The code looks good @jimgrace, I just left a question for you.

Thanks,
Maikel

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
jimgrace: Thanks @maikelarabori, good question. A program indicator has a (main) expression and a filter expression. When this code is called, `expression` will be set to one of these two. What we want to find ultimately is the value of `expression`. If we're evaluating the main expression and it contains a `V{value_count}` or `V{zero_pos_value_count}` we want to see how many of the data elements or attributes present in the main expression have a value. However if we're evaluating the filter expression, we also need to see how many of the data elements or attributes present in the main expression have a value. So to find set of data element and attribute identifiers, we should always look in the main expression, no matter whether we are evaluating the main expression or the filter expression.

maikelarabori: Thanks for the clarification @jimgrace :+1:

PR #10532, opened 2022-04-19
fix: Disk space utilization for analytics temp tables spiking in [2.36-DHIS2-13085]

d-bernat:
Sometimes the temporary partition table are not dropped and staying in database after the analytics update. Normally the temp partitions are dropped by swapping the old for new ones. If this swap process failed (silently, there is just debug level log), the old data remaining in the partition table. Subsequently new analytics update process will just add the more or less same data into temp table (with reindexing - time and space consuming).
On the beginning of the process we are trying to drop  all remaining temp tables. The implementation did not remove (drop) partition temp tables (year handling was incorrect)
This problem was fixed (unintentionally) by Maikel in https://github.com/dhis2/dhis2-core/pull/10014/files and backported down to the 2.37 (PR 9th of March 2022).
2.36 was not backported, we are using another partitioning there, so it is not possible just backport it. It means <=2.36 is still buggy.
Because the reason why the swap process is not able to remove temp table, is not clear to me (probably continuous analytics job process  locked the table, maybe forever when hanging), there is no guarantee the problem will gone. Good message is we should have the log indicates there is something wrong.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/AbstractJdbcTableManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/AbstractJdbcTableManager.java
@@ -74,6 +74,7 @@ import org.hisp.dhis.setting.SystemSettingManager;
 import org.hisp.dhis.system.database.DatabaseInfo;
 import org.hisp.dhis.util.DateUtils;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DataAccessException;
 import org.springframework.jdbc.BadSqlGrammarException;
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/AbstractJdbcTableManager.java
```diff
@@ -387,7 +393,7 @@ protected void executeSilently( String sql )
         }
         catch ( BadSqlGrammarException ex )
```
maikelarabori, line 394: In "master" we are, know, catching `DataAccessException`, should it be the case here too?

maikelarabori (approved):
Looks good @d-bernat. I just left one comment for your review before merging it.

Cheers,
Maikel

gnespolino (approved):

PR #10538, opened 2022-04-20
feat: adds datetime support to dataelements [DHIS2-13115]

gnespolino:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -719,17 +722,40 @@ protected String getSelectSql( QueryFilter filter, QueryItem item, Date startDat
...
+                        // known formats
+                        "yyyy-MM-dd'T'HH.mm",
+                        "yyyy-MM-dd'T'HH.mm.ss" ),
+                    // postgres format
+                    "yyyy-MM-dd HH:mm:ss" );
+            }
+        }
+        catch ( ParseException pe )
+        {
+            log.warn( "Unparsable date using known formats: " + filter );
```
maikelarabori, line 742: What if we use the same approach we already have in this class. Something like:
```
log.warn( ErrorCode.E7131.getMessage(), ex );
throw new QueryRuntimeException( ErrorCode.E7131, ex );
```
where
```
ErrorCode.E7131 = "Unparsable date/time. Unknown format"
```
Just a suggestion, so the clients can have an idea about what could be wrong.

maikelarabori (approved):
Looks good @gnespolino, just left a suggestion for your review.

Thanks,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
gnespolino: sounds good.

d-bernat (approved):

PR #10546, opened 2022-04-20
fix: TrackerPreheat defaults support idSchemes DHIS2-12563

teleivo:
Relates to both https://jira.dhis2.org/browse/DHIS2-12563 and https://jira.dhis2.org/browse/DHIS2-12759

defaults for metadata like Category, CategoryOptionCombo, ... where not returned by

https://github.com/dhis2/dhis2-core/pull/10546/files#diff-8bcab5fc74e034ebd337f7c3a069adb167a1451414163c9b7938aafaab26dc11L379-L380

since imports using idScheme other than UID failed to insert the metadata into the TrackerPreheat.map. So even though `TrackerPreheat.defaults` contained the defaults `return this.get( defaultClass, uid );` (which accesses `TrackerPreheat.map`) did not.

Add default metadata into the preheat using a dedicated supplier. The logic is irrespective of the idScheme chosen during the import. Defaults can be retrieved using their name (`default`). Its important to map the metadata to avoid lazy initialization exceptions we get from relying on hibernate proxies. Note: the manager.getDefaults() still returns proxies (parents are not, but children collections are)!

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/mappers/CategoryOptionMapper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/mappers/CategoryOptionMapper.java
@@ -28,7 +28,9 @@
 package org.hisp.dhis.tracker.preheat.mappers;
 
 import org.hisp.dhis.category.CategoryOption;
+import org.mapstruct.BeanMapping;
 import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
 import org.mapstruct.factory.Mappers;
 
 @Mapper( uses = DebugMapper.class )
@@ -36,5 +38,10 @@ public interface CategoryOptionMapper extends PreheatMapper<CategoryOption>
 {
     CategoryOptionMapper INSTANCE = Mappers.getMapper( CategoryOptionMapper.class );
 
+    @BeanMapping( ignoreByDefault = true )
+    @Mapping( target = "id" )
+    @Mapping( target = "uid" )
+    @Mapping( target = "name" )
+    @Mapping( target = "code" )
     CategoryOption map( CategoryOption categoryOption );
 }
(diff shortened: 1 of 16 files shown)


ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/mappers/CategoryOptionMapper.java
```diff
@@ -28,13 +28,25 @@
...
 import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
 import org.mapstruct.factory.Mappers;
 
 @Mapper( uses = DebugMapper.class )
 public interface CategoryOptionMapper extends PreheatMapper<CategoryOption>
 {
     CategoryOptionMapper INSTANCE = Mappers.getMapper( CategoryOptionMapper.class );
 
+    @BeanMapping( ignoreByDefault = true )
```
teleivo, line 41: @enricocolasante would be in favor of adding this explicit mapping?

The reason why I originally added it was to prevent the lazy init exception we got from mapping the collection fields. We learned that this was due to the `manager.getDefaults()` not really returning a detached entity. Its collection fields were still hibernate proxies, and since these were cached in another thread the session was closed => lazy init exception.

Now that I wrote a specific `DefaultsSupplier` that uses the manager to fetch the entities, this will not occur.
This is why I am asking. So this change is not needed to fix defaults. It would only align with being explicit which we do in most places. We could of course make this change its own PR at some point. In all remaining mappers.

teleivo: I removed the change since it is not needed to fix defaults with respect to idSchemes. Forgetting to map a field can lead to another bug. So if we want to use explicit mapping everywhere, we should do so in another task.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/supplier/DefaultsSupplier.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.stereotype.Component;
+
+@RequiredArgsConstructor
+@Component
+public class DefaultsSupplier extends AbstractPreheatSupplier
+{
+
+    private static final int CACHE_TTL = 60;
+
+    private static final long CACHE_CAPACITY = 1000;
```
stian-sandvold, line 58: The number of default values should be exactly 4 in the case of category, categorycombo, categoryoption and categoryoptioncombo. I think a more relevant capacity would be 4 or maybe 10, if we assume we might add more defaults later?

Also the defaults _never_ change in a normal instance, so the TTL could be way higher :)

stian-sandvold (approved):

PR #10601, opened 2022-04-26
feat: Add custom data entry form API endpoint

larshelge:
### Overview

Adds a custom data entry form API endpoint at the following API path, where the `id` refers to the data set UID. The response type is JSON.

```
GET /api/dataEntry/customForms/{id}
```

Removes unused method from `DataEntryFormService`.

### Testing

* Unit tests added.
* Test URL for valid form: `/api/dataEntry/customForms/lyLU2wR22tC`
* Test URL for non-custom form: `/api/dataEntry/customForms/BfMAe6Itzgt`
* Test URL for non-existing data set: `/api/dataEntry/customForms/lyLU2wR22tX`

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -67,7 +67,10 @@ public interface IdentifiableObjectManager
 
     <T extends IdentifiableObject> T get( Class<T> type, long id );
 
-    <T extends IdentifiableObject> T get( Class<T> type, String uid );
+    <T extends IdentifiableObject> T get( Class<T> type, String uid )
+        throws IllegalQueryException;
+
+    <T extends IdentifiableObject> T getAndValidate( Class<T> type, String uid );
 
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
```diff
@@ -67,7 +67,10 @@
 
     <T extends IdentifiableObject> T get( Class<T> type, long id );
 
-    <T extends IdentifiableObject> T get( Class<T> type, String uid );
+    <T extends IdentifiableObject> T get( Class<T> type, String uid )
```
jbee, line 70: Just as food for thought: If we have a `get` that makes sure things exist a lookup that allows for things to not exist is usually called `find` or something like that. We might want to look into establishing semantics consistent with method names used.

jbee (approved):
Look good.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
larshelge: Yes I agree, it would be good with a more succinct notation for null vs exception. Could be good to do a separate refactor PR.

PR #10613, opened 2022-04-27
feat: user data store impersonate users as superuser [DHIS2-7851]

jbee:
### Summary
Adds an optional  `username` parameter to all endpoints of the user data store that can be used by a superuser to impersonate another user, that means run the operation as the other user affecting that users data-store. 

This feature can only be used by superusers. Non superusers get a conflict should they use the `username` parameter.

### Automatic Testing
Added a new unit test testing all endpoints for the scenario where superusers impersonate a user.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/userdatastore/UserDatastoreEntry.java
```diff
@@ -117,7 +117,7 @@ public void setKey( String key )
     @JsonProperty
     public String getValue()
     {
-        return encrypted ? encryptedValue : jbPlainValue;
+        return isEncrypted() ? getEncryptedValue() : getJbPlainValue();
```
jbee, line 120: This is actually a bugfix. I think we can backport this entire PR to 2.38 and maybe even 2.37 as it is very local. If we decide not to backport we need to backport this line change as a fix.

jbee: The issue here is that not using the getter can result in `null` even if the `value` field is non-null. This is because the getters return the `value` field with priority as long as it is non-null.

larshelge (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserDatastoreController.java
```diff
@@ -219,18 +231,44 @@ public WebMessage updateUserValue(
...
+        }
+        if ( !currentUser.isSuper() )
+        {
+            throw new IllegalQueryException(
+                "Only superusers can read or write other users data using the `user` parameter." );
+        }
+        User user = userService.getUserByUsername( username );
+        if ( user == null )
+        {
+            throw new IllegalQueryException( "No user with username " + username + " does exist." );
```
benguaraldi, line 270: "No user with username X exists." or "There is no user with username X." might be better.

jbee: I try to sneak in this wording change in some of my next PRs ;)

benguaraldi: Good plan! :)

Thanks!

PR #10617, opened 2022-04-27
fix: implicit OR conditions are now applicated for filters [DHIS2-13160]

gnespolino:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -897,12 +897,12 @@ else if ( header.getValueType() == ValueType.REFERENCE )
...
 
     /**
-     * Return SQL string bnased on Items a its params
+     * Return SQL string based on Items and its params
      *
      * @param params a {@see EventQueryParams}
      * @param hlp a {@see SqlHelper}
      */
-    protected String getItemsSql( EventQueryParams params, SqlHelper hlp )
+    protected String getItemsAndItemFiltersSql( EventQueryParams params, SqlHelper hlp )
```
maikelarabori, line 905: Maybe we could rename this method to something like "getStatementForDimensionsAndFilters".

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -897,12 +897,12 @@ else if ( header.getValueType() == ValueType.REFERENCE )
     }
 
     /**
-     * Return SQL string bnased on Items a its params
+     * Return SQL string based on Items and its params
```
maikelarabori, line 900: Regarding the Javadoc, I think we could also make clear that the method takes into consideration "dimensions" and "filters".

maikelarabori (approved):
Looks good @gnespolino, just two minor comments. :+1: 

Thanks!

enricocolasante (approved):

PR #10623, opened 2022-04-27
fix: app hub returns 503 when internal app hub request fails [DHIS2-13107]

jbee:
### Summary
The best fix here seemed to me to map the `RestClientException` to a `WebMessage` with `503` HTTP response code.
This exception was previously not mapped resulting in a HTTP 500.
Mapping all "client" (requests done by the server internally to other web services) to a 503 might be a bit broad but it surely is better than a 500 and doing very detailed error handling just seems overkill for this and similar problems.

### Automatic Testing
Added some controller tests that verify the 503 by changing the configuration for the app hub API URL to some non-existing URL. This is not a timeout as encountered in the original ticket but should be a similar type of scenario.

I also added two "success" path scenarios which work with the default URL which means the test do a request to the https://apps.dhis2.org/api URL. We have to see if this works without issues on CI.


larshelge (approved):

larshelge:
Nice solution :+1:

PR #10627, opened 2022-04-28
ci: we do not allow merge commits

teleivo:
so our dependabot workflow should also squash its commits when the PR is merged


teleivo:
please someone with admin rights overrule the failed sonarcloud check as the quality gate actually passed. and this is just a yaml change 😋

stian-sandvold (approved):

PR #10629, opened 2022-04-28
fix: remove unused RollingFileAppenders in log4j2.xml DHIS2-12986

teleivo:
The "fileLogger" defined in the xml config is never "attached" to a
logger. Not in the config or in code. Its thus unused.
Log4JLogConfigInitializer creates its own RollingFileAppenders unless a
user provides their own logging config.

Just to understand the root cause of the exception on start of DHIS2:
The "fileLogger" in the log4j2.xml config was responsible for it. The
property 'createOnDemand' controls when the RollingFileAppender creates
the log file. Since that is false by default, the "fileLogger" is trying
to create the file once log4j2 initializes. log4j2 does not have
permissions to create the dhis.log specified in "fileLogger".

### DIFF
--- a/dhis-2/dhis-support/dhis-support-test/src/main/resources/log4j2.xml
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/resources/log4j2.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Configuration status="warn">
     <Appenders>
-
         <Console name="console" target="SYSTEM_OUT">
             <PatternLayout  pattern="* %-5p %d{ABSOLUTE} %m (%F [%t])%n" />
             <Filters>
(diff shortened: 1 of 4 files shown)


ameenhere (approved):
Approved it but I was wondering if it makes sense to make it "used" instead of removing it?

teleivo:
> Approved it but I was wondering if it makes sense to make it "used" instead of removing it?

I dont see how that would be useful. I am open for someone convincing me otherwise 😋 The config is shipped by us in the war. We actually configure log4j2 in our code https://github.com/dhis2/dhis2-core/blob/c7a59a167ac899f3306da2ba1f8ce248c1a6ef66/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/log/Log4JLogConfigInitializer.java#L283-L300 At least when it comes to logging to files.

If a user does not like the logging configuration they can provide their own log4j2 configuration, in which case our log4j2.xml will also play no role

ameenhere:
> At least when it comes to logging to files.

Ah. I missed this. Yes if we have the file appender in place in the programmatic java log4j2 configuration, then that is fine. My immediate reaction was because I wrongly assumed that we were skipping file-based logging completely. LGTM!

netroms (approved):

PR #10632, opened 2022-04-28
fix: expected reports for daily datasets in last month was wrong [DHIS2-13134]

gnespolino:
TY @maikelarabori and @claudemamo for support on reproducing the issue


maikelarabori (approved):
Thanks for looking into it @gnespolino :+1:

zubaira (approved):

PR #10594, opened 2022-04-26
fix: Performance: relationships endpoint is slow when there are many relationships [ 2.39 ]

lucaCambi77:
Improve get relationship by an entity ( tei, pi, psi ) in order to actually use the index in relationshipitem table.
Access plan should now  be like https://explain.dalibo.com/plan/blEn

PR #10642, opened 2022-04-29
fix: Performance: relationships endpoint is slow when there are many relationships [ 2.38 ]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/10594


lucaCambi77:
try to create a new one

PR #10670, opened 2022-05-03
fix: Remove ANTRL versions warning from logs

enricocolasante:
dhis2-antlr-expression-parser was using different versions for the antlr plugin and the antlr library and this was throwing a warning in the logs


teleivo (approved):
PR description explains what the problem was :)

teleivo:
@enricocolasante can you please use a maven property for the antler version in the parser lib. This will prevent this from happen again :)

gnespolino (approved):

PR #9841, opened 2022-02-17
fix: NTI relationship-endpoint only returning uids DHIS2-12625

teleivo:
## What

Tracker endpoints:

1. POST /api/tracker
2. GET /api/tracker/enrollments
3. GET /api/tracker/events
4. GET /api/tracker/trackedEntities
5. GET /api/tracker/relationships

1-4 will still take and respond with only UIDs in the relationships `from` and `to` `trackedEntity`, `enrollment` and `event` properties.

5. will now respond with the full `trackedEntity`, `enrollment` and `event` as in the individual endpoints 2-4

so before

```json
{
  "relationship": "Nva3Xj2j75W",
  "relationshipName": "Mother-Child_a-to-b_(Person-Person)",
  "relationshipType": "dDrh5UyCyvQ",
  "createdAt": "2018-10-01T12:17:30.163",
  "updatedAt": "2018-10-01T12:17:30.163",
  "bidirectional": false,
  "from": {
    "trackedEntity": "Ea0rRdBPAIp"
  },
  "to": {
    "trackedEntity": "G1afLIEKt8A"
  }
}
```

after

```json
{
  "relationship": "Nva3Xj2j75W",
  "relationshipName": "Mother-Child_a-to-b_(Person-Person)",
  "relationshipType": "dDrh5UyCyvQ",
  "createdAt": "2018-10-01T12:17:30.163",
  "updatedAt": "2018-10-01T12:17:30.163",
  "bidirectional": false,
  "from": {
    "trackedEntity": {
      "trackedEntity": "Ea0rRdBPAIp",
      "trackedEntityType": "nEenWmSyUEp",
      "createdAt": "2013-09-08T21:40:28.000",
      "createdAtClient": "2013-09-08T21:40:28.000",
      "updatedAt": "2014-07-23T12:53:31.638",
      "orgUnit": "DiszpKrYNg8",
      "inactive": false,
      "deleted": false,
      "potentialDuplicate": false,
      "relationships": [],
      "attributes": [
...
```

* fixed another bug where /api/tracker/relationships?enrollment could not find relationships see TrackerRelationshipsExportController.relationshipRetrievers which was `ProgramStage` to `getRelationshipsByProgramInstance` instead of `ProgramInstance`
* consistent error messages. Stick to `trackedEntity`
* add tests

## How

Since we now map the full tei, enrollment, event in the `from` and `to` properties of the RelationshipItem in endpoint `GET /api/tracker/relationships`. This endpoint needs some purpose built classes.

[...]

PR #10676, opened 2022-05-03
chore: simplify RelationshipItem in domain DHIS2-12563

teleivo:
## This PR

Now that we have separate models for our domain and the view (JSON exposed to users) only the TEI, enrollment, event ids are needed in the tracker domain RelationshipItem.

See https://github.com/dhis2/dhis2-core/pull/9841 if you need more context why these fields where introduced before we had domain and view models.

## Background on changes in [DHIS2-12563](https://jira.dhis2.org/browse/DHIS2-12563)

* tracker view (JSON import/export) and domain models are split since users only specify idSchemes in the query parameters (not in the body) and only on import. The metadata identifier fields in the user-facing import/export JSON are strings containing only the identifier without idScheme.

* Attribute.attribute
* DataValue.dataElement
* Event.programStage :heavy_check_mark:
* Event.attributeOptionCombo :heavy_check_mark:
* Event.attributeCategoryOptions
* TrackedEntity.trackedEntityType
* Relationship.relationshipType
* {Enrollment :heavy_check_mark:,Event :heavy_check_mark:}.program
* {Enrollment :heavy_check_mark:,Event :heavy_check_mark:,TrackedEntity :heavy_check_mark:}.orgUnit

will become idScheme aware.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/RelationshipValidationUtils.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/RelationshipValidationUtils.java
@@ -40,15 +40,15 @@ public class RelationshipValidationUtils
 {
     public static TrackerType relationshipItemValueType( RelationshipItem item )
     {
-        if ( item.getTrackedEntity() != null && StringUtils.isNotEmpty( item.getTrackedEntity().getTrackedEntity() ) )
+        if ( StringUtils.isNotEmpty( item.getTrackedEntity() ) )
         {
             return TrackerType.TRACKED_ENTITY;
         }
-        else if ( item.getEnrollment() != null && StringUtils.isNotEmpty( item.getEnrollment().getEnrollment() ) )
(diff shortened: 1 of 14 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/RelationshipValidationUtils.java
```diff
@@ -59,15 +59,15 @@ public static Optional<String> getUidFromRelationshipItem( RelationshipItem item
...
         {
-            return Optional.of( item.getEnrollment().getEnrollment() );
+            return Optional.of( item.getEnrollment() );
         }
         else if ( item.getEvent() != null )
         {
-            return Optional.of( item.getEvent().getEvent() );
+            return Optional.of( item.getEvent() );
         }
         return Optional.empty();
```
enricocolasante, line 72: Alternative code:

`return Optional.ofNullable(ObjectUtils.firstNonNull(item.getTrackedEntity(), item.getEnrollment(), item.getEvent()));`

enricocolasante (approved):
Just an alternative code suggestion

jbee (approved):

PR #10682, opened 2022-05-03
feat: Add deleted field to relationship [DHIS2-11320]

enricocolasante:
This is the first of a few PR to make relationships soft deletable.
In this PR I am just shaping the model in the DB and in dhis-api.

Relationship extends `BaseNameableObject` even though any of its nameable fields are ever exposed, so I changed it to a softDeletableObject that extends `BaseIdentifiableObject`. I therefore removed a test that checked translation of relationship displayName.
@stian-sandvold Does it make sense?


dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipService.java
```diff
@@ -74,6 +74,15 @@
      */
     Relationship getRelationship( long id );
 
+    /**
+     * Checks for the existence of a relationship by UID. Takes into account
```
teleivo, line 78: Maybe

`Checks if relationship for given UID exists (including deleted relationships).`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipService.java
```diff
@@ -74,6 +74,15 @@
...
+     * also the deleted relationships.
+     *
+     * @param uid Relationship UID to check for.
+     * @return relationship exists or not.
```
teleivo, line 82: I know the return one in the style of

`return true if relationship exists, false otherwise`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipService.java
```diff
@@ -74,6 +74,15 @@
...
+     */
+    boolean relationshipExistsIncludingDeleted( String uid );
```
teleivo, line 84: do we have any convention for methods with boolean return types like this one
like?

`hasRelationship`
`isRelationship`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipStore.java
```diff
@@ -80,6 +80,15 @@ List<Relationship> getByProgramStageInstance( ProgramStageInstance psi,
      */
     Relationship getByRelationship( Relationship relationship );
 
+    /**
+     * Checks for the existence of a relationship by UID. Takes into account
```
teleivo, line 84: if you update the service than please update this one as well

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/relationship/hibernate/HibernateRelationshipStore.java
```diff
@@ -202,6 +203,16 @@ public List<Relationship> getByRelationshipType( RelationshipType relationshipTy
 
     }
 
+    @Override
+    public boolean existsIncludingDeleted( String uid )
+    {
+        Query query = getSession().createNativeQuery( "select count(*) from relationship where uid=:uid" );
```
teleivo, line 209: I would not use count if all we want is know whether there is a relationship with the UID and there is a possibility for more than one relationship with the UID (so no unique index).

Something like

`select uid from relationship where uid=:uid limit 1;`

would be enough I assume

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/translation/TranslationServiceTest.java
```diff
@@ -138,35 +132,6 @@ void testFormNameTranslationForOption()
...
-        to.setTrackedEntityInstance( trackedEntityInstance );
-        relationship.setFrom( from );
-        relationship.setTo( to );
-        relationship.setRelationshipType( relationshipType );
-        relationship.setKey( RelationshipUtils.generateRelationshipKey( relationship ) );
-        relationship.setInvertedKey( RelationshipUtils.generateRelationshipInvertedKey( relationship ) );
-        manager.save( relationship );
-        String translatedValue = "RelationShip FormName Translated";
-        Set<Translation> translations = new HashSet<>( relationship.getTranslations() );
-        translations.add( new Translation( locale.getLanguage(), "FORM_NAME", translatedValue ) );
```
teleivo, line 165: did that test relate to it being a BaseNameableObject before?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/relationship/AbstractRelationshipService.java
```diff
@@ -307,8 +307,6 @@ public ImportSummaries updateRelationships( List<Relationship> relationships, Im
             {
                 importSummaries.addImportSummary( updateRelationship( relationship, importOptions ) );
             }
-
-            clearSession();
```
teleivo, line 311: why did you remove this? just asking what it was for

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_8__Add_soft_delete_feature_relationship.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE relationship ADD COLUMN IF NOT EXISTS deleted boolean DEFAULT false;
+
+ALTER TABLE relationship drop constraint fk_relationship_from_relationshipitemid;
```
teleivo, line 3: drop constraint should support `if exists` https://www.postgresql.org/docs/current/sql-altertable.html just in case the constraint is not there.

foreign key constraints also supports ALTER CONSTRAINT
but the drop with if exists/add might be more resilient if it did not exist.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipService.java
enricocolasante: Not really. Even less for the include deleted ones

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/translation/TranslationServiceTest.java
enricocolasante: Yes.
relationship.getDisplayFormName() is the only time we use relationship as a BaseNameableObject

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/relationship/AbstractRelationshipService.java
enricocolasante: This was causing a failure in the tests after my changes.
Anyway, we shouldn't clear the session during a transaction. I found out that generally this clearSession() calls were added to make some test pass when more than one call was involved.

The right solution is to clear the session if needed in the test itself between api calls

enricocolasante:
> seems sensible to me 😄 just a few minor things/questions. and can we make such a change from BaseNameableObject to SoftDeletableObject in Relationships without affecting other teams?

We will try to answer that in the next design meeting.

stian-sandvold:
> 

The relationship itself is not translatable and doesnt have any native names, so displayName should in theory never be used. This should be fine to change :)

stian-sandvold:
> 

As far as I know, Relationships are used by Tracker (backend), Web and Android. Additionally, it's used in the GIS app I think. The change itself should not affect any other code in theory. I'll tag @mortenoh and @vietnguyen the metadata/generic model experts to confirm or not my assumptions :)

stian-sandvold (approved):

teleivo (approved):

PR #10690, opened 2022-05-03
feat: Add data entry data and min-max values API endpoint [DHIS2-13184]

larshelge:
Adds new API endpoint for data values and min-max values, intended for the new data entry web app, at:

```
/api/dataEntry/dataValues
```

Returns data values and min-max values for the specific form. Example query:

```
/api/dataEntry/dataValues?ds=BfMAe6Itzgt&pe=202204&ou=DiszpKrYNg8
```

Adds new API endpoint for POST and DELETE of min-max value at:

```
/api/dataEntry/minMaxValues
```


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dataentry/DataSetValueController.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private final DataValidator dataValidator;
+
+    @GetMapping( "/dataValues" )
+    public DataValuesDto getDataValueSet( DataSetValueQueryParams params )
+    {
+        DataSet ds = dataValidator.getAndValidateDataSet( params.getDs() );
+        Period pe = dataValidator.getAndValidatePeriod( params.getPe() );
+        OrganisationUnit ou = dataValidator.getAndValidateOrganisationUnit( params.getOu() );
+        CategoryOptionCombo ao = dataValidator.getAndValidateAttributeOptionCombo( params.getCc(), params.getCp() );
```
jbee, line 77: Makes me think a validation for `DataSetValueQueryParams` would be nicer but I guess this is also used as a way to map uids to objects.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/datavalue/DataValuesDto.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * DTO which represents data values and min-max values for a data entry form.
+ *
+ * @author Lars Helge Overland
+ */
+@Getter
+@Setter
```
jbee, line 46: Do these need to be mutable? I assume this is more of a return value DTO?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/datavalue/MinMaxValueDto.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * DTO which represents a min-max value.
+ *
+ * @author Lars Helge Overland
+ */
+@Getter
+@Setter
```
jbee, line 43: Again, may these be immutable?

jbee (approved):
Mostly wonder if we can have less mutable state objects.

mortenoh (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/datavalue/MinMaxValueDto.java
larshelge: This one is used for POST as well.

PR #10716, opened 2022-05-04
fix: Log4JLogConfigInitializer should check log4j version 2 config DHIS2-13163 [2.35]

teleivo:
backport of https://github.com/dhis2/dhis2-core/pull/10641 except for the exclusion of log4j version 1. debezium prior to 1.9 brings in a dependency on log4j version 1. Since debezium is responsible for replication backporting an upgrade of debezium seems risky.

Log4JLogConfigInitializer will not make any config changes if the user provides
their own logging configuration. The check relies on the log4j config property.
Log4JLogConfigInitializer should check the log4j version 2 config not version 1, as
we rely on version 2.


teleivo:
unclear what this test has to do with these minor log related changes. its fixed in 2.36-2.39

PR #10743, opened 2022-05-09
fix: Fix sonar issue in github actions

enricocolasante:
Way to go Sonar :(
https://community.sonarsource.com/t/urgent-analyzing-functions-never-stops/62775

They changed the plugin and add a new analyser that is not working for us, I have no clue why. Disabling the new analyser and everything works again as before

### DIFF
--- a/.github/workflows/analyse-pr.yml
+++ b/.github/workflows/analyse-pr.yml
@@ -39,7 +39,7 @@ jobs:
         if: github.event_name == 'pull_request'
         run: |
           mvn -f dhis-2/pom.xml clean install --batch-mode --no-transfer-progress -Psonarqube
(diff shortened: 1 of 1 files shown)


.github/workflows/analyse-pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
       - name: Set up JDK 11
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v2
```
teleivo, line 28: is that one needed? or are you just debugging 😋

enricocolasante: Just shooting in the air a little bit, reverting the last commits

stian-sandvold (approved):

PR #10747, opened 2022-05-09
feat: Add endpoint to hard delete soft deleted relationships [DHIS2-11320]

enricocolasante:


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
```diff
@@ -146,6 +146,16 @@ public int deleteSoftDeletedProgramStageInstances()
         return result;
     }
 
+    @Override
+    public int deleteSoftDeletedRelationships()
+    {
+        int result = maintenanceStore.deleteSoftDeletedRelationships();
+
+        log.info( "Permanently deleted soft deleted relationships: " + result );
```
teleivo, line 154: I think such a log statement is more for devs. So, I would see this as a debug log. I would not add this line. I think we log too much. As an operator of DHIS2 I don't have to act on this. All is fine, so I would not need to read this line.

Use parametrized messages to not construct the string if the log level is disabled.
https://www.slf4j.org/faq.html#logging_performance

You also have an audit statement later on, so this information is available in the audit already.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -137,6 +138,33 @@ public int deleteSoftDeletedProgramStageInstances()
...
     }
 
+    @Override
+    public int deleteSoftDeletedRelationships()
+    {
+
+        List<String> deletedRelationships = getDeletionEntities(
+            "(select uid from relationship where deleted is true)" );
+
+        String relationships = "(select relationshipid from relationship where deleted is true)";
```
teleivo, line 148: if we are not reusing this var, please inline it.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -137,6 +138,33 @@ public int deleteSoftDeletedProgramStageInstances()
         return result;
     }
 
+    @Override
+    public int deleteSoftDeletedRelationships()
+    {
+
+        List<String> deletedRelationships = getDeletionEntities(
+            "(select uid from relationship where deleted is true)" );
```
teleivo, line 146: When is the deletion going to run? Asking since I don't know how many rows this might return, from looking at this it could be many 🤷🏻 Do we have a mechanism for "pacing" such an operation like a DB rate limiter 😋 Or is this something only an admin will do, so they will take care and do this when appropriate.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -137,6 +138,33 @@ public int deleteSoftDeletedProgramStageInstances()
...
+        String[] sqlStmts = new String[] {
+            "delete from relationshipitem where relationshipid in " + relationships,
+            "delete from relationship where deleted is true" };
+
+        int result = jdbcTemplate.batchUpdate( sqlStmts )[sqlStmts.length - 1];
+
+        if ( result > 0 && !deletedRelationships.isEmpty() )
+        {
+            auditHardDeletedEntity( deletedRelationships, Relationship.class );
+
```
teleivo, line 162: extra newline

dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
```diff
@@ -210,6 +210,40 @@ public void setUpTest()
...
+        Relationship r = new Relationship();
+        RelationshipItem rItem1 = new RelationshipItem();
+        rItem1.setProgramInstance( programInstance );
+        RelationshipItem rItem2 = new RelationshipItem();
+        rItem2.setTrackedEntityInstance( entityInstance );
+        r.setFrom( rItem1 );
+        r.setTo( rItem2 );
+        r.setRelationshipType( rType );
+        r.setKey( RelationshipUtils.generateRelationshipKey( r ) );
+        r.setInvertedKey( RelationshipUtils.generateRelationshipInvertedKey( r ) );
```
teleivo, line 231: Could you please group this and the other tests into 3 blocks (arrange, act, assert). Its not obvious what method is being tested.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/MaintenanceController.java
```diff
@@ -167,6 +167,15 @@ public void deleteSoftDeletedProgramStageInstances()
         maintenanceService.deleteSoftDeletedProgramStageInstances();
     }
 
+    @RequestMapping( value = "/softDeletedRelationshipRemoval", method = { RequestMethod.PUT,
+        RequestMethod.POST } )
+    @PreAuthorize( "hasRole('ALL') or hasRole('F_PERFORM_MAINTENANCE')" )
+    @ResponseStatus( HttpStatus.NO_CONTENT )
```
teleivo, line 173: A bit odd that we don't return a bit of info on the operation like how many deletions took place. Since we don't do this for the other endpoints its consistent not to return anything here 👌🏻

teleivo (approved):
Some small suggestions 😄

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
stian-sandvold: This is a once in a blue moon action, and we also log for all similar operations. I think logging is fine, but as @teleivo  says, info might not be the right level. But thats something that should be changed for all the logs in that case.

To summarize, I think it's fine to add it with info as it is now, since thats what the routine has been for this kind of action. If we do decide to increase log level, we should create a new ticket.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
stian-sandvold: Only admins will do this. But the number of records is hard to predict. I would expect people only to do this in special cases. We dont know the number of rows though, so its not a bad idea to think about how to pace this. But again, as my previous comment, this would be relevant for all of these "hard delete soft deleted values" actions.

stian-sandvold (approved):

PR #10752, opened 2022-05-10
fix: DataValue.dataElement is a MetadataIdentifier DHIS2-12563

teleivo:
## This PR (final field migration :tada:)

* migrate DataValue.dataElement to MetadataIdentifier

## Background on changes in [DHIS2-12563](https://jira.dhis2.org/browse/DHIS2-12563)

* tracker view (JSON import/export) and domain models are split since users only specify idSchemes in the query parameters (not in the body) and only on import. The metadata identifier fields in the user-facing import/export JSON are strings containing only the identifier without idScheme.

* Attribute.attribute :heavy_check_mark:
* DataValue.dataElement :heavy_check_mark:
* Event.programStage :heavy_check_mark:
* Event.attributeOptionCombo :heavy_check_mark:
* Event.attributeCategoryOptions :heavy_check_mark:
* TrackedEntity.trackedEntityType :heavy_check_mark:
* Relationship.relationshipType :heavy_check_mark:
* {Enrollment :heavy_check_mark:,Event :heavy_check_mark:}.program
* {Enrollment :heavy_check_mark:,Event :heavy_check_mark:,TrackedEntity :heavy_check_mark:}.orgUnit

will become idScheme aware.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EventTrackerConverterService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EventTrackerConverterService.java
@@ -42,6 +42,7 @@ import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.hisp.dhis.category.CategoryOption;
 import org.hisp.dhis.category.CategoryOptionCombo;
+import org.hisp.dhis.common.BaseIdentifiableObject;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.event.EventStatus;
 import org.hisp.dhis.eventdatavalue.EventDataValue;
@@ -138,7 +139,7 @@ public class EventTrackerConverterService
                 DataValue value = new DataValue();
                 value.setCreatedAt( DateUtils.instantFromDate( dataValue.getCreated() ) );
                 value.setUpdatedAt( DateUtils.instantFromDate( dataValue.getLastUpdated() ) );
-                value.setDataElement( dataValue.getDataElement() );
+                value.setDataElement( MetadataIdentifier.ofUid( dataValue.getDataElement() ) );
                 value.setValue( dataValue.getValue() );
                 value.setProvidedElsewhere( dataValue.getProvidedElsewhere() );
                 value.setStoredBy( dataValue.getStoredBy() );
@@ -190,9 +191,14 @@ public class EventTrackerConverterService
             return eventDataValues;
         }
 
+        // Normalize identifiers as EventDataValue.dataElement are UIDs and
+        // payload dataElements can be in any idScheme
         Set<String> dataElements = event.getDataValues()
             .stream()
             .map( DataValue::getDataElement )
+            .map( preheat::getDataElement )
+            .filter( java.util.Objects::nonNull )
+            .map( BaseIdentifiableObject::getUid )
             .collect( Collectors.toSet() );
         for ( EventDataValue eventDataValue : programStageInstance.getEventDataValues() )
         {
@@ -279,7 +285,7 @@ public class EventTrackerConverterService
             eventDataValue.setProvidedElsewhere( dataValue.isProvidedElsewhere() );
             // ensure dataElement is referred to by UID as multiple
             // dataElementIdSchemes are supported
-            DataElement dataElement = preheat.get( DataElement.class, dataValue.getDataElement() );
+            DataElement dataElement = preheat.getDataElement( dataValue.getDataElement() );
             eventDataValue.setDataElement( dataElement.getUid() );
             eventDataValue.setLastUpdatedByUserInfo( UserInfoSnapshot.from( preheat.getUser() ) );
             eventDataValue.setCreatedByUserInfo( UserInfoSnapshot.from( preheat.getUser() ) );
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/SetMandatoryFieldValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/SetMandatoryFieldValidator.java
@@ -27,8 +27,6 @@
  */
 package org.hisp.dhis.tracker.programrule.implementers;
 
-import static org.hisp.dhis.tracker.validation.hooks.ValidationUtils.validateMandatoryDataValue;
-
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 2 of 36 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EventTrackerConverterService.java
```diff
@@ -190,12 +190,19 @@ private List<EventDataValue> getProgramStageInstanceDataValues( TrackerPreheat p
...
         for ( EventDataValue eventDataValue : programStageInstance.getEventDataValues() )
         {
+            // TODO() what is
+            // org.hisp.dhis.eventdatavalue.EventDataValue.getDataElement()
+            // String? always a UID?
+            // we compare to data values from the payload which can be in any
+            // idScheme
+            // can we first get the DataValue out of the preheat, so we can
+            // compare using UIDs?
             if ( !dataElements.contains( eventDataValue.getDataElement() ) )
```
teleivo, line 206: @stian-sandvold and @enricocolasante I need your input on this

org.hisp.dhis.eventdatavalue.EventDataValue.getDataElement() is of type String

Its not clear to me what it represents, I assume a UID. Is that a safe assumption?
If we can safely assume its a UID, I would need to fetch the DataElement from the preheat. Otherwise, I cannot compare it to a DataElement in the payload which can be of any idScheme.

This means that a DataElement in the payload might not be in the preheat => does not exist.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/SetMandatoryFieldValidator.java
```diff
@@ -118,10 +118,12 @@ private List<ProgramRuleIssue> checkMandatoryDataElement( Event event, List<Even
     {
         ProgramStage programStage = bundle.getPreheat().getProgramStage( event.getProgramStage() );
 
+        // TODO() what is this String, always a UID?
         Map<String, EventActionRule> mandatoryDataElementsByActionRule = actionRules.stream()
```
teleivo, line 122: @stian-sandvold and @enricocolasante I need your input on this

I want to make the "String" key a MetadataIdentifier since in `validateMandatoryDataValue` we compare against DataElements from the payload which can be in any idScheme.

What I dont know is what the String represents. Always a UID?
If so same as above. I would need to fetch that from the preheat so I can map it to a MetadataIdentifier. In that case the DataElement from the EventActionRule might not be found in the prehehat.

enricocolasante: I really think here the key is always as UID as in rule engine we do not have the idScheme concept.
So if a user creates a rule involving a data element it must always use the UID to identify it

teleivo: And do we ensure the user can only create a rule with data element that exists in the DB (thus tracker preheat)? Because I need to validate based on the idScheme in the payload, thus fetch the rule data element from the preheat

enricocolasante: I think we are sure that this is true when they create the rule. I am not sure we ensure that this will be true forever, let's say if someone delete the data element, I don't think we are deleting/invalidating the rule.

But @zubaira or @Bekkalizer should know better about it

Bekkalizer: Hey! Rule actions have a foreign key constraint for both data elements, tracked entity attributes and other metadata object types. So if there is a data element linked to the rule action, it would always exist in the database. We would not be certain that the user can read it(with the metadata:read sharing level), but it would always be in the database.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/EventTrackerConverterService.java
stian-sandvold: eventdatavalue.getDataElement would be a String of type UID. We had an issue where it could be stored with any identifier, but I dont see the point of creating any extra complicated code for that when it should be fixed as its own thing :)

enricocolasante: @stian-sandvold will know better but I cannot see how dataElement here could be something different from UID.
If at any time we saved this value as any idScheme that was used in the importer it seems to me that we lost this information now

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHook.java
```diff
@@ -97,16 +98,15 @@ private void validateMandatoryDataValues( Event event, ValidationErrorReporter r
...
-            .map( de -> preheat.getIdSchemes().getDataElementIdScheme()
-                .getIdentifier( de.getDataElement() ) )
+            .map( de -> preheat.getIdSchemes().toMetadataIdentifier( de.getDataElement() ) )
             .collect( Collectors.toList() );
-        List<String> missingDataValue = validateMandatoryDataValue( programStage, event,
+        List<MetadataIdentifier> missingDataValue = validateMandatoryDataValue( programStage, event,
             mandatoryDataElements );
         missingDataValue
-            .forEach( de -> reporter.addError( event, E1303, de ) );
+            .forEach( de -> reporter.addError( event, E1303, de.getIdentifierOrAttributeValue() ) );
```
enricocolasante, line 109: We can either pass the identifier as an arg for the error or we can add a parser for MetadataIdentifier in `TrackerReportUtils`.
Personally, I prefer the second option so we do not risk to forget to get the identifier and make the error difficult to read

enricocolasante (approved):
Just one suggestion to think about

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventDataValuesValidationHook.java
teleivo: I am going to address this in https://jira.dhis2.org/browse/DHIS2-12519
I a also like the 2nd option better. The metadata identifier has all the context we need to display the id well 😊

lucaCambi77 (approved):

PR #10758, opened 2022-05-10
fix: Fix SQL queries for deleted relationships [DHIS2-11320]

enricocolasante:
Relationship is now a soft deletable entity. Stores and other services that are using plain sql to retrieve relationships now need to be aware if the relationship is soft deleted or not.
Changes are needed:

- In the old tracker importer when retrieving TEIs and linked Relationships. We need to check `includeDeleted` param.
- In the old tracker importer when retrieving Enrollments and linked Relationships. We need to check `includeDeleted` param.
- In the old tracker importer when retrieving Events and linked Relationships. We need to check `includeDeleted` param.
- When counting relationships using `d2:relationshipCount` function for indicators. @jimgrace please have a careful look at this part

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/function/D2RelationshipCount.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/function/D2RelationshipCount.java
@@ -80,6 +80,8 @@ public class D2RelationshipCount
 
         return "(select count(*) from relationship r" + relationshipIdConstraint +
             " join relationshipitem rifrom on rifrom.relationshipid = r.relationshipid" +
-            " join trackedentityinstance tei on rifrom.trackedentityinstanceid = tei.trackedentityinstanceid and tei.uid = ax.tei)";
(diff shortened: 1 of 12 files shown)


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/function/D2RelationshipCount.java
```diff
@@ -80,6 +80,8 @@ public Object getSql( ExprContext ctx, CommonExpressionVisitor visitor )
 
         return "(select count(*) from relationship r" + relationshipIdConstraint +
             " join relationshipitem rifrom on rifrom.relationshipid = r.relationshipid" +
-            " join trackedentityinstance tei on rifrom.trackedentityinstanceid = tei.trackedentityinstanceid and tei.uid = ax.tei)";
+            " join trackedentityinstance tei on rifrom.trackedentityinstanceid = tei.trackedentityinstanceid and tei.uid = ax.tei"
+            +
```
jimgrace, line 84: Just cosmetic: Looks like spotless may have put the + on a new line. You could put the string continuation on the same line or, maybe better yet, just make a long string on the previous line including the where clause (like in ProgramIndicatorServiceD2FunctionTest.)

PR #10763, opened 2022-05-10
feat: i18n for data integrity checks [DHIS2-12643]

jbee:
### Summary 
Adds i18n support for data integrity description details.
This works both for the programmatic and YAML based checks.

The i18n properties use the pattern

    data_integrity.{name}.{property}

The programmatic checks in addition allow to use the i18n properties to set the `severity` (override the default `WARNING`) as well as to set the `section` (override the default `Other`).

As requested a `displayName` property was added to the `DataIntegrityCheck` record.
The `displayName` is either provided as `data_integrity.{name}.name` in i18n files or falls back to the check name where `_` is replaced with space.

Default `section` for programmatic checks was changed from `Legacy` to `Other`.

Detail text for all programmatic checks were provided to the best of my knowledge since these otherwise would not have that information. YAML based checks usually provide the details so no texts were added. Still there could the internationalized in the same manner.

### Added Filters
While testing this it felt clumsy that one could not filter the list of all checks returned by `GET /api/dataIntegrity`.
Therefore the `checks` filter that is available for any of the _run_ methods in the API as also added to `GET`, for example

    GET /api/dataIntegrity?checks=program*

In addition a `section` filter was added, for example

    GET /api/dataIntegrity?section=Program Rules

These two can also be combined.

### Automatic Tests
New test scenarios were added to test the added filters and the mapping of the translation texts to properties.

### Manual Testing
* test the URLs provided in the above description
* observe the difference between the matches
* check that all checks `GET /api/dataIntegrity` have a `name`, `displayName`, `section` and `description`

### Documentation
https://github.com/dhis2/dhis2-docs/pull/1000


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
```diff
@@ -517,22 +519,36 @@ private void registerNonDatabaseIntegrityCheck( DataIntegrityCheckType type,
...
+            log.error( "Failed to register data integrity check " + type, ex );
+        }
     }
 
     /**
      * Maps all "hard coded" checks to their implementation method and registers
      * a {@link DataIntegrityCheck} to perform the method as
      * {@link DataIntegrityDetails}.
      */
-    @PostConstruct
```
jbee, line 535: FYI: I needed to defer this further as now when i18n is used this would need setting which would need hibernate which would fail at the stage this needed to happen.

larshelge (approved):

mortenoh (approved):

PR #10770, opened 2022-05-11
fix: Initialize all collections related to the current user

netroms:
Jira: https://jira.dhis2.org/browse/DHIS2-13256

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
@@ -34,6 +34,9 @@ import java.util.Map;
 
 import lombok.extern.slf4j.Slf4j;
 
+import org.hisp.dhis.category.CategoryOptionGroupSet;
+import org.hisp.dhis.common.DimensionalObject;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.core.userdetails.UserDetails;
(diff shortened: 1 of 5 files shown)


jbee (approved):

larshelge:
Tests are failing. Did we consider the performance impact of the eager fetching? I assume this only happens once on login?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthenticationListener.java
```diff
@@ -137,7 +137,14 @@ private void registerSuccessfulLogin( String username )
...
         {
             user.updateLastLogin();
-            userService.updateUser( user );
+            try
+            {
+                userService.updateUser( user );
+            }
+            catch ( Exception e )
+            {
+                log.warn( "Failed to update the user!", e );
```
vilkg, line 146: We should make a note of this and check the logs to see what the exception is. A user should be updated with last logged in info, so if it's not it would be a functional bug.

vilkg (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -173,20 +176,96 @@ public static <T> T getUserSetting( UserSettingKey key )
...
+        {
+            if ( dimension == null )
+            {
+                continue;
+            }
+            initializeAndUnproxy( dimension.getItems() );
+            initializeAndUnproxy( dimension.getAttributeValues() );
+            initializeAndUnproxy( dimension.getFavorites() );
+            initializeAndUnproxy( dimension.getUserAccesses() );
+            initializeAndUnproxy( dimension.getUserGroupAccesses() );
```
vietnguyen, line 252: no need to initialize `UserAccesses` and `UserGroupAccesses` here. Those are just DTO collections, mapped from jsonb `dimension.sharing.users` and `dimension.sharing.userGroups`. 

Please note that  each element in collection  `dimension.getUserGroupAccesses()` contains only `uid` and `access` string

```json
 "userGroupAccesses": [
            {
                "id": "hj0nnsVsPLU",
                "access": "rw------"
            },
            {
                "id": "qMjBflJMOfB",
                "access": "r-------"
            }
        ]
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthenticationListener.java
netroms: This is due to the fact that the OAuth2 client/user is not a real user and hence the lookup will fail. So in any other cases this should not happen. Consider this a hack, since there is no other reliable way to know what user type it is right now. This problem will disappear when we remove the old OAuth2 support or we do a bigger refactor to include User "type" information into the User "object".

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -173,20 +176,96 @@ public static <T> T getUserSetting( UserSettingKey key )
...
-        initializeAndUnproxy( user.getDataViewOrganisationUnits() );
+        for ( UserGroup group : user.getGroups() )
+        {
+            if ( group == null )
+            {
+                continue;
+            }
+            initializeAndUnproxy( group.getManagedByGroups() );
+            initializeAndUnproxy( group.getManagedGroups() );
+            initializeAndUnproxy( group.getMembers() );
```
larshelge, line 216: This can potentially be very expensive. On some instances they can have many thousand users/members, and many hundreds of user groups, e.g. for PEPFAR. This will be very heavy to fetch on every login.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
netroms: Ok, good point. I will remove those.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
netroms: Maybe we can do some tests on some DBs that is similar to get a idea how expensive it is. Another thing to consider might be to refactor the usage of user groups to be used through a service, so we do more efficient caching etc.

PR #10772, opened 2022-05-11
fix: prevent duplicate relationships from/to events DHIS2-13263

teleivo:


dhis-2/dhis-api/src/test/java/org/hisp/dhis/relationship/RelationshipKeyTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.jupiter.api.Test;
+
+class RelationshipKeyTest
+{
+
+    @Test
+    void asStringForRelationshipTeiToTei()
+    {
+
+        RelationshipKey key = RelationshipKey.of( "dDrh5UyCyvQ",
```
lucaCambi77, line 41: Maybe we can also avoid empty lines here?

lucaCambi77 (approved):

PR #10777, opened 2022-05-11
fix: Indicator summing boolean values DHIS2-13227

jimgrace:
See [DHIS2-13227](https://jira.dhis2.org/browse/DHIS2-13227). Since 2.36.7, expressions can process non-numeric data elements. To support this, expressions do type-checking based on the data element's ValueType. For example, a boolean data element can be used as the true or false test in an if() function within a validation rule or predictor. However this introduced a problem because previously a boolean data data element could be used in an expression to produce a numeric value. This is because booleans are stored as numeric values (1 or 0) in the analytics tables and can be aggregated to show the count of true values (1's). The type checking broke this functionality.

The fix is to allow for boolean values that will be aggregated in an indicator to be considered to return a numeric value for the purpose of type checking.

However boolean, text, and date values within subexpressions are parsed as the types they are, to be used that way in subexpressions.

In `DimItemDataElementAndOperand`, SQL is added to cast a boolean value (stored as a double value) to boolean so it can be used that way in subexpressions. (It must be cast first to an int and then to a bool, since PostgreSQL does not allow casting directly from double to bool.)

Integration test cases were added to `AnalyticsServiceTest` to verify the handling of booleans as numeric when aggregated in indicators, and accessing boolean, text, and date values as the types they are within subexpressions.

`AnalyticsServiceTest` was also refactored to allow the same indicator to be updated for various indicator tests, to make it easier to add additional indicator tests without having to add a new indicator for each test with a new identifier character. The indicator tests were renamed to say what functionality they are testing, not which indicator they are using.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceTest.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/AnalyticsServiceTest.java
@@ -32,6 +32,7 @@ import static org.hisp.dhis.analytics.AggregationType.COUNT;
 import static org.hisp.dhis.analytics.AggregationType.NONE;
 import static org.hisp.dhis.analytics.AggregationType.SUM;
 import static org.hisp.dhis.common.ValueType.BOOLEAN;
+import static org.hisp.dhis.common.ValueType.DATE;
 import static org.hisp.dhis.common.ValueType.INTEGER;
 import static org.hisp.dhis.common.ValueType.TEXT;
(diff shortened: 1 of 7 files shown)


maikelarabori (approved):
Hi @jimgrace, it looks good to me 👍 

I have just one question:
- I saw you made a small change in the validation. Reading the description of your PR, I remembered we had an issue with the validation of boolean values/expressions in the UI itself. I believe it consumes a validation service behind the scenes. The issue is https://jira.dhis2.org/browse/DHIS2-12657.
The validation changes you made in this PR should also fix that ticket? Or is it related to another validation service/flow?

Thanks!
Maikel

larshelge (approved):

PR #10780, opened 2022-05-12
refactor: RelationshipKey is the only responsible to create keys [DHIS2-11320]

enricocolasante:
We were generate relationship keys in 2 different ways to search them in the DB and to put them in the Preheat.
This refactor remove this duplication and cleans up the way we are managing relationship keys to prepare the implementation of idScheme for RelationshipType.

Changes:

- We are not collecting anymore relationship keys in the `TrackerIdentifierCollector`
- A new supplier to retrieve duplicate relationships was created. Removed the logic from `RelationshipStrategy`
- Moved some logic from `RelationshipKeySupport` to `RelationshipUtils` where it belongs
- In the `DuplicateRelationshipSupplier` we look for relationships that have the same key as the relationships in the payload and if we find them, we store only the key in the Preheat. If a relationship is an update, the key will be present in the preheat but it will never be accessed as relationship cannot be updated and when tracker detect that it is an update it will skip validation.

The key that we save in the DB and the key that we save and use in the Preheat are created using the UID of RelationshipType. When looking in the DB for duplicate relationships we need to make sure that we use the UID of RelationshipType to create the key.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/RelationshipTrackerConverterService.java
```diff
@@ -179,8 +180,10 @@ else if ( relationshipType.getToConstraint().getRelationshipEntity().equals( PRO
 
         toRelationship.setFrom( fromItem );
         toRelationship.setTo( toItem );
-        toRelationship.setKey( RelationshipUtils.generateRelationshipKey( toRelationship ) );
-        toRelationship.setInvertedKey( RelationshipUtils.generateRelationshipInvertedKey( toRelationship ) );
+        RelationshipKey relationshipKey = RelationshipKeySupport.getRelationshipKey( fromRelationship,
```
teleivo, line 183: why do we now use `fromRelationship` and before we were using `toRelationship` to get the RelationshipKey?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/TrackerPreheat.java
```diff
@@ -268,14 +269,15 @@ public MetadataIdentifier getCategoryOptionComboIdentifier( CategoryCombo catego
...
+     * Internal set of all relationship keys and inverted keys already present
+     * in the DB. This is used to validate only newly create relationships as
+     * update is not allowed for relationships. The key is a string
+     * concatenating the relationshipType uid, the uid of the `from` entity and
+     * the uid of the `to` entity. The inverted key is a string concatenating
+     * the relationshipType uid, the uid of the `to` entity and the uid of the
+     * `from` entity.
      */
-    private final Map<String, Relationship> duplicatedRelationships = new HashMap<>();
+    private final Set<String> duplicatedRelationships = new HashSet<>();
```
teleivo, line 280: I just thought that the name is a bit misleading. I mean we do not allow duplicate relationships. So `existingRelationships` is more accurate, no?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/TrackerPreheat.java
```diff
@@ -643,11 +645,10 @@ public Relationship getRelationship( org.hisp.dhis.tracker.domain.Relationship r
 
     public boolean isDuplicate( org.hisp.dhis.tracker.domain.Relationship relationship )
     {
-        RelationshipType relationshipType = get( RelationshipType.class, relationship.getRelationshipType() );
-
         if ( hasRelationshipKey( relationship ) )
         {
-            RelationshipKey relationshipKey = getRelationshipKey( relationship );
+            RelationshipType relationshipType = get( RelationshipType.class, relationship.getRelationshipType() );
+            RelationshipKey relationshipKey = getRelationshipKey( relationship, relationshipType.getUid() );
```
teleivo, line 651: do we always call `relationshipType.getUid()` when calling `getRelationshipKey()`?

if so we should move that `getUid()` inside of `getRelationshipKey()`
I was just reminded because in the PR description you said we should make sure to always store the UID of the relationship key. This way we make sure we do.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/supplier/DuplicateRelationshipSupplier.java
```diff
@@ -0,0 +1,75 @@
...
+    private final RelationshipStore relationshipStore;
+
+    @Override
+    public void preheatAdd( TrackerImportParams params, TrackerPreheat preheat )
+    {
+        List<org.hisp.dhis.relationship.Relationship> relationships = retrieveRelationshipKeys(
+            params.getRelationships() );
+
+        relationships.forEach( r -> {
+            preheat.addDuplicatedRelationship( r.getKey() );
```
teleivo, line 57: why did you move the logic out of `addDuplicatedRelationship` and in here?

I think it belongs in `addDuplicatedRelationship`. Only then can you make sure that what is stored is not just any String but a RelatiohshipKey. As well as that the inverted key is stored for bidirectional rels.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/util/RelationshipKeySupport.java
```diff
@@ -89,35 +64,11 @@ public static RelationshipKey.RelationshipItemKey getRelationshipItemKey(
...
-    }
-
-    private static String getUidOrEmptyString( BaseIdentifiableObject baseIdentifiableObject )
-    {
-        return Objects.isNull( baseIdentifiableObject ) ? ""
-            : StringUtils.trimToEmpty( baseIdentifiableObject.getUid() );
-    }
-
-    public static boolean isRelationshipPreheatKey( String s )
+    public static RelationshipKey getRelationshipKey( Relationship relationship, String relationshipType )
```
teleivo, line 67: I am not a fan of utils 😋 

IMHO these are nicer factories for RelationshipKey/RelationshipKey.RelationshipItemKey
Since they depend on our tracker domain I agree that they should not live in `org.hisp.dhis.relationship.`.

In what packages is this class used? If its only used in preheat I think preheat is a good place.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/DuplicateRelationshipSupplierTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private TrackedEntityInstance teiA, teiB, teiC;
+
+    @Mock
+    private TrackerPreheat preheat;
+
+    @Mock
+    private RelationshipStore relationshipStore;
+
+    @InjectMocks
+    private DuplicateRelationshipSupplier supplierToTest;
```
teleivo, line 103: for me just `supplier` is meaningful enough

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/RelationshipUtils.java
```diff
@@ -44,9 +49,7 @@
      */
     public static String generateRelationshipKey( Relationship relationship )
```
teleivo, line 50: why is this class needed?

I think we have too many "utils"
* RelationshipUtils
* RelationshipKeySupport

It would be great if these could be in one place/package. The issue with things getting out of sync like we had them before might also one of ease of finding things.

teleivo (commented):
just a few small questions before ✅ 
do we still need the parsing function (complement of asString) => https://github.com/dhis2/dhis2-core/blob/31838dd72325a3156bb4e421b0e770edd8c487f6/dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipKey.java#L57-L79

its unused and out of sync with how the key looks like. So you cannot call fromString(key.asString())

Its an unused method, so if we are not going to need it please remove it 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/util/RelationshipKeySupport.java
enricocolasante: It shouldn't be in Preheat package as it is also used in the converter.
They could live in tracker domain objects but now that we are passing the RelationshipType to create the key it seems forced have them there.

`getRelationshipItemKey` could stay in `RelationshipItem` object as it doesn't depend on anything else

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/RelationshipUtils.java
enricocolasante: We are generating keys in a lot of places outside tracker.
We could remove `RelationshipKeySupport` but then we need to convert tracker domain `Relationship` to api domain `Relationship` every time we need to generate a key

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/RelationshipTrackerConverterService.java
enricocolasante: I think it was conceptually wrong before even though it was working.
We are converting `fromRelationship` into `toRelationship` so we should use `fromRelationship` to generate all the fields to create the target object.

Using `toRelationship` is working only if we generate the key after having populated all the other fields, `fromRelationship` already have all those fields populated

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/RelationshipUtils.java
teleivo: Can we merge these two classes? I dont fully get why we need RelationshipUtils and RelationshipKeySupport

Maybe there is a reason these 2 classes have to exist separately. The current names just don't tell me what they are for. I would also not know where to add changes based on the names.

I haven't looked at them in details so I dont have a more concrete suggestion. We can also leave them as is until we have a better idea.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preheat/supplier/DuplicateRelationshipSupplier.java
enricocolasante: I think that in the end I was doing like that because it was easier to test 😬 
But I agree that it should be put back in the Preheat

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/RelationshipUtils.java
enricocolasante: We really cant merge them as they have different scopes.
RelationshipKeySupport does the same exact things of RelationshipUtils but on tracker objects.
We could later think of removing `RelationshipUtils` and put the logic inside `org.hisp.dhis.relationship.Relationship` but then I don't think we can do the same for `RelationshipKeySupport` and `org.hisp.dhis.tracker.domain.Relationship` as we are creating the keys using an external relationshipType due to the idScheme mechanism

teleivo (approved):

lucaCambi77 (approved):

PR #10826, opened 2022-05-16
fix: Slow PSQL queries when retrieving /events [ DHIS2-12039 ]

lucaCambi77:
This is an attempt to improve performance while accessing /events endpoint.
An index has been added to programstageinstance table for **lastupdated** column, which is ordering the result set.

Main reasons for this attempt :
1) This is significantly reducing the cost of the queries
2) During local performance tests, the distribution of the response time during around 5 minutes has moved a bit to lower percentiles and reduced the average response time over a much higher Request per second.
Locally psi table is not big as in a production env though, so results need to be validated

Also, a minor update not to retrieve a full Tei for existence

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventSearchParams.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventSearchParams.java
@@ -45,7 +45,6 @@ import org.hisp.dhis.program.Program;
 import org.hisp.dhis.program.ProgramStage;
 import org.hisp.dhis.program.ProgramStatus;
 import org.hisp.dhis.program.ProgramType;
-import org.hisp.dhis.trackedentity.TrackedEntityInstance;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.webapi.controller.event.mapper.OrderParam;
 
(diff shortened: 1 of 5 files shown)


teleivo:
@lucaCambi77 I think it would be helpful if you would post some numbers/summary from your local performance tests. Descriptions like `a bit lower`, `much higher` do not tell us much. Adding an index also comes with a cost. So that should be warranted.

@vilkg can we run the changes from this branch on the setup you have with a bigger dataset? To validate this does improve things before we merge the changes.

lucaCambi77:
@teleivo I am not able to post an HTML report as it is not supported here. There are some improvements locally, we just need to test if this makes improvements in a bigger dataset. If not we can just revert the change.

teleivo:
> @teleivo I am not able to post an HTML report as it is not supported here. There are some improvements locally, we just need to test if this makes improvements in a bigger dataset. If not we can just revert the change.

you can upload a picture of the HTML 😄

lucaCambi77:
It is a report with different parts, not easy to take a pic. Let's see if there is an improvement. If we can avoid merging first it is of course better

vilkg:
> 
> @vilkg can we run the changes from this branch on the setup you have with a bigger dataset? To validate this does improve things before we merge the changes.

yeah, Luca already asked me. We will. :)

enricocolasante (approved):

vilkg:
I have compared dev and this branch and I can't see any significant improvements with the requests that I'm trying ( from the issue description). The sql queries also look identical. @lucaCambi77

lucaCambi77:
yes @vilkg the query is the same, only a new index has been added. We'll try something different

larshelge (approved):

PR #10828, opened 2022-05-16
fix: Re-fetch detached current user OrgUnits

netroms:
Fix issue with detached org units in [DHIS2-13287](https://jira.dhis2.org/browse/DHIS2-13287)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationStoreConfiguration.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationStoreConfiguration.java
@@ -33,6 +33,7 @@ import org.hisp.dhis.association.CategoryOptionOrganisationUnitAssociationsQuery
 import org.hisp.dhis.association.DataSetOrganisationUnitAssociationsQueryBuilder;
 import org.hisp.dhis.association.ProgramOrganisationUnitAssociationsQueryBuilder;
 import org.hisp.dhis.cache.CacheProvider;
+import org.hisp.dhis.organisationunit.OrganisationUnitService;
 import org.hisp.dhis.user.CurrentUserService;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationStoreConfiguration.java
```diff
@@ -46,28 +47,29 @@
 
     @Bean( "jdbcProgramOrgUnitAssociationsStore" )
     public JdbcOrgUnitAssociationsStore jdbcProgramOrgUnitAssociationStore( CurrentUserService currentUserService,
-        JdbcTemplate jdbcTemplate )
+        JdbcTemplate jdbcTemplate, OrganisationUnitService unitService )
```
larshelge, line 50: Could we name this `organisationUnitService`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
```diff
@@ -119,7 +122,14 @@ public SetValuedMap<String, String> getOrganisationUnitsAssociations( Set<String
 
     private Set<String> getUserOrgUnitPaths()
     {
-        Set<String> allUserOrgUnitPaths = currentUserService.getCurrentUserOrganisationUnits().stream()
+        Set<OrganisationUnit> currentUserOrganisationUnits = currentUserService.getCurrentUserOrganisationUnits();
```
larshelge, line 125: Just curious, why is this necessary? Are we getting detached or cached org units from the current user service?

netroms: Yes, right now the user is detached after re-trivial on login. This is possible the disable very easy, right now I think we should consider this an ongoing experiment.

PR #10842, opened 2022-05-18
fix: update Option.sortOrder resulted with null sortOrders

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-12828

### Issue
- By default, if we update an OptionSet with Option's sortOrder gaps then Hibernate will auto re-organizes the sortOrders. However, Hibernate can't do that if we execute the update from _Option side of the association_.  Means that we only update Option without touching its OptionSet.

### Fix
- I think the best way to solve this for the metadata import service is to remove and add the updating Option again to OptionSet. 
- Then at then end of the import process when we call `session.flush()` the OptionSet will be updated and all sortOrders will be re-organized.
- This is already handled for create new Option. Currently issue only happens with updating.

Example: [1,2,null,null,null,9] will be re-organized by Hibernate to [1,2,3]

### Test
- Added controller test


jbee (approved):

mortenoh (approved):

larshelge:
Failing integration test seems related:

```
Error:    CsvMetadataImportTest.testOptionSetMergeDuplicate:173->runImport:265 » NullPointer
```

PR #10843, opened 2022-05-18
chore: Remove unused json files in tracker tests [TECH-890]

enricocolasante:


ameenhere (approved):

teleivo (approved):
curious, how come we have so many unused test fixtures 🤔

enricocolasante:
> curious, how come we have so many unused test fixtures 🤔

I guess because they were used at some point then the test or the feature was either changed or deleted and the linked json was never removed

PR #10846, opened 2022-05-18
fix: Fix tracker integration tests to use tracker import service [TECH-890]

enricocolasante:


teleivo:
> A few integration tests were calling `TrackerBundleService` instead of `TrackerImportService`

Can you please explain why that is a problem?

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/event_events_and_enrollment.json
```diff
@@ -555,40 +421,6 @@
       ],
       "completedBy": "admin",
       "completedAt": "2019-01-28T00:00:00.000",
-      "dataValues": [
```
teleivo, line 558: why are you deleting all of these?

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/te_program_with_tea_delete_data.json
```diff
@@ -1,102 +1,12 @@
 {
-  "importMode": "COMMIT",
```
teleivo, line 2: I dont think this should be deleted. These tracker integration tests use the TrackerImportParams JSON.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/bundle/TrackedEntityAttributeTest.java
```diff
@@ -70,6 +73,7 @@ protected void initTest()
         throws IOException
     {
         setUpMetadata( "tracker/te_with_tea_metadata.json" );
+        injectAdminUser();
```
teleivo, line 76: is it necessary/a good idea to use an admin user in our tests?

teleivo (commented):
some questions

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/event_events_and_enrollment.json
enricocolasante: Because they are not relevant for this test and before the validation was not running.
To maintain them here I would need to change a lot of things in the metadata to make them work

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/te_program_with_tea_delete_data.json
enricocolasante: Are you talking about the `importMode` or `idSchemes`?
I really want to have the payload as little as possible and not to deal with unnecessary details here

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/bundle/TrackedEntityAttributeTest.java
enricocolasante: I think this is a big question. And I agree that we should deal with the whole users mechanism in a much smarter way in the tests.
But I think we should discuss about it when we discuss a little more in depth on test strategy.
At the moment, we have almost all integration tests using admin users.
We select a different user only if we are really testing the permissions part.

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/te_program_with_tea_delete_data.json
teleivo: I've added them to all our payloads, since that reflects reality AFAIK. I added them in the migration by deserializing and serializing the JSON again (using the adaptations I made to the TrackerImportParams).

teleivo (approved):
I think you should leave the importMode, idSchemes, importStrategy, ... in the test fixtures. At least for now, so its consistent with all other fixtures. We can have another look on whether we can remove it in all fixtures.

dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/te_program_with_tea_delete_data.json
enricocolasante: I put them back.
We can have another look at them later

teleivo (approved):

PR #10857, opened 2022-05-19
fix: Remove disabled from tracker test [TECH-890]

enricocolasante:


dhis-2/dhis-services/dhis-service-tracker/src/test/resources/tracker/validations/enrollments_te_attr-non-existing.json
```diff
@@ -0,0 +1,93 @@
...
+      "followUp": false,
+      "deleted": false,
+      "storedBy": "admin",
+      "events": [],
+      "relationships": [],
+      "attributes": [
+        {
+          "attribute": {
+            "idScheme": "UID",
+            "identifier": "--BIpOStK--"
```
teleivo, line 59: I don't think that is a valid UID. Can you please change that. Just so we dont risk the test failing because of it being an invalid UID. You can generate a "non-existing" using the CodeGenerator

enricocolasante, line 59:
```suggestion
            "identifier": "QfUVllTs6XX"
```

enricocolasante: Fixed, even though we are not currently validating UIDs for attributes

teleivo (approved):

lucaCambi77 (approved):

PR #10864, opened 2022-05-19
chore: refactoring to makes all things work together [DHIS2-13232]

gnespolino:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/linelisting/trackedentityinstance/TeiAnalyticsQueryService.java
```diff
@@ -65,18 +64,23 @@
...
-        final SqlQueryResult result = (SqlQueryResult) queryExecutor.execute( query );
-        final List<GridHeader> headers = from( ((SqlQuery) query).getColumns() );
+        final SqlQuery query = (SqlQuery) teiJdbcQuery.from( teiLineListingParams );
+        final SqlQueryResult result = queryExecutor.execute( query );
+        final List<GridHeader> headers = from( query.getColumns() );
 
         return gridAdaptor.createGrid( headers, result.result() );
     }
+
+    public void validateRequest( TeiLineListingRequest request )
```
maikelarabori, line 82: In our last discussion, I think we decided to move it into a separate component.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/linelist/trackedentity/TrackedEntityLineListingController.java
```diff
@@ -59,15 +63,15 @@
     @GetMapping( "query/{trackedEntityType}" )
     Grid getGrid(
         @PathVariable String trackedEntityType,
-        TrackedEntityLineListingRequest request,
+        TeiLineListingRequest request,
```
maikelarabori, line 66: I think we decided to pass in two different request objects, so we could remove the class extension.

maikelarabori (commented):
Looks good @gnespolino, I just raised two points we had discussed recently. Please take a look.

Also, I'm not sure that we should use "line list/listing" in the backend. This endpoint can be used by any front-end application/client. I don't think it should be tight to a specific application.

Thanks,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/linelisting/trackedentityinstance/TeiAnalyticsQueryService.java
gnespolino: Ok, I'll extract it into a dedicated tei validation service

PR #10868, opened 2022-05-20
fix: Revert the current user session detachment

netroms:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserDetailsImpl.java
```diff
@@ -0,0 +1,112 @@
...
+    private final boolean accountNonLocked;
+
+    private final boolean credentialsNonExpired;
+
+    private final Collection<GrantedAuthority> authorities;
+
+    private final Map<String, Serializable> userSettings;
+
+    @Override
+    public Map<String, Serializable> getUserSettings()
```
jbee, line 60: class level `@Getter` can get rid of all the getter methods

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
```diff
@@ -429,4 +429,9 @@
      * authority.
      */
     List<User> getUsersWithAuthority( String authority );
+
+    CurrentUserDetails validateAndCreateUserDetails( User user, String password );
+
+    CurrentUserDetailsImpl createUserDetails( User user, String password, boolean accountNonLocked,
```
jbee, line 435: Maybe use `@Builder` on the `CurrentUserDetailsImpl` ?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -105,20 +109,24 @@
...
+    private final SecurityService securityService;
+
     private final Cache<String> userDisplayNameCache;
 
     public DefaultUserService( UserStore userStore, UserGroupService userGroupService,
         UserRoleStore userRoleStore,
         CurrentUserService currentUserService, SystemSettingManager systemSettingManager,
         CacheProvider cacheProvider,
-        @Lazy PasswordManager passwordManager, @Lazy SessionRegistry sessionRegistry )
+        @Lazy PasswordManager passwordManager, @Lazy SessionRegistry sessionRegistry,
```
larshelge, line 120: Is this lazy to due circular dependencies? If so, would be good to resolve them properly.

larshelge (approved):

larshelge (approved):

jbee (approved):

PR #10869, opened 2022-05-20
performance: Try to speed up tracker tests [TECH-890]

enricocolasante:
Tracker tests were quite slow because we were running the setup before every single test case.
Now we are running the setup only once per class and the tests are running into a transaction that is then rolled back by junit framework after the test is complete.

This is also fixing https://jira.dhis2.org/browse/DHIS2-13068

Relevant stats:
`TrackedEntityImportValidationTest` was ~ 20s, now is ~ 4.5s
`EnrollmentImportValidationTest` was 14s, now is ~ 3.8s
`EventImportValidationTest` was 30s, now is ~ 5.6s


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleIntegrationTest.java
```diff
@@ -179,7 +188,7 @@ void testImportEventInProgramStageSuccessWithWarningRaised()
         assertEquals( 3, warningReports.size() );
         assertEquals( TrackerErrorCode.E1308, warningReports.get( 0 ).getWarningCode() );
         assertEquals(
-            "Generated by program rule (`ProgramRule`) - DataElement `DATAEL00001` is being replaced in event `EVENT123456`",
+            "Generated by program rule (`ProgramRulC`) - DataElement `DATAEL00001` is being replaced in event `EVENT123456`",
```
teleivo, line 191: why is that change necessary?

teleivo (commented):
Do I understand correctly that with this PR all changes made to the DB by each test will only be cleaned up after the test class?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleIntegrationTest.java
enricocolasante: We need to talk about this...
There are a few places that are quite difficult to test because of Hibernate Session.
This particular one has something to do with the fact that Hibernate cache doesn't work well if we retrieve a filtered list of program rule actions linked to a program rule.

PR #10876, opened 2022-05-23
chore: ValidationErrorReporter only needs idSchemes DHIS2-12519

teleivo:
* TrackerErrorReport/TrackerWarningReport/ValidationErrorReporter only needs idSchemes to add a new error. Not the entire TrackerBundle. First step in decoupling ValidationErrorReporter from the TrackerBundle. Later PR will look into removing its dependency on it.

* removed unused TrackerBundle.identifier which was always UID and did not reflect the actual idSchemes set by the user. Using it in TrackerReportUtils lead to reports always being in idScheme UID instead of user defined one.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
@@ -139,14 +139,7 @@ public class PreCheckSecurityOwnershipValidationHook
             if ( tei.getProgramInstances().stream().anyMatch( pi -> !pi.isDeleted() )
                 && !user.isAuthorized( Authorities.F_TEI_CASCADE_DELETE.getAuthority() ) )
             {
-                TrackerErrorReport error = TrackerErrorReport.builder()
-                    .uid( trackedEntity.getUid() )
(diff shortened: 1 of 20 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
```diff
@@ -170,7 +163,7 @@ private void checkTeiTypeWriteAccess( ValidationErrorReporter reporter, String t
                 .errorCode( TrackerErrorCode.E1001 )
                 .addArg( user )
                 .addArg( trackedEntityType )
-                .build( bundle );
+                .build( bundle.getPreheat().getIdSchemes() );
```
enricocolasante, line 166: Pass the TEI to the method and use `reporter.addError( trackedEntity, E1001, user, trackedEntityType );`

enricocolasante (approved):
Just a small suggestion

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckSecurityOwnershipValidationHook.java
teleivo: perfect!

lucaCambi77 (approved):

PR #10877, opened 2022-05-23
fix: Potential duplicate status is not validated [ DHIS2-11534 ]

lucaCambi77:
- add booleans for deduplication status for different cases (create, filter, merge)
- fix tests

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java
@@ -27,10 +27,26 @@
  */
 package org.hisp.dhis.deduplication;
 
+import lombok.Getter;
+
+@Getter
 public enum DeduplicationStatus
 {
-    ALL,
-    OPEN,
-    INVALID,
-    MERGED
+    ALL( true, false, false ),
+    OPEN( true, true, false ),
+    INVALID( true, false, false ),
+    MERGED( true, false, true );
+
+    private final boolean forFilter;
+
+    private final boolean forCreation;
+
+    private final boolean forMerge;
+
+    DeduplicationStatus( boolean forFilter, boolean forCreation, boolean forMerge )
+    {
+        this.forFilter = forFilter;
+        this.forCreation = forCreation;
+        this.forMerge = forMerge;
+    }
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java
@@ -32,21 +32,8 @@ import lombok.Getter;
 @Getter
 public enum DeduplicationStatus
 {
-    ALL( true, false, false ),
-    OPEN( true, true, false ),
-    INVALID( true, false, false ),
-    MERGED( true, false, true );
-
-    private final boolean forFilter;
-
-    private final boolean forCreation;
-
-    private final boolean forMerge;
-
-    DeduplicationStatus( boolean forFilter, boolean forCreation, boolean forMerge )
-    {
-        this.forFilter = forFilter;
-        this.forCreation = forCreation;
-        this.forMerge = forMerge;
-    }
+    ALL,
+    OPEN,
+    INVALID,
+    MERGED
 }
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
@@ -67,7 +67,7 @@ public class PotentialDuplicatesActions
         return uid;
     }
 
-    public ApiResponse createPotentialDuplicate( String teiA, String teiB, String status )
+    public ApiResponse postPotentialDuplicate( String teiA, String teiB, String status )
     {
         JsonObject object = new JsonObjectBuilder()
             .addProperty( "original", teiA )
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
@@ -44,7 +44,7 @@ public class PotentialDuplicatesActions
         super( "/potentialDuplicates" );
     }
 
-    public ApiResponse createPotentialDuplicate( String teiA, String teiB)
+    public ApiResponse createPotentialDuplicate( String teiA, String teiB )
     {
         JsonObject object = new JsonObjectBuilder()
             .addProperty( "original", teiA )
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
@@ -44,20 +44,20 @@ public class PotentialDuplicatesActions
         super( "/potentialDuplicates" );
     }
 
-    public ApiResponse createPotentialDuplicate( String teiA, String teiB, String status )
+    public ApiResponse createPotentialDuplicate( String teiA, String teiB)
     {
         JsonObject object = new JsonObjectBuilder()
             .addProperty( "original", teiA )
             .addProperty( "duplicate", teiB )
-            .addProperty( "status", status )
+            .addProperty( "status", "OPEN" )
             .build();
 
         return this.post( object );
     }
 
-    public String createAndValidatePotentialDuplicate( String teiA, String teiB, String status )
+    public String createAndValidatePotentialDuplicate( String teiA, String teiB )
     {
-        return createPotentialDuplicate( teiA, teiB, status )
+        return createPotentialDuplicate( teiA, teiB )
             .validateStatus( 200 )
             .extractString( "id" );
     }
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
@@ -44,7 +44,7 @@ public class PotentialDuplicatesActions
         super( "/potentialDuplicates" );
     }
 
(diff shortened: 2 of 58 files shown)


dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -29,99 +29,90 @@
...
     }
 
-    @CsvSource( {
-        "MERGED,INVALID,false",
-        "OPEN,INVALID,true",
-        "OPEN,MERGED,false",
-        "INVALID,OPEN,true",
-        "MERGED,OPEN,false"
-    } )
+    @CsvSource( { "MERGED" } )
```
teleivo, line 104: If your parametrized test only has one value, make it a simple test like you did in the above test.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/deduplication/DefaultDeduplicationService.java
```diff
@@ -261,7 +262,17 @@ private boolean sameAttributesAreEquals( Set<TrackedEntityAttributeValue> tracke
     @Override
     @Transactional
     public void addPotentialDuplicate( PotentialDuplicate potentialDuplicate )
+        throws PotentialDuplicateConflictException
     {
+        if ( !potentialDuplicate.getStatus().isForCreation() )
+        {
+            throw new PotentialDuplicateConflictException(
+                "Potential Duplicate creation must have one of following status : "
```
teleivo, line 270: `"Potential Duplicate creation must have one of following status : "`

maybe

`"Invalid status %s, creating potential duplicate is allowed using: "`

replacing the `%s` with the one that was set. Adding the actual value gives more context (for example to a stacktrace).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java
```diff
@@ -27,10 +27,26 @@
...
-    ALL,
-    OPEN,
-    INVALID,
-    MERGED
+    ALL( true, false, false ),
+    OPEN( true, true, false ),
+    INVALID( true, false, false ),
+    MERGED( true, false, true );
+
+    private final boolean forFilter;
```
teleivo, line 40: I would not add these fields for the following reasons:

* `forFilter` is never used. All status have `forFilter` set to true.
* `forCreation` is only true for OPEN. Most of the code added in DefaultDeduplicationService can be replaced by just stating that the status has to be OPEN when creating a potential duplicate.
* `forMerge` is only true for MERGE. So it just repeats what we already know.

so to me, this adds complexity (duplicate information) that is not needed.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/deduplication/DefaultDeduplicationService.java
```diff
@@ -261,7 +262,17 @@ private boolean sameAttributesAreEquals( Set<TrackedEntityAttributeValue> tracke
     @Override
     @Transactional
     public void addPotentialDuplicate( PotentialDuplicate potentialDuplicate )
+        throws PotentialDuplicateConflictException
     {
+        if ( !potentialDuplicate.getStatus().isForCreation() )
```
teleivo, line 267: check if status == OPEN and simplify creating the exception below

teleivo (changes requested):
I think the fix can be much simpler as explained in dhis-2/dhis-api/src/main/java/org/hisp/dhis/deduplication/DeduplicationStatus.java

I am open to hearing others thoughts of course 😄

teleivo:
@vilkg I think your review of the e2e test changes would be great 😄

vilkg (changes requested):
I'm going to check out the code a bit later, but from a quick look it seems that we are loosing some coverage here. Some tests were setting up the status ( not OPEN) intentionally and the state was relevant to the tests, but now all duplicates are created with OPEN status. I would prefer PotentialDuplicateActions to still support status param and then change the status of the duplicate via the API within those methods if requested status is not OPEN. 

something like : 
```
 public ApiResponse createPotentialDuplicate( String teiA, String teiB, String status )
    {
        JsonObject object = new JsonObjectBuilder()
            .addProperty( "original", teiA )
            .addProperty( "duplicate", teiB )
            .addProperty( "status", "OPEN" )
            .build();

        String uid = this.post( object );
        
        if ( status == "MERGED") {
           this.autoMergePotentialDuplicate(uid); 
        }

        if ( status == "INVALID") {
            this.invalidatePotentialDuplicate(uid); // iI think it's just a request to /potentialDuplicates/<uid>/invalidate
         }
         return uid;
    }

```

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -29,99 +29,90 @@
...
+    public void shouldFilterByStatus()
     {
-        String teiA = createTei();
-        String teiB = createTei();
+        // OPEN
+        potentialDuplicatesActions.createAndValidatePotentialDuplicate( createTei(), createTei() );
 
-        potentialDuplicatesActions.createPotentialDuplicate( teiA, teiB, status ).validate().statusCode( 200 );
+        // INVALID
+        String duplicateId = potentialDuplicatesActions.createAndValidatePotentialDuplicate( createTei(), createTei() );
```
vilkg, line 58: why is this invalid?

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -29,99 +29,90 @@
...
 
-        ApiResponse response = potentialDuplicatesActions.get( "", new QueryParamsBuilder().add( "status=" + status ) );
-        response
+        potentialDuplicatesActions.update( duplicateId + "?status=" + "INVALID", new JsonObjectBuilder().build() )
             .validate()
-            .body( "identifiableObjects", hasSize( greaterThanOrEqualTo( 1 ) ) )
```
vilkg, line 67: Looks like you had to comment these assertions because there are no longer invalid duplicates. That means the intention of the test has changed and we lost coverage.

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -29,99 +29,90 @@
...
+            .body( "identifiableObjects", hasSize( greaterThanOrEqualTo( 3 ) ) )
             .body( "identifiableObjects.status",
                 allOf( hasItem( "OPEN" ), hasItem( "INVALID" ), hasItem( "MERGED" ) ) );
+
+        potentialDuplicatesActions.get( "", new QueryParamsBuilder().add( "status=" + "OPEN" ) )
+            .validate()
+            .body( "identifiableObjects", hasSize( greaterThanOrEqualTo( 1 ) ) )
+            .body( "identifiableObjects.status", everyItem( equalTo( "OPEN" ) ) );
+
+        potentialDuplicatesActions.get( "", new QueryParamsBuilder().add( "status=" + "INVALID" ) )
```
vilkg, line 82: I don't really see where `INVALID` potential duplicate was created? Only 1 `MERGED` should exist here since you only create one potential duplicate and change the status to `MERGED`.

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
vilkg: We lost some coverage here again and yes, if there's only value there's no need for parametrization

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/actions/tracker/PotentialDuplicatesActions.java
```diff
@@ -44,22 +44,38 @@ public PotentialDuplicatesActions()
...
+        if ( status.equals( "INVALID" ) )
+        {
+            this.update( uid + "?status=INVALID", new JsonObjectBuilder().build() ).validateStatus( 200 );
+        }
+
+        return uid;
     }
 
-    public String createAndValidatePotentialDuplicate( String teiA, String teiB, String status )
+    public ApiResponse createPotentialDuplicate( String teiA, String teiB, String status )
```
vilkg, line 70: I think the naming here is a bit misleading. Since the contract changed, only OPEN status will **create** a duplicate. I would change the name to "postPotentialDuplicate", or add a javadoc explaining that.

teleivo (approved):
Code LGTM. I leave the approval of the e2e test changes to @vilkg

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -29,60 +29,59 @@
...
+public class PotentialDuplicatesTests extends PotentialDuplicatesApiTest
 {
     @BeforeEach
     public void beforeEach()
     {
         loginActions.loginAsAdmin();
     }
 
-    @ParameterizedTest
-    @ValueSource( strings = { "OPEN", "INVALID", "MERGED" } )
```
vilkg, line 56: Is there a reason why you wanted to get rid of the parameters? In my mind it makes the test look a bit cleaner and also runs the test for every status regardless of failures in other statuses. With the change, validation errors in, say OPEN filter, will block other checks from executing.

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicatesTests.java
```diff
@@ -93,35 +92,37 @@ public void shouldRequireBothTeis()
             .statusCode( equalTo( 400 ) )
             .body( "status", equalTo( "ERROR" ) )
             .body( "message", containsStringIgnoringCase( "missing required input property" ) );
+
+        potentialDuplicatesActions.createPotentialDuplicate( createTei(), null, "OPEN" )
```
vilkg, line 96: I'm wondering if this duplicate check is necessary or is it better to cover it in unit tests. e2e tests are generally time-consuming and checking both inputs instead of one might not add any meaningful coverage.

vilkg (approved):

PR #10887, opened 2022-05-24
chore: simplify TrackerErrorReport/WarningReport DHIS2-12519

teleivo:
* ReportUtils (now MessageFormatter) is actually responsible for formatting the report message. Thus, name it MessageFormatter (`report` context is already part of the package).
* TrackerErrorReport/WarningReport do not need builders. All their fields are required. So it should only have an all args constructor. Builders are great with lots of optional fields.
* TrackerErrorReport/WarningReport do not need to depend on idSchemes. The ValidationErrorReporter is already idScheme aware. ValidationErrorReporter can instantiate the TrackerErrorReport using a MessageFormatter formatted message.
* add docs and tests to MessageFormatter

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/MessageFormatter.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/MessageFormatter.java
@@ -30,9 +30,9 @@ package org.hisp.dhis.tracker.report;
 import java.text.DateFormat;
 import java.text.MessageFormat;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.hisp.dhis.category.CategoryOption;
@@ -61,11 +61,6 @@ class MessageFormatter
         // not meant to be inherited from
     }
 
-    protected static List<String> buildArgumentList( TrackerIdSchemeParams params, List<Object> arguments )
-    {
-        return arguments.stream().map( arg -> parseArgs( params, arg ) ).collect( Collectors.toList() );
-    }
-
     /**
      *
      * @param idSchemes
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/MessageFormatter.java
```diff
@@ -77,36 +103,36 @@ else if ( MetadataIdentifier.class.isAssignableFrom( argument.getClass() ) )
...
         else if ( ProgramStage.class.isAssignableFrom( argument.getClass() ) )
         {
-            return getIdAndName( idSchemeParams.toMetadataIdentifier( (ProgramStage) argument ),
+            return getIdAndName( idSchemes.toMetadataIdentifier( (ProgramStage) argument ),
                 (ProgramStage) argument );
         }
         else if ( IdentifiableObject.class.isAssignableFrom( argument.getClass() ) )
         {
-            return getIdAndName( idSchemeParams.toMetadataIdentifier( (IdentifiableObject) argument ),
+            return getIdAndName( idSchemes.toMetadataIdentifier( (IdentifiableObject) argument ),
```
enricocolasante, line 135: All the previous Metadata classes should implement IdentifiableObject.
Would be safe to just remove them?

enricocolasante (approved):
Just a suggestion

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/MessageFormatter.java
teleivo: We support overriding the general `idScheme` param with metadata specific one like `dataElementIdScheme`, ... see https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/new-tracker.html#request-parameters

Since generic type information is not available at runtime I used method overloading.
idSchemes.toMetadataIdentifier( IdentifiableObject metadata ) will use the `idScheme` parameter.

idSchemes.toMetadataIdentifier( DataElement metadata ) will use the `dataElementIdScheme` and so on.

So I cannot remove them, otherwise it would not report in the correct idScheme.

ameenhere (approved):

PR #10898, opened 2022-05-25
fix: Fix audit and import error message for relationships [DHIS2-11320]

enricocolasante:
- Audit was not working for hard deleted relationships
- Soft deleted relationships were not considered when validating in the old importer

### DIFF
--- a/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/teis/RelationshipsTest.java
+++ b/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/teis/RelationshipsTest.java
@@ -47,7 +47,6 @@ import org.hisp.dhis.actions.tracker.RelationshipActions;
 import org.hisp.dhis.actions.tracker.TEIActions;
 import org.hisp.dhis.dto.ApiResponse;
 import org.hisp.dhis.helpers.QueryParamsBuilder;
-import org.hisp.dhis.helpers.TestCleanUp;
 import org.hisp.dhis.helpers.file.FileReaderUtils;
 import org.hisp.dhis.tracker.TrackerApiTest;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/teis/RelationshipsTest.java
```diff
@@ -149,6 +148,37 @@ public void duplicateRelationshipsShouldNotBeAdded()
...
+
+        response.validate().statusCode( 409 )
+            .body( "status", equalTo( "ERROR" ) )
+            .body( "response.status", equalTo( "ERROR" ) )
+            .body( "response.ignored", equalTo( 1 ) )
+            .body( "response.total", equalTo( 1 ) )
+            .rootPath( "response.importSummaries[0]" )
+            .body( "status", equalTo( "ERROR" ) )
+            .body( "description", Matchers.containsString(
+                "Relationship '" + createdRelationship + "' is already deleted and cannot be modified." ) )
```
teleivo, line 178: Why did you go for an e2e test here? Could we not test this in a unit test of AbstractRelationshipService

I just imagine making a change to AbstractRelationshipService that breaks this e2e test and how long it takes me to realize that 😅 might be 20min vs < 1min

teleivo (approved):
code LGTM, the only thing I wonder is whether we should write a unit test instead of the e2e test. not sure if the e2e test offers us more safety

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/teis/RelationshipsTest.java
enricocolasante: I agree with you.
I think I am quite skeptical when I am working with the old importer, I thought a unit test was going to be much more difficult.

ameenhere (approved):

PR #10901, opened 2022-05-26
fix: Potential bug with analytics api when requesting event data from two programs that share a data element[2.39-DHIS2-12466]

d-bernat:
please see: [https://jira.dhis2.org/browse/DHIS2-12466](https://jira.dhis2.org/browse/DHIS2-12466)


maikelarabori (approved):
The code looks good @d-bernat.

gnespolino (approved):

PR #10906, opened 2022-05-27
fix: collect TrackedEntity/RelationshipType from payload

teleivo:
so we fetch them with attributes and values in case the idScheme is
ATTRIBUTE. Otherwise, idScheme ATTRIBUTE will not work for these
metadata types.

Caching is currently disabled entirely, so we will not have any
performance penalty. When caching is fixed it should be with idSchemes
in mind.


enricocolasante (approved):
Maybe we could add a test to assert that we are collecting the TrackedEntityType and the RelationshipType identifiers

ameenhere (approved):

PR #10914, opened 2022-05-28
fix: Fix audit for relationship deletion [DHIS2-11320]

enricocolasante:


teleivo (commented):
Do we have a test to prevent a regression?

enricocolasante:
> Do we have a test to prevent a regression?

No, we don't.
Testing audit is one of the thing that is not easy to test and we do not have a way to test it for now.
It should probably be a separate issue

teleivo (approved):

vilkg (approved):

PR #10918, opened 2022-05-29
fix: Prog. Ind. boolean as bool/number DHIS2-12657

jimgrace:
See [DHIS2-12657](https://jira.dhis2.org/browse/DHIS2-12657). Analytics boolean values are stored as numbers (1 = true, 0 = false). This is done for both aggregate data which can be queried by indicators and event data which can be queried by program indicators.

In the case of aggregate data, ticket [DHIS2-13227](https://jira.dhis2.org/browse/DHIS2-13227) allows boolean data to be used in indicator expressions as either numeric or boolean. It does so by considering the datatype of boolean data to be numeric. When a boolean values is used, such as in the first argument to an if() statement inside an expression, the integer value is automatically converted to boolean (zero = false, nonzero = true). This was implemented because existing indicators had relied on the fact that boolean data elements are represented as numbers, and their values can be summed within an indicator for uses such as "count all the organisation units where the value is true".

Likewise, this PR allows boolean event data to be used in program indicators as either numeric or boolean. This is because some existing program indicators had relied on the fact that boolean data elements are represented as numbers. In the example in ticket [DHIS2-12657](https://jira.dhis2.org/browse/DHIS2-12657), a boolean data element is multiplied by a numeric data element to return the numeric value if true (1 * value), or zero if false (0 * value). This change also makes sense because it allows for consistent evaluation of boolean values between indicators and program indicators.

The implementation of this PR is different from the implementation of [DHIS2-13227](https://jira.dhis2.org/browse/DHIS2-13227) because of the differences in the way indicators and program indicators are processed. With indicators, data is first fetched from the database while being aggregated, and then the indicator expression is evaluated in Java using the fetched, aggregated values. As the expression is evaluated in Java, boolean values (stored as numeric) can easily be converted to boolean on the fly if a boolean value is expected.

By contrast, program indicator expressions (as well as the new indicator subexpressions in 2.38.1) are translated into SQL and evaluated in the database engine against individual analytics values before they are aggregated. Unlike indicator values which can be converted from numeric to boolean on the fly depending on the value type, program indicator values must know ahead of time whether a boolean or numeric value is expected at a given point in the expression, so the SQL can be generated with the correct type. Fortunately this is knowable. Here are notes on how it is done:

- `ExpressionParams.dataType` is initially set to the dataType expected of the expression as a whole. `DefaultProgramIndicatorService` now specifies that the main expression of a program indicator should return `DataType.NUMERIC`, and the filter expression of a PI should return `DataType.BOOLEAN`.

- `FunctionIf` always expects its first argument to return a BOOLEAN value. So the `getSql` method calls `visitor.sqlBooleanVisit()` to force the expected datatype for that argument to boolean. The second and third arguments are expected to return the same dataType that is expected of the if() function as a whole, therefore when evaluating these arguments the dataType is not changed. Likewise, `D2Condition` sets the type to BOOLEAN when evaluating the conditional expression.

- Comparison operators `OperatorCompareEqual, ...GreaterThan, ...GreaterThanOrEqual, ...LessThan,  ...LessThanOrEqual`, and `...NotEqual` always return a boolean value, but the items they compare may be integer, boolean, or string. (string includes dates in string format 'yyyy-mm-dd'). When walking the parse tree for a comparison operator, it is not knowable ahead of time what data type the values will have to the left and right of the operator. But this is not necessary. The comparison operators request that the left and right sides be generated in SQL as NUMERIC if they are numeric or boolean. This allows two numeric values to be compared, two booleans to be compared (as numeric 1 and 0), or even a numeric value to be compared with a boolean (as 1 or 0). (String and date values will still be generated as SQL text and compared as such.)

- `CommonExpressionVisitor` adds two public methods for the convenience of the above conditional and comparison classes: sqlBooleanVisit, to generate SQL for a subtree returning BOOLEAN, and sqlNumericVisit, to generate SQL for a subtree returning NUMERIC. They both use the private method visitWithDataType which regenerates the visitor object with the specified type and uses it to visit the subtree.

- `ProgramExpressionItem` generates the correct SQL according to datatype: If the valueType of a Program DataElement or Attribute is boolean or numeric, then SQL is generated for either dataType BOOLEAN or NUMERIC depending on what is desired at this point in the expression, with a default value of 0 or false, respectively. SQL for other datatypes (including text and date) are processed as text (the SQL column is text already), with a default value of the empty string ''.

Unit tests have been modified and added as appropriate. Manual testing was also done.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
@@ -28,7 +28,6 @@
 package org.hisp.dhis.expression;
 
 import static org.hisp.dhis.expression.MissingValueStrategy.NEVER_SKIP;
-import static org.hisp.dhis.util.ObjectUtils.firstNonNull;
 
 import java.util.HashMap;
 import java.util.List;
@@ -188,6 +187,8 @@ public class ExpressionParams
 
     public DataType getDataType()
     {
-        return firstNonNull( dataType, parseType.getDataType() );
+        return (dataType != null)
+            ? dataType
+            : parseType.getDataType();
     }
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramIndicatorService.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramIndicatorService.java
@@ -31,6 +31,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import org.hisp.dhis.analytics.DataType;
+
 /**
  * @author Chau Thu Tran
  * @author Jim Grace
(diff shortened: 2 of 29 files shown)


maikelarabori (approved):
Nice work @jimgrace :+1:

enricocolasante (approved):

jbee (approved):

PR #10921, opened 2022-05-30
fix: Add CSV format for analytics data value set API [DHIS2-13351]

larshelge:
Adds CSV format for analytics data value set endpoint. JSON and XML formats are already supported.

```
/api/analytics/dataValueSet.csv
```

Added unit tests. Sample URL for manual testing:

```
/api/analytics/dataValueSet.csv
?dimension=dx:fbfJHSPpUQD;cYeuwXTCPkU;Jtf34kNZhzP
&dimension=pe:LAST_12_MONTHS
&dimension=ou:O6uvpzGd5pu;fdc6uOvgoji;lc3eMKXaEfw
```


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GridHeader.java
```diff
@@ -101,6 +102,17 @@ public GridHeader( String name, String column )
         this.column = column;
     }
 
+    /**
+     * @param name formal header name.
+     * @param valueType the header value type.
+     */
+    public GridHeader( String name, ValueType valueType )
+    {
+        this( name );
```
jbee, line 111: I think this is bad practice to call a less argument constructor from one with more. It should be the other way around so that the one with less basically just provides defaults for the parameters that are not passed.

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/ListGrid.java
```diff
@@ -309,7 +309,13 @@ public List<GridHeader> getMetadataHeaders()
     @Override
     public int getIndexOfHeader( String name )
     {
-        return headers.indexOf( new GridHeader( name, null ) );
+        return headers.indexOf( new GridHeader( name ) );
+    }
+
+    @Override
+    public boolean headerExists( String name )
```
jbee, line 316: This could have been a `default` method in the interface implemented like this to avoid having to reimplement this e.g. in testing mocks etcetera.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GridHeader.java
larshelge: I think I can agree. One challenge is if you want to have "sensible" defaults for certain properties and you want to avoid spreading those default values out to many constructors. You could use a variable of course. We have the more-calling-less pattern for all constructors already so I think I don't want to go into a refactor in this PR.

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/ListGrid.java
larshelge: Okay will update.

PR #10928, opened 2022-05-30
test: adds test for import with non default id schemes

vilkg:
DHIS2-12563: 
- adds a test for complete tracker data import ( tei, relationship, event, enrollment) with different id schemes; 
- removes the old id schemes tests for event import


dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
```diff
@@ -0,0 +1,242 @@
...
+        TestData data = new TestData();
+
+        JsonObject event = new EventDataBuilder()
+            .setProgram( new TestData( programIdScheme ).getEventProgram() )
+            .setProgramStage( new TestData( programStageIdScheme ).getEventProgramStage() )
+            .setOu( new TestData( ouIdScheme ).getOrgUnit() )
+            .setAttributeCategoryOptions( Arrays.asList( data.getCategoryOption() ) )
+            .array();
+
+        // act
```
teleivo, line 160: I don't think we need arrange, act, assert comments. You've grouped the code nicely, so its clear to me 😄

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
```diff
@@ -0,0 +1,242 @@
...
+        {
+        }
+
+        public TestData( String idScheme )
+        {
+            String propertyName;
+            if ( idScheme.toLowerCase().contains( "attribute" ) )
+            {
+                propertyName = "attributeValues.value[0]";
+            }
```
teleivo, line 203:
```suggestion
            } else {
```

the newline in between if and else looks odd

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
```diff
@@ -0,0 +1,242 @@
...
+        EnrollmentDataBuilder enrollmentDataBuilder = new EnrollmentDataBuilder()
+            .setOu( data.getOrgUnit() )
+            .setProgram( data.getTrackerProgram() )
+            .addEvent( eventDataBuilder );
+
+        JsonObject payload = new TeiDataBuilder()
+            .setId( teiId )
+            .setTeiType( data.getTrackedEntityType() )
+            .setOu( data.getOrgUnit() )
+            .addAttribute( data.trackedEntityAttribute, DataGenerator.randomString() )
```
teleivo, line 113: Is there any reason you are not using the getters sometimes? like in `data.trackedEntityAttribute`

If not, I would keep it consistent 😄

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
```diff
@@ -0,0 +1,242 @@
...
+        String programIdScheme = "CODE";
+        String ouIdScheme = "NAME";
+        String programStageIdScheme = "ATTRIBUTE:" + ATTRIBUTE_ID;
+        TestData data = new TestData();
+
+        JsonObject event = new EventDataBuilder()
+            .setProgram( new TestData( programIdScheme ).getEventProgram() )
+            .setProgramStage( new TestData( programStageIdScheme ).getEventProgramStage() )
+            .setOu( new TestData( ouIdScheme ).getOrgUnit() )
+            .setAttributeCategoryOptions( Arrays.asList( data.getCategoryOption() ) )
```
teleivo, line 157: I would inline the `data` variable if its only used once.

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
```diff
@@ -0,0 +1,242 @@
...
+    @Test
+    public void shouldImportEventsWithDifferentProgramSchemes()
+    {
+        String programIdScheme = "CODE";
+        String ouIdScheme = "NAME";
+        String programStageIdScheme = "ATTRIBUTE:" + ATTRIBUTE_ID;
+        TestData data = new TestData();
+
+        JsonObject event = new EventDataBuilder()
+            .setProgram( new TestData( programIdScheme ).getEventProgram() )
```
teleivo, line 154: I might be wrong 😋 In `TestData(IdScheme)` we are making ~ 10 GET requests. Since we call this constructor 3 times in this test and 1 time in each of the above 2 tests (each parameterized with 3 values). That's 10 GET requests * (3+2*3) ~ 90 GET requests just to get the data we have in `idSchemesMetadata.json`

Do we parse/get such identifiers from test fixtures in other tests or do we always go via the API?

teleivo (commented):
I have a question regarding the requests me make in TestData. Other than that it looks very well 😄

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
vilkg: You are not wrong, that's a good observation :) I can add a cache for responses if you want, or parse the values from the file, but all the tests take 10s here, so I don't think it's too bad.

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
vilkg: no reason, just clumsiness. Fixed now 👍

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/IdSchemeTests.java
teleivo: I would prefer reading them from the JSON file 😄 10s does not sound too bad, but this will add up over time 😬

teleivo (approved):
awesome, thank you very much @vilkg ! 🎉

enricocolasante (approved):

PR #10934, opened 2022-05-31
fix: FileResourceSupplier checking DataElement DHIS2-12563

teleivo:
which should also be in an idScheme aware manner, not assuming only UIDs

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/FileResourceSupplierTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/FileResourceSupplierTest.java
@@ -196,7 +196,6 @@ class FileResourceSupplierTest extends DhisConvenienceTest
 
     private Attribute numericAttribute()
     {
-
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

lucaCambi77 (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/FileResourceSupplierTest.java
```diff
@@ -53,148 +59,223 @@
...
+        {
+            dvs.add( dv );
+        }
+        return dvs;
+    }
+
+    private DataValue dataValue( String uid, String value )
+    {
+
+        return DataValue.builder()
```
lucaCambi77, line 276: also here remove empty lines

teleivo: either we are using a poor formatter or its not well configured if we still have to talk about formatting even when its used 😢 
removed them 👍🏻

PR #10935, opened 2022-05-31
fix: OrgUnitValueTypeSupplier checking DataElement DHIS2-12563

teleivo:
which should also be in an idScheme aware manner, not assuming only UIDs

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/OrgUnitValueTypeSupplierTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/OrgUnitValueTypeSupplierTest.java
@@ -1,0 +1,279 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/preheat/supplier/OrgUnitValueTypeSupplierTest.java
```diff
@@ -0,0 +1,279 @@
...
+    private TrackerImportParams.TrackerImportParamsBuilder params( TrackerIdSchemeParams idSchemes )
+    {
+        return TrackerImportParams.builder().idSchemes( idSchemes );
+    }
+
+    private Attribute numericAttribute()
+    {
+
+        return Attribute.builder()
+            .attribute( MetadataIdentifier.ofUid( "numeric" ) )
```
lucaCambi77, line 199: we should avoid line breaks evenly in code

teleivo: what do you mean by that? You can use the `Add a suggestion` feature to show how you would format the code 😄

lucaCambi77: in general no empty lines at the beginning and end of a method, we can see if we can add that in the formatter?

lucaCambi77 (approved):

PR #10958, opened 2022-06-01
fix: User defined names for time attributes in headers [DHIS2-13136]

maikelarabori:
This fix aims to apply the correct usage of the labels for each time attribute that has a custom name defined in enrollments or events.

The GET request below, for example, should not bring "Enrollment date" in the response header. It should be the user-defined value "Start of treatment date".

```
https://play.dhis2.org/dev/api/analytics/events/query/ur1Edk5Oe2n.html+css?dimension=ou:USER_ORGUNIT&outputType=EVENT&tableLayout=true&columns=ou;enrollmentDate&enrollmentDate=LAST_12_MONTHS&headers=ouname,enrollmentdate&dataIdScheme=NAME&paging=false&asc=ouname
```

For more details about the mapping rule, see DHIS2-13136.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CommonRequestMapper.java
```diff
@@ -51,11 +51,11 @@
 import org.hisp.dhis.common.QueryItem;
 import org.hisp.dhis.i18n.I18nManager;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
-import org.springframework.stereotype.Service;
+import org.springframework.stereotype.Component;
 
 import com.google.common.collect.ImmutableMap;
 
-@Service
+@Component
```
maikelarabori, line 58: Not related to this fix, just a minor improvement

gnespolino (approved):

d-bernat (approved):

PR #10963, opened 2022-06-02
fix: GeoJson import - always return status OK [DHIS2-7171]

jbee:
### Summary
GeoJSON import API now always returns HTTP status code 200 OK. 
This is to be consistent with other imports in DHIS2.

However, the import status is reflected in the `status` property of the response object.

### Automatic Testing
Existing tests were adjusted to the change of behaviour.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/GeoJsonImportController.java
```diff
@@ -198,7 +204,7 @@ public void after()
         public void call()
         {
             notifier.clear( config );
-            notifier.notify( config, NotificationLevel.INFO, "GeoJSON import stared", true );
+            notifier.notify( config, NotificationLevel.INFO, "GeoJSON import stared", false );
```
jbee, line 207: FYI: This is an unrelated small fix.

PR #10975, opened 2022-06-02
fix: Analytics api when requesting event data from two programs that share a data element [2.37-DHIS2-12466-backport]

d-bernat:
Please see: https://jira.dhis2.org/browse/DHIS2-12466


maikelarabori (approved):

gnespolino (approved):

larshelge:
Could we add a test which asserts the bug and the fix?

d-bernat:
> Could we add a test which asserts the bug and the fix?

Yep, good idea, I am going to create new ticket for it, this is just backport

PR #8438, opened 2021-07-26
feat: Implement field filtering using Jackson

mortenoh:
Part 1 of changes to switch out the the field filtering system to a new parser/filter based on Jackson only.

This PR includes

* New module `dhis-support-field-filtering`
* Removes old `dhis-support-json-filtering` module
* New `FieldFilterParser` for fully parsing out field filters + transformers.
* New Jackson only `FieldFilterManager` that handles filtering using Jackson nodes.
* Jackson `BeanPropertyFilter` for applying include/exclude patterns.

None of this is currently applied to any endpoint, so this PR will not functionally change anything in the system.

[TECH-650](https://jira.dhis2.org/browse/TECH-650)

PR #10979, opened 2022-06-02
fix: jclouds connection to aws-s3

jimgrace:


dhis-2/pom.xml
```diff
@@ -1653,15 +1656,26 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.apache.jclouds.provider</groupId>
```
larshelge, line 1660: Use two-space indentation for pom.xml.

jimgrace: @larshelge This PR is for 2.38 where there is a four-space indentation. I have a separate PR for 2.39 with two-space indentation at [pull/10978](https://github.com/dhis2/dhis2-core/pull/10978). I've submitted this as two independent PRs rather than trying to backport and fight the indentation change. :)

(Both PRs are running into the unrelated integration test troubles that are noted in Slack.)

jimgrace: Ah, it seems that dhis-2/pom.xml was changed from 2-space to 4-space [here](https://github.com/dhis2/dhis2-core/pull/8438/files#diff-4caf3d06d60c2fda63c091b5d0a1f29e447208cae834b134367e1342cf5ffb40) before 2.38 was released.

This was fixed back from 4-space to 2-space [here](https://github.com/dhis2/dhis2-core/pull/10789/files#diff-c7edcca96f2e943ecc9b19588780e38a30ed8b1f95c86796c51367b55b967968) for 2.39 but was not fixed in 2.38.

That's why they're currently different.

larshelge: Understand.

PR #10983, opened 2022-06-02
feat: added a hook for request pre-processing [DHIS2-13379]

gnespolino:
feat: validate TEI Param inside the mapper [DHIS2-13379]
feat: added a hook for request pre-processing [DHIS2-13379]

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CommonRequestPreProcessor.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CommonRequestPreProcessor.java
@@ -51,11 +51,12 @@ public class CommonRequestPreProcessor
         //
         // 2 - in event/enrollments we supported _OR_ separator to handle
         // enhanced conditions
-        // left both as TODO for now
+        // TODO: DHIS2-13383
         return commonQueryRequest;
     }
 
-    // TODO: we would really like to have all criteria/request/params to be
+    // TODO: DHIS2-13384 we would really like to have all
+    // criteria/request/params to be
     // immutable, but PagingCriteria is not
     // returning it for now, should be converted to use builders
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CommonRequestPreProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class CommonRequestPreProcessor
+{
+    private final SystemSettingManager systemSettingManager;
+
+    public CommonQueryRequest preProcessCommonRequest( CommonQueryRequest commonQueryRequest )
+    {
+        // here we should refactor/preprocess dimensions and filters for 2
```
larshelge, line 45: Could we clean this Javadoc up? Sounds like something we should fix before merging to master.

gnespolino: We're developing a completely new service for cross program line listing reporting.
We cannot do everything in one ticket and we decided to use master to develop rather than a feature branch to help qa test it soon.
Ideally a dedicated, deployable branch would have been much better, then merge to master when it looked good, but in the end it seemed to be overkill.
Long story short: this part will be implemented in a separate ticket, here is just a placeholder.

gnespolino:
All TODOs in the code were updated with ticket number

larshelge (approved):

maikelarabori (approved):

d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CommonRequestPreProcessor.java
larshelge: That's fine @gnespolino .

PR #10984, opened 2022-06-02
chore: avoid id parameter concat in deletion handlers [DHIS2-13359]

jbee:
Goal is to not use string concatenation to insert IDs into the SQL used by the deletion handlers.
While these are not user input and also numeric they cannot be exploited for an SQL attack but they still get flagged by static analysis tools.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DataElementDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DataElementDeletionHandler.java
@@ -27,11 +27,12 @@
  */
 package org.hisp.dhis.dataelement;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static org.hisp.dhis.category.CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME;
-import static org.hisp.dhis.system.deletion.DeletionVeto.ACCEPT;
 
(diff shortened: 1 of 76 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DataElementDeletionHandler.java
```diff
@@ -138,8 +124,7 @@ private void deleteLegendSet( LegendSet legendSet )
 
     private DeletionVeto allowDeleteOptionSet( OptionSet optionSet )
     {
-        String sql = "SELECT COUNT(*) FROM dataelement WHERE optionsetid = " + optionSet.getId();
-
-        return jdbcTemplate.queryForObject( sql, Integer.class ) == 0 ? ACCEPT : VETO;
+        String sql = "SELECT COUNT(*) FROM dataelement WHERE optionsetid = :id";
```
larshelge, line 127: Do we want to change to lower-case SQL while we are at it?

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DataElementDeletionHandler.java
jbee: :+1: I can make another PR for that

PR #10986, opened 2022-06-02
feat: Add referral parameter to program stage [DHIS2-13301]

zubaira:


dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/program/ProgramStageServiceTest.java
```diff
@@ -78,8 +80,13 @@ void testSaveProgramStage()
     {
         long idA = programStageService.saveProgramStage( stageA );
         long idB = programStageService.saveProgramStage( stageB );
+
         assertNotNull( programStageService.getProgramStage( idA ) );
         assertNotNull( programStageService.getProgramStage( idB ) );
+
+        ProgramStage persistedProgramStage = programStageService.getProgramStage( idA );
```
teleivo, line 87: Can you move these assertions up to so they are grouped with `assertNotNull( programStageService.getProgramStage( idA ) );`. Add a newline before `assertNotNull( programStageService.getProgramStage( idB ) );`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerErrorCode.java
```diff
@@ -141,6 +141,8 @@
     E1309( "Generated by program rule (`{0}`) - Unable to assign value to attribute `{1}`. " +
         "The provided value must be empty or match the calculated value `{2}`" ),
     E1310( "Generated by program rule (`{0}`) - Attribute `{1}` is being replaced in tei `{2}`" ),
+    E1311( "Event: `{0}`, is referral event and needs to have at least one complete relationship" ),
```
teleivo, line 144: Could you please write it as a full sentence including punctuation like

"Event: `{0}` is a referral event that needs to have at least one complete relationship."

Same for the other error message.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralCheckValidationHook.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ *
+ * Validation hook to make sure referral events contain Relationship and both
+ * sides of Relationship are provided
+ *
+ * @author Zubair Asghar
+ */
+
+@Component
+public class EventReferralCheckValidationHook extends AbstractTrackerDtoValidationHook
```
teleivo, line 53: Check and Validation are redundant. So please rename to `EventReferralValidationHook`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralCheckValidationHook.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public void validateReferralRelationshipLinkage( ValidationErrorReporter reporter, TrackerBundle bundle,
+        Event event, ProgramStage programStage )
+    {
+        if ( programStage.isReferral() )
+        {
+            List<Relationship> relationships = event.getRelationships();
+
+            reporter.addErrorIf( relationships::isEmpty, event, E1311,
+                "programStage" );
```
teleivo, line 73: Passing arg "programStage" will lead to error message

"Event: `programStage` is a referral event that needs to have at least one complete relationship."

is that what you want? Same for the error message below.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_16__Add_referral_parameter_to_program_stage.sql
```diff
@@ -0,0 +1,10 @@
+-- https://jira.dhis2.org/browse/DHIS2-13301
+-- Adding referral parameter to ProgramStage
+
+ALTER TABLE programstage ADD COLUMN IF NOT EXISTS referral BOOLEAN;
```
teleivo, line 4: I think Enrico recently added a migration like this. Does it look the same @enricocolasante ?

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/DhisConvenienceTest.java
```diff
@@ -1780,6 +1780,7 @@ public static ProgramStage createProgramStage( char uniqueCharacter, int minDays
         programStage.setDescription( "description" + uniqueCharacter );
         programStage.setMinDaysFromStart( minDays );
         programStage.setRepeatable( repeatable );
+        programStage.setReferral( false );
```
teleivo, line 1783: referral will be false by default unless you added so I this line is not needed

teleivo (changes requested):
I have some minor comments 😄 and please also address the linting issues from Sonar/GitHub

enricocolasante (changes requested):
Validator is not added to the config file and so it is never really used.
If we add a validator I would say that we must create a unit test for it

gnespolino (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralCheckValidationHook.java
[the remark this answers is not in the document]
zubaira: resolved

zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerErrorCode.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerErrorCode.java
```diff
@@ -141,8 +141,8 @@
     E1309( "Generated by program rule (`{0}`) - Unable to assign value to attribute `{1}`. " +
         "The provided value must be empty or match the calculated value `{2}`" ),
     E1310( "Generated by program rule (`{0}`) - Attribute `{1}` is being replaced in tei `{2}`" ),
-    E1311( "Event: `{0}`, is referral event and needs to have at least one complete relationship" ),
-    E1312( "Event: `{0}`, is referral event and needs to have both sides of a relationship" ),
+    E1311( "Event: `{0}`, is a referral event and needs to have at least one complete relationship." ),
```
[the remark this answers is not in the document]
zubaira: I think having one complete sentence with all the required information is more readable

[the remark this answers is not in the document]
teleivo: The majority of the errors in https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerErrorCode.java are not using the style you suggest. As the entity that is invalid is already provided in the error report. I would keep it consistent for the user and therefore remove the redundant `{0}`. You could propose your style in the design meeting 😄 If thats what we want, we should adapt all errors.

teleivo (changes requested):
Can you please fix the linting issues/CodeQL issue 😄

zubaira:
> 

resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHook.java
```diff
@@ -0,0 +1,81 @@
...
+    public void validateEvent( ValidationErrorReporter reporter, TrackerBundle bundle, Event event )
+    {
+        ProgramStage programStage = bundle.getPreheat().getProgramStage( event.getProgramStage() );
+
+        checkNotNull( programStage, TrackerImporterAssertErrors.PROGRAM_STAGE_CANT_BE_NULL );
+
+        validateReferralRelationshipLinkage( reporter, bundle, event, programStage );
+    }
+
+    public void validateReferralRelationshipLinkage( ValidationErrorReporter reporter, TrackerBundle bundle,
```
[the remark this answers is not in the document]
teleivo: @zubaira can you please fix this one and also check sonarqube which reports this issue and another one.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHook.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public void validateReferralRelationshipLinkage( ValidationErrorReporter reporter,
+        Event event, ProgramStage programStage )
+    {
+        if ( programStage.isReferral() )
+        {
+            List<Relationship> relationships = event.getRelationships();
+
+            reporter.addErrorIf( relationships::isEmpty, event, E1311,
+                event.getEvent() );
```
teleivo, line 73: now that the error message has no argument anymore you should also remove the arg event.getEvent() here https://github.com/dhis2/dhis2-core/pull/10986/files#diff-5bda4b7c88a18af614d4dfde1ba2055b19e4ee287b096db8c53e1491f3cb077aR73 and here https://github.com/dhis2/dhis2-core/pull/10986/files#diff-5bda4b7c88a18af614d4dfde1ba2055b19e4ee287b096db8c53e1491f3cb077aR77

teleivo (changes requested):
@enricocolasante correctly pointed out that EventReferralValidationHook is missing a unit test.

zubaira:
> @enricocolasante correctly pointed out that EventReferralValidationHook is missing a unit test.

resolved

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
```diff
@@ -0,0 +1,254 @@
...
+            .programStage( MetadataIdentifier.ofUid( PROGRAM_STAGE_REFERRAL_UID ) )
+            .orgUnit( MetadataIdentifier.ofUid( ORG_UNIT_UID ) )
+            .attributeOptionCombo( MetadataIdentifier.EMPTY_UID )
+            .enrollment( ENROLLMENT_UID )
+            .build();
+
+        subject.validateEvent( reporter, bundle, event );
+
+        assertEquals( 1, reporter.getReportList().size() );
+        assertTrue( reporter.hasErrorReport( r -> r.getErrorCode() == TrackerErrorCode.E1311 ) );
```
teleivo, line 206: when checking size and content of a list I would use https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertAll-org.junit.jupiter.api.function.Executable...-

like

```suggestion
        assertAll( () -> assertEquals( 1, reporter.getReportList().size() ),
                   () -> assertTrue( reporter.hasErrorReport( r -> r.getErrorCode() == TrackerErrorCode.E1311 ) )
        );
```

so that in case the list contains 2 errors we still check whether it at least contains the error we want.

This gives us better error output. This way we don't have to re-run tests/debug them to figure out whats wrong. Super helpful also in CI when we have flaky tests and or with integration tests that just take so long to run.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
```diff
@@ -0,0 +1,254 @@
...
+        Event event = Event.builder()
+            .event( CodeGenerator.generateUid() )
+            .program( MetadataIdentifier.ofUid( program ) )
+            .programStage( MetadataIdentifier.ofUid( PROGRAM_STAGE_REFERRAL_UID ) )
+            .orgUnit( MetadataIdentifier.ofUid( ORG_UNIT_UID ) )
+            .attributeOptionCombo( MetadataIdentifier.EMPTY_UID )
+            .enrollment( ENROLLMENT_UID )
+            .build();
+
+        subject.validateEvent( reporter, bundle, event );
```
teleivo, line 203: Just as an example regarding the overall review comment on tests.
I think the above `when()` with

```suggestion
        subject.validateEvent( reporter, bundle, new Event() );
```

and the below assertion is all that is needed to test the actual implementation with respect to this particular scenario.

teleivo (changes requested):
Thank you very much for addressing our comments 😄 

I think the tests have quite a lot of setup that is not needed. The implementation you test only needs

* `bundle.getPreheat().getProgramStage( event.getProgramStage() );` to be non-null and a referral or not.
* then an event with a relationship with both, none or only one side populated or no relationship at all

Our backend tests in general need a lot of setup, which when the complexity is needed is what we have to do. In this case the validator you wrote is super simple. So the tests should be simple as well. I would remove all setup that is not needed, even if the event would not be configured in the way a "real" event in an implementation would look like. The only thing we want to test is the unit of the validator.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
```diff
@@ -0,0 +1,254 @@
...
+        init();
+    }
+
+    @Test
+    public void verifyValidationSuccessForNonReferralEvent()
+    {
+        when( preheat.getProgramStage( MetadataIdentifier.ofUid( PROGRAM_STAGE_UID ) ) )
+            .thenReturn( programStage );
+
+        Event event = Event.builder()
```
enricocolasante, line 130: You have events that you need for the test that will only differ by programStage and relationship. You can a private method that will get this 2 parameter and create the event to test.
This way is good to clean up the test and remove the focus on building object details.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        subject.validateEvent( reporter, bundle, event );
+
+        assertEquals( 1, reporter.getReportList().size() );
+        assertTrue( reporter.hasErrorReport( r -> r.getErrorCode() == TrackerErrorCode.E1311 ) );
+    }
+
+    private void init()
+    {
+        orgUnit = createOrganisationUnit( "A" );
```
enricocolasante, line 211: Do you really need to initialise all these metadata for the test?
Unnecessary details make it difficult to understand what is being tested and what is important for the test.
I would suggest to remove any metadata that will not change the result of the test

enricocolasante (commented):
Minor comments on the unit test

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHook.java
```diff
@@ -0,0 +1,78 @@
...
+    public void validateEvent( ValidationErrorReporter reporter, TrackerBundle bundle, Event event )
+    {
+        ProgramStage programStage = bundle.getPreheat().getProgramStage( event.getProgramStage() );
+
+        checkNotNull( programStage, TrackerImporterAssertErrors.PROGRAM_STAGE_CANT_BE_NULL );
+
+        validateReferralRelationshipLinkage( reporter, event, programStage );
+    }
+
+    public void validateReferralRelationshipLinkage( ValidationErrorReporter reporter,
```
stian-sandvold, line 65: To simplify the logic, we don't need this validation; As discussed with Markus, the client will handle the logic; We just provide the metadata flags :)

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHookTest.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/EventReferralValidationHook.java
zubaira: resolved

larshelge (approved):

PR #10987, opened 2022-06-02
fix: UserRoleDeletionHandler DHIS2-13336

jimgrace:
@larshelge I quite agree with [your comment](https://github.com/dhis2/dhis2-core/pull/10900#issuecomment-1141995568) on the first PR for [DHIS2-13336](https://jira.dhis2.org/browse/DHIS2-13336). For consistency with the rest of the code base this is better done in the deletion handler. In fact the logic is already there in `UserRoleDeletionHandler`.

What apparently happened is that the `UserRoleDeletionHandler` logic was broken by another code refactor. Instead of fixing it, `UserObjectBundleHook#preDelete` was added to remove the user roles before deletion. However this logic was flawed because it tried to remove a Set member while iterating through the Set. My [first PR for the ticket](https://github.com/dhis2/dhis2-core/pull/10900) fixed this flaw.

But the original cause of the problem seems to be that `UserRoleDeletionHandler` had removed the user from the roles, but had not removed the roles from the user. Presumably because of the user refactor, this logic no longer worked and the attempt at deleting the user resulted in a foreign key violation between the user and the user role. (Did Hibernate re-persist the connections to roles that were still attached in Java to the User object???)

In this PR, I have cleared the user roles from the user (after removing the user from the roles). When deleting a user with one or more roles, `UserRoleDeletionHandler` now works again, as I have tested. I have also deleted the unneeded `UserObjectBundleHook#preDelete`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserRoleDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserRoleDeletionHandler.java
@@ -27,6 +27,8 @@
  */
 package org.hisp.dhis.user;
 
+import java.util.HashSet;
+
 import lombok.AllArgsConstructor;
 
 import org.hisp.dhis.system.deletion.DeletionHandler;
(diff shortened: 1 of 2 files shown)


larshelge (approved):

larshelge:
Nice spot @jimgrace . Great to get the root cause fixed here. Yes for bi-directional associations we need to remove from both sides.

PR #10989, opened 2022-06-03
test: use assertNoImportErrors for better context

teleivo:
in case an import fails we shouldn't be left wondering why and re-run CI over and over again.
Re-use our own `assertNoImportErrors`. Adjust `assertNoImportErrors` so its message more readable.

so a test like

```java
        TrackerImportReport trackerImportReport = trackerImportService.importTracker(fromJson( "tracker/validations/events-with-registration.json" ));

        assertNoImportErrors( trackerImportReport );
```

not expecting any import errors would fail with for example

```
org.opentest4j.AssertionFailedError: Expected import with status OK, instead got:
E1123: Missing required event property: `orgUnit`.
E1123: Missing required event property: `programStage`.
E1123: Missing required event property: `program`.
 ==> 
Expected :OK
Actual   :ERROR
```

instead of just

```
Expected :OK
Actual   :ERROR
```


larshelge (approved):

larshelge:
Good idea.

PR #11000, opened 2022-06-03
test: improve assertions of flaky DefaultMetadataVersionServiceTest

teleivo:
DefaultMetadataVersionServiceTest.testShouldReturnTheLatestVersion has been flaky on GitHub. 

Improve assertions we can figure out why. Also remove unused mockito annotation its a pure integration test that is not using any mocks.

Instead of assertion errors like `expected true got false` we now get

in case of one of the versions being null

```
org.opentest4j.AssertionFailedError: Either both versions or none must be null. v1: null, v2: MetadataVersion{importDate=null, type=BEST_EFFORT, name='Version_2', hashCode='abcdef'}
```

in case the versions do not match

```
name must match ==> expected: <Version_1> but was: <Version_2>
Comparison Failure: 
Expected :Version_1
Actual   :Version_2
<Click to see difference>



type must match ==> expected: <ATOMIC> but was: <BEST_EFFORT>
Comparison Failure: 
Expected :ATOMIC
Actual   :BEST_EFFORT
<Click to see difference>



org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
	org.opentest4j.AssertionFailedError: names must match ==> expected: <Version_1> but was: <Version_2>
	org.opentest4j.AssertionFailedError: types must match ==> expected: <ATOMIC> but was: <BEST_EFFORT>
```


enricocolasante (approved):

larshelge (approved):

larshelge:
Good initiative.

PR #11007, opened 2022-06-03
feat: Add referral field to RelationshipType [DHIS2-13300]

enricocolasante:


teleivo:
please rebase after https://github.com/dhis2/dhis2-core/pull/11010 is merged so we can test if jacoco collects coverage across module boundaries

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipType.java
```diff
@@ -59,6 +59,8 @@
 
     private String toFromName;
 
+    private boolean referral = false;
```
teleivo, line 62: I think this can be removed as false is the default anyway.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE relationshiptype ADD COLUMN IF NOT EXISTS referral BOOLEAN;
+
+UPDATE relationshiptype SET referral = FALSE WHERE referral IS NULL;
+
+ALTER TABLE relationshiptype ALTER COLUMN referral SET NOT NULL;
```
teleivo, line 5: would it make sense to add a default?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipType.java
enricocolasante: Fixed

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
enricocolasante: I had the same doubt when I was creating the script but then I saw that in all similar case we are doing in this way.
Maybe @stian-sandvold would know more and would know if there is any reason not to have default values on the DB.
Otherwise I am happy to change it

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipType.java
enricocolasante, line 62:
```suggestion
    private boolean referral;
```

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
stian-sandvold: As far as I know, there is no practical reason for not adding a default, other than the fact that we rarely do it. You can add it as a default instead if you want :)

stian-sandvold (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
enricocolasante: Changed to default

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
```diff
@@ -1,6 +1,2 @@
-ALTER TABLE relationshiptype ADD COLUMN IF NOT EXISTS referral BOOLEAN;
-
-UPDATE relationshiptype SET referral = FALSE WHERE referral IS NULL;
-
-ALTER TABLE relationshiptype ALTER COLUMN referral SET NOT NULL;
+ALTER TABLE relationshiptype ADD COLUMN IF NOT EXISTS referral BOOLEAN DEFAULT false;
```
teleivo, line 1: A reason for doing it the way it was before would be if there is a chance the column exists. It would then still make sure to update the null values to false and add the not null constraint.

So now it would not do the latter steps. If you also want to account for that fact, you would need to adapt it.

@ameenhere maybe you have some more thoughts on this as you've seen a few migrations that failed in the last months I think 😬

ameenhere: Shouldn't we still set the NOT NULL constraint on the column even though we set the default?

The column can otherwise still contain null values.

ameenhere: Although practically it may not be possible to update it to null through hibernate as booleans are primitives. But it does leave a small window open to either have null values via jdbctemplate or directly updating in the db.

Apart from that tiny window, the script is idempotent and looks good to me.

enricocolasante: Look at the suggested change that I put.
I think that to be really safe we need to make sure we remove null values (as @teleivo said, the column could already be present and allow null values, if we just add if not exist, it will be skipped and null values would still be allowed) and we need to set the not null to be sure it cannot be manually changed

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
```diff
@@ -0,0 +1,2 @@
...
```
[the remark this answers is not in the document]
ameenhere: I think you can have the SET DEFAULT (3rd script) inlined with your first script. It will still be idempotent and will not fail if the column already exists and even if it has null values.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_14__Add_referral_column_to_relationshiptype.sql
enricocolasante: It will not fail but it will not be applied if the column exists

ameenhere: Ah good point.

PR #11008, opened 2022-06-03
feat: Persist fileResource owner from file side

abyot:


stian-sandvold (approved):
@abyot looks good!

Do you have a plan for data values for aggregate, and file resources used outside of tracker? Should we potentially add a new issue for those?

abyot:
> @abyot looks good!
> 
> Do you have a plan for data values for aggregate, and file resources used outside of tracker? Should we potentially add a new issue for those?

may be deal that in a separate issue?

enricocolasante (approved):

PR #11009, opened 2022-06-03
fix: UserGroupDeletionHandler drop groups from user

jimgrace:
In `UserGroupDeletionHandler#deleteUser`, after removing user from groups, also remove groups from user.

This follows the (new) pattern in `UserRoleDeletionHandler#deleteUser`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserGroupDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserGroupDeletionHandler.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.user;
 
+import java.util.HashSet;
 import java.util.Set;
 
 import lombok.AllArgsConstructor;
@@ -62,6 +63,8 @@ public class UserGroupDeletionHandler extends DeletionHandler
             group.getMembers().remove( user );
             idObjectManager.updateNoAcl( group );
         }
+
+        user.setGroups( new HashSet<>() );
     }
 
     private void deleteUserGroup( UserGroup userGroup )


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserGroupDeletionHandler.java
```diff
@@ -62,6 +63,8 @@ private void deleteUser( User user )
             group.getMembers().remove( user );
             idObjectManager.updateNoAcl( group );
         }
+
+        user.setGroups( new HashSet<>() );
```
larshelge, line 67: I am just curious why this is necessary, as the association is managed from the user group side, and we don't do this in other places. See e.g. DataElementDeletionHandler. To be correct we should update the association from both sides in the application so not saying it is wrong.

Another issue is by setting groups to empty set we are relying on the Hibernate dirty-checking feature which will automatically update the user when some other save operation is done as part of the session, which is not ideal since we rely on ORM-provider specific behavior.

jimgrace: @larshelge Good point about sided-ness. I'm trying to compare and contrast this with `UserRoleDeletionHandler#deleteUser` where the role relationship is owned by user. That code, which follows the same pattern, worked for years until the user refactor seems to have broken it. Then by adding a single line to clear the roles from the user it was fixed. But I wasn't really sure why it was broken. Maybe hibernate was re-persisting the relationship since it wasn't deleted from user? You mentioned in https://github.com/dhis2/dhis2-core/pull/10987 that "for bi-directional associations we need to remove from both sides". My intention in submitting this PR was out of an abundance of caution, not because it appears to be needed. Though perhaps it could have unintended consequences and we should just let things stay as they are.

But this does make me more curious about why the UserRoleDeletionHandler fix was needed. I see in 2.37 UserCredentials.hbm.xml:

```
    <set name="userAuthorityGroups" table="userrolemembers">
      <cache usage="read-write" />
      <key column="userid" foreign-key="fk_userrolemembers_userid" />
      <many-to-many column="userroleid" class="org.hisp.dhis.user.UserAuthorityGroup"
        foreign-key="fk_userrolemembers_userroleid" />
    </set>
```
Whereas in the current User.hbm.xml:

```
    <set name="userRoles" table="userrolemembers" cascade="all">
      <cache usage="read-write"/>
      <key column="userid" foreign-key="fk_userrolemembers_userid"/>
      <many-to-many column="userroleid" class="org.hisp.dhis.user.UserRole"/>
    </set>
```

Notably, cascade="all" has been added. Whatever the reason for adding it, I wonder if this is what broke UserRoleDeletionHandler. Perhaps as hibernate was preparing to delete the user with cascade="all", it decided that it first needed to persist the role relationship (that still existed from the user side) to the roles?

I'm curious why we use cascade="all" on some relationships but not others. (It's not there for example on the UserGroup side of the UserGroup-User relationship.) But I suppose I really don't need to know. I'm fine with just closing this PR if you think it's better not to try and fix something that ain't broke. :)

larshelge: Good spot @jimgrace , I agree. I suspect the cascade=all is not necessary here @netroms , what do you think? Adding cascade=all can have the effect that objects are "re-saved" even when you try to remove them from associations.

larshelge: It appears removing cascade=all introduces a host of problems. Merging.

larshelge: Morten had a look and it seems removing cascade=all for user roles cause a lot of issues and test failures.

PR #11012, opened 2022-06-05
feat: Predictions by Cat Option Combo DHIS2-9832

jimgrace:
See [DHIS2-9832](https://jira.dhis2.org/browse/DHIS2-9832). This implements the ability for predictors to make independent predictions for every data element disaggregation (category option combo). So, for example, if an input data element is disaggregated by age and sex, and so is the output data element, one prediction will be made for females under 5, another prediction will be made for males under five, and so on.

No additional metadata is needed to support this. Disaggregated predictions will happen when the output data element has a CategoryCombination that includes disaggregations, and the output CategoryOptionCombo is null. When the predictor expression contains one or more input data elements (not data element operands with explicit catOptionCombos), and the input data element's category option combinations are compatible with those of the output data element, then the predictor expression will be evaluated for each output catOptionCombo using the corresponding catOptionCombo values from the compatible input data elements. Any input data elements whose disaggregations are not compatible with the output will just be evaluated in the usual way -- their same values will be supplied to each output disaggregation calculation.

A frontend change will be made to make this configuration explicit ([DHIS2-12203](https://jira.dhis2.org/browse/DHIS2-12203)). Currently if no output catOptionCombo is specified, the default catOptionCombo is filled in, so it is actually impossible currently in the frontend to configure a predictor with a null catOptionCombo.

Most of code for this feature is in two new classes `PredictionDisaggregator` and `CategoryOptionComboMapGenerator` and their respective unit test classes. Other classes have been modified to fit in with these changes.

### PredictionDisaggregator

This builds on the relatively recent refactoring of predictors that created the `PredictionContext` and `PredictionContextGenerator` classes. These two classes simplified the prediction engine, reducing the number of nested loops it contained. A `PredictionContext` contains the dimensions (within a single organisation unit) in which one single prediction is made, such as attributeOptionCombo, prediction output period, and the data on which to make the prediction. Added to this PredictionContext now is categoryOptionCombo, which allows the prediction engine in `DefaultPredictionService` to focus on a single disaggregation at a time, without having to add another level of logic to the prediction engine itself.

The `PredictionContextGenerator`, which generates the other dimensions of the prediction context, calls this new class `PredictionDisaggregator` to manage the additional logic for when `PredictionContexts` are (or are not) generated for each disaggregation.

### CategoryOptionComboMapGenerator

In large installations for which disaggregated predictions would be useful, disaggregations sometimes change from year to year. For example data elements used in one year may have age ranges (0-14, 15-30, >30), but in another year have age ranges (0-4, 5-10, 11-15, etc.) Already these installations often use category option groups so they can provide reporting across years for category options that can be mapped to each other. This implementation of disaggregated predictions also allows for such disaggregation changes through time in the following way: If the output data element contains disaggregations that can cover all the combinations of category options present in the input data elements, then they are considered compatible. For instance if different input data elements have the disaggregations in the example above, they can map to an output data element if it has catOptionCombos for all the date ranges (0-14, 15-30, >30, 0-4, 5-10, 11-15, etc.) Analytics based on the output data element can use category option groups to group options in much the same way that is currently done for multiple data elements.

Note that different CategoryOptionCombos are typically generated for different CategoryCombos, even for the same combination of category options. Because of this, the CategoryOptionComboMapGenerator does not rely on the CategoryOptionCombos being the same; it looks at the CategoryOptions within these CategoryOptionCombos. It creates a map for each input data element CategoryCombo that has compatible options with the output data element CategoryCombo. This map tells how to map each input CategoryOptionCombo to an output CategoryOptionCombo, based on the CategoryOptions contained within each CategoryOptionCombo.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
+++ b/dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.predictor;
+
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toMap;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.hisp.dhis.category.CategoryCombo;
+import org.hisp.dhis.category.CategoryOption;
+import org.hisp.dhis.category.CategoryOptionCombo;
+import org.hisp.dhis.common.MapMap;
+
+/**
+ * Generator of a map that, for each category combo (CC) found in an input data
+ * element, returns a map from the category option combinations of that CC to
+ * the category option combinations of the output category combo.
+ *
+ * @author Jim Grace
+ */
+public class CategoryOptionComboMapGenerator
+{
+    private CategoryOptionComboMapGenerator()
+    {
+    }
+
+    /**
+     * Returns a map that, for each input category combo (CC), returns a map
+     * from the category option combinations of that CC to the category option
+     * combinations of the output category combo.
+     * <p>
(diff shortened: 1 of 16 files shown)


dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
```diff
@@ -0,0 +1,140 @@
...
+ * element, returns a map from the category option combinations of that CC to
+ * the category option combinations of the output category combo.
+ *
+ * @author Jim Grace
+ */
+public class CategoryOptionComboMapGenerator
+{
+    private CategoryOptionComboMapGenerator()
+    {
+    }
```
jbee, line 53: Usually a private utility class throws a `UnsupportedOperationException("util")`.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
```diff
@@ -0,0 +1,140 @@
...
+     * combinations of the output category combo.
+     * <p>
+     * The map has entries only for those CCs where all the COCs can be mapped
+     * to output COCs.
+     *
+     * @param outputCombo predictor output data element category combination.
+     * @param inputCategoryCombos data element cat combos found in the input.
+     * @return the map.
+     */
+    public static MapMap<String, String, String> getCcMap( CategoryCombo outputCombo,
```
jbee, line 67: The `MapMap` approach has the drawback of being a poor way of communicating what the data represents or how it is organized. I think with lombok allowing us to define data structures with just a class with some fields and some class level annotations it often a better option to introduce a new "record" type. That way the semantics get encoded.

jbee (approved):
Looks good. Have a think about replacing `MapMap` with a dedicated record type to better communicate the role and structure of the data.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
jimgrace: @jbee Thanks, I hadn't caught onto that yet. Done!

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/CategoryOptionComboMapGenerator.java
jbee: It surely is easier to grasp now. Extending collection or map classes comes with its own issues but as long as this is used in a small scope those aren't that bad. Something that would be better than maps of maps is if this can be modelled as list of records. If this problem is not suited for that either extending a map or creating a record with a map field is fine. I personally would rather have a map field but here I think extension is fine too.

jimgrace: @jbee Understood. Ultimately a Map is needed so that lookups can happen quickly as data (potentially lots of it) is processed. The MapMap (or Map of Map) is nice because two different decisions are needed as the data is processed:
1. Should the COCs of a CategoryCombo (CC) be translated? Easy, see if the top map has that key.
2. To what output COC should an input COC (off a CC) be translated? Access both levels of map.

For the second use case, a single level of Map could be used with the key being perhaps a concatenation of the CC UID with the COC UID. Then for the first use case a Set might be used to contain the set of CCs that have map entries (combined with COCs) in the Map.

It's perhaps a tradeoff being complexity and more code vs. being more transparent. A judgement call to be sure. More code in the sake of greater transparency can also increase the time to understand and modify the code correctly which is, in the end, what we are after.

If you think the code now is reasonable enough, please re-approve the PR. (I had re-requested a review from you.) I need to move on to other things between now and the Annual Conference. :)

I appreciate the time and care you take in reviewing my PRs. I believe I am improving my craft as a result!

jbee: I missed that you re-requested but you had my approval ;)

PR #11050, opened 2022-06-08
fix: Validate enrollment attributes values based on valueType in NTI [DHIS2-12669] (2.39)

ameenhere:
In NTI, program attributes can be specified as part of the Enrollment tracker Dto. Validation based on valueType for enrollment attributes was missing in the NTI. This PR fixes it by moving the `validateAttributeValue` method from `TrackedEntityAttributeValidationHook` class into its base abstract class `AttributeValidationHook`, so that the second subclass `EnrollmentAttributeValidationHook` can reuse the same method.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/AttributeValidationHook.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/AttributeValidationHook.java
@@ -28,13 +28,20 @@
 package org.hisp.dhis.tracker.validation.hooks;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.hisp.dhis.system.util.ValidationUtils.dataValueIsValid;
+import static org.hisp.dhis.tracker.report.TrackerErrorCode.E1077;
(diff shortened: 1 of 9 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/AttributeValidationHook.java
```diff
@@ -103,6 +115,31 @@ else if ( valueType.equals( ValueType.USERNAME ) )
...
+    {
+        checkNotNull( tea, TRACKED_ENTITY_ATTRIBUTE_VALUE_CANT_BE_NULL );
+        checkNotNull( value, TRACKED_ENTITY_ATTRIBUTE_VALUE_CANT_BE_NULL );
+
+        // Validate value (string) don't exceed the max length
+        reporter.addErrorIf( () -> value.length() > Constant.MAX_ATTR_VALUE_LENGTH, trackerDto,
+            E1077, value,
+            Constant.MAX_ATTR_VALUE_LENGTH );
+
+        // Validate if that encryption is configured properly if someone sets
```
stian-sandvold, line 130: Did we do this before? I'm not sure this is the correct place to complain about the encryption configuration issue

stian-sandvold (approved):

PR #11052, opened 2022-06-08
feat: DimensionParam to wrap DimensionalObject and QueryItems [DHIS2-13399]

gnespolino:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/analytics/common/dimension/DimensionParam.java
```diff
@@ -0,0 +1,109 @@
...
+@Data
+@Builder
+@AllArgsConstructor( access = AccessLevel.PRIVATE )
+public class DimensionParam
+{
+    private final DimensionalObject dimensionalObject;
+
+    private final QueryItem queryItem;
+
+    private final DimensionParamType dimensionParamType;
```
maikelarabori, line 56: Maybe we could rename only the attribute in this case, to `type`.

Otherwise we will end-up with `dimensionParam.getDimensionParamType()`. It seems redundant.
The form `dimensionParam.geType()` sounds a bit nicer.

maikelarabori (approved):
Looks good. I left one minor comment only.
Apart from that, I would just add Javadocs for all classes and public methods, and the most complex private ones. Also some unit tests (unless we are planning to add them later when things are in better shape).

Cheers,
Maikel

d-bernat (approved):

PR #11100, opened 2022-06-15
feat: MULTI_TEXT value type [DHIS2-429]

jbee:
### Summary
Adds new `ValueType` `MULTI_TEXT`.

This type **must** be used with an `OptionSet` which provides the valid values.
The values are stored as a `,` separated text value.

Main changes affect the validation of data elements and data values.
New checks make sure:
* data elements with value type `MULTI_TEXT` must have an option set
* value type of data element and its option set must be identical (for any value types)

Data values of type `MULTI_TEXT` are only accepted when 
* the data element and option set are valid (see above)
* the data value only contains valid, `,` separated options from the option set

### Automatic Testing
New test were added for all of the new validations.

### Manual Testing
* create an option set with value type `MULTI_TEXT`
* create some options for the option set
* create a data element of type `MULTI_TEXT` using the created option set
* create data values for the data element using valid and invalid option both in the single data value API and the data value set import

Failure scenarios:
* try to create or update a data element of type  `MULTI_TEXT` without an option set
* try to create or update a data element of type  `MULTI_TEXT` with an option set of another type
* try to create or update a data value for a  `MULTI_TEXT` data element using non existing options


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DefaultDataElementService.java
```diff
@@ -80,18 +70,39 @@ public DefaultDataElementService( DataElementStore dataElementStore,
...
+        {
+            if ( dataElement.getValueType() == ValueType.MULTI_TEXT )
+            {
+                throw new IllegalQueryException( new ErrorMessage( ErrorCode.E1116, dataElement.getUid() ) );
+            }
+            return;
+        }
+        // need to reload the options in case this is a create and the options
+        // is a shallow reference object
+        OptionSet options = optionSetStore.getByUid( dataElement.getOptionSet().getUid() );
```
vietnguyen, line 99: Seems the comment above doesn't match with the code here. Maybe we don't need to reload OptionSet.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -75,7 +75,13 @@
     URL( String.class, false ),
     FILE_RESOURCE( String.class, true, FileTypeValueOptions.class ),
     IMAGE( String.class, false, FileTypeValueOptions.class ),
-    GEOJSON( GeoJSON.class, false );
+    GEOJSON( GeoJSON.class, false ),
+    MULTI_TEXT( String.class, true );
+
+    public static List<String> splitMultiText( String value )
+    {
```
larshelge, line 82: Move separate to constant, to make it clear which character is the separator.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -75,7 +75,13 @@
...
+        return value == null ? List.of() : List.of( value.split( "," ) );
```
larshelge, line 83: Question: We use `,` as separator which is a fairly commonly used character. Do we want to use `|` or some less used character as separator?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DefaultDataElementService.java
jbee: In what why doesn't it match? the `OptionSet` is reloaded directly after the comment as it states. It was needed because otherwise the set might be a reference what does not hold the value type information. This occurs when a new data element with option set reference is created as the comment says. I assume this also would occur when an existing data element gets assigned a option set.

vietnguyen: ok then its all good. 
Was a bit confused with the variable name `OptionSet options = ...`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -75,7 +75,13 @@
...
+    }
```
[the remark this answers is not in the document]
jbee: I did not think that escaping would be useful as a code with a collision of the separator character is unlikely to be possible to change to add the escaping.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -213,10 +219,7 @@ public boolean isAggregatable( AggregationType aggregationType )
         {
```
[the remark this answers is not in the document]
jbee: There is no general validation on codes yet. If I remember correctly @mortenoh wanted codes to be "extended alphanumeric" in the future which also would solve this issue.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
jbee: The `,` was chosen as it should still be a unusual character to use as part of a code but it is very readable when displayed as string for a data value. I think we should assume that we will not be able to adjust the UI in every view to deal with multi-text data values and display them nicely. So whatever separator we chose should render "human readable". Comma was the obvious choice and in regards to usage as actual character of a code is no worse or better than alternatives like `;`, `:` or `+`.

jbee:
I added additional validations as discussed on Tuesday.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dataelement/DataElementServiceTest.java
```diff
@@ -98,6 +142,44 @@ void testUpdateDataElement()
...
+    void testUpdateDataElement_WithOptionSetHavingCodeWithSeparator()
+    {
+        DataElement de = createDataElement( 'A' );
+        dataElementService.addDataElement( de );
+        de.setValueType( ValueType.MULTI_TEXT );
+        de.setOptionSet( addOptionSet( ValueType.MULTI_TEXT, "A", "B", "C,D" ) );
+
+        IllegalQueryException ex = assertThrows( IllegalQueryException.class,
+            () -> dataElementService.updateDataElement( de ) );
+        assertEquals( ErrorCode.E1117, ex.getErrorCode() );
```
larshelge, line 177: Very minor, we have `org.hisp.dhis.utils.Assertions.assertThrowsErrorCode` which could make these error code assertions a bit more compact.

larshelge (approved):

larshelge:
Nice work :+1:

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dataelement/DataElementServiceTest.java
jbee: nice, will use it where I don't check the message as well

PR #11104, opened 2022-06-16
fix: Performance: relationships endpoint is slow when there are many relationships [ DHIS2-11935 ]

lucaCambi77:
[DHIS2-11935](https://jira.dhis2.org/browse/DHIS2-11935)

Second step to improve performance for Get relationships endpoint.

- Remove a circular dependency with relationships while getting an event.
- It also add a basic criteria for pagination. Front end is not using pagination for relationships so we'll need to integrate proper logic later for both (fe - be)


teleivo (commented):
The issue you refer to

DHIS2-12039 Slow PSQL queries when retrieving /events

does not mention relationships. I see some changes relate to events in your PR, but your PR describes mainly adapting the relationships endpoint. That's confusing to me 😅 Can you please explain? Did you link to the wrong issue?

vilkg:
performance looks good! For the same case and the same number of relationships, memory allocation went from 2g to 100mb ( with `skipPaging=true` query param, since Luca introduced paging). I think with paging we should also add `pager` object in the response to avoid confusion.

enricocolasante (approved):

gnespolino (approved):

PR #11105, opened 2022-06-16
fix: flaky test because of created date

vietnguyen:
### Issue
 - This `testShouldReturnTheLatestVersion` fails occasionally  with below error 
 - I couldn't reproduce it, but i guess the reason is because the query from `versionService.getCurrentVersion()` try to get the latest version by `created` date. But from the `assertEqualMetadataVersions`, we can see that both version 1 & 2 have same `created` date. 

```
[ERROR] Failures:
[ERROR]   DefaultMetadataVersionServiceTest.testShouldReturnTheLatestVersion:139->assertEqualMetadataVersions:301 Mult
res (2 failures)
        org.opentest4j.AssertionFailedError: name must match ==> expected: <Version_2> but was: <Version_1>
        org.opentest4j.AssertionFailedError: type must match ==> expected: <BEST_EFFORT> but was: <ATOMIC>
```

### Fix
- I set `created` date explicitly when create new `Version` objects in order to avoid `new Date()` return same value when code are run parallel.

### Test
- Need to observer for a few days to check if the error occur again on master before create backports for the fix.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/version/DefaultMetadataVersionServiceTest.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/version/DefaultMetadataVersionServiceTest.java
@@ -74,6 +74,10 @@ class DefaultMetadataVersionServiceTest extends TransactionalIntegrationTest
 
     private MetadataVersion versionB;
 
+    private Date startDate;
+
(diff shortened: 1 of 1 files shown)


teleivo (approved):
great analysis @vietnguyen 🎉 thank you very much for looking into this!

teleivo:
Just FYI PR is based on an older commit, so codecov report should look more accurate after a rebase. There is no need to rebase the PR though!

larshelge (approved):

PR #11155, opened 2022-06-22
fix: add more details for referencesCheck errorReport

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-7864
### Issue
- The ReferenceCheck doesn't provide `displayName` and `errorProperty` to the `ErrorReport` like in other validation checks of the import service.
### Fix
- Add necessary properties to the `ErrorReport` and add unit test.


dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/ReferencesCheckTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Indicator indicator = new Indicator();
+        indicator.setUid( "indicatorUid" );
+        indicator.setName( "indicatorA" );
+        IndicatorType indicatorType = new IndicatorType();
+        indicator.setName( "indicatorTypeA" );
+        indicator.setIndicatorType( indicatorType );
+
+        ReferencesCheck referencesCheck = new ReferencesCheck();
+        TypeReport report = referencesCheck.check( bundle, Indicator.class, Collections.emptyList(),
+            Lists.newArrayList( indicator ), ImportStrategy.CREATE, validationContext );
```
jbee, line 101: You can now use `List.of` instead of `Lists.newArrayList` as long as you know for sure that the elements are all non-`null`.

jbee (approved):

PR #11177, opened 2022-06-24
chore: Handle H2 test for schema and setting modules [TECH-1208]

enricocolasante:
After some discussion with the backend team we agreed that tests that need to access the DB should use postgres if there is no strong reason to use H2. (This doesn't apply to `dhis-web-api-test` module as H2 is used to prevent mocking the whole persistence layer).

In this PR the tests that access the DB are transformed to integration tests and moved to `dhis-test-integration`.
The ones that are not accessing the DB are transformed into unit tests


teleivo (approved):
@enricocolasante you can dismiss the codeql alerts on the tests. CodeQL check is currently failing.

ameenhere (approved):

PR #11193, opened 2022-06-27
test: Analytics e2e tests structure

maikelarabori:
This PR aims to introduce e2e tests for Analytics APIs.

**Testing data**
The tests will run on top of a specific Sierra-Leone database, and this will become the baseline for all new tests down the road.
This is required, so we keep a baseline regarding the data (they will not change, ensuring the same results).
As part of this initial work, we choose to go with Sierra-Leone 2.39.0.

With this approach, we have plenty of scenarios for testing out-of-the-box. We won't need to craft custom JSON files as in all e2e tests. Also, if some test breaks, we can reproduce it in Play or locally. It will make it much easier to identify the issue and debug for problems.

**Test execution**
By default, all analytics e2e tests are disabled on the PRs. If one wishes to execute them, the label `run-api-analytics-tests` has to be added to the respective PR. This is because of the extra time required to run the analytics e2e tests. We need to download the Sierra-Leone DB, unzip it and import it into the DB in the current container. After that execute the analytics tests, which are likely to be heavier than others given the nature of execution, responses and the number of assertions required.

**Main technical changes**
In `docker-compose.yml` I defined a new variable responsible to hold the volume mounting point: `${DB_MOUNT:-.:/app}`.
It gives us the option to set or not that variable. If not set, we default to the usual mounting point `.:/app`. **_This is useful to ensure that the analytics e2e tests will only mount and populate the Sierra Leone DB when the variable `DB_MOUNT` is set. Currently, it occurs only when the label `run-api-analytics-tests` is applied_**.

A new git workflow file was created `run-api-analytics-tests.yml`. This file contains a new step named `Download Sierra Leone DB` where the Sierra Leone DB is downloaded, unzipped and copied to a temp folder. There is also caching in place in order to avoid downloading the same DB file every time. This is achieved through the action `actions/cache@v3`.

**Basic assumptions for all analytics e2e tests**
All analytics e2e tests assume that there is a Sierra-Leone DB correctly populated, based on the expected assertions.
Also, there is an additional general check that asserts that each analytics test should not take longer than 2 seconds to execute. This could help catch major performance issues, as an execution over 2 seconds would mean a serious performance degradation. This number can be adjusted or even removed later (if we conclude that this is not really useful).


teleivo (commented):
Could you please write in the PR description what it is that you are trying to achieve?

* when should analytics e2e tests run? this might differ when thinking about a devs workflow and when tests run in CI
* ...

This way we can help out with the setup 😄 I would try to avoid the duplication I currently see (another workflow, another docker compose file). For example using Maven surefire to steer what tests (tags/groups) are executed by default, on CI, ...
Like we already do in for example https://github.com/dhis2/dhis2-core/blob/a32e9ddf4c15521d5a41c0df72e1241a68b6a486/dhis-2/pom.xml#L356


Docker compose also supports environment variables with defaults.

So there is lots of flexibility available and depending on your goals we can decide on which route to take 😄

maikelarabori:
> Could you please write in the PR description what it is that you are trying to achieve?
> 
>     * when should analytics e2e tests run? this might differ when thinking about a devs workflow and when tests run in CI
> 
>     * ...
> 
> 
> This way we can help out with the setup 😄 I would try to avoid the duplication I currently see (another workflow, another docker compose file). For example using Maven surefire to steer what tests (tags/groups) are executed by default, on CI, ... Like we already do in for example
> 
> https://github.com/dhis2/dhis2-core/blob/a32e9ddf4c15521d5a41c0df72e1241a68b6a486/dhis-2/pom.xml#L356
> 
> Docker compose also supports environment variables with defaults.
> 
> So there is lots of flexibility available and depending on your goals we can decide on which route to take 😄

Thanks @teleivo for your comments. There are no details or descriptions yet because this is just a PoC. I'm not sure if this is the right path to follow and what would be the best approach.

This is just an initial implementation so we can start a discussion. I have a meeting today exactly to discuss it.
Once we have the definitive approach and more clarity about the final expectations I will add a good description for the WIP PR 😄

I will probably reach you in the next few days to get some help.

Cheers,
Maikel

PR #11200, opened 2022-06-29
feat: Event/Enroll Ownership Analytics DHIS2-11597

jimgrace:
See [DHIS2-11597](https://jira.dhis2.org/browse/DHIS2-11597) including the comment from 29/Jun/22.

This PR extends the concept of orgUnitField to include other options for selecting the organisation unit to be returned by event (and enrollment) analytics queries. Until now, orgUnitField was an optional query string holding the UID of a program attribute or data element with `ValueType.ORGANISATION_UNIT`. When present, this orgUnit would be used for analytics reporting rather than the event or enrollment orgUnit (depending on whether it was an event or enrollment query).

The orgUnitField query string may now also have one of the following string values:

`REGISTRATION` - Registration orgUnit
`ENROLLMENT` - Enrollment orgUnit
`OWNER_AT_START` - Owning orgUnit at start of period (or report)
`OWNER_AT_END` - Owning orgUnit at end of period (or report)

For reports that are aggregated by period (as in Event Visualizer), the owner orgUnit is determined at the start or end of each reporting period. For reports not aggregated by period (as in Line Listing), the owner orgUnit is determined at the start or end of the entire report.

Following are selective comments on new or modified classes:

### OrgUnitField and OrgUnitFieldType

These classes hold the expanded semantics of orgUnitField. Now the optional orgUnitField String is converted to an `OrgUnitField` object with type `OrgUnitFieldType`: `DEFAULT` (no orgUnitField specified), `ATTRIBUTE` (UID of attribute or data element), `REGISTRATION`, `ENROLLMENT`, `OWNER_AT_START`, or `OWNER_AT_END`.

These two classes determine the table alias / column combination to be used for the organisation unit when querying the analytics event or enrollment tables.

### JdbcOwnershipAnalyticsTableManager

This implements a new type of analytics table: `analytics_ownership_{programUid}`. This table is needed because the owner orgUnit is determined by the query period and cannot be known beforehand for a given event or enrollment.

The ownership analytics table records the ownership history of all TEI/Program combinations that change ownership orgUnits. Each record shows a period of ownership history for a TEI/Program combination with a start and end date. If a program/TEI only ever has one owner orgUnit, it will not be recorded in the table to save space and time because the enrollment orgUnit will be the owner orgUnit. Otherwise, this table can be queried to show the ownership on any date.

The ownership analytics table loads the values found in the `programownershiphistory` table when entries are present (they are not always), and also from the `programinstance` table to show enrollment ownership through time when `programownershiphistory` rows might not be present. The ownership enddate from `programownershiphistory` is ignored in case it is not correct. Ownership is extended until the next `programownershiphistory` or the next `programinstance` enrollment date, whichever is next.

There will be no gaps and no overlaps in the ownership analytics tables regardless of the information found in the `programownershiphistory` and `programinstance` tables. The end date for each ownership interval is set to the day before the next ownership interval starts. The first ownership interval always starts on 1000-01-01 (to cover any event dates before enrollment) and the last ownership interval ends on 9999-12-31 (to cover any future event dates).

If there is more than one enrollment and/or ownership change recorded on a given date, only the last such record is considered. Most precisely, the table shows the ownership at the end of any given day (the last ownership before midnight), and is used in analytics with that in mind. The ownership at the start of the next day is taken to be the same as the ownership at the end of the day before, before any changes made on the new day including any changes made at 12:00 am.

The current code is the third revision, to improve performance:

1. The ownership analytics tables were populated through an INSERT from a SELECT query, as are the other analytics tables, and then the start/end dates were adjusted as needed with UPDATEs (which included joins). With SL Demo on my machine this took 5 hours, presumably because there were a lot of updates with joins to a table with no indexes yet.

2. A SELECT query was used for reading into memory and the ownership analytics values were written using a `BatchHandler` (see `MappingBatchHandler` notes below). By querying the values in the right order, only two rows were needed in memory at any time before writing them out because I just needed to see if each row is the first for that TEI/Program (to set the start date way in the past and the end date just before the start of the next row), and to see if it is the last row for that TEI/Program (to set the end date way in the future). With SL Demo on my machine this took about 2 minutes.

3. I realized that for TEI/Program combinations where the ownership never changed, I didn't need to write out any rows to the ownership table at all because the enrollment orgUnit could be used for events at any time. With SL Demo on my machine this reduced the time to 5 seconds (because no rows were written out; there were no ownership changes).

Note: The ownership tables are only useful with tracker programs. However, they are created for both tracker and single event programs. With single event programs they are always empty. This was done to avoid requiring different table join logic if the owning orgUnit is requested for a single event program. (This would add further complexity to the logic that is already needed to fetch columns and join tables differently depending on whether the query is for events or enrollments.) For single event programs the event orgUnit will return the event reporting orgUnit as requested in [DHIS2-11597](https://jira.dhis2.org/browse/DHIS2-11597).

### JdbcOwnershipWriter

Implements the logic described in the previous section for writing the ownership analytics table rows while only needing two rows in memory at a time.

### MappingBatchHandler

So far in DHIS2 there have only been `BatchHandlers` that implement tables with a fixed number of columns as defined by fixed Java objects. This is not possible with the new ownership analytics tables because they contain columns that are dynamically defined at runtime. The `MappingBatchHandler` gets around this limitation by implementing a table for the object `Map<String, Object>`. This is general-purpose and can be used for any table schema. We might even consider adding it to the `quick` repository.

### JdbcEventAnalyticsTableManager

Two new columns were added to `analytics_event_{programUid}` tables:

`registrationou` - the TEI registration orgUnit for the event
`enrollmentou` - the TEI enrollment orgUnit for the event

### JdbcEnrollmentAnalyticsTableManager

One new column was added to `analytics_enrollment_{programUid}` tables:

`registrationou` - the TEI registration orgUnit for the event

### JdbcEventAnalyticsManager and JdbcEnrollmentAnalyticsManager

The queries were modified to fetch events and enrollments, respectively, with the help of `OrgUnitField`, `OrgUnitFieldType`, and `OrgUnitTableJoiner` (see below)

### OrgUnitTableJoiner

Joins the various orgUnit tables when/as needed for analytics event and enrollment queries:

- When using the default orgUnit from the analytics tables (event ou for events or enrollment ou for enrollments), no extra orgUnit tables are joined.

- When any other orgUnit is used, the `_orgunitstructure` is joined against that ou so that ou levels can be correctly queried.

- When any other orgUnit is used and orgUnitGroupSets are selected, the `_organisationunitgroupsetstructure` table is also joined to return the right groupSet membership. Note that in 2.38 the base analytics groupset columns were queried which gave the wrong results when orgUnitField was used. This has been fixed.

- When the ownership analytics table is being joined and data is aggregated to periods, the `_periodstructure` table is joined first on the period found in the `periodtype` column, so the ownership table can be joined for a date interval that includes the start (or end) of the period.

- When the ownership analytics table is being joined and data just has start/end dates (like for a line listing), the ownership table is joined for a date interval including the `startDate` (or `endDate`) of the report. The `_periodstructure` table is not needed.

### DateUtils

Added a `minusOneDay` method. At first I patterned this after `plusOneDay` which used `Date#toInstant`. However, I ran into trouble with dates returned from the database query because they were `java.sql.Date` objects. `java.sql.Date` is a subclass of `java.util.Date` but throws an exception when `toInstant` is called. This, according to the Java library code comment, is _"because SQL Date values do not have a time component"_. However `java.sql.Date` will return `Date#getTime`. So I coded both `plusOneDay` and `minusOneDay` to use `getTime` instead, which resulted in smaller and faster code as well as being compatible with `java.sql.Date`.

### EventAnalyticsServiceTest

I refactored this to use `SingleSetupIntegrationTestBase`. Originally it had 3 tests, one of which did nothing. Now it has 20. Instead of the one the test that did nothing there are now 18 tests that verify the correct return of all six `OrgUnitField` types while getting (1) an aggregated grid, (2) a list of events (`EventService`), and (3) a list of Enrolments (`EnrolmentService`).

The test that did nothing was `testGridAggregation`. It expected a value of 6.0 when calling `AnalyticsTestUtils.assertResultGrid`. However `assertResultGrid` only checked to see if any row was returned it would have a value of 6.0. No rows were returned so the test succeeded. Any test value besides 6.0 would also have worked! There was a test setup method `parseEventData` that read csv test event data into memory objects but never persisted them. So no events were entered into the system and the analytics tables did not contain any! I ripped out `parseEventData` and the csv file, and created test data using code.

The other two existing tests, `testDimensionRestrictionSuccessfully` and `testDimensionRestrictionWhenUserCannotReadCategoryOptions`, appear to have worked but needed refactoring to accommodate the single setup. Before the refactor, the `testDimensionRestrictionSuccessfully` method had successfully passed the dimension restriction check but then returned nothing because the analytics tables weren't built. (The analytics tables were only built for the test that did nothing.) In my first attempt at running the test with real analytics tables built, the test successfully completed the dimension restriction check but then threw an exception because the analytics query included the dimension restrictions but the analytics tables didn't have those columns. This behaviour is related to [DHIS2-12388](https://jira.dhis2.org/browse/DHIS2-12388) when the program has a category combination (like default) that does not use the dimension restriction categories. To avoid the problem, I assigned the test program to a category combination containing these dimensions. The categories and combinations had to be generated before the analytics tables were built, so I had to created them in the test setup.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/OrgUnitField.java
```diff
@@ -0,0 +1,266 @@
...
+                this.type = OrgUnitFieldType.ENROLLMENT;
+            }
+            else if ( "OWNERATSTAR".equals( field ) )
+            {
+                this.type = OrgUnitFieldType.OWNER_AT_START;
+            }
+            else if ( "OWNERATENDD".equals( field ) )
+            {
+                this.type = OrgUnitFieldType.OWNER_AT_END;
+            }
```
jbee, line 113: Maybe better to add an alias field to the type and do a lookup of exact name first, if that fails lookup by alias or something similar? Or why not name the enums so that there is no need for this alias?

jbee (approved):
This is a big change that is hard to follow if you do not know the modified code. It looks to me that you given it extra effort to make it more readable and modular but I still have a hard time seeing how things fit which I hope is mostly because I know little about this code or what the overall process or semantics are. Therefore I think it is important the 2nd review is done by someone who does know this better.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/OrgUnitField.java
jimgrace: Good spot. This was actually code used for testing that I forgot to remove. The real names for these options are matched by the Enum names and can be parsed with `Enum.getIfPresent()`. Since the front end doesn't yet generate these real names, but is useful for displaying the output of the analytics queries, I had created dummy data elements in my database with 11-character UIDs REGISTRATIO, ENROLLMENTT, OWNERATSTAR, and OWNERATENDD that could be selected by the frontend. I've now removed this code. Thanks!

jimgrace:
@jbee Thanks, and agreed. The other reviewers have all worked in this part of the code. Because of the nature of the changes, I hope that @larshelge can find the time to look at this even if others do also.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/programindicator/DefaultProgramIndicatorSubqueryBuilder.java
```diff
@@ -121,8 +122,9 @@ private String getAggregateClauseForPIandRelationshipType( ProgramIndicator prog
 
     private String getFrom( ProgramIndicator pi )
     {
-        return " FROM " + ANALYTICS_TABLE_NAME + "_" + pi.getAnalyticsType().getValue() + "_"
-            + pi.getProgram().getUid().toLowerCase() + " as " + SUBQUERY_TABLE_ALIAS;
+        String baseTableName = ANALYTICS_TABLE_NAME + "_" + pi.getAnalyticsType().getValue();
+
+        return " FROM " + PartitionUtils.getTableName( baseTableName, pi.getProgram() ) + " as " + SUBQUERY_TABLE_ALIAS;
```
maikelarabori, line 127: Tiny recommendation: change "FROM" to "from"

maikelarabori (approved):
Looks good @jimgrace, nice implementation :+1:

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/programindicator/DefaultProgramIndicatorSubqueryBuilder.java
jimgrace: Thanks @maikelarabori. I agree that I would like "from" better. However, "FROM" matches all the other uppercase SQL keywords in this class so I think we should change them all if we change any of them. To avoid further extending the scope of this PR, I suggest we change none of them for the moment.

d-bernat (approved):

gnespolino (approved):
Looks good!

PR #11202, opened 2022-06-29
fix: exlucde not-readable property from audit record (2.38.1)

vietnguyen:
backport https://github.com/dhis2/dhis2-core/pull/11197


Philip-Larsen-Donnelly (approved):
Approving as a backport

PR #11205, opened 2022-06-29
fix: Refactor Event SQL to use named parameters - Simple bindings [ DHIS2-13306 ]

lucaCambi77:
https://jira.dhis2.org/browse/DHIS2-13306

Add sql parameter map for simple bindings e.g. = (equals).  
More complex bindings will be checked in https://jira.dhis2.org/browse/DHIS2-13455 also to break the work of using named parameters in at least 2 issues


gnespolino (approved):
looks good, but I see only changes in test code, is this correct ?

enricocolasante (approved):

PR #11218, opened 2022-07-04
fix: Improve data element deletion performance - sections [DHIS2-5761]

jbee:
### Summary
This PR improves the deletion of data elements, specifically the "unlinking" for `Section`s.
Instead of iterating all sections, only those affected by the deletion are loaded and updated.

This is one of multiple improvments to reduce time it takes to delete data elements.

### Automatic Tests
Added new test to make sure the SQL for the new method selection the affected methods works as intended.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/hibernate/HibernateSectionStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/hibernate/HibernateSectionStore.java
@@ -73,7 +73,8 @@ public class HibernateSectionStore
             + " left join sectiongreyedfields sgf on s.sectionid = sgf.sectionid"
             + " left join dataelementoperand deo on sgf.dataelementoperandid = deo.dataelementoperandid"
             + ", dataelement de"
-            + " where de.uid = :Id and (sde.dataelementid = de.dataelementid or deo.dataelementid = de.dataelementid);";
-        return getSession().createNativeQuery( hql, Section.class ).setParameter( "Id", dataElementUid ).list();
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/hibernate/HibernateSectionStore.java
```diff
@@ -63,4 +64,16 @@ public Section getSectionByName( String name, DataSet dataSet )
...
+    @Override
+    public List<Section> getSectionsByDataElement( String dataElementUid )
+    {
+        String hql = "select * from section s"
+            + " left join sectiondataelements sde on s.sectionid = sde.sectionid"
+            + " left join sectiongreyedfields sgf on s.sectionid = sgf.sectionid"
+            + " left join dataelementoperand deo on sgf.dataelementoperandid = deo.dataelementoperandid"
+            + ", dataelement de"
+            + " where de.uid = :Id and (sde.dataelementid = de.dataelementid or deo.dataelementid = de.dataelementid);";
+        return getSession().createNativeQuery( hql, Section.class ).setParameter( "Id", dataElementUid ).list();
```
larshelge, line 77: Perhaps use `:dataElementId` as parameter just to make it explicit what it refers to and prepare for additional parameters later.

mortenoh (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/hibernate/HibernateSectionStore.java
jbee: done

PR #11222, opened 2022-07-04
feat: general dimension parsing [DHIS2-13467]

gnespolino:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/analytics/common/dimension/DimensionParam.java
```diff
@@ -49,9 +49,11 @@
  */
 @Data
 @Builder( access = PRIVATE )
-@AllArgsConstructor( access = PRIVATE )
+@RequiredArgsConstructor( access = PRIVATE )
 public class DimensionParam
```
maikelarabori, line 53: Would be nice to have some Javadoc here.
It could explain why to we need this class and how it fits in the flow.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/DimensionSpecificationService.java
```diff
@@ -0,0 +1,195 @@
...
+            if ( uidWithOffsets.size() == 1 )
+            {
+                return DimensionSpecificationUids.of( null, null, uidWithOffsets.get( 0 ) );
+            }
+
+            throw new IllegalArgumentException( "Malformed parameter: " + fullDimensionId );
+
+        }
+
+        private static List<UidWithOffset> parseFullDimensionId( String fullDimensionId )
```
maikelarabori, line 163: I would add a Javadoc here explaining how many elements can result from the split of `fullDimensionId`, and examples of valid `fullDimensionId` that this method will accept.

maikelarabori (approved):
Hi @gnespolino, looks good in general.
I'm approving with some comments regarding Javadoc/documentation. Please take a look.
As we discussed the unit tests and small fixes if applicable will come as part of other PR.

Thanks,
Maikel

gnespolino:
Close, will create a new one

PR #11252, opened 2022-07-07
fix: Restore debezium-connector-postgres TECH-1259

jimgrace:
See [TECH-1259](https://dhis2.atlassian.net/browse/TECH-1259).

debezium-connector-postgres was removed from pom.xml in [pull/10436](https://github.com/dhis2/dhis2-core/pull/10436). However, it is needed for PostgreSQL cache invalidation in a DHIS2 cluster.

This PR adds it back in along with a test that will break if it is removed again.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-cache-invalidation/pom.xml
+++ b/dhis-2/dhis-support/dhis-support-cache-invalidation/pom.xml
@@ -43,6 +43,10 @@
       <groupId>io.debezium</groupId>
       <artifactId>debezium-embedded</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-connector-postgres</artifactId>
+    </dependency>
 
     <dependency>
       <groupId>org.springframework</groupId>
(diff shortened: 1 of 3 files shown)


jbee (approved):

dhis-2/dhis-support/dhis-support-cache-invalidation/pom.xml
```diff
@@ -43,6 +43,11 @@
       <groupId>io.debezium</groupId>
       <artifactId>debezium-embedded</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-connector-postgres</artifactId>
+      <scope>test</scope>
```
enricocolasante, line 49: If the scope is test the library will not be available when running the application.
Does it make sense?

If it needs to be available at runtime but it is not directly used, we need to add it as an exception in `maven-dependency-plugin` as a `ignoredNonTestScopedDependency`

jimgrace: @enricocolasante Thanks, good catch. I fixed the build error by adding scope test but it broke the running. I've now done as you suggest and it runs.

enricocolasante (approved):

jbee (approved):

PR #11271, opened 2022-07-11
fix: Performance optimisation when fetching events from Event Programs [ DHIS2-13443 ] 

lucaCambi77:
https://jira.dhis2.org/browse/DHIS2-13443


lucaCambi77:
@stian-sandvold @vilkg @Bekkalizer this fix looks good to me, let me know what you think so we can do more testing if required and backport it

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -2044,7 +2059,7 @@ private String getOrgUnitSql( SqlHelper hlp, EventSearchParams params, List<Orga
 
         if ( params.getOrgUnit() != null && !params.isPathOrganisationUnitMode() )
         {
-            orgUnitSql.append( " ou.organisationunitid = " )
+            orgUnitSql.append( " psi.organisationunitid = " )
```
ameenhere, line 2062: Immediate Reaction:  I think this change should also be conditional. Since the joining condition either uses the ownership org unit or the psi org unit. For Tracker programs (WITH_REGISTRATION), this clause shall remain as `ou.organisationunitid`. For Event Programs (WITHOUT_REGISTRATION), since ownership is completely out of scope it is okay to change to `psi.organisationunitid`. 


Reaction after thinking for a bit: Does this specific line change really introduce a performance benefit? Even though HISP Bangladesh states that this change in query seems to improve performance, it might be worth confirming with actual stress test numbers.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -2093,7 +2108,7 @@ else if ( params.isOrganisationUnitMode( OrganisationUnitSelectionMode.CHILDREN
                 {
                     orgUnitSql.insert( 0, " (" );
                     orgUnitSql.append( orHlp.or() )
-                        .append( " (ou.organisationunitid = " )
+                        .append( " (psi.organisationunitid = " )
```
ameenhere, line 2111: Same as previous comment.

ameenhere (changes requested):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
stian-sandvold: On lines 1070 - 1087, we already "link" ou.organisationunitid to psi or po (Depending on condition). This line should in theory not actually make a difference, the set should already be the same :)

Worst case, without a condition for WITH/WITOUT registration, this will return the wrong result (Since po.organisationunitid != psi.organisationunitid; We would filter on the wrong thing).

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
stian-sandvold: Agree, same as above :)

stian-sandvold (changes requested):

stian-sandvold:
In the case of this optimization, we can do a few more things:

Assuming we are working with "WITHOUT_REGISTRATION", we can:
1. remove the `LEFT JOIN trackedentityprogramowner po ON ( pi.trackedentityinstanceid = po.trackedentityinstanceid )
`
2. Rewrite `INNER JOIN organisationunit ou ON ( COALESCE ( po.organisationunitid, psi.organisationunitid ) = ou.organisationunitid )`. with item 1, we no longer have po, so we can rewrite the COALESCE.

stian-sandvold (approved):

ameenhere (approved):

PR #11273, opened 2022-07-11
fix: (NEEDS LARS REVIEW, DO NOT MERGE) Table partitioning in Analytics [DHIS2-13429]

maikelarabori:
This PR aims to restore the expected Analytics export and partitioning (as release 2.36).

**Overview**
Since release 2.37, some behaviours were not correctly covered when we introduced partition tables.

Pepfar identified and raised those issues in three JIRA tickets: DHIS2-13429, DHIS2-DHIS2-13478 and DHIS2-13479. In summary, those tickets describe the following problems.

**Problems**
1) The table `analytics_0` never gets cleaned. It causes data duplication when the application hits the query below, in `JdbcAnalyticsManager.getFromSourceClause`.
```
 select ap.*
      from analytics_0 as ap
      union all
      select ap.*
      from analytics_2019 as ap
      union all
      select ap.*
      from analytics_2020 as ap
```

2) After running the Analytics job (export) for the `last_years=x` all partitions are removed and replaced by the partitions related to `last_years` only. It causes old generated data to be lost and obliges the user to run the full export to recover all data, which takes much longer.

3) The method `JdbcAnalyticsManager.getFromSourceClause` always assumes we have a `0` partition (depending on the input params). If the `0` partition is not there for any reason, it causes a SQL syntax error. In this case, the partition `0` for analytics should always be generated, so the query never fails with a syntax error.

4) When the Analytics job is triggered using `last_years=0`, all years/partitions are removed. This obliges the user to run a full export so they can recover all Analytics tables.

**Resolution**
The changes present on this PR should fix those issues and mimic the same behaviour before version 2.37, per the items above.

**Impediments**
It's important to mention that the table `0` cannot be added as a partition of the main table. This happens because the table `0` can store rows for any specific year, making it impossible to map this partition with a single year. Even if we could, year `0` would not make much sense. For this reason, it's kept as a separate table.

****Pending - NEEDS TO BE FIXED BEFORE MERGING IT**
As mentioned in "_Impediments_", the table `0` for `events`, `enrollments` and others, cannot be attached as a partition (as explained above). So the data stored on these tables are never queried. Before, they could be queried because they were part of the `inheritance` approach. We need to discuss it and find a good solution for this case. We could always do a `union` with the table `0` for ALL queries, but this is a very bad idea in my opinion. The `inheritance` approach seemed more adapted for our use case, maybe we need to rethink it a bit more.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/partition/JdbcPartitionManager.java
```diff
@@ -65,8 +65,12 @@ public Set<String> getAnalyticsPartitions( AnalyticsTableType tableType )
             return analyticsPartitions.get( tableType );
         }
 
+        // TODO: Temporary fix to no break the application after analytics
```
maikelarabori, line 68: It's needed so we can test this branch correctly.
This issue was introduced recently, and this `workaround` should be removed as soon as the issue is correctly fixed.

gnespolino (approved):

d-bernat (approved):

larshelge:
Closing after discussion with Maikel.

PR #11288, opened 2022-07-12
fix: Audit created only for changed data value [DHIS2-13451]

zubaira:


teleivo:
@zubaira sonarqube points to an unused variable you can remove 😄

gnespolino (approved):

teleivo:
@zubaira can you please rebase. tests on master are passing again 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/bundle/TrackedEntityDataValueAuditTest.java
```diff
@@ -120,6 +120,7 @@ void testTrackedEntityDataValueAuditCreate()
             assertEquals( a.getProgramStageInstance().getUid(), psi.getUid() );
         } );
 
+        assertEquals( 1, updatedAudit.size() );
```
teleivo, line 123: We are regularly having issues with tests failing on master only on CI with little to no context.

We should absolutely avoid assertions like this for this reason. Imagine this test fails on GitHub only telling you `expected 1 got 2`. What are you going to do? 😅 Multiple devs lost a few days on multiple occasions trying to debug tests with such assertions 😬 

Please use for example
https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertAll(java.lang.String,java.util.Collection)

like in https://github.com/dhis2/dhis2-core/blob/0d367e9b804ae5476694ab9545840e21b4cd68e5/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/trigramsummary/TrigramSummaryControllerTest.java#L243-L245

to group the assertion for checking the size and the content below into one assertAll

This would allow us to at least know that if the `updatedAudit` contains 3 elements if the one we expect is in there.

Also try failing the test locally. Would you know what do do with the error message if it were to fail on GitHub only?

Adding a supplier to JUnit assertions like
https://github.com/dhis2/dhis2-core/blob/6e83eba877030575456f919de1bce55fcae8e517/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java#L44-L59

helps a lot with adding context, in case the assertion itself does not do that for us.

teleivo (changes requested):
Please rethink your assertion. I added a lot of info to the corresponding code because I had to deal with such issues to often already in DHIS2 😢

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/bundle/TrackedEntityDataValueAuditTest.java
zubaira: resolved.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/bundle/TrackedEntityDataValueAuditTest.java
```diff
@@ -107,30 +114,24 @@ void testTrackedEntityDataValueAuditCreate()
...
+        assertAuditCollection( updatedAudit, AuditType.UPDATE, ORIGINAL_VALUE );
+        assertAuditCollection( deletedAudit, AuditType.DELETE, UPDATED_VALUE );
+
+    }
+
+    private void assertAuditCollection( List<TrackedEntityDataValueAudit> audits, AuditType auditType,
+        String expectedValue )
+    {
+        assertAll( () -> assertFalse( audits.isEmpty() ),
+            () -> assertEquals( auditType, audits.get( 0 ).getAuditType(),
```
teleivo, line 129: You might have misunderstood what I meant 😅

So assertions in integration tests should give us the context we need to potentially fix them with only that information and nothing else for example when the test only fails on Github.

You have duplicated the information we already get from `assertEquals` in a message supplier to it.

In your case, imagine `audits` has 2 or more elements and what you expect `expectedValue`/`auditType` is in the list, just not at position `0`. Unfortunately, this is happening regularly to us 😓 In such an error case with assertions structured like this you will only see an assertion message telling you that the `auditType` of element 0 is not what we expect. You will not know that maybe the element you are looking for is in the `audits` but just at position 2.

Try setting up a unit test with a list of audits and the assertion `assertAuditCollection` you wrote. Then add more than just one element, with the expected one in any position other than 0. This way you can iterate quickly and come up with assertions that give a useful error message. https://github.com/dhis2/dhis2-core/pull/11326 is another example of how adjusted some assertions using `assertAll`.

PR #11289, opened 2022-07-13
fix: delete option linked with OptionGroup [DHIS2-7908-39]

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-7908

- Add deletionHandler to remove `Option` from `OptionGroup` when `deleteOption()` is triggered.
- Note that `Option` is automatically removed from `OptionSet` because of cascade="all" defined in `OptionSet.hbm`. So we don't need to add deletionHandler for this association.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/option/OptionGroupDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/option/OptionGroupDeletionHandler.java
@@ -29,19 +29,38 @@ package org.hisp.dhis.option;
 
 import java.util.List;
 
-import org.hisp.dhis.common.DataDimensionType;
-import org.hisp.dhis.common.GenericDimensionalObjectStore;
+import lombok.AllArgsConstructor;
 
-/**
- * @author Viet Nguyen <viet@dhis2.org>
- */
+import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/option/OptionGroupDeletionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    protected void register()
+    {
+        whenDeleting( Option.class, this::deleteOption );
+    }
+
+    private void deleteOption( Option option )
+    {
+        List<OptionGroup> optionGroup = optionGroupStore.getOptionGroups( option.getUid() );
```
larshelge, line 54: I think ideally we could use a query which fetches only the option groups which contains the given option, as opposed to retrieving all option groups and iterating in memory. 

Typically there will not be a high number of option groups, but as we know deletion can be slow, so as a good principle we should try to pull only the necessary objects from the database vs all.

vietnguyen: The method `optionGroupStore.getOptionGroups( option.getUid() )` queries only `optionGroup` which contains given option ID. I changed the method name to `getOptionGroupsByOptionId()` to make it less confusing.

mortenoh (approved):

stian-sandvold (approved):

PR #11290, opened 2022-07-13
fix DatastoreEntryStoreTest

vietnguyen:


gnespolino (approved):

enricocolasante:
> This `DatastoreEntryStoreTest` has been converted to `SingleSetupIntegrationTestBase`, so the test data created in each test should be unique. This PR attempt to provide unique name for each namespace in the test. An alternative way to solve similar issue is to implement `tearDownTest()`.

`SingleSetupIntegrationTestBase` is only assuring that the setup is run once. In this case there is not setup, so it shouldn't make a difference.
Then every single test is running inside a transaction that when the test ends is rollbacked, so nothing that is persisted in thee test should affect other tests

jbee:
The cause is in another test as fixed by https://github.com/dhis2/dhis2-core/pull/11293

PR #11292, opened 2022-07-13
feat: custom dates and or conditions [DHIS2-13383]

gnespolino:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CustomDateHelper.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Arrays;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class CustomDateHelper
```
maikelarabori, line 44: Would be nice adding a Javadoc for the class and add some unit tests for this helper.

maikelarabori (approved):
Looks good @gnespolino, just added one minor comment.

Cheers,
Maikel

d-bernat (approved):

PR #11296, opened 2022-07-14
chore: improvements around data element deletion [DHIS2-5761]

jbee:
### Summary
Some more improvements related to deletion of data elements to move work towards the database and avoid loading unnecessary data into memory.

### Automatic Testing
Existing tests

### Manual Testing
Simple scenario
* create new data element
* delete newly created data element => success

Scenario with predictor
* create new data element
* create a new predictor and set the newly created data element as output of the predictor
* try to delete newly created data element => not allowed

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/deletion/JdbcDeletionHandler.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/deletion/JdbcDeletionHandler.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.system.deletion;
 
+import java.util.List;
 import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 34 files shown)


mortenoh (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/deletion/JdbcDeletionHandler.java
```diff
@@ -63,4 +64,14 @@ protected final int delete( String sql, Map<String, Object> parameters )
     {
         return npTemplate.update( sql, parameters );
     }
+
+    protected final String firstMatch( String sql, Map<String, Object> parameters )
+    {
+        if ( !sql.toLowerCase().contains( "limit 1" ) )
```
larshelge, line 70: A minor improvement here would be to do a regex check for `limit [0-9]+` as in theory the SQL could contain a limit clause for another limit (e.g. 10), and if appending another limit clause, it would crash.

larshelge (approved):

PR #11300, opened 2022-07-18
fix: Enable some disabled tests [TECH-1254]

enricocolasante:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/EventImportTest.java
```diff
@@ -166,12 +166,6 @@ class EventImportTest extends TransactionalIntegrationTest
 
     private static final SimpleDateFormat simpleDateFormat = new SimpleDateFormat( DateUtils.ISO8601_NO_TZ_PATTERN );
 
-    @Override
-    public boolean emptyDatabaseAfterTest()
```
teleivo, line 170: Is this unused? Or not needed anymore?
Cleaning up after tests seems like something we want 😬

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/sync/TrackerSynchronizationTest.java
```diff
@@ -31,180 +31,76 @@
...
     }
 
     @Override
     public void setUpTest()
     {
         userService = _userService;
-        currentSession = sessionFactory.getCurrentSession();
         User user = createUserWithAuth( "userUID0001" );
-        currentSession.save( user );
-        subject = new JacksonTrackedEntityInstanceService( teiService, trackedEntityAttributeService,
```
teleivo, line 202: is the subject never used?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/audit/AuditRepositoryTest.java
```diff
@@ -360,30 +359,10 @@ void testCompressDecompress()
...
-            String uid = CodeGenerator.generateUid();
-            String code = CodeGenerator.generateUid();
-            Audit audit = Audit.builder().auditType( AuditType.CREATE ).auditScope( AuditScope.AGGREGATE )
-                .createdAt( LocalDateTime.of( 2000, 1, 1, 0, 0 ) ).createdBy( "test-user" )
-                .klass( DataElement.class.getName() ).uid( uid ).code( code ).data( "{}" ).build();
-            audits.add( audit );
-        } );
-        Timer timer = new Timer().start();
-        audits.forEach( audit -> auditRepository.save( audit ) );
-        System.err.println( "Single Insert: " + timer.getTimeInS() + "s" );
```
teleivo, line 377: Looks like these 2 tests are testing different things (just briefly looking at the assertion/println). Why did you decide to remove one?

teleivo (commented):
added some questions 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/audit/AuditRepositoryTest.java
enricocolasante: I removed the single insert 200K because it is already covered and it is more a performance test, it takes 4 minutes in my machine.
I enabled the batch one and decreased the number of insert in order to have an acceptable run time and still testing the save of multiple audits that was not covered

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/sync/TrackerSynchronizationTest.java
enricocolasante: It is used but we can autowire it instead of manually created and manually inject all the dependencies

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/EventImportTest.java
enricocolasante: Sure, this is why empty the database is the default behaviour.
I am not even sure we should allow to override it.

jbee (approved):

stian-sandvold (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/EventImportTest.java
teleivo: I don't think we should allow overriding it 🤔

PR #11307, opened 2022-07-19
fix: abort deletion on exception in handlers [DHIS2-13499]

jbee:
### Summary
Current behaviour is:

* if a deletion veto handler throws an exception (that is not the `DeleteNotAllowedException`) during veto phase the deletion handling is aborted effectively skipping other veto handler and certainly skipping all deletion handler (that change/remove linkage)
* if a deletion handler throws an exception (any) deletion handling is aborted and all further deletion handlers are skipped
Since no exception is re-thrown the code triggering the deletion event will continue and delete the object triggering the deletion event (unless this fails due to a FK constraint in the DB).

This can leave the DB in an inconsistent state that is mostly unpredictable and certainly undesirable. 

The target behaviour is:

* if an exception is thrown (not `DeleteNotAllowedException`) either during veto or deletion loop the manager will throw a `DeleteNotAllowedException` to effectively prevent the deletion to keep the database in an consistent state. 

This is certainly not ideal either as any issue in one of the deletion handler will prevent deletion from happening. This might cause situations where deletion is not possible to perform because the reasons of the exception are due to an programming error or a database state that cannot be repaired using the UI.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
@@ -228,6 +228,9 @@ public enum ErrorCode
     E4052( "For program rule variable with name `{0}` following keywords are forbidden : and , or , not" ),
     E4053( "Program stage `{0}` must reference a program." ),
 
+    /* Metadata Validation (continued) */
+    E4060( "Object could not be deleted: {0}" ),
+
     /* SQL views */
     E4300( "SQL query is null" ),
     E4301( "SQL query must be a select query" ),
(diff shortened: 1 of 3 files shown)


jbee:
FYI: The failing tests are unrelated, see https://dhis2.slack.com/archives/G4513PWGY/p1658238316990209

mortenoh (approved):

larshelge (approved):

larshelge:
Sounds good @jbee . I think it is safest to abort deletion if any exception happens.

There is one integration test failure which may or may not be related, good to check out:

```
Error:  org.hisp.dhis.tracker.programrule.ProgramRuleIntegrationTest.testImportEventInProgramStageSuccessWithWarningRaised  Time elapsed: 0.19 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <E1308> but was: <E1300>
```

jbee:
That might indeed be related, I check when I am back from vacation.

larshelge:
Seems like the failures are unrelated, will go ahead and merge.

PR #11308, opened 2022-07-19
ci: use jib to build DHIS2 Docker image [TECH-1279]

teleivo:
## TL;DR

* build Docker image using [Jib](https://github.com/GoogleContainerTools/jib) specifically its [Maven plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#jib---containerize-your-maven-project)
* run e2e tests using the Jib image

what we gain

* no bespoke build code needed to build Docker image using best practices
* faster DHIS2 startup due to pre-exploded war. On the instance manager DHIS2 2.38.1 startup is 8.168s faster on average (or 9.189s using max startup times).
* [multi-platform support](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#platform-object) by adjusting the Jib configuration in the `pom.xml`
* reduced network traffic (pulling/pushing from/to Dockerhub) when maven dependencies stay the same between subsequent builds due to the [dependencies image layer](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-are-jib-applications-layered)

what we loose

* build time: **locally** with warm caches = 3.80 ± 0.25 times slower build; 16.742s slower on average. Note that the Jib build took 23.7s max locally, which I would still consider a fast Docker build. A sample run on GitHub on this PR took 01:09 min.
* as Jib does not provide a GitHub action and [caches its image layers differently](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#system-properties) (without Docker) we cannot use the [Docker action cache support](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api). We would need to cache using the [GitHub action cache](https://github.com/actions/cache/blob/main/examples.md#examples) explicitly. The Jenkins build might also need to be changed to leverage Jib caching.

## DHIS2 startup time

I deployed [dhis2/core:2.38.1.1-tomcat-9.0-jdk11-openjdk-slim](https://hub.docker.com/layers/core/dhis2/core/2.38.1.1-tomcat-9.0-jdk11-openjdk-slim/images/sha256-d678ca68476b90a0ee360c4b75bdc6979daabf3e997d0e89d6ca741105d18c6a?context=explore) and [core-dev/jib-2.39-SNAPSHOT](https://hub.docker.com/layers/core-dev/dhis2/core-dev/jib-2.39-SNAPSHOT/images/sha256-17215ab797ad94f7c80441b1f005dc2a1c28172df9afc20aced81301a0b2df01?context=explore) 5 times using the instance manager. Then got the startup time from the tomcat log like `org.apache.catalina.startup.Catalina.start Server startup in [61209] milliseconds`. DB was [2.38.1](https://databases.dhis2.org/sierra-leone/2.38.1/dhis2-db-sierra-leone.sql.gz) and no migrations had to be run.

| | dhis2/core:2.38.1.1-tomcat-9.0-jdk11-openjdk-slim | core-dev/jib-2.39-SNAPSHOT |
| ----------- | ----------- | ----------- |
| max (seconds) | 63.166 | 53.977 |
| avg (seconds) | 61.3758 | 53.2078 |
| stdev (seconds) | 1.105682233 | 0.5268924938 |

So at least in the current configuration (core version, DB, instance manager setup (nodes, ...)) we save 8.168s on average or 9.189s using the max startup times.

### Method

The `Build core image` step in our GitHub workflow is building `dhis.war` and the Docker image. The existing Docker image build inside this step does not output any timings. Comparing a few runs of the current build and the Jib build is obviously far from scientific and prone to bias.

I therefore run benchmarks locally using https://github.com/sharkdp/hyperfine which you can rerun for yourself 🧑‍🔬 

I built images locally with [Docker BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/).

Build `dhis.war`

```sh
mvn clean install -f dhis-2/pom.xml -pl -dhis-web-embedded-jetty
mvn clean install -f dhis-2/dhis-web/pom.xml
```

Then run hyperfine

#### Using warm caches

```sh
hyperfine --export-markdown docker-build-benchmark.md \
  --warmup 3 \
  'mvn --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild' \
  'dhis-2/build-dev.sh'
```

#### Using cold caches

I am unsure if a scenario with cold caches will apply and if so how often. I added it for completeness.

Added `--no-cache` https://docs.docker.com/engine/reference/commandline/build to the `dhis-2/build-dev.sh` docker build command. Let hyperfine remove the jib cache for the application layers before running the jib build. Both approaches still leverage their caches for the base image.

```sh
hyperfine \
  --prepare 'rm -rf dhis-2/dhis-web/dhis-web-portal/target/jib-*' \
  'mvn --batch-mode --no-transfer-progress -DskipTests -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild' \
  'dhis-2/build-dev.sh'
```

#### Using warm caches

Benchmark 1: mvn --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild
  Time (mean ± σ):     22.726 s ±  0.515 s    [User: 44.002 s, System: 3.465 s]
  Range (min … max):   21.991 s … 23.700 s    10 runs

Benchmark 2: dhis-2/build-dev.sh
  Time (mean ± σ):      5.984 s ±  0.364 s    [User: 4.284 s, System: 0.863 s]
  Range (min … max):    5.626 s …  6.713 s    10 runs

Summary
  'dhis-2/build-dev.sh' ran
    3.80 ± 0.25 times faster than 'mvn --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild'

#### Using cold caches

Benchmark 1: mvn --batch-mode --no-transfer-progress -DskipTests -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild
  Time (mean ± σ):     22.039 s ±  0.190 s    [User: 42.973 s, System: 3.326 s]
  Range (min … max):   21.714 s … 22.320 s    10 runs

Benchmark 2: dhis-2/build-dev.sh
  Time (mean ± σ):     17.556 s ±  0.937 s    [User: 4.005 s, System: 0.832 s]
  Range (min … max):   16.401 s … 19.156 s    10 runs

Summary
  'dhis-2/build-dev.sh' ran
    1.26 ± 0.07 times faster than 'mvn --batch-mode --no-transfer-progress -DskipTests -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild'

### Method

Compared image sizes using [dive](https://github.com/wagoodman/dive).

You can get it using `docker pull wagoodman/dive`.

Pulled and analyzed the existing DHIS2 image of 2.38.1. I don't know if we publish one for 2.39 as well. The Jib image is built using 2.39.

```sh
docker run --rm -it \
    -v /var/run/docker.sock:/var/run/docker.sock \
    wagoodman/dive:latest dhis2/core:2.38.1-tomcat-9.0-jdk11-openjdk-slim
```

```sh
docker run --rm -it \
    -v /var/run/docker.sock:/var/run/docker.sock \
    wagoodman/dive:latest dhis2/core-dev:jib-2.39-SNAPSHOT
```

### Results

The [DHIS2 war itself has ~300MB](https://dhis2.org/downloads/). Locally on master it was 280MB for me. Both images use the same base image `tomcat:9.0-jdk11-openjdk-slim`. This explains why there is not much difference in terms of total image size.

Image name: dhis2/core:2.38.1-tomcat-9.0-jdk11-openjdk-slim
Total Image size: 1.1 GB
Potential wasted space: 8.1 MB
Image efficiency score: 99 %

Image name: dhis2/core-dev:jib-2.39-SNAPSHOT
Total Image size: 1.0 GB
Potential wasted space: 5.9 MB
Image efficiency score: 99 %

Dive allows us to see how [Jib layers the image](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-does-the-war-command-work). The first layers are the base image and should thus be the same with the previous Docker images. Jib then creates a layer for the dependencies as is common best practice in Docker images.

image

It then creates a layer for the [snapshots dependencies, resources and other files](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-does-the-war-command-work). Note that it pre-explodes the war file, which is why we get faster startup times 😄 

image

See in comparison the [COPY ./artifacts/dhis.war /usr/local/tomcat/webapps/ROOT.war
](https://github.com/dhis2/dhis2-core/blob/655a95b8838f191714212c88710c2de44e3ff902/docker/tomcat-debian/Dockerfile#L32) in the previous image

image

So Jibs layering allows us to reuse the dependency or resource layer if these were not changed.

## Running as non-root

Run DHIS2

```sh
docker compose up
```

The container has its own process space. Tomcat is running as PID 1

```sh
docker compose exec web cat /proc/1/cmdline
/usr/local/openjdk-11/bin/java-Djava.util.logging.config.file=/usr/local/tomcat/conf/logging.properties-Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager-Djdk.tls.ephemeralDHKeySize=2048-Djava.protocol.handler.pkgs=org.apache.catalina.webresources-Dorg.apache.catalina.security.SecurityListener.UMASK=0027-Dignore.endorsed.dirs=-classpath/usr/local/tomcat/bin/bootstrap.jar:/usr/local/tomcat/bin/tomcat-juli.jar-Dcatalina.base=/usr/local/tomcat-Dcatalina.home=/usr/local/tomcat-Djava.io.tmpdir=/usr/local/tomcat/temporg.apache.catalina.startup.Bootstrapstart
```

https://tomcat.apache.org/tomcat-9.0-doc/class-loader-howto.html

Owned by user uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)

```sh
docker compose exec web ls -l /proc/1/
total 0
-r--r--r--   1 nobody nogroup 0 Jul 19 09:52 arch_status
dr-xr-xr-x   2 nobody nogroup 0 Jul 19 09:49 attr
-rw-r--r--   1 nobody nogroup 0 Jul 19 09:52 autogroup
-r--------   1 nobody nogroup 0 Jul 19 09:52 auxv
-r--r--r--   1 nobody nogroup 0 Jul 19 09:49 cgroup
...
```

## Understanding Jib

If you want to dig deeper in what Jib does or how a particular aspect works refer to

https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#frequently-asked-questions-faq

Adding `-X -Djib.serialize=true` to the Jib build will output useful details like layer ids you can then also find in the different cache directories like `dhis-2/dhis-web/dhis-web-portal/target/jib-cache/layers`

```sh
mvn -X -Djib.serialize=true --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild
```

## What's next

If we agree on building with Jib we should

* use Jib to build the Docker images we currently publish to Dockerhub
* remove any directories/files that are not needed anymore like `docker`
* decide on what platforms we want to support and adapt our pipeline to publish those as well

### DIFF
--- a/.github/workflows/run-api-tests.yml
+++ b/.github/workflows/run-api-tests.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:    
   api-test:
     env:
-      CORE_IMAGE_NAME: "dhis2/core-dev:jib-2.39-SNAPSHOT"
+      CORE_IMAGE_NAME: "dhis2/core-dev:jib-2.40-SNAPSHOT"
       TEST_IMAGE_NAME: "dhis2/tests:local"
       PR_NUMBER: ${{ github.event.number }}
       DOCKER_CHANNEL: "dhis2/core-pr"
--- a/.github/workflows/run-api-tests.yml
+++ b/.github/workflows/run-api-tests.yml
@@ -36,21 +36,21 @@ jobs:
           mvn clean install --threads 2C --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f dhis-2/dhis-web/pom.xml
           mvn --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true -f ./dhis-2/dhis-web/dhis-web-portal/pom.xml jib:dockerBuild
 
-      # - name: Login to Docker Hub
-      #   if: "contains(github.event.pull_request.labels.*.name, 'publish-docker-image')"
-      #   uses: docker/login-action@v2
-      #   with:
-      #     username: ${{ secrets.DHIS2_BOT_DOCKER_HUB_USERNAME }}
-      #     password: ${{ secrets.DHIS2_BOT_DOCKER_HUB_PASSWORD }}
+      - name: Login to Docker Hub
+        # if: "contains(github.event.pull_request.labels.*.name, 'publish-docker-image')"
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DHIS2_BOT_DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DHIS2_BOT_DOCKER_HUB_PASSWORD }}
       
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
        
-      # - name: Publish docker image
-      #   if: "contains(github.event.pull_request.labels.*.name, 'publish-docker-image')"
-      #   run: |
-      #     docker tag $CORE_IMAGE_NAME $DOCKER_CHANNEL:$PR_NUMBER
-      #     docker push $DOCKER_CHANNEL:$PR_NUMBER
+      - name: Publish docker image
+        # if: "contains(github.event.pull_request.labels.*.name, 'publish-docker-image')"
+        run: |
+          # docker tag $CORE_IMAGE_NAME $DOCKER_CHANNEL:$PR_NUMBER
+          docker push $CORE_IMAGE_NAME
 
       - name: Build test image
         uses: docker/build-push-action@v3
--- a/.github/workflows/run-api-tests.yml
+++ b/.github/workflows/run-api-tests.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:    
   api-test:
     env:
-      CORE_IMAGE_NAME: "dhis2/core:local"
+      CORE_IMAGE_NAME: "dhis2/core-dev:jib-2.39-SNAPSHOT"
       TEST_IMAGE_NAME: "dhis2/tests:local"
       PR_NUMBER: ${{ github.event.number }}
       DOCKER_CHANNEL: "dhis2/core-pr"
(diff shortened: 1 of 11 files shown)


stian-sandvold (approved):

teleivo:
moving from spike to implementation. closing this PR used for the spike in favor of https://github.com/dhis2/dhis2-core/pull/11981

PR #11312, opened 2022-07-19
fix: Do not delete analyticsdataexchange table

enricocolasante:
This PR https://github.com/dhis2/dhis2-core/pull/10775 created `analyticsdataexchange` table.
The naming of the table conflicts with the naming of analytics tables that are created at some point and can also be dropped.
Depending on the order of the tests, `analyticsdataexchange` is dropped and then other tests need it to run.

In this PR I just prevent the deletion of this specific table, maybe we will need to think of something more organic to prevent this from happening again as this kind of bugs in the tests are quite difficult to find.


jbee (approved):

stian-sandvold (approved):

maikelarabori (approved):
As a temporary fix, this should be ok for the moment.
We need to find a definitive fix later.

Cheers,
Maikel

PR #11319, opened 2022-07-21
feat: Add headers param support in CPL [DHIS2-13511]

maikelarabori:
This feature aims to add support for the `headers` request param.

When this param is set, the response should respect the `headers` specified, returning only the defined ones in the respective order. The main change is the code related to this class `dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/shared/GridHeaders.java`

I also did a small code clean-up as part of this PR.

_You will notice a small commit from "mortenoh", that can be ignored. This is because I merged `master` into my commit._

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAttribute.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAttribute.java
@@ -27,9 +27,8 @@
  */
 package org.hisp.dhis.security.apikey;
 
-import java.io.Serializable;
-
 import org.hisp.dhis.common.DxfNamespaces;
+import org.hisp.dhis.common.EmbeddedObject;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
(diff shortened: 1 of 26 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAttribute.java
```diff
@@ -45,13 +44,18 @@
     @JsonSubTypes.Type( value = RefererAllowedList.class, name = "RefererAllowedList" ),
     @JsonSubTypes.Type( value = MethodAllowedList.class, name = "MethodAllowedList" ) } )
 @JacksonXmlRootElement( localName = "apiTokenAttribute", namespace = DxfNamespaces.DXF_2_0 )
-public abstract class ApiTokenAttribute implements Serializable
+public abstract class ApiTokenAttribute implements EmbeddedObject
```
d-bernat, line 47: why did you remove Serializable interface?

d-bernat (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAttribute.java
maikelarabori: As discussed, this change is not part of this PR.

gnespolino (approved):

PR #11321, opened 2022-07-22
feat: Add configurable max limit for TEI query [DHIS2-10679]

zubaira:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.hisp.dhis.user.UserService;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import com.google.common.collect.Sets;
+
+/**
+ * @author Zubair Asghar
+ */
+class TrackedEntityInstanceQueryLimitTest extends TransactionalIntegrationTest
```
teleivo, line 61: @enricocolasante what is the preferred one now, SingleSetupIntegrationTestBase? I think this one is slower at least when we would add more tests into this test class, right?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
```diff
@@ -0,0 +1,172 @@
...
+        params.setProgram( program );
+        params.setOrganisationUnits( Sets.newHashSet( orgUnitA ) );
+        params.setOrganisationUnitMode( OrganisationUnitSelectionMode.ALL );
+        params.setUser( user );
+        params.setSkipPaging( true );
+
+        List<Long> teis = trackedEntityInstanceService.getTrackedEntityInstanceIds( params,
+            false, false );
+
+        assertThat( teis, hasSize( is( 3 ) ) );
```
teleivo, line 169: Please remove the assertion on the size since the later hamcrest contains assertion will fail if there is an element which is not in your call to `contains`.

For example

```java
        assertThat(Arrays.asList("foo", "bar", "free"), contains("foo", "bar"));
```

leads to

```
java.lang.AssertionError: 
Expected: iterable containing ["foo", "bar"]
     but: not matched: "free"
```

which gives us the necessary context to fix an issue with our test/implementation. If the test were to become flaky/only fail on Github.

With the current assertions the integration test would fail on unequal size of expected/actual collections with no more context.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -145,22 +147,26 @@
...
-        checkNotNull( statementBuilder );
-        checkNotNull( organisationUnitStore );
+        checkNotNull( instanceStoreServicesSupplier );
 
-        this.statementBuilder = statementBuilder;
-        this.organisationUnitStore = organisationUnitStore;
+        this.instanceStoreServicesSupplier = instanceStoreServicesSupplier;
+    }
+
+    @PostConstruct
```
teleivo, line 166: `@PostConstruct` should not be used IMHO 😄 I think this leads to bad design. In order to for example test such a class in a unit test, you need to call init after having constructed an object of the class. If you forget to do that in a test your test diverges from what happens in production.

It looks to me as if HibernateTrackedEntityInstanceStore and the new TrackedEntityInstanceStoreServicesSupplier both depend on the StatementBuilder. If so let them both depend on it and Spring inject the StatementBuilder. If I did not make any mistake in reading the code we would then not need the PostConstruct 😅

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceStoreServicesSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+import org.hisp.dhis.organisationunit.OrganisationUnitStore;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author rajazubair
+ */
+@Component
+@RequiredArgsConstructor
+@Getter
+public class TrackedEntityInstanceStoreServicesSupplier
```
teleivo, line 45: Do we even need this class? If I understand the code correctly you would only need to add the DhisConfigurationProvider to the HibernateTrackedEntityInstanceStore 😅

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -1360,10 +1371,21 @@ private String getFromSubQueryLimitAndOffset( TrackedEntityInstanceQueryParams p
     {
         StringBuilder limitOffset = new StringBuilder();
         int limit = params.getMaxTeiLimit();
+        int teiQueryLimit = Integer
+            .parseInt( instanceStoreServicesSupplier.getConfigurationProvider()
```
teleivo, line 1375: It seems to me that we do not seem to validate the config and fail DHIS2 on startup if the config is invalid. This is unfortunate since a user can put in a String like `foo` into this config and it will only manifest when the first user makes a request that calls this store. If we do have the ability to validate the config and fail DHIS2 startup then lets do it 😄 If not, can we at least parse this only once maybe in the constructor and store it in a field?

teleivo (changes requested):
I have some suggestions 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
enricocolasante: Agreed.
With one test there is no difference but if we will had more tests the other one is better

enricocolasante (commented):
Nothing to add to what @teleivo already commented on

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
```diff
@@ -0,0 +1,162 @@
...
+        TrackedEntityInstanceQueryParams params = new TrackedEntityInstanceQueryParams();
+        params.setProgram( program );
+        params.setOrganisationUnits(Sets.newHashSet( orgUnitA ) );
+        params.setOrganisationUnitMode( OrganisationUnitSelectionMode.ALL );
+        params.setUser( user );
+        params.setSkipPaging(true);
+
+        List<Long> teis = trackedEntityInstanceService.getTrackedEntityInstanceIds( params, false, false );
+
+        assertThat( teis, Matchers.hasSize( Matchers.is( 3) ) );
```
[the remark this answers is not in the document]
zubaira: yeah agreed. that is was i am looking into how we can use different values for ConfigurationKey for different tests.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceStoreServicesSupplier.java
zubaira: yes we do need this. Otherwise just injecting DhisConfigurationProvider in constructure will introduce some sonarqube  code smell. This was one of the ways to avoid that without having to introduce much bigger change.

zubaira: there is a limit set on the number of parameters we can have in the constructor.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
stian-sandvold: Since we are testing that the limit is 3, would it make sense to assert the size is not >3?  @teleivo 

I would say potentially in this case the contains is more flaky, incase the default ordering is changing (Which we might actually do); Or @zubaira you can specify the order in the test to make sure its deterministic throughout changes like that

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -145,22 +147,28 @@
     // Dependencies
     // -------------------------------------------------------------------------
 
-    private final OrganisationUnitStore organisationUnitStore;
```
teleivo, line 148: Why are you removing the `final` on the dependencies? I think we should keep it

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -1360,10 +1373,19 @@ private String getFromSubQueryLimitAndOffset( TrackedEntityInstanceQueryParams p
     {
         StringBuilder limitOffset = new StringBuilder();
         int limit = params.getMaxTeiLimit();
+        int teiQueryLimit = systemSettingManager.getIntSetting( SettingKey.TRACKED_ENTITY_MAX_LIMIT );
 
         if ( limit == 0 && !params.isPaging() )
         {
-            return "";
+
+            return Optional.of( teiQueryLimit ).filter( a -> teiQueryLimit == 0 ).map( a -> "" )
```
teleivo, line 1381: I think we should simplify this into

```java
        if (teiQueryLimit > 0) {
            limitOffset
                    .append("limit")
                    .append(" ")
                    .append(teiQueryLimit)
                    .append(" ");
        }

        return limitOffset.toString();
```

I also assume a limit `teiQueryLimit < 0` does not make sense hence the `teiQueryLimit > 0` check.

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -702,7 +702,6 @@
      * other session when a new session is started.
      */
     MAX_SESSIONS_PER_USER( "max.sessions.per_user", "10", false ),
-
```
teleivo, line 705: Can you please remove unrelated changes. Might be accidental 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
teleivo: @zubaira now that you added more tests can you please change to SingleSetupIntegrationTestBase. I assume your tests are fine if `setUpTest` is only run once.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
```diff
@@ -0,0 +1,215 @@
...
+        userService.addUser( user );
+    }
+
+    @Test
+    void testConfiguredMaxTeiLimit()
+    {
+        systemSettingManager.saveSystemSetting( SettingKey.TRACKED_ENTITY_MAX_LIMIT, 3 );
+        TrackedEntityInstanceQueryParams params = new TrackedEntityInstanceQueryParams();
+        params.setProgram( program );
+        params.setOrganisationUnits( Sets.newHashSet( orgUnitA ) );
```
teleivo, line 166: Please prefer Java's `Set.of` for something like that 😄 (mind https://github.com/dhis2/wow-backend/blob/master/guides/java11_pitfalls.md)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentityinstance/TrackedEntityInstanceQueryLimitTest.java
```diff
@@ -0,0 +1,215 @@
...
+        params.setOrganisationUnitMode( OrganisationUnitSelectionMode.ALL );
+        params.setUser( user );
+        params.setSkipPaging( true );
+
+        List<Long> teis = trackedEntityInstanceService.getTrackedEntityInstanceIds( params,
+            false, false );
+
+        assertNotNull( teis );
+        assertEquals( 4, teis.size() );
```
teleivo, line 192: assertContainsOnly already checks the size using assertEquals 😄 So please remove assertEquals.

teleivo (changes requested):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
teleivo: @zubaira can you please address this 😄

teleivo (changes requested):
@zubaira you said in the standup that the PR conversations have been resolved, but there are still a bunch of open conversations. Can you please have a look 😄

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -1360,10 +1373,18 @@ private String getFromSubQueryLimitAndOffset( TrackedEntityInstanceQueryParams p
     {
         StringBuilder limitOffset = new StringBuilder();
         int limit = params.getMaxTeiLimit();
+        int teiQueryLimit = systemSettingManager.getIntSetting( SettingKey.TRACKED_ENTITY_MAX_LIMIT );
 
         if ( limit == 0 && !params.isPaging() )
         {
-            return "";
+
+            return Optional.of( teiQueryLimit ).filter( a -> teiQueryLimit > 0 ).map( a -> limitOffset
```
teleivo, line 1381: This one https://github.com/dhis2/dhis2-core/pull/11321/files/3f2d605b48a848b713fe96f2d85c88e2cb5049bc#r955991464
is not resolved.

Pasting it here in case it got lost for you 😄 

I think we should simplify this into

```java
        if (teiQueryLimit > 0) {
            limitOffset
                    .append("limit")
                    .append(" ")
                    .append(teiQueryLimit)
                    .append(" ");
        }

        return limitOffset.toString();
```

I also assume a limit `teiQueryLimit < 0` does not make sense hence the `teiQueryLimit > 0` check.

teleivo (changes requested):
one issue is still in 😄

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
zubaira: existing code is doing the same thing with the only difference of using `Optional` instead of `If` statement

enricocolasante: I really think that the if version is much more readable.
@zubaira You are right that they are doing the same thing but the use of Optional here is just confusing.
Optional should be used for variables that maybe are present and maybe they are not (null), in this case `teiQueryLimit ` can never be null, especially because it is an `int`.
What you want here is to do something if `teiQueryLimit` is greater than 0 and do nothing otheerwise.
An if statement here is much more human readable than the use of Optional

teleivo (approved):
Thank you for the change @zubaira 😄

lucaCambi77 (approved):

teleivo:
@stian-sandvold the smells are actually one smell that we need to tackle. I think we can merge this PR, but we should definitely schedule this work.

PR #11322, opened 2022-07-22
Feat: Refactor Event SQL to use named parameters - Complex bindings [ Dhis2-13455 ]

lucaCambi77:
[DHIS2-13455](https://jira.dhis2.org/browse/DHIS2-13455)

Add named parameters to events query for :
- filters
- sharing

Filters are used for data values that are in JSON so now the data type is no longer relevant.
Also, the events grid will be fixed later with [DHIS2-13508](https://jira.dhis2.org/browse/DHIS2-13508). For now, there is an empty map in case we use the same methods.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
@@ -44,9 +44,13 @@ import org.hisp.dhis.dataelement.DataElementService;
 import org.hisp.dhis.dxf2.events.event.EventSearchParams;
 import org.hisp.dhis.dxf2.events.event.EventService;
 import org.hisp.dhis.dxf2.events.event.Events;
+import org.hisp.dhis.event.EventStatus;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.program.Program;
+import org.hisp.dhis.program.ProgramInstance;
 import org.hisp.dhis.program.ProgramStage;
+import org.hisp.dhis.program.ProgramStatus;
+import org.hisp.dhis.program.ProgramType;
 import org.hisp.dhis.user.User;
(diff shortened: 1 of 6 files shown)


stian-sandvold (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
```diff
@@ -95,4 +115,207 @@ void testExportEventsWhenFilteringByEnrollment()
...
+
+        DataElement datael00001 = dataElement( "DATAEL00001" );
+
+        params.setDataElements( Set.of(
+            new QueryItem( datael00001, QueryOperator.IN, "value00001;value00002", datael00001.getValueType(),
+                null,
+                null ) ) );
+
+        Events events = eventService.getEvents( params );
+        assertNotNull( events );
```
enricocolasante, line 209: We are regularly having issues with tests failing on master only on CI with little to no context.

We should absolutely avoid assertions like this for this reason. Imagine this test fails on GitHub only telling you `expected 1 got 2`. What are you going to do? 😅 Multiple devs lost a few days on multiple occasions trying to debug tests with such assertions 😬 

Please use for example
https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertAll(java.lang.String,java.util.Collection)

like in https://github.com/dhis2/dhis2-core/blob/0d367e9b804ae5476694ab9545840e21b4cd68e5/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/trigramsummary/TrigramSummaryControllerTest.java#L243-L245

to group the assertion for checking the size and the content below into one assertAll

enricocolasante (commented):
A comment on the assertion of the test.
Because of flaky tests we are losing a lot of time to debug failing tests. More informative assertions will help to speed up on this

gnespolino (approved):

PR #11335, opened 2022-07-27
feat: add validations for metadata attribute

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-8915
https://jira.dhis2.org/browse/DHIS2-5537

- This PR add validations for Metadata Attribute which are called in the metadata import service.
- Added multiple unit tests and few integration tests.
- Had to modified the test data files  because they contains invalid attribute values, which are not relevant to the tests using those files.


netroms (approved):

stian-sandvold (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -263,6 +263,21 @@
     E6003( "Could not patch object with id {0}." ),
     E6004( "Attribute `{0}` has invalid GeoJson value." ),
     E6005( "Attribute `{0}` has unsupported GeoJson value." ),
+    E6006( "AttributeValue `{0}` is an invalid Integer value." ),
```
larshelge, line 266: I think we can use lower-case for the value types, like "integer" and "positive integer value".

vietnguyen: ok @larshelge , i will update those in another PR

PR #2996, opened 2019-03-12
Fix integration test failing with out of memory postgres

ameenhere:
Extended testcontainer classes to create a custom DhisPostgreSQLContainer that can be used to provide custom configurations for postgres when the postgres container starts for integration tests. max_locks_per_transaction is now set to 100 for postgres test containers.

PR #11339, opened 2022-07-27
chore: Upgrade testcontainers library

enricocolasante:
In this PR we are using the Junit Jupiter API for testcontainers to simplify the code and let testcontainers be responsible to create and manage the container for the DB.

All the "magic" happens in the JDBC url as explained here: https://www.testcontainers.org/modules/databases/jdbc


dhis-2/pom.xml
```diff
@@ -2324,6 +2326,11 @@
         <artifactId>tree</artifactId>
         <version>${tree.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>junit-jupiter</artifactId>
```
teleivo, line 2331: does it have scope test? if not we should set it 😄 maybe even be explicit about it even it if already is

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
```diff
@@ -42,16 +40,19 @@
...
 /**
  * @author Gintare Vilkelyte <vilkelyte.gintare@gmail.com>
  */
 @Slf4j
 @Configuration
 @ComponentScan( "org.hisp.dhis" )
+@Testcontainers
 public class IntegrationTestConfig
 {
+    public static final String POSTGIS_VERSION = "10-2.5-alpine";
```
teleivo, line 54: Not for this PR but should we update? This looks like postgres 10

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
```diff
@@ -105,21 +103,4 @@ public DhisConfigurationProvider dhisConfigurationProvider()
 
         return dhisConfigurationProvider;
     }
-
-    private JdbcDatabaseContainer<?> initContainer()
-    {
-        // NOSONAR
-        DhisPostgreSQLContainer<?> postgisContainer = ((DhisPostgreSQLContainer<?>) new DhisPostgisContainerProvider()
-            .newInstance()) // NOSONAR
-                .appendCustomPostgresConfig( "max_locks_per_transaction=100" )
```
teleivo, line 114: Do we still have this config with the changed setup? Not sure why we are using it. Maybe a performance tuning thing for tests 🤔

dhis-2/pom.xml
enricocolasante: This is not a test dependency, it is used in `IntegrationTestConfig` that is not in the test folder.

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
enricocolasante: I agree, maybe we should even find a way to sync it with our postgres dependencies

enricocolasante:
> NICE 🎉
> 
> So I see we are using https://www.testcontainers.org/modules/databases/jdbc to speed up tests
> 
> Whats not so clear to me is if we still use the singleton pattern like before https://www.testcontainers.org/test_framework_integration/junit_5/, so one container is started, all integration tests are run, then the container is stopped
> 
> https://www.testcontainers.org/modules/databases/jdbc talks about Running container in daemon mode

As discussed we will have a deeper look at the singleton patter later. It seems to me that this is still a Singleton because of the way we are defining the Spring bean that is providing the data source to access the DB.

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
enricocolasante: I think @ameenhere can help here
I can see from this testcontainers issue that this was a problem at some point https://github.com/testcontainers/testcontainers-java/issues/1278

What do we need this configuration for?
And do we have a way to configure it from JdbcUrl?

dhis-2/pom.xml
teleivo: Should we set it to test here and override it to compile for the `dhis-support-test` module? Just to make sure anyone that would use it directly does not pull it in as a compile time dependency?

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
teleivo: Do you mean the driver https://github.com/dhis2/dhis2-core/blob/f788463e90e9c1c285e752ec3a255a068a9abd3f/dhis-2/pom.xml#L100 ?

It works with https://jdbc.postgresql.org/download.html 
| supports PostgreSQL 8.2 or newer

Postgis https://github.com/dhis2/dhis2-core/blob/f788463e90e9c1c285e752ec3a255a068a9abd3f/dhis-2/pom.xml#L101 probably needs to align with the Postgis version we use

dhis-2/pom.xml
enricocolasante: I think we have the `maven-dependency-plugin` that is helping us with it.
If someone is using it directly in a test and add it as a compile time dependency will get an error from the analyser

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
teleivo: Resolving this since this will not be part of this PR 😄

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/config/IntegrationTestConfig.java
teleivo: I found the original PR that added it https://github.com/dhis2/dhis2-core/pull/2996
Since we got OOM errors back then we can still merge this PR and see if they reappear. The PR was from `Mar 13, 2019` so lots might have changed in the meantime 😋

teleivo (approved):

stian-sandvold (approved):

PR #11385, opened 2022-08-03
fix: Visualization download to remove redundant 0 decimals [DHIS2-13460]

maikelarabori:
**_[Backport from master/23.9]_**

The download of a Visualization is showing "0" digits for numbers like "237.0", "55.0", "13300.0" and so on.

Instead, it should remove `0` decimal digits, as they are redundant. So the numbers above would became: "237", "55", "13300".

This happens when exporting Data Elements of Value Type "Number" because the number of decimal digits is not configurable (in Indicators this option exists).

This PR aims to remove the redundant `zeros` for the Data Element results.

The GET request below can be used to reproduce the issue in master, and also to display the expected results on this branch.
```
http://localhost:8080/dhis/api/analytics.json?dimension=dx:fbfJHSPpUQD;cYeuwXTCPkU;Jtf34kNZhzP;hfdmMSPBgLG&dimension=pe:LAST_12_MONTHS&showHierarchy=false&hierarchyMeta=false&includeMetadataDetails=true&includeNumDen=true&skipRounding=false&completedOnly=false&outputIdScheme=NAME&filter=ou:ImspTQPwCqd
```

**_This backport includes also the PR related to `DHIS2-8184`, where we apply the same behaviour for Reporting Rates for consistency. As the fixes are very close and related, I decided to backport both on the same ticket. See ticket https://jira.dhis2.org/browse/DHIS2-8184_**


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsUtils.java
```diff
@@ -263,7 +268,34 @@ else if ( params.isSkipRounding() )
...
+     * -232.0000 -> true
+     *
+     * 133.25 -> false
+     *
+     * -1045.00000001 -> false
+     *
+     * @param value
+     * @return true if the value has "0" as decimal digits, false otherwise
+     */
+    public static boolean endsWithZeroAsDecimalDigit( final double value )
```
larshelge, line 296: Maybe rename to `endsWithZeroAecimal`, shorter and same meaning.

larshelge (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsUtils.java
maikelarabori: Done

PR #11386, opened 2022-08-03
fix: Upgrade rule engine version [DHIS2-13468]

enricocolasante:


stian-sandvold (approved):

larshelge (approved):

larshelge:
Let's remember to set it back to non-snapshot version before release.

PR #11388, opened 2022-08-03
fix: Remove decimals zeros in rep. rates/val. rules [DHIS2-8184]

maikelarabori:
This change is a complement of https://github.com/dhis2/dhis2-core/pull/11199

It will also remove decimal "zeros" from report rates and validation rules for consistency.

So, numbers like `123.0` will simply become `123`.


gnespolino (approved):
Looks good 👍

d-bernat (approved):

PR #11400, opened 2022-08-04
fix : Fix tracker data view query [DHIS2-12249]

zubaira:
PR will fix the issue of wrongly populated uids.
Previous `ProgramTrackedEntityAttribute` and `ProgramStageDataElement` uids were being populated instead of uids of `TrackedEntityAttribute` and `DataElement`.


stian-sandvold (approved):

larshelge:
Is it okay to change an existing, released Flyway migration? I assume this would crash for all implementations with a checksum mismatch, without enabling Flyway repair, which we have not told any implementations to do?

zubaira:
> Is it okay to change an existing, released Flyway migration? I assume this would crash for all implementations with a checksum mismatch, without enabling Flyway repair, which we have not told any implementations to do?

That's right @larshelge
For exactly this reason I wanted to raise this as draft PR and keep looking for a way forward. But accidentally this was raised as review PR.

teleivo (approved):

PR #11417, opened 2022-08-09
fix: log leader election job on debug [DHIS2-13562]

jbee:
End goal is to have the leader election job logging its "information" log entries on debug level. This is solved by introducing a general flag per job type to switch the log level of general messages from INFO to DEBUG.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
@@ -149,6 +149,15 @@ public enum JobType
         this.relativeApiElements = relativeApiElements;
     }
 
+    /**
+     * @return when true, general information on job progress should be logged
+     *         on debug level, otherwise when false it is logged on info level
+     */
+    public boolean isDefaultLogLevelDebug()
+    {
+        return this == LEADER_ELECTION;
+    }
+
     public boolean isUsingNotifications()
     {
         return this == RESOURCE_TABLE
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
```diff
@@ -149,6 +149,15 @@
         this.relativeApiElements = relativeApiElements;
     }
 
+    /**
+     * @return when true, general information on job progress should be logged
+     *         on debug level, otherwise when false it is logged on info level
+     */
+    public boolean isDefaultLogLevelDebug()
+    {
+        return this == LEADER_ELECTION;
```
larshelge, line 158: I think this is okay. Just wondering, would it be slightly more elegant if we had a (default) method in the JobConfiguration  called `getDefaultLogLevel` or similar, which could be overridden by the node leader election job? We could avoid the job type being aware of particular job types that way.

larshelge (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
jbee: We could. But so far this sort of configuration comes from the type and having this one being elsewhere would be inconsistent. Also Slf4j does not have a log level enum or methods that would accept such enum so I decided we just have a switch to change between info and debug.

PR #11443, opened 2022-08-15
fix: Update regex not to be greedy

stian-sandvold:
Reported by Sonarcloud as a potential problem:
When matching large strings, greedy regex can be overloaded and cause issues. By adding "+" after "*", we make the matching possesive rather than greedy.

Tested using existing unit tests (program rules, textpattern) and running a fresh db (flyway).

### DIFF
--- a/dhis-2/dhis-api/src/test/java/org/hisp/dhis/textpattern/TestTextPatternMethod.java
+++ b/dhis-2/dhis-api/src/test/java/org/hisp/dhis/textpattern/TestTextPatternMethod.java
@@ -38,8 +38,9 @@ class TestTextPatternMethod
     @Test
     void testValidateText()
     {
-        String[] valid = { "\"Hello world!\"", "\"Hello \\\"world\\\"\"" };
-        String[] invalid = { "Hello world", "Hello \" world", "\"Hello world", "Hello world\"" };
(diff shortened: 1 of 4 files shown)


larshelge (approved):

dhis-2/dhis-api/src/test/java/org/hisp/dhis/textpattern/TestTextPatternMethod.java
```diff
@@ -38,8 +38,9 @@
     @Test
     void testValidateText()
     {
-        String[] valid = { "\"Hello world!\"", "\"Hello \\\"world\\\"\"" };
-        String[] invalid = { "Hello world", "Hello \" world", "\"Hello world", "Hello world\"" };
+        String[] valid = { "\"Hello world!\"", "\"Hello \\\"world\\\"\"",
+            "\"This is a text with more than two words\"" };
+        String[] invalid = { "Hello world", "Hello \" world", "\"Hello world", "Hello world\"", "\"Hello \"world\"\"" };
```
teleivo, line 43: I think this test should be replaced by a parametrized test so each case can fail independently.

teleivo (approved):
Can you check the test and maybe follow up with another PR to update it.

PR #11444, opened 2022-08-15
fix: reuse one instance of Random, instead of creating new each time

stian-sandvold:
Suggestion by sonarCloud, see issue for info

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/MathUtils.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/MathUtils.java
@@ -71,6 +71,8 @@ public class MathUtils
 
     private static final Pattern ZERO_PATTERN = Pattern.compile( "^0(\\.0*)?$" );
 
+    private static final Random RANDOM = new Random();
+
     /**
      * Indicates whether the given double valid, implying it is not null, not
      * infinite and not NaN.
@@ -478,7 +480,7 @@ public class MathUtils
      */
     public static int getRandom()
     {
-        return new Random().nextInt( 999 );
+        return RANDOM.nextInt( 999 );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/MathUtils.java
```diff
@@ -71,6 +71,8 @@
 
     private static final Pattern ZERO_PATTERN = Pattern.compile( "^0(\\.0*)?$" );
 
+    private static final Random RANDOM = new Random();
+
```
netroms, line 75: Might be better to use a ThreadLocalRandom.
See Random.class java doc: Consider instead using java.util.concurrent.ThreadLocalRandom in multithreaded designs.

netroms (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/MathUtils.java
stian-sandvold: Thanks, I'll check it out!

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/MathUtils.java
```diff
@@ -71,6 +72,8 @@
 
     private static final Pattern ZERO_PATTERN = Pattern.compile( "^0(\\.0*)?$" );
 
+    private static final Random RANDOM = ThreadLocalRandom.current();
+
```
larshelge, line 76: Just wondering if this is a great idea, as a class scope variable means that all threads will use the same instance.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html

From the javadoc it seems the recommended usage is:

```
ThreadLocalRandom.current().nextX(...) 
```
I.e. `ThreadLocalRandom.current()` returns the current thread's ThreadLocalRandom.


[current](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html#current--)()
Returns the current thread's ThreadLocalRandom.

stian-sandvold: I think you are right, I'll remedy this right away!

larshelge (approved):

PR #11446, opened 2022-08-15
fix: reorder code to deal with closing connection in try-catch

stian-sandvold:


teleivo (approved):
Does the issue then show up as fixed in Sonarcloud? It does not show up in the report. Would be nice if it also gave some stats on how much a PR improves existing code 😄

larshelge (approved):

PR #11448, opened 2022-08-15
feat: attribute ID columns for metadata CSV [DHIS2-11867]

jbee:
Adds attribute ID columns to metadata CSV export.

Also the CSV generation no longer makes the intermediate step of creating a `List<Map<?,?>>` for the rows.
Instead each row is filled into an array (which is reused) and then written directly to the string output via a mapped object writer.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
@@ -297,16 +297,16 @@ public abstract class AbstractFullReadOnlyController<T extends IdentifiableObjec
             // We just split on ',' here, we do not try and deep dive into
             // objects using [], if the client provides id,name,group[id]
             // then the group[id] part is simply ignored.
-            for ( String splitField : field.split( "," ) )
+            for ( String fieldName : field.split( "," ) )
             {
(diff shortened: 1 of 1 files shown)


netroms (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
```diff
@@ -292,19 +303,29 @@ public ResponseEntity<String> getObjectListCsv(
...
+                if ( (property.isCollection() && property.itemIs( PropertyType.REFERENCE )) )
                 {
-                    schemaBuilder.addColumn( property.getName() );
-                    properties.add( property );
+                    schemaBuilder.addArrayColumn( property.getCollectionName() );
+                    obj2valueByProperty.put( property.getCollectionName(), obj -> {
+                        Object value = ReflectionUtils.invokeMethod( obj, property.getGetterMethod() );
+                        @SuppressWarnings( "unchecked" )
+                        Collection<IdentifiableObject> collection = (Collection<IdentifiableObject>) value;
+                        return collection.stream().map( PrimaryKeyObject::getUid ).collect( toList() );
```
larshelge, line 321: Could this part be factored out into a method with javadoc to help readability?

larshelge (approved):

PR #11455, opened 2022-08-16
feat: save user settings as part of user payload [DHIS2-8601]

jbee:
### Summary
Allows to send a `UserSettings` as part of the `User` object to store/update user settings in same operation as the user.

The `UserSettings` object has a field for each `UserSettingKey`. This is so the type conversion is handled by jackson.
When a new key is added the accessor (getter) is provided with the enum linking the two so it is hard to forget to now update both places in sync.

Each field allows the name with the `key` prefix as expected by the `/userSettings` API. 
For readability it also allows names without the `key` prefix starting with a lower case letter.

### Limitations
Only non-`null` values will be considered. That means this feature cannot be used to clear a setting.
This sematic was chosen to avoid accidentally clearing settings in case they were not included in the payload.

### Automatic Testing
A new test scenario was added where a user is updated with a setting.

### Manual Testing
1. fetch a user as JSON, GET http://localhost:8080/api/users/CotVI2NX0rI
2. extend the user JSON with a `settings` member, for example:
  ```json
"settings": {
    "keyUiLocale": "de"
  },
  ```
3. patch or put the updated user JSON, PUT http://localhost:8080/api/users/CotVI2NX0rI
4. check the setting has updated as provided with the modified user JSON, http://localhost:8080/api/userSettings/keyUiLocale?userId=CotVI2NX0rI

Try similar changes with the settings name without `key` prefix (`keyUiLocale` => `uiLocale`), try updating multiple settings in one go, try storing a new user with settings.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
@@ -119,7 +119,7 @@ public class UserSettingController
         return result;
     }
 
-    @GetMapping( value = "/{key}" )
+    @GetMapping( value = "/{key}", produces = "text/plain" )
     public String getUserSettingByKey(
         @PathVariable( value = "key" ) String key,
         @RequestParam( required = false, defaultValue = "true" ) boolean useFallback,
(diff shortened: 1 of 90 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
```diff
@@ -119,7 +119,7 @@ public Map<String, Serializable> getAllUserSettings(
         return result;
     }
 
-    @GetMapping( value = "/{key}" )
+    @GetMapping( value = "/{key}", produces = "text/plain" )
```
jbee, line 122: FYI: this is just making clear what the actual behaviour always has been. This method does not produce JSON as the returned string is not quoted.

larshelge (approved):

larshelge:
Looks good :+1:

netroms (approved):

PR #11456, opened 2022-08-16
fix: Only restrict page size to max limit if greater than zero [DHIS2-13606]

larshelge:
The event analytics query engine restricted the page size in the paging logic to the max limit of records from the `ANALYTICS_MAX_LIMIT` even when that value as 0. The value 0 in this case represents unlimited records, and the restriction should only happen if the max limit of records is greater than 0. Currently if the max limit is set to unlimited, no events are ever returned from the event analytics query endpoint.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/PagingCriteriaProcessor.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/PagingCriteriaProcessor.java
@@ -43,7 +43,6 @@ import org.springframework.stereotype.Component;
 @RequiredArgsConstructor
 public class PagingCriteriaProcessor implements Processor<AnalyticsPagingCriteria>
 {
-
     private final SystemSettingManager systemSettingManager;
 
     // TODO: DHIS2-13384 we would really like to have all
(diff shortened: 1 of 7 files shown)


maikelarabori (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/PagingCriteriaProcessor.java
```diff
@@ -43,19 +43,17 @@
...
-
     private final SystemSettingManager systemSettingManager;
 
     // TODO: DHIS2-13384 we would really like to have all
     // criteria/request/params to be
     // immutable, but PagingCriteria is not
     // returning it for now, should be converted to use builders
     @Override
-    public AnalyticsPagingCriteria process( final AnalyticsPagingCriteria pagingCriteria )
+    public AnalyticsPagingCriteria process( AnalyticsPagingCriteria pagingCriteria )
```
maikelarabori, line 53: I realized that a few `final keywords were removed. I usually use it everywhere to make it clear they are immutable.
Personally, I found it easier to read and a good way to avoid side effects.
If you guys prefer to not use `final`, please let me know. I will stop using it and will start removing them instead =D

larshelge: Cool. It is fine per se, sorry for removing. In this case, it does not really make the argument immutable, as the inner state (properties) of `AnalyticsPagingCriteria` could still be changed, so it would give sort of a false sense of immutability.

maikelarabori:
Yes, in that particular case the object can be changed, but the reference cannot be reassigned.
That’s why I still like using it...

PR #11460, opened 2022-08-17
Revert "fix: Filter not EQ in analytics event query filters out null [DHIS2-13563] (#11430)"

jbee:
This reverts commit c85a386ccdc7aa7865bc9d850f61f6b530328b86.

For some reason there was a conflict so I just kept the alternative of that conflict.


jbee:
Revert has not the outcome we hoped for.

PR #11464, opened 2022-08-17
fix: cron expression validation [DHIS2-9768]

jbee:
### Summary
`CronSequenceGenerator` did not accept the valid expression `0 0 1 ? * TUE-SUN`.
As it was deprecated I tried `CronExpression.isValidExpression` instead as our scheduling is also based on `CronExpression` class. That one does accept the expression.

### Manual Testing
* create a job in the scheduler app using a cron expression like `0 0 1 ? * TUE-SUN` that has a day-of-the-week range. 
* try to safe the job successfully


netroms (approved):

larshelge (approved):

larshelge:
Nice solution :+1:

larshelge (approved):

PR #11465, opened 2022-08-17
fix: update regex to fix sonarcloud issue [DHIS2-13595]

zubaira:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/listener/CommandSMSListener.java
```diff
@@ -72,7 +72,7 @@
     extends
     BaseSMSListener
 {
-    private static final String DEFAULT_PATTERN = "([^\\s|=]+)\\s*\\=\\s*([^|=]+)\\s*(\\=|$)*\\s*";
+    private static final String DEFAULT_PATTERN = "([^\\s|=]+)\\s*\\=\\s*([^|=]+)\\s*(\\=|$)*+\\s*";
```
larshelge, line 75: Just curious, is `*+` really meaningful in regex? If it is supposed to be greedy, should it be `*?` ?

stian-sandvold: Apprently *+ makes it posessive, avoiding overloading the regex parsing for big strings (It was suggested and highlighted in Sonarcloud).

zubaira: https://sonarcloud.io/project/issues?issues=AX_PYZLN5EDiRQZeL591&open=AX_PYZLN5EDiRQZeL591&id=dhis2_dhis2-core

stian-sandvold (approved):

larshelge (approved):

PR #11466, opened 2022-08-17
fix: Failing H2 tests

stian-sandvold:
This change fixes the test-fails locally for me. I have been unable to understand why the problem happened in the first case, but this change seem to resolve the issue.


teleivo (approved):

vilkg (approved):

teleivo:
@netroms can you please have a look at this change and follow up with us on the failing h2 tests. We do not understand how ApiTokenAuthenticationTest was passing before without this `ConfigProviderConfiguration` 🤷🏻 Also we found base classes DhisControllerWithApiTokenAuthTest and DhisControllerWithJwtTokenAuthTest only being used once. @stian-sandvold and I think they should be merged into the only tests needing that setup as we already have quite a complex base class setup.

PR #11468, opened 2022-08-17
fix: Fix enrollment error message [DHIS2-11410]

muilpp:
If tei doesn't exist, show the entity instead of the program uid.


stian-sandvold (approved):
Perfect, great job!

vilkg (approved):

PR #11471, opened 2022-08-17
fix: NPE when changing avatar image without F_USER_VIEW authority [DHIS2-9761]

jbee:
NPE occured because of the `currentUser.getAvatar().equals( fileResource )` where `currentUser.getAvatar()` could be null if it was never set. The NPE was indirectly prevented by the `F_USER_VIEW` authority as the condition would already be true so that the second part would never be evaluated.

With this fix in place the file itself can be uploaded but a user lacking authority might still not be able to then set the image file resource as avatar as he cannot modify its own profile (`User` object). This is a more general issue that maybe is best addressed in another PR.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
@@ -27,8 +27,6 @@
  */
 package org.hisp.dhis.user;
 
-import static java.util.Collections.emptySet;
-import static java.util.Collections.unmodifiableSet;
 import static org.springframework.beans.BeanUtils.copyProperties;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


netroms (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -287,16 +289,20 @@ public String getUserRoleNames()
      */
     public Set<String> getAllAuthorities()
     {
+        if ( userRoles.isEmpty() )
+        {
+            return emptySet();
+        }
+        if ( userRoles.size() == 1 )
```
larshelge, line 296: Just curious, why do we need this optimization for size == 1?

jbee: It is not needed. Maybe it is unrealistic that users have just one role? I thought for installations with many users most might just have one role and there it would help to avoid copying collections.

larshelge: Okay understand, thanks for the explanation. I think in most cases users will have more than one role, and we might be making the code a bit harder to read for a small performance benefit. Not a big deal of course.

jbee: Reverted it now

PR #11474, opened 2022-08-17
fix: Conflict when storing a file resource with existing UID [DHIS2-6288]

jbee:
Mainly this should provide a nicer server response then the 500 caused by FK containt violation in case a file resource with a UID already exist when the client sends the UID.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.webapi.utils;
 
-import static java.lang.String.format;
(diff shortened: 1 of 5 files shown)


vietnguyen (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
```diff
@@ -192,8 +193,11 @@ public FileResource saveFileResource( String uid, MultipartFile file, FileResour
 
         File tmpFile = toTempFile( file );
 
+        if ( uid != null && fileResourceService.fileResourceExists( uid ) )
+        {
+            throw new WebMessageException( conflict( format( "A resource with Id %s already exists", uid ) ) );
```
larshelge, line 198: Do we want to use an `ErrorCode` here? I know we have many places without but we could try to at least use error codes for new exceptions?

larshelge (approved):

PR #11491, opened 2022-08-18
refactor: Metadata options logic in event analytics

maikelarabori:
This refactor aims to simplify the logic of finding the correct Option objects based on the current event params and filters.

The main change can be found at [QueryItemHelper.java](https://github.com/dhis2/dhis2-core/compare/meta_options_logic?expand=1#diff-55f50751bb06e3ad35a87f23d1c36472f04cbc28a79043f3329e0bd3604ecd0d).
The logic was extracted into this class as it seemed a better home while making the unit testing easier.

I also improved the unit tests, as some of them were testing different methods/operations on the same test case. The global setup was also removed as it leads to confusion long term. It's better to have each unit test fully independent with its own mocking/stubbing scoped per scenario.

Also, notice that I found it better to represent the list of `Option` objects as `Set` instead of `List`.
As we cannot have repeated `Option` objects, the `Set` interface seems more appropriate. At the same time, the `LinkedHashSet` implementation will keep the insert and iteration order.

Javadocs were also improved where it made sense. The `final` keyword was not used and was removed where applicable.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
```diff
@@ -294,35 +291,56 @@ protected void addMetadata( EventQueryParams params, List<DimensionItemKeywords.
     {
         if ( !params.isSkipMeta() )
         {
-            final Map<String, Object> metadata = new HashMap<>();
+            Map<String, Object> metadata = new HashMap<>();
+            Map<String, List<Option>> dimensionOptions = getItemOptions( grid, params );
```
larshelge, line 295: Here we could use a ListValuedMap potentially.

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/ListValuedMap.html

larshelge (approved):

PR #11501, opened 2022-08-19
fix: boolean value from redis not comparable

vietnguyen:
https://jira.dhis2.org/browse/DHIS2-13592

- Below code return false for TRUE value if server uses Redis. It works fine without redis enabled.

```
getSystemSetting( key, Boolean.class ) == Boolean.TRUE
```

Screen Shot 2022-08-19 at 15 41 39

- Code return TRUE with the fix

Screen Shot 2022-08-19 at 15 41 55

### DIFF
--- a/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SystemSettingManager.java
+++ b/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SystemSettingManager.java
@@ -175,7 +175,7 @@ public interface SystemSettingManager
     @Transactional( readOnly = true )
     default boolean getBoolSetting( SettingKey key )
     {
-        return getSystemSetting( key, Boolean.class ) == Boolean.TRUE;
+        return Boolean.TRUE.equals( getSystemSetting( key, Boolean.class ) );
     }
 
     @Transactional( readOnly = true )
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SystemSettingManager.java
```diff
@@ -175,7 +175,7 @@ default Boolean getBooleanSetting( SettingKey key )
     @Transactional( readOnly = true )
     default boolean getBoolSetting( SettingKey key )
     {
-        return getSystemSetting( key, Boolean.class ) == Boolean.TRUE;
+        return getSystemSetting( key, Boolean.class ) == Boolean.TRUE.booleanValue();
```
larshelge, line 178: I think this is good. Though it seems we get an object Boolean back from `getSystemSetting`, so by doing == boolean primitive we rely on auto-boxing for this to work. Would this be a slightly better approach, ` Boolean.TRUE.equals` will compare object Booleans and also tolerate nulls?

`return Boolean.TRUE.equals( getSystemSetting( key, Boolean.class ) );`

vietnguyen: Agree, using `equals()` is a cleaner approach. Fixed now.

larshelge (approved):

PR #11512, opened 2022-08-21
test: transactional tests do not need to empty DB manually

teleivo:
as this is done by `@Transactional`

Needs
- [ ] https://github.com/dhis2/dhis2-core/pull/11510


teleivo:
fixed in https://github.com/dhis2/dhis2-core/pull/11612 😄

PR #11523, opened 2022-08-22
fix: Allow fileResources to be deleted from data entry app [DHIS2-9378]

abyot:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/datavalue/DataValueController.java
```diff
@@ -407,6 +416,18 @@ private void saveDataValueInternal( DataValueDto dataValue, User currentUser )
         }
     }
 
+    private String getFileResourceOwner( String de, String co, String ao, String pe, String ou )
```
larshelge, line 419: It feels like this method should be centralized and documented somehow, e.g. since the order of the identifiers is significant, so that we make sure we generate the file resource owner identifier in the same way everywhere.

stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/datavalue/DataValueController.java
abyot: yes, we need to document. otherwise, I felt this is pretty much an internal stuff... unless one manipulates file resources externally.

larshelge: Right. Just checking, do we create such identifiers anywhere else in the code base?

larshelge (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/datavalue/DataValueController.java
abyot: this combination of identifier is only for aggregate. otherwise, we are using fileResourceOwner for both tei and event datavalues. but there we just use event and tei uids.

PR #11524, opened 2022-08-22
ci: run all tests on Jenkins

teleivo:
by excluding the default profile with `-default` surefire will pick up all tests no matter what group/tag they belong to.

Note that we are then running without any profile, which you can check using

```sh
mvn help:active-profiles -P -default -f dhis-2/pom.xml
```

Adapted pipeline to use the pom.xml instead of the pom-full.xml since having 2 has the risk of them diverging as they have. pom-full.xml only being used on Jenkins and pom.xml locally and on GitHub.

## Validation

Surefire reports are uploaded to GitHub for every test type (unit, integrationH2, integration). For example run
https://github.com/dhis2/dhis2-core/actions/runs/2904511567

The surefire reports from GitHub show:
3113 unit
2330 integration
872  integrationH2
6315 total

The same amount of total tests is shown to run on Jenkins
https://ci.dhis2.org/job/dhis2-core-dev/job/ci%252Ftests-not-running-on-master/18/testReport/


teleivo:
If you @vilkg and @radnov agree with [this change](https://github.com/dhis2/dhis2-core/pull/11524/files#diff-a1d2e6c7c735f6ceb48974c9a41a7fa8aafbf865ef13e1a08d15e95513dd79d3L47-R48) I will adapt jenkinsfiles/dev to only have the same diff. We should then be able to merge this 😄

radnov (approved):
[The changes](https://github.com/dhis2/dhis2-core/pull/11524/files#diff-a1d2e6c7c735f6ceb48974c9a41a7fa8aafbf865ef13e1a08d15e95513dd79d3L47-R48) look good, nice work 👍

vilkg:
Looks good @teleivo !

vilkg (approved):

PR #11538, opened 2022-08-23
ci: run all tests on Jenkins [2.38]

teleivo:
backport of https://github.com/dhis2/dhis2-core/pull/11524

We did not run all tests on Jenkins since our default profile which we
activate by default only runs unit tests (excludes tests tagged with
integration).

by excluding the default profile with -default surefire will pick up all
tests no matter what group/tag they belong to.

Note that we are then running without any profile, which you can check using

mvn help:active-profiles -P -default -f dhis-2/pom.xml

Adapted pipelines to use the pom.xml instead of the pom-full.xml since having
2 has the risk of them diverging as they have. pom-full.xml only being used on
Jenkins and pom.xml locally and on GitHub.

## Validation

Surefire reports are uploaded to GitHub for every test type (unit, integration). For example run
https://github.com/dhis2/dhis2-core/actions/runs/2911600044

GitHub surefire reports show:
4940 unit (note here unit tests include what we call integrationH2 tests on master https://github.com/dhis2/dhis2-core/pull/11524)
 950 integration
  17 integration analytics
5907 total

Jenkins https://ci.dhis2.org/view/dhis2-core/job/dhis2-core-dev/job/2.38 shows:
4940 total **before these changes apply to Jenkins**. So **967** tests do not seem to run at the moment.


radnov (approved):
LGTM 👍

larshelge (approved):

PR #11540, opened 2022-08-23
fix: Refactor Event SQL to use named parameters - Events grid - [ DHIS2-13508 ]

lucaCambi77:
https://jira.dhis2.org/browse/DHIS2-13508
Add named parameter to Events grid. Overall clean code to make it more even with event SQL.
Add _assertAll_ to tests


lucaCambi77:
showing strange errors in tests

PR #11546, opened 2022-08-23
fix: SqlView JSON and XML response with pager [DHIS2-9813]

jbee:
### Summary
Reverts the first attempt to fix the issue in #11234 as it was lacking the root object with pager level.

New solution adds the `GridResponse` as root object using the existing `Pager`.
For JSON Jackson takes care of the serialisation. As Jackson support XML was removed the `GridUtils` got extended so the can handle a `GridResponse` and write the additional fields to XML.

### Manual Testing 
* login as `system` user (or any other with `ALL` authority)
* create a new SQL query view
* check JSON output with pager `/api/29/sqlViews/{uid}/data?paging=true`
* check JSON output without pager `/api/29/sqlViews/{uid}/data?paging=false`
* check XML output with pager `/api/29/sqlViews/{uid}/data.xml?paging=true` 
* check XML output without pager `/api/29/sqlViews/{uid}/data.xml?paging=false`

check here means to check it looks like a valid JSON/XML document and compare the structure to that returned by older versions.They should have the same structure.

### Example Output
JSON (shortened for main structure, `headers` and `rows` usually not empty)
```json
{
  "pager": {
    "page": 1,
    "pageCount": 1,
    "total": 22,
    "pageSize": 50
  },
  "listGrid": {
    "headerWidth": 15,
    "width": 15,
    "title": "test",
    "height": 22,
    "headers": [],
    "rows": []
  }
}
```

XML (also shortened)
```xml
<metadata xmlns="http://dhis2.org/schema/dxf/2.0">
<pager>
<page>5</page>
<pageCount>11</pageCount>
<total>22</total>
<pageSize>2</pageSize>
</pager>
<listGrid title="test" width="15" height="2">
<headers>
<header name="categoryid" column="categoryid" type="java.lang.String" hidden="false" meta="false"/>
<!-- ... -->
</headers>
<rows>
<row>
<!-- ... -->
</row>
</rows>
</listGrid>
</metadata>
```


jimgrace (approved):

netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GridResponse.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Getter
+@AllArgsConstructor
+public final class GridResponse
+{
+    @JsonProperty
+    private final Pager pager;
+
+    @JsonProperty
+    private final Grid listGrid;
```
larshelge, line 46: Re the name `listGrid`, I think `grid` would be a better name, as the interface is called grid and `ListGrid` is just an implementation of it using an array list. Is it too bad of a breaking change to use `grid`? Has it always been `listGrid`?

benguaraldi: @larshelge I think `grid` would be a better name, too, but it's definitely easier on us if it isn't changed.  I estimate that it will take PEPFAR at least 20 hours of development time (and maybe several times that) to incorporate this into all of our DATIM and publicly-available apps.  I imagine other places with SQL view apps would have similar amounts of work.

Note that apps with SQL views that need to work on multiple versions of DHIS2 will need to have internal logic that could use either `listGrid` or `grid`.

I don't know what it was prior to DHIS 2.27, but I think it's been listGrid at least since then.

larshelge: Okay thanks Ben, agreed. Looking at the code I think the maintenance app also depends on that name, so let's leave it.

PR #11566, opened 2022-08-24
fix: add index.action suffix for data entry redirect [DHIS2-6705]

jbee:
### Summary
This is a workaround so that the `/dhis-web-dataentry/` URL becomes `/dhis-web-dataentry/index.action` when doing the redirct 
after login. This is important so have the data entry work as start app.

### Manual Testing
* goto System settings => Appearance and select _Data Entry_ as _Start Page_
* logout
* login again, check the data entry app is displayed correctly as the first page the user sees

### DIFF
--- a/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
+++ b/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
@@ -77,10 +77,6 @@ public class RedirectAction
                     if ( app.getShortName().equals( startModule.substring( "app:".length() ) ) )
                     {
                         redirectUrl = app.getLaunchUrl();
-                        if ( redirectUrl.endsWith( "dhis-web-dataentry/" ) )
-                        {
-                            redirectUrl += "index.action";
-                        }
                         return SUCCESS;
                     }
                 }
--- a/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
+++ b/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
@@ -77,6 +77,10 @@ public class RedirectAction
                     if ( app.getShortName().equals( startModule.substring( "app:".length() ) ) )
                     {
                         redirectUrl = app.getLaunchUrl();
+                        if ( redirectUrl.endsWith( "dhis-web-dataentry/" ) )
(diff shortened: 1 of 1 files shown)


vietnguyen (approved):

dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
```diff
@@ -77,13 +77,21 @@ public String execute()
                     if ( app.getShortName().equals( startModule.substring( "app:".length() ) ) )
                     {
                         redirectUrl = app.getLaunchUrl();
+                        if ( redirectUrl.endsWith( "dhis-web-dataentry/" ) )
```
larshelge, line 80: Could we move `dhis-web-dataentry` to a final variable and re-use?

dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
```diff
@@ -77,13 +77,21 @@ public String execute()
...
+                        {
+                            redirectUrl += "index.action";
+                        }
```
larshelge, line 83: Just curious, if the `startModule` is a Struts module/the data entry module, does it start with `app:`?

Just wondering what the if / else logic is good for. I.e. if the data entry module starts with `app:`, then we don't need to put this special fix in both branches of the if / else.

```
if ( startModule.startsWith( "app:" ) )
```

jbee: I think it does not but I could not debug the class as I don*t have a working IDE config for the web modules.

larshelge: Okay, perhaps just a good old system.out.println?

jbee: removed it from that branch and tested it still works which it does.

dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/about/action/RedirectAction.java
jbee: now there is only one occurence.

larshelge (approved):

PR #11567, opened 2022-08-24
fix: Add name to source request for ADE [DHIS2-13105]

larshelge:
Adds a mandatory `name` property and an optional `visualization` property to the source request of aggregate data exchange. Adds Javadoc to data model and service methods.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
```diff
@@ -172,6 +173,19 @@ public static List<PeriodType> getAvailablePeriodTypes()
...
 
+    /**
+     * Returns an immutable list of the names of all period types in their
+     * natural order.
+     *
+     * @return an immutable list of the names of all period types.
+     */
+    public static List<String> getAvailablePeriodTypeNames()
+    {
+        return List.copyOf( PERIOD_TYPES.stream()
```
jbee, line 184: unnecessary `copyOf`. `collect` already create a new list.

jbee (approved):
Looks good overall. Just the unnecessary copying.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
jbee: For that there is `toUnmodifyableList()` collector.

larshelge: Right, I will update it.

PR #11607, opened 2022-08-26
test: fix DataValueSetServiceIntegrationTest assertions depending on order

teleivo:
DataValueSetServiceIntegrationTest failed on Jenkins in a commit with completely unrelated changes while passing on GitHub. This suggests that the order of DataValues might have been different. [DataValueService.getAllDataValues()](https://github.com/dhis2/dhis2-core/blob/4380176ede3e773fa5a4b1dd2bc733e5f782a993/dhis-2/dhis-api/src/main/java/org/hisp/dhis/datavalue/DataValueService.java#L163-L168) does not state that it guarantees the order of DataValues.
Our assertion should therefore not expect a certain order.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetServiceIntegrationTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetServiceIntegrationTest.java
@@ -44,6 +44,7 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.time.DateUtils;
@@ -587,8 +588,9 @@ class DataValueSetServiceIntegrationTest extends IntegrationTestBase
         assertNotNull( dataValues );
         assertEquals( 3, dataValues.size() );
         assertTrue( dataValues.contains( new DataValue( deA, peA, ouA, ocDef, ocDef ) ) );
-        assertEquals( "10002", dataValues.get( 1 ).getValue() );
-        assertEquals( "10003", dataValues.get( 2 ).getValue() );
+        assertEquals( Set.of( "10001", "10002", "10003" ),
+            dataValues.stream().map( d -> d.getValue() ).collect( Collectors.toSet() ),
+            "mismatch in dataValues values" );
 
         List<Executable> audits = dataValues.stream()
             .map( dv -> ((Executable) () -> assertEquals( List.of(), dataValueAuditService.getDataValueAudits( dv ) )) )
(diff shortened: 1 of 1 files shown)


stian-sandvold (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetServiceIntegrationTest.java
```diff
@@ -587,8 +588,9 @@ void testImportDataValueSetXml()
         assertNotNull( dataValues );
         assertEquals( 3, dataValues.size() );
         assertTrue( dataValues.contains( new DataValue( deA, peA, ouA, ocDef, ocDef ) ) );
-        assertEquals( "10002", dataValues.get( 1 ).getValue() );
-        assertEquals( "10003", dataValues.get( 2 ).getValue() );
+        assertEquals( Set.of( "10001", "10002", "10003" ),
```
jbee, line 591: there is also DHIS2 `Assertions.assertContainsOnly`

larshelge (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetServiceIntegrationTest.java
larshelge: Which is also more helpful during debugging as it logs which items were expected and not presents, and which items were present but not expected.

teleivo: True, forgot about it, thank you 😄 `assertContainsOnly` fails on the first item not found which does then leave out information about whether the other items I expect to be in are in or not. With flaky tests only failing in CI that is handy to know.

We can update `assertContainsOnly` to help with that. I'll also ask again in the backend team meeting about our stance on hamcrest or not. I feel like some of the things we do is already well done by hamcrest.

teleivo: > Which is also more helpful during debugging as it logs which items were expected and not presents, and which items were present but not expected.

`assertEquals( Set.of(...), a)` gives you more information than `Assertions.assertContainsOnly`. It will allow you to figure that out what is in but should not be and what is not in and should be 😄 Granted it does not print a sentence like `Expected value %s not found in %s`.

`Assertions.assertContainsOnly` will not tell you the extra values that are not expected to be in the collection. It just tells you collection sizes don't match. It also fails on the first item that is not found. So there might be more items that are not in the collection but should be. All of which are super important to know in tests that tend to fail only on CI.

We can obviously address these issues and improve `assertContainsOnly`. Assertion libraries like hamcrest/assertj already provide assertions that take care of all of that. So maybe we should discuss this in the backend team meeting to settle on an approach 😄

PR #11620, opened 2022-08-30
feat: Improve predictions by C.O.C. DHIS2-9832

jimgrace:
See [DHIS2-9832](https://dhis2.atlassian.net/browse/DHIS2-9832).

In further discussion with PEPFAR a shortcoming was identified in how predictor disaggregation was originally coded. A key part of predictor disaggregation is to map a catOptionCombo from input data to the catOptionCombo that should be used for the predictor output data element. This was done by identifying which catCombos should be translated--by seeing if all their combinations of options (equivalent catOptionCombos) are also present in the output data element's option combo. Then for each qualifying catCombo a map was creating from that catCombo's catOptionCombos to the output catCombo's catOptionCombos.

The shortcoming of this approach is that in complex implementations, sometimes catCombos change over the years such that catOptionCombos in the data may be no longer present in the catCombo. Further, data elements may use different catCombos in different data sets. So it is entirely possible that data fetched for a data element may have catOptionCombos that do not map to the catCombo currently defined for that data element. In the first coding, such data would be ignored because it didn't fall in the pre-generated mapping from input catOptionCombo to output catOptionCombo. But as long as there are catOptionCombos in the data--for qualifying data elements--which can still be mapped to output catOptionCombos, PEPFAR would like them to be mapped, as presumably other installations would as well.

This PR refactors the code as follows: It still selects for disaggregation the data elements having catCombos that map to the output data element catCombo, as before. But it no longer pre-generates maps from input catOptionCombo to output catOptionCombo. Instead, it pre-generates a map from sorted option uids to output catOptionCombo. When encountering data for data elements that can be disaggregated, it now finds the sorted option uids for the data's catOptionCombo, and uses the pre-generated map to find the output catOptionCombo. If there is no match for an output catOptionCombo, it adds the data element operand (data element + catOptionCombo) to a list of those for which no match was found. At the end of the prediction, a warning message is issued if there are such data element operands for which a match was tried but not found. (Depending on the use case this may indicate a configuration error--which is why the warning is given--or it may be expected behaviour.)


maikelarabori (approved):
I'm not familiar with Predictors, but the code and tests look good :+1:

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
```diff
@@ -269,4 +270,51 @@ public static <T> boolean addIfNotNull( Collection<T> collection, T object )
...
+     * @param k7 the seventh mapping's key
+     * @param v7 the seventh mapping's value
+     * @param k8 the eighth mapping's key
+     * @param v8 the eighth mapping's value
+     * @param k9 the ninth mapping's key
+     * @param v9 the ninth mapping's value
+     * @param k10 the tenth mapping's key
+     * @param v10 the tenth mapping's value
+     * @param k11 the eveventh mapping's key
+     * @param v11 the eleventh mapping's value
```
jbee, line 298: I think these don't provide much - should be pretty clear without

jbee (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
jimgrace: Agreed. See [pull/11717](https://github.com/dhis2/dhis2-core/pull/11717). (I merged this PR to be sure it made it before the 2.39 feature freeze.)

PR #11623, opened 2022-08-31
feat: user lookup restricted to capture units [DHIS2-12586]

jbee:
### Summary
Adds `orgUnitBoundary` parameter to both `/users` and `/userLoopup` endpoints.
The parameter accepts any of the `UserOrgUnitType` enum values and restricts the search to the users that share an organisation unit with the current user for the selected boundary.

The existing parameter `&userOrgUnits=true` would be identical to `&orgUnitBoundary=DATA_CAPTURE` and is kept for backwards compatibility. If `userOrgUnits` is true it takes precedence over any `orgUnitBoundary` provided additionally.

### Automatic Testing
A new controller test scenario was added.

### Manual Testing
* create a user and set its capture organisation unit to a specific unit
* create another user and set its capture organisation unit to the same unit
* login as one of those users
* use `/userLookup` search to find the users sharing a capture organisation unit using `query` and `orgUnitBoundary=DATA_CAPTURE` parameters
* check other users that otherwise would match the `query` are not found as they do not share the capture organisation unit
* repeat the test for `/users` and other boundary types

### Documentation
https://github.com/dhis2/dhis2-docs/pull/1036

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
@@ -163,21 +163,6 @@ public class UserQueryParams
         return this;
     }
 
-    public boolean hasOrganisationUnits()
-    {
-        return !organisationUnits.isEmpty() && orgUnitBoundary == UserOrgUnitType.DATA_CAPTURE;
-    }
-
-    public boolean hasDataViewOrganisationUnits()
-    {
-        return !organisationUnits.isEmpty() && orgUnitBoundary == UserOrgUnitType.DATA_OUTPUT;
-    }
-
-    public boolean hasTeiSearchOrganisationUnits()
-    {
-        return !organisationUnits.isEmpty() && orgUnitBoundary == UserOrgUnitType.TEI_SEARCH;
-    }
-
     public UserQueryParams setOrganisationUnits( Set<OrganisationUnit> units )
     {
         this.organisationUnits = units;
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
@@ -33,21 +33,29 @@ import java.util.Set;
 
 import lombok.Getter;
 
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+import lombok.experimental.Accessors;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
-
-import com.google.common.base.MoreObjects;
 
 /**
  * @author Lars Helge Overland
  */
 @Getter
+@Setter
+@Accessors(chain = true)
+@ToString(onlyExplicitlyIncluded = true)
+@NoArgsConstructor
 public class UserQueryParams
 {
     /**
      * The user query string.
      */
+    @ToString.Include
     private String query;
 
+    @ToString.Include
     private String phoneNumber;
 
     /**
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -330,9 +331,21 @@ private void handleUserQueryParams( UserQueryParams params )
...
         }
 
-        if ( params.isUserOrgUnits() && params.hasUser() )
+        if ( params.hasUser() )
         {
-            params.setOrganisationUnits( params.getUser().getOrganisationUnits() );
+            UserOrgUnitType orgUnitBoundary = params.getOrgUnitBoundary();
+            if ( params.isUserOrgUnits() || orgUnitBoundary == UserOrgUnitType.DATA_CAPTURE )
+            {
+                params.setOrganisationUnits( params.getUser().getOrganisationUnits() );
```
larshelge, line 339: I am thinking we could use just one collection, the one called `organisationUnits`, for all these boundaries. The type of org units can be inferred from the `UserOrgUnitType` in the SQL query. I might be wrong.

jbee: yes, I think that should work. I'll update.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
```diff
@@ -101,241 +125,91 @@
...
-        return this;
+        return !organisationUnits.isEmpty() && orgUnitBoundary == UserOrgUnitType.TEI_SEARCH;
     }
 
-    public UserQueryParams setDisjointRoles( boolean disjointRoles )
+    public UserQueryParams setOrganisationUnits( Set<OrganisationUnit> units )
     {
-        this.disjointRoles = disjointRoles;
+        this.organisationUnits = units;
+        if ( orgUnitBoundary == null )
```
jbee, line 184: Note that this setter is now a bit ambiguous. In the old semantics this was explicitly filtering data capture. In the new semantics this setter is potentially used for any of the boundaries so we just set the type of boundary if it is still null. This is so this can be used in combination with `setOrgUnitBoundary` in any order.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
```diff
@@ -245,7 +245,7 @@ private Query<?> getUserQuery( UserQueryParams params, List<String> orders, bool
             {
                 hql += hlp.whereAnd() + " (";
 
-                for ( OrganisationUnit ou : params.getDataViewOrganisationUnits() )
+                for ( OrganisationUnit ou : params.getOrganisationUnits() )
```
larshelge, line 248: This is better. Sorry for harping on this. I think it is actually possible to do away with 2 of the 3 if code blocks for org unit filtering entirely. The only piece that is now different across the 3 blocks is the join field alias. I.e. we have:

`left join u.organisationUnits ou `
`left join u.dataViewOrganisationUnits dwou`
`left join u.teiSearchOrganisationUnits tsou`
 
 This can be parameterized to:
 
 `left join u.${org_unit_field} ou`
 
 and `${org_unit_field}` can be inferred from the user org unit type.
 
 The "query parameters" part further down in the method must be updated too. 
 
 That way we collapse 3 blocks of if/for into 1 it could seem.

larshelge (approved):

PR #11625, opened 2022-08-31
fix: Show attributes in csv response [DHIS2-13621] [2.38]

muilpp:
This PR aims to fix the issue of writing tracked entity attributes into a csv file.
The attributes were not written into the file because they are not simple properties, they are complex so they were not included.
The PR pushes the fix to include them from now on, plus some unit tests.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
+++ b/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
@@ -92,7 +92,7 @@ public class CsvNodeSerializer extends AbstractNodeSerializer
                     {
                         schemaBuilder.addColumn( property.getName() );
                     }
-                    if ( property.getName().equalsIgnoreCase( "attributes" ) )
+                    if ( property.getName().equals( "attributes" ) )
                     {
                         property.getChildren().stream().findFirst()
                             .ifPresent( fc -> fc.getChildren().forEach( p -> schemaBuilder.addColumn( p.getName() ) ) );
@@ -140,7 +140,7 @@ public class CsvNodeSerializer extends AbstractNodeSerializer
                 simpleNodeList.add( (SimpleNode) property );
             }
 
-            if ( property.getName().equalsIgnoreCase( "attributes" ) )
+            if ( property.getName().equals( "attributes" ) )
             {
                 hasAttributes = true;
                 writeComplexNode( property.getChildren(), simpleNodeList, csvGenerator );
--- a/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
+++ b/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
@@ -88,12 +88,14 @@ public class CsvNodeSerializer extends AbstractNodeSerializer
 
                 for ( Node property : node.getChildren() )
                 {
-                    if ( property.isSimple())
+                    if ( property.isSimple() )
                     {
                         schemaBuilder.addColumn( property.getName() );
                     }
-                    if (property.getName().equalsIgnoreCase("attributes")) {
-                        property.getChildren().stream().findFirst().ifPresent(fc -> fc.getChildren().forEach(p -> schemaBuilder.addColumn(p.getName())));
+                    if ( property.getName().equalsIgnoreCase( "attributes" ) )
+                    {
+                        property.getChildren().stream().findFirst()
+                            .ifPresent( fc -> fc.getChildren().forEach( p -> schemaBuilder.addColumn( p.getName() ) ) );
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


larshelge:
Please add a proper semantic PR description.

dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/node/serializers/CsvNodeSerializer.java
```diff
@@ -91,6 +92,11 @@ protected void startSerialize( RootNode rootNode, OutputStream outputStream )
                     {
                         schemaBuilder.addColumn( property.getName() );
                     }
+                    if ( property.getName().equalsIgnoreCase( "attributes" ) )
```
mortenoh, line 95: Don't need to ignore case here, it should always be lowercase.

mortenoh (approved):

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+
+    private RootNode createCollectionWithSimpleNodes()
+    {
+        CollectionNode collectionRootNode = new CollectionNode( "Complex node" );
+        CollectionNode firstCollectionNode = new CollectionNode( "First collection child" );
+        CollectionNode secondCollectionNode = new CollectionNode( "Second collection child" );
+
+        SimpleNode firstSimpleNode = new SimpleNode( "First simple node", "First value simple child" );
```
teleivo, line 135: I would inline vars like `firstSimpleNode` since they are only used once. This makes me have to read/remember less IMHO 😄

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
```diff
@@ -0,0 +1,193 @@
...
+        SimpleNode secondSimpleNode = new SimpleNode( "Second simple node", "Second value simple child" );
+        SimpleNode thirdSimpleNode = new SimpleNode( "Third simple node", "Third value simple child" );
+
+        secondCollectionNode.addChild( firstSimpleNode );
```
teleivo, line 139: I am unsure if I read correctly, but is `addChild` returning the argument? It would be nice to be able to write this more fluently like

```java
CollectionNode firstCollectionNode = (new CollectionNode( "First collection child" ))
.addChild( new SimpleNode( "First simple node", "First value simple child" ) )
.addChild( new SimpleNode( "Second simple node", "Second value simple child" ) )
.addChild( new SimpleNode( "Third simple node", "Third value simple child" ) );
```

What I like personally is looking at the test setup and assertions and seeing why some values are expected at one glance. So preferably in the test. This is a bit hard with `createCollectionWithSimpleNodes()` right now as I keep jumping from here up to the test to understand why we assert on certain values 😄

teleivo (commented):
I am not familiar with the CsvNodeSerializer. I would need more time to review the implementation meaningfully 😬
 
I have some suggestions for the test though 😄

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
```diff
@@ -0,0 +1,193 @@
...
+class CsvNodeSerializerTest
+{
+
+    private final CsvNodeSerializer csvNodeSerializer = new CsvNodeSerializer();
+
+    @Test
+    void CsvFileIsWrittenWhenOnlySimpleNodesAreProvided()
+        throws Exception
+    {
+        csvNodeSerializer.serialize( createCollectionWithSimpleNodes(), new FileOutputStream( "output.csv" ) );
```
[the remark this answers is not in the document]
muilpp: yep, totally right, this way it's less error prone

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
muilpp: this one I can clean up a bit, but the way I understood the Node system works, I'm not sure it gives a lot of room for improvement

muilpp: actually, I just tried but it's not possible because what it returns is the child you add. 
In this case it needs to be a simpleNode, and such a node can't have children

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
```diff
@@ -30,9 +30,6 @@
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.*;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
 
 import lombok.extern.slf4j.Slf4j;
```
teleivo, line 34: I think there is no need for logging now that there is no cleanup that can go wrong 😄

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
```diff
@@ -56,96 +52,81 @@
...
+        String[] resultArray = outputStream.toString().split( "\n" );
 
-        assertEquals( 2, values.size() );
-        assertEquals( 3, values.get( 0 ).size() );
-        assertEquals( 3, values.get( 1 ).size() );
-        assertEquals( "\"First simple node\"", values.get( 0 ).get( 0 ) );
-        assertEquals( "\"Second value simple child\"", values.get( 1 ).get( 1 ) );
-        assertEquals( "\"Third simple node\"", values.get( 0 ).get( 2 ) );
+        assertEquals( 2, resultArray.length );
+        assertEquals( "\"First simple node\",\"Second simple node\",\"Third simple node\"", resultArray[0] );
```
teleivo, line 61: Out of curiosity 😋 I have not used it yet myself but I wonder if https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertLinesMatch(java.util.List,java.util.List) or the variants for streams are a good fit since you are asserting lines in the output stream. Did you try using it?

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
muilpp: true!

dhis-2/dhis-services/dhis-service-node/src/test/java/org/hisp/dhis/node/serializer/CsvNodeSerializerTest.java
muilpp: I didn't try it, I guess it would work if I transform the output stream into a list first. 
Regarding the method with the stream signature, it supports monadic streams, so I guess it would need to be transformed into a list as well.

enricocolasante (approved):

PR #11634, opened 2022-09-01
feat: test setup helper

jbee:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/datavalue/DataValueServiceIntegrityTest.java
```diff
@@ -54,88 +43,39 @@
...
-
-    private Category categoryA;
-
-    private Category categoryB;
+    private MetadataSetupServiceExecutor setupExecutor;
 
-    private CategoryOptionCombo categoryOptionComboAB;
-
-    private CategoryOptionCombo categoryOptionComboAC;
+    private final MetadataSetup setup = new MetadataSetup();
```
larshelge, line 52: Should perhaps the `MetadataSetup` be freshly instantiated for each test to avoid pollution in between tests?

jbee: This should not be needed here. The object is only changed by the `setUpTest` method. Tests are just reading its state.

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/setup/MetadataSetupServiceExecutor.java
```diff
@@ -0,0 +1,338 @@
...
+    private Attribute createAttribute( AttributeSetup setup )
+    {
+        Attribute obj = new Attribute( setup.getName(),
+            orElse( setup, AttributeSetup::getValueType, ValueType.TEXT ) );
+        obj.setUid( setup.getUid() );
+        obj.setUnique( setup.isUnique() );
+        obj.setMandatory( setup.isMandatory() );
+        setup.getObjectTypes().forEach( type -> obj.setAttribute( type, true ) );
+        obj.setAutoFields();
+        attributeService.addAttribute( obj );
```
larshelge, line 175: Use `IdentifiableObjectManager` where possible to reduce the number of dependencies?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/datavalue/DataValueServiceTest.java
```diff
@@ -104,27 +112,35 @@ class DataValueServiceTest extends TransactionalIntegrationTest
...
+        setup.addDataElement( "C", withDefaultSetup );
+        setup.addDataElement( "D", withDefaultSetup );
+        setup.addPeriod( "2022-09-05", withDefaultSetup );
+        setup.addPeriod( "2022-09-06", withDefaultSetup );
+        setup.addPeriod( "2022-09-07", withDefaultSetup );
+        setup.addPeriod( "2022-09-27", withDefaultSetup );
+        setup.addOrganisationUnit( "A", withDefaultSetup );
+        setup.addOrganisationUnit( "B", withDefaultSetup );
+        setup.addOrganisationUnit( "C", withDefaultSetup );
+        setup.addOrganisationUnit( "D", withDefaultSetup );
```
larshelge, line 126: Introduce a new method overload to avoid specifying `withDefaultSetup` everywhere?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/datavalue/DataValueServiceTest.java
```diff
@@ -104,27 +112,35 @@ class DataValueServiceTest extends TransactionalIntegrationTest
...
-        peX = createPeriod( getDay( 27 ), getDay( 28 ) );
-        ouA = createOrganisationUnit( 'A' );
-        ouB = createOrganisationUnit( 'B' );
-        ouC = createOrganisationUnit( 'C' );
-        ouD = createOrganisationUnit( 'D' );
-        organisationUnitService.addOrganisationUnit( ouA );
-        organisationUnitService.addOrganisationUnit( ouB );
-        organisationUnitService.addOrganisationUnit( ouC );
-        organisationUnitService.addOrganisationUnit( ouD );
+        setup.addDataElement( "A", withDefaultSetup );
```
larshelge, line 115: Consider offering a more opinionated service/interface which automatically creates a set of metadata, where you can specify just the number of objects you want and you get what you get. In many cases it does not matter so much exactly what the metadata is called, you just need a reasonable set of metadata as basis for tests.

stian-sandvold (approved):

PR #11635, opened 2022-09-01
fix: import user with legacy UserCredentials [DHIS2-13675] [2.39]

vietnguyen:
- Follow up on https://github.com/dhis2/dhis2-core/pull/11627
- This PR fix an issue of importing User with old `UserCredentials`. 
The error occurs when the `UserCredentials`  has references to persisted `UserRole` which are not included in the payload.
The `PreheatService` will not pick the transient property `UserCredentialsRaw`, hence the `UserRole` will also not being added to the preheat map.
- Added integration test

### DIFF
--- a/dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/jsonpatch/operations/RemoveByIdOperation.java
+++ b/dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/jsonpatch/operations/RemoveByIdOperation.java
@@ -66,7 +66,7 @@ public class RemoveByIdOperation extends JsonPatchOperation
 
         if ( !nodePathExists( node ) )
         {
-            throw new JsonPatchException( String.format( "Invalid path %s", path ) );
+            return node;
         }
 
         final JsonNode parentNode = node.at( path );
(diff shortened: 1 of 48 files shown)


larshelge (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/jsonpatch/operations/RemoveByIdOperation.java
```diff
@@ -66,7 +66,7 @@ public JsonNode apply( JsonNode node )
 
         if ( !nodePathExists( node ) )
         {
-            return node;
+            throw new JsonPatchException( String.format( "Invalid path %s", path ) );
```
mortenoh, line 69: Maybe add some tests to both Remove operations? Just so we our covered if anyone changes this behavior again

netroms (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/jsonpatch/operations/RemoveByIdOperation.java
vietnguyen: @mortenoh sorry, this change supposed to be in another PR, reverted now.

PR #11665, opened 2022-09-05
feat: Enforce 2FA setup on next login [DHIS2-13333]

netroms:
# Summary:
This adds support for: 
* Admins to set up mandatory self enrolment of 2FA for users.
As an admin, you can now add what is called a role restriction to a user role, when this restriction is R_ENABLE_2FA, any user that is a member will automatically be redirected to 2FA setup after login, if the user not already has enrolled and setup 2FA. 
* Admins to disable/reset 2FA for any user.
* Generating QR code on the server itself, instead of relying on an external third party service.

The concept of role restrictions is realized with a new one-to-many table called “userrolerestrictions”, in similar fashion as the authorities with the “userroleauthorities” table.
Right now, only one restriction exists, the “R_ENABLE_2FA”. 
This concept is meant to be flexible and used for any situations when you want to enforce some criteria on all members of a role.

### 2FA Enrolment:
1. As an admin/superuser, create a new role.
2. Use the API or edit directly the DB to add the restriction “R_ENABLE_2FA” to the new role. 
3. As an admin/superuser, create a new user and assign the new role with the restriction.
4. Log in as the new user and observe that you are redirected to 2FA setup on login.
5. Setup 2FA with your phone and enter the 2FA code to complete the setup.
6. Log out and try to log in again and observe that the 2FA code field is show.
7. Enter your 2FA code and log in as normal.

### Admin disabling/reset of user's 2FA:
1. Enrol a user with 2FA as described in the previous steps
2. As superuser call the URL: `POST http://localhost:8080/api/users/USER_ID/twoFA/disabled`

## Automatic tests
* TwoFactorControllerTest
* UserControllerTest
 
## Missing front-end parts:
* Refactor user/account UI, enable/disable 2FA. Needs to support new QR flow.
* Role restriction on role UI
* User enrolment/setup page, after login, needs proper styling.
* Disable/reset 2FA on user admin UI.

JIRA: [DHIS2-13333](https://dhis2.atlassian.net/browse/DHIS2-13333)

[DHIS2-13333]: https://dhis2.atlassian.net/browse/DHIS2-13333


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_24__Create_table_userrolerestrictions.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE userrolerestrictions
```
jbee, line 1: Should we do a drop table before we try to create it?

jbee (approved):
That was a big change but looks good!

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.39/V2_39_24__Create_table_userrolerestrictions.sql
vietnguyen: `create table if not exists`

vietnguyen (approved):

vietnguyen (approved):

PR #11667, opened 2022-09-06
fix: NPE in OptionObjectBundleHook.preUpdate

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-13640
- According to the bug reported, user got NPE when trying to add `Option` to an `OptionSet`.
- I couldn't reproduce it but the issue can be fixed in backend with a `null` check.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
@@ -35,6 +35,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.hisp.dhis.common.BaseIdentifiableObject;
 import org.hisp.dhis.common.DxfNamespaces;
 import org.hisp.dhis.common.IdScheme;
@@ -120,6 +121,16 @@ public class OptionSet
     public void removeAllOptions()
     {
         options.clear();
+    }
+
+    public void removeOption( Option option )
+    {
+        if ( options == null || CollectionUtils.isEmpty( options ) )
+        {
+            return;
+        }
+
+        options.remove( option );
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
```diff
@@ -122,6 +123,16 @@ public void removeAllOptions()
         options.clear();
     }
 
+    public void removeOption( Option option )
+    {
+        if ( options == null || CollectionUtils.isEmpty( options ) )
```
jbee, line 128: Main reason to use a static `isEmpty` is that it allows for `null` so it does not need a `null` check before

vietnguyen: fixed

mortenoh (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
```diff
@@ -122,6 +123,14 @@ public void removeAllOptions()
         options.clear();
     }
 
+    public void removeOption( Option option )
```
larshelge, line 126: This is a little bit worrying as we might be hiding underlying issues. Do we know why the options list is null here?

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
vietnguyen: @larshelge  I talked with Pamod yesterday. They can't reproduce the issue after restarting the sever. All OptionSets are now working normally on their instance. 
My theory is that after executing the clean_up_option_sortOrder script, it reset all the OptionSet.options references in the cache. But I couldn't replicate this behaviour on my local. So not really sure what happened on their instance.
I will update the clean up script document to add a final step for restarting server.

PR #11669, opened 2022-09-06
fix: Use inheritance instead of partitions in analytics export [DHIS2-13429] [2.37]

maikelarabori:
This PR aims to restore the expected Analytics export and table _partitioning_ using `inheritance` (as release 2.36).

**Overview**
Since release 2.37, some behaviours were not correctly covered when we introduced partition tables.

Pepfar identified and raised those issues in three JIRA tickets: DHIS2-13429, DHIS2-DHIS2-13478 and DHIS2-13479. In summary, those tickets describe the following problems.

**Problems**
1) The table `analytics_0` never gets cleaned after a full export. It causes data duplication when the application hits the query below, in `JdbcAnalyticsManager.getFromSourceClause`.
```
 select ap.*
      from analytics_0 as ap
      union all
      select ap.*
      from analytics_2019 as ap
      union all
      select ap.*
      from analytics_2020 as ap
```

2) After running the Analytics job (export) for the `last_years=x` all partitions are removed and replaced by the partitions related to `last_years` only. It causes old generated data to be lost and obliges the user to run the full export to recover all data, which takes much longer.

3) When the Analytics job is triggered using `last_years=0`, all years/partitions are removed. This obliges the user to run a full export so they can recover all Analytics tables.

**Resolution**
The changes present on this PR should fix those issues and mimic the same behaviour before version 2.37, per the items above.
Also, before this fix, it's important to mention that the table `_0`could never be added as a partition of the main table. This happened because the table `_0` could contain rows for any specific year, making it impossible to map this partition with a single year. Even if we could, year `0` would not make much sense. For this reason, we had to restore the previous behaviour, where we used `table inheritance` instead of `partitioned tables`.

**Scenarios of tests run**
The following testing _scenarios_ **-> results** were run to check if the process is working as expected:

1) _Run the analytics job specifying `lastYears=0`, ie: `localhost:8080/dhis/api/resourceTables/analytics?lastYears=0&skipEvents=false&skipEnrollment=false`:_
**-> It creates and populates the analytics `_0` table and keeps all existing tables in place.**

2) _Run continuous analytics job:_
**-> It creates the analytics `_0` table and keeps all existing tables in place.**

3) _Manually delete the table `_0`:_
**-> The queries to analytics still work fine.**

4) _Full export all tables, all years:_
**-> All tables for all years are generated and populated. The tables `_0` are removed.**

5) _Partial export (2 years):_
**-> Generates tables only for specific the last 2 years. The tables `_0` are not removed.**

6) _Export table setting numbers of years to include  to `1`:_
**-> Generates tables only for the current year. The tables `_0` are not removed.**

7) _Running full export on top of `partition tables`:_
**-> Process removes all existing tables and recreates them using inheritance.**

8) _Partial export on top of `partition tables`:_
**-> Process removes all existing tables and recreates them using inheritance.**

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/AbstractJdbcTableManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/AbstractJdbcTableManager.java
@@ -66,6 +66,7 @@ import org.hisp.dhis.commons.util.TextUtils;
 import org.hisp.dhis.dataapproval.DataApprovalLevelService;
 import org.hisp.dhis.jdbc.StatementBuilder;
 import org.hisp.dhis.organisationunit.OrganisationUnitGroupSet;
+import org.hisp.dhis.organisationunit.OrganisationUnitLevel;
 import org.hisp.dhis.organisationunit.OrganisationUnitService;
 import org.hisp.dhis.period.PeriodType;
 import org.hisp.dhis.resourcetable.ResourceTableService;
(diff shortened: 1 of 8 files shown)


d-bernat (approved):

Philip-Larsen-Donnelly (approved):
Approving as approved on 2.37dev branch

PR #11673, opened 2022-09-06
feat: Add data entry data set completion endpoint [TECH-1264]

larshelge:
Adds a new API endpoint for completion of data sets. API endpoint:

```
POST /api/dataEntry/dataSetCompletion
```

Request payload format:

```
{
  "dataSet": "<id>",
  "period": "<iso-period>",
  "orgUnit": "<id>",
  "attribute": {
    "combo": "<id>",
    "options": ["<id>", "<id>"]
  }
}
```

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/CompleteDataSetRegistration.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/CompleteDataSetRegistration.java
@@ -62,9 +62,9 @@ public class CompleteDataSetRegistration
 
     private CategoryOptionCombo attributeOptionCombo;
 
-    private Date date; // TODO rename to created
+    private Date date; // TO-DO rename to created
 
-    private String storedBy; // TODO rename to createdBy
+    private String storedBy; // TO-DO rename to createdBy
 
     private Date lastUpdated;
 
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/CompleteDataSetRegistration.java
```diff
@@ -62,9 +63,9 @@
 
     private CategoryOptionCombo attributeOptionCombo;
 
-    private Date date; // TODO rename to created
+    private Date date; // TO-DO rename to created
 
-    private String storedBy;
+    private String storedBy; // TO-DO rename to createdBy
```
jbee, line 68: Is there a reason these TODOs cannot be done in this PR?

jbee (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/CompleteDataSetRegistration.java
larshelge: Mainly that it is a breaking API change and will break existing data integrations, including the DHIS 2 data sync feature. So we will have to do it in a backwards-compatible way, which is a bigger and independent task.

PR #11676, opened 2022-09-06
feat: add backend function for generating QR code [DHIS2-13659]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-13659
- Add a new endpoint ` /api/2fa/qrCode` which returns an "image/png"
- After this is merged, need to integrate it with current UI and remove the deprecated functions/endpoint.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -170,6 +170,9 @@
     E3022( "User already has two factor authentication enabled, disable 2FA before you create a new QR code" ),
     E3023( "Invalid 2FA code" ),
     E3024( "Not allowed to disable 2FA" ),
+    E3025( "User must have a secret" ),
+    E3026( "Couldn't generate QR code" ),
```
larshelge, line 174: Change to "Could not".

larshelge (approved):

larshelge:
Nice work!

mortenoh (approved):

PR #11679, opened 2022-09-07
feat: Indicator .yearToDate() DHIS2-13320

jimgrace:
See [DHIS2-13320](https://dhis2.atlassian.net/browse/DHIS2-13320), especially the `Expression syntax` section of the May 24, 2022 comment.

This PR implements the following expression syntax for indicators:

- yearToDate() modifier to add any previous values in the calendar year

- [periodInYear] returns the number of the period in the calendar year

- [yearlyPeriodCount] returns the total number of periods of this type in the calendar year

The two classes that do most of the heavy lifting are:

### DataHandler

The approach is similar to periodOffset. Before the indicator database query, year-to-date processing adds periods if needed to fetch the values from previous periods within the same calendar year(s) as any requested periods. After the database query, previous calendar year data is added to the results to be returned.

> The alternative approach would have been to do the year-to-date summing in the SQL query itself. This might have been efficient in a way, but might have resulted in more queries and would not allow the reuse of data. If an analytics query returns year-to-date values for a sequence of months, the same monthly figures can be added to several consecutive year-to-date values.

`getAggregatedDataValueMap` now gets the Grid of values. The code that assembles the ValueMap from the Grid has been separated into the following method `getAggregatedValueMapFromGrid` to reduce method complexity and size.

The `Grid rows loop` now accounts for year-to-date values. As it makes a single pass through the grid rows, any values that contribute to one or more year-to-date values are added to a growing collection of `yearToDateRows`. Values not contributing to year-to-date values are just added to the ValueMap within the loop. After the Grid rows loop is complete, then the `yearToDateRows` are added to the ValueMap.

Processing of the non-year-to-date values has been simplified in the Grid rows loop by calling the `addRowToValueMap` method which adjusts the period if needed for a periodOffset and then adds the row to the ValueMap. Generating the key and value and adding it to the map is now just done in one place here, instead of in two places as was previously done: (1) within the grid row loop for non-period values, and (2) within `addItemBasedOnPeriodOffset` for period values. The same `addRowToValueMap` is also used to add the accumulated year-to-date values after looping through all the Grid rows.

The `getAdjustedRow` method has much of the same logic that used to be in `addItemBasedOnPeriodOffset`, and also has the `hasPeriod` test that had been in the Grid rows loop. This method just adjusts the row if needed for period offset, but doesn't add the values to the Grid; that's now done only in `addRowToValueMap`.

### PeriodOffsetUtils

Most of the year-to-date specific logic has been added here. `addShiftedPeriods` now adds any periods needed for year-to-date values as well as for periodOffset values. `buildYearToDateRows` and `addYearToDateRow` do the job of accumulating year-to-date values during the Grid rows loop. `yearToDatePeriods` generates a list of periods in the calendar year that will be needed for year-to-date values. (Weekly and BiWeekly periods that span reporting years will be included in one year or the other depending on how many days they have in each year.)

Note that periodOffset and yearToDate can be used in combination. For example if the period is April and periodOffset is -1, it will fetch year-to-date data from January through March and report it in April.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemPeriodInYear.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemPeriodInYear.java
@@ -29,7 +29,6 @@ package org.hisp.dhis.expression.dataitem;
 
 import static java.lang.Integer.parseInt;
 import static java.util.Calendar.DAY_OF_YEAR;
-import static java.util.Calendar.MONTH;
 import static org.hisp.dhis.parser.expression.ParserUtils.DOUBLE_VALUE_IF_NULL;
 import static org.hisp.dhis.parser.expression.antlr.ExpressionParser.ExprContext;
 
@@ -40,10 +39,12 @@ import java.util.regex.Pattern;
 
 import org.hisp.dhis.parser.expression.CommonExpressionVisitor;
 import org.hisp.dhis.parser.expression.ExpressionItem;
+import org.hisp.dhis.period.BiMonthlyPeriodType;
 import org.hisp.dhis.period.DailyPeriodType;
 import org.hisp.dhis.period.MonthlyPeriodType;
 import org.hisp.dhis.period.Period;
 import org.hisp.dhis.period.PeriodType;
+import org.hisp.dhis.period.YearlyPeriodType;
 
 /**
  * Expression item [periodInYear]
@@ -78,11 +79,12 @@ public class ItemPeriodInYear
         {
             return dayOfYear( period );
         }
-        else if ( periodType instanceof MonthlyPeriodType )
+        else if ( periodType instanceof MonthlyPeriodType
+            || periodType instanceof BiMonthlyPeriodType )
         {
-            return monthOfYear( period );
+            return monthOrBiMonthOfYear( period );
         }
-        else if ( periodType.getFrequencyOrder() >= 365 )
+        else if ( periodType.getFrequencyOrder() >= YearlyPeriodType.FREQUENCY_ORDER )
         {
             return 1.0;
         }
@@ -94,25 +96,31 @@ public class ItemPeriodInYear
     // Supportive methods
     // -------------------------------------------------------------------------
 
+    /**
+     * Finds the day of the year for a daily period type.
+     */
     private double dayOfYear( Period period )
     {
-        return getCal( period ).get( DAY_OF_YEAR );
-    }
-
-    private double monthOfYear( Period period )
-    {
-        return 1.0 + getCal( period ).get( MONTH );
-    }
-
-    private Calendar getCal( Period period )
-    {
         Calendar cal = Calendar.getInstance();
-
         cal.setTime( period.getStartDate() );
 
-        return cal;
+        return cal.get( DAY_OF_YEAR );
     }
 
+    /**
+     * For Monthly yyyyMM and BiMonthly yyyyMMB types, the period number is in
+     * the fourth and fifth characters of the period's ISO Date.
+     */
+    private double monthOrBiMonthOfYear( Period period )
+    {
+        return parseInt( period.getIsoDate().substring( 4, 6 ) );
+    }
+
+    /**
+     * For many period types, the period number can be found in the trailing
+     * digits of the period's ISO Date, for example Weekly yyyWn, Quarterly
+     * yyyyQn, SixMonthly yyyySn, etc.
+     */
     private double trailingDigits( Period period )
     {
         Matcher m = TRAILING_DIGITS.matcher( period.getIsoDate() );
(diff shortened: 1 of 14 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemYearlyPeriodCount.java
```diff
@@ -0,0 +1,118 @@
...
+        return 1.0;
+    }
+
+    // -------------------------------------------------------------------------
+    // Supportive methods
+    // -------------------------------------------------------------------------
+
+    private double weeksInYear( Period period )
+    {
+        return testLastPeriodInYear( period, 52 );
```
maikelarabori, line 96: I think we could have some constants for some of the integers used in this class.
For someone reading the code, it's difficult to understand what `28` means. Why not `29` or `30`for example?
`12` could be assigned to a constant named `DECEMBER`, for example.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemYearlyPeriodCount.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private double biWeeksInYear( Period period )
+    {
+        return testLastPeriodInYear( period, 26 );
+    }
+
+    private double testLastPeriodInYear( Period period, int count )
```
maikelarabori, line 104: Would be nice to have a Javadoc for this method.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemYearlyPeriodCount.java
```diff
@@ -0,0 +1,118 @@
...
+
+        if ( periodType instanceof WeeklyAbstractPeriodType )
+        {
+            return weeksInYear( period );
+        }
+        else if ( periodType instanceof BiWeeklyAbstractPeriodType )
+        {
+            return biWeeksInYear( period );
+        }
+        else if ( periodType instanceof CalendarPeriodType && periodType.getFrequencyOrder() < 365 )
```
maikelarabori, line 82: For years that have 366 days, do we need some specific checks?

maikelarabori (approved):
Looks good @jimgrace, just left a few minor comments.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemPeriodInYear.java
```diff
@@ -0,0 +1,127 @@
...
+
+        if ( periodType instanceof DailyPeriodType )
+        {
+            return dayOfYear( period );
+        }
+        else if ( periodType instanceof MonthlyPeriodType )
+        {
+            return monthOfYear( period );
+        }
+        else if ( periodType.getFrequencyOrder() >= 365 )
```
enricocolasante, line 85: Same here.
Is it going to be correct for leap years?

enricocolasante (approved):
Just double check the leap year problem

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemYearlyPeriodCount.java
jimgrace: Excellent comments, thanks! I've put all the magic numbers into constants and/or JavaDoc, and added to the JavaDoc so things will be easier for the next readers.

Also, as I was writing the JavaDoc I noticed that I'd not parsed `BiMonthly` periods correctly. I've fixed that now and added a test for it. I guess that's one of the benefits of writing JavaDoc!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemYearlyPeriodCount.java
jimgrace: Yes, thanks. The period count for 366 daily periods is handled below within `countPeriodsWithinThisYear`. I've replaced `< 365` to `< YearlyPeriodType.FREQUENCY_ORDER` so hopefully it's more clear that I'm filtering for period types that are shorter than yearly periods types.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/dataitem/ItemPeriodInYear.java
jimgrace: Good question. Same answer. By changing `365` to `YearlyPeriodType.FREQUENCY_ORDER` hopefully it will be clearer that we're just returning 1 for any period types that are yearly or longer.

Leap years only affect the numbering of daily periods. They are handled by `cal.get( DAY_OF_YEAR )` and tested.

PR #11691, opened 2022-09-07
fix: only create data value audit if value changes [DHIS2-13705]

jbee:
Only create a data value update audit when the value has changed.

From the report of the ticket I also deduced that there are pure followup updates which should mean we only want to skip the update if the followup has not changed.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
@@ -935,7 +935,8 @@ public class DefaultDataValueSetService
         }
         if ( !internalValue.isDeleted()
             && Objects.equals( existingValue.getValue(), internalValue.getValue() )
-            && Objects.equals( existingValue.getComment(), internalValue.getComment() ) )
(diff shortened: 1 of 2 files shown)


larshelge (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -935,15 +935,16 @@ private void saveDataValueUpdate( ImportContext context, ImportCount importCount
         }
         if ( !internalValue.isDeleted()
             && Objects.equals( existingValue.getValue(), internalValue.getValue() )
-            && Objects.equals( existingValue.getComment(), internalValue.getComment() ) )
+            && Objects.equals( existingValue.getComment(), internalValue.getComment() )
+            && existingValue.isFollowup() == internalValue.isFollowup() )
```
larshelge, line 939: Good catch here, agreed. Now should we not due == checks on object Booleans and instead use equals (taking null into account)? It will work in practice but I recall that regression around boolean system settings and Redis.

jbee: These are `boolean`s

mortenoh (approved):

PR #11705, opened 2022-09-08
fix: user settings value by key set header accept all [DHIS2-13711]

jbee:
Sets the `Content-Type` header while any `Accept` is mapped to the method.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
@@ -119,7 +119,7 @@ public class UserSettingController
         return result;
     }
 
-    @GetMapping( value = "/{key}", produces = "text/plain" )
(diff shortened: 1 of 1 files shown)


larshelge (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/UserSettingController.java
```diff
@@ -135,6 +135,7 @@ public String getUserSettingByKey(
             .get( key );
 
         response.setHeader( ContextUtils.HEADER_CACHE_CONTROL, CacheControl.noCache().cachePrivate().getHeaderValue() );
+        response.setHeader( "Content-Type", ContextUtils.CONTENT_TYPE_TEXT );
```
larshelge, line 138: I suggest we add a static final for the content type header just for consistency/"type safety":  `ContextUtils#HEADER_CONTENT_TYPE`. There is also `HttpHeaders` from Spring.

mortenoh (approved):

PR #11722, opened 2022-09-09
test: add assertStartsWith to Assertions DHIS2-13648

teleivo:
according to our new guideline on Assertions
https://github.com/dhis2/wow-backend/blob/8e9bd4ccdfd8997b1f2f03d9e21543c902d225a6/guides/testing.md?plain=1#L93

# Example Failures

```java
assertStartsWith("prefix",null);

org.opentest4j.AssertionFailedError: expected string to start with 'prefix', got null instead ==> expected: not <null>
```

```java
assertStartsWith("prefix", "is not contained");

org.opentest4j.AssertionFailedError: expected string to start with 'prefix', got 'is not contained' instead ==> 
Expected :true
Actual   :false
```

### DIFF
--- a/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapperTest.java
+++ b/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapperTest.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.webapi.controller.tracker.export;
 
 import static org.hisp.dhis.utils.Assertions.assertContainsOnly;
+import static org.hisp.dhis.utils.Assertions.assertStartsWith;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapperTest.java
```diff
@@ -539,10 +534,9 @@ void testFilterAttributesWhenSameTEAHasMultipleFilters()
         assertNotNull( exception.getMessage() );
```
enricocolasante, line 534: We can get rid of this assertNotNull, right?
It should be included in the new assertion

enricocolasante (approved):
One small point

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapperTest.java
teleivo: good catch, will remove it right away 😄

teleivo: oh, sorry just saw that there are other assertions in the assertAll that would get a NPE if I remove it. I have a follow up where I do remove it as I also add an assertContains. So it will be removed, just not in this PR 😋

enricocolasante: 👍

jbee (approved):

PR #11725, opened 2022-09-11
fix: NPE in /query/events [2.38-DHIS2-13573-backport]

d-bernat:
backport, please see https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/boards/104?modal=detail&selectedIssue=DHIS2-13573&quickFilter=497

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
@@ -493,6 +493,15 @@ public abstract class AbstractAnalyticsService
 
             if ( item.hasOptionSet() )
             {
+                // if itemOptions.get( item.getItem().getUid() ) returns null,
+                // getDimensionItemUidList
+                // returns Lists.newArrayList() which is
+                // equal to no option set and no legend set.
+                // This should be ok, query item can't have both legends and
+                // options
+                // E7215( "Query item cannot specify both legend set and option
+                // set:
+                // `{0}`" )
                 dimensionItems.put( itemUid,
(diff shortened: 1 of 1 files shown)


gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
```diff
@@ -539,12 +548,14 @@ private List<String> getDimensionItemUidList( EventQueryParams params, QueryItem
         {
             return item.getOptionSetFilterItemsOrAll();
         }
-        else
+        else if ( itemOptions != null )
```
maikelarabori, line 551: I think you could rewrite this Javadoc to something like:

    /**
     * Return a list of dimension item uids.
     * If itemOptions is null, it returns an empty List, which means no options set and no legend set.
     *
     * @param params EventQueryParams.
     * @param item QueryItem
     * @param itemOptions itemOtion list.
     * @return a list of uids.
     */

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsService.java
```diff
@@ -493,6 +493,15 @@ private Map<String, List<String>> getDimensionItems( EventQueryParams params,
 
             if ( item.hasOptionSet() )
             {
+                // if itemOptions.get( item.getItem().getUid() ) returns null,
```
maikelarabori, line 496: I think you could make this comment a bit clearer. I would split it into two. The first part is below, and the second part is at the end of the code review.

```
// The call itemOptions.get( item.getItem().getUid() ) can return null.
// This should be ok, the query item can't have both legends and options.
```

maikelarabori (approved):
Looks good, I just added a suggestion to improve a bit the javadoc.

PR #11733, opened 2022-09-12
chore: move tracker CSV package close to its usage DHIS2-13648

teleivo:
its only used in the export, so moving it there. The high level view of the tracker web package

now more closely resembles our domain. Tracker supports import/export both of which have a package. The view package exists to tigh import/export together as we support importing exported data directly.

If you want to know more about each feature you can drill down in the packages. csv was not at the same level of abstraction

image


teleivo:
@enricocolasante since you are the author could you have a look at the sonar issues. Maybe mark some as won't fix if they do not make sense, otherwise could you assign yourself to them. I think you'll be faster at resolving them since you have more context 😄

enricocolasante (approved):

teleivo:
@stian-sandvold I only moved a package. The sonar issues are existing ones that Enrico has assigned himself to. We will tackle them, just not in this PR. Can you please merge this PR, thank you!

PR #11737, opened 2022-09-12
chore: fast and simple pattern based random value generation [TECH-1195]

jbee:
### Summary
This started out as an attempt removing `@MockitoSettings( strictness = Strictness.LENIENT )` from tests.
In case of `TimeToLiveTest` this was straight forward as the mocking wasn't needed any longer at all.

Looking into `RandomGeneratorServiceTest` quickly went down a rabbit hole. 
The mocking was only needed because the way the code was organized so I started reorganizing it turning the generation into static utility. Doing this and cleaning the test a couple of issues became apparent. 

* recreation of `SecureRandom` (expensive)
* many intermediate collections and string conversions (expensive)
* mutable state was manipulated and read in a way that was hard to follow (and I think also had a multi-threading issue)
* use of parallel execution (complex)
* unnecessarily limiting used letters to fist 10 of the alphabet hurting quality of randomness

Addressing this ended with gradually replacing the implementation.

The new implementations randomness is based on a very simple XOR based random generator.
This is crazy fast and cheap and almost passes most randomness quality test suits.
As we do not need this for security but only want unique random values this is a much better trade-off than using the expensive secure random. Also this is cheap to create so each caller/thread will use its own instance seeded with a seed that is also guaranteed to be unique based on current time and a counter.

### Automatic Testing
Test were refactored but they test exactly the existing scenarios and do some additional asserts in some cases.
Also a performance and quality of randomness test scenario was added generating 300k values and requiring that at least 150K of these are unique. This is a very conservative assertion and the actual quality is much better but I don't want this to be flaky due to bad luck. This test takes less than 0.5s on 5 year old hardware.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/RandomPatternValueGenerator.java
```diff
@@ -0,0 +1,203 @@
...
+ * </pre>
+ *
+ * @author Jan Bernitt
+ */
+@NoArgsConstructor( access = AccessLevel.PRIVATE )
+public class RandomPatternValueGenerator
+{
+    private static final String STAR = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+    private static final AtomicLong NEXT_SEED = new AtomicLong( currentTimeMillis() );
```
enricocolasante, line 60: Just a question as we would need something similar to create repeatable performance tests in locust.
Using the same `NEXT_SEED` will lead to the same list of random generated values?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/ValueGeneratorService.java
```diff
@@ -45,70 +48,49 @@
...
         }
 
-        return generatedValues;
+        return generatedNumbers
+            .stream()
+            .map( n -> String.format( "%0" + segment.getParameter().length() + "d", n ) )
+            .collect( Collectors.toList() );
+    }
+
+    private List<String> generateRandomValues( TextPatternSegment segment, int numberOfValues )
```
enricocolasante, line 88: Maybe I am not following the code in the right way but lets say we want to generate 10 random values, if we have variable `numberOfValues` equals 10, then will we have this code generating 10 random values 10 times? So 100 values?

enricocolasante (commented):
Just one doubt on one part of the code and one question on how the code works.
Anyway, this is a really useful PR

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/RandomPatternValueGenerator.java
jbee: Same seed always produces the same pseudo-random sequence.

The seed here is a function of current time and a counter. That means it is the current time millis unless that has already been used before. then we add to it so the next requested seed is also unique. When the next seed is requested and some time has passed we again use the current time and so forth.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/ValueGeneratorService.java
jbee: Good question. I also find this super confusing. From what I gathered each call to random value generation creates a chunk. I just lifted the size of that chunk `RANDOM_GENERATION_CHUNK` out so it becomes an argument. Then we might do this a number of times so yes, the total result are `numberOfValues * RANDOM_GENERATION_CHUNK` values. Which is just a simplified view because generation might also output 2x the chunk, 2x chunk +1 or 3x chunk size depending on the actual pattern. This odd logic was thankfully covered by tests so I am sure I kept it as it is.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/RandomPatternValueGenerator.java
jbee: If I understand you correctly you would want to have the possibility to set the seed for testing. Which is no problem but that code needs to be added.

enricocolasante: Nice.
Yes, I will need this in another project but it should be as easy as injecting the seed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/ValueGeneratorService.java
enricocolasante: :+1:

enricocolasante (approved):

lucaCambi77 (approved):

PR #11743, opened 2022-09-13
wip: use Spring Converter to parse fields request parameter

teleivo:
This allows devs to access FieldPath in their controllers. Enables for example
only fetching things from the DB which will also be returned in the JSON after
field filtering has been applied. Currently, parsing needs to be done twice
as FieldPath are not exposed.

This PR is just a POC. Far from polished 😋

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,13 +3,13 @@
 **/.gitignore
 .github
 .idea
-docker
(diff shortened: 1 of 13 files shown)


teleivo:
will open up a cleaner PR with tests after this spike 🥳

PR #11752, opened 2022-09-13
feat: Add ordering on tei attributes for events DHIS2-13648 [2.39]

enricocolasante:
* feat: Add ordering on tei attributes for events [DHIS2-13648]

* feat: Fix query for ordering on attributes without filter [DHIS2-13648]

* feat: Fix test [DHIS2-13648]

chore: move tracker CSV package close to its usage (#11733)

its only used in the export, so moving it there. The high level view of the tracker web package

now more closely resembles our domain. Tracker supports import/export both of which have a package. The view package exists to tigh import/export together as we support importing exported data directly.

If you want to know more about each feature you can drill down in the packages. csv was not at the same level of abstraction

test: use suppliers in assertContainsOnly (#11732)

and fix 2 sonar issues of redundant call to toString()

test: add assertContains to Assertions (#11723)

according to our new guideline on Assertions
https://github.com/dhis2/wow-backend/blob/8e9bd4ccdfd8997b1f2f03d9e21543c902d225a6/guides/testing.md?plain=1#L93

test: clean up assertions DHIS2-13648 (#11736)

* test: use DateUtils.parseDate and UID constants

* test: use assertIsEmpty

* test: use assertContainsOnly as order does not matter

we did sort the UIDs in the parametrized test functions so that we could assert on UIDs irrespective of order. assertContainsOnly already provides that, so no need to order first.

Remove superfluous assertNotNull as assertContainsOnly already does that

feat: Add ordering attribute column in select [DHIS2-13648] (#11750)


teleivo (approved):

larshelge (approved):

larshelge:
SonarCloud failing on this one @enricocolasante . Approved it.

teleivo:
> SonarCloud failing on this one @enricocolasante . Approved it.

Yes, that is technical debt we already had, that surfaced again in Sonarcloud because we moved files. @Stian did merge the same changes into master. We put these issues on our list of things to tackle after the release.

teleivo (approved):

PR #11762, opened 2022-09-14
feat: Add SourceParams to AggregateDataExchange [DHIS2-13105]

larshelge:
Adds new object `SourceParams` to `AggregateDataExchange` > `Source`. This class represents which period types are allowed for period requet parameters. This refers to the ability to override the period dimension of a data exchange with request parameters.

Adds support for additional source output identifier schemes.

Fixes a bug where output ID scheme was not taken into account for source request.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/Status.java
```diff
@@ -33,6 +33,7 @@
 public enum Status
 {
     OK,
+    SUCCESS,
```
jbee, line 36: How is this different from `OK`? Wasn't that the _success_ state already?

mortenoh (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -117,6 +118,18 @@ public static <E> void assertIsEmpty( Collection<E> actual )
         assertTrue( actual.isEmpty(), actual.toString() );
     }
 
+    /**
+     * Asserts that the given collection is not null and not empty.
+     *
+     * @param <E>
+     * @param actual the collection.
+     */
+    public static <E> void assertNotEmpty( Collection<E> actual )
```
jbee, line 127: If a type parameter is not needed for consistent type checking it is better to use `Collection<?>` instead.

larshelge: Agree, fixed.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/Status.java
larshelge: Good question, so I am getting `SUCCESS` from the data value set API in 2.37 and `OK` in 2.38.

So when targeting older DHIS 2 version for data exchange, it breaks. I can add a new enum to the data exchange project to contain this if that is better.

jbee: I see, I think this is due to changing the type of report we return or report vs. web message response. Maybe the extra enum is the best compromise to keep this odd exception local.

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/Status.java
larshelge: Makes sense. Fixed.

PR #11775, opened 2022-09-14
test: be more lenient in cache header assertion

teleivo:
A few tests like `NO_CACHE` and `CACHE_1_MINUTE` are enough to assert that the `CacheStrategy` passed to configureResponse is making it into the max-age=60 response header value. Most other durations just delegate to TimeUnit for making the calculation which we do not need to test.

`CACHE_6AM_TOMORROW` calls new Date(). So testing this we make 2 calls to it once in the test for the expected value and once in the implementation. Since we cannot control the time in the test, we can be a bit more lenient in the expectation of the max-age seconds.

## Example errors

```java
assertWithinRange(100, 20, 100);
```

leads to

```
org.opentest4j.AssertionFailedError: lower bound 100 must be < than the upper bound 20 ==> 
Expected :true
Actual   :false
```

```java
assertWithinRange(10, 20, 21);
```

leads to

```
Expected actual 21 to be <= than upper bound 20 ==> expected: <true> but was: <false>
Comparison Failure: 
Expected :true
Actual   :false
```

```java
assertWithinRange(10, 20, 9);
```

leads to

```
Expected actual 9 to be >= than lower bound 10 ==> expected: <true> but was: <false>
Comparison Failure: 
Expected :true
Actual   :false
```

### DIFF
--- a/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
@@ -147,4 +147,46 @@ public final class Assertions
         assertTrue( actual.contains( expected ), () -> String
             .format( "expected actual to contain '%s', got '%s' instead", expected, actual ) );
     }
+
+    /**
+     * Asserts that the given value is within the range of lower and upper bound
+     * (inclusive i.e. [lower, upper]).
+     *
+     * @param lower lower bound
+     * @param upper upper bound
+     * @param actual actual value to be checked
+     */
+    public static void assertWithinRange( long lower, long upper, long actual )
+    {
+        assertTrue( lower < upper,
+            () -> String.format( "lower bound %d must be < than the upper bound %d", lower, upper ) );
+
+        assertAll(
+            () -> assertGreaterOrEqual( lower, actual ),
+            () -> assertLessOrEqual( upper, actual ) );
+    }
+
+    /**
+     * Asserts that the given value is greater or equal than lower bound.
+     *
+     * @param lower lower bound
+     * @param actual actual value to be checked
+     */
+    public static void assertGreaterOrEqual( long lower, long actual )
+    {
+        assertTrue( actual >= lower,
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -147,4 +147,51 @@ public static void assertContains( CharSequence expected, String actual )
...
+     * Asserts that the given value is within the range of lower and upper bound
+     * (inclusive i.e. [lower, upper]).
+     *
+     * @param lower lower bound
+     * @param upper upper bound
+     * @param actual actual value to be checked
+     */
+    public static void assertWithinRange( long lower, long upper, long actual )
+    {
+        // TODO should this be an assertion error? its more of an
```
jbee, line 162: remove TODO

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -147,4 +147,51 @@ public static void assertContains( CharSequence expected, String actual )
...
+     * @param actual actual value to be checked
+     */
+    public static void assertGreaterOrEqual( long lower, long actual )
+    {
+        assertTrue( actual >= lower,
+            () -> String.format( "Expected actual %d to be >= than lower bound %d", actual, lower ) );
+    }
+
+    /**
+     *
```
jbee, line 186: Remove empty javadoc line at start (also above)

jbee (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -147,4 +147,47 @@ public static void assertContains( CharSequence expected, String actual )
...
```
maikelarabori, line 182: I believe you can remove the blank line at the top.

maikelarabori (approved):

PR #11792, opened 2022-09-15
feat: Add test for order on multiple attributes [DHIS2-13648]

enricocolasante:
Add test for order on multiple attributes. Check that if the first order attribute is equal then the order is based on the second one

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
@@ -670,6 +670,40 @@ class EventExporterTest extends TrackerTest
     }
 
     @Test
+    void testOrderEventsOnMultipleAttributesDesc()
+    {
+        EventSearchParams params = new EventSearchParams();
+        params.setOrgUnit( orgUnit );
+
+        params.setFilterAttributes( List.of( queryItem( "toUpdate000" ), queryItem( "toDelete000" ) ) );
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java
```diff
@@ -669,6 +669,40 @@ void testOrderEventsOnAttributeDesc()
...
+        params.setAttributeOrders( List.of( new OrderParam( "toDelete000", OrderParam.SortDirection.DESC ),
+            new OrderParam( "toUpdate000", OrderParam.SortDirection.DESC ) ) );
+
+        List<String> trackedEntities = eventService.getEvents( params ).getEvents().stream()
+            .map( Event::getTrackedEntityInstance )
+            .collect( Collectors.toList() );
+
+        assertEquals( List.of( "QS6w44flWAf", "dUE514NMOlo" ), trackedEntities );
+    }
+
```
lucaCambi77, line 688: if the order is no relevant, as @Ivo is suggesting we should do like https://github.com/dhis2/dhis2-core/pull/11736 and use assertContainsOnly ?

enricocolasante: The order is exactly what is tested here so I cannot use `assertContainsOnly`

teleivo (approved):

stian-sandvold (approved):

PR #11797, opened 2022-09-15
chore: extract validation in EventRequestToSearchParamsMapper DHIS2-13648

teleivo:
* extract validation methods in EventRequestToSearchParamsMapper and ensure that we validate before building EventSearchParam (where we can). Filters/filterAttributes are being validated later on as its mixed with transformations
* TrackerEventCriteria should only represent the request parameters to `/tracker/event`. Move any parsing/validation into EventRequestToSearchParamsMapper which translates request parameters (TrackerEventCriteria) to the repository layer (EventSearchParam)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapper.java
```diff
@@ -127,66 +134,35 @@ void setSchema()
...
+        validateFilter( eventCriteria.getFilter(), eventIds, eventCriteria.getProgramStage(), programStage );
 
-        if ( eventIds == null )
-        {
-            eventIds = new HashSet<>();
-        }
+        Set<String> assignedUserIds = parseUids( eventCriteria.getAssignedUser() );
+        validateAssignedUsers( eventCriteria.getAssignedUserMode(), assignedUserIds );
+
+        EventSearchParams params = new EventSearchParams();
```
enricocolasante, line 165: I wonder if we can manage to not instantiate EventSearchParams and  maybe use a builder later in the code.
To do so we will need to change how we add filters, filterAttributes and dataElements

enricocolasante (approved):
There is already a lot in the PR but I left a suggestion to have the validation and the build of `EventSearchParams` even more separated

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/EventRequestToSearchParamsMapper.java
teleivo: Agree 👍 We'll explore it tomorrow 😊

stian-sandvold (approved):

PR #11798, opened 2022-09-15
ci: Check Jenkins build log for failures before sending Slack notification [skip ci]

radnov:
When there are failing tests, the Maven Surefire plugin output contains "There are test failures.", which could be used to make sure the build failed due to test errors and not something else (e.g. dependency transfer issues, compilation errors, etc).   
Looking for strings in the build log seems to be the only consistent way to find out the stage failure reason and this particular string [seems to be part of the Maven Surefire plugin since 2010](https://github.com/apache/maven-surefire/commit/5a06237fcfd40e16b356f3af0f42c44e381cd12b#diff-bf7f174589af8f08638da17d6aefbe897133b7b77ff363a1e12439c37ea8683eR44), so it should be reliable enough.

Example:
```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project dhis-web-api-test: There are test failures.
...
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M7:test (default-test) on project dhis-web-api-test: There are test failures.
...
Caused by: org.apache.maven.plugin.MojoFailureException: There are test failures.
```

This PR aims to get rid of false positive test failure notifications sent to the backend channel by checking whether the build output contains that string.

Related to https://github.com/dhis2/jenkins-pipeline-library/pull/13

---

Tested by faking a failing test in the `dhis-api` module and sending the notification to myself:
image
image

[Build with fake failing test](https://ci.dhis2.org/job/test-notifications/job/fix-test-failure-false-positive-notifications/2/).

Also tested with a fake stage failure by using the Jenkins pipeline [error step](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#error-error-signal), to make sure that Build stage failures that are not caused by test failures (like dependency transfer issues or compilation errors) will not send a notification.
image

[Build with fake stage failure](https://ci.dhis2.org/job/test-notifications/job/fix-test-failure-false-positive-notifications/4/).


teleivo (commented):
How would we notice if the string we are looking for changes? However unlikely that may be 😋 

And can we make such a change without having to create a PR in the library repo as well? Just curious

radnov:
> How would we notice if the string we are looking for changes? However unlikely that may be 😋
> 
> And can we make such a change without having to create a PR in the library repo as well? Just curious

@teleivo If the string changes - we will start getting false positive notifications in the backend channel again. 😅 
The library repo PR only provides a way to look into the current build's logs, so any changes in the string we're looking for would be still in the jenkinsfile for the pipeline.

radnov:
> How would we notice if the string we are looking for changes? However unlikely that may be 😋
> 
> And can we make such a change without having to create a PR in the library repo as well? Just curious

@teleivo That's a good point. Right now I can't think of another way besides keeping an eye on the general jenkins notifications channel we have (where we get notifications regardless of the failure reason) and checking the failure reasons (which I do at least once a day 😅). Do you have any ideas? 🤔 

The library repo PR only provides a way to look into the current build's logs, so any changes in the string we're looking for would be still in the jenkinsfile for the pipeline.

teleivo (approved):
> Do you have any ideas?
Not right now 😓 

Lets try it

Philip-Larsen-Donnelly (approved):
Great!

PR #11801, opened 2022-09-15
fix: metadata export with dependencies is broken [DHIS2-13744]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-13744
#### Issue:

-  Metadata export with dependencies doesn't work with api request extension `json.zip` or `xml.zip`.
- This is due those endpoint have not been updated to return `ResponseEntity<MetadataExportParams>`. Which in result are not processed by `MetadataExportParamsMessageConverter`

#### Fix

- Updated all controllers which supports export metadata with dependencies.
- Add web api test.
- Update `MvcTestConfig` so it support testing a GET request with extension `*.zip`.
- Some codes updated to pass code smells.

#### Test
- In Import/Export app, do metadata export with dependencies for all metadata type, verify if the downloading compressed file is working.


mortenoh (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/MetadataExportParams.java
```diff
@@ -246,4 +253,19 @@ public boolean getSkipSharing()
...
     }
+
+    public boolean isExportWithDependencies()
+    {
+        return objectExportWithDependencies != null;
+    }
+
+    public IdentifiableObject getObjectExportWithDependencies()
+    {
+        return objectExportWithDependencies;
```
jbee, line 264: I am a bit confused by the naming here. As far as I understand the object is the dependency. Wouldn't it be more clear to call this field simply "dependency" or "exportDependency"? (The `isExportWithDependencies` makes sense as is)

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/MetadataExportService.java
```diff
@@ -99,6 +99,21 @@ void getMetadataAsObjectNodeStream( MetadataExportParams params, OutputStream ou
...
+     * of the specified export parameters are used for the metadata with
+     * dependencies export. All objects are written to given outputStream
+     *
+     * @param object The {@link IdentifiableObject} to be exported with
+     *        dependencies.
+     * @param params {@link MetadataExportParams}
+     * @param outputStream Streaming target.
+     * @throws IOException
+     */
+    void getMetadataWithDependenciesAsNodeStream( IdentifiableObject object,
```
jbee, line 113: `get` is maybe not the best verb here. I also had trouble with "write" in other exports since it is unclear if we mean write to database (=import) or write to output (=export). So I started naming these `export...`.

jbee (approved):
Looks good. Two naming suggestions.

larshelge (approved):

PR #11805, opened 2022-09-16
chore: do not fetch optional ids from DB if empty

teleivo:
* do not fetch optional ids from DB if empty like `program`, `programStage`, ...
* reduce chance of mutation in TrackerEventCriteria by replacing `@Data` which adds setters. Use a `@Builder` as lots of these fields are optional. Return unmodifiable copies of the collections. Note that the date fields are still mutable. We need to use an explicit constructor with the parent class fields to get the Lombok builder to work. https://projectlombok.org/features/experimental/SuperBuilder is experimental and not on our https://github.com/dhis2/wow-backend/blob/master/guides/lombok.md#allowed-annotations list yet.


teleivo:
closing as we are trying it a different way

PR #11810, opened 2022-09-16
chore: split old/new tracker enrollment export mappers TECH-1414

teleivo:
so we can safely adapt the new tracker endpoints without affecting old tracker endpoints

* duplicate mappers so we can have one dedicated for old and one for new tracker. Deliberate duplication, just so you know sonarqube 😋 
* adhere to
https://github.com/dhis2/wow-backend/blob/master/guides/spring_conventions.md#bean-id


teleivo:
@stian-sandvold issues are since we duplicate classes that have technical debt in them. I assigned myself the issues in the "new tracker code". Will have a look at them later, so the diff will become clear. So please allow merging even though we have supposedly 2 "new" issues (sonarqube might need some time to update its comment here).

enricocolasante (approved):

stian-sandvold (approved):

PR #11811, opened 2022-09-16
fix: Empty event table when OU is not defined in PSI [DHIS2-13754]

maikelarabori:
Without this fix, the analytics event table for some Programs may not be populated. This small change should address this problem. This can be reproduced in Sierra Leone (Program `1150221`).

Some ProgramInstances may not have an OrgUnit associated (defined as null in such cases).
In Sierra Leone we have this particular case for Program `1150221`:
```
SELECT * FROM programinstance p WHERE organisationunitid IS NULL AND programid = 1150221;
```
This query will bring the column `organisationunitid` set to `NULL`.

Because of such cases, that might show up in production, we are changing how we join the respective statement (see the code changes). Instead of a `inner join` we are now using `left join`.

This brings back the population of the analytics events tables even in cases where there are no OrgUnit associated with a ProgramInstance.


gnespolino (approved):

d-bernat (approved):

jimgrace (approved):
Looks good, thanks!

PR #11822, opened 2022-09-19
fix: Slow PSQL queries when retrieving /events [ DHIS2-12039 ]

lucaCambi77:
Following performance issues in Bangladesh for events.
Move the dataelementcategoryoption access check for non-superusers to a case in select.

old plan https://explain.dalibo.com/plan/q6F#plan
new plan https://explain.dalibo.com/plan/mvq#plan


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -1738,23 +1734,22 @@ private String addLastUpdatedFilters( EventSearchParams params, MapSqlParameterS
         return sqlBuilder.toString();
     }
 
-    private String getCategoryOptionSharingForUser( User user, MapSqlParameterSource mapSqlParameterSource )
+    private String getCategoryOptionSharingForUser( User user, EventSearchParams params )
```
teleivo, line 1737: Seems like EventSearchParams is mostly first in the method signatures in this class. I would keep it that way to be consistent.

teleivo (commented):
Do I read the https://explain.dalibo.com/plan/q6F#plan and https://explain.dalibo.com/plan/mvq#plan correctly that before the query took 6min and then 25s 🤔 

This improvement only applies to superusers right?

teleivo (approved):

stian-sandvold (approved):

PR #11831, opened 2022-09-19
feat: support application zip archives with a top-level directory [DHIS2-13760]

amcgee:
Fixes [DHIS2-13760](https://dhis2.atlassian.net/browse/DHIS2-13760)

This loops through all files in the archive to determine if there is a common directory prefix, then strips that prefix from the beginning of each file path when extracting it.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
```diff
@@ -0,0 +1,81 @@
...
+            else if ( prefix == null )
+            {
+                prefix = "";
+            }
+            else
+            {
+                return "";
+            }
+        }
+        return prefix == null ? "" : prefix;
```
jbee, line 79: Not sure if that works but have you tried take the first, then use streams `allMatch`?

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
amcgee: Nice, that will probably be much cleaner.  I'll update.

amcgee:
@jbee updated!

stian-sandvold (approved):

PR #11834, opened 2022-09-20
chore: map directly to TrackedEntityInstanceQueryParams in new tracker TECH-1414

teleivo:
Before we split old/new tracker web related code we used TrackerTrackedEntityCriteriaMapper to map from `TrackedEntityCriteria` (coming from `/tracker/trackedEntities`) to map to `TrackedEntityInstanceCriteria` (which represents the old trackedEntities endpoint query params).

* Now `TrackedEntityCriteria` from request to new `/tracker/trackedEntities` are mapped directly to `TrackedEntityInstanceQueryParams` which we use to fetch TEI from the DB.

* Initialize collections to reduce defensive null checking. Inline some temp vars only used in the setter. This makes sonarqube happy as it reduces complexity.

Less indirection => easier to follow 🎉 

## Process

* Made sure we have good test coverage first
* Used `TrackerTrackedEntityCriteriaMapper` to make sure to use the correct fields that then go into `TrackedEntityInstanceQueryParams`.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntityCriteriaMapper.java
```diff
@@ -384,4 +369,17 @@ private void validateOrderParams( List<OrderParam> orderParams, Map<String, Trac
...
     }
+
+    private Set<String> getAssignedUsers( TrackerTrackedEntityCriteria criteria )
+    {
+        Set<String> assignedUsers = new HashSet<>();
+
+        if ( criteria.getAssignedUser() != null && !criteria.getAssignedUser().isEmpty() )
+        {
+            assignedUsers = TextUtils.splitToSet( criteria.getAssignedUser(), TextUtils.SEMICOLON ).stream()
+                .filter( CodeGenerator::isValidUid ).collect( Collectors.toSet() );
```
enricocolasante: I am just wondering why we remove invalid UIDs only for assigned users but not for OrgUnits and TrackedEntityInstances

enricocolasante (commented):
Just a comment to reason about invalid uids.
I do not mark as approved just because auto-merge is on

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntityCriteriaMapper.java
teleivo: I was also wondering. I have no clue 🤷🏻 In this PR I just move code around. I took the code from the TrackedEntityInstanceCriteria. So this is how it was.

I think we should discuss this topic. I don't think its a good idea to remove invalid UIDs in any case. As a user I would expect all my parameters to be taken into account. If my request is not valid I expect an error.

teleivo: We will discuss this in todays design meeting 😄 Lets keep the behavior as it is for now. I would see this as a dedicated task/issue where we need to go through all our endpoints and ensure we return an error instead of filtering the UIDs out.

enricocolasante (approved):

stian-sandvold (approved):

PR #11837, opened 2022-09-20
fix: /tracker/enrollments does not return attributes, events and relationships [ Dhis2-12906 ]

lucaCambi77:
add attributes, events, and relationships to /tracker/enrollments endpoint
It also fixes the same issue for the old tracker
For the old tracker, tei params are initialized to false and triggered to true only the ones that are requested to avoid side effects


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerEnrollmentsExportControllerTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+
+@ExtendWith( MockitoExtension.class )
+public class TrackerEnrollmentsExportControllerTest
```
teleivo, line 55: Can you please have a look at the web api tests we already have and extend them 😄 We added extensive tests when we fixed field filtering. For example for /tracker/trackedEntities see https://github.com/dhis2/dhis2-core/blob/52d6c7f8d06e7013e6fa00d5cfb928db6387a5e5/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerTrackedEntitiesExportControllerTest.java#L141-L181

And for /tracker/enrollments see https://github.com/dhis2/dhis2-core/blob/52d6c7f8d06e7013e6fa00d5cfb928db6387a5e5/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerEnrollmentsExportControllerTest.java#L92-L110

There are tests for making a request with no fields, checking that the default fields are returned and tests for some field filter combinations. You can add some cases for the issues you have fixed 😄 This is more meaningful than mocking our interactions 😬

teleivo (changes requested):
I just noticed that you did not add to the tests we already have. Maybe you did not know they existed. Please add cases for the issues you fixed to the existing tests (this is regarding new tracker). See comment.

teleivo (commented):
## Minor details

* remove `toString` on `EnrollmentParams` and `TrackedEntityInstanceParams` as we already have an implementation via `@Value`
* if you use `@Service/@Component` add a bean id https://github.com/dhis2/wow-backend/blob/master/guides/spring_conventions.md#bean-id

## Design

When I worked on fixing field filtering I first encountered `TrackedEntitiesSupportService`. It existing did not feel right, but I did not have the time to do anything about it and wasn't to sure what to do yet :yum:. What is its responsibility? What kind of "Support"? This suggests that `TrackedEntitiesService` does not do what we actually need. So ideally we adapt our service to do what we actually need or rename it/extract explicit responsibilities from `TrackedEntitiesSupportService`.

Now that we add another SupportService I think its time to act on that 😄 It becomes clear that the responsibility of the SupportServices related to this issue is to map `List<String> fields` (soon https://dhis2.atlassian.net/browse/DHIS2-13731 List<FieldPath>) to `InstanceParams`.

So I suggest we create classes

1. `TrackerEnrollmentFieldsParamMapper` with signature `public static EnrollmentParams map(List<String> fields)`

2. `TrackerTrackedEntityFieldsParamMapper` with signature `public static TrackedEntityInstanceParams map(List<String> fields)`

which only contain pure functions (static) without any Spring annotations. They should be functions as they do not need any state as they just purely map from one type to another. We don't need Spring to create singletons for us.

This separation/naming makes it clear what these classes are responsible for IMHO. Its easy to spot where a change to the behavior of what we fetch from the DB in response to a certain `fields` query param values has to go.

I would place the classes in `org.hisp.dhis.webapi.controller.tracker.export` as that's the only place where they are used. Creating a package with a generic name like `support` will only attract unrelated classes IMHO.

I would also use composition instead of inheriting from an abstract class like `InstanceFieldsSupportService`. `InstanceFieldsSupportService` just contains pure functions we can easily reuse in the above mapper classes. In this simple case I don't see an advantage in inheriting IMHO it makes it harder to read/understand.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerEnrollmentsExportControllerTest.java
```diff
@@ -118,6 +156,179 @@ void getEnrollmentByIdNotFound()
...
+
+        assertWithRelationshipResponse(
+            (GET( "/tracker/enrollments/{id}?fields=relationships", programInstance.getUid() )
+                .content( HttpStatus.OK )) );
+    }
+
+    @Test
+    void getEnrollmentByIdWithEventsFields()
+    {
+        programInstance.setProgramStageInstances( Sets.newHashSet( programStageInstance ) );
```
teleivo, line 181: use `Set.of` instead of `com.google.common.collect.Sets`

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/fields/EnrollmentFieldsParamMapperTest.java
```diff
@@ -41,10 +42,8 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith( MockitoExtension.class )
```
teleivo, line 44: mockito should not be needed since we are just testing a pure function 😄

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/fields/TrackedEntityFieldsParamMapperTest.java
```diff
@@ -50,19 +45,13 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith( MockitoExtension.class )
```
teleivo, line 47: same here no mockito should be needed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/InstanceFieldsParamMapper.java
```diff
@@ -38,17 +38,17 @@
...
 /**
- * Abstract class that provides basic methods to transform input fields into
- * {@link FieldPath } based on {@link FieldFilterParser }. It follows the
- * principles of {@link org.hisp.dhis.fieldfiltering.FieldFilterService}
+ * Provides basic methods to transform input fields into {@link FieldPath }
+ * based on {@link FieldFilterParser }. It follows the principles of
+ * {@link org.hisp.dhis.fieldfiltering.FieldFilterService}
  */
-abstract class InstanceFieldsSupportService
+public interface InstanceFieldsParamMapper
```
teleivo: I don't think we need an interface. The mapping functions in for example TrackerTrackedEntityFieldsParamMapper are just pure functions. We don't need to mock these in tests of the TrackerTrackedEntityController. We also just have one implementation for each type of controller. So IMHO the reasons I look to when creating an interface are not given here. All that we need is for functions like TrackerTrackedEntityFieldsParamMapper.map to be able to reuse some common functions that are in here 😄 

I assume the prefix `InstanceFieldsParamMapper` comes from `TrackedEntityInstance`? If so that does not apply here as this class only knows about field paths. Prefix in `InstanceParams` is probably also not accurate 🤔

teleivo (changes requested):
Thanks @lucaCambi77 for your changes 😄 
I think we should change some naming, unused mocking and remove the interface.

Just curious, what benefit do you see in us having a `fields` package?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/InstanceFieldsParamMapper.java
lucaCambi77: In commit [999f705](https://github.com/dhis2/dhis2-core/pull/11837/commits/999f705ce044507edcd8a8d876a3d956966f8a23) i've already moved it to a class. 
I've renamed it to `FieldsPathParamsHelper`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/FieldsPathParamsHelper.java
```diff
@@ -42,35 +41,30 @@
  * based on {@link FieldFilterParser }. It follows the principles of
  * {@link org.hisp.dhis.fieldfiltering.FieldFilterService}
  */
-public interface InstanceFieldsParamMapper
+public class FieldsPathParamsHelper
```
teleivo, line 44: Since you created a fields package this can be package private as its only used by the 2 mapping classes 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/FieldsPathParamsHelper.java
```diff
@@ -42,35 +41,30 @@
...
-     * should be fetched from the DB.
-     *
-     * @param fields fields query parameter values
-     * @return InstanceParams
-     */
-    InstanceParams map( List<String> fields );
+    protected static final String FIELD_ATTRIBUTES = "attributes";
 
-    static Map<String, FieldPath> getRoots( List<String> fields )
+    protected static Map<String, FieldPath> getRoots( List<String> fields )
```
teleivo, line 56: Do these need to be `protected`? If the class is package private these can only be accessed from within the package.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/FieldsPathParamsHelper.java
```diff
@@ -42,35 +41,30 @@
  * based on {@link FieldFilterParser }. It follows the principles of
  * {@link org.hisp.dhis.fieldfiltering.FieldFilterService}
  */
-public interface InstanceFieldsParamMapper
+public class FieldsPathParamsHelper
 {
-    String FIELD_RELATIONSHIPS = "relationships";
+    private FieldsPathParamsHelper()
```
teleivo, line 46: I think its a good idea to prevent instantiation (by mistake) and document why you added a private constructor via `throw new IllegalStateException( "Utility class" );`.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/params/InstanceParams.java
```diff
@@ -0,0 +1,45 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.dxf2.events.params;
+
+public interface InstanceParams
```
teleivo, line 30: Can we get away without adding this interface? Instead use the Enrollment/TrackedEntityInstanceParams directly

I wonder if for example we will use the same technique in `/tracker/events` and `/tracker/relationships` to only fetch whats necessary. Will that interface make sense for these as well?

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/EnrollmentServiceTest.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Spy
+    AbstractEnrollmentService enrollmentService;
+
+    String enrollmentId = CodeGenerator.generateUid();
+
+    @BeforeEach
+    void setUp()
+    {
+        ReflectionTestUtils.setField( enrollmentService, "currentUserService", currentUserService );
```
teleivo, line 107: I think this is bad practice and was also the source of flaky tests in the past 😬 Note how you do not clean up your mutations. In integration (h2 or not) tests this will influence other tests with us not knowing why when it fails. 

If I understand correctly you are trying to test that given some `InstanceParam` we only fetch the necessary information.

If so can we write an integration test that shows that?

If not, then why not use `@InjectMocks` or call the constructor of an EnrollmentService where you pass in the mocked dependencies. If `@InjectMocks` does not work on an abstract class you can create a private subclass of it in this test class and test through that. Calling its constructor with the mocks.

teleivo (changes requested):
I think we should adapt EnrollmentServiceTest

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
```diff
@@ -288,27 +290,20 @@ public List<Enrollment> getEnrollments( Iterable<ProgramInstance> programInstanc
...
     @Override
-    public Enrollment getEnrollment( ProgramInstance programInstance )
-    {
-        return getEnrollment( currentUserService.getCurrentUser(), programInstance, TrackedEntityInstanceParams.FALSE,
-            false );
-    }
-
-    @Override
-    public Enrollment getEnrollment( ProgramInstance programInstance, TrackedEntityInstanceParams params )
+    public Enrollment getEnrollment( ProgramInstance programInstance, InstanceParams params )
```
enricocolasante, line 300: Why was `getEnrollment( ProgramInstance programInstance )` removed?
I can see that in other classes we are calling a lot of times `getEnrollment( programInstance, TrackedEntityInstanceParams.FALSE )`.
Wouldn't make sense to maintain one method without the params that are default to FALSE or we would change some behavior that I cannot see

enricocolasante (commented):
A couple of comments

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/FieldsPathParamsHelper.java
lucaCambi77: you can't instantiate the class unless you change the visibility so I don't see the benefit to add that

lucaCambi77:
@teleivo @enricocolasante I forgot to add the params for /enrollments end point. Now we have` TrackedEntityInstanceParams.FALSE` only for relationships, so we may want to have a look at that in a separate issue.

teleivo:
> @teleivo @enricocolasante I forgot to add the params for /enrollments end point. Now we have` TrackedEntityInstanceParams.FALSE` only for relationships, so we may want to have a look at that in a separate issue.

What is EnrollmentParams for then?

This PR is already quite big as you also made changes to the old endpoints which are not part of the issue. So I would not want any more changes in here. Important is that https://dhis2.atlassian.net/browse/DHIS2-12906 is fixed with these changes. If there are more issues or things we can improve then lets create another issue 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fields/FieldsPathParamsHelper.java
teleivo: Joshua Bloch explains it in Effective Java as:

> the exception isn’t strictly required, but it provides insurance in case the constructor is accidentally invoked from within the class. It guarantees the class will never be instantiated under any circumstances.

The message `"Utility class"` also acts as a documentation of intent. We also use this idiom a lot, so it would simply be consistent.

lucaCambi77:
@enricocolasante @teleivo I've closed this issue as I am feeling we are wasting time here. Feel free to pick it up in the next milestone as maybe I am missing something for this work. We can use it as a reference for future works

PR #11867, opened 2022-09-22
fix: remove non-persisted properties from identifiable token query

vietnguyen:
Issue:
- `/api/eventVisualizations?filter=identifiable:token:child` return 0 results.
- eventVisualizations doesn't have shortName but the query tries to filter the token with shortName property by the in-memory query.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/DefaultJpaQueryParser.java
```diff
@@ -105,7 +105,7 @@ private void handleIdentifiablePath( Schema schema, String operator, Object arg,
         disjunction.add( getRestriction( schema, "code", operator, arg ) );
         disjunction.add( getRestriction( schema, "name", operator, arg ) );
 
-        if ( schema.haveProperty( "shortName" ) )
+        if ( schema.getPersistedProperty( "shortName" ) != null )
```
mortenoh, line 108: Maybe use `schema.havePersistedProperty("shortName")` instead?

vietnguyen: ok will update

mortenoh (approved):

PR #11873, opened 2022-09-22
feat: support application zip archives with a top-level directory [DHIS2-13760] (2.39)

amcgee:
2.39 backport of #11831

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
@@ -59,6 +59,7 @@ import org.hisp.dhis.external.conf.ConfigurationKey;
 import org.hisp.dhis.external.conf.DhisConfigurationProvider;
 import org.hisp.dhis.external.location.LocationManager;
 import org.hisp.dhis.external.location.LocationManagerException;
+import org.hisp.dhis.util.ZipFileUtils;
 import org.jclouds.ContextBuilder;
 import org.jclouds.blobstore.BlobRequestSigner;
 import org.jclouds.blobstore.BlobStore;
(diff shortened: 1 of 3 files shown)


jbee (approved):

stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -338,10 +339,17 @@ public App installApp( File file, String filename, Cache<App> appCache )
         try ( ZipFile zip = new ZipFile( file ) )
         {
             // -----------------------------------------------------------------
-            // Parse ZIP file and it's manifest.webapp file.
+            // Determine top-level directory name, if the zip file contains one
             // -----------------------------------------------------------------
 
-            ZipEntry entry = zip.getEntry( MANIFEST_FILENAME );
+            String prefix = ZipFileUtils.getTopLevelDirectory( zip.entries().asIterator() );
+            log.debug( "Detected top-level directory '" + prefix + "' in zip" );
```
larshelge, line 346: Very minor, you can do this with slf4j to avoid string concatenation:

> log.debug( "Detected top-level directory '{}' in zip archive", prefix );

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
```diff
@@ -0,0 +1,79 @@
...
+    {
+        if ( !zipEntries.hasNext() )
+        {
+            return "";
+        }
+
+        ZipEntry firstEntry = zipEntries.next();
+
+        Matcher m = TOP_LEVEL_DIRECTORY_PREFIX_PATTERN.matcher( firstEntry.getName() );
+        if ( m.find() )
```
larshelge, line 61: I wonder if `matches` is more appropriate than `find` if you are trying to match the entire string just once, and are not looking for multiple matches.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
amcgee: Good to know, thanks @larshelge !  I was copying the style used elsewhere in this file, slf4j is definitely nicer!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ZipFileUtils.java
amcgee: It looks like `.matches()` would work as well.  All the examples of parsing regex capture groups I found online used `.find()` so I went with that - in this case it should be pretty much identical since the pattern starts with `^`

amcgee: @larshelge should I switch this to `.matches`?

PR #11875, opened 2022-09-22
fix: Trying a fix for IllegalAccessException [DHIS2-13758]

maikelarabori:
This is another attempt to fix an issue that only manifests in "debug/dev".

This exception does not happen in other environments.

```
SQL state [null]; error code [0]; IllegalAccessException:
com/sun/rowset/providers/RIOptimisticProvider;
nested exception is javax.sql.rowset.spi.SyncFactoryException:
IllegalAccessException: com/sun/rowset/providers/RIOptimisticProvider
```


jimgrace (approved):
Looks good. I like the "o.owndate" comment. :)

PR #11880, opened 2022-09-23
fix: allow mutation of query item collections TECH-1414

teleivo:
We cannot yet fully deal with immutable collections all the way to the DB fetching code. Service/repo code is mutating the structures that represent the users request. Go back to mutable collections (at least for now). If we manage to move this responsibility into only the criteria mappers we might be able to make the param structures immutable.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
@@ -87,7 +87,7 @@ class RequestParamUtils
     {
         return parseUidString( input )
             .filter( CodeGenerator::isValidUid )
-            .collect( Collectors.toUnmodifiableSet() );
+            .collect( Collectors.toSet() );
     }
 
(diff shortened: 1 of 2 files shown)


jbee:
I understand this approach as a workaround but long term this should not be a pattern. We should decide for general semantics that avoid having to make defensive copies of collections.

teleivo:
> I understand this approach as a workaround but long term this should not be a pattern. We should decide for general semantics that avoid having to make defensive copies of collections.

Totally agree 👍🏻 and this is the plan 😋 Thanks for your comment! It got me thinking again. I was in a rush due to the 🥶 freeze and someone being blocked. I went back to mutable structures the way it was before (see diff). Long-term I would love for some of these to be immutable. If we find time, we'd better do this for an entire tracker endpoint's criteria/params in one go.

We have mappers in the web part of the code whos responsibility it is to map from a structure representing the request parameters (we call `criteria`) to one representing the DB query (we call `param`). I thought the responsibility is only in the criteria mappers (`criteria` -> `param`) we have. I did not know that this is actually not the case.

enricocolasante (approved):

lucaCambi77 (approved):

PR #11890, opened 2022-09-24
feat: Add new end point for TEA file resource  [DHIS2-13657]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-13657


stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityInstanceController.java
```diff
@@ -287,6 +271,66 @@ public void getAttributeImage(
         }
     }
 
+    @GetMapping( value = "{teiId}/{attributeId}/files" )
```
enricocolasante, line 274: If in the response there will always be one file, maybe we should change the URL
```suggestion
    @GetMapping( value = "{teiId}/{attributeId}/file" )
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityInstanceController.java
```diff
@@ -216,7 +218,7 @@ public void getAttributeImage(
                 unauthorized( "You're not authorized to access the TrackedEntityInstance with id: " + teiId ) );
         }
 
-        if ( attribute.size() == 0 )
+        if ( attribute.isEmpty() )
```
enricocolasante, line 221: I know it was this way before, but would it make more sense to rename it to `attributes`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityInstanceController.java
```diff
@@ -287,6 +271,66 @@ public void getAttributeImage(
...
+        {
+            throw new WebMessageException(
+                unauthorized( "You're not authorized to access the TrackedEntityInstance with id: " + teiId ) );
+        }
+
+        Set<TrackedEntityAttributeValue> attributeValues = tei.getTrackedEntityAttributeValues();
+
+        Optional<TrackedEntityAttributeValue> optionalAttributeValue = attributeValues.stream()
+            .filter( att -> att.getAttribute().getUid().equals( attributeId ) )
+            .findFirst();
```
enricocolasante, line 299: Here you can use the same syntax as the tei before and use `orElseThrow` and you will not need to use `get()` later in the code

enricocolasante (commented):
Small code fixes

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityInstanceController.java
zubaira: I think this was `attribute` because the list will always have either 0 or 1 object.

zubaira: otherwise it make sense to have `attributes` rather `attribute`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityInstanceController.java
zubaira: resolved

enricocolasante (approved):

PR #11891, opened 2022-09-25
fix: check for empty objectUids

mortenoh:
## Summary

Checks for empty object collection before running SQL query. Needed so that we don't do invalid `in (...)` queries.

[TECH-1437](https://dhis2.atlassian.net/browse/TECH-1437)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
@@ -58,7 +58,7 @@ public class JdbcOrgUnitAssociationsStore
     {
         Set<String> userOrgUnitPaths = getUserOrgUnitPaths();
 
-        if ( userOrgUnitPaths.isEmpty() || objectUids.isEmpty() )
+        if ( userOrgUnitPaths.isEmpty() )
         {
             return new HashSetValuedHashMap<>();
         }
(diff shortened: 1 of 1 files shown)


netroms (approved):

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
```diff
@@ -58,6 +58,11 @@ public SetValuedMap<String, String> getOrganisationUnitsAssociationsForCurrentUs
     {
         Set<String> userOrgUnitPaths = getUserOrgUnitPaths();
 
+        if ( objectUids.isEmpty() )
+        {
+            return new HashSetValuedHashMap<>();
+        }
+
```
larshelge, line 65: Looks good.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
```diff
@@ -58,6 +58,11 @@ public SetValuedMap<String, String> getOrganisationUnitsAssociationsForCurrentUs
...
         return jdbcTemplate.query(
```
larshelge, line 66: What about the other method `getOrganisationUnitsAssociations`, should we have the same fix there, as it seems it has the same weakness?

PR #11908, opened 2022-09-26
feat: Program Attributes are now all presents with programUid prefix [DHIS2-13814]

gnespolino:
Before this was implemented, CPL query dimension endpoint was returning
- Tracked Entity Attributes as single-level UID identifier
- Program Attribute as UID prefixed with programUid (but removing all tracked entity attribute already present in the above line).

Now we only return PA prefixed with programUid.

I also added an optional program parameter in the endpoint

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/DefaultTeiAnalyticsDimensionsService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/DefaultTeiAnalyticsDimensionsService.java
@@ -27,10 +27,6 @@
  */
 package org.hisp.dhis.analytics.tei;
 
-import static org.hisp.dhis.analytics.event.data.DimensionsServiceCommon.OperationType.QUERY;
-import static org.hisp.dhis.analytics.event.data.DimensionsServiceCommon.filterByValueType;
-import static org.hisp.dhis.common.PrefixedDimensions.ofItemsWithProgram;
-
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tei/TeiQueryController.java
```diff
@@ -136,6 +136,7 @@ Grid getGrid(
     @GetMapping( "/query/dimensions" )
     public AnalyticsDimensionsPagingWrapper<ObjectNode> getQueryDimensions(
         @RequestParam String trackedEntityType,
+        @RequestParam( required = false ) List<String> program,
```
maikelarabori, line 139: I would use a Set, as the program cannot be repeated.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/DefaultTeiAnalyticsDimensionsService.java
```diff
@@ -73,37 +66,29 @@ class DefaultTeiAnalyticsDimensionsService implements TeiAnalyticsDimensionsServ
     private final ProgramService programService;
 
     @Override
-    public List<PrefixedDimension> getQueryDimensionsByTrackedEntityTypeId( String trackedEntityTypeId )
+    public List<PrefixedDimension> getQueryDimensionsByTrackedEntityTypeId( String trackedEntityTypeId,
+        Collection<String> programUids )
```
maikelarabori, line 70: Instead of using Collection, I would use something more specific, like Set in this case (as the program cannot be repeated).

maikelarabori (approved):
Looks good, just two minor comments @gnespolino :+1: 

Cheers,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/DefaultTeiAnalyticsDimensionsService.java
gnespolino: done

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tei/TeiQueryController.java
gnespolino: done

PR #11915, opened 2022-09-27
fix: Properly validate enum values in the event exporter endpoint

enricocolasante:
This PR is just a POC to check if it would make sense to use javax validation framework to return better error responses to the user.
I added a validator for enums that will return a 400 HTTP Status if the value assigned to the field represented by an Enum is not a valid value


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerEventCriteria.java
```diff
@@ -59,17 +59,20 @@ class TrackerEventCriteria extends PagingAndSortingCriteriaAdapter
 
     private String programStage;
 
-    private ProgramStatus programStatus;
+    @ValueOfEnum( enumClass = ProgramStatus.class )
+    private String programStatus;
```
teleivo, line 63: Going from an enum to a String in our type is a step back in type safety. I assume this is necessary to capture the value the user actually supplied if its not one of the possible enum values, right?

teleivo (commented):
Thank you for exploring how we can improve our error messages/validation 😄 

I thought about using javax validation some time ago. I remember discussing it with @jbee at least with regards to the tracker importer 🤔 If I recall correctly @jbee concern was that it would cause the validations to be split between javax validation on the controller and validations in our validators.

In our GET endpoints we don't have the same validators but right now validation seems to be done in the criteria mappers. What I don't know is if we have some validations that we cannot implement in javax validations. This would cause such a split between validations via annotations on the criteria classes and some implemented in the criteria mappers.

For example: can we validate `filter` only contains TEA UIDs that exist in the DB?

enricocolasante:
> Thank you for exploring how we can improve our error messages/validation smile
> 
> I thought about using javax validation some time ago. I remember discussing it with @jbee at least with regards to the tracker importer thinking If I recall correctly @jbee concern was that it would cause the validations to be split between javax validation on the controller and validations in our validators.
> 
> In our GET endpoints we don't have the same validators but right now validation seems to be done in the criteria mappers. What I don't know is if we have some validations that we cannot implement in javax validations. This would cause such a split between validations via annotations on the criteria classes and some implemented in the criteria mappers.
> 
> For example: can we validate `filter` only contains TEA UIDs that exist in the DB?

Thank you for the feedback, I created this PR to look how it would work on the code having javax validators and to understand pro and cons.
I agree with @jbee concern, things will be split in 2 but if you have a deeper look at our validators in tracker we are implicitly separate the concepts of syntax and semantic validators.
In my head we could have javax validators to do the syntax validation in an almost transparent way, so we will have our criteria object, with annotations and at least we need that if we arrive to the mapper stage all our formats are correct, and from there we can validate accessing the DB and other validations.
Of course I don't like that in the PR we have everything as a String but I couldn't find a better way and I am not sure there is one.

Talking about Enums, how it is now in the API is something that we should change anyway, if the client use a value that is not present in the Enum the server respond with 500 and a very verbose message that is not really clear.
Maybe is just a matter of intercepting the right Exception in the `CrudControllerAdvice`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerEventsExportController.java
```diff
@@ -106,7 +107,7 @@
 
     @GetMapping( produces = APPLICATION_JSON_VALUE )
     public PagingWrapper<ObjectNode> getEvents(
-        TrackerEventCriteria eventCriteria, HttpServletRequest request,
+        @Valid TrackerEventCriteria eventCriteria, HttpServletRequest request,
```
jbee, line 110: What if this `@Valid` would be missing? Would a non valid value be processed?

jbee (commented):
Generally I can get on board with doing syntax level validation in a different way and phase than the semantic validation. But if we use `String`s this falls a bit flat as there is no guarantee that the strings we pass and or call next level with are valid syntax wise so to be correct the service would have to again make the syntax level changes at which point this feels silly. 
It would be nice if we use this as an argument to create and use more specific types.
The syntax level validation can occur in the constructor of the type making sure only valid values of the type can be produced. Which is in a way what `enum`s do in a basic form.
Wen such values are then passed to a service the syntax level validation has for sure occurred and does not need to be repeated. 

What we need to find is a good way to have this integrated in spring/jackson so that errors caused by syntax level problems result in nice messages and proper error codes.

When you said you could not find another way than using `String` for the enum fields is that to make validation work or for other reasons? Our goal should be to have true enum fields where possible and adjust our validation it, not the other way around.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerEventCriteria.java
larshelge: This might not be an ideal solution, i.e. moving from a strongly-typed to weaker typing. Could there be another solution where we generically handle this problem for enums?

PR #11916, opened 2022-09-27
fix: Use custom serializer which skips API credentials [DHIS2-13105]

larshelge:
Introduces a custom serializer `ApiSerializer` which skips API credentials for aggregate data exchange.

This serializer exists because`Api` is persisted as a JSONB data type which relies on creating a deep copy by writing and reading the class to JSON, in which case the use of `JsonProperty.Access.WRITE_ONLY` at the property level leads to the property value being lost in the copy process.


jbee (approved):

mortenoh (approved):
Approved. Would be good to use the normal Jackson functionality here, but as you said that's not possible in this case.

PR #11920, opened 2022-09-28
fix: check for empty objectUids (#11891)

mortenoh:
## Summary

Checks for empty object collection before running SQL query. Needed so that we don't do invalid `in (...)` queries.

[TECH-1437](https://dhis2.atlassian.net/browse/TECH-1437)


Philip-Larsen-Donnelly:
Approved for 2.39.0 by RCB:
Lars, Phil

PR #11921, opened 2022-09-28
fix: Filter can have multiple value [DHIS2-13676]

abyot:
Event filter can have multiple values e.g.`filter=qrur9Dvnyt5:ge:10:le:20`


enricocolasante (approved):
I just edit the test that I created in the previous PR to cover the case that you fixed

PR #11932, opened 2022-09-29
fix: Use custom serializer which skips API credentials [DHIS2-13105]

larshelge:
Introduces a custom Jackson serializer `ApiSerializer` which skips API credentials for aggregate data exchange.

This serializer exists becauseApi is persisted as a JSONB data type which relies on creating a deep copy by writing and reading the class to JSON, in which case the use of `JsonProperty.Access.WRITE_ONLY` at the property level leads to the property value being lost in the copy process.

Registering a custom serializer for the mapper used in the API, but not in JSONB conversion, solves the problem. This is not the ideal solution and a better, more standard solution could be investigated.


Philip-Larsen-Donnelly:
Approved for 2.39.0 by RCB:
Lars, Morten, Phil

PR #11958, opened 2022-09-30
fix: Avoid sql operator replacement in tracker [DHIS2-13840]

muilpp:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java
```diff
@@ -113,21 +113,44 @@ public boolean isOperator( QueryOperator op )
     }
 
     public String getSqlOperator()
+    {
+        return getSqlOperator( false );
+    }
+
+    /**
+     *
+     * @param isOperatorSubstitutionAllowed whether the operator should be
```
maikelarabori, line 122: I think it would be nice to have the full Javadoc. A general description + the return definition.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java
```diff
@@ -113,21 +113,44 @@ public boolean isOperator( QueryOperator op )
...
         {
             return null;
         }
 
-        return safelyGetOperator();
+        return safelyGetOperator( isOperatorSubstitutionAllowed );
     }
 
-    private String safelyGetOperator()
+    private String safelyGetOperator( boolean isOperatorSubstitutionAllowed )
```
maikelarabori, line 136: Should we also use a more specific name for this flag, as you did below? `isNullValueSubstitutionAllowed`

maikelarabori (approved):
Thanks, @muilpp, it looks good 👍 I just left two small comments. Please, see what you think.

Cheers,
Maikel

stian-sandvold (approved):
Looks good to me!
I agree with the comments of @maikelarabori as well, a short comment on the new method would be great!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java
muilpp: I named both flags different because I wanted to highlight the substitution of an operator and the substitution of the text NV.
From my point of view I think this flag name is fine, but I've been working in this ticket, so maybe it's still fresh in my mind. Do you have a better suggestion for a name?

maikelarabori: Right, it makes sense Marc.

PR #11993, opened 2022-10-03
chore: avoid mockito strictness LENIENT [TECH-1195]

jbee:
Made sense to check that for an enabled filter the session is used while it is not if disabled.

### DIFF
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/servlet/filter/RequestIdentifierFilterTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/servlet/filter/RequestIdentifierFilterTest.java
@@ -34,6 +34,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import java.io.IOException;
+import java.util.function.Consumer;
 
 import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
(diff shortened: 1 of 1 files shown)


enricocolasante (approved):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/servlet/filter/RequestIdentifierFilterTest.java
```diff
@@ -88,23 +87,24 @@ void testIsEnabled()
...
     {
         init( true );
-        doFilter();
+        doFilter( request -> {
+            HttpSession session = mock( HttpSession.class );
+            when( request.getSession() ).thenReturn( session );
+            when( session.getId() ).thenReturn( "ABCDEFGHILMNO" );
+        } );
 
         assertNotNull( MDC.get( "sessionId" ) );
```
teleivo, line 96: would it also make sense that we put the session id `ABCDEFGHILMNO` into the context?

teleivo (approved):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/servlet/filter/RequestIdentifierFilterTest.java
jbee: Not sure what you suggest. Please elaborate.

teleivo: I don't know the code we are testing. I just based it on the class name and that we are asserting on MDC. I assume what we are testing is responsible for putting the session id into the MDC (context) that is then used to log the session id.

https://www.baeldung.com/mdc-in-log4j-2-logback

That's why I thought should we not also check that the value of `MDC.get( "sessionId" )` is the session id `ABCDEFGHILMNO`

jbee: Ah, yes - that makes sense to me.

jbee: Added the change to #11997 as that so far had just comment changes.

teleivo: Thank you 😄

PR #11995, opened 2022-10-03
chore: avoid mockito strictness LENIENT #5 [TECH-1195]

jbee:
`noLinks` test scenario did unnecessary mocking that I simply removed.

Also updated to use `List.of` and `Map.of` which showed that the pager next/prev URLs are dependent on the order in the map of parameters. To preserve the order that map presents the parameters assuming this is the order they where present in the URL originally I opted to add a dedicated new assert method that would compare relative URLs for functional equality allowing the parameters to occur in a different order than in the expected URL string.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
@@ -27,6 +27,8 @@
  */
 package org.hisp.dhis.utils;
 
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toUnmodifiableList;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -37,6 +39,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 import org.hisp.dhis.common.ErrorCodeException;
 import org.hisp.dhis.commons.collection.CollectionUtils;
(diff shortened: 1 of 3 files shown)


enricocolasante (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -200,4 +203,36 @@ public static void assertLessOrEqual( long upper, long actual )
...
+     *
+     * <pre>
+     * /context/endpoint?a=b&c=d
+     * /context/endpoint?c=d&a=b
+     * </pre>
+     *
+     * @param expected the expected URL with path and optional parameters
+     * @param actual the actual URL with path and optional parameters
+     */
+    public static void assertEquivalentRelativeUrls( String expected, String actual )
```
teleivo, line 221: nice 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/service/DefaultLinkService.java
```diff
@@ -69,18 +70,9 @@ public class DefaultLinkService implements LinkService
 
     private final ContextService contextService;
 
-    public DefaultLinkService( SchemaService schemaService, ContextService contextService )
-    {
-        checkNotNull( schemaService );
```
teleivo, line 74: Should the dependencies then get a `@NonNull` annotation if you remove the check? as in https://github.com/dhis2/wow-backend/blob/master/guides/spring_conventions.md#injection

teleivo (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/service/DefaultLinkService.java
jbee: We removed these checks in beans when switching to lombok annotations. Use of `@NonNull` (or any equivalent annotation) is a general question we might want to discuss in Thursday meeting. I think we could benefit using it where we feel it helps but we should agree which of the available annotations we start using.

PR #11996, opened 2022-10-03
chore: avoid mockito strictness LENIENT #6 [TECH-1195]

jbee:
Easiest seemed to unmock common setup using `reset` and just add the needed `when` back for the single case where common setup would not work as this tests a error case that exists early.


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/user/UserControllerTest.java
```diff
@@ -297,6 +290,8 @@ void updateUserExpireRequiresShareBasedAuthority()
     {
         setUpUserExpireScenarios();
         when( aclService.canUpdate( currentUser, user ) ).thenReturn( false );
+        reset( userService );
```
teleivo, line 293: setUp does not setup any mocks. So I assume setUpUserExpireScenarios contains mocks we do not need. Can we adapt that instead of using `reset()`.

jbee: Can yes. Best idea I have is to pass a flag and skip the part we reset here if that flag is set but didn't found that better in this case. Other ideas are welcome. Maybe we could do what I did in another of these PRs and pass a consumer so the mocking of the particular service can be customised. But here it is more than one service and it seems quite arbitrary that one can be customised while the others can't.

teleivo: We also said if the strict mode does not make the test more readable we are just going to keep the lenient setting. So I vote for keeping it lenient instead of the use of `reset()` 😄 I would not want to encourage new tests being written relying on `reset`.

jbee: Her it was one scenario out of many failing with strict so I felt it was worth adjusting that scenario.

teleivo: I understand, its just that I would not want someone writing a new test using `reset`. Instead have them only setup the mocks they need. That's why I would not want to promote `reset` by putting it into any test. I abstain from my vote and let @enricocolasante and @netroms make the decision 😋

enricocolasante: I would rather create a new setup only for this test or if we do not want to duplicate this code just for one test I would put the lenient in front of the guilty mock method like
`lenient().when( userService.canAddOrUpdateUser( any(), any() ) ).thenReturn( true );`

The use of reset to make the test being strict seems quite weird to me

jbee: I like this one:

> lenient().when( userService.canAddOrUpdateUser( any(), any() ) ).thenReturn( true );

enricocolasante (approved):

teleivo (approved):

PR #12006, opened 2022-10-04
fix: Name csv in zip file events.csv [DHIS-10376]

muilpp:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/EventController.java
```diff
@@ -652,6 +654,7 @@ public void getCsvEvents(
             response.addHeader( ContextUtils.HEADER_CONTENT_TRANSFER_ENCODING, "binary" );
             outputStream = new GZIPOutputStream( outputStream );
             response.setContentType( "application/csv+gzip" );
+            response.setHeader( HEADER_ENTITY_TYPE, "events" );
```
mortenoh, line 657: Would it make more sense to parse this out from the `HEADER_CONTENT_DISPOSITION` header ? (if present)

mortenoh (approved):

stian-sandvold (approved):

PR #12007, opened 2022-10-04
fix: Check for potential null values - [ DHIS2-13673 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-13673
fix npe mostly using Optional


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java
```diff
@@ -135,14 +136,10 @@ public String getSqlOperator( boolean isOperatorSubstitutionAllowed )
 
     private String safelyGetOperator( boolean isOperatorSubstitutionAllowed )
     {
-        Function<Boolean, String> operatorFunction = OPERATOR_MAP.get( operator );
-        if ( operatorFunction != null )
```
teleivo, line 139: Does this not already handle null? This change does not seem necessary 😄

dhis-2/dhis-api/src/main/java/org/hisp/dhis/message/MessageConversation.java
```diff
@@ -148,13 +148,15 @@ public void addMessage( Message message )
         if ( message != null )
         {
             message.setAutoFields();
-        }
 
-        this.messages.add( message );
```
teleivo, line 153: This changes the behavior int that we now will not add `null` messages. As I don't know what this code is for it might make sense, but might also not.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/resolver/CategoryOptionGroupResolver.java
```diff
@@ -97,17 +97,16 @@ private String getExpression( String expression, DimensionalItemId id, List<Stri
     {
         List<String> cocUidIntersection = getCategoryOptionCombosIntersection( cogUidList, dataElementId );
 
-        if ( isEmpty( cocUidIntersection ) )
+        if ( Optional.ofNullable( cocUidIntersection ).filter( i -> !i.isEmpty() ).isPresent() )
```
teleivo, line 100: Empty strings will still be in the `resolved` collection below in case there are also non-empty strings in `cocUidIntersection`. I wonder if this filtering should not rather move into `getCategoryOptionCombosIntersection` instead.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/fileresource/DefaultImageProcessingService.java
```diff
@@ -77,7 +78,8 @@ public Map<ImageFileDimension, File> createImages( FileResource fileResource, Fi
                     continue;
                 }
 
-                ImageSize size = IMAGE_FILE_SIZES.get( dimension );
+                ImageSize size = Optional.ofNullable( IMAGE_FILE_SIZES.get( dimension ) )
```
teleivo, line 81: I would go for a good old if/else instead of an Optional in cases like this. I find it easier to read 😅. Also wrapping something in an Optional to be unwrapped immediately does not make much sense to me if it does not improve readability.

```java
ImageSize size = IMAGE_FILE_SIZES.get( dimension );
if (size == null) {
    size = new ImageSize( 0, 0 );
}
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramInstanceService.java
```diff
@@ -387,23 +388,9 @@ public ProgramInstance prepareProgramInstance( TrackedEntityInstance trackedEnti
         programInstance.setOrganisationUnit( organisationUnit );
         programInstance.enrollTrackedEntityInstance( trackedEntityInstance, program );
 
-        if ( enrollmentDate != null )
```
teleivo, line 390: I would not change code that handles null already just for the sake of transitioning to Optional.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsGateway.java
```diff
@@ -138,29 +138,22 @@ public HttpStatus send( String urlTemplate, HttpEntity<?> request, HttpMethod ht
 
     public OutboundMessageResponse wrapHttpStatus( HttpStatus httpStatus )
     {
-        GatewayResponse gatewayResponse;
+        GatewayResponse gatewayResponse = getGatewayResponse( httpStatus );
 
         OutboundMessageResponse status = new OutboundMessageResponse();
-
-        if ( OK_CODES.contains( httpStatus ) )
```
teleivo, line 145: Is this just a refactoring? To me it seems that the code already behaves in the same way, unless I miss a detail 😅

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataImportService.java
```diff
@@ -223,7 +224,7 @@ public MetadataImportParams getParamsFromMap( Map<String, List<String>> paramete
         {
             User overrideUser = null;
 
-            if ( parameters.containsKey( "overrideUser" ) )
+            if ( Optional.ofNullable( parameters ).filter( p -> p.containsKey( "overrideUser" ) ).isPresent() )
```
teleivo, line 227: Is `parameters` ever null? Our code can be pretty defensive at times. I often found the solution is to make sure collections are non-null by default.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/context/ProgramInstanceSupplier.java
```diff
@@ -80,26 +81,25 @@ public Map<String, ProgramInstance> get( ImportOptions importOptions,
...
-            // Create a bi-directional map enrollment uid -> event uid
-            Multimap<String, String> programInstanceToEvent = HashMultimap.create();
-            for ( Event event : events )
-            {
-                programInstanceToEvent.put( event.getEnrollment(), event.getUid() );
-            }
+            .map( Event::getEnrollment )
+            .filter( StringUtils::isNotEmpty ).collect( Collectors.toSet() );
+
+        Map<String, ProgramInstance> programInstances = Optional.of( programInstanceUids )
```
teleivo, line 87: Is `programInstanceUids` ever null? Should it not rather be an empty set if there are no events with a non empty enrollment 🤔

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/context/ProgramInstanceSupplier.java
```diff
@@ -151,7 +151,7 @@ private void mapExistingEventsToProgramInstances( ImportOptions importOptions, L
         // Collect all the Program Instances by event uid
         final Map<String, ProgramInstance> programInstancesByEvent = getProgramInstanceByEvent( importOptions, events );
```
teleivo, line 152: Same here as in another place, I prefer we ensure that `programInstancesByEvent` is never null. This makes the code safer and easier to read. We are too defensive 😬

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/context/ProgramSupplier.java
```diff
@@ -150,7 +150,7 @@ public Map<String, Program> get( ImportOptions importOptions, List<Event> eventL
 
             programMap = loadPrograms( importOptions.getIdSchemes() );
 
-            if ( !MapUtils.isEmpty( programMap ) )
+            if ( Optional.ofNullable( programMap ).filter( pm -> !MapUtils.isEmpty( pm ) ).isPresent() )
```
teleivo, line 153: `MapUtils.isEmpty` is already null safe. I might not get why this is needed 😅

teleivo (commented):
Some of the changes do not seem necessary to me. They just turn if/else statements into an Optional chain. IMHO that often makes the code harder to read 😅 One could also argue less efficient as we instantiate an Optional to unwrap it immediately anyway.

AFAIK the issue is about making the code safer. So to me the PR should only contain necessary changes to accomplish that. Can we for example ensure that the collections we operate on are never null. That would make the code safer and more readable.

Maybe we should discuss the use of Optional in the backend meeting as it feels like there is a camp of devs going all in and another that is more Optional skeptic 😂

lucaCambi77:
I personally prefer using Optional as I feel it makes it less prone to error but that is my personal opinion. I've followed sonar and ide suggestions and honestly, I am not going through all the issues again as it takes time. So also I'd suggest, feel free to make changes to this branch or discard it.

jbee (changes requested):
We talked about use of `Optional` in today's Thursday meeting. There was mostly consent that `Optional` should not be used as a replacement for if/else and alike within the body of a method implementation. As a rule of thumb it was suggested to think twice before using `Optional.ofNullable` as this usually is the case when replacing other language statement types. `Optional` primary use case is to avoid returning `null` from a method and to communicate the possibility of "no value" to the caller.

I might be wrong but I think the basis of for your changes are sonar warnings of potential NPEs. This analysis isn't always correct as the assumptions are based on a mix of code analysis and accepting that annotations express. We should think about striking a balance between defensive coding and readability. Adding too much defensive extra code makes code harder to read which in turn usually comes with an increased risk of introducing other issues. In our discussion we also had conses that it is better to design methods so they would never return `null` instead so we don't have to get defensive. We also talked about using javax `@NotNull` to support static analysis so it can help us find potential NPEs and that we should use more `Optional` as return type when a method might not return anything, or have it throw an exception.

Last but not least, NPEs are annoying but they are also easy to find and fix. Usually code matures quickly once it is tested. The weight of defensive code on the other hand stays.

PR #12009, opened 2022-10-04
feat: Program Attribute and TEI OU filter implementation [DHIS2-13777,DHIS2-13828]

gnespolino:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/analytics/common/dimension/DimensionParamItem.java
```diff
@@ -81,13 +81,9 @@ public static List<DimensionParamItem> ofStrings( List<String> items )
     private static QueryOperator getOperator( String operator )
     {
         return Arrays.stream( QueryOperator.values() )
-            .filter( queryOperator -> queryOperator.name().equals( operator ) )
+            .filter( queryOperator -> equalsIgnoreCase( queryOperator.name(), operator ) )
             .findFirst()
-            .orElseThrow( IllegalAccessError::new );
+            .orElseThrow( () -> new IllegalArgumentException( "Invalid operator: " + operator ) );
```
maikelarabori, line 86: I believe we could throw something like `new IllegalQueryException( ErrorCode.E2035 );`
This should be automatically handled by the controller layer.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/InConditionRenderer.java
```diff
@@ -0,0 +1,65 @@
...
+    public String render()
+    {
+        if ( values.size() > 1 )
+        {
+            return doubleQuote( field ) + SPACE + QueryOperator.IN.getValue() + " ("
+                + queryContext.bindParamAndGetIndex( valueTypeMapping.convertMany( values ) ) + ")";
+        }
+        else
+        {
+            return EqConditionRenderer.of( field, List.of( values.get( 0 ) ), valueTypeMapping, queryContext ).render();
```
maikelarabori, line 62: I'm not sure if we should use `EqConditionRenderer` inside this class (InConditionRenderer).
Maybe we could only generate IN statement here?

maikelarabori (approved):
Approved with a few comments for review @gnespolino :+1: 
_There are some build issues to be checked as well._

Thanks!

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/shared/SqlQuery.java
```diff
@@ -51,6 +53,9 @@ public SqlQuery( final String statement, final Map<String, Object> params )
 
         this.statement = statement;
         this.params = params;
+
+        log.info( "STATEMENT: " + statement );
```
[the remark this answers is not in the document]
gnespolino: yes for sure, I need to configure my local logger to print debug

d-bernat (approved):

PR #12017, opened 2022-10-05
fix: Event program relationships can now be added through maintenance app

stian-sandvold:
In the case of a program without registration, there is only one program stage. As such, the UI only presents the program, and sends it in.

This fix will populate the ProgramStage value with that single program stage of an event program if its not present.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/RelationshipTypeController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/RelationshipTypeController.java
@@ -27,6 +27,9 @@
  */
 package org.hisp.dhis.webapi.controller.event;
 
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.program.ProgramService;
+import org.hisp.dhis.relationship.RelationshipConstraint;
 import org.hisp.dhis.relationship.RelationshipType;
 import org.hisp.dhis.schema.descriptors.RelationshipTypeSchemaDescriptor;
 import org.hisp.dhis.webapi.controller.AbstractCrudController;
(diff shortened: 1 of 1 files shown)


enricocolasante (approved):

lucaCambi77 (commented):
Any way to add a test?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/RelationshipTypeController.java
```diff
@@ -41,4 +44,53 @@
...
+     * We do this because programs without registration only have a single
+     * program stage
+     *
+     * @param constraint
+     */
+    private void handleEventProgramRelationshipConstraint( RelationshipConstraint constraint )
+    {
+        if ( constraint.getProgram() != null && constraint.getProgramStage() == null )
+        {
+            Program program = programService.getProgram( constraint.getProgram().getUid() );
```
teleivo, line 88: Can we safely assume the idScheme is UID here?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/RelationshipTypeController.java
```diff
@@ -41,4 +44,53 @@
...
+
+            if ( program.isWithoutRegistration() )
+            {
+                constraint.setProgramStage( program.getProgramStageByStage( 1 ) );
```
teleivo, line 92: What does the `1` mean here? Might be worth documenting in a constant.

teleivo (commented):
This change only affects old tracker, is that intentional? 😄 
👍🏻 to @lucaCambi77 for adding a test 😋

PR #12021, opened 2022-10-05
fix: various DataValueAudit updates

mortenoh:
## Summary

- [x] Use `dv.value` instead of `dv.auditValue` when undeleting a value.
- [x] Only set `created` for completely new data values.
- [x] DIsplay audits even if `dv.deleted` is `true`.

[TECH-1352](https://dhis2.atlassian.net/browse/TECH-1352)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/datavalue/DataValueAudit.java
```diff
@@ -116,6 +116,22 @@ public static DataValueAudit from( DataValueAudit dataValueAudit )
             dataValueAudit.getAuditType() );
     }
 
+    public static DataValueAudit from( DataValueAudit dataValueAudit, Date created )
```
jbee, line 119: Feels like `@Builder` with `toBuilder()` might be a more concise way doing this.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/datavalue/DataValueStore.java
```diff
@@ -109,6 +109,21 @@
...
+     * @param dataElement the DataElement of the DataValue.
+     * @param period the Period of the DataValue.
+     * @param source the Source of the DataValue.
+     * @param categoryOptionCombo the category option combo.
+     * @param attributeOptionCombo the attribute option combo.
+     * @param includeDeleted Include deleted data values
+     * @return the DataValue which corresponds to the given parameters, or null
+     *         if no match.
+     */
+    DataValue getDataValue( DataElement dataElement, Period period, OrganisationUnit source,
```
jbee, line 124: Following what we said during Thursday meeting this is a "find" type of method but I guess there are already exist other and naming these now inconsistent is also bad. We should try to update this at some point to follow the get/find pattern.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/hibernate/HibernateDataValueStore.java
```diff
@@ -143,9 +150,16 @@ public DataValue getDataValue( DataElement dataElement, Period period, Organisat
...
 
-        String hql = "select dv from DataValue dv  where dv.dataElement =:dataElement and dv.period =:period and dv.deleted = false  "
-            +
-            "and dv.attributeOptionCombo =:attributeOptionCombo and dv.categoryOptionCombo =:categoryOptionCombo and dv.source =:source ";
+        String includeDeletedSql = "";
+
+        if ( !includeDeleted )
+        {
+            includeDeletedSql = "and dv.deleted = false   ";
+        }
```
jbee, line 158:
```suggestion
String includeDeletedSql = includeDeleted ? "" : "and dv.deleted = false ";
```

jbee (approved):
Some suggestions:

larshelge (approved):

PR #12022, opened 2022-10-05
feat: Implement sorting by TEI OU in Cross Program Line listing Cpl development 13778

d-bernat:
see: https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/boards/104?modal=detail&selectedIssue=DHIS2-13778&sprint=333&quickFilter=497


maikelarabori (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/TeiFields.java
```diff
@@ -149,4 +169,18 @@ public static Set<GridHeader> getGridHeaders( final TeiQueryParams teiQueryParam
...
+     * Static element is term for the parameter used directly as a database
+     * table column name (example: OU, uidlevel1, ..) Dynamic elements are for
+     * example columns with uid strings
+     *
+     * @param p AnalyticsSortingParas
+     * @return boolean
+     */
+    private static boolean isDynamicElement( DimensionIdentifier<Program, ProgramStage, DimensionParam> p )
+    {
+        return (p.hasProgram() || p.hasProgramStage()) &&
```
gnespolino, line 183: this is fine, however no dimension exists having only the programStage (so if program is null, then also programStage should be null)

gnespolino (approved):
Looks ok, most probably we'll need to change something at later time, but it's good for now

PR #12023, opened 2022-10-05
fix: Broken user group add and removal in user app

netroms:
### Summary

The user app was broken and did not work when trying to add or remove a user group on the user.
This was broken because the new patch method in the AbstractCrudController, that now the UI uses, don't support inverted relationship operations. 
To fix this, the relationship had be inverted and let the UserGroup service handle the operations, by using the pre and post patch hooks.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
+++ b/dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
@@ -40,6 +40,8 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * Utility methods for operations on various collections.
@@ -290,5 +292,17 @@ public class CollectionUtils
             Map.entry( k9, v9 ),
             Map.entry( k10, v10 ),
             Map.entry( k11, v11 ) );
+    }
+
+    /**
+     * Convert an Iterator to a Stream.
+     *
+     * @param iterator The Iterator to convert to a stream.
+     * @return A stream of the iterable.
+     */
+    public static <E> Stream<E> iterableToStream( Iterator<E> iterator )
+    {
+        Iterable<E> iterable = () -> iterator;
+        return StreamSupport.stream( iterable.spliterator(), false );
     }
 }
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
@@ -50,6 +50,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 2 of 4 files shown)


dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
```diff
@@ -291,4 +293,16 @@ public static <K, V> Map<K, V> mapOf(
...
+    /**
+     * Convert an Iterator to a Stream.
+     *
+     * @param iterator The Iterator to convert to a stream.
+     * @return A stream of the iterable.
+     */
+    public static <E> Stream<E> iterableToStream( Iterator<E> iterator )
+    {
+        Iterable<E> iterable = () -> iterator;
+        return StreamSupport.stream( iterable.spliterator(), false );
```
jbee, line 306: Better use : 
```java
StreamSupport.stream(spliteratorUnknownSize(iter, Spliterator.ORDERED | Spliterator.NONNULL | Spliterator.IMMUTABLE), false);
```
`Spliterator.NONNULL` might not apply.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -816,13 +823,31 @@ protected void prePatchEntity( User oldEntity, User newEntity )
...
         {
-            userService.expireActiveSessions( user );
+            userService.expireActiveSessions( entityAfter );
+        }
+
+        if ( entityAfter != null && patch != null )
+        {
+            for ( JsonPatchOperation op : patch.getOperations() )
+            {
+                JsonPointer userGroups = op.getPath().matchProperty( "userGroups" );
```
mortenoh, line 839: I wonder if its better to match on `/userGroups` as we just want this supported at the root I guess?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -816,13 +823,31 @@ protected void prePatchEntity( User oldEntity, User newEntity )
...
+                String opName = op.getOp();
+                if ( userGroups != null && opName.equals( "add" ) )
```
mortenoh, line 841: Check for `replace` also, as in this case they have the same semantics

mortenoh (changes requested):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
netroms: Yeah, that's a valid point. I will change.

jbee (approved):
We should think about how we can solve situation like this without adding workarounds.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
netroms: ok

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
netroms: Ok

mortenoh (approved):

PR #12026, opened 2022-10-05
chore: Rename group to userGroup

netroms:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -225,7 +225,7 @@
 
     private Date lastCheckedInterpretations;
 
-    private Set<UserGroup> groups = new HashSet<>();
+    private Set<UserGroup> userGroups = new HashSet<>();
```
larshelge, line 228: Is there any backwards API compatibility issues with this change? Or the JSON data model stays the same?

netroms: JSON model should stay the same. This is just for code clarity. So the code matches the output model naming.

PR #12035, opened 2022-10-06
chore: use @Nonnull and @CheckForNull in generic store APIs

jbee:
I thought we need to start somewhere with `@Nonnull` why not start at the low level APIs.

I also found one case of a `find` method that could rather use `Optional` as return type to avoid `null`.

There were some methods called `getXzy` that would return `null` and hence should either become a _find_ using `Optional` or throw an exception but I wanted to keep this change mostly about adding `@Nonnull`. 

There are some cases where I wasn't 100% sure if null would be possible return value where I did not add the annotation.
Some of the parameters would actually handle null but semantically should never be called with `null` in which case I did annotate with `@Nonnull`. 

**Edit:**
After first CI run it became apparent that some of the generics were not properly propagated.
I then started doing that work and found a handful of methods that were not returning empty lists but `null` and 1 which was not throwing an exception but returned `null` when it shouldn't.

**Edit2:**
`@NotNull` was wrong annotation to use specifically for java validation. 
Therefore switched to javax `@Nonnull` which also has a counterpart `@CheckForNull`.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -198,8 +198,8 @@ public interface IdentifiableObjectManager
         throws IllegalQueryException;
 
     @NotNull
-    List<IdentifiableObject> getByUid(
-        @NotNull Collection<Class<? extends IdentifiableObject>> types,
+    <T extends IdentifiableObject> List<T> getByUid(
+        @NotNull Collection<Class<? extends T>> types,
         @NotNull Collection<String> uids );
 
     <T extends IdentifiableObject> @NotNull List<T> getById( @NotNull Class<T> type, @NotNull Collection<Long> ids );
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -78,7 +78,7 @@ public interface IdentifiableObjectManager
      * @return The {@link IdentifiableObject} with the given UID
      */
     @NotNull
-    Optional<IdentifiableObject> find( @NotNull String uid );
+    Optional<? extends IdentifiableObject> find( @NotNull String uid );
 
     <T extends IdentifiableObject> @NotNull T get( @NotNull Class<T> type, long id );
 
@@ -120,10 +120,10 @@ public interface IdentifiableObjectManager
 
     <T extends IdentifiableObject> boolean exists( @NotNull Class<T> type, @NotNull String uid );
 
-    <T extends IdentifiableObject> T get( @NotNull Collection<Class<? extends IdentifiableObject>> types,
+    IdentifiableObject get( @NotNull Collection<Class<? extends IdentifiableObject>> types,
         @NotNull String uid );
 
-    <T extends IdentifiableObject> T get( @NotNull Collection<Class<? extends IdentifiableObject>> types,
+    IdentifiableObject get( @NotNull Collection<Class<? extends IdentifiableObject>> types,
         @NotNull IdScheme idScheme,
         String value );
 
@@ -198,7 +198,8 @@ public interface IdentifiableObjectManager
         @NotNull Collection<String> uids )
         throws IllegalQueryException;
 
-    <T extends IdentifiableObject> @NotNull List<T> getByUid(
+    @NotNull
+    List<IdentifiableObject> getByUid(
         @NotNull Collection<Class<? extends IdentifiableObject>> types,
         @NotNull Collection<String> uids );
 
@@ -279,16 +280,15 @@ public interface IdentifiableObjectManager
     <T extends IdentifiableObject> List<T> getByAttributeAndValue( @NotNull Class<T> type, @NotNull Attribute attribute,
         @NotNull String value );
 
-    <T extends IdentifiableObject> boolean isAttributeValueUnique( @NotNull Class<? extends IdentifiableObject> type,
+    <T extends IdentifiableObject> boolean isAttributeValueUnique( @NotNull Class<T> type,
         @NotNull T object,
         @NotNull AttributeValue attributeValue );
 
-    <T extends IdentifiableObject> boolean isAttributeValueUnique( @NotNull Class<? extends IdentifiableObject> type,
-        @NotNull T object,
-        @NotNull Attribute attribute, @NotNull String value );
+    <T extends IdentifiableObject> boolean isAttributeValueUnique( @NotNull Class<T> type,
+        @NotNull T object, @NotNull Attribute attribute, @NotNull String value );
 
     @NotNull
-    List<? extends IdentifiableObject> getAllByAttributeAndValues( @NotNull Class<? extends IdentifiableObject> type,
+    <T extends IdentifiableObject> List<T> getAllByAttributeAndValues( @NotNull Class<T> type,
         @NotNull Attribute attribute, @NotNull List<String> values );
 
     @NotNull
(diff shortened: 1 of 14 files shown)


jbee:
Found that there are many type parameters that need propagation. Will add this to this PR.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -327,100 +317,78 @@ public <T extends IdentifiableObject> boolean exists( Class<T> type, String uid
...
-    @SuppressWarnings( "unchecked" )
     public <T extends IdentifiableObject> T loadByCode( Class<T> type, String code )
         throws IllegalQueryException
     {
-        IdentifiableObjectStore<IdentifiableObject> store = getIdentifiableObjectStore( type );
+        IdentifiableObjectStore<T> store = getIdentifiableObjectStore( type );
 
         if ( store == null )
         {
-            return null;
```
jbee, line 405: Definitely a bug

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -675,32 +635,30 @@ public <T extends IdentifiableObject> List<T> loadByUid( Class<T> type, Collecti
...
     @Transactional( readOnly = true )
-    @SuppressWarnings( "unchecked" )
     public <T extends IdentifiableObject> List<T> getById( Class<T> type, Collection<Long> ids )
     {
-        IdentifiableObjectStore<IdentifiableObject> store = getIdentifiableObjectStore( type );
+        IdentifiableObjectStore<T> store = getIdentifiableObjectStore( type );
 
         if ( store == null )
         {
-            return null;
```
jbee, line 685: OBS! could be considered a bug

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1152,97 +1073,93 @@ public <T extends IdentifiableObject> void updateNoAcl( T object )
...
+            return List.of();
         }
 
-        IdentifiableObjectStore<IdentifiableObject> store = getIdentifiableObjectStore( type );
+        IdentifiableObjectStore<T> store = getIdentifiableObjectStore( type );
 
         if ( store == null )
         {
-            return null;
```
jbee, line 1215: OBS! could be considered a bug

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -943,11 +871,11 @@ public <T extends IdentifiableObject> List<T> getObjects( Class<T> type, Identif
             {
                 return store.getByUid( identifiers );
             }
-            else if ( IdentifiableProperty.CODE.equals( property ) )
+            if ( IdentifiableProperty.CODE.equals( property ) )
```
teleivo, line 874: since this is an enum should we move to `==` instead of `equals`?
just because you touched this line, I know its not part of the scope of your PR. so you can also leave it 😋

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/organisationunit/OrganisationUnitController.java
```diff
@@ -278,9 +279,7 @@
 
     @GetMapping( "/{uid}/parents" )
     public @ResponseBody List<OrganisationUnit> getEntityList( @PathVariable( "uid" ) String uid,
-        @RequestParam Map<String, String> parameters, Model model, TranslateParams translateParams,
-        HttpServletRequest request, HttpServletResponse response )
-        throws Exception
+        @RequestParam Map<String, String> parameters, TranslateParams translateParams )
```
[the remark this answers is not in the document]
teleivo: seems like you can remove `parameters` 😄

teleivo (commented):
There does not seem to be support in Intellij for `@NotNull` highlighting when one passes null 😢 Its still useful for documentation purposes. Spring also supports it in its validation framework.

Regarding the places where you did highlight that this might be a bug: I think we should throw an exception instead of returning an empty collection as this could delay finding the issue. How would we discern the cases that these methods did not find anything vs there is no store for the type of object 😬

jbee:
> There does not seem to be support in Intellij for @NotNull

Also seems sonar does not use it. Maybe we should find an annotation that is picked up and use that.

teleivo:
> > There does not seem to be support in Intellij for @NotNull
> 
> Also seems sonar does not use it. Maybe we should find an annotation that is picked up and use that.

Its weird that there seems to be a rule that suggests using `@NotNull` https://github.com/SonarSource/sonar-java/blob/5f6684297582c1919aaf0fb52c610127aa24aaff/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1695.html

When searching for NotNull https://github.com/SonarSource/sonar-java/search?q=NotNull in sonar I find references to all the existing ones 😂

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/organisationunit/OrganisationUnitController.java
```diff
@@ -212,16 +211,18 @@ else if ( anyQueryPropertySet )
...
         {
-            for ( OrganisationUnit unit : list )
+            Optional<? extends IdentifiableObject> member = manager.find( memberObject );
+            if ( member.isPresent() )
             {
-                Long count = organisationUnitService.getOrganisationUnitHierarchyMemberCount( unit, member,
-                    memberCollection );
+                for ( OrganisationUnit unit : list )
+                {
+                    Long count = organisationUnitService.getOrganisationUnitHierarchyMemberCount( unit, member.get(),
```
enricocolasante, line 221: Not part of this PR but just to highlight this method that just add complexity.
Why return a Long if then the null value is evaluated 0, instead of just returning a long with default value 0?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
```diff
@@ -72,223 +75,246 @@
      * {@link org.hisp.dhis.common.IdentifiableObjectManager#get(Class, String)}.
      *
      * @param uid a UID of an object of unknown type
-     * @return The {@link IdentifiableObject} with the given UID or null if no
-     *         such object exists
+     * @return The {@link IdentifiableObject} with the given UID
      */
-    IdentifiableObject find( String uid );
+    @NotNull
```
enricocolasante, line 80: This is just Java being Java.
We are returning an Optional but technically we still need to be sure that it is not null :(

enricocolasante (commented):
When we find an annotation that will be caught in the static analysis in Sonar we can better evaluate the usefulness of this PR.

My concern is that we will need to annotate 99% of the method that are not void to compensate for a missing feature in the programming language.

teleivo (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
jbee: Yes, `Optional` is not privileged in any way and will require the annotation.

enricocolasante (approved):

PR #12039, opened 2022-10-07
fix: Use callback handler for query execution [DHIS2-13873]

maikelarabori:
This PR aims to fix the out-of-memory error that happens in some environments.

The original approach where we were using `queryForRowSet` was using a local cache behind the scenes (see `CachedRowSet`), to load all DB rows. This was probably causing out-of-memory when the volume of data was larger than the JVM would expect.

Locally, I was able to reproduce the out-of-memory problem at the same point, by tweaking the Tomcat settings to something like this:
```
export CATALINA_OPTS="$CATALINA_OPT -Xms320m -Xmx320m -Xss1m"
```

With these changes, even with the low memory settings above, we are able to complete the analytics export process which used to fail due to out-of-memory.

We also tested this fix on the performance environment (`https://test.performance.dhis2.org/2.39.0-rc-2`), where we deployed a "patched" WAR to test the changes and confirm that the problem is gone.

More details about it can be found on the JIRA `https://dhis2.atlassian.net/browse/DHIS2-13873`

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ResourceTableController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ResourceTableController.java
@@ -102,11 +102,6 @@ public class ResourceTableController
             skipTableTypes.add( AnalyticsTableType.ENROLLMENT );
         }
 
-        if ( skipEvents && skipEnrollment )
-        {
-            skipTableTypes.add( AnalyticsTableType.OWNERSHIP );
-        }
-
         if ( skipOrgUnitOwnership )
         {
             skipTableTypes.add( AnalyticsTableType.OWNERSHIP );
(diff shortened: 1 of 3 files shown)


jimgrace (approved):
Nice work, @maikelarabori!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ResourceTableController.java
```diff
@@ -106,9 +107,13 @@ public WebMessage analytics(
             skipTableTypes.add( AnalyticsTableType.OWNERSHIP );
         }
 
+        if ( skipOrgUnitOwnership )
```
jimgrace, line 110: Now that we have a proper way of skipping the ownership tables, please remove these lines just above:

        if ( skipEvents && skipEnrollment )
        {
            skipTableTypes.add( AnalyticsTableType.OWNERSHIP );
        }

This will allow users to build the ownership tables only at a different time if they want to do that, and then generate the events and enrollments tables at a different time.

Otherwise, looks great, thanks!!!!!

jimgrace (changes requested):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ResourceTableController.java
maikelarabori: Done

jimgrace (approved):

gnespolino (approved):
Look great!

PR #11981, opened 2022-10-03
ci: use jib to build DHIS2 Docker image TECH-1433

teleivo:
## Test

You can build and run DHIS2 and the Docker image using Jib with

```sh
./dhis-2/build-dev.sh
docker compose up
```

built as `dhis2/core-dev:local`

Get image from this PR as its labeled with `publish-docker-image`

```sh
docker pull dhis2/core-pr:11981
docker inspect dhis2/core-pr:11981 -f "{{ .Config.Labels }}"
map[DHIS2_VERSION:2.40-SNAPSHOT]
```

## Details

* PRs are e2e tested against a Docker image built using Jib
* PRs labeled with `publish-docker-image` will get their image which is built in the e2e test run published to https://hub.docker.com/r/dhis2/core-pr/tags with tag `dhis2/core-pr:$PR_NUMBER` i.e 11981 for this PR `docker pull dhis2/core-pr:11981`
* `dhis2/build-dev.sh` builds core and then Docker image using Jib
* README.md contains instructions on building/running the Docker image
* Jenkins dev build is now building the Docker image and testing different Tomcat variants. Its structured as

image

Previously, we were publishing dhis.war and updating `play.dhis2.org` even if the e2e tests were failing. Not anymore 😋 

* label Docker image with DHIS2_VERSION, DHIS2_BUILD_REVISION so we can debug issues more easily

```sh
docker pull dhis2/core-dev:master
docker inspect dhis2/core-dev:master -f "{{ .Config.Labels }}"
map[DHIS2_BUILD_REVISION:031cd0a47d9e436f243d6eb3264a1894c414afe9 DHIS2_VERSION:2.40-SNAPSHOT]
```

* updated naming scheme for Docker images to reflect what we already did before this PR in addition to changes we already agreed on https://github.com/dhis2/operations-handbook/pull/6

Next PRs:
- build & push images for arm https://dhis2.atlassian.net/browse/TECH-1454
- can we adapt the docker-compose.yml in a way to allow users to start DHIS2 with the Sierra Leone demo DB without checking it into the repository

To discuss:
- re-evaluate whether we can simplify the Docker images we publish. We have a lot of repos/tags like dhis2/core, dhis2/core-dev, dhis2/core-pr, canary?

PR #12046, opened 2022-10-08
ci: support running Docker image on arm64 TECH-1454

teleivo:
## To Test

* Run `DHIS2_IMAGE=dhis2/core-dev:2.40-rc-8.5-jre11 docker compose up`. This should pull the necessary images, demo DB and start DHIS2
* `docker inspect dhis2/core-dev:2.40-rc-8.5-jre11 -f "{{ .Architecture }}"` should show arm64 if you are on this architecture

## Details

Building on https://github.com/dhis2/dhis2-core/pull/11981

This PR adds support for building multi-arch Docker images (`linux/amd64` and `linux/arm64`) via Jib and [this profile](https://github.com/dhis2/dhis2-core/pull/12046/files#diff-8a1e9dcacfbca843e23f920dcd5957fbd340b220b140fc56dea6e4d4020ed39bR18-R43), as briefly explained [here](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image).

Both the `dev` and the `stable` pipelines will build the Docker images twice:

* The first time the images are tagged with the base tomcat image variant we're running the e2e API tests against (currently `8.5-jre11` and `9.0-jre11`). 
* The second time (only if the tests have passed) the image based on the `9.0-jre11` tomcat variant is built again and tagged with the appropriate branch/tag version (like `latest`, `2.39`, `2.39.0`, etc)

We do this in order to use the convenience of `jib:build` for building multi-arch images, instead of having to tag images separately with their architecture and then combining them in one manifest list.

The `dev` pipeline will use the `core-dev` repo for both of the builds.
The `stable` one will use the `core-dev` repo the first time we build (before the tests) and then once the images have passed the tests - it will build and push to `core`. The reasoning for this is that we shouldn't push untested images or images failing the tests to the main `core` repo.


amcgee:
> Note that the other tags are like teleivo/dhis2:master-linux-arm64 will also be visible.

One option we could consider, if we don't want to pollute the main image with a bunch of tags - we could create a dedicated image just for the arch-specific builds, something like `dhis2/core-singlearch` (terrible name, and maybe not worth the effort, but I think this could still be referenced from a manifest pushed to `dhis2/core`)

teleivo:
> > Note that the other tags are like teleivo/dhis2:master-linux-arm64 will also be visible.
> 
> One option we could consider, if we don't want to pollute the main image with a bunch of tags - we could create a dedicated image just for the arch-specific builds, something like `dhis2/core-singlearch` (terrible name, and maybe not worth the effort, but I think this could still be referenced from a manifest pushed to `dhis2/core`)

I also thought of it 👌🏻 We can try and compare the pros/cons of both approaches. Is the image working for you? I am curious if it does work 😄

radnov (commented):
Left a few comments, in case we keep PR as is.

radnov (approved):

tonsV2 (approved):

amcgee (commented):
This won't support building on a local machine right?

tonsV2:
If you run build-dev.sh, and by such invoke jib:dockerBuild, the artifact will be pushed to your locally running docker daemon (unless you configured a remote one of course). This won't be a multi arch manifest.
If you run jib:build the artifact will be pushed to a (very likely) remote registry. If you run jib:build -PjibBuild, the same will happen but with a multi arch artifact.
So yes, it works locally but depending on what you do, you might have to pull the artifact afterwards

teleivo:
> If you run build-dev.sh, and by such invoke jib:dockerBuild, the artifact will be pushed to your locally running docker daemon (unless you configured a remote one of course). This won't be a multi arch manifest. If you run jib:build the artifact will be pushed to a (very likely) remote registry. If you run jib:build -PjibBuild, the same will happen but with a multi arch artifact. So yes, it works locally but depending on what you do, you might have to pull the artifact afterwards

We should be able to improve the ergonomics of this 😅 using

https://maven.apache.org/guides/introduction/introduction-to-profiles.html

which allows us to activate profiles by default for example based on architecture. build-dev.sh would then build the Docker image for the appropriate architecture.

I think we can do this in another PR.

jenkinsfiles/dev
```diff
@@ -142,17 +146,16 @@ pipeline {
         stage('Publish images') {
             steps {
                 withDockerRegistry([credentialsId: "docker-hub-credentials", url: ""]) {
-                    script {
-                        // Docker images are built in a matrix of different base images
-                        // all these images are used for testing but only one is published and endorsed for use in production
-                        // which is the DOCKER_IMAGE_NAME_PUBLISH_SOURCE
-                        sh "docker tag ${DOCKER_IMAGE_NAME_PUBLISH_SOURCE} ${DOCKER_IMAGE_NAME_PUBLISH_TARGET}"
-                        sh "docker push ${DOCKER_IMAGE_NAME_PUBLISH_TARGET}"
```
teleivo, line 150: We previously pushed images tagged as master and latest which we do not anymore if I understand the changes correctly. If this is ok you will need to adapt https://github.com/dhis2/operations-handbook/pull/6/files

jenkinsfiles/dev
```diff
@@ -142,17 +146,16 @@ pipeline {
...
+                    if (env.GIT_BRANCH == 'master') {
+                        tag = 'latest'
+                    } else {
+                        tag = env.GIT_BRANCH
+                    }
+
+                    withMaven(options: [artifactsPublisher(disabled: true)]) {
+                        sh "mvn --batch-mode --no-transfer-progress -DskipTests -Dmaven.test.skip=true \
+                                        -f dhis-2/dhis-web/dhis-web-portal/pom.xml jib:build -PjibBuild \
+                                        -Djib.from.image=${DOCKER_IMAGE_NAME_PUBLISH_SOURCE} -Djib.to.image=${DOCKER_IMAGE_NAME}:${tag}"
```
teleivo, line 158: I would also add the labels here

jenkinsfiles/stable
```diff
@@ -56,31 +57,36 @@ pipeline {
                         values '9.0-jre11', '8.5-jre11'
                     }
                 }
+
                 environment {
                     DOCKER_IMAGE_NAME_BASE_FULL = "${DOCKER_IMAGE_NAME_BASE}:${DOCKER_IMAGE_TAG_BASE}"
-                    DOCKER_IMAGE_NAME_FULL = "${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}-${DOCKER_IMAGE_TAG_BASE}" // used to test against different Tomcat variants
+                    DOCKER_IMAGE_NAME_FULL = "${DOCKER_IMAGE_NAME_DEV}:${DOCKER_IMAGE_TAG}-${DOCKER_IMAGE_TAG_BASE}" // used to test against different Tomcat variants
```
teleivo, line 63: I think we need some kind of slug (maybe the build number) in the image name to guard against concurrent builds overwriting each others images 🤔 Before this was not the case as we did not push/pull intermediary images.

teleivo (commented):
I cannot select "request changes" as its "my" PR 😂
I think there are a few things we might need to change unless they are not a concern of course 😋

.gitignore
```diff
@@ -31,8 +31,9 @@ overlays/
...
-node_modules
+node_modules
 **/rebel.xml
-
-# ignore generated artifact directory
-docker/artifacts
+
+# ignore generated artifact directory
+docker/artifacts
+postgresql-persistence
```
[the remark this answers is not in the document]
tonsV2: I'm honestly not sure. I thought it was a side effect of the new docker compose file you introduced.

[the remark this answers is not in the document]
teleivo: I am using the docker compose file every day for some time now and I have never seen this file being created. Maybe its the baosystems image that creates it 🤷🏻

jenkinsfiles/dev
teleivo: Maybe I should add that I do agree with this change 😋 @radnov and I just opted for keeping the repo/tag structure as it is when working on jib. We thought this might need its own discussion/PR which is not related to this work.

jenkinsfiles/stable
tonsV2: Do we really need this? The is the stable build which is invoked manually and not very often

teleivo: @radnov do you see any chance of what I described happening? If not you can mark this as resolved 😋

radnov: Concurrent builds shouldn't really happen, as the stable pipeline is triggered manually by Phil for both branches (release candidates) and tags (releases). 👀

teleivo:
@Philip-Larsen-Donnelly @amcgee I am also accepting this PR from @tonsV2 and @radnov 😄 Are we good to merge it?

.gitignore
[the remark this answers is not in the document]
amcgee: A docker image shouldn't be able to interact with the host filesystem though...?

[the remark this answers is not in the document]
teleivo: It could AFAIK if we mount a volume. This change was removed. I never see this file appear on my machine 👌🏻 @amcgee

Philip-Larsen-Donnelly (approved):

Philip-Larsen-Donnelly:
Let's merge this now. If there are remaining issues, we can deal with them down the line.

PR #12047, opened 2022-10-08
ci: use db healthcheck to wait for e2e DB

teleivo:
to become ready
https://github.com/compose-spec/compose-spec/blob/master/spec.md\#long-syntax-1

which is more accurate in understanding whether the DB initialization of the DHIS2 DB is done. Rather than just relying on the tcp port to be ready. Note that the pg_isready was not chosen as it does not take the DB actually existing into account. https://dba.stackexchange.com/questions/298643/monitoring-a-specific-database-with-pg-isready


radnov (approved):
Nice addition 👍

enricocolasante (approved):

PR #12065, opened 2022-10-10
ci: use JRE base image for DHIS2 Docker image TECH-1457

teleivo:
as we do not need to build DHIS2 within the image but only run it

```sh
docker pull dhis2/core-pr:12065
docker image ls dhis2/core-pr:12065
REPOSITORY      TAG       IMAGE ID       CREATED        SIZE
dhis2/core-pr   12065     c71ae7349c0f   52 years ago   865MB
```

compared to image from https://github.com/dhis2/dhis2-core/pull/11981

```sh
docker pull dhis2/core-pr:11981
docker image ls dhis2/core-pr:11981
REPOSITORY      TAG       IMAGE ID       CREATED        SIZE
dhis2/core-pr   11981     2eb020f18ceb   52 years ago   1.04GB
```

**175MB reduction**

### DIFF
--- a/dhis-2/dhis-web/dhis-web-portal/pom.xml
+++ b/dhis-2/dhis-web/dhis-web-portal/pom.xml
@@ -97,7 +97,7 @@
     <properties>
         <rootDir>../../</rootDir>
         <jib.version>3.3.0</jib.version>
-        <jib.from.image>tomcat:9.0-jdk11-openjdk-slim</jib.from.image>
+        <jib.from.image>tomcat:9.0-jre11</jib.from.image>
         <jib.to.image>dhis2/core-dev:local</jib.to.image>
         <!-- uid=65534(nobody) present in Tomcat image -->
         <jib.container.user>65534</jib.container.user>
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -25,7 +25,7 @@ pipeline {
         DHIS2_VERSION = readMavenPom(file: 'dhis-2/pom.xml').getVersion()
         DOCKER_IMAGE_NAME = "${DOCKER_HUB_OWNER}/core-dev"
         DOCKER_IMAGE_TAG = "${env.GIT_BRANCH}"
-        DOCKER_IMAGE_NAME_PUBLISH_SOURCE = "${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}-9.0-jdk11-openjdk-slim" // source of image to publish to Dockerhub (one of the matrix axes)
+        DOCKER_IMAGE_NAME_PUBLISH_SOURCE = "${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}-9.0-jre11" // source of image to publish to Dockerhub (one of the matrix axes)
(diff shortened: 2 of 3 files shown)


radnov (commented):
The `*-jre11` tag seems to point to a Temurin & Jammy variant
image

However, the Debian tag we are using now doesn't seem to be supported anymore, either:
https://hub.docker.com/_/tomcat

🤔

teleivo:
> The `*-jre11` tag seems to point to a Temurin & Jammy variant image
> 
> However, the Debian tag we are using now doesn't seem to be supported anymore, either: https://hub.docker.com/_/tomcat
> 
> thinking

Correct. Does it matter to us whether its debian/ubuntu? @tonsV2 @amcgee

tonsV2 (approved):

teleivo:
> > The `*-jre11` tag seems to point to a Temurin & Jammy variant image
> > However, the Debian tag we are using now doesn't seem to be supported anymore, either: https://hub.docker.com/_/tomcat
> > thinking
> 
> Correct. Does it matter to us whether its debian/ubuntu? @tonsV2 @amcgee

Actually check https://github.com/docker-library/tomcat/tree/master/9.0/jdk11
Isn't this already the case for the image we are using https://github.com/docker-library/tomcat/tree/master/9.0/jdk11 ?

radnov:
> > > The `*-jre11` tag seems to point to a Temurin & Jammy variant image
> > > However, the Debian tag we are using now doesn't seem to be supported anymore, either: https://hub.docker.com/_/tomcat
> > > thinking
> > 
> > 
> > Correct. Does it matter to us whether its debian/ubuntu? @tonsV2 @amcgee
> 
> Actually check https://github.com/docker-library/tomcat/tree/master/9.0/jdk11 Isn't this already the case for the image we are using https://github.com/docker-library/tomcat/tree/master/9.0/jdk11 ?

Nope, we are currently using a non-supported Debian based tag - https://hub.docker.com/_/tomcat/tags?page=1&name=9.0-jdk11-openjdk-slim

And given that there's no more Debian based image, I suppose there's no alternative really. But at least we should probably stick to Ubuntu Focal, rather than Jammy. 🤔

teleivo:
> And given that there's no more Debian based image, I suppose there's no alternative really. But at least we should probably stick to Ubuntu Focal, rather than Jammy. thinking

Why? Isn't Jammy the newer LTS release, so shouldn't we go with that? And by using `9.0-jre11` I assume we are on Jammy.

radnov:
> > And given that there's no more Debian based image, I suppose there's no alternative really. But at least we should probably stick to Ubuntu Focal, rather than Jammy. thinking
> 
> Why? Isn't Jammy the newer LTS release, so shouldn't we go with that? And by using `9.0-jre11` I assume we are on Jammy.

Yeah, I'm just thinking that since Focal is still much more widespread - it will match more environments that might exist now. 

But you are right, since this shouldn't really matter for the end user that just wants "a docker image".

tonsV2:
Regarding Debian vs. Ubuntu. Commercial backed distros tend to update faster in regards to security fixes

Philip-Larsen-Donnelly (approved):

PR #12071, opened 2022-10-10
chore: only use @Nonnull and ban similar annotations

jbee:
Uses `@Nonnull` (javax annotation) instead of `@NotNull` (javax validation) or `@NonNull` (lombok).

Switching to `@Nonnull` also caused hibernate to run nullness analysis in tests so that all non-null database fields need to have a value when stored. This required some more changes to tests to provide a value.

In case of `IndicatorGroup.groupSet` this mapping was not updated similar to what had been done to `DateElementGroup`.
A `IndicatorGroup` can be member of multiple `IndicatorGroupSet`s. Therefore the `IndicatorGroup.groupSet` was changed to a `Set<IndicatorGroupSet>` called `IndicatorGroup.groupSets`. This change was mande now as hibernate otherwise considered the relation as non-null and complained about missing value in tests.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventvisualization/SimpleDimensionHandler.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventvisualization/SimpleDimensionHandler.java
@@ -139,9 +139,7 @@ public class SimpleDimensionHandler
     private SimpleDimension createSimpleEventDimensionFor( final DimensionalObject dimensionalObject,
         final Attribute attribute )
     {
-        final SimpleDimension simpleDimension = new SimpleDimension();
-        simpleDimension.setParent( attribute );
-        simpleDimension.setDimension( dimensionalObject.getUid() );
+        final SimpleDimension simpleDimension = new SimpleDimension( attribute, dimensionalObject.getUid() );
 
         if ( isNotEmpty( dimensionalObject.getItems() ) )
         {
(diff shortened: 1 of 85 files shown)


teleivo (approved):

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventvisualization/SimpleDimensionHandler.java
```diff
@@ -139,9 +139,7 @@ private List<BaseDimensionalItemObject> loadDimensionalItems( final String dimen
     private SimpleDimension createSimpleEventDimensionFor( final DimensionalObject dimensionalObject,
         final Attribute attribute )
     {
-        final SimpleDimension simpleDimension = new SimpleDimension();
```
teleivo, line 142: thank you very much 🎉

teleivo (approved):

teleivo (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventvisualization/SimpleDimensionHandler.java
teleivo: annoying that simple things often lead to more problems 😬 constructors seem to have gone out of fashion at some point 😂

gnespolino (approved):

PR #12074, opened 2022-10-10
ci: download demo DB for local dev setup

teleivo:
Use the init container style pattern with the `db-dump` container running first to ensure a demo DB exists in the db-dump volume. `db-dump` volume is shared with the db so it will be picked up during postgres initialization.

* Allow us to start an arbitrary DHIS2 version via `DHIS2_IMAGE=dhis2/core-dev:2.39 docker compose up`

```sh
docker compose images
Container           Repository          Tag                 Image Id            Size
core-db-1           postgis/postgis     10-2.5-alpine       82cadaebf2ad        215MB
core-web-1          dhis2/core-dev      2.39                43115e22363c        1.03GB
```

also useful for the PR images we push to https://hub.docker.com/r/dhis2/core-pr/tags

* Use `DHIS2_IMAGE` instead of `IMAGE_NAME` to be more specific which service this affects. I took `IMAGE_NAME` from our e2e docker compose file.
* Allow us to init DB with an arbitrary DB dump via `DHIS2_DB_DUMP_URL=https://databases.dhis2.org/sierra-leone/2.39/dhis2-db-sierra-leone.sql.gz docker compose up`
* DB dump is is only downloaded the first time

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,43 @@
-version: "3"
+version: "3.8"
 
 services:
   web:
-    image: dhis2/core-dev:local
+    image: "${DHIS2_IMAGE:-dhis2/core-dev:local}"
     ports:
       - "8080:8080"
     volumes:
       - ./docker/dhis.conf:/opt/dhis2/dhis.conf:ro
     depends_on:
-      - db
+      db:
+        condition: service_healthy
 
   db:
     image: postgis/postgis:10-2.5-alpine
+    volumes:
+      - db-dump:/docker-entrypoint-initdb.d/
     environment:
       POSTGRES_USER: dhis
       POSTGRES_DB: dhis2
-      POSTGRES_PASSWORD: dhis
+      POSTGRES_PASSWORD: &postgres_password dhis
+      PGPASSWORD: *postgres_password # needed by psql in healthcheck
+    healthcheck:
+      test: ["CMD-SHELL", "psql --no-password --quiet --username $$POSTGRES_USER postgres://127.0.0.1/$$POSTGRES_DB -p 5432 --command \"SELECT 'ok'\" > /dev/null"]
+      start_period: 120s
+      interval: 1s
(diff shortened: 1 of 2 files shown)


radnov (commented):
Looks good, but I think parametrizing the demo DB url is essential for this change, as well as having a sane default for the SL database version. The lowest supported DHIS2 version maybe? 🤔 

Also - the healthcheck overall period (about ~40s if I read [the options](https://docs.docker.com/engine/reference/builder/#healthcheck) right) wasn't enough on my machine. Note that I've allocated very little resources to the Docker Engine, but I think having a longer start period might be a good idea, just because there shouldn't be any downside to it.

enricocolasante (approved):

docker-compose.yml
```diff
@@ -1,18 +1,43 @@
...
+      retries: 5
+    depends_on:
+      db-dump:
+        condition: service_completed_successfully # make sure the DB dump has been downloaded
+
+  db-dump:
+    image: busybox # busybox wget version does not have --no-clobber, so we need to do the [ -f ] test
+    command: sh -c '[ -f dump.sql.gz ] && echo "dump.sql.gz exists" || wget --output-document dump.sql.gz $$DHIS2_DB_DUMP_URL' # only download file if it does not exist
+    environment:
+      DHIS2_DB_DUMP_URL: "${DHIS2_DB_DUMP_URL:-https://databases.dhis2.org/sierra-leone/2.38.1/dhis2-db-sierra-leone.sql.gz}"
```
radnov, line 37: Should the default version here be `2.38.1`?

radnov (commented):
Looks good :+1:, just one question below.

docker-compose.yml
teleivo: ah, I wanted to use `dev` will update ASAP. Thank you 😄

radnov (approved):

PR #12080, opened 2022-10-11
fix: support application/zip for /api/events

mortenoh:


vietnguyen (approved):

muilpp (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/EventController.java
```diff
@@ -652,6 +654,14 @@ public void getCsvEvents(
             outputStream = new GZIPOutputStream( outputStream );
             response.setContentType( "application/csv+gzip" );
         }
+        else if ( ContextUtils.isAcceptCsvZip( request ) )
+        {
```
larshelge, line 658: I get a bit of a "centralize this" feeling but it might not be worth it.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/ContextUtils.java
```diff
@@ -344,6 +344,21 @@ public static boolean isAcceptCsvGzip( HttpServletRequest request )
...
+     * compressed response.
+     *
+     * @param request the HttpServletRequest.
+     * @return whether the given requests indicates that it accepts a compressed
+     *         response.
+     */
+    public static boolean isAcceptCsvZip( HttpServletRequest request )
+    {
+        return request != null && ((request.getPathInfo() != null && request.getPathInfo().endsWith( ".zip" ))
+            || (request.getHeader( "Accept" ) != null
```
larshelge, line 358: Use Spring HttpHeaders constants for header names instead of strings ("Accept")?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/EventController.java
mortenoh: No, I don't in this case it's not so important. This code is on the way out, and as you see the tracker team has rewritten it to not use the message converters anymore.

PR #12083, opened 2022-10-11
ci: improve e2e test Docker workflow

teleivo:
* Allow running e2e tests using docker compose within one command.

`docker compose -f docker-compose.yml -f docker-compose.e2e.yml up --remove-orphans --exit-code-from test`

* Build e2e tests automatically

* Allow passing/pulling of an arbitrary image `DHIS2_IMAGE` like an image pushed from a PR like this one `dhis2/core-pr:12083`. Such images will only be pulled if missing. This is to avoid pulling an image in CI where we only want to test the images built on CI.

`DHIS2_IMAGE=dhis2/core-pr:12083 docker compose -f docker-compose.yml -f docker-compose.e2e.yml up --remove-orphans --exit-code-from test`

* Renamed `IMAGE_NAME` to `DHIS2_IMAGE` to be more explicit as to which image we are targeting.

* Improve the README by explaining all the ways in which the e2e tests can be run. I never remember it when I have to run them after a while. Document how to potentially debug any failing e2e tests.

* Only compile e2e once instead of twice. They were compiled once in the Docker build of the test image and again when running the tests as part of the `test` container command. Since compilation is fast lets just compile during `mvn test` when running the e2e tests. If we need the speed we can take on the added complexity of the Dockerfile compilation/preventing re-compilation during `mvn test`

* Do not copy the `dhis.conf` in the `test` container. Not sure why that was done. Our e2e tests are testing DHIS2 from the outside via HTTP only.

* Consistently use DB name `dhis` for our local dev setup, integration and e2e tests. So devs can connect to any of these DBs using the same connection settings.


dhis-2/dhis-e2e-test/docker-compose.e2e.yml
```diff
@@ -1,9 +1,12 @@
 version: "3"
```
enricocolasante, line 1: Version here is different from the one in the other docker compose file. Is it a problem?

dhis-2/dhis-e2e-test/README.md
```diff
@@ -1,54 +1,149 @@
...
+```
+
+### Inside Docker
+
+If you do not have DHIS2 running, you can either run
+
+1. only DHIS2 in Docker and the tests outside of Docker
+
+```sh
+docker-compose up --detach
```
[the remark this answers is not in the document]
teleivo: yes good point 😄 both will work, but without the dash is the new way 👍🏻

dhis-2/dhis-e2e-test/docker-compose.e2e.yml
teleivo: Will adjust it so it is the same as in the other file 👍🏻

enricocolasante (approved):

radnov (changes requested):

radnov (approved):

PR #12102, opened 2022-10-12
chore: @Nonnull and @CheckForNull in DefaultIdentifiableObjectManager  [TECH-1460]

jbee:
### Summary
Adds nullness annotations to `DefaultIdentifiableObjectManager` and some that got overlooked in the `IdentifiableObjectManager`.

Adding the annotations already helped the IDE to highlight some potential NPEs and inconsistencies that I addressed in this PR.

### Automatic Testing
Added two tests that check the changed behaviour of `store.get` to either return a non-null value or throw an exception.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
@@ -72,9 +72,11 @@ public interface GenericStore<T>
      *
      * @param id the object identifier.
      * @return the object identified by the given identifier.
+     * @throws IllegalQueryException when no such object exists in the database
      */
     @Nonnull
-    T get( long id );
+    T get( long id )
+        throws IllegalQueryException;
 
     long countAllValuesByAttributes( @Nonnull List<Attribute> attributes );
 
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
@@ -29,6 +29,7 @@ package org.hisp.dhis.common;
 
 import java.util.List;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 import org.hisp.dhis.attribute.Attribute;
@@ -71,12 +72,10 @@ public interface GenericStore<T>
      * look in the current Session, then hit the database if not existing.
      *
      * @param id the object identifier.
-     * @return the object identified by the given identifier.
-     * @throws IllegalQueryException when no such object exists in the database
+     * @return the object identified by the given identifier or null
      */
-    @Nonnull
-    T get( long id )
-        throws IllegalQueryException;
+    @CheckForNull
+    T get( long id );
 
     long countAllValuesByAttributes( @Nonnull List<Attribute> attributes );
 
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -37,7 +37,6 @@ import java.util.Set;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
-import org.hibernate.ObjectNotFoundException;
 import org.hisp.dhis.attribute.Attribute;
 import org.hisp.dhis.attribute.AttributeValue;
 import org.hisp.dhis.feedback.ErrorCode;
@@ -83,16 +82,16 @@ public interface IdentifiableObjectManager
     Optional<? extends IdentifiableObject> find( @Nonnull String uid );
 
     /**
+     * Lookup objects of a specific type by database ID.
      *
      * @param type the object class type.
      * @param id object's database ID
      * @return the found object
-     * @throws ObjectNotFoundException when no such object exists in the
-     *         database
+     * @throws IllegalQueryException when no such object exists in the database
      */
     @Nonnull
     <T extends IdentifiableObject> T get( @Nonnull Class<T> type, long id )
-        throws ObjectNotFoundException;
+        throws IllegalQueryException;
 
     /**
      * Retrieves the object of the given type and UID, or null if no object
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -87,11 +87,9 @@ public interface IdentifiableObjectManager
      * @param type the object class type.
      * @param id object's database ID
      * @return the found object
-     * @throws IllegalQueryException when no such object exists in the database
      */
-    @Nonnull
-    <T extends IdentifiableObject> T get( @Nonnull Class<T> type, long id )
-        throws IllegalQueryException;
+    @CheckForNull
+    <T extends IdentifiableObject> T get( @Nonnull Class<T> type, long id );
 
     /**
      * Retrieves the object of the given type and UID, or null if no object
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
@@ -37,6 +37,7 @@ import java.util.Set;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
+import org.hibernate.ObjectNotFoundException;
 import org.hisp.dhis.attribute.Attribute;
 import org.hisp.dhis.attribute.AttributeValue;
 import org.hisp.dhis.feedback.ErrorCode;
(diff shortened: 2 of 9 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectManager.java
```diff
@@ -81,8 +82,17 @@
     @Nonnull
     Optional<? extends IdentifiableObject> find( @Nonnull String uid );
 
+    /**
+     *
+     * @param type the object class type.
+     * @param id object's database ID
+     * @return the found object
+     * @throws ObjectNotFoundException when no such object exists in the
```
jbee, line 90: I think this was the behaviour all along as this should be how hibernate responds to lookup by ID when the object does not exist. This should just make this clear.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1138,57 +1165,51 @@ public <T extends IdentifiableObject> List<T> getByAttributeAndValue( Class<T> t
...
     @Transactional( readOnly = true )
     public Map<Class<? extends IdentifiableObject>, IdentifiableObject> getDefaults()
     {
-        Long catId = defaultObjectCache.get( Category.class.getName(),
-            key -> getByName( Category.class, DEFAULT ).getId() );
-        Long cateComboId = defaultObjectCache.get( CategoryCombo.class.getName(),
-            key -> getByName( CategoryCombo.class, DEFAULT ).getId() );
-        Long catOptionId = defaultObjectCache.get( CategoryOption.class.getName(),
-            key -> getByName( CategoryOption.class, DEFAULT ).getId() );
-        Long catOptionComboId = defaultObjectCache.get(
```
jbee, line 1176: FYI: In theory this was a potential NPE as `getId()` could be invoked on `null` returned by `getByName`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1138,57 +1165,51 @@ public <T extends IdentifiableObject> List<T> getByAttributeAndValue( Class<T> t
...
-            .put( Category.class, Objects.requireNonNull( get( Category.class, catId ) ) )
-            .put( CategoryCombo.class, Objects.requireNonNull( get( CategoryCombo.class, cateComboId ) ) )
-            .put( CategoryOption.class, Objects.requireNonNull( get( CategoryOption.class, catOptionId ) ) )
-            .put( CategoryOptionCombo.class,
-                Objects.requireNonNull( get( CategoryOptionCombo.class, catOptionComboId ) ) )
-            .build();
+        BiFunction<Map<Class<?>, Long>, Class<? extends IdentifiableObject>, Long> getIdByName = ( cache,
+            type ) -> defaultObjectCache.computeIfAbsent( type, key -> {
+                IdentifiableObject obj = getByName( type, DEFAULT );
+                return obj == null ? -1 : obj.getId();
```
jbee, line 1202: FYI: For the case where the default object actually does not exist `-1` is returned as ID which later will cause an exception when the ID is used to lookup the object. Effectively a NPE changed to a object not found exception.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1239,66 +1255,40 @@ public void removeUserGroupFromSharing( String userGroupUid )
...
+                log.warn( "No IdentifiableObjectStore found for class: '{}'", type );
             }
-        }
-
-        return (GenericDimensionalObjectStore<T>) store;
+            return store;
+        } );
     }
 
-    private void initMaps()
```
jbee, line 1283: This was not thread-safe and also unnecessarily complicated. The new "cache" is a concurrent map which is initializing on demand. In a real start of the server this map should quickly fill with all stores and behave the same as a map that is pre-initialized.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -84,69 +89,54 @@
...
     public static final String DEFAULT = "default";
 
     /**
-     * Cache for default category objects. Disabled during test phase.
+     * "Cache" for default category objects IDs. The IDs never change but the
+     * objects might. The cache exists to avoid doing a lookup by name by
+     * remembering the database ID, so it can be used in subsequent lookups.
      */
-    private final Cache<Long> defaultObjectCache;
+    private final Map<Class<?>, Long> defaultObjectCache = new ConcurrentHashMap<>( 4 );
```
jbee, line 103: FYI: Using an actual `Cache` to cache 4 longs by class name which also would never expire (become invalid or change) was unnecessarily complicated so I used a simple map instead.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -224,51 +214,53 @@ public void updateTranslations( IdentifiableObject persistedObject, Set<Translat
...
     @Transactional( readOnly = true )
-    public <T extends IdentifiableObject> T get( Class<T> type, long id )
+    public <T extends IdentifiableObject> T get( @Nonnull Class<T> type, long id )
     {
         IdentifiableObjectStore<T> store = getIdentifiableObjectStore( type );
 
         if ( store == null )
         {
-            return null;
+            throw new ObjectNotFoundException( id, "No store registered for objects of type: " + type );
```
jbee, line 254: OBS! Either this method should return an object or throw an exception. This makes it consistent.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -280,22 +272,25 @@ public <T extends IdentifiableObject> T get( Class<T> type, String uid )
...
-    public <T extends IdentifiableObject> T load( Class<T> type, String uid )
+    public <T extends IdentifiableObject> T load( @Nonnull Class<T> type, @Nonnull String uid )
         throws IllegalQueryException
     {
         IdentifiableObjectStore<T> store = getIdentifiableObjectStore( type );
 
         if ( store == null )
         {
-            return null;
+            throw new IllegalQueryException( "No store registered for objects of type: " + type );
```
jbee, line 284: OBS! Either this method should return an object or throw an exception. This makes it consistent.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -901,58 +924,62 @@ public <T extends IdentifiableObject> List<T> getObjects( Class<T> type, Collect
...
+            return store.getByCode( value );
+        }
+        if ( idScheme.is( IdentifiableProperty.NAME ) )
+        {
+            return store.getByName( value );
+        }
+        if ( idScheme.is( IdentifiableProperty.ATTRIBUTE ) )
+        {
+            Attribute attribute = get( Attribute.class, idScheme.getAttribute() );
+            if ( attribute == null )
```
jbee, line 964: OBS! In theory attribute can be null - handled by throwing an exception

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1239,66 +1254,40 @@ public void removeUserGroupFromSharing( String userGroupUid )
...
-        Class<T> type )
+    private <T extends IdentifiableObject> IdentifiableObjectStore<T> getIdentifiableObjectStore( @Nonnull T object )
     {
-        initMaps();
-
-        IdentifiableObjectStore<? extends IdentifiableObject> store = identifiableObjectStoreMap.get( type );
-
-        if ( store == null )
-        {
-            store = identifiableObjectStoreMap.get( type.getSuperclass() );
```
jbee, line 1251: Note the odd `type.getSuperclass()`. Presumably this was to "unproxy" the `Class` but this is already done by the callers of the `getIdentifiableObjectStore(Class)` method. There is a `getIdentifiableObjectStore(Object)` method which uses `getRealClass(obj)`. So the cleaned code uses `type` not  `type.getSuperclass()` as lookup to find the corresponding store.

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
```diff
@@ -72,9 +72,11 @@
      *
      * @param id the object identifier.
      * @return the object identified by the given identifier.
+     * @throws IllegalQueryException when no such object exists in the database
      */
     @Nonnull
-    T get( long id );
+    T get( long id )
+        throws IllegalQueryException;
```
jbee, line 79: I thought this always would have thrown some low level exception but it actually would not. I changed this now so it either returns a value or throws an `IllegalQueryException`.

vietnguyen: Should we add the `find` methods as a part of this refactoring ? 
I think in some places we might expect `null` result instead of the exception.

vietnguyen:
I think in many places we have been using the `get(id)` method in a wrong manner, which should have been `find(id)`. For example: 
```
 DataSet dataSet = manager.get( DataSet.class, uid );

        if ( dataSet == null )
        {
            throw new WebMessageException( conflict( "Data set does not exist: " + uid ) );
        }
```

Those places are currently expect a `null` result instead of an exception. How would this PR affects them ?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/GenericStore.java
jbee: reverted this.

PR #12103, opened 2022-10-12
[WIP] feat: validation service API

jbee:


teleivo (commented):
Added some thoughts because I am curious/working on tracker validation 😋 Knowing that this is WIP 😄 

Maybe the `ValidationScenario` could be part of the context to reduce the validate parameters. To me it also feels like a good place since the scenario is part of the context in which the validation occurs 🤔 

I found that in tracker we sometimes call `new Date()` in our validations. For tracker I am thinking of putting something like a `validationDate` into our validation context. So validations always use that to compare. This might prevent some weird effects if multiple validations call `new Date()` to compare dates. Error reports containing `now` date will then also contain the same one which is likely clearer.

larshelge:
Closing for now, please open a new PR when ready :+1:

PR #12105, opened 2022-10-12
feat: sort code refactored [DHIS2-13929]

gnespolino:


d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/ProgramAttributeCondition.java
```diff
@@ -40,6 +40,7 @@
 import org.hisp.dhis.analytics.shared.query.BinaryConditionRenderer;
 import org.hisp.dhis.analytics.shared.query.Field;
 import org.hisp.dhis.analytics.shared.query.OrCondition;
+import org.hisp.dhis.analytics.tei.query.context.QueryContext;
```
maikelarabori, line 43: I'm not sure if we need another package-level "context". We are going to seven levels...

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/context/SortingContextUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+import org.hisp.dhis.analytics.common.dimension.DimensionIdentifier;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.program.ProgramStage;
+import org.hisp.dhis.trackedentity.TrackedEntityType;
+
+@NoArgsConstructor( access = AccessLevel.PRIVATE )
+class SortingContextUtils
```
maikelarabori, line 42: We could add a TODO here, so we can revisit and use the query builder objects instead of the plain SQL.

maikelarabori (approved):
Looks good @gnespolino, I just left a few minor comments.
Thanks!

PR #12106, opened 2022-10-12
fix: Keep event datavalue dates if value is not changed [DHIS2-13632]

abyot:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/bundle/persister/EventPersister.java
```diff
@@ -217,10 +217,14 @@ private void handleDataValues( Session session, TrackerPreheat preheat, Set<Data
         } );
     }
 
-    private void handleDataValueCreatedUpdatedDates( DataValue dv, EventDataValue eventDataValue )
+    private void handleDataValueCreatedUpdatedDates( ValuesHolder valuesHolder, DataValue dv,
+        EventDataValue eventDataValue )
     {
-        eventDataValue.setCreated( getFromOrNewDate( dv, DataValue::getCreatedAt ) );
-        eventDataValue.setLastUpdated( getFromOrNewDate( dv, DataValue::getUpdatedAt ) );
+        if ( valuesHolder.getAuditType() != null )
```
larshelge, line 223: Just wondering, by looking at `ValuesHolder#getAuditParameters`, it seems like audit type will be not null even for updated values. I think this issue was to ensure that created date is _never changed for existing values_. At the same time, the last updated date _should_ change on updates. I am trying to see how the fix solves this.

stian-sandvold (approved):

PR #12113, opened 2022-10-13
fix: potential NPEs found by nullness analysis [TECH-1460]

jbee:
Addresses issues found https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=BUG&id=dhis2_dhis2-core

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/resolver/CategoryOptionGroupResolver.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/resolver/CategoryOptionGroupResolver.java
@@ -30,6 +30,7 @@ package org.hisp.dhis.analytics.resolver;
 import static org.hisp.dhis.commons.collection.CollectionUtils.isEmpty;
 import static org.hisp.dhis.expression.ParseType.INDICATOR_EXPRESSION;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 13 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/resolver/CategoryOptionGroupResolver.java
```diff
@@ -113,21 +114,23 @@ private String getExpression( String expression, DimensionalItemId id, List<Stri
...
 
         for ( String cogUid : cogUidList )
         {
-            CategoryOptionGroup cog = categoryOptionGroupStore
-                .getByUid( cogUid );
-
+            CategoryOptionGroup cog = categoryOptionGroupStore.getByUid( cogUid );
+            if ( cog == null )
+            {
+                continue;
```
larshelge, line 124: I think we should probably not silently ignore this and instead fail early, by using `loadByUid`, and throw an exception. I think we should encapsulate these null checks a bit more.

jbee: I was also thinking about adding a `loadById(long)` that would throw an exception so we can use lookup by long and have the exception behaviour. WDYT?

larshelge: Yes I think that sounds good.

jbee: Yes, meant in general - there were other cases I think.

netroms (approved):

jbee:
I used `loadByUid` now where single objects were fetched. In loops I used null check so objects that are not found are skipped/ignored.

vietnguyen (approved):

PR #12129, opened 2022-10-14
fix: Updating relationships through /trackedentityinstances endpoint shows error 500 [ DHIS2-13919 ] 

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-13919
In the relationship update, we were persisting an object with missing references to the DB object making it invalid.
In this PR we fix only using the DB object updating what is in the payload without creating any transient object.


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/relationship/AbstractRelationshipService.java
```diff
@@ -363,20 +363,77 @@ else if ( daoRelationship.isDeleted() )
...
+     * @param relationshipInput relationship in the payload
+     * @return relationshipDb with updated relationshipItems
+     */
+    private org.hisp.dhis.relationship.Relationship updatedDAORelationship(
+        org.hisp.dhis.relationship.Relationship relationshipDb, Relationship relationshipInput )
+    {
+        RelationshipType relationshipType = relationshipTypeCache.get( relationshipInput.getRelationshipType() );
+        relationshipDb.setRelationshipType( relationshipType );
+
+        RelationshipItem fromItem = relationshipDb.getFrom();
```
stian-sandvold, line 389: Setting the from and to could probably be done in a separate method to reduce the code duplication :)

stian-sandvold (changes requested):
Looks good, remember to add tests as well! :)

stian-sandvold (approved):

enricocolasante (approved):

PR #12132, opened 2022-10-15
Cpl development 13779

d-bernat:


gnespolino (approved):
Looks good!
Maybe we can just add a specific test to check if produced SQL is as we expect.
(this kind of rest is missing for enrollments and events actually)

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/tei/query/TeiFullQueryTest.java
```diff
@@ -78,7 +77,7 @@ void testSqlQueryRenderingWithOrgUnitNameObject()
         // then
         assertTrue( sql.contains( "ouname" ) );
 
-        assertContains( "order by t_1.\"OUNAME\" ASC", sql );
+        assertContains( "order by t_1.\"OUNAME\" ASC".toLowerCase(), sql.toLowerCase() );
```
maikelarabori, line 80: Should we remove the "toLowerCase()", and match the correct string/case expected?

maikelarabori (approved):

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/tei/query/TeiFullQueryTest.java
d-bernat: I do not think so, it is just about if the string fraction is there and there is nothing case sensitive

PR #12140, opened 2022-10-16
fix: Use null instead of default ID scheme for ADE [DHIS2-13105]

larshelge:
This PR is about aggregate data exchange and is a backport from master/2.39.

For the source analytics request and the target data value import request, for ID schemes, the system would use the default ID scheme (UID) when no ID scheme (null) was specified. This was unfortunate in the (typical) use-case where only a general ID scheme was specified for the source or target request, as the system would inject the UID as the specific data element and org unit ID schemes, effectively overriding the general ID scheme and causing unintentional behavior.

The data value set Java API import options don't support supplying null for specific ID schemes, and setting null would lead to no data values being imported. Adjusted the code to not set the specific ID scheme if it was not set in the aggregate data exchange object.


Philip-Larsen-Donnelly:
Approved for merge by Release Control Board

stian-sandvold (approved):

PR #12146, opened 2022-10-17
chore: add @Nonnull to generic store implementation [TECH-1460]

jbee:
This PR propagates annotations from interface to the base implementation of the hibernate store.

As the store implementation would often handle `null` when collections of identifiers where passed as parameters but the annotation set is `@Nonnull` I also propagated that annotation upwards to the services so we better see and sonar checks that the caller would not call this with a `null` collection. I think this is not 100% safe change but long term this is what we want to have so should there be callers found that can pass null we have to fix that as we discover them. I mostly suspect that not all controllers always make sure input parameters used are non-null collections.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
@@ -36,6 +36,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.persistence.TypedQuery;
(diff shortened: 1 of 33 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
```diff
@@ -128,51 +128,45 @@ private void save( T object, User user, boolean clearSharing )
     {
         String username = user != null ? user.getUsername() : "system-process";
 
-        if ( IdentifiableObject.class.isAssignableFrom( HibernateProxyUtils.getRealClass( object ) ) )
```
jbee, line 131: FYI: there were a a few similar checks but `T` is already bound to `BaseIdentifiableObject` so check must always be true.

mortenoh (approved):

stian-sandvold (approved):

PR #12163, opened 2022-10-18
fix: Code is displayed instead of name for numeric option sets [2.40-DHIS2-13872.2]

d-bernat:


gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -961,7 +962,45 @@ protected void addGridValue( Grid grid, GridHeader header, int index, SqlRowSet
             }
             else
             {
-                grid.addValue( params.isSkipRounding() ? value : MathUtils.getRoundedObject( value ) );
+                if ( header.hasOptionSet() )
+                {
+                    Optional<Option> option = header.getOptionSetObject()
```
maikelarabori, line 967: I think that the code inside this IF statement should be extracted into a method. So you can add a better description of what's being done. The comment inside `.filter` could be improved and become part of the new method description.

maikelarabori (commented):
Thanks @d-bernat, I left a comment for review.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
d-bernat: done

maikelarabori (approved):

PR #12183, opened 2022-10-20
fix: Show error message if tei limit reached [DHIS2-13281]

muilpp:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityInstanceService.java
```diff
@@ -757,16 +757,20 @@ public void validateSearchScope( TrackedEntityInstanceQueryParams params, boolea
...
-            }
-
+            checkIfMaxTeiLimitIsReached( params, maxTeiLimit );
             params.setMaxTeiLimit( maxTeiLimit );
         }
     }
 
+    private void checkIfMaxTeiLimitIsReached( TrackedEntityInstanceQueryParams params, int maxTeiLimit )
+    {
+        int instanceCount = trackedEntityInstanceStore.getTrackedEntityInstanceCountForGrid( params );
```
enricocolasante, line 767: The previous implementation didn't seem correct at all but this new one can have a big impact on performance.
We need to be sure that performance are not impacted too much by this change.

For sure one thing that can already be improved is to call the count only if maxTeiLimit > 0

enricocolasante (commented):
Counting TEIs was always a problem from the performance point of view.
@stian-sandvold how can we make sure this implementation will not impact too much on the system or do we have any other strategy used somewhere else to limit the count?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityInstanceService.java
muilpp: Good idea to run the query only if maxTeiLimit > 0. Other than that, I added a clause to limit the count to maxTeiLimit+1, I guess this will improve the performance a bit. Any other suggestion is welcome.

stian-sandvold (approved):
Looks good!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityInstanceService.java
stian-sandvold: Yeah, doing the limit+1 is probably the most efficient way to do this. I think taking into account pagesize actually allowed for circumventing; Ignoring paging looks like the best way to go

enricocolasante (approved):

PR #12184, opened 2022-10-20
fix: sonar warnings - Object#equals

jbee:
Had a look at sonar issues of type [_Bug_](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=dhis2_dhis2-core).

Sonar correctly found that `BaseIdentifiableObject#equals` was not making a class check and therefore is very likely to violate
```java
a.equals(b)
// must imply:
b.equals(a)
```
I figured that the reason a `isAssignableFrom` check was done was to handle hibernate proxies so I did a exact match compare after getting the actual class
```java
getRealClass(a) == getRealClass(b)
```

`BaseIdentifiableObject#equals` was also overridden so I had a look at all the overrides and addressed similar issues found there.

I also made the implementations more consistent by 
* always comparing specific class field within a extra method called `objectEquals` (found in one of them so I continued the pattern).
* use `Objects.equals` to do null-safe value `equals`
* always call parameter of `equals` `obj` and `other` after the cast
* use `this == obj || obj instanceof X && objectEquals( (X) ob)` instead of a sequence of if-return-true/false.

Please note that some types call the `super.equals( obj )` while others don't. I would assume this is wrong but I kept it "as is" for now.

**Edit:**
Reverted the semantic change to `ConfigurablePeriod`. 
This means a `ConfigurablePeriod` can be equal to a `Period` but the the other way around a `Period` is very unlikely to be equal to a `ConfigurablePeriod` (which is obviously not how `equals` is supposed to work but we don't want to break whatever issue the `ConfigurablePeriod` is working around).

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/ConfigurablePeriod.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/ConfigurablePeriod.java
@@ -28,14 +28,14 @@
 package org.hisp.dhis.period;
 
 import java.util.Date;
+import java.util.Objects;
 
 /**
  * @author Lars Helge Overland
  */
-public class ConfigurablePeriod
-    extends Period
+public class ConfigurablePeriod extends Period
 {
-    private String value;
+    private final String value;
 
     public ConfigurablePeriod( String value )
     {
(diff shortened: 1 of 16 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/ConfigurablePeriod.java
```diff
@@ -63,26 +63,14 @@ public int hashCode()
...
-        if ( !(o instanceof Period) )
-        {
-            return false;
-        }
-
-        final Period other = (Period) o;
+        return this == obj || obj instanceof ConfigurablePeriod && objectEquals( (ConfigurablePeriod) obj );
+    }
 
-        return value.equals( other.getIsoDate() );
```
jbee, line 85: @Lars Not quite sure if this is intentional or by accident. There is just one use of the `ConfigurablePeriod` which seams to just read. I assumed that at no point we would compare a `Period` object with a `ConfigurablePeriod`.

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/ConfigurablePeriod.java
larshelge: I am not quite sure, but we use ConfigurablePeriod in analytics (BaseAnalyticalObject) so I am a bit hesitant to just take it out without understanding unintended effects.

jbee: I reverted the equals semantics to the one that were present.

netroms (approved):

PR #12210, opened 2022-10-26
chore: move new tracker web api into service module

teleivo:
We want to completely isolate tracker code into the tracker module. So the tracker team can make changes as independently from the rest of the product teams as possible.

* duplicated FieldTranslatorSupport as its used by old and new tracker. We deliberately duplicate code like this so that we can simply delete old tracker once its reached its EOL. This allows us to make changes to new tracker without fear of adding a regression to old tracker.

- [ ] TrackerOwnershipManagerTest integration needs a `SystemAuthoritiesProvider` implementation which are all in `dhis-web-commons` module. Why are all implementations in this web-commons module?

We would need to do
`mvn install --update-snapshots -f ./dhis-2/dhis-web/pom.xml` before running integration tests as even though we have dependencies between non-web modules and the web these are not declared in our `pom.xml`. This adds some minutes to the integration test run 😞 

`dhis-web-api` solves this by creating a `SystemAuthoritiesProvider` bean 
https://github.com/dhis2/dhis2-core/blob/808f67257778ea712ff971825e77852341945645/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/WebTestConfiguration.java#L218-L222

Integration tests ideally use a real `SystemAuthoritiesProvider` implementation.


teleivo:
closing this for now as I don't have time to follow up at the moment 😅

PR #12211, opened 2022-10-26
fix: validate duplicate programRule name

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-13502

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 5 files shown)


enricocolasante (approved):
Do we need a script or something else to fix or spot duplication already present in the DB?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/ProgramRuleControllerTest.java
```diff
@@ -57,4 +60,20 @@ void testValidateCondition_NoSuchProgram()
...
+        manager.save( program );
+        assertStatus( HttpStatus.OK, GET( "/programs/{id}", program.getUid() ) );
+        assertStatus( HttpStatus.OK,
+            POST( "/metadata/", "{'programRules':[{'name':'test', 'program':{ 'id':'" + program.getUid() + "'}}]}" ) );
+
+        JsonImportSummary response = POST( "/metadata/",
+            "{'programRules':[{'name':'test', 'program':{ 'id':'" + program.getUid() + "'}}]}" )
+                .content( HttpStatus.CONFLICT ).get( "response" ).as( JsonImportSummary.class );
+        assertEquals( "The Program Rule name test already exist in Program ProgramA",
+            response.getTypeReports().get( 0 ).getObjectReports().get( 0 ).getErrorReports().get( 0 ).getMessage() );
```
jbee, line 77: FYI: Digging down to a particular report might become unstable should more conflicts be found when new validations are added. There are `find` methods inherited from `JsonObject` where you can essentially grab a conflict by the error code and then check its message. I have done this in some tests.

jbee (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
```diff
@@ -0,0 +1,73 @@
...
+import org.hisp.dhis.dxf2.metadata.objectbundle.ObjectBundle;
+import org.hisp.dhis.feedback.ErrorCode;
+import org.hisp.dhis.feedback.ErrorReport;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.programrule.ProgramRule;
+import org.hisp.dhis.programrule.ProgramRuleService;
+import org.springframework.stereotype.Component;
+
+@Component
+@AllArgsConstructor
```
mortenoh, line 44: Maybe better to use `@RequiredArgsConstructor`

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
```diff
@@ -0,0 +1,73 @@
...
+        Program program = programRule.getProgram();
+
+        if ( program == null )
+        {
+            return;
+        }
+
+        program = bundle.getPreheat().get( bundle.getPreheatIdentifier(), program );
+
+        List<ProgramRule> rules = programRuleService.getProgramRule( program );
```
mortenoh, line 61: Should program rules also be part of the preheat?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/ProgramRuleControllerTest.java
vietnguyen: fixed

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
vietnguyen: fixed

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleObjectBundleHook.java
vietnguyen: I don't think so, by design `Program` object  doesn't have a collection of `ProgramRule`.
If you have only one `ProgramRule` in payload, then preheat will only contain the referenced `Program` object.

vietnguyen:
> Do we need a script or something else to fix or spot duplication already present in the DB?

@enricocolasante since we are not backporting this PR and this change won't break any function in program rule engine, I'm thinking just adding it to the 2.40 Upgrade Notes document. So user will be aware of the change and update duplicate names accordingly when they get the error message in Maintenance app.

PR #12221, opened 2022-10-31
fix: NPE in Interpretations API [DHIS2-14010]

maikelarabori:
This fix will prevent an NPE that was recently introduced as part of one of the non-null clean-ups (https://github.com/dhis2/dhis2-core/commit/add6ee3fa78ffe2c7e6864a611b92913ca636a2a)

As now there is no null check anymore for this particular case, we have to ensure we do not pass null objects down the flow.
Basically, we pass a default empty set instead of null, if we do not have any subscribers.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
@@ -54,6 +54,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import javax.annotation.CheckForNull;
+
 import org.hisp.dhis.analytics.AggregationType;
 import org.hisp.dhis.category.CategoryDimension;
 import org.hisp.dhis.category.CategoryOptionCombo;
@@ -1424,8 +1426,14 @@ public abstract class BaseAnalyticalObject
         return subscribers;
     }
 
+    @CheckForNull
     public void setSubscribers( Set<String> subscribers )
     {
+        if ( subscribers == null )
+        {
+            subscribers = new HashSet<>();
+        }
+
         this.subscribers = subscribers;
     }
 
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/interpretation/DefaultInterpretationService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/interpretation/DefaultInterpretationService.java
@@ -27,6 +27,9 @@
  */
 package org.hisp.dhis.interpretation;
 
+import static java.util.Collections.emptySet;
+import static org.apache.commons.lang3.ObjectUtils.firstNonNull;
+
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashSet;
@@ -118,8 +121,6 @@ public class DefaultInterpretationService
     {
         User user = currentUserService.getCurrentUser();
 
-        Set<User> users = new HashSet<>();
-
         if ( user != null )
         {
             interpretation.setCreatedBy( user );
@@ -130,7 +131,7 @@ public class DefaultInterpretationService
             interpretation.setPeriod( periodService.reloadPeriod( interpretation.getPeriod() ) );
         }
 
-        users = MentionUtils.getMentionedUsers( interpretation.getText(), userService );
+        Set<User> users = MentionUtils.getMentionedUsers( interpretation.getText(), userService );
         interpretation.setMentionsFromUsers( users );
         updateSharingForMentions( interpretation, users );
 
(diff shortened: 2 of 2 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/interpretation/DefaultInterpretationService.java
```diff
@@ -287,7 +288,8 @@ private void notifySubscribers( Interpretation interpretation, InterpretationCom
         if ( interpretableObjectSchema.isSubscribable() )
         {
             SubscribableObject object = (SubscribableObject) interpretableObject;
-            Set<User> subscribers = new HashSet<>( userService.getUsers( object.getSubscribers() ) );
+            Set<User> subscribers = new HashSet<>(
+                userService.getUsers( firstNonNull( object.getSubscribers(), emptySet() ) ) );
```
jbee, line 292: Unfortunately we will see a bit of these mismatches in a transition phase. I think to follow the idea here we should not fix this usage of  `object.getSubscribers()` but change the class(es) that have the method returning a collection to never return a `null` collection. So most likely what these lack is being initialized with an empty collection. Or is is the result of deserialisation?

maikelarabori: I thought about it, and I agree it would be better to never return null in such cases. It's a bit complicated in this case though...

In this case, sometimes it returns `null` and other times it returns an empty set. This happens because this column is of type JSON and sometimes it is set to NULL and other times to `[]`, in the database (Sierra Leone, likely on some production DB as well). This could happen for a few reasons, and the code should safely deal with it.

_"have the method returning a collection to never return a `null` collection"_
If one sets the collection to be `null` for any reason, it should still return `null` when we do a get on it. Forcing it to return an empty collection if the respective collection is `null` does not sound 100% consistent to me, but could potentially be done.

_"most likely what these lack is being initialized with an empty collection"_
The referred object has the collection initialized as most of the other entities in the project:
`protected Set<String> subscribers = new HashSet<>()`.

maikelarabori: Thanks for your concern @jbee. If you think we should fix it in a different way feel free to reach out.

jbee: It depends. If we want/need a semantic distinction between this collection being null or empty we obviously need to handle both. If null/empty have the same consequences we should change the code so even a null in the database ends up with an empty collection in the database and equally an empty collection in the code is stored as either null or empty in the database. Both would need to be handled. 

Because if we fix it by adding code that handles the null returned form the getter we would have to make the same fix for all getters and we are likely to introduce new bugs in the future as the rule is that getters returning collections do not return null. So if we deem this situation to be special and have it digress from this expectation we should also go ahead and add the according annotation `@CheckForNull` to make this clear to the readers/callers.

maikelarabori: We don't know if in all cases today or in the future, null and empty will mean the same thing. As you said, it depends on each use case.
If you want to generalize it we will lose the ability to make the distinction. Obviously, we can always add it back by creating a new type (a bit overkill though).

But as I see you really prefer forcing the same semantic, let me check if there is a "good" way to always return empty `[]`.

maikelarabori: After having a conversation with @jbee, we agreed to add a null check in the setter method and create a new instance of the collection if the given parameter is null.

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
```diff
@@ -1424,8 +1426,14 @@ public Set<String> getSubscribers()
...
     }
 
+    @CheckForNull
     public void setSubscribers( Set<String> subscribers )
     {
+        if ( subscribers == null )
+        {
+            subscribers = new HashSet<>();
+        }
+
```
gnespolino, line 1436: I think we should not change the input parameter, rather check if it's null then set this.subscribers to new HashSet.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
```diff
@@ -1424,8 +1426,14 @@ public Set<String> getSubscribers()
         return subscribers;
     }
 
+    @CheckForNull
```
gnespolino, line 1429: From CheckForNull javadoc:
```

/**
 * The annotated element might be null, and uses of the element should check for null.
 * <p>
 * When this annotation is applied to a method it applies to the method return value.
 */

```
so what is the effect of it on a void method ?

gnespolino (approved):
2 small comments

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
maikelarabori: Good point, it will be safer.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
maikelarabori: It should be part of the parameter actually. Good catch.

PR #12235, opened 2022-11-02
fix: consider all data sets of a data element during data import [DHIS2-7836]

jbee:
### Problem
So far we would select a single data set - the "approval data set" and use its properties to validate multi-dataset scenarios. This has several shortcomings potentially leading to valid data being rejected and invalid data potentially being accepted.

### Solution
In a multi-dataset scenario all relevant datasets need to be considered.

In this context the relevant datasets are those that
* are linked to the data element of the imported data value
* are linked to the organisation unit of the imported data value
* are linked the same category combo as the imported data value attribute category combo

When this selection process results in more than one relevant dataset there are two ways to look at this:
1. **strict**: if any of the datasets does not allow the value the value is rejected
2. **permissive**: if any of the datasets permit the value the value is accepted

If checks act strict or permissive is configured via system setting. The default is **strict**.

For the for affected checks 3 individual settings were added since it does make sense to only want to use permissive settings when it comes to data input periods while still using a strict settings related to locking or data approval.

The three new settings are:
* strict data approval
* strict data input periods
* strict data locking (lock exceptions)

As before, if there are no datasets or no existing data set is left after the filtering with the target selection criteria the 4 checks in question are skipped. In other words: Values with no relevant linked datasets are always permitted to be imported with regards to the 4 checks concerning datasets. 

### Automatic Testing
Existing tests for the validations were adjusted.
Also added a multi-dataset scenario for open period validation which was the issue in the reported jira ticket.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
```diff
@@ -646,6 +646,17 @@ public Period createPeriod( String isoDate )
     // CalendarPeriodType
     // -------------------------------------------------------------------------
 
+    public final Period getFuturePeriod( int periodOffset )
```
jimgrace, line 649: I'm not sure what is gained by moving this logic from `DataElement`. It still has the funny concept of "Future" which is different from "Next" by -1. If moved, this public method needs JavaDoc to explain what it does.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
```diff
@@ -646,6 +646,17 @@ public Period createPeriod( String isoDate )
...
+    {
+        Period period = createPeriod();
+
+        // Rewind one as 0 open periods implies current period is locked
```
jimgrace, line 653: This comment may be out of place here in `PeriodType` to talk about locking. (Perhaps you can explain this better in JavaDoc for the method.)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
```diff
@@ -646,6 +646,17 @@ public Period createPeriod( String isoDate )
...
+
+        period = getPreviousPeriod( period );
+
+        return getNextPeriod( period, periodOffset );
```
jimgrace, line 657: We talked about this logic. I see the need for it now. If periodOffset is 0, then this returns the previous period. If you just replace these two lines with `return getNextPeriod( period, periodOffset - 1 )` that would return the current period when periodOffset is 0 because -1 results in no shift within `getNextPeriod`.

However you could replace these two lines with `return getShiftedPeriod( period, periodOffset - 1 )`, though it won't backport unless you backport `getShiftedPeriod` too -- I just added it recently to work with yearToDate and periodOffset stuff, thinking it might be useful for other things some day!

jimgrace (approved):
Looks good. @jbee and I already agreed on the approach, and I like the way it's been implemented. Just minor comments about open future periods.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
jbee: I needed to move this since I call this from validation were I don't use the data element's period type.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
jbee: I would if I knew :D I just moved this code.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
jbee: Thanks for providing this. For now I just keep it the way it is.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
jbee: done

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -1070,6 +1070,12 @@ private ImportContext createDataValueSetImportContext( ImportOptions options, Da
                 || settings.getBoolSetting( SettingKey.DATA_IMPORT_STRICT_ATTRIBUTE_OPTION_COMBOS ) )
             .strictOrgUnits( options.isStrictOrganisationUnits()
                 || settings.getBoolSetting( SettingKey.DATA_IMPORT_STRICT_ORGANISATION_UNITS ) )
+            .strictDataSetApproval( options.isStrictDataSetApproval()
+                || settings.getBoolSetting( SettingKey.DATA_IMPORT_STRICT_DATA_SET_APPROVAL ) )
+            .strictDataSetLocking( options.isStrictDataSetLocking()
+                || settings.getBoolSetting( SettingKey.DATA_IMPORT_STRICT_DATA_SET_LOCKING ) )
+            .strictDataSetLocking( options.isStrictDataSetInputPeriods()
```
jimgrace, line 1077: Copy/Paste/NoEdit error: Should be `.strictDataSetInputPeriods( options...`

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/ImportContext.java
```diff
@@ -222,12 +270,38 @@ public String getStoredBy( DataValueEntry dataValue )
...
+     * The set of {@link DataSet} that needs to be considered contains all
+     * {@link DataSet}s whose values intersect with the checked value's
+     * dimensions.
+     *
+     * These have
+     * <ul>
+     * <li>the same {@link DataElement} as the value</li>
+     * <li>include the same {@link OrganisationUnit} as the value</li>
+     * <li>the same {@link org.hisp.dhis.category.CategoryCombo} as the value's
+     * attribute option combo</li>
```
jimgrace, line 283: To be more accurate I suggest:
```
     * <li>the same {@link org.hisp.dhis.category.CategoryCombo} as the category
     * combo for the value's attribute option combo</li>
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
```diff
@@ -375,8 +375,7 @@ private boolean objectEquals( Period other )
     @Override
     public String toString()
     {
-        return "[" + (dateField == null ? "DEFAULT" : dateField) + " | "
-            + (periodType == null ? "" : periodType.getName() + ": ") + startDate + " - " + endDate + "]";
+        return getIsoDate();
```
jimgrace, line 378: Why did you make this change? Was it because `toString` was ugly, unintuitive and overly-complicated? If so I agree with these reasons. However if `dateField` is present then `toString` won't show the difference whereas `equals` will. I'm not sure this is good. How about:

        return getIsoDate() + (isDefault() ? "" : ":" + dateField);

(This is more like the how the dateField was first parsed in `IsoPeriodHolder`.)

jimgrace (changes requested):
Nice code. I tried to do a careful review and I like it, just a couple of comments. Jan and I agreed on the three new system settings in a lengthy Slack Huddle design session. Personally I think some installations may want to be strict on data approvals (approved data will not change) but loose on open periods (if some DataSet has an open period, it is to allow data to be entered there). So hopefully all the new settings will satisfy everyone.

Jan and I agree that in the future we should refactor the validation code used by `DataValueController` to be consistent with this logic and the new system settings.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
jbee: Nice catch. Fixed.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/ImportContext.java
jbee: I'll take it :D

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
jbee: I found myself having to expand into the details of an period object to read what period it was referring to so I thought this can be more readable/informative simply by returning its string form, the ISO date. 

Added the dateField as you suggested.

jbee:
Thanks @jimgrace for the detailed review. Fixed issues you pointed out and went with your suggestions.

jimgrace (approved):
Looks good, thanks!

larshelge (approved):

PR #12236, opened 2022-11-02
fix: Remove usage of ConfigurablePeriod class [TECH-1473]

maikelarabori:
The class `ConfigurablePeriod` is used in a single place and can be replaced by `Period`.
This will prevent equality issues and make the code more consistent.
It does not seem to cause any side effects in my local tests. If something arises this change can be easily rolled back.

The main change can be seen at line: https://github.com/dhis2/dhis2-core/pull/12236/files#diff-6c1b0854268e6e35f253669d0e957e67ea2d627a768d987a675c0e915bbfd2f2R678

The other changes are related to removing the `final` keyword, as this is not considered good practice in our project.


jbee (approved):
Nice!

d-bernat (approved):

gnespolino (commented):
Looks good!

PR #12247, opened 2022-11-04
feat: Multi-calendar support in resource tables and analytics engine

abyot:
- Add Ethiopian calendar i18n translation for month and day names
- Use available data for earliest and latest year calculation
- Add 5 years (+/-) on available data year for resource tables calculation


larshelge:
Closing due to inactivity, please open a new PR when ready :+1:

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
```diff
@@ -255,4 +258,29 @@ public void dropAllSqlViews( JobProgress progress )
...
+        if ( availableDataYears.isEmpty() )
+        {
+            availableDataYears = new ArrayList<>();
+            availableDataYears.add( now().getYear() );
+        }
+
+        int firstYear = availableDataYears.get( 0 );
+        int lastYear = availableDataYears.get( availableDataYears.size() - 1 );
+
+        for ( int i = 0; i < 5; i++ )
```
maikelarabori, line 276: I believe would be nice to extract the number `5` into a constant to give it a better context.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
```diff
@@ -255,4 +258,29 @@ public void dropAllSqlViews( JobProgress progress )
         progress.startingStage( "Drop SQL views", nonQueryViews.size(), SKIP_ITEM );
         progress.runStage( nonQueryViews, SqlView::getViewName, sqlViewService::dropViewTable );
     }
+
+    @Override
+    public List<Integer> generateDataYears()
```
maikelarabori, line 263: Would be nice to have a unit test for this public method covering the possible scenarios.

maikelarabori (approved):
Thanks for the changes @abyot. I just left two minor comments.
Please take a look.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
abyot: this is a good point, done.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/jdbc/JdbcResourceTableStore.java
```diff
@@ -211,4 +211,22 @@ public void batchUpdate( int columns, String tableName, List<Object[]> batchArgs
 
         jdbcTemplate.batchUpdate( builder.toString(), batchArgs );
     }
+
+    @Override
+    public List<Integer> getAvailableDataYears()
```
larshelge, line 216: I am wondering if the JdbcResourceTableStore is not the right place for this method, as it does not relate to resource tables?

larshelge:
@abyot you think you could add a description to the PR which explains what this change does and why it is needed?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
```diff
@@ -353,6 +360,10 @@ protected List<String> getPartitionChecks( AnalyticsTablePartition partition )
     @Override
     protected void populateTable( AnalyticsTableUpdateParams params, AnalyticsTablePartition partition )
     {
+        List<Integer> availableDataYears = resourceTableService.generateDataYears();
```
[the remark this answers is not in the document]
abyot: I am happy to get a suggestion for the right place. 

Both methods were using constants - 1975 for earliest and current year + 25 for latest. This has made things unnecessarily slow and failing in the case of Nepal. What the method does is provide latest and earliest years based on available data (± 5 years). We are calling this methods everywhere we need earliest and latest years.

abyot:
> @abyot you think you could update the title and add a description to the PR which explains what this change does and why it is needed?

done.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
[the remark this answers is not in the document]
maikelarabori: This makes me wonder if we should not permit more years in the past.
If we have clients that have very old events/enrollments they won't be able to see them anymore. We could always start in a defined year in the past (that never changes), and permit the current year +5 in the future to ensure we won't cause issues related to the past years.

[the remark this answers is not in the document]
abyot: ± 5 is just a buffer ... otherwise those very old events will be part of the data and the method will return their year.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
```diff
@@ -255,4 +259,31 @@ public void dropAllSqlViews( JobProgress progress )
...
+    {
+        List<Integer> availableDataYears = new ArrayList<>();
+
+        availableDataYears.addAll( resourceTableStore.getAvailableDataYears() );
+
+        if ( availableDataYears.isEmpty() )
+        {
+            availableDataYears = new ArrayList<>();
```
larshelge, line 272: Do we need to instantiate a new list here?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTableService.java
```diff
@@ -116,4 +118,11 @@
      * Drop all SQL views.
      */
     void dropAllSqlViews( JobProgress progress );
+
+    /**
+     * Returns all data years available in the system (both from aggregate and
+     * tracker) with 5 years previous and future additions.
+     * [extra_5_previous_years, data_years, extra_5_future_year]
+     */
+    List<Integer> generateDataYears();
```
larshelge, line 127: Rename to `getDataYears()` as we are simply returning a result of the query.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTableStore.java
```diff
@@ -63,4 +63,11 @@
      * @param batchArgs the arguments to use for the update statement.
      */
     void batchUpdate( int columns, String tableName, List<Object[]> batchArgs );
+
+    /**
+     * Returns all data years available in the system (both from aggregate and
+     * tracker) with 5 years previous and future additions.
+     * [extra_5_previous_years, data_years, extra_5_future_year]
+     */
+    List<Integer> getAvailableDataYears();
```
larshelge, line 72: Just curious what the difference between `generateDataYears` and `getAvailableDataYears` are, as they have the same Javadoc?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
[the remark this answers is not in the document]
larshelge: I can see we need the data periods method both for resource tables and analytics service/code. I think we could contain these methods in a new Java service, something like `DataPeriodProvider`.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/DatePeriodResourceTable.java
```diff
@@ -93,10 +93,8 @@ public Optional<List<Object[]>> getPopulateTempTableContent()
 
         List<Object[]> batchArgs = new ArrayList<>();
 
-        // TODO Create a dynamic solution instead of having fixed dates
-
-        Date startDate = new Cal( FIRST_YEAR_SUPPORTED, 1, 1, true ).time();
-        Date endDate = new Cal( LATEST_YEAR_SUPPORTED + 1, 1, 1, true ).time();
+        Date startDate = new Cal( objects.get( 0 ), 1, 1, true ).time();
```
maikelarabori, line 96: We could create variables here as well, to make it more explicit, ie:
``` 
Date startDate = new Cal( objects.get( FIRST_YEAR ), 1, 1, true ).time();
Date endDate = new Cal( objects.get( LAST_YEAR ), 1, 1, true ).time();
```

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/DatePeriodResourceTable.java
```diff
@@ -48,11 +48,11 @@
  * @author Lars Helge Overland
  */
 public class DatePeriodResourceTable
-    extends ResourceTable<Period>
+    extends ResourceTable<Integer>
 {
     private final String tableType;
 
-    public DatePeriodResourceTable( List<Period> objects, String tableType )
+    public DatePeriodResourceTable( List<Integer> objects, String tableType )
```
maikelarabori, line 55: We could name the List<Integer> as `years`, instead of `objects`.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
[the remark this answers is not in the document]
abyot: `DataYearProvider` or `DataPeriodProvider` ? also where to put his service?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
abyot: correct, we don't need that.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/DatePeriodResourceTable.java
abyot: ok, can add +1. I thought the extra 5 years covers for that.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/DatePeriodResourceTable.java
abyot: I think we are going with `objects` because of the superclass?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/DatePeriodResourceTable.java
```diff
@@ -93,10 +93,8 @@ public Optional<List<Object[]>> getPopulateTempTableContent()
...
         List<Object[]> batchArgs = new ArrayList<>();
 
-        // TODO Create a dynamic solution instead of having fixed dates
-
-        Date startDate = new Cal( FIRST_YEAR_SUPPORTED, 1, 1, true ).time();
-        Date endDate = new Cal( LATEST_YEAR_SUPPORTED + 1, 1, 1, true ).time();
+        Date startDate = new Cal( objects.get( 0 ), 1, 1, true ).time();
+        Date endDate = new Cal( objects.get( objects.size() - 1 ), 1, 1, true ).time();
 
         List<Period> dailyPeriods = new DailyPeriodType().generatePeriods( startDate, endDate );
```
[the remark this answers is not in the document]
abyot: This is the same as above and fixed.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodDataProvider.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Component responsible for fetching, extracting and providing specific data
+ * related to existing periods in the database.
+ *
+ * @author maikel arabori
+ */
+@Component
+@RequiredArgsConstructor
+public class PeriodDataProvider
```
larshelge, line 50: Could we move this class to the dhis-service-core project? I don't think we want services with database interaction in the API project.

dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/resourcetable/table/PeriodDataProviderTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Unit tests for {@link PeriodDataProvider}.
+ *
+ * @author Abyot Asalefew Gizaw
+ */
+@ExtendWith( MockitoExtension.class )
+class PeriodDataProviderTest
```
larshelge, line 56: We should put this test in the same project as the service itself. We could put both `PeriodDataProvider` and this test class in dhis-service-core.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/jdbc/JdbcResourceTableStore.java
```diff
@@ -197,4 +197,22 @@ public void batchUpdate( int columns, String tableName, List<Object[]> batchArgs
...
```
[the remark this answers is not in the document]
maikelarabori: Actually, I forgot to remove this method. It is the one moved into `PeriodDataProvider`.
I removed it now :+1:

dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/resourcetable/table/PeriodDataProviderTest.java
maikelarabori: Done.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodDataProvider.java
maikelarabori: Right, that makes sense. Done.

PR #12255, opened 2022-11-06
feat: OpenAPI generator and endpoint [DHIS2-14004]

jbee:
### Summary

* adds general OpenAPI tool-chain
* adds new OpenAPI controller ("life" analysis)
* adds command line tool ("static" analysis)
* adds `@OpenApi` family of annotations to tweak the generated document in the controllers
* adds a group tag to every controller (via `@OpenApi.Tags`) so they can be split into named groups
* replaces "invisible" parameters for Gist API with a dedicated parameter object
* replaces "invisible" parameters for Datastore query API with a dedicated parameter object
* adds `@OpenApi` annotations to CRUD controller base classes (so inherited methods are mostly described)
* adds some more `@OpenApi` annotations to individual controllers mostly to correct the response type (where it is not visible from the method signature)

Many controllers will need further tweaking by adding `@OpenApi` annotations but the general mechanisms should exist.

### Automatic Testing
The controllers where the API was actually changed by adding dedicated parameter objects is tested by many tests.
New tests were added to test the added OpenAPI document generation endpoints.

### Manual Testing
* make sure controller classes are compiled
* run `OpenApiTool` via `main` method

### Examples
Include all controllers and group them into documents (files) by their first tag (on the controller class):
```bash
--group .

Generated Documents
  openapi-analytics.json         [ 12 controllers, 114 schemas]
  openapi-data.json              [ 29 controllers, 217 schemas]
  openapi-messaging.json         [  7 controllers, 107 schemas]
  openapi-metadata.json          [ 51 controllers, 280 schemas]
  openapi-system.json            [ 20 controllers, 173 schemas]
  openapi-tracker.json           [ 33 controllers, 291 schemas]
  openapi-ui.json                [ 12 controllers, 132 schemas]
  openapi-user.json              [ 14 controllers,  99 schemas]
```

Generate a document just for `/users` path/controller:
```bash
/users openapi-users-only.json

Generated Documents
  openapi-users-only.json        [  1 controllers,  61 schemas]
```

Generate a document with all controllers tagged `"user"`:
```bash
user openapi-user.json

Generated Documents
  openapi-user.json              [ 14 controllers,  99 schemas]
```


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractCrudController.java
```diff
@@ -165,6 +167,9 @@
     // OLD PATCH
     // --------------------------------------------------------------------------
 
+    @OpenApi.Ignore
```
larshelge, line 170: Just curious about the effect of ignoring and adding params annotations at the same time?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractGistReadOnlyController.java
```diff
@@ -94,47 +103,65 @@
...
+    public void getObjectListGistAsCsv( GistParams params, HttpServletResponse response )
         throws IOException
     {
-        gistToCsvResponse( response, createGistQuery( request, getEntityClass(), GistAutoType.S )
+        gistToCsvResponse( response, createGistQuery( params, getEntityClass(), GistAutoType.S )
             .toBuilder().typedAttributeValues( false ).build() );
     }
 
+    @OpenApi.Response( status = BAD_REQUEST, value = WebMessage.class )
+    @OpenApi.Response( status = NOT_FOUND, value = WebMessage.class )
```
larshelge, line 158: Speculative comment, should we have a global way way of defining the behavior for 400, 404, 409? Seems to be the same for many endpoints?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractCrudController.java
jbee: Ignore wins, I first went through adding parameters, later decided to ignore this one. We could remove the added parameter annotations now.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractGistReadOnlyController.java
jbee: Maybe. I was mostly thinking about using this to provide better descriptions as to why this might happen. But so far I did not add such texts to them. But that was the plan.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractCrudController.java
larshelge: Got it.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractGistReadOnlyController.java
larshelge: Right. Not a big issue. Can imagine the reason for 404 is probably mostly the same, and starting to explain 409 might be a lot of text. All good.

larshelge (approved):

netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -78,7 +77,7 @@
     URL( String.class, false ),
     FILE_RESOURCE( String.class, true, FileTypeValueOptions.class ),
     IMAGE( String.class, false, FileTypeValueOptions.class ),
-    GEOJSON( GeoJSON.class, false ),
+    GEOJSON( String.class, false ),
```
jbee, line 80: Just to highlight: This was needed because using GeoJSON here can become an issue with reflection from/in modules that do not have the dependency on the geotools artefact. Also this had no actual function apart from controlling the header in the grid view of values.

PR #12276, opened 2022-11-09
fix: Sharing settings bug when Dashboard is edited [DHIS2-6556]

maikelarabori:
In the very past, some workaround was added to the Dashboard, so users that do not have access to certain DashboardItems could still be able to see them.
The edit/update operation of the Dashboard was not taken into consideration though.

This causes DashboardItems to "disappear" from the Dashboard when users that do not have access to that particular item execute an update operation on the respective Dashboard.

This fix should address this problem by allowing DashboardItems of a Dashboard to bypass the sharing settings checking, so the items will not disappear and remain in the Dashboard as the user had full access.

@mortenoh, as I know you were involved in the initial workaround, I would like your point of view before making this PR official. I'm open to other solutions/fixes. Thanks!


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/DefaultQueryService.java
```diff
@@ -143,7 +149,7 @@ private long countObjects( Query query )
         return criteriaQueryEngine.count( pQuery );
     }
 
-    private List<? extends IdentifiableObject> queryObjects( Query query )
+    private List<? extends IdentifiableObject> queryObjects( Query query, boolean ignoreSharingSettings )
```
vietnguyen, line 152: Should we add `ignoreSharingSettings` as a property of `Query` object instead of new method parameter.

vietnguyen:
@maikelarabori thanks for the PR, code looks good, just one minor comment regarding the `ignoreSharingSettings` property.

From 2.37 we have introduced the [Cascade Sharing for Dashboard](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-237/sharing.html#cascade-sharing-for-dashboard), and this use case seems to be applicable for that function. We could add a validation in `DashboardObjectBundleHook` to validate user permission when update a Dashboard. Then user would need to ask the admin to do Cascade Sharing for them if needed. 

So maybe this PR can be merged to 2.39 and older versions. Then from 2.40 we could implement the validation approach.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/DefaultQueryService.java
maikelarabori: Done

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/preheat/DefaultPreheatService.java
```diff
@@ -432,6 +442,65 @@ public void validate( PreheatParams params )
...
+        Map<Class<? extends IdentifiableObject>, List<IdentifiableObject>> objects )
+    {
+        if ( MapUtils.isEmpty( objects ) )
+        {
+            return emptySet();
+        }
+
+        Set<String> itemsUid = new HashSet<>();
+
+        for ( Map.Entry<Class<? extends IdentifiableObject>, List<IdentifiableObject>> entry : objects.entrySet() )
```
vietnguyen, line 462: can replace this for loop with `objects.get( Dashboard.class ) `

maikelarabori: Done

vietnguyen (approved):

maikelarabori:
This is being fixed by the Platform team using a different approach. So I'm closing this PR.

PR #12282, opened 2022-11-14
fix: getSystemSetting causes performance issue (2.37)

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14114

### Issue: 
- `SystemSettingManager.getSystemSetting()` use `settingCache`. Currently `@Transaction( readOnly = true )` commit the transaction for every function call even with cache hits, which is causing performance issue when we call this method in a for loop.
- We want to initiate the transaction only when cache miss.

### Fix:
- Use `TransactionTemplate.execute()` to manually initiate transaction when cache misses in `SystemSettingManager.getSystemSettingOptional()`

### Test:
- The affected function of this issue is in Aggregate Data Entry app. Just need to test if the forms can be loaded successfully.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/DefaultSystemSettingManager.java
+++ b/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/DefaultSystemSettingManager.java
@@ -50,6 +50,7 @@ import org.jasypt.encryption.pbe.PBEStringEncryptor;
 import org.jasypt.exceptions.EncryptionOperationNotPossibleException;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.transaction.support.TransactionTemplate;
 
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/hibernate/HibernateSystemSettingStore.java
```diff
@@ -50,6 +51,13 @@ public HibernateSystemSettingStore( SessionFactory sessionFactory, JdbcTemplate
         super( sessionFactory, jdbcTemplate, publisher, SystemSetting.class, true );
     }
 
+    @Override
+    @Transactional
```
jbee, line 55: Should this be readonly?

vietnguyen: yes, updated

stian-sandvold (approved):

mortenoh (approved):

PR #12283, opened 2022-11-14
refactor: Clean up module dependencies

enricocolasante:
After creating `dhis-test-integration` module a few dependencies that were needed before to run the tests in every module are now redundant and the analyzer was not complaining because we had a ignore rules on such modules.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/pom.xml
+++ b/dhis-2/dhis-services/dhis-service-administration/pom.xml
@@ -34,10 +34,6 @@
       <artifactId>dhis-service-schema</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.hisp.dhis</groupId>    
-      <artifactId>dhis-service-validation</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.hisp.dhis</groupId>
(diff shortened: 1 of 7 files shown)


teleivo (approved):

teleivo:
Just curious why do we for example have an ignore declared unused config for a module like dhis-service-dxf2?
https://github.com/dhis2/dhis2-core/pull/12283/files#diff-4caf3d06d60c2fda63c091b5d0a1f29e447208cae834b134367e1342cf5ffb40L715

enricocolasante:
> Just curious why do we for example have an ignore declared unused config for a module like dhis-service-dxf2? https://github.com/dhis2/dhis2-core/pull/12283/files#diff-4caf3d06d60c2fda63c091b5d0a1f29e447208cae834b134367e1342cf5ffb40L715

It is used by jetty
https://github.com/dhis2/dhis2-core/blob/72f1c3f7467a0909060c90a5a5f01cd4717bafe2/dhis-2/dhis-web-embedded-jetty/pom.xml#L177

I think at some point we should review that long list of ignore rules

stian-sandvold (approved):

PR #12285, opened 2022-11-14
fix: maintain order for data elements/attributes in tracker data view [DHIS2-14084]

zubaira:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackerdataview/TrackerDataView.java
```diff
@@ -60,10 +60,10 @@ public class TrackerDataView implements Serializable
     @JsonProperty
     @JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
     @Builder.Default
-    private Set<String> attributes = new HashSet<>();
+    private Set<String> attributes = new LinkedHashSet<>();
```
larshelge, line 63: You think you could explain the reason for not using a List? Uniqueness?

enricocolasante: I think it will be useful add a test in `web-api-test` that goes through the different possible cases (uniqueness, order...)

enricocolasante (changes requested):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/RelationshipTypeControllerTest.java
```diff
@@ -0,0 +1,109 @@
...
+                    + "'}, 'program': {'id': '" + program + "' }},'name': 'test-rel'}" ) );
+
+        JsonList<JsonString> attributes = GET(
+            "/relationshipTypes/" + relationshipTypeId + "?fields=toConstraint[trackerDataView[attributes]]" ).content()
+                .getList( "toConstraint.trackerDataView.attributes", JsonString.class );
+
+        assertEquals( 3, attributes.size() );
+        assertEquals( attrA, attributes.get( 0 ).string() );
+        assertEquals( attrB, attributes.get( 1 ).string() );
+        assertEquals( attrC, attributes.get( 2 ).string() );
```
jbee, line 107: We have `org.hisp.dhis.utils.Assertions#assertContainsOnly` to assert set semantics.
If order matters here I think you should rather do one assert like this:
```java
assertEquals(List.of(attrA, attrB, attrC), attributes.toList(JsonString::string));
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackerdataview/TrackerDataView.java
jbee: Remember that if the field is still a `Set` the setter can be called with any implementation including those that do not care about order. This might be a problem. Can't say without knowing the case here.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/RelationshipTypeControllerTest.java
zubaira: I will try this. but I suspect `attributes.toList` is converting it to `List` which might not preserve any specific order

jbee: It is a `JsonList` => `List` transformation. The order is preserved.

zubaira: resolved

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/RelationshipTypeObjectBundleHookTest.java
```diff
@@ -143,33 +143,38 @@ public void setUp()
         personConstraint.setTrackedEntityType( personTrackedEntityType );
         personConstraint.setRelationshipEntity( RelationshipEntity.TRACKED_ENTITY_INSTANCE );
         personConstraint.setTrackerDataView(
-            TrackerDataView.builder().attributes( Sets.newHashSet( trackedEntityAttribute.getUid() ) ).build() );
+            TrackerDataView.builder()
+                .attributes( Sets.newLinkedHashSet( Sets.newHashSet( trackedEntityAttribute.getUid() ) ) ).build() );
```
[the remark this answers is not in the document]
zubaira: `Sets.newLinkedHashSet` needs `Iterable`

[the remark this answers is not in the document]
jbee: Right, then I personally would rather use `new LinkedHashSet(Set.of( trackedEntityAttribute.getUid() ))` as it is less overhead and long term we want to get away from the `Sets` util but it is still fine to you if you prefer it.

jbee (approved):

enricocolasante (approved):

PR #12289, opened 2022-11-15
refactor: Clean up more module dependencies

enricocolasante:


dhis-2/dhis-web-api-test/pom.xml
```diff
@@ -21,6 +21,11 @@
         <artifactId>maven-dependency-plugin</artifactId>
         <configuration>
           <failOnWarning>true</failOnWarning>
+          <ignoredUnusedDeclaredDependencies>
+            <ignoredUnusedDeclaredDependency>com.h2database:h2</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>org.skyscreamer:jsonassert</ignoredUnusedDeclaredDependency>
```
teleivo, line 26: Why do we need to ignore it as unused declared? Its imported in TrackerImportParamsSerdeTest 🤔

dhis-2/dhis-web-api-test/pom.xml
```diff
@@ -21,6 +21,11 @@
...
+            <ignoredUnusedDeclaredDependency>com.cronutils:cron-utils</ignoredUnusedDeclaredDependency>
```
teleivo, line 27: What uses this one and does not import it?

dhis-2/pom.xml
```diff
@@ -672,29 +672,23 @@
...
             <failOnWarning>true</failOnWarning>
             <ignoredUnusedDeclaredDependencies>
-              <ignoredUnusedDeclaredDependency>org.projectlombok:lombok</ignoredUnusedDeclaredDependency>
               <ignoredUnusedDeclaredDependency>org.junit.jupiter:junit-jupiter</ignoredUnusedDeclaredDependency>
-              <ignoredUnusedDeclaredDependency>org.springframework.security:spring-security-core</ignoredUnusedDeclaredDependency>
               <ignoredUnusedDeclaredDependency>net.sf.jasperreports:jasperreports</ignoredUnusedDeclaredDependency>
               <ignoredUnusedDeclaredDependency>org.mockito:mockito-inline</ignoredUnusedDeclaredDependency>
               <ignoredUnusedDeclaredDependency>org.apache.logging.log4j:log4j-slf4j-impl</ignoredUnusedDeclaredDependency>
+              <ignoredUnusedDeclaredDependency>javax.servlet:javax.servlet-api</ignoredUnusedDeclaredDependency>
+              <ignoredUnusedDeclaredDependency>org.apache.jclouds.provider:aws-s3</ignoredUnusedDeclaredDependency>
```
teleivo, line 680: `dhis-service-core` seems to be the one declaring most jclouds dependencies. I think it would be great if we could move the dependencies that are specific to individual modules inside those modules. So we don't add these ignore configurations globally. This should lead to cleaner poms over time.

Adding a `maven-dependency-plugin` config in `dhis-service-core` with this attribute
https://www.baeldung.com/maven-plugin-override-parent#2-append-parent-configuration
might work

dhis-2/dhis-web-api-test/pom.xml
enricocolasante: I need to comment on the ignore unused as they are not intuitive at all.
On this particular one, I now removed the runtime dependency.
It was happening in https://github.com/dhis2/dhis2-core/blob/8298dbaa92923ece2228bb5e9516ae4d65d9d4ef/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/StaticContentControllerTest.java#L209

Changing the json assert to a string assert removed the need to have `jsonassert` dependency.
The error was `java.lang.NoClassDefFoundError: org/skyscreamer/jsonassert/JSONAssert`

dhis-2/dhis-web-api-test/pom.xml
enricocolasante: Added a comment in the pom explaining the issue and where it is happening.
This would probably require some more investigation, I am not sure why this is happening

dhis-2/pom.xml
enricocolasante: I agree with the idea. I just think that before that we should really validate that such dependencies belong in the module where they are declared.
Now it is difficult to really understand it because they are only needed at runtime.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/StaticContentControllerTest.java
```diff
@@ -206,7 +206,7 @@ void verifyErrorWhenStoringInvalidMimeType()
         final String error = buildResponse( "Unsupported Media Type", 415, "WARNING", null );
         mvc.perform( multipart( URL + LOGO_BANNER )
             .file( new MockMultipartFile( "file", "testlogo.png", IMAGE_JPEG.toString(), "image".getBytes() ) )
-            .session( session ) ).andExpect( content().json( error ) )
+            .session( session ) ).andExpect( content().string( error ) )
```
teleivo, line 209: String comparison of 2 JSON sounds like it could lead to a brittle test just because of whitespace issues. I think this here can be replaced with a json tree assertion? Like we do in the convenience controller tests. `buildResponse` builds the expected JSON using Gson. So we can replace that with json-tree.

enricocolasante: It makes sense.
I am focusing just on the dependencies now, I am putting back the jsonassert dependency and add a comment explaining what is the issue

teleivo (approved):

PR #12291, opened 2022-11-15
DHIS2-12036 Use search scope if no owner found

muilpp:
If a program has no owner, use search scope instead of capture scope to allow registering a TEI to the program


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
```diff
@@ -254,15 +253,20 @@ public boolean hasAccess( User user, TrackedEntityInstance entityInstance, Progr
             return true;
         }
 
-        OrganisationUnit ou = getOwner( entityInstance.getId(), program, entityInstance::getOrganisationUnit );
+        OrganisationUnit ownerOrgUnit = getOwner( entityInstance.getId(), program );
+        boolean hasOwner = ownerOrgUnit != null;
 
-        if ( program.isOpen() || program.isAudited() )
+        if ( program.isOpen() || program.isAudited() || !hasOwner )
```
enricocolasante, line 259: Maye it will be more readable extracting the if (!hasOwner logic) so having it as a guard clause and then another if(!program.isOpen() && !program.isAudited())

enricocolasante (approved):
Just one minor suggestion

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
```diff
@@ -254,15 +253,21 @@ public boolean hasAccess( User user, TrackedEntityInstance entityInstance, Progr
...
 
-        OrganisationUnit ou = getOwner( entityInstance.getId(), program, entityInstance::getOrganisationUnit );
+        OrganisationUnit ownerOrgUnit = getOwner( entityInstance.getId(), program );
+        boolean hasOwner = ownerOrgUnit != null;
 
-        if ( program.isOpen() || program.isAudited() )
+        if ( !hasOwner )
         {
-            return organisationUnitService.isInUserSearchHierarchyCached( user, ou );
+            ownerOrgUnit = entityInstance.getOrganisationUnit();
```
teleivo, line 261: I would inline `entityInstance.getOrganisationUnit()`. Assigning it to `ownerOrgUnit`  is confusing to me since the program has no owner.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerOwnershipManagerTest.java
```diff
@@ -136,4 +143,20 @@ void testTransferOwnership()
         assertFalse( trackerOwnershipAccessManager.hasAccess( userA, entityInstanceA1, programA ) );
         assertTrue( trackerOwnershipAccessManager.hasAccess( userB, entityInstanceA1, programA ) );
     }
+
+    @Test
+    void testAccessToClosedProgramWithoutOwnershipOnlyIfUserInSearchScope()
+    {
+        assertTrue( trackerOwnershipAccessManager.hasAccess( userA, entityInstanceA1, programB ) );
```
teleivo, line 150: Its hard to understand why one should be true and the other false just from reading the test 😅 It's often difficult with how our tests are structured.

Is the reason why userB does not have access because its a closed program?
userA has access because programB is in search scope right? Where is that set?

IMHO I like to see these important bits in the test. So a little setup that shows these.

If you wanted we could pair on it to see if there is a way to structure the tests differently 😄

teleivo (commented):
Added some small comments.

My main concern is this:

> If a program has no owner, use search scope instead of capture scope to allow registering a TEI to the program

Reads to me as `capture scope` is loosing its meaning. Maybe I misunderstand 🤔 

I would also want the review of @abyot as this change relates to ownership where Abyot has more experience.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerOwnershipManagerTest.java
muilpp: I refactored this and now I'm testing one thing at a time with a more explicit naming, so hopefully it's a bit easier to understand what each test does. If you think there's still room for significant improvement, I'd be happy to pair program and find a better solution.

muilpp:
@teleivo this was discussed in the design meeting last week, it was decided that when a program has no enrollments, we would allow anyone with search scope access to register a new TEI.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
```diff
@@ -254,15 +253,20 @@ public boolean hasAccess( User user, TrackedEntityInstance entityInstance, Progr
...
 
-        OrganisationUnit ou = getOwner( entityInstance.getId(), program, entityInstance::getOrganisationUnit );
+        OrganisationUnit ownerOrgUnit = getOwner( entityInstance.getId(), program );
+        boolean hasOwner = ownerOrgUnit != null;
 
-        if ( program.isOpen() || program.isAudited() )
+        if ( !hasOwner )
         {
-            return organisationUnitService.isInUserSearchHierarchyCached( user, ou );
+            return organisationUnitService.isInUserSearchHierarchyCached( user, entityInstance.getOrganisationUnit() );
```
abyot, line 261: I think we go to search scope if program is open or audited, otherwise we restrict to capture scope.

stian-sandvold: The discussion we had in the design meeting, was about creating an enrollment in a CLOSED program, when there is no ownership within the context of that program.

As we agreed, ownership only makes sense in the case of a program. If no enrollment exists for a program, ownership goes out of the window (Since we don't have ownership in the context of that program). In that case we fall back to the TEI ou.

In all cases (Let me know if I am wrongly assuming here), the user needs CAPTURE scope access for a OU to create any data there, including an enrollment.

In this case:
TEI.ou = A
Enrollment.ou = B

A needs to be in the search scope (We need to be able to see the TEI)
B needs to be in the capture scope (We need "capture access" to the ou where we create the enrollment).


In the case over, we determine if a user has access to a tei through the context of a program. In our usecase, the tei don't have an owner for the provided program, so we fall back to the tei ou - Which means, the tei.ou needs to be in the search scope for the user to have access.

An additional check is required in another place to make sure the Enrollment.ou is actually in the users capture scope.

Both of these are regardless of access level. The access level is only relevant for reading, or for updating existing data that has an owner.

To me, the code over seems correct to me. It's the same as before (logic) with the exception that we now check if ownership exists or not, and create this specific exception

teleivo: Thank you @stian-sandvold for your detailed explanation 😄 So statement

> If a program has no owner, use search scope instead of capture scope to allow registering a TEI to the program.

is only related to this check

> TEI.ou = A: A needs to be in the search scope (We need to be able to see the TEI)

And

> In all cases (Let me know if I am wrongly assuming here), the user needs CAPTURE scope access for a OU to create any data there, including an enrollment.

is checked somewhere else. Did I get that right?

And is the definition of a closed program one that has no owner? Or one in which you cannot enroll?

muilpp:
There's been a bit of confusion in this ticket, but in the end the result is that to determine the access to the data of a closed program, we will use the capture scope. That's already what we are currently doing, so I'm closing this PR.

PR #12294, opened 2022-11-16
ci(deps): add dependabot updates to 2.39

teleivo:
as discussed in the backend team meeting

`target-branch` does not allow an array of branches

https://github.com/dependabot/dependabot-core/issues/2511

it seems that duplicating the config on the main branch itself per `target-branch` should configure dependabot to update other branches (security fixes and regular updates).

Trying it first on 2.39 before adding it to 2.38, 2.37.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,7 +26,6 @@ updates:
     labels:
       - "dependencies"
       - "java"
-      - "run-api-tests"
     ignore:
       - dependency-name: "org.eclipse.jetty:*" # Jetty 11 moved some code to a different package, updated the Servlet version which requires more work
         versions:
@@ -58,7 +57,6 @@ updates:
     labels:
       - "dependencies"
       - "java"
-      - "run-api-tests"
     ignore:
       - dependency-name: "org.eclipse.jetty:*" # Jetty 11 moved some code to a different package, updated the Servlet version which requires more work
         versions:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,6 @@
 version: 2
 updates:
+  # master
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -31,3 +31,36 @@ updates:
...
+      timezone: "Europe/Oslo"
+    pull-request-branch-name:
+      separator: "-"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+    labels:
+      - "dependencies"
+      - "java"
+      - "run-api-tests"
```
enricocolasante, line 61: Do we need `run-api-test` label? I think we are using inverse logic in the workflow https://github.com/dhis2/dhis2-core/blob/3808769b1edc43b88fff523cf870e7f65524c6c2/.github/workflows/run-api-tests.yml#L23

I found that sometimes api-tests are not running and I need to remove that label and then put it back to have them running, not sure why

enricocolasante (commented):
Just a question about api-test label

.github/dependabot.yml
teleivo: You are right, we do not need it. Maybe this changed at some point from opt-in to opt-out 🤷🏻 

I also experienced what you describe. I am not sure if this issue has to do with https://github.com/dhis2/dhis2-core/blob/37afd8bc755c2f5ff5c3cec179c2e6edb4b9a610/.github/workflows/run-api-tests.yml#L3-L5 The event when the workflows are triggered.

teleivo: Will remove the label 👌🏻

enricocolasante (approved):

jbee (approved):

netroms (approved):

stian-sandvold (approved):

PR #12234, opened 2022-11-01
feat: add DhisControllerIntegrationTest

vietnguyen:
- Add DhisControllerIntegrationTest for supporting writing controller tests using postgres db

PR #12304, opened 2022-11-16
fix: /tracker/enrollments does not return attributes, events and relationships

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-12906
This fix missing fields returned for enrollments/, both single or multiple instances requests.
It uses an `EnrollmentParams` which is now a nested property for `tracker/trackedEntities` requests.
So for trackedEntities, if we want enrollment attributes, we need to add enrollments[attributes] in the fields, but if we still want to keep it like an input from `TrackedEntityInstanceParams` we can revert those changes.
The old tracker remains with previous behavior


dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/fieldsmapper/TrackerEnrollmentFieldsMapperTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.stream.Stream;
+
+import org.hisp.dhis.dxf2.events.TrackedEntityInstanceParams;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith( MockitoExtension.class )
```
teleivo, line 44: I don't think Mockito is needed for this test

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
```diff
@@ -126,6 +130,88 @@ public void shouldReturnRequestedFields( String endpoint, String fields, String
             } );
     }
 
+    private Stream<Arguments> shouldReturnRequestedTeiEnrollmentsFields()
+    {
+        return Stream.of( Arguments.of( "enrollments[events, relationships, attributes]", not( emptyIterable() ),
```
teleivo, line 135: Do you think we need to write an e2e test for every permutation of these `fields` query parameter values? I think this might add more cost than benefit in terms of security.

You have a unit test for these values translating into the correct object like `TrackerTrackedEntityFieldsParamMapper` that is passed into the querying code.

We have the MockMVC/h2 tests like dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerTrackedEntitiesExportControllerTest.java which also test the request is met with a response which has these fields set to something non null/empty. This one tests the code querying the DB and using the information we extract out of `fields`.

IMHO these should be enough to ensure we do return these nested fields.

https://github.com/dhis2/dhis2-core/pull/12234 we can now also write controller tests using Postgres. In case you need to test /tracker/enrollments without providing a specific `uid`. As these requests usually rely on JSONB which does not work in the convenience controller tests relying on h2 DB.

lucaCambi77: i don't see any problem with e2e tests, also I've added some of the tests you are mentioning

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerEventsExportTests.java
```diff
@@ -0,0 +1,93 @@
...
+    public void shouldReturnAscOrderedEventByTEIAttribute()
+    {
+        ApiResponse response = trackerActions.get( "events?order=dIVt4l5vIOa:asc" );
+        response
+            .validate()
+            .statusCode( 200 )
+            .body( "instances", hasSize( equalTo( 2 ) ) );
+        List<String> events = response.extractList( "instances.event.flatten()" );
+        assertEquals( List.of( "ZwwuwNp6gVd", "olfXZzSGacW" ), events, "Events are not in the correct order" );
+    }
```
enricocolasante, line 92: Should these tests be removed from TrackerTrackedEntitiesExportTests?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/EnrollmentParams.java
```diff
@@ -0,0 +1,97 @@
...
+
+    boolean includeEvents;
+
+    boolean includeRelationships;
+
+    boolean includeAttributes;
+
+    boolean includeDeleted;
+
+    boolean dataSynchronizationQuery;
```
enricocolasante, line 54: Why are these not private?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/TrackedEntityInstanceParams.java
```diff
@@ -39,26 +39,29 @@
...
+    boolean includeProgramOwners;
 
-    private final boolean includeProgramOwners;
+    boolean includeAttributes;
 
-    private final boolean includeDeleted;
+    boolean includeDeleted;
 
-    private final boolean dataSynchronizationQuery;
+    boolean dataSynchronizationQuery;
```
enricocolasante, line 64: Why are these not private?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/trackedentity/store/EnrollmentStore.java
```diff
@@ -66,4 +67,13 @@
      *         List of {@see Relationship} objects
      */
     Multimap<String, Relationship> getRelationships( List<Long> ids, AggregateContext ctx );
+
+    /**
+     * Fetches all the attributes
+     *
+     * @param ids a list of enrollment ids
+     * @return a MultiMap where key is a {@see Enrollment} uid and the key a
```
enricocolasante, line 75: The return description is not clear

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerEnrollmentsExportControllerTest.java
```diff
@@ -64,6 +78,18 @@ class TrackerEnrollmentsExportControllerTest extends DhisControllerConvenienceTe
...
+
+    private Relationship relationship;
+
+    private ProgramStage programStage;
+
+    private ProgramStageInstance programStageInstance;
+
+    private DataElement dataElement;
+
+    private static final String attributeValue = "value";
```
enricocolasante, line 91: Not sure if we have a rule for that but generally constant names are all uppercase

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fieldsmapper/TrackerEnrollmentFieldsParamMapper.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import org.hisp.dhis.dxf2.events.EnrollmentParams;
+import org.hisp.dhis.dxf2.events.TrackedEntityInstanceParams;
+import org.hisp.dhis.fieldfiltering.FieldPath;
+import org.hisp.dhis.fieldfiltering.FieldPreset;
+
+public class TrackerEnrollmentFieldsParamMapper
```
enricocolasante, line 42: Do we really need the tracker prefix here?
We are in tracker/export package, I think `EnrollmentFieldsParamMapper` is clear enough

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/fieldsmapper/TrackerTrackedEntityFieldsParamMapper.java
```diff
@@ -0,0 +1,212 @@
...
+import static org.hisp.dhis.webapi.controller.tracker.export.fieldsmapper.FieldsParamMapper.FIELD_RELATIONSHIPS;
+
+import java.util.List;
+import java.util.Map;
+
+import org.hisp.dhis.dxf2.events.TrackedEntityInstanceParams;
+import org.hisp.dhis.fieldfiltering.FieldPath;
+import org.hisp.dhis.fieldfiltering.FieldPreset;
+
+public class TrackerTrackedEntityFieldsParamMapper
```
enricocolasante, line 41: Same as vefore

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/EnrollmentParams.java
teleivo: Might be because if you do and add `@Value` that it shows sonarlint issues that these are already private final by lombok 🤔 I remember something like it. I still write them as private final as I want this to be clear not only from reading some annotation. I find the warning annoying 😋

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/TrackerEnrollmentsExportControllerTest.java
teleivo: IntelliJ also does this by default when extracting a constant in its refactoring.

teleivo (changes requested):
I don't think we should write an e2e test for all these cases. e2e tests are expensive to run, more likely to break and if they do harder to fix.

See my comment in dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerEnrollmentExportTests.java

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerEnrollmentExportTests.java
```diff
@@ -0,0 +1,110 @@
...
+            .extractImportedTeis()
+            .get( 0 );
+        enrollmentId = response.extractImportedEnrollments().get( 0 );
+        importRelationshipEnrollmentToTei( enrollmentId, teiId ).validateSuccessfulImport();
+    }
+
+    private Stream<Arguments> shouldReturnRequestedEnrollmentFields()
+    {
+        return Stream.of(
+            Arguments.of( "events, relationships, attributes", not( emptyIterable() ), not( emptyIterable() ),
```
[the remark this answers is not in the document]
lucaCambi77: as mentioned in another comment, i've already added tests in TrackerEnrollmentsExportControllerTest and also in TrackerTrackedEntitiesExportControllerTest. Still, i don't see anything wrong with e2e tests, these are very simple, but I'll remove them

lucaCambi77:
@enricocolasante @teleivo I'd like to come to an end with this PR. It has been out there for months. Still, I see a lot of comments for tests and naming but none about the main changes. One of them is about enrollments as a nested field for tei export. **If we agree with this should we do the same for events**? I'd like with PRs to focus on the logic or main changes, we can't spend months here, and PR comments are difficult to follow.

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/12491

PR #12324, opened 2022-11-17
fix: default AssignedUserSelectionMode to ANY

teleivo:
For TEI exports only if mode is PROVIDED will a subquery for the events assignedUser be made. If mode is CURRENT it will be set to PROVIDED with the assignedUsers to the current users id. Validations exist in the web layer and the service layer.

For event exports only if assignedUsers is not empty will a subquery be added. Validation that mode == PROVIDED with assignedUsers non empty is only done in the web layer.

Remove some of the defensive null checks. Replace enum equals comparison by using == which is null safe


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -438,7 +406,10 @@ public TrackedEntityInstanceQueryParams addAttributes( List<QueryItem> attrs )
 
     public boolean hasFilterForEvents()
     {
-        return hasAssignedUsers() || isIncludeOnlyAssignedEvents() || isIncludeOnlyUnassignedEvents()
+        // in the new class)
```
enricocolasante, line 409: Lost comment?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -438,7 +406,10 @@ public TrackedEntityInstanceQueryParams addAttributes( List<QueryItem> attrs )
...
+        return this.getAssignedUserQueryParam().hasAssignedUsers()
+            || AssignedUserSelectionMode.ANY == this.getAssignedUserQueryParam().getMode()
+            || AssignedUserSelectionMode.NONE == this.getAssignedUserQueryParam().getMode()
```
enricocolasante, line 412: Maybe here would be easier to understand with 

```suggestion
return AssignedUserSelectionMode.ALL != this.getAssignedUserQueryParam().getMode()
```

?

enricocolasante (commented):
2 minor comments

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
teleivo: haha, good catch 😅

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
teleivo: Good find! That's my end goal 😋 yes 😄 

I did not yet add `AssignedUserSelectionMode.ALL`. I need to make another PR for EventSearchParams. Then I'll add AssignedUserSelectionMode.ALL. I think creating separate PRs will make it easier to follow.

teleivo: If doing this in steps is ok for you @enricocolasante can you please mark this as resolved 😄

enricocolasante (approved):

PR #12233, opened 2022-11-01
fix: Return event assigned user on /trackedEntityInstances [DHIS2-13422]

muilpp:

PR #12356, opened 2022-11-17
fix: return event assigned user [DHIS2-13422]

muilpp:


enricocolasante (commented):
There are only changes to tests in the PR
Is this intentional or some commit got lost in the way?

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/aggregates/TrackedEntityInstanceAggregateTest.java
```diff
@@ -492,6 +490,15 @@ void testEventMapping()
         assertThat( event.getCompletedBy(), is( "[Unknown]" ) );
     }
 
+    private void assertUserProperties( Event event )
+    {
+        assertThat( event.getAssignedUserFirstName(), is( FIRST_NAME + TEST_USER ) );
```
teleivo, line 495: Since this test class uses hamcrest I get that you adhere to the style. I would do the same 😋 In this case what you wrote can and should be replaced with `assertEquals`. In the backend team we discussed that we will not use hamcrest just for the sake of its syntactic sugar but stick with JUnit 5 assertions (where possible).

See https://github.com/dhis2/wow-backend/blob/master/guides/testing.md#assertions

https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java is where we try to build up useful assertions mainly based on JUnit assertions. If JUnit does not provide an assertion and another library has one that is very useful we would create a wrapper in the Assertions class.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/aggregates/TrackedEntityInstanceAggregateTest.java
```diff
@@ -492,6 +490,15 @@ void testEventMapping()
         assertThat( event.getCompletedBy(), is( "[Unknown]" ) );
     }
 
+    private void assertUserProperties( Event event )
```
teleivo, line 493: Maybe call it `assertAssignedUser` as its more specific than `Properties`

teleivo (changes requested):
Some small adjustments I would make to a test 😄 

Is there any reason why you did not write the test with the changes https://github.com/dhis2/dhis2-core/pull/12233 ? This is how we would normally do it or should do it 😅

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/aggregates/TrackedEntityInstanceAggregateTest.java
muilpp: good to know, thanks!

muilpp:
@teleivo the idea was to commit this along with the fix for [12233](https://github.com/dhis2/dhis2-core/pull/12233), but it was my mistake because I opened the PR with the fix, and by the time it took me to write the test, someone else merged the PR. Lesson learned.

teleivo (approved):

muilpp:
There's been a bit of confusion in this ticket, but in the end the result is that to determine the access to the data of a closed program, we will use the capture scope. That's already what we are currently doing, so I'm closing this PR.

teleivo:
> There's been a bit of confusion in this ticket, but in the end the result is that to determine the access to the data of a closed program, we will use the capture scope. That's already what we are currently doing, so I'm closing this PR.

Did you mean this PR https://github.com/dhis2/dhis2-core/pull/12291 maybe 🤔

PR #12388, opened 2022-11-22
fix: multiOrgUnit openingDate/closingDate/name [DHIS2-13088]

tomzemp:
See [comment on DHIS2-13088](https://dhis2.atlassian.net/browse/DHIS2-13088?focusedCommentId=178868).

Some of the logic in form.js assumes that organisationUnits object will contain the organisationUnits needed when a multiOrganisationUnit form is loaded. However, ouwt.js does not seem to populate those organisationUnits until one has clicked on an organisation unit at that level. See for example this error in the console:
image

Because of errors, the data values will not necessarily populate in the multiOrgUnit forms.

I've added a fix to populate the relevant information (openingDate, closingDate, name) into the `organisationUnitList` variable (rather than relying on the standard string representation inherited from BaseNameableObject) and then use the orgUnit information from organistationUnitList rather than retrieving from organisationUnits[id].

I now do not notice any other issues when testing these multiOrgUnit forms, so hopefully the end of this ticket 😔

### DIFF
--- a/dhis-2/dhis-web/dhis-web-dataentry/src/main/webapp/dhis-web-dataentry/javascript/form.js
+++ b/dhis-2/dhis-web/dhis-web-dataentry/src/main/webapp/dhis-web-dataentry/javascript/form.js
@@ -1892,11 +1892,11 @@ function insertDataValues( json )
     	
     	$.each( organisationUnitList, function( idx, item )
         {    		
-    		orgUnitClosed = dhis2.de.validateOrgUnitOpening( organisationUnits[item.uid], period ) ;
+    		orgUnitClosed = dhis2.de.validateOrgUnitOpening( item, period ) ;
     		
     		if( orgUnitClosed )
     		{    			  	        
-    			return;
+    			return false;
     		}            
         } );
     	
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web/dhis-web-dataentry/src/main/webapp/dhis-web-dataentry/javascript/form.js
```diff
@@ -1889,15 +1889,14 @@ function insertDataValues( json )
...
-    		orgUnitClosed = dhis2.de.validateOrgUnitOpening( organisationUnits[item.uid], period ) ;
+        {
+            orgUnitClosed = dhis2.de.validateOrgUnitOpening( item, period ) ;
     		
     		if( orgUnitClosed )
-    		{    			  	        
-    			return;
-    		}            
+			{
+				return false;
```
tomzemp, line 1898: jQuery requires that you return false to break an each loop (see https://api.jquery.com/jquery.each/):
> We can break the $.each() loop at a particular iteration by making the callback function return false. Returning non-false is the same as a continue statement in a for loop; it will skip immediately to the next iteration.

(I can't seem to match the original styling whatever I do, which is why it's picking up styling changes as well)

dhis-2/dhis-web/dhis-web-dataentry/src/main/webapp/dhis-web-dataentry/javascript/form.js
```diff
@@ -1892,11 +1892,11 @@ function insertDataValues( json )
     	
     	$.each( organisationUnitList, function( idx, item )
         {    		
-    		orgUnitClosed = dhis2.de.validateOrgUnitOpening( organisationUnits[item.uid], period ) ;
+    		orgUnitClosed = dhis2.de.validateOrgUnitOpening( item, period ) ;
     		
     		if( orgUnitClosed )
     		{    			  	        
-    			return;
+    			return false;
```
tomzemp, line 1899: jQuery requires that you return false to break an each loop (see https://api.jquery.com/jquery.each/):

> We can break the $.each() loop at a particular iteration by making the callback function return false. Returning non-false is the same as a continue statement in a for loop; it will skip immediately to the next iteration

vietnguyen (approved):

vietnguyen:
Thanks for fixing those issues in the old data entry app @tomzemp

larshelge (approved):

PR #12392, opened 2022-11-22
fix: TEIs are not filterable by potentialDuplicate property [ DHIS2-11625 ]

lucaCambi77:
Adds potentialDuplicate as a request parameter. For teis we support the _filter_ only for attribute ids


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -277,6 +277,12 @@
      */
     private Date skipChangedBefore;
 
+    /**
+     * Indicates whether tracked entity instance is marked as a Potential
+     * Duplicate
+     */
+    private Boolean potentialDuplicate;
```
teleivo, line 284: I am not sure we have a convention for booleans so I might find examples for different styles 😅 

I think the name `potentialDuplicate` for a request parameter is a little confusing. I would rather name it `includePotentialDuplicate`. Since it should indicate whether we return such TEIs and not what you describe in the comment above. So the comment should definitely change (could you please include in the comment what `null` means. That is often not clear). I think the query param as well but someone else can also add their view.

Like we do for

- `includeDeleted`
- `includeAllAttributes`

see https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/new-tracker.html#request-parameters

teleivo (commented):
Does this also affect new tracker? I know the jira issue does not say so. It would be good to know/check as this fix is only for old tracker.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
lucaCambi77: mmh i see what you mean but in this case, I'd stick with potentialDuplicate because it is a tei property. 
- `inlucdeDeleted` , it defaults to false because I think we are not potentially interested in deleted tei?
- `includeAllAttributes`, still defaults to false, I see it is in use to check ownership

Potential duplicate shouldn't have a default,  we may want to filter for true or false. That is why is null by default and in that case, we get all. We can add it to the doc for sure. 
Anyways, I'll wait also for others to give an opinion

dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/teis/TEIimportTest.java
```diff
@@ -90,36 +92,73 @@ public void teisShouldBeUpdatedAndDeletedInBulk()
...
 
     }
+
+    @Test
+    public void shouldGetTeiByPotentialDuplicateParam()
+    {
+        JsonObject tei = object.getAsJsonArray( "trackedEntityInstances" ).get( 0 ).getAsJsonObject();
+
+        updatePotentialDuplicateAndFindTei( tei, true );
+        updatePotentialDuplicateAndFindTei( tei, false );
```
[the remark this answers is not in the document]
lucaCambi77: yeah, so I am not sure why I did that test :), probably to cover different combinations as we don't have a starting default. I agree, now i see the best way is to start from e.g. 2 teis with different status and test cases for true, false and null. And also integration tests would be easier

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
enricocolasante: I think here `potentialDuplicate` is more like `followUp` parameter.
The naming follow the pattern of the other parameters, so I don't think it is a problem. I would improve the comment though explain exactly what it happens if it is TRUE, FALSE and null.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -734,6 +740,15 @@ public boolean hasUser()
         return user != null;
     }
 
+    /**
+     * Indicates whether this parameters specifies if tei is a potential
+     * duplicate. It can be true or false.
```
enricocolasante, line 745: It can be true or false is implicit in the return value and it can be removed from the comment.
What is confusing me here is the `has` vs `is`. The name of the method is `hasPotentialDuplicate` while in the comment it says that tei `is a potential duplicate`.
I think here this method is not telling you anything about the tei but it is telling if the parameters specify a potential duplicate filter.

teleivo: > I think here this method is not telling you anything about the tei but it is telling if the parameters specify a potential duplicate filter.

Isn't that the same point about the query parameter `potentialDuplicate` above? To me the query parameter does not convey a clear meaning.

lucaCambi77: `hasPotentialDuplicate` is about the parameter in the class. So I'd leave the method name like that and explain it in the doc correctly as you are pointing out it is not correct

teleivo: IMHO the query parameter and all related fields should be called `includePotentialDuplicates`. To me this clearly says what it will do all the way down to the DB. Not much docs needed then 😋

enricocolasante: I think the include word is here used with a different meaning.
We have `includeDeleted` that if false by default and if true it will include all the deleted teis.
Here the `potentialDuplicate` is used to filter the result, and it is not a binary value, it can also be null.
I am not sure what it the best name here but if we put `includePotentialDuplicates` I would expect that if it is false, I would get all the teis but the ones that have potential duplicate equals true and if it is true I would get all the teis, also the ones with potential duplicate equals false

teleivo: Mhm, ok got it 👌🏻 `includePotentialDuplicates` is not good then.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
stian-sandvold: By default, we return teis regardless of their "potentialDuplicate" value. That means, true or false is not the only options as Luca mentions, but also null (ignore/dont filter by this property).

I agree, @lucaCambi77 please improve the comment, stating each case of true/false/null and the intended result; That will make it clear why we are using a Boolean, and how null is significant.

I think the naming is fine; And if null was not valid in this case, I agree it should be boolean :)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
stian-sandvold: I think the name and comment is a bit confusing, although I understand the intention in the context of the whole file.

I would suggest improving the naming and comments as follows:
Name: "hasPotentialDuplicateFilter" or similar. Currently, the name confuses me with the actual property of the tei.

Comment: Describe it more like "Checks if the Parameters include potentialDuplicate as a filter. If the property is null, it implies we do not care whether a tei is a potentialDuplicate or not."

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/webrequest/TrackedEntityInstanceCriteria.java
```diff
@@ -186,6 +186,12 @@ public class TrackedEntityInstanceCriteria extends PagingAndSortingCriteriaAdapt
      */
     private boolean includeAllAttributes;
 
+    /**
+     * Indicates whether the Tracked Entity Instance is marked as a Potential
+     * Duplicate.
+     */
```
stian-sandvold, line 192: I think the comment here is misleading? This is about a filter parameter, not about the flag of the tei itself - correct?

stian-sandvold (changes requested):

stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/webrequest/TrackedEntityInstanceCriteria.java
lucaCambi77: ok i've updated methods javadoc

PR #12395, opened 2022-11-22
fix: system info field from json serialization

mortenoh:


Philip-Larsen-Donnelly (approved):
Approved by for 2.38.2 by RCB (regression):
Phil, Lars, Morten S

PR #12403, opened 2022-11-24
fix: update metadata without createdBy field

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14191

### Issue:
- When updating metadata, if createdBy property is not provided in payload then the createdBy value of existing object will be changed to currentUser.
This applies to all metadata objects.

### Fix:
- Add logic to `IdentifiableObjectBundleHook` to make sure createdBy field is immutable when updating.
- Only update the createdBy field when the existing object has createdBy=null ( which should not happen, and could be a data issue ) 

### Test case
1. Create a new DataElement, or use existing one. Prepare a payload of this object for updating. Note down the createdBy user UID.
2. In prepared payload, delete the createdBy field.
3. Send a PUT request with prepared payload to api/dataElements/{uid} to update dataelement without createdBy field.
4. Check the createdBy field of updated DataElement, it should be the same as in step 1.


mortenoh (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
```diff
@@ -127,6 +126,29 @@ private void handleAttributeValues( IdentifiableObject identifiableObject, Objec
...
+     * @param bundle The {@link ObjectBundle} of the importing service.
+     */
+    private void handleCreatedByProperty( IdentifiableObject object, IdentifiableObject persistedObject,
+        ObjectBundle bundle )
+    {
+        if ( persistedObject.getCreatedBy() == null )
+        {
+            object.setCreatedBy( object.getCreatedBy() != null ? object.getCreatedBy() : bundle.getUser() );
+        }
+        else if ( object.getCreatedBy() == null
```
larshelge, line 144: Just an idea, would it make sense for us to have a method `IdentifiableObjectUtils.equals( IdentifiableObject, IdentifiableObject )` with unit tests where the null logic is encapsulated? This will be similar to Java `Objects.equals`. Then we would not have to repeat all these null checks in application code/more compact code, and less risk of mistakes/nullpointer exceptions.

larshelge (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
vietnguyen: fixed

PR #12404, opened 2022-11-24
ci(deps): ignore Spring(security) framework 6 updates

teleivo:
as Spring 6 requires minimum JDK 17 so it will require more effort/time until we update it.

For Spring and Spring security both need to be on the same major version
but are released independently.

Other libraries released by Spring follow a different release schedule
where the major version number differs.


teleivo:
closing see https://github.com/dhis2/dhis2-core/pull/12404

enricocolasante (approved):

larshelge (approved):

PR #12405, opened 2022-11-24
fix: Accept lower case enum as request parameter

enricocolasante:
Following the suggestion of the backend meeting I was looking for a way to ignore case in enum parameters in the request.
The solution to load all the enum classes in the `CrudControllerAdvice` and for all of them register a custom editor that will convert the param string representing an enum to upper case.

It is still a draft because I still need to find a good controller to create a specific h2 test for this


dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/events/EventExportTests.java
```diff
@@ -141,9 +141,9 @@ public void beforeEach()
     Stream<Arguments> shouldUseCorrectScopeWhenOuIsProvided()
     {
         return Stream.of(
-            Arguments.of( "OU: root", "SELECTED", rootOu, false, null ),
-            Arguments.of( "OU: capture", "SELECTED", captureOu, true, Arrays.asList( captureOu ) ),
-            Arguments.of( "OU: search", "SELECTED", searchOu, false, null ),
+            Arguments.of( "OU: root", "selected", rootOu, false, null ),
```
enricocolasante, line 144: That was really just to test. I am looking for a better place in web-api-test to add a relevant test for it

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -116,11 +119,20 @@
...
+        this.enumKlasses = new ClassGraph().acceptPackages( "org.hisp.dhis" )
+            .enableClassInfo().scan().getAllClasses().getEnums().loadClasses();
+    }
+
     @InitBinder
     protected void initBinder( WebDataBinder binder )
     {
         binder.registerCustomEditor( Date.class, new FromTextPropertyEditor( DateUtils::parseDate ) );
         binder.registerCustomEditor( IdentifiableProperty.class, new FromTextPropertyEditor( String::toUpperCase ) );
+        this.enumKlasses.forEach( c -> binder.registerCustomEditor( c, new ConvertEnum( c ) ) );
```
jbee, line 135: No need for `K` in `Klasses` if the word is not exactly and only `Class` ;)

jbee (approved):
Looks good. 
You can test this with any endpoint with a `filter` and `rootJunction=or` - There are some existing tests for Gist API where I recently had to change the test to use `OR` instead of `or` because I started using the enum field instead of `String`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
enricocolasante: True

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -467,4 +479,23 @@ public void setAsText( String text )
             setValue( fromText.apply( text ) );
         }
     }
+
+    private static final class ConvertEnum<T extends Enum<T>> extends PropertyEditorSupport
+    {
+        private Class<T> enumClass;
```
teleivo, line 485: could/should be final?

teleivo (commented):
I think I spotted a mistake in a test. Added some small comments and questions :smi

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -116,11 +119,20 @@
 
     private static final String GENERIC_ERROR_MESSAGE = "An unexpected error has occured. Please contact your system administrator";
 
+    private final List<Class<?>> enumClasses;
+
+    public CrudControllerAdvice()
+    {
+        this.enumClasses = new ClassGraph().acceptPackages( "org.hisp.dhis" )
```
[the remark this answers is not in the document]
enricocolasante: The enum will not be registered as a custom editor and lower case parameters will be invalid

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/EnumBindingTest.java
```diff
@@ -0,0 +1,123 @@
...
+class EnumBindingTest
+{
+    private final static String ENDPOINT = "/enum";
+
+    private MockMvc mockMvc;
+
+    @BeforeEach
+    public void setUp()
+    {
+        EnumController controller = new EnumController();
```
teleivo, line 56: you could inline this one

teleivo (approved):
This is beautiful 🎉 Nicely tested!!!

Only if `criteria` is not a generally understood term in our code base (not sure if its only tracker specific 🤷🏻) should we change it for something else like maybe `QueryParamClass` or maybe Spring has a term for it already.

enricocolasante:
> This is beautiful tada Nicely tested!!!
> 
> Only if `criteria` is not a generally understood term in our code base (not sure if its only tracker specific 🤷🏻) should we change it for something else like maybe `QueryParamClass` or maybe Spring has a term for it already.

Don't think we have an agreement on that in the code base and Spring doesn't seem to have a term as well. I would leave as it is and maybe we should try to stick with the criteria term all over the place

PR #12418, opened 2022-11-25
fix: broken password generator in invite user

netroms:


vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/DefaultSecurityService.java
```diff
@@ -252,7 +252,7 @@ public void prepareUserForInvite( User user )
             user.setUsername( username );
         }
 
-        String rawPassword = CodeGenerator.getRandomSecureToken( INVITED_USER_PASSWORD_LENGTH_BYTES );
+        String rawPassword = "Inv!t3#" + CodeGenerator.getRandomSecureToken( INVITED_USER_PASSWORD_LENGTH_BYTES );
```
enricocolasante, line 255: Is it too long now? Should it be something like this?

```suggestion
        String rawPassword = "Inv!t3#" + CodeGenerator.getRandomSecureToken( INVITED_USER_PASSWORD_LENGTH_BYTES - 7);
```

enricocolasante (commented):
Test is still failing, probably because now the password is too long

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/DefaultSecurityService.java
netroms: Yeah, it was too long. Thanks Enrico.

enricocolasante (approved):

PR #12419, opened 2022-11-25
feat: support EventHook pub/sub

mortenoh:
## Summary

* [x] Implement `EventHook` type
* [x] Implement publisher/subscriber services
* [x] Implement metadata CRUD event publishing
* [x] Implement `webhook` target
* [x] Implement `console` target
* [x] Implement `jms` target
* [x] Implement `kafka` target
* [x] Custom `TaskExecutor` for `EventHooks`
* [x] Implement event publishing for Scheduler

[DHIS2-12194](https://dhis2.atlassian.net/browse/DHIS2-12194) | [discussions/337](https://github.com/dhis2/notes/discussions/337)

[DHIS2-12194]: https://dhis2.atlassian.net/browse/DHIS2-12194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/ConsoleTarget.java
```diff
@@ -0,0 +1,56 @@
...
+ * @author Morten Olav Hansen
+ */
+@Getter
+@Setter
+@EqualsAndHashCode( callSuper = true )
+@Accessors( chain = true )
+public class ConsoleTarget extends Target
+{
+    @JsonProperty( required = true )
+    private String clientId = "dhis2-" + CodeGenerator.generateUid();
```
larshelge, line 50: Perhaps move the initialization of this property to the constructor, and mark it as final? This to make it clear that this value should be initialized exactly once (assuming that is the case).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/auth/HttpBasicAuth.java
```diff
@@ -0,0 +1,73 @@
...
+    @JsonProperty( required = true )
+    private String password;
+
+    public HttpBasicAuth()
+    {
+        super( "http-basic" );
+    }
+
+    @Override
+    public void apply( MultiValueMap<String, String> headers )
```
larshelge, line 63: Perhaps use `HttpHeaders` over `MultiValueMap`, since the former extends the latter, and makes the code a bit more specific and descriptive?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/auth/ApiTokenAuth.java
```diff
@@ -0,0 +1,71 @@
...
+    @JsonProperty( required = true )
+    private String token;
+
+    public ApiTokenAuth()
+    {
+        super( "api-token" );
+    }
+
+    @Override
+    public void apply( MultiValueMap<String, String> headers )
```
larshelge, line 62: Very minor, I am wondering if `apply` is a little bit misleading here, since we are not applying the headers to the authorization class, it is the authorization class which is applied to the headers.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookListener.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Async( "eventHookTaskExecutor" )
+    @TransactionalEventListener( classes = Event.class, phase = TransactionPhase.AFTER_COMMIT, fallbackExecution = true )
+    public void eventListener( Event event )
+        throws EventHookException,
+        JsonProcessingException
+    {
+        for ( EventHook eventHook : eventHooks )
+        {
+            if ( eventHook.getSource().getPath().startsWith( eventHook.getSource().getPath() ) )
```
larshelge, line 86: Might be an oversight here, comparing the same value.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookListener.java
```diff
@@ -0,0 +1,169 @@
...
+            {
+                if ( !targets.containsKey( eventHook.getUid() ) || targets.get( eventHook.getUid() ).isEmpty() )
+                {
+                    continue;
+                }
+
+                if ( event.getObject() instanceof Collection )
+                {
+                    Collection<ObjectNode> objects = new ArrayList<>();
```
larshelge, line 95: Use List interface to get more specific methods.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/ConsoleTarget.java
mortenoh: This is the default value which they would hopefully override. I'll add some javadocs (I know it's missing many places, it's coming)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/auth/HttpBasicAuth.java
mortenoh: I didn't really want to add `spring-web` to the `dhis-api` module..

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/auth/ApiTokenAuth.java
mortenoh: My thinking we are applying the state of the `ApiTokenAuth` to the `MultiValueMap<>`, I'm open to name suggestions?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/EventUtils.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        return metadata( object, "update" );
+    }
+
+    public static Event metadataDelete( Class<?> type, String uid )
+    {
+        String name = camelCase( type.getSimpleName() );
+
+        return Event.builder()
+            .path( "metadata." + name + "." + uid )
```
larshelge, line 58: Use `String.format` ?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/eventhook/EventHookPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Morten Olav Hansen
+ */
+@Component
+@RequiredArgsConstructor
+public class EventHookPublisher
+{
+    private static final List<?> BLACKLIST = List.of(
```
larshelge, line 47: I guess the term these days is `BLOCKLIST`.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookException.java
```diff
@@ -0,0 +1,39 @@
...
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.eventhook;
+
+/**
+ * @author Morten Olav Hansen
+ */
+public class EventHookException extends Exception
```
larshelge, line 33: Better to extend `ErrorCodeException` and use error codes from the start?

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookListener.java
```diff
@@ -0,0 +1,161 @@
...
+            if ( !eh.isEnabled() )
+            {
+                continue;
+            }
+
+            targets.put( eh.getUid(), new ArrayList<>() );
+
+            for ( Target target : eh.getTargets() )
+            {
+                if ( "webhook".equals( target.getType() ) )
```
larshelge, line 137: Use an enum for the various target types? Would be nice to have a central place to store target types I would think.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_5__add_evenhook_table.sql
```diff
@@ -0,0 +1,32 @@
+
+-- Add eventhook table
+
+drop table if exists eventhook;
+
+create table eventhook (
```
larshelge, line 6: Use "create table if not exists" for (at least part) idempotence.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookException.java
mortenoh: This exception class is not in use and will be removed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/eventhook/EventHookPublisher.java
mortenoh: Sure, we do use `blacklist` in multiple other places though. I have changed it now.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_5__add_evenhook_table.sql
mortenoh: Sure, was just following the other flyway scripts, updated now.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookListener.java
mortenoh: Switch to using constants for json sub-type types.

amcgee:
@mortenoh as we discussed in Vietnam, we should do some performance stress tests on this.  What's the plan for that?  cc @Philip-Larsen-Donnelly

amcgee:
>  Implement event publishing for Scheduler

Am I correct in assuming that this will not (as currently implemented) trigger event hooks if i.e. the analytics table export is run manually via a call to the API (POST /api/resourceTables/analytics)?

mortenoh:
> > Implement event publishing for Scheduler
> 
> Am I correct in assuming that this will not (as currently implemented) trigger event hooks if i.e. the analytics table export is run manually via a call to the API (POST /api/resourceTables/analytics)?

They will trigger in all cases

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookService.java
```diff
@@ -0,0 +1,79 @@
...
+    public List<EventHook> getAll()
+    {
+        List<EventHook> eventHooks = new ArrayList<>();
+
+        for ( EventHook eventHook : eventHookStore.getAll() )
+        {
+            try
+            {
+                // TODO: this is a bit of a hack, but we need to decrypt the password and not change the object in the database
+                EventHook eh = objectMapper.readValue( objectMapper.writeValueAsString( eventHook ), EventHook.class );
```
jbee, line 67: Better safe then sorry, but it is declared as only a read-only transaction - but with hibernate you never know...

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/Handler.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @author Morten Olav Hansen
+ */
+@FunctionalInterface
+public interface Handler
+{
+    void run( EventHook eventHook, Event event, String payload );
+
+    default void close()
```
jbee, line 38: I think you can the `Handler` let extend `AutoClosable` and then override `close` here without the `throws Exception` which should be the same as now just that it can be used in advanced try-catch blocks.

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/descriptors/EventHookSchemaDescriptor.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public Schema getSchema()
+    {
+        Schema schema = new Schema( EventHook.class, SINGULAR, PLURAL );
+        schema.setRelativeApiEndpoint( API_ENDPOINT );
+        schema.setOrder( 1900 );
+
+        schema.add( new Authority( AuthorityType.CREATE_PUBLIC,
+            Lists.newArrayList( "F_EVENT_HOOK_PUBLIC_ADD" ) ) );
```
jbee, line 60: Please use `List.of`

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/EventHookControllerTest.java
```diff
@@ -0,0 +1,156 @@
...
+class EventHookControllerTest extends DhisControllerIntegrationTest
+{
+    @Test
+    void testGetEventHooks()
+    {
+        JsonObject eventHooks = GET( "/eventHooks" ).content( HttpStatus.OK );
+        assertTrue( eventHooks.isObject() );
+        assertTrue( eventHooks.has( "eventHooks" ) );
+        assertTrue( eventHooks.get( "eventHooks" ).isArray() );
+        assertTrue( eventHooks.get( "eventHooks" ).asList( JsonObject.class ).isEmpty() );
```
jbee, line 59: For a empty check you can just do `getArray( "eventHooks" ).isEmpty()`.

You do not need the `assertTrue( eventHooks.has( "eventHooks" ) );` because `eventHooks.get( "eventHooks" ).isArray()` implies that.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/EventHookControllerTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertTrue( eventHook.has( "id", "name", "source", "targets" ) );
+        assertEquals( "bRNvL6NMQXb", eventHook.getString( "id" ).string() );
+        assertEquals( "WebhookApiToken", eventHook.getString( "name" ).string() );
+        assertEquals( "metadata", eventHook.get( "source" ).asObject().getString( "path" ).string() );
+        assertEquals( "id,name", eventHook.get( "source" ).asObject().getString( "fields" ).string() );
+
+        JsonList<JsonObject> targets = eventHook.get( "targets" ).asList( JsonObject.class );
+        assertFalse( targets.isEmpty() );
+
+        JsonObject target = targets.get( 0 ).asObject();
```
jbee, line 79: `getObject(0)` instead of `get( 0 ).asObject()` is less "casty". There are a couple more of similar cases where `getObject` could be used instead.

jbee (commented):
A dozen small things. Mostly naming. Some cases where a more suitable method exists.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/ReloadEventHookListener.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.hisp.dhis.eventhook;
+
+/**
+ * Event that reloads the internals of the event listener. Should be published
+ * when EventHooks are created, updated, deleted.
+ *
+ * @author Morten Olav Hansen
+ */
+public class ReloadEventHookListener
```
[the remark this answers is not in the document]
jbee: I later understood the listener name. But suggest to rename that and then this accordingly with the `Reload` prefix.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/EventHookService.java
mortenoh: This 100% changed the object every time.. I don't really want to create another wrapper around it, but I also feels its not a great way of doing it (could add another DTO layer.. but its seems kinda useless in this case)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/EventUtils.java
```diff
@@ -0,0 +1,110 @@
...
+import org.hisp.dhis.scheduling.JobStatus;
+
+import com.google.common.base.CaseFormat;
+
+/**
+ * @author Morten Olav Hansen
+ */
+public final class EventUtils
+{
+    public static Event metadataCreate( BaseIdentifiableObject object )
```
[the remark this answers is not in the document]
mortenoh: Generaly we end it with `Utils` not too strongly for or against..

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/eventhook/EventHookPublisher.java
```diff
@@ -0,0 +1,67 @@
...
+import org.hisp.dhis.user.User;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Morten Olav Hansen
+ */
+@Component
+@RequiredArgsConstructor
+public class EventHookPublisher
```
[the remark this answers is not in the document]
jbee: Are you creating an interface for this?

jbee (commented):
I think the publisher should have an interface.

As I say `EventHookListener` is confusing. Could we use some other suffix than _Listener_ that more focuses on the this being the active component that dispatches the events to the targets?

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventhook/targets/auth/ApiTokenAuth.java
larshelge: Fair enough.

larshelge (approved):

PR #12430, opened 2022-11-29
fix: safely set AssignedUserSelectionMode EventSearchParams DHIS2-13765

teleivo:
follow up after https://github.com/dhis2/dhis2-core/pull/12324


teleivo:
closing in favor of https://github.com/dhis2/dhis2-core/pull/12446

PR #12431, opened 2022-11-29
fix: Fetch nested event notes in trackedEntityInstance endpoint [DHIS2-11731]

zubaira:


enricocolasante (changes requested):
A bug fix like this should have a test to clearly show what was fixed and what is the current behaviour

zubaira:
> A bug fix like this should have a test to clearly show what was fixed and what is the current behaviour

resolved

enricocolasante (approved):

stian-sandvold (approved):

PR #12432, opened 2022-11-29
chore: add default AssignedUserSelectionMode.ALL

teleivo:
follow up after https://github.com/dhis2/dhis2-core/pull/12430

Add default AssignedUserSelectionMode.ALL in place of null indicating that all events should be returned irrespective of whether a user is assigned or not. This makes using the params safer and is analogous to OrganisationUnitSelectionMode which already has mode ALL.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AssignedUserQueryParam.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AssignedUserQueryParam.java
@@ -42,7 +42,8 @@ import org.hisp.dhis.user.User;
 public class AssignedUserQueryParam
 {
 
-    public static final AssignedUserQueryParam ALL = new AssignedUserQueryParam( null, null, null );
+    public static final AssignedUserQueryParam ALL = new AssignedUserQueryParam( AssignedUserSelectionMode.ALL, null,
+        Collections.emptySet() );
 
     private final AssignedUserSelectionMode mode;
 
@@ -87,6 +88,11 @@ public class AssignedUserQueryParam
         {
             this.mode = AssignedUserSelectionMode.PROVIDED;
             this.assignedUsers = assignedUsers;
+        }
+        else if ( mode == null )
+        {
+            this.mode = AssignedUserSelectionMode.ALL;
+            this.assignedUsers = Collections.emptySet();
         }
         else
         {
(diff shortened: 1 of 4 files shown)


muilpp (approved):

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AssignedUserQueryParam.java
```diff
@@ -88,6 +89,11 @@ else if ( (mode == null || mode == AssignedUserSelectionMode.PROVIDED)
             this.mode = AssignedUserSelectionMode.PROVIDED;
             this.assignedUsers = assignedUsers;
         }
+        else if ( mode == null )
```
stian-sandvold, line 92: Looks correct, but it occurs to me this whole block might be ripe for a facelift. But I think that is out of the scope of this review :)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -403,9 +403,7 @@ public TrackedEntityInstanceQueryParams addAttributes( List<QueryItem> attrs )
 
     public boolean hasFilterForEvents()
     {
-        return this.getAssignedUserQueryParam().hasAssignedUsers()
-            || AssignedUserSelectionMode.ANY == this.getAssignedUserQueryParam().getMode()
-            || AssignedUserSelectionMode.NONE == this.getAssignedUserQueryParam().getMode()
+        return this.getAssignedUserQueryParam().getMode() != AssignedUserSelectionMode.ALL
```
stian-sandvold, line 406: This seems correct to me. It occurs to me the previous solution would be open to miss data; I think you might have solved a potential bug as well :)

stian-sandvold (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AssignedUserQueryParam.java
teleivo: Yes 😅 there might be a cleaner solution. Feel free to give it a try after I merged the PR. I added ample tests. So refactoring should be safe and fun 🎉

PR #12447, opened 2022-11-30
feat: Support unlogged tables in analytics export [DHIS2-14208]

maikelarabori:
This set of changes is adding support for a new `dhis.conf` property named `analytics.table.unlogged`.

When this property is set to `on`, the analytics table export process will create tables using the `unlogged` flag. This will make the table export much faster (about ~40% faster on my local desktop). The SQL will look like this:
```
create unlogged table ...
```

The drawbacks of this approach are that tables generated in this mode cannot be replicated, and if Postgres crashes or is shut down abruptly all `unlogged` tables are truncated. So this flag is not recommended if clients need replication or don't want to have analytics tables eventually truncated.

If the flag is not present or is set to `off`, the tables are generated as usual. ie.:
```
create table ...
```

As most clients do not use or need analytics tables replication, this new property might give them a good performance boost.

_I also applied some cosmetic refactoring, making use of StringBuilder instead of String concatenations._


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/CategoryOptionComboNameResourceTable.java
```diff
@@ -63,10 +66,13 @@ public ResourceTableType getTableType()
     @Override
     public String getCreateTempTableStatement()
     {
-        return "create table " + getTempTableName() +
-            " (categoryoptioncomboid bigint not null primary key, " +
-            "categoryoptioncomboname varchar(255), approvallevel integer, " +
-            "startdate date, enddate date)";
+        StringBuilder sql = new StringBuilder();
```
larshelge, line 69: I am not quite sure if this makes the code more readable. I would actually prefer the way it was before. Internally, the JVM will optimize string memory usage so there is not much to gain from a memory perspective either.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/analytics/AnalyticsExportSettings.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Returns the respective string that represents the table type to be
+     * exported. Two types are supported: UNLOGGED and DEFAULT. See
+     * {@link AnalyticsTableType}
+     *
+     * @return the string representation of {@link AnalyticsTableType}.
+     */
+    public String getTableType()
+    {
+        if ( cachedTableType != null )
```
larshelge, line 81: I don't think we need the cached table type parameter. The config values in DhisConfigurationProvider are read on startup and kept in memory, so lookups are quite cheap I think?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/analytics/AnalyticsExportSettings.java
```diff
@@ -0,0 +1,97 @@
...
+public class AnalyticsExportSettings
+{
+    private final DhisConfigurationProvider dhisConfigurationProvider;
+
+    private String cachedTableType;
+
+    public enum AnalyticsTableType
+    {
+        DEFAULT( EMPTY ),
+        UNLOGGED( "unlogged" );
```
larshelge, line 57: Since the underlying config property is "unlogged" on or off, I am wondering if an enum is helpful. I.e. we cannot add more options to the "unlogged" property - it is on or off. If we add another config property later, there will be a lot of combinations to support across these options.

So I think either remove this class entirely, or change the config property to support named options, like `default`, `unlogged`, which allows to be extended later.

maikelarabori: Done

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/analytics/AnalyticsExportSettings.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @author maikel arabori
+ */
+@Component
+@RequiredArgsConstructor
+public class AnalyticsExportSettings
+{
+    private final DhisConfigurationProvider dhisConfigurationProvider;
+
+    private String cachedTableType;
```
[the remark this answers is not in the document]
maikelarabori: Done

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/table/CategoryOptionComboNameResourceTable.java
maikelarabori: Done

larshelge (approved):

PR #12448, opened 2022-11-30
fix: Remove formatting for line comments

enricocolasante:
This is removing the automatic formatting for line comments.
Now inline comments after some code can be as long as we want, they will not be broken in multiple lines.
The existing multiline line comments can be messed up by the plugin though


dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/Constants.java
```diff
@@ -59,13 +59,13 @@
         .setProgramStages( Arrays.asList( "PaOOjwLVW23", "nlXNK4b7LVr" ) );
 
     public static String TRACKER_PROGRAM_ID = "f1AyMswryyQ"; // todo: remove and
-                                                             // use
```
teleivo, line 62: Its nice that it joins comments like the one above for `13 characters` https://github.com/dhis2/dhis2-core/pull/12448/files#diff-06993b172f593b50d3502e60bbc91f28ba566e5f1a11cc4296fff1b28276aea5R72

This change here is weird as it looks like unaligns the comments. I assume that the comment is formatted like this is due to the rule you remove, right? So it should improve new comments 😄

enricocolasante: The `13 characters` inline is something I manually fixed just to check that the formatter was not messing that up. Sorry I mixed up the apply from the plugin and some manual fixes, the plugin never inlines comments that are on multiple lines
The formatter now is messing line comments that are broken on more lines, the idea would be that a line comment should be on one line or if it is on multiple lines there is no code on those lines.

jbee (approved):

dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/Constants.java
teleivo: Got it ☺️

stian-sandvold (approved):
Fixed the really bad indentations; Good job!

Was a lot of issues fixing that last file, I think the formatter was playing me, only letting me keep 1 change each commit. 🤔

PR #12457, opened 2022-12-01
fix: turn removeOnError into skipOnError DHIS2-14213

teleivo:
## Why

https://dhis2.atlassian.net/browse/DHIS2-14213

## How

Before we iterated over hooks (in [DefaultTrackerValidationService](https://github.com/dhis2/dhis2-core/blob/1c8287b0aa9334c31547c0f9685a7c1de3cb601b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationService.java#L89-L113)) each iterating over entities (in [AbstractTrackerDtoValidationHook](https://github.com/dhis2/dhis2-core/blob/f21a72c4d48acd7a80d31ed47e9def27ce785779/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/AbstractTrackerDtoValidationHook.java#L159-L191)) and removing an invalid one from the bundle.

Now we iterate over entities then hooks and skip subsequent hooks if the hook deems the entity invalid and has skipOnError set to true. All in a single place `DefaultTrackerValidationService`.

This is an intermediate step in a series of refactorings. The TrackerValidationHook interface will not only contain default methods.

The following changes to the tests highlight that we did not find all possible errors before this change 🎉 

## Tests

Some tests had to be adapted or added. This explains why some tests needed to be changed.

### ReportSummaryIntegrationTest.testStatsCountForOneUpdateEnrollmentAndOneCreatedEnrollmentAndOneInvalidEnrollment

Enrollment with UID `INVALIDUID-TOOLONG` was previously removed from the bundle via PreCheckUidValidationHook as it had `removeOnError=true`. Both enrollments with UID `INVALIDUID-TOOLONG` and `TvctPPhpD8u` have program `BFcipDERJnf` which only allows one active/completed enrollment. This means the test setup was incorrect and did not fit the test description/intention.

So before the change we got

* TrackerErrorReport{message=Object: `INVALIDUID-TOOLONG`, uid: `INVALIDUID-TOOLONG`, has an invalid uid format., errorCode=E1048, trackerEntityType=ENROLLMENT, uid=INVALIDUID-TOOLONG}

while after

* TrackerErrorReport{message=Object: `INVALIDUID-TOOLONG`, uid: `INVALIDUID-TOOLONG`, has an invalid uid format., errorCode=E1048, trackerEntityType=ENROLLMENT, uid=INVALIDUID-TOOLONG}
* TrackerErrorReport{message=TrackedEntityInstance: `CREATEDTEI1`, already has an active Enrollment in Program `BFcipDERJnf`., errorCode=E1015, trackerEntityType=ENROLLMENT, uid=NENROLLMENT}

Fix: create 3 TEIs each is enrolled so we can create an enrollment, update one and ignore one since its UID is invalid.

### AtomicModeTests

Both e2e tests use the same invalid payload to test `AtomicMode.ALL` or `OBJECT`.

https://github.com/dhis2/dhis2-core/blob/50b4ae59f67746ec9d298025f01a75fc053ac321/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/AtomicModeTests.java#L93-L100

Removing the `trackedEntityType` triggers
[PreCheckMandatoryFieldsValidationHook](https://github.com/dhis2/dhis2-core/blob/50b4ae59f67746ec9d298025f01a75fc053ac321/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/hooks/PreCheckMandatoryFieldsValidationHook.java#L59) which previously had `removeOnError=true`. Now that its not removed from the bundle more validations will run.

before we got

```json
"errorReports": [
  {
      "message": "Missing required tracked entity property: `trackedEntityType`.",
          "errorCode": "E1121",
          "trackerType": "TRACKED_ENTITY",
          "uid": "AAR6N7hYSPC"
  },
  {
      "message": "Relationship: `dEsKfzHDQJz` cannot be persisted because trackedEntity AAR6N7hYSPC referenced by this relationship is not valid.",
          "errorCode": "E4011",
          "trackerType": "RELATIONSHIP",
          "uid": "dEsKfzHDQJz"
  }
]
```

after

```json
"errorReports": [
   {
       "message": "Missing required tracked entity property: `trackedEntityType`.",
       "errorCode": "E1121",
       "trackerType": "TRACKED_ENTITY",
       "uid": "hyt391ByqW0"
   },
   {
       "message": "Relationship Type `from` constraint requires a Tracked Entity having type `Q9GufDoplCL` but `null` was found.",
       "errorCode": "E4014",
       "trackerType": "RELATIONSHIP",
       "uid": "R6PK2CUAqA4"
   },
   {
       "message": "Relationship: `R6PK2CUAqA4` cannot be persisted because trackedEntity hyt391ByqW0 referenced by this relationship is not valid.",
       "errorCode": "E4011",
       "trackerType": "RELATIONSHIP",
       "uid": "R6PK2CUAqA4"
   }
]
```

Error `E4014` "Relationship Type `from` constraint requires a Tracked Entity having type `Q9GufDoplCL` but `null` was found." being new.

### EventImportValidationTest.testEnrollmentAndEventFailIndependently

Added test showing that due to `removeOnError` we invalidated a valid event because of its invalid enrollment which was removed from the payload.


stian-sandvold (approved):
Looks good! Good comments and PR description made it easy to review! 👍

enricocolasante (approved):

PR #12463, opened 2022-12-02
chore: default AssignedUserSelectionMode to ALL

teleivo:
in controllers. Remove private final which is already added by lombok


teleivo:
This cannot be done as easily since not passing the `mode` is ok even though you provided `assignedUsers` as we then set the mode to `PROVIDED`. So setting `mode` to `ALL` by default will result in mode `ALL` and non-empty `assignedUsers` in this case which is invalid.

PR #12467, opened 2022-12-02
chore: remove unused duplicate reporting logic DHIS2-14213

teleivo:
Previously, I thought `TrackerValidationReport` and `ValidationErrorReporter` should be merged. I think I was wrong 😁 They have different responsibilities.

`TrackerValidationReport` is for representational purposes only. It aggregates different reports of its kind and presents them as JSON to the user as part of the `TrackerImportReport`.

`ValidationErrorReporter`s responsibility is to collect errors/warnings that occur during the validation and keep track of which `TrackerDto`s are invalid.

The confusion might have come from the proximity of `TrackerValidationReport` and `ValidationErrorReporter` (both in `report` package).

* moved `ValidationErrorReporter` and `MessageFormatter` into `validation` package as this is where they are mostly used. That mostly imports in the `validation` package need to change shows that.
* added a package javadoc for `report` to explain that its classes are to create a `TrackerImportReport`.

## Future changes

* I think we need to find a better name to distinguish `TrackerValidationReport` and `ValidationErrorReporter`.
* I think we need to relocate some other classes to get the dependencies straight. I think `validation` should not have to depend on `report`. `report` should depend on `validate`.


teleivo:
closing in favor of https://github.com/dhis2/dhis2-core/pull/12486

PR #12473, opened 2022-12-04
refactor: clean ValidationErrorReporter DHIS2-14213

teleivo:
* fix linting issues (redundant access modifiers added by lombok, missing javadoc)
* collect invalid dtos in set. As we do not need to keep track of how many errors a dto has.
* tracker invalid dtos in enummap. As its more efficient since amount of keys is known beforehand
* don't allow direct access to invalid dtos. It's an internal structure the reporter keeps track of. Access is given via its `isInvalid` methods.
* group methods related to fields together in call order. So if one like `addErrorIf` calls another one addErrorIf should be above it. So we can read/navigate from top to bottom.
* name fields simply error and warning 
  * `reportList` makes it unclear that these are actually errors. warnings
  are also reports since both are TrackerError/WarningReports
  * no need to add suffixes like list as this is clear from its type

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationErrorReporter.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationErrorReporter.java
@@ -35,6 +35,8 @@ import java.util.Set;
 import java.util.function.BooleanSupplier;
 import java.util.function.Predicate;
 
+import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.Value;
 
 import org.hisp.dhis.tracker.TrackerIdSchemeParams;
@@ -59,6 +61,7 @@ public class ValidationErrorReporter
 
     TrackerIdSchemeParams idSchemes;
 
+    @Getter( AccessLevel.NONE )
     /*
      * Keeps track of all the invalid Tracker objects (i.e. objects with at
      * least one TrackerErrorReport in the ValidationErrorReporter) encountered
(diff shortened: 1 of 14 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationErrorReporter.java
```diff
@@ -94,31 +97,37 @@ public ValidationErrorReporter( TrackerIdSchemeParams idSchemes, boolean failFas
      */
     public ValidationErrorReporter( TrackerIdSchemeParams idSchemes )
     {
-        this.reportList = new ArrayList<>();
-        this.warningsReportList = new ArrayList<>();
-        this.invalidDTOs = new HashMap<>();
+        this.errors = new ArrayList<>();
```
enricocolasante, line 100: We can call the other constructor here
`super(idSchemes, false);`

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationErrorReporter.java
```diff
@@ -48,23 +51,23 @@
  * @author Morten Svanæs <msvanaes@dhis2.org>
  */
 @Value
-// TODO: should this be "ValidationReporter" since it does not only report
-// errors ?
 public class ValidationErrorReporter
 {
-    private final List<TrackerErrorReport> reportList;
+    List<TrackerErrorReport> errors;
```
stian-sandvold, line 56: Renaming errors and warnings make sense to me!

stian-sandvold (approved):
FYI: There is a rogue file in your PR: dhis-2/dhis-services/dhis-service-tracker/.attach_pid646883

PR #12478, opened 2022-12-05
Additional metadata checks for organisation units

jason-p-pickering:


olavpo (approved):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/AbstractDataIntegrityIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+        if ( expectedDetailsComments != null )
+        {
+            Set detailsComments = issues.stream().map( issue -> issue.getComment() ).collect( Collectors.toSet() );
+            assertEquals( expectedDetailsNames, detailsComments );
+        }
+
+        assertEquals( issueType, details.getIssuesIdType() );
+    }
+
+    public final void DataIntegrityDivideByZeroTestTemplate( String detailsIssueType, String check )
```
jbee, line 155: It is not clear from the name what this does. 
Maybe a name like `assertNoIssues`.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/AbstractDataIntegrityIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+            .content()
+            .as( JsonDataIntegrityDetails.class );
+        assertTrue( details.exists() );
+        assertTrue( details.isObject() );
+        JsonList<JsonDataIntegrityDetails.JsonDataIntegrityIssue> issues = details.getIssues();
+        assertTrue( issues.exists() );
+        assertEquals( 0, issues.size() );
+    }
+
+    public final void DataIntegrityNegativeTestTemplate( String detailsIssueType, String check )
```
jbee, line 179: This looks almost the same as the one above just that it expects a percent 0 instead percent null. 
You can make them one method with an argument for the expected percent:
```java
void assertNoIssues( tring detailsIssueType, String check, boolean percentageIsPresent )
```

jbee (changes requested):
Overall it looks good. A couple of cleanup/code-style/convention things:

* method names should not start with an upper case letter
* don't use `null` for any collection value, use empty ones and check `!isEmpty()` instead of `!= null`, e.g. `Set.of()`
* fields in test classes should be `private`

olavpo (approved):

jbee (approved):

PR #12481, opened 2022-12-05
chore: make tracker RelationshipItem API safer

teleivo:
* `RelationshipItem` does not guard its state-space. Exactly one of the fields trackedEntity, enrollment, event must be set! So we should not use a builder since they are not all optional.

=> create constructors and factories that only allow creating a valid `RelationshipItem`

* Once you have an item you don't know which of the fields contains the UID. You also don't know what the item represents. This makes it hard to just answer a question like `does this entity exist in the DB?` without either trying all fields or actually figuring out what it is. The figuring out part is a lot of work 😅 Leading down a rabbit hole of classes back to the  [Relationship.relationshipType](https://github.com/dhis2/dhis2-core/blob/3b2d94a99479c414ca93272c1c1fa7660bfeeede/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/domain/Relationship.java#L57) -> [RelationshipType](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipType.java#L50-L52) -> [RelationshipConstraint](https://github.com/dhis2/dhis2-core/blob/3b2d94a99479c414ca93272c1c1fa7660bfeeede/dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipConstraint.java#L51) -> [RelationshipEntity](https://github.com/dhis2/dhis2-core/blob/3b2d94a99479c414ca93272c1c1fa7660bfeeede/dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipEntity.java#L35-L39). The latter not being very useful as its just some String telling me I've got a TEI, enrollment or event. This entire flow means you need to leave our domain to recover information we actually had when the `RelationshipItem` was instantiated. 

=> Add type information so we can programmatically access the identifier that is set, without knowing exactly whether its a trackedEntity, enrollment or event.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/RelationshipTrackerConverterService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/converter/RelationshipTrackerConverterService.java
@@ -87,13 +87,20 @@ public class RelationshipTrackerConverterService
 
     private RelationshipItem convertRelationshipType( org.hisp.dhis.relationship.RelationshipItem from )
     {
-        RelationshipItem relationshipItem = new RelationshipItem();
-        relationshipItem.setEnrollment( from.getProgramInstance() != null ? from.getProgramInstance().getUid() : null );
-        relationshipItem
-            .setEvent( from.getProgramStageInstance() != null ? from.getProgramStageInstance().getUid() : null );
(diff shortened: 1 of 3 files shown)


teleivo:
cannot be done as easily. right now the tracker domain model is also used to represent invalid user input. it can thus not be refactored to provide a safe API as one would expect 😬

PR #12489, opened 2022-12-07
fix: NPE when updating programRuleActions

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14105

### Issue:
- Program is not provided in payload, so Preheat doesn't have  it.

Error log

```
java.lang.NullPointerException
	at org.hisp.dhis.programrule.action.validation.ProgramRuleActionValidationContextLoader.load(ProgramRuleActionValidationContextLoader.java:85)
	at org.hisp.dhis.programrule.action.validation.ProgramRuleActionValidationContextLoader$$FastClassBySpringCGLIB$$23bc3578.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
```

Sample payload : 

```
{
  "programRuleActions":
  [
    {
      "lastUpdated": "2017-05-23T00:29:24.454",
      "id": "UUwZWS8uirn",
      "href": "https://play.dhis2.org/2.37.8/api/29/programRuleActions/UUwZWS8uirn",
      "created": "2016-04-12T15:58:55.041",
      "content": "The hemoglobin value cannot be above 99",
      "externalAccess": false,
      "displayContent": "The hemoglobin value cannot be above 99",
      "programRuleActionType": "SHOWERROR",
      "evaluationTime": "ALWAYS",
      "favorite": false,
      "access":
      {
        "read": true,
        "update": true,
        "externalize": false,
        "delete": true,
        "write": true,
        "manage": true
      },
      "programRule":
      {
        "id": "dahuKlP7jR2"
      },
      "lastUpdatedBy":
      {
        "displayName": "John Traore",
        "name": "John Traore",
        "id": "xE7jOejl9FI",
        "username": "admin"
      },
      "dataElement":
      {
        "id": "vANAXwtLwcT"
      },
      "sharing":
      {
        "external": false
      },
      "favorites":
      [],
      "evaluationEnvironments":
      [
        "WEB",
        "ANDROID"
      ],
      "translations":
      [],
      "userGroupAccesses":
      [],
      "attributeValues":
      [],
      "userAccesses":
      []
    }
  ]
}
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/action/validation/ProgramRuleActionValidationContextLoader.java
+++ b/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/action/validation/ProgramRuleActionValidationContextLoader.java
@@ -32,6 +32,7 @@ import java.util.List;
 
 import javax.annotation.Nonnull;
 
+import org.hisp.dhis.common.IdentifiableObjectManager;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.option.Option;
 import org.hisp.dhis.option.OptionGroup;
@@ -57,10 +58,13 @@ public class ProgramRuleActionValidationContextLoader
     @Nonnull
     private final ProgramRuleActionValidationService validationService;
 
+    private final IdentifiableObjectManager objectManager;
+
     public ProgramRuleActionValidationContextLoader(
-        @Nonnull ProgramRuleActionValidationService actionValidationService )
+        @Nonnull ProgramRuleActionValidationService actionValidationService, IdentifiableObjectManager manager )
     {
         this.validationService = actionValidationService;
+        this.objectManager = manager;
     }
 
     @Transactional( readOnly = true )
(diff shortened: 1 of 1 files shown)


mortenoh (approved):

jbee (approved):

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/action/validation/ProgramRuleActionValidationContextLoader.java
```diff
@@ -72,6 +76,12 @@ public ProgramRuleActionValidationContext load( Preheat preheat, PreheatIdentifi
 
         Program program = preheat.get( preheatIdentifier, Program.class, rule.getProgram() );
 
+        if ( program == null )
```
larshelge, line 79: Just wondering, do we know the root cause? This seems a little bit like a patch/workaround, though I can be work. Could we figure out why the program is not part of the preheat in the first place?

vietnguyen: @larshelge Please check the attached payload. User wants to update ProgramRuleAction, so they only post only that object with all the necessary references. 
However with current model design, we have ProgramRule -- ProgramRuleAction, so there no link to Program.

larshelge: Okay understand. Would there be a benefit in adding the program to the bundle in the beginning of the process, to avoid fetching it many times or similar? Just checking, might be wrong.

vietnguyen: Yes, I'm working on new Preheat hook so developers can add on demand objects UID to the preheat process. Currently we will need to add an if-else block to the `PreheatService` which is not good.

PR #12501, opened 2022-12-09
chore: Cleanup tracker reporter

enricocolasante:
Make some cleanup in `TrackerValidationReport`
- errorReportsByCode can be changed to a list as is never actually used as a map
- timings is moved to `ValidationErrorReporter` as they are never really exposed to the API, they are @JsonIgnore annotated. We could think of remove them once for all

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerObjectReport.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerObjectReport.java
@@ -28,9 +28,7 @@
 package org.hisp.dhis.tracker.report;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerObjectReport.java
```diff
@@ -63,7 +61,7 @@
     @JsonProperty
     private String uid;
 
-    private Map<TrackerErrorCode, List<TrackerErrorReport>> errorReportsByCode = new HashMap<>();
+    private List<TrackerErrorReport> errorReports = new ArrayList<>();
```
teleivo, line 64: I think we can name this field even `errors`. We can also do this later when we rename the `TrackerErrorReport` though 😄

teleivo (approved):

PR #12502, opened 2022-12-09
fix: Allow ordering TEIs by attribute [DHIS2-13981]

muilpp:
When trying to order by the following values: updatedAtClient, createdAtClient and enrolledAt, the query fails because we are not including any of these values in the select clause nor in the subquery.


lucaCambi77 (commented):
Good job, however, I would add more java docs to private methods so they are clear

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +486,55 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+                        .append( addStaticColumn( orderParam.getField(), select.toString() ).replace( "pi", "tei" ) );
+                }
+                else
+                {
+                    orderQuery.append( addStaticColumn( orderParam.getField(), select.toString() ) );
+                }
+            }
+        }
+
+        select.append( SPACE );
```
teleivo, line 521: Is this needed? This is also done right after the method returns in https://github.com/dhis2/dhis2-core/pull/12502/files#diff-f960125887353fe285946328173a4e612632e40c0bca283e9d30c824348764dbR497

If not you could pass in the select as `String` and save making two calls to build the string. This would also make it clear that you are not mutating the select.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +486,55 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+        {
+            select.append( getOrderByQuery( params, select ) );
+        }
+
         select.append( SPACE );
 
         return select.toString();
     }
 
+    private String getOrderByQuery( TrackedEntityInstanceQueryParams params, StringBuilder select )
```
teleivo, line 500: This method name is confusing to me 🤔 If I understand the code correctly: you are adding select expressions for the columns that we will order by http://forcedotcom.github.io/phoenix/#select

There is already `getQueryOrderBy` which is actually generating the ORDER BY clause.

These two method names are too similar while doing different things.

Maybe we can call it `getQuerySelectOrderByColumns`, `getQuerySelectOrderBy`, ...

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +486,55 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+
+        for ( OrderParam orderParam : params.getOrders() )
+        {
+            if ( isStaticColumn( orderParam.getField() ) )
+            {
+                if ( ENROLLED_AT.getPropName().equalsIgnoreCase( orderParam.getField() ) )
+                {
+                    //In the main query, we need to use the alias "tei" to fetch the value of the field enrolledAt
+                    orderQuery
+                        .append( addStaticColumn( orderParam.getField(), select.toString() ).replace( "pi", "tei" ) );
```
teleivo, line 512: Do I understand correctly that if someone renames the alias in one place (i.e. main query) but not the other we will have a bug?

If so we should tie the alias from the main query together with the alias used here. For example via constant.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -909,6 +984,26 @@ else if ( !params.isOrganisationUnitMode( OrganisationUnitSelectionMode.ALL ) )
...
+     * order by is enrolledAt, we need to join the program instance table to be
+     * able to select and order by this value
+     *
+     * @param params
+     * @return a SQL INNER JOIN for program instances
+     */
+    private String getFromSubQueryJoinProgramInstanceConditions( TrackedEntityInstanceQueryParams params )
+    {
+
+        if ( params.getOrders() != null
```
teleivo, line 998: Would be great to add a sensible default (empty collection) to params.orders so we do not have to do any null checks.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -909,6 +984,26 @@ else if ( !params.isOrganisationUnitMode( OrganisationUnitSelectionMode.ALL ) )
...
+            && params.getOrders().stream().anyMatch( p -> ENROLLED_AT.getPropName().equalsIgnoreCase( p.getField() ) ) )
```
teleivo, line 999: If we are doing something like `ENROLLED_AT.getPropName().equalsIgnoreCase` we could maybe add this to the OrderColumn.

`ENROLLED_AT.isEqualToProperty(String)` or `ENROLLED_AT.isProperty(String)`

or so

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStoreTest.java
```diff
@@ -0,0 +1,233 @@
...
+        params.setOrders( List.of( new OrderParam( "updatedAtClient", OrderParam.SortDirection.ASC ) ) );
+
+        //when
+        store.getTrackedEntityInstanceIds( params );
+
+        //then
+        Mockito.verify( jdbcTemplate, Mockito.atLeast( 1 ) ).queryForRowSet( sqlQueryCaptor.capture() );
+        String sqlQuery = sqlQueryCaptor.getValue();
+        assertAll( () -> assertEquals( 2, countMatches( sqlQuery, ", tei.lastUpdatedAtClient" ) ),
+            () -> assertEquals( 2, countMatches( sqlQuery, "ORDER BY tei.lastUpdatedAtClient ASC" ) ) );
```
teleivo, line 159: While its interesting to write a unit test and assert on the SQL string I feel like we don't really know whether ordering works.

The fact that `ORDER BY tei.lastUpdatedAtClient ASC` appears twice in the SQL does not tell us whether the result is ordered by `lastUpdatedAtClient`.

muilpp: I agree, with this test we only assert that the element we want to order by is present in the query, and is present the expected amount of times, but we don't even know if the query will be successful.
My original idea was to write an integration test for this issue, but the convenience test doesn't support postgres and the controllerIntegrationTest have the problem of the async call with the transaction manager.

There was also the option of writing and e2e, but I don't think this feature is worth an e2e test. To me e2e should be used for features that if broken, deploys to production freeze until the issue is fixed. This one has been broken for a while and no one reported it...

That's why I wrote this unit test, to at least make sure that the param is present in the query, and as soon as we find a solution for the controllerIntegrationTest, we can write an integration test where we assert that the TEIs in the result are sorted as expected.

teleivo: But you can write an integration test using postgres for this store. This allows you to test that the you get the result in the correct order. Like in https://github.com/dhis2/dhis2-core/blob/373a5bb1b4a3ad14d44eefc57701dfdb716915fe/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/EventExporterTest.java#L682-L696

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -1268,6 +1272,11 @@ public static boolean isStaticColumn( String propName )
                 .anyMatch( orderColumn -> orderColumn.getPropName().equals( propName ) );
         }
 
+        public static boolean isFieldEqualToEnrolledAt( String fieldName )
```
teleivo, line 1275: If you make this a method on OrderColumn (non-static) than it can be reused for other values on the enum. For example like `ENROLLED_AT.isProperty("enrolledat")`. You could also add one for the column if we are also comparing to that `isColumn()`

lucaCambi77 (changes requested):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/HibernateTrackedEntityInstanceStoreTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @Test
+    void whenOrderingByCreatedAtThenEntitiesAreSortedById()
+    {
+        params.setOrders( List.of( new OrderParam( "createdAt", OrderParam.SortDirection.ASC ) ) );
+
+        List<Long> list = teiStore.getTrackedEntityInstanceIds( params );
+
+        assertEquals( "[1, 2, 3, 4]", list.toString() );
+    }
+
```
lucaCambi77: We should make assertEquals comparing lists instead of strings. For this, we can create `TrackedEntityInstance` in a specific order before each test or in the test so we have objects to compare

lucaCambi77 (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -1268,6 +1272,11 @@ public static boolean isStaticColumn( String propName )
...
+        {
+            return ENROLLED_AT.getPropName().equalsIgnoreCase( fieldName );
+        }
+
```
lucaCambi77, line 1306: A method with the name `isPropertyEqualTo`  and then it only compares ENROLLED_AT doesn't make much sense to me

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -1277,14 +1277,43 @@ public static boolean isFieldEqualToEnrolledAt( String fieldName )
...
+         */
+        public static String getColumn( String property, String alias )
+        {
+            String column = Arrays.stream( values() )
                 .filter( orderColumn -> orderColumn.getPropName().equals( property ) )
                 .map( OrderColumn::getColumn )
                 .findFirst()
                 .orElseThrow(
                     () -> new IllegalArgumentException( "Property" + property + " is not defined as order column" ) );
+
```
lucaCambi77, line 1337: The getColumn method should just return a value and not an exception. If we want to add an exception we should add a validation when we collect the order values from a request. At this level, I think we should log

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -1344,15 +1364,33 @@ private String getQueryGroupBy( TrackedEntityInstanceQueryParams params )
...
-                    .append( statementBuilder.columnQuote( orderAttribute.getItemId() ) )
-                    .append( SPACE );
+                if ( isStaticColumn( orderParam.getField() ) )
+                {
+                    //In the main query, we need to use the main query alias to fetch the value of the field enrolledAt
+                    if ( isFieldEqualToEnrolledAt( orderParam.getField() ) )
+                    {
+                        groupBy
+                            .append( addStaticColumn( orderParam.getField(), groupBy.toString(), MAIN_QUERY_ALIAS ) );
+                    }
```
lucaCambi77, line 1370: the addStaticColumn method here is used only with MAIN_QUERY_ALIAS

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +487,90 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+    /**
+     * Gets all the static fields from the order parameters and appends them to
+     * the string builder, so they can be used in the query select clause
+     *
+     * @param params
+     * @param select current select statement
+     * @return all the static fields that are not yet present in the select
+     *         statement
+     */
+    private String getOrderParamForSelectClause( TrackedEntityInstanceQueryParams params, StringBuilder select )
```
teleivo, line 507: Please pass in a `String select`. Passing in a `StringBuilder` signals to me that you'll add to it. Since you are just interested in `select.toString()` pass in a `String`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +487,90 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+    {
+        StringBuilder orderQuery = new StringBuilder();
+
+        for ( OrderParam orderParam : params.getOrders() )
+        {
+            if ( isStaticColumn( orderParam.getField() ) )
+            {
+                if ( ENROLLED_AT.isPropertyEqualTo( orderParam.getField() ) )
+                {
+                    //In the main query, we need to use the main query alias to fetch the value of the field enrolledAt
```
teleivo, line 517: That should be clear from you now using a constant `MAIN_QUERY_ALIAS`. I don't think the comment is needed.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -484,11 +487,90 @@ private String getQuerySelect( TrackedEntityInstanceQueryParams params, boolean
...
+    /**
+     * Checks if a field is a valid column and whether is present in the current
+     * query.
+     *
+     * @param field name of the field to add in the query
+     * @param sqlQuery current sql query
+     * @return the mapped column name if not present in the query, empty
+     *         otherwise
+     */
+    private String addStaticColumn( String field, String sqlQuery )
```
teleivo, line 540: This method is confusing me 😅 

The javadocs say `Checks if a field is a valid column and whether is present in the current query.`. The method returns a `String` and the method name is `addStaticColumn` suggesting that something is added to something. A method that just checking something should probably return a boolean. It should also not have a word like `add` suggesting something is changed. Reminds me of https://en.wikipedia.org/wiki/Command%E2%80%93query_separation

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/HibernateTrackedEntityInstanceStoreTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @ParameterizedTest
+    @MethodSource( "provideOrderParamsAndExpectedResult" )
+    void whenOrderingByBasicStaticFieldsThenEntitiesAreSortedAccordingly( String input, String expected )
+    {
+        params.setOrders( List.of( new OrderParam( input, OrderParam.SortDirection.ASC ) ) );
+
+        List<Long> list = teiStore.getTrackedEntityInstanceIds( params );
+        list = reduceListValuesToStartCountingFromOne( list );
```
teleivo, line 86: If you are doing these two in every test you might as well create a private wrapper function `getTrackedEntityInstanceIds` so this just becomes one call like `List<Long> list = getTrackedEntityInstanceIds( params );`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/HibernateTrackedEntityInstanceStoreTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        assertEquals( expected, list.toString() );
```
teleivo, line 88: We should assert using `List<Long>` instead of String.

teleivo (commented):
Great that you got the integration test to work 🎉 

I feel like we are doing too many operations only using Strings instead of types like `OrderColumn`. This might also be due to how `OrderColumn` is written. We should just not simply accept that since we can change it :yum:

For example

```java
for ( OrderParam orderParam : params.getOrders() )
{
    if ( isStaticColumn( orderParam.getField() ) )
    {
        if ( ENROLLED_AT.isPropertyEqualTo( orderParam.getField() ) )
        {
            //In the main query, we need to use the main query alias to fetch the value of the field enrolledAt
            orderQuery
                .append( addStaticColumn( orderParam.getField(), select.toString(), MAIN_QUERY_ALIAS ) );
        }
        else
        {
            orderQuery.append( addStaticColumn( orderParam.getField(), select.toString() ) );
        }
    }
}
```

`orderParam.getField()` is a String that I assume is coming from user input. We should translate that into an `OrderColumn`. If we don't we can never be certain its valid. We also never have the information we actually need like its `propertyName` or `column`. The above code for example is only concerned with a column that is an `OrderColumn`. Instead we pass Strings around to methods like `addStaticColumn` or check again on a String inside `ENROLLED_AT.isPropertyEqualTo`.

If we have something like

```java
public static Optional<OrderColumn> findColumn(String propertyName) {
    return Arrays.stream(values())
            .filter( oc -> oc.getPropName().equalsIgnoreCase( propertyName ) )
            .findFirst();
}
```

You would not only not need `isStaticColumn` you would also not need `isPropertyEqualTo`. You can then also operate more safely/less defensively in the rest of the code.

We need to work on our types, they are there to help us 😄

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityInstanceService.java
```diff
@@ -248,19 +248,19 @@ public List<Long> getTrackedEntityInstanceIds( TrackedEntityInstanceQueryParams
...
+        List<QueryItem> sortableAttributes = params.getAttributesAndFilters().stream()
+            .filter( queryItem -> ordersIdentifier.contains( queryItem.getItemId() ) )
+            .collect( Collectors.toList() );
 
-            // adding to attributes conditionally if they are also not present
-            // in filters.
-            params.addAttributesIfNotExist( QueryItem.getQueryItems( sortAttributes ).stream()
-                .filter( sAtt -> !params.getFilters().contains( sAtt ) ).collect( Collectors.toList() ) );
-        }
+        List<TrackedEntityAttribute> sortAttributes = sortableAttributes.stream()
```
muilpp, line 258: we only use ordersIdentifier and sortableAttributes once, could we avoid declaring these two lists and then map, filter and map all in one statement?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -637,60 +568,48 @@ private String getFromSubQuery( TrackedEntityInstanceQueryParams params, boolean
...
 
         for ( OrderParam orderParam : params.getOrders() )
         {
-            if ( isStaticColumn( orderParam.getField() ) )
+            Optional<TrackedEntityInstanceQueryParams.OrderColumn> orderColumn = getColumn( orderParam.getField() );
+
+            if ( orderColumn.isPresent() )
             {
-                orderAttributes.append( addStaticColumn( orderParam.getField(), query ) );
+                columns.add( orderColumn.get().getTableAlias() + "." + orderColumn.get().getColumn() );
```
muilpp, line 582: Maybe it would make sense to have a method in the OrderColumnd enum that build this, we could reuse it while constructing the order by sql too. Something like:

```
public static String getColumnStatement() {
    return getTableAlias() + "." + getColumn()
} 
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -637,60 +568,48 @@ private String getFromSubQuery( TrackedEntityInstanceQueryParams params, boolean
...
+     *
+     * @param params
+     * @return List of QueryItem
+     */
+    private List<QueryItem> sortableAttributesAndFilters( TrackedEntityInstanceQueryParams params )
+    {
+        List<String> ordersIdentifier = params.getOrders().stream()
+            .map( OrderParam::getField )
+            .collect( Collectors.toList() );
+        return params.getAttributesAndFilters().stream()
```
muilpp, line 610: same as above, why don't we avoid the creation of the list and map and filter in the same statement instead?

lucaCambi77:
We worked with @muilpp to make the code cleaner and readable. It should be ok now, We moved the tests to integration testing in `TrackedEntityInstanceServiceTest`. Further improvements for the ordering will follow up with https://dhis2.atlassian.net/browse/TECH-1492.
However, with this PR we are adding an inner join with the ProgramInstance table which can potentially cause performance issues. There is already in the query an EXISTS clause that checks for the existence of the enrollment. So i'd say, before merging this PR, to make sure there are no performance issue with this new query

lucaCambi77 (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -562,44 +576,48 @@ private String getFromSubQuery( TrackedEntityInstanceQueryParams params, boolean
...
-                .append( ".value AS " )
-                .append( statementBuilder.columnQuote( orderAttribute.getItemId() ) );
-        }
-
-        return orderAttributes.toString();
+        List<String> ordersIdentifier = params.getOrders().stream()
+            .map( OrderParam::getField )
+            .collect( Collectors.toList() );
+        return params.getAttributesAndFilters().stream()
+            .filter( queryItem -> ordersIdentifier.contains( queryItem.getItemId() ) )
```
teleivo, line 619: I would rather use a Set then a List (even if ordersIdentifier might not contain large N items)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceServiceTest.java
```diff
@@ -335,18 +340,195 @@ void testTrackedEntityInstanceGridWithNoDisplayAttributes()
...
+        entityInstanceD1.setCreated( DateTime.now().plusDays( 2 ).toDate() );
+        addEntityInstances();
+
+        TrackedEntityInstanceQueryParams params = new TrackedEntityInstanceQueryParams();
+        params.setOrganisationUnits( Sets.newHashSet( organisationUnit ) );
+        params.setOrders( List.of( new OrderParam( "createdAt", OrderParam.SortDirection.ASC ) ) );
+
+        List<Long> teiIdList = entityInstanceService.getTrackedEntityInstanceIds( params, true, true );
+
+        Assertions.assertEquals( 4, teiIdList.size() );
```
teleivo, line 368: please use a static import for `Assertions`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceServiceTest.java
```diff
@@ -335,18 +340,195 @@ void testTrackedEntityInstanceGridWithNoDisplayAttributes()
...
+        Assertions.assertEquals( entityInstanceA1.getId(), teiIdList.get( 0 ) );
```
teleivo, line 369: Please replace your assertions with (so remove asserting on size as well)

`assertEquals(List.of(entityInstanceA1.getId(), entityInstanceB1.getId(), ...), teiIdList);`

as order matters. In cases order does not matter use https://github.com/dhis2/dhis2-core/blob/221cce4b8e9154e50c9a6458b713ba405cf27111/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java#L67

This asserts the size is 4 and the ids are in the expected order. If the assertion is not met we will get an error message with the context needed to figure out whats wrong. The way you have written it it will fail with

```
org.opentest4j.AssertionFailedError: 
Expected :1
Actual   :3
```

which is not helpful when this test only fails on GitHub/Jenkins.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceServiceTest.java
```diff
@@ -335,18 +340,195 @@ void testTrackedEntityInstanceGridWithNoDisplayAttributes()
...
+    {
+        injectSecurityContext( superUser );
+        addEntityInstances();
+
+        TrackedEntityInstanceQueryParams params = new TrackedEntityInstanceQueryParams();
+        params.setOrganisationUnits( Sets.newHashSet( organisationUnit ) );
+
+        List<Long> teiIdList = entityInstanceService.getTrackedEntityInstanceIds( params, true, true );
+
+        Assertions.assertEquals( 4, teiIdList.size() );
```
teleivo, line 471: Size is already asserted on by `assertEquals` below so please remove it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityInstanceQueryParams.java
```diff
@@ -1273,36 +1251,56 @@ public void setTrackedEntityTypes( List<TrackedEntityType> trackedEntityTypes )
...
         {
-            return Arrays.stream( OrderColumn.values() )
-                .anyMatch( orderColumn -> orderColumn.getPropName().equals( propName ) );
+            return propName.equalsIgnoreCase( property );
         }
 
-        public static String getColumn( String property )
+        /**
+         * @param property
+         * @return an Optinal of an OrderColumn matching by property name
```
teleivo, line 1279: typo in Optinal

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -576,18 +571,21 @@ private String getFromSubQuery( TrackedEntityInstanceQueryParams params, boolean
...
+
+        LinkedHashSet<String> columns = new LinkedHashSet<>( defaultSelectFields );
 
         for ( OrderParam orderParam : params.getOrders() )
         {
             Optional<TrackedEntityInstanceQueryParams.OrderColumn> orderColumn = findColumn( orderParam.getField() );
 
-            if ( orderColumn.isPresent() )
+            if ( orderColumn.isPresent()
+                && !defaultSelectFields.contains( orderColumn.get().getSqlColumnWithTableAlias() ) )
```
teleivo, line 586: I don't think its a good idea to do these substring searches in loops. Where you not using LinkedHashSet explicitly to account for duplicates? 🤔 This should turn the operation into O(1). Same above in `getQuerySelect`.

muilpp:
I created a new test in the locust repository to validate the overall performance of running queries to retrieve tracked entities order by specific parameters: https://github.com/dhis2/performance-tests-locust/pull/63

After running it, I see no impact on the performance. I also run it manually in the instance where the performance database is seeded and the response time was not increased.

teleivo (approved):

PR #12524, opened 2022-12-13
chore: add smaller Validator interface DHIS2-14298

teleivo:
First step in breaking up the large `TrackerValidationHook`s into type specific ones. Most of our hooks are already type specific i.e. only validating trackedEntity, enrollment, ... Transitioned the first one `TrackedEntityAttributeValidator` over to the new `Validator` interface.

Smaller Validator interface

`void validate( ValidationErrorReporter reporter, TrackerBundle bundle, T input );`

allows us to define hooks only for trackedEntity, enrollment, ...

All the type specific hooks have `skipOnError=false` so we can simply run them after the `PreCheck` hooks in case they succeeded or have `skipOnError=false`.

I will first migrate the type specific hooks https://github.com/dhis2/dhis2-core/blob/2f607f53c7c920bab48832e9effb2de8696652a3/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/config/TrackerValidationConfig.java#L112-L134

After that I will migrate/break up the `PreCheck` hooks.

Since rule engine validations are different we need 2 `Validators` configurations. Like we already have in our `TrackerValidationConfig`. Aggregating the type specific `Validators` in an interface/implementations will make the transition easier. It also makes testing easier. A `Validator` that needs a service like the `aclService` is easier to build this way as well then what we currently have in the `TrackerValidationConfig`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationService.java
@@ -63,19 +63,26 @@ public class DefaultTrackerValidationService
     @Qualifier( "ruleEngineValidationHooks" )
     private final List<TrackerValidationHook> ruleEngineValidationHooks;
 
+    @Qualifier( "org.hisp.dhis.tracker.validation.DefaultValidators" )
+    private final Validators validators;
+
+    @Qualifier( "org.hisp.dhis.tracker.validation.RuleEngineValidators" )
+    private final Validators ruleEngineValidators;
+
     @Override
     public TrackerValidationReport validate( TrackerBundle bundle )
     {
-        return validate( bundle, validationHooks );
+        return validate( bundle, validationHooks, validators );
     }
 
     @Override
     public TrackerValidationReport validateRuleEngine( TrackerBundle bundle )
     {
-        return validate( bundle, ruleEngineValidationHooks );
+        return validate( bundle, ruleEngineValidationHooks, ruleEngineValidators );
     }
 
-    private TrackerValidationReport validate( TrackerBundle bundle, List<TrackerValidationHook> hooks )
+    private TrackerValidationReport validate( TrackerBundle bundle, List<TrackerValidationHook> hooks,
+        Validators validators )
     {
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationService.java
```diff
@@ -141,10 +149,30 @@ private void validateTrackedEntities( TrackerBundle bundle, List<TrackerValidati
...
-                        break; // skip subsequent validation hooks for this invalid entity
+                        failed = true;
+                        break; // skip subsequent validation for this invalid entity
                     }
                 }
             }
+
+            if ( failed )
+            {
+                continue;
```
enricocolasante, line 160:
```suggestion
                continue; // skip specific validations for this invalid entity
```

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/Validator.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Validates given input and adds errors and warnings to {@code reporter}.
+     *
+     * @param reporter aggregates errors and warnings
+     * @param bundle tracker bundle
+     * @param input input to validate
+     */
+    void validate( ValidationErrorReporter reporter, TrackerBundle bundle, T input );
+
+    default boolean needsToRun( TrackerImportStrategy strategy )
```
stian-sandvold, line 46: Would it make sense to rename this "runForStrategy"? needsToRun sounds like it should always run, or it doesn't matter to my ears/eyes 👀

stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/Validator.java
teleivo: We can definitely rename this. I just copied it over from the old TrackerValidationHook.
Maybe `runOnStrategy` 🤔

PR #12525, opened 2022-12-13
chore: Rename and clean up report model objects [TECH-14298]

enricocolasante:
Some renaming in the report model.

- `TrackerImportReport` -> `ImportReport`
- `TrackerBundleReport` -> `PersistenceReport`
- `TrackerObjectReport` -> `Entity`
- `TrackerStatus` -> `Status`
- `TrackerStats` -> `Stats`
- `TrackerTimingStats` -> `TimingStats`
- `TrackerStats` -> `Stats`

Also removed and reorganized some redudant information that was included in the report, like status in the ObjectReport that was always OK and the stats that were repeated

I didn't change the json fields that are exposed to the client.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/Entity.java
```diff
@@ -56,27 +56,27 @@
...
-    public TrackerObjectReport( TrackerType trackerType )
+    public Entity( TrackerType trackerType )
     {
         this.trackerType = trackerType;
     }
 
-    public TrackerObjectReport( TrackerType trackerType, String uid, Integer index )
+    public Entity( TrackerType trackerType, String uid, Integer index )
     {
         this.trackerType = trackerType;
```
teleivo, line 73: Should we use the `TrackerDto` or at least accept it in the constructor? Maybe that makes things easier. The TrackerDto is the type and its UID.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/Entity.java
```diff
@@ -41,10 +41,10 @@
  * @author Morten Olav Hansen <mortenoh@gmail.com>
  */
 @Data
-public class TrackerObjectReport
+public class Entity
```
teleivo, line 44: What does `TrackerObjectReport` represent? Where is it used? Just wondering about the name. My understanding so far is that the things that were called `report` are actually `items` on a report to me. So the report aggregates those items.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/PersistenceReport.java
```diff
@@ -45,40 +45,30 @@
...
+@Getter
 @NoArgsConstructor
 @AllArgsConstructor
-public class TrackerBundleReport
+public class PersistenceReport
 {
-    @JsonProperty
-    @Builder.Default
-    private TrackerStatus status = TrackerStatus.OK;
+    public static PersistenceReport emptyReport()
```
teleivo, line 53: Is it a good idea to have/useful to have an empty report?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerTypeReport.java
```diff
@@ -65,27 +65,27 @@ public TrackerTypeReport( TrackerType trackerType )
 
     @JsonCreator
     public TrackerTypeReport( @JsonProperty( "trackerType" ) TrackerType trackerType,
-        @JsonProperty( "stats" ) TrackerStats stats,
+        @JsonProperty( "stats" ) Stats stats,
         @JsonProperty( "sideEffectDataBundles" ) List<TrackerSideEffectDataBundle> sideEffectDataBundles,
-        @JsonProperty( "objectReports" ) List<TrackerObjectReport> objectReports )
+        @JsonProperty( "objectReports" ) List<Entity> objectReports )
```
teleivo, line 70: `objectReports` should then also become `entityReport`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerTypeReport.java
```diff
@@ -102,22 +102,22 @@ public boolean isEmpty()
         return getErrorReports().isEmpty();
     }
 
-    public void addObjectReport( TrackerObjectReport objectReport )
+    public void addEntity( Entity objectReport )
```
teleivo, line 105: `objectReport` should then be `entity`

Why does this `Entity` have an index? Why does the position matter?

teleivo (commented):
Did you add https://github.com/dhis2/dhis2-core/pull/12525/files#diff-f611e64a95d360b6daf45df814cd9d659070d31a0f22eab617966b56711646dcR61 on purpose?

I am wondering about the name of `Entity` and what the purpose is of the formerly `ObjectReport`. I totally agree that the use of `report` to me does not fit in our code. As what we call report is usually an item in a report that aggregates those items. Maybe you can add a sample response JSON we give to the user so its easier to understand which JSON property is coming from what class/field.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/TrackerTypeReport.java
enricocolasante: I am not sure.
I am just renaming in this PR, I will check on these details on a later PR

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/PersistenceReport.java
enricocolasante: Maybe not, I will however think about it in next PR(s) as we need to have the case when this report is empty/null that is when there are validation errors and nothing was persisted

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/Entity.java
enricocolasante: I think we will need to rethink the whole Entity concept in the Report. At the moment we are being really verbose just to give the client a list of UIDs by type. That is actually a list of TrackerDtos

enricocolasante:
> Did you add https://github.com/dhis2/dhis2-core/pull/12525/files#diff-f611e64a95d360b6daf45df814cd9d659070d31a0f22eab617966b56711646dcR61 on purpose?
> 
> I am wondering about the name of `Entity` and what the purpose is of the formerly `ObjectReport`. I totally agree that the use of `report` to me does not fit in our code. As what we call report is usually an item in a report that aggregates those items. Maybe you can add a sample response JSON we give to the user so its easier to understand which JSON property is coming from what class/field.

The @jsonIgnore annotation is there temporarly I hope.
I just wanted to remove it from the report as it is alreay present in the ImportReport object itself but it is calculated in PersistenceReport. I will think of a better way to calculate stats.

The idea of this PR was just to rename some of the report fields that were really repetitive and misleading.
My next PR would really be about the report itself and how we present it to the client

teleivo (approved):

PR #12527, opened 2022-12-13
chore: migrate event validators DHIS2-14298

teleivo:
next in line after https://github.com/dhis2/dhis2-core/pull/12526

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/RuleEngineValidators.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/RuleEngineValidators.java
@@ -80,7 +80,7 @@ public class RuleEngineValidators implements Validators
     public List<Validator<Event>> getEventValidators()
     {
         return List.of(
-            eventDataValuesValidator,
-            eventRuleValidator );
+            eventRuleValidator,
+            eventDataValuesValidator );
     }
 }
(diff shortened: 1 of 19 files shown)


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/RuleEngineValidators.java
```diff
@@ -68,4 +75,12 @@ public List<Validator<Enrollment>> getEnrollmentValidators()
             enrollmentRuleValidationHook,
             enrollmentAttributeValidator );
     }
+
+    @Override
+    public List<Validator<Event>> getEventValidators()
+    {
+        return List.of(
+            eventDataValuesValidator,
+            eventRuleValidator );
```
enricocolasante, line 84: I think it is important to maintain order here.
So `eventRuleValidator` should run before `eventDataValuesValidator`

enricocolasante (commented):
Order of validators in rule engine validation

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/RuleEngineValidators.java
teleivo: good catch! that is actually what I am trying to do 😄 will adapt

enricocolasante (approved):

PR #12528, opened 2022-12-13
feat: named job queues [DHIS2-14314]

jbee:
### Summary
Adds named job queues to the scheduling.

All REST API interaction with a queue is handled by a new controller `JobSchedulerController` which is backed by a new service `JobQueueService`.

A queue
* has a unique name of 1-50 characters
* a job sequence of two or more jobs
* a trigger CRON expression

Internally the queue is stored as part of the `JobConfiguration`s.
All configurations that are assigned to a queue have a `queueName` and `queuePosition` (0-n) set.

When a job is removed from a queue its CRON expression is always reset and the job gets disabled.
This is to force user action to configure the job properly before it will run.

### Endpoints

*  `GET  /api/scheduler/queues`: List all names of existing queues
* `GET  /api/scheduler/queues/{name}`: Show queue
* `POST /api/scheduler/queues/{name}` : Create a new queue
* `PUT  /api/scheduler/queues/{name}` : Update an existing queue
* `DELETE  /api/scheduler/queues/{name}` : Delete an existing queue
* `GET /api/scheduler`: List scheduler entries (to support app listing with queues)
* `GET  /api/scheduler/queueable` : List all job entries that can be added to a queue (same data per entry as in above)

### Automatic Testing
A new controller test was added covering all new controller methods with success and error scenarios.


vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -332,6 +332,11 @@
     E7006( "Failed to execute job: `{0}`" ),
     E7007( "Delay must be not null for job with scheduling type FIXED_DELAY: `{0}`" ),
     E7010( "Failed to validate job runtime: `{0}`" ),
+    E7020( "Job Queue does not exist: `{0}`" ),
```
larshelge, line 335: Use lower-case "q" in "queue" for constistency.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobQueueService.java
```diff
@@ -0,0 +1,211 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+/**
+ *
+ *
+ * @author Jan Bernitt
+ */
+@Service
+@AllArgsConstructor
```
larshelge, line 59: Minor, I think we could favor `@RequiredArgsConstructor` over `@AllArgsConstructor` since it forces the dependencies to be `final`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultSchedulingManager.java
```diff
@@ -101,21 +103,28 @@ public DefaultSchedulingManager( JobService jobService, JobConfigurationService
...
     {
-        log.info( String.format( "Scheduling job: %s", configuration ) );
+        boolean isCronTriggered = configuration.getJobType().getSchedulingType() == SchedulingType.CRON;
+        if ( isCronTriggered && isNullOrEmpty( configuration.getCronExpression() ) )
+        {
+            log.info(
+                format( "Ignoring CRON schedule for %s as it is not ready to run.", configuration.getJobType() ) );
+            return;
+        }
+        log.info( format( "Scheduling job: %s", configuration ) );
```
larshelge, line 113: With `@Slf4j` we can use `{}` to indicate variables and skip `String.format` entirely.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
```diff
@@ -0,0 +1,274 @@
...
+@RestController
+@RequestMapping( value = "/scheduler" )
+@AllArgsConstructor
+public class JobSchedulerController
+{
+    private final JobConfigurationService jobConfigurationService;
+
+    private final JobQueueService jobQueueService;
+
+    @Value
```
larshelge, line 87: Nice use of `@Value` for immutability.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * API for scheduler list and named queues (sequences).
+ *
+ * @author Jan Bernitt
+ */
+@OpenApi.Tags( "system" )
+@RestController
+@RequestMapping( value = "/scheduler" )
+@AllArgsConstructor
```
larshelge, line 80: Minor, I think we could favor `@RequiredArgsConstructor` over `@AllArgsConstructor` since it forces the dependencies to be final.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
```diff
@@ -0,0 +1,274 @@
...
+public class JobSchedulerController
+{
+    private final JobConfigurationService jobConfigurationService;
+
+    private final JobQueueService jobQueueService;
+
+    @Value
+    static class SchedulerEntry
```
larshelge, line 88: Just an opinion, I think it may make the controller cleaner by moving these classes to their own files.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
```diff
@@ -0,0 +1,274 @@
...
+                config.getJobStatus(),
+                config.isEnabled(),
+                config.getJobType().isConfigurable(),
+                List.of( SchedulerEntryJob.of( config ) ) );
+        }
+
+        static SchedulerEntry of( List<JobConfiguration> queue )
+        {
+            JobConfiguration trigger = queue.get( 0 );
+            if ( trigger.getQueueName() == null )
```
larshelge, line 130: Perhaps add a method with Javadoc to `JobConfiguration` to encapsulate the null check and have the method explicitly indicate the meaning of this property being null?

larshelge (approved):

larshelge:
Code looks great, nice and compact @jbee :+1:

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
jbee: I moved them out because they have many fields and some initialiser logic. 

Generally, if such dedicated controller "DTO" types are small I think it is more readable having them right there where they are used. But there is a limit when this becomes too much and I think you are right that this was clearly reached in this case.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
jbee: I added a `isUsedInQueue()`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
larshelge: Thanks, agreed.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
larshelge: Looks great.

PR #12558, opened 2022-12-14
chore: Decouple validation package from report package [DHIS2-14298]

enricocolasante:
Validation step is now returning a `ValidationResult` object that exposes a errors and warnings through a `Validation` interface.
Validation step now is almost completely independent from report step (still need to move TrackerErrorCode, will do in next PR).
Report step is responsible to transform the `ValidationResult` in a `ValidationReport` that will be used in the `ImportReport` and presented to in the response to the client.
Once TrackerErrorCode is moved as well, I will try to make validation objects access as strict as possible.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/Validation.java
```diff
@@ -0,0 +1,39 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.validation;
+
```
teleivo, line 29: Can you please add javadocs to the public things that we expose from the validation package. I think it would also be great if they are linked to from the package-info.java 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/Error.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.hisp.dhis.tracker.TrackerType;
+import org.hisp.dhis.tracker.report.TrackerErrorCode;
+
+@Value
+@Builder
+@RequiredArgsConstructor
+public class Error implements Validation
+{
+    private final String errorMessage;
```
teleivo, line 42: I think we can shorten the field names to `message`, `code`, `type`. I would also do the same in `Warning`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
+
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+/**
+ * @author Morten Olav Hansen <mortenoh@gmail.com>
```
teleivo, line 39: I think it deviated enough that Morten is not the author anymore 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private final List<Warning> warnings;
+
+    public ValidationResult()
+    {
+        this.errors = new ArrayList<>();
+        this.warnings = new ArrayList<>();
+    }
+
+    // -----------------------------------------------------------------------------------
```
teleivo, line 55: I would remove this comment

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -0,0 +1,149 @@
...
+    {
+        addErrorIfNotExisting( error );
+        return this;
+    }
+
+    public ValidationResult addErrors( List<Error> errors )
+    {
+        for ( Error error : errors )
+        {
+            addErrorIfNotExisting( error );
```
teleivo, line 85: This always puzzled me, why are we doing this instead of declaring it a `Set<Error>` 🤷🏻 
We can adapt this later on.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationServiceConfigOrderTest.java
```diff
@@ -73,10 +72,10 @@ void hooksAreExecutedInTrackerValidationConfigOrder()
             .events( Collections.singletonList( event ) )
             .build();
 
-        ValidationReport report = trackerValidationService.validate( bundle );
+        ValidationResult report = trackerValidationService.validate( bundle );
 
         assertTrue( report.hasErrors() );
         assertEquals( 1, report.getErrors().size() );
-        assertEquals( E1048, report.getErrors().get( 0 ).getErrorCode() );
+        assertEquals( E1048.name(), report.getErrors().get( 0 ).getCode() );
```
teleivo, line 79: If we have many of these we could add an assertion to our assertion helpers. We have some for the two reports. Maybe we can straighten them after this PR. They were always super close but not quite as ValidationErrorReporter and the other don't share an interface.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/AssertionsTest.java
```diff
@@ -49,58 +51,55 @@
...
     @Test
     void testAssertHasOnlyErrorsFailsIfReportHasMoreErrors()
     {
 
-        ValidationReport report = new ValidationReport();
-        report.addError( TrackerErrorReport.builder().errorCode( TrackerErrorCode.E1000 ).build() );
-        report.addError( TrackerErrorReport.builder().errorCode( TrackerErrorCode.E1019 ).build() );
-        report.addError( TrackerErrorReport.builder().errorCode( TrackerErrorCode.E1041 ).build() );
+        ValidationReport report = new ValidationReport(
+            List.of( Error.builder().errorCode( TrackerErrorCode.E1000.name() ).build(),
```
teleivo, line 64: At some point we should remove the Error.builder as all fields are required. We can add a factory if that makes it nicer to read than `new Error()`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -0,0 +1,149 @@
...
+
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+/**
+ * @author Morten Olav Hansen <mortenoh@gmail.com>
+ */
+@ToString
+@EqualsAndHashCode
+public class ValidationResult
```
teleivo, line 43: Could we make this package private and expose only an interface like we do with `Validation`?

So `ValidationResult` has 2 `List<Validation>` one for error and one for warning.

This would be great so once this leaves the validation package it can only be read.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
enricocolasante: I think that now that we decoupled validation from reporting we can have it as a Set.
I think before it was not a Set just to preserve the order

lucaCambi77 (approved):

PR #12560, opened 2022-12-14
fix: broken Redis cache invalidation for non-long identifiers

netroms:


dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/PostCacheEventPublisher.java
```diff
@@ -94,6 +118,30 @@ public void onPostDelete( PostDeleteEvent postDeleteEvent )
...
+        else if ( entity instanceof CompleteDataSetRegistration )
+        {
+            id = getCompleteDataSetRegistrationId( entity );
+        }
+        else if ( entity instanceof DataStatisticsEvent )
+        {
+            DataStatisticsEvent dataStatisticsEvent = (DataStatisticsEvent) entity;
+            id = dataStatisticsEvent.getId();
+        }
+        else if ( IdentifiableObject.class.isAssignableFrom( entity.getClass() ) )
```
jbee, line 139: can be `instanceof` check instead

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/CacheInvalidationListener.java
```diff
@@ -85,7 +97,7 @@ private void handleMessage( String message )
         if ( serverInstanceId.equals( uid ) )
         {
             log.debug( "Message came from this server, ignoring." );
-            return;
+            //            return;
```
jbee, line 100: Is this intentionally? If so remove the line entirely.

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/PostCacheEventPublisher.java
```diff
@@ -114,6 +162,47 @@ private void publishMessage( Class<?> realClass, String message )
...
+    {
+        DataValue dataValue = (DataValue) entity;
+
+        long dataElementId = dataValue.getDataElement().getId();
+        long periodId = dataValue.getPeriod().getId();
+        long organisationUnitId = dataValue.getSource().getId();
+        long categoryOptionComboId = dataValue.getAttributeOptionCombo().getId();
+        long attributeOptionComboId = dataValue.getAttributeOptionCombo().getId();
+
+        return dataElementId + ";" + periodId + ";" + organisationUnitId + ";" + categoryOptionComboId + ";"
```
jbee, line 175: At some point we should think about using dedicated ID classes for this combination of IDs. Joining and splitting is relying on too much implicit knowledge. I guess we could equally well capture these in a small inner class like this? 
```java
@Value
static class DataValueId {
  long dataElementId;
  long periodId;
  long organisationUnitId;
  long categoryOptionComboId;
  long attributeOptionComboId;
}
```
Then you can simply do:
```java
return new DataValueId(dataElementId, periodId, organisationUnitId, categoryOptionComboId, attributeOptionComboId);
```
And so on for all the other IDs you need.

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/PostCacheEventPublisher.java
```diff
@@ -114,6 +157,46 @@ private void publishMessage( Class<?> realClass, String message )
...
+    private Serializable getCompleteDataSetRegistrationId( Object entity )
+    {
+        CompleteDataSetRegistration completeDataSetRegistration = (CompleteDataSetRegistration) entity;
+
+        long dataSetId = completeDataSetRegistration.getDataSet().getId();
+        long periodId = completeDataSetRegistration.getPeriod().getId();
+        long organisationUnitId = completeDataSetRegistration.getSource().getId();
+        long categoryOptionComboId = completeDataSetRegistration.getAttributeOptionCombo().getId();
+
+        return dataSetId + ";" + periodId + ";" + organisationUnitId + ";" + categoryOptionComboId;
```
larshelge, line 197: I think this part can be written better by joining e.g. with `String.join(";", List.of(dataSetId, periodId, ..))`.

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/PostCacheEventPublisher.java
netroms: The problem here is that I need to serialize the IDs for the message passing through Redis. Without having to serialize and deserialize the messages, just using colon and semicolon separated IDs is by far the easiest.

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/PostCacheEventPublisher.java
netroms: This does not work without having to cast to charseq

larshelge (approved):

PR #12581, opened 2022-12-15
chore: Only expose immutable result from validation package [DHIS2-14298]

enricocolasante:
- Create a `ValidationResult` interface that just expose the Error and Warning lists.
- Change the flow in the `DefaultTrackerImporterService` to use `ValidationReport` instead of `ValidationResult`

`ValidationResult` is an interface that expose the output of validation step.
`ValidationReport` is an object containing the validation results in a format that is ready to be presented to the client.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationReport.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationReport.java
@@ -71,8 +71,8 @@ public class ValidationReport
     public static ValidationReport fromResult( ValidationResult validationResult )
     {
         ValidationReport validationReport = new ValidationReport();
-        validationReport.addErrors( convertToError( validationResult.getErrors() ) );
-        validationReport.addWarnings( convertToWarning( validationResult.getWarnings() ) );
+        validationReport.addErrors( convertToError( List.copyOf( validationResult.getErrors() ) ) );
+        validationReport.addWarnings( convertToWarning( List.copyOf( validationResult.getWarnings() ) ) );
         return validationReport;
     }
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationReport.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationReport.java
@@ -59,16 +59,16 @@ public class ValidationReport
         return new ValidationReport();
     }
 
-    public static ValidationReport merge( ValidationResult result, ValidationResult anotherResult )
+    public static ValidationReport merge( ValidationResult ValidationResult, ValidationResult anotherValidationResult )
     {
         return new ValidationReport();
     }
 
-    public static ValidationReport fromResult( ValidationResult result )
+    public static ValidationReport fromResult( ValidationResult ValidationResult )
     {
         ValidationReport validationReport = new ValidationReport();
-        validationReport.addErrors( convertToError( result.getErrors() ) );
-        validationReport.addWarnings( convertToWarning( result.getWarnings() ) );
+        validationReport.addErrors( convertToError( ValidationResult.getErrors() ) );
+        validationReport.addWarnings( convertToWarning( ValidationResult.getWarnings() ) );
         return validationReport;
     }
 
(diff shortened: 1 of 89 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -27,116 +27,15 @@
...
 import java.util.List;
-import java.util.function.Predicate;
 
-import lombok.EqualsAndHashCode;
-import lombok.ToString;
-
-@ToString
-@EqualsAndHashCode
-public class ValidationResult
+public interface ValidationResult
```
teleivo, line 32: Can you please add some javadocs explaining what its for. As we are exposing this publicly. Would also be great if you can link to it from the package-info.java.

So if someone wants to understand a tracker package they can start by opening package-info.java It states the responsibilities and the main entry point. For validation that is the ValidationService. And maybe what comes out when computations are done in this phase. For validation that is the ValidationResult.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/Result.java
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return Collections.unmodifiableList( errors );
+    }
+
+    public List<Validation> getWarnings()
+    {
+        return Collections.unmodifiableList( warnings );
+    }
+
+    public Result addError( Error error )
```
teleivo, line 68: Are there some methods we can remove? 🤔
And make these errors and warnings a Set? (or would you rather do that separately 👌🏻 )

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/report/ValidationReport.java
```diff
@@ -59,16 +59,20 @@ public static ValidationReport emptyReport()
         return new ValidationReport();
     }
 
-    public static ValidationReport merge( ValidationResult result, ValidationResult anotherResult )
+    public static ValidationReport merge( ValidationResult validationResult, ValidationResult anotherValidationResult )
     {
-        return new ValidationReport();
+        ValidationReport validationReport = fromResult( validationResult );
```
teleivo, line 64: Can we improve the API so that we just do `ValidationReport validationReport = fromResult( validationResult, anotherValidationResult);`? So there is no mutation of the ValidationReport

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultTrackerValidationService.java
```diff
@@ -112,9 +110,10 @@ private ValidationResult validate( TrackerBundle bundle, Validators validators )
         bundle.setEvents( persistables.getEvents() );
         bundle.setRelationships( persistables.getRelationships() );
 
-        reporter.getErrors().addAll( persistables.getErrors() );
-
-        return new ValidationResult().addErrors( reporter.getErrors() ).addWarnings( reporter.getWarnings() );
+        return new Result()
```
teleivo, line 113: Maybe now or later 😋 Would be great if `Result` is created in one go without adding to it. For only adding errors and warnings that would be as easy as `new Result(errors, warnings)`. Since we have 2 separate errors `new Result(errors, othererrors, warnings)` looks a bit weird 🤔

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/DefaultValidationServiceTest.java
```diff
@@ -179,44 +179,11 @@ void warningsDoNotInvalidateAndRemoveEntities()
         when( validators.getTrackedEntityValidator() ).thenReturn( v1 );
         service = new DefaultValidationService( validators, ruleEngineValidators );
 
-        ValidationResult result = service.validate( bundle );
+        ValidationResult report = service.validate( bundle );
```
teleivo, line 182: `result` seems more appropriate for type `ValidationResult`

stian-sandvold (approved):

PR #12597, opened 2022-12-16
chore: single validator DHIS2-14298

teleivo:
* first commit: create one main `Validator<TrackerBundle>` out of the per type specific ones like `Validator<Enrollment>`
* second commit: replace hook(s) with validator as the old hook interface does not exist anymore


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultValidationService.java
```diff
@@ -49,24 +49,24 @@
 {
 
     @Qualifier( "org.hisp.dhis.tracker.validation.DefaultValidators" )
-    private final Validators validators;
+    private final Validator<TrackerBundle> validator;
```
enricocolasante, line 52: I think the qualifiers are wrong.
org.hisp.dhis.tracker.validation.validator.DefaultValidator should be on validator
and org.hisp.dhis.tracker.validation.validator.RuleEngineValidator should be on ruleEngineValidator

enricocolasante (commented):
Check the qualifiers for validator components

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/DefaultValidationService.java
teleivo: Thank you for spotting this 😂

enricocolasante (approved):
Check the qualifiers for validator components

stian-sandvold (approved):

stian-sandvold (approved):

PR #12600, opened 2022-12-18
fix: Handle numeric value types for event data items [DHIS2-14355]

larshelge:
The code for checking whether a value dimension was numeric used `getValueType() == ValueType.NUMBER` which only covers one of the many numeric value types. The right approach is `getValueType().isNumeric()`.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -819,19 +818,6 @@ protected String getSelectSql( QueryFilter filter, QueryItem item, EventQueryPar
...
-     * Checks if the ValueType, in the given parameters, is of type NUMERIC.
-     *
-     * @param params the {@link EventQueryParams}.
-     * @return true if the parameter value type is numeric, false otherwise.
-     */
-    private boolean isParamsValueTypeNumeric( EventQueryParams params )
-    {
-        return params != null &&
-            params.getValue() instanceof ValueTypedDimensionalItemObject &&
-            ((ValueTypedDimensionalItemObject) params.getValue()).getValueType() == ValueType.NUMBER;
```
larshelge, line 832: There are more numeric value types than just `NUMBER`. Use `valueType.isNumeric()` instead.

PR #12604, opened 2022-12-19
test: make assertHasError reusable using Validation DSHI2-14298

teleivo:
## Changes to existing error/warning assertions

Now that we have the Validation interface implemented by errors and warnings we can use it to create assertions on Validation. On top of that we create small wrappers that we use in our tests.

These are some of the assertions we now have

```java
assertHasError( reporter, event, E1118 );
assertHasError( reporter, event, E1118, "contains this message" );
```

You can use it with a `Reporter` or a `ValidationResult`. There is one for errors and one for warnings.

* All use the same order of parameters with `ValidationCode` and if provided
a Validation.message.
* Add `assertHasWarning` so its symmetric and as easy to assert errors
* Since we always assert that a `TrackerDto` has a given error/warning allow passing it directly instead of only its type and uid.

```diff
- hasTrackerError( reporter, E1118, TrackerType.EVENT, event.getUid() );
+ assertHasError( reporter, event, E1118 );
```

## Changes to existing tests

We are now using the above assertions in all our validator tests.

* remove redundant assertion `assertTrue( reporter.hasErrors() )` like in 

```java
assertTrue( reporter.hasErrors() );
assertHasError( reporter, enrollment, E1016 );
```

`assertHasError` will make sure there are errors, otherwise there would not even be an error like `E1016` in the `errors`.

* Use `assertHasError` to ensure the entire message or a substring

```java
assertHasError( reporter, enrollment, E1127 );
assertThat( reporter.getErrors().get( 0 ).getMessage(), containsString( "program" ) );
```

is equal to `assertHasError( reporter, enrollment, E1127, "program" );`

* wrap multiple assertions in an `assertAll` like

```java
assertAll(
        () -> assertEquals( 1, reporter.getErrors().size() ),
        () -> assertHasError( reporter, enrollment, ValidationCode.E1076 )
);
```

Asserting on size without also asserting on `assertHasError` will leave out useful context in case of an error.

* Use `assertIsEmpty` to assert lack of errors

```diff
assertThat( reporter.getErrors(), hasSize( 0 ) );
assertIsEmpty( reporter.getErrors() );
```

This ensures any errors are printed that we did not expect. Makes tests easier to debug.


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/EnrollmentAttributeValidatorTest.java
```diff
@@ -176,8 +176,9 @@ void shouldFailValidationWhenValueIsNullAndAttributeIsMandatory()
 
         validator.validate( reporter, bundle, enrollment );
 
-        assertThat( reporter.getErrors(), hasSize( 1 ) );
-        hasTrackerError( reporter, ValidationCode.E1076, TrackerType.ENROLLMENT, enrollment.getUid() );
+        assertAll(
```
enricocolasante, line 179: If this is a very common pattern for an assertion, would it make sense to create something like `assertHasOnlyError` to wrap this logic?

enricocolasante (approved):

stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/EnrollmentAttributeValidatorTest.java
teleivo: Often enough, yes we could create it. I just wonder whether we should even assert on the error size at all 🤔

PR #12620, opened 2022-12-20
chore: don't mutate bundle in validation DHIS2-14298

teleivo:
Return ValidationResult with the persistable entities so that the import does the mutation. Later on we should pass the persistables to the persist step without mutating the bundle at all.

* remove unused methods and factories only used in tests from `Result`. It should be a simple container of persistable entities, errors and warnings.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
@@ -27,14 +27,36 @@
  */
 package org.hisp.dhis.tracker.validation;
 
+import java.util.List;
 import java.util.Set;
 
+import org.hisp.dhis.tracker.TrackerImportStrategy;
+import org.hisp.dhis.tracker.domain.Enrollment;
+import org.hisp.dhis.tracker.domain.Event;
+import org.hisp.dhis.tracker.domain.Relationship;
+import org.hisp.dhis.tracker.domain.TrackedEntity;
+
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
```diff
@@ -27,14 +27,32 @@
...
+import org.hisp.dhis.tracker.domain.Event;
+import org.hisp.dhis.tracker.domain.Relationship;
+import org.hisp.dhis.tracker.domain.TrackedEntity;
+
 /**
- * The result of the validation step. It contains a set of {@link Validation}s
- * divided in errors and warnings.
+ * ValidationResult is the result of the validation step. The tracked entities,
+ * enrollments, events and relationships can be persisted (created, updated,
+ * deleted) and thus trusted to be in a valid state. Any entity that passed
```
enricocolasante, line 41: Even if we risk to be repetitive I would specify here that we call these entities, `persistable entities`.
Then in the last sentence, Persistable entities it is cristal clear what we are talking about.

enricocolasante (approved):
Just one comment on a comment

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationResult.java
teleivo: done 😄

stian-sandvold (approved):

PR #12635, opened 2022-12-21
fix: (2.39) Unable to load persisted SMS configurations [DHIS2-14351]

zubaira:


lucaCambi77 (commented):
Any way to add a test here? What is this change fixing?

zubaira:
> Any way to add a test here? What is this change fixing?

SMS configurated is unable to load once Tomcat is restarted and this makes it a bit tricky to test. If there is any way we can simulate tomcat restart functionality in the test then we can add some tests for this.

stian-sandvold (approved):

PR #12644, opened 2022-12-22
feat: Last and first aggregation type for text value type [DHIS2-13449]

larshelge:
Introduces two new aggregation types:

* `LAST_LAST_ORG_UNIT`: Uses last value/event in the time/period dimension and last value/event in the org unit/all other dimensions.
* `FIRST_FIRST_ORG_UNIT`: Uses first value/event in the time/period dimension and first value/event in the org unit/all other dimensions.

This implies that no aggregation takes place. Instead the last or first data value/event is selected across all dimensions (e.g. across the org unit and time/period dimensions).

Adds support in the aggregate analytics engine for program data elements of value type `TEXT` and aggregation type `LAST_LAST_ORG_UNIT` and `FIRST_FIRST_ORG_UNIT`, meaning that the last/first text value will be returned.

The selection of last/first data value/event happens in a sub query which uses a window function to rank data values/events in time, either ascending or descending. The sub query will return exactly one data value/event per dimension item combination as specified in the analytics query. The outer query will filter by the top ranked result from the inner query, returning the last/first record.

Note: One Sonarqube warning exists but I suggest ignoring it, as it does not have any practical impact and is not related to this PR.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
```diff
@@ -222,10 +222,14 @@ public boolean isAggregatable( AggregationType aggregationType )
         {
             return false;
         }
+
         if ( this == TEXT )
         {
-            return aggregationType == AggregationType.NONE;
+            return Set.of(
```
maikelarabori, line 228: Not a big deal, but I believe would be nicer to extract this "Set" into a constant.

maikelarabori (approved):
Looks good @larshelge, just left a minor comment.

Thanks

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java
larshelge: Agreed, it is not ideal to create that set on every invocation.

PR #12657, opened 2022-12-29
feat: Add program stage working list [DHIS2-14367]

muilpp:
This is the implementation of a new endpoint that will allow us storing the definition of program stage working lists. That is, we will be able to create, edit, update and delete filters for a specific working list.

Link to the ticket: https://dhis2.atlassian.net/browse/DHIS2-14367


enricocolasante (changes requested):
One thing that would help a lot to review any feature is to have a description in the PR, if not a description at least a link to the Jira Issue. Another thing that would help is to have the issue linked with the Epic.

I am confused and I think we should have a chat about this PR.
From what I understand there are 2 different ways to import (and even more important validate) metadata, through metadata importer (directly using AbstractCrudController) or creating a separate endpoint.
This PR makes it possible to import a `ProgramStageWorkingListDefinition` using the corresponding endpoint but it also allow it to import through metadata endpoint, right?
If this is the case, using PreEntity and PostEntity to validate instead of the metadata validation is allowing to import such an entity without passing through validation.

We should understand if we need to have this entity to also be imported through metadata importer and if this is the case I think we should implement validation in there.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
```diff
@@ -0,0 +1,142 @@
...
+import org.hisp.dhis.translation.Translatable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
+
+@Setter
+@JacksonXmlRootElement( localName = "programStageWorkingListDefinition", namespace = DxfNamespaces.DXF_2_0 )
+public class ProgramStageWorkingListDefinition extends BaseIdentifiableObject
```
stian-sandvold, line 48: I see you use the name in different ways different places: ProgramStageWorkingList and ProgramStageWorklingListDefinition.

I would prefer the first everywhere, making it a bit shorter and easier to read.

Is there any specific reason to include "Definition"?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @JsonProperty
+    @JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
+    public ProgramStageQueryCriteria getProgramStageQueryCriteria()
+    {
+        return programStageQueryCriteria;
+    }
+
+    @Override
+    public boolean equals( Object o )
```
stian-sandvold, line 112: Any special reason to include an override for equals and hashcode?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinitionService.java
```diff
@@ -0,0 +1,81 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.programstageworkinglistdefinition;
+
+import java.util.List;
+
+public interface ProgramStageWorkingListDefinitionService
+{
+    String ID = ProgramStageWorkingListDefinition.class.getName();
```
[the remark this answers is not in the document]
stian-sandvold: Most likely this is related to logging; Since we're not logging, it should not be needed :)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ValidationUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class ValidationUtil
+{
+
+    private ValidationUtil()
+    {
+    }
+
+    public static void validateAttributeValueFilters( List<String> errors, List<AttributeValueFilter> attributes,
+        TrackedEntityAttributeService teaService )
```
stian-sandvold, line 48: Not sure about passing services as arguments to an utils class. Would it be better to try to fetch all attributes before validation, then passing the attributes instead (Any missing attributes = wrong uid or not existing)?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_10__Create_table_program_stage_working_list_definition.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE if not exists programstageworkinglistdefinition (
+    programstageworkinglistdefinitionid integer NOT NULL,
+    uid character varying(11) NOT NULL,
+    code character varying(50),
+    created timestamp without time zone NOT NULL,
+    lastupdated timestamp without time zone NOT NULL,
+    lastupdatedby integer,
```
stian-sandvold, line 7: You are missing "CreatedBy", its defined in the hbm file, but not here :)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ProgramStageWorkingListController.java
```diff
@@ -0,0 +1,76 @@
...
+        this.workingListDefinitionService = workingListDefinitionService;
+    }
+
+    @Override
+    public void preCreateEntity( ProgramStageWorkingListDefinition workingListDefinition )
+    {
+        List<String> errors = workingListDefinitionService.validate( workingListDefinition );
+        if ( !errors.isEmpty() )
+        {
+            throw new IllegalQueryException( errors.toString() );
```
stian-sandvold, line 62: Will these validations always be triggered when importing the data? I am not completely sure myself, but it should be tested if you havent already :)

lucaCambi77 (commented):
A quick note. We should avoid this kind of PRs, they have a lot of different commits and are difficult to read. Let's focus on starting from unit tests and then moving to the logic and the context we are working in. Otherwise, we'll end up copying and pasting from other code that already exists. In any case, in order to make the PR easier to review for others, we can also make more than one until it is clear enough to go through

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageQueryCriteria.java
```diff
@@ -0,0 +1,230 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.programstageworkinglistdefinition;
```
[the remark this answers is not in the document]
muilpp: I'm fine with that, but I'm following the package convention we already have in the project, for example:
 - programstagefilter
 - trackedentityattributevalue
 - trackedentitycomment
 - trackedentitydatavalue
 - trackedentityfilter
 
 If the new convention is to not flatten concepts, I'll be happy to split it in two packages.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
muilpp: The reason to include "definition" at the end is to make clear that this is not a working list, this is just its "metadata", it's just a bunch of properties that will be applied to filter and sort a working list accordingly.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
muilpp: If I don't include them Sonar complaints with S2160: Subclasses that add fields should override "equals"

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_10__Create_table_program_stage_working_list_definition.sql
muilpp: I just added it, but I see we are actually not using it in any of the other filters, should I also not add it in this one?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/util/ValidationUtil.java
muilpp: I agree, the thing is that before I get each attribute, I need to iterate the list and add an error in case any of them is empty, so some code would be duplicated. Would you be fine using a functional interface instead?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ProgramStageWorkingListController.java
muilpp: Now there's the bundle hook class, so the validations will be run in both cases.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
enricocolasante: I also would be in favor to remove the "definition" part from the name.
All the metadata that we have do not use "definition" to make it clear that is a metadata, having it just here for me increase the confusion instead of make things more clear.
On the other end, if you want to differentiate between data and metadata, we often use the "Instance" in the data to make it clear that is an instance of such metadata

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListDefinitionObjectBundleHook.java
```diff
@@ -0,0 +1,186 @@
...
+import org.hisp.dhis.programstageworkinglistdefinition.ProgramStageWorkingListDefinition;
+import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
+import org.hisp.dhis.trackedentity.TrackedEntityAttributeService;
+import org.hisp.dhis.trackedentityfilter.AttributeValueFilter;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+
+@Component( "org.hisp.dhis.dxf2.metadata.objectbundle.hooks.ProgramStageWorkingListDefinitionObjectBundleHook" )
+@AllArgsConstructor
+public class ProgramStageWorkingListDefinitionObjectBundleHook
```
enricocolasante, line 61: Generally I would prefer to have a unit test for a validation class and then add some meaningful test case in the integration test.
Not really a suggestion, just a note to remember that we still need to have a discussion on test strategy to agree on where to test what.

enricocolasante (commented):
For me now everything is working fine.
I would just consider again removing "definition" from the name

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListDefinitionObjectBundleHook.java
muilpp: You are right, I added as many unit tests as possible and I left the integration tests needed to validate the rest. I agree we need to have a discussion to agree on a test strategy as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinition.java
JoakimSM: @muilpp: @enricocolasante makes an excellent point. Skipping "Definition" makes sense to me now.

muilpp: I've removed the word definition, now it's called programStageWorkingList everywhere

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/programstageworkinglist/hibernate/ProgramStageWorkingListDefinition.hbm.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
+<!DOCTYPE hibernate-mapping PUBLIC
+  "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
+  "http://www.hibernate.org/dtd/hibernate-mapping-3.0.dtd"
+  [<!ENTITY identifiableProperties SYSTEM "classpath://org/hisp/dhis/common/identifiableProperties.hbm">]
```
enricocolasante, line 5: Still definition in the name

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHook.java
```diff
@@ -0,0 +1,186 @@
...
+{
+
+    private final DataElementService dataElementService;
+
+    private final OrganisationUnitService organisationUnitService;
+
+    private final TrackedEntityAttributeService teaService;
+
+    @Override
+    public void validate( ProgramStageWorkingList workingListDefinition, ObjectBundle bundle,
```
enricocolasante, line 72: Still definition in the name.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHook.java
```diff
@@ -0,0 +1,186 @@
...
+        Consumer<ErrorReport> addReports )
+    {
+
+        ProgramStageQueryCriteria queryCriteria = workingListDefinition.getProgramStageQueryCriteria();
```
enricocolasante, line 76: Same here

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_12__Create_table_program_stage_working_list_definition.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE if not exists programstageworkinglist (
```
enricocolasante, line 1: Still definition in the name

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/ProgramStageWorkingListControllerTest.java
```diff
@@ -0,0 +1,223 @@
...
+    {
+        programId = assertStatus( HttpStatus.CREATED, POST( "/programs/",
+            "{'name': 'ProgramTest', 'shortName': 'ProgramTest', 'programType': 'WITHOUT_REGISTRATION'}" ) );
+
+        programStageId = assertStatus( HttpStatus.CREATED,
+            POST( "/programStages/", "{'name': 'ProgramStageTest', 'program':" + "{'id': '" + programId + "'}}" ) );
+    }
+
+    @Test
+    void shouldReturnWorkingListIdDefinitionWhenWorkingListCreated()
```
enricocolasante, line 59: In most of the test name definition is still there

enricocolasante (approved):
Some lost "definition" here and there

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageQueryCriteria.java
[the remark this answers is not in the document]
muilpp: I think you reviewed outdated code, now it's called programstageworkinglist, it's under `dhis` in its own package because it's a new feature. Even though it's similar to programstagefilter and trackedentityfilter, these two will need to be refactored, that's why I didn't want to put them together under one package.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ProgramStageWorkingListController.java
```diff
@@ -0,0 +1,76 @@
...
+@OpenApi.Tags( "tracker" )
+@Controller
+@RequestMapping( value = ProgramStageWorkingListDefinitionSchemaDescriptor.API_ENDPOINT )
+@ApiVersion( include = { DhisApiVersion.ALL, DhisApiVersion.DEFAULT } )
+public class ProgramStageWorkingListController
+    extends AbstractCrudController<ProgramStageWorkingListDefinition>
+{
+    private final ProgramStageWorkingListDefinitionService workingListDefinitionService;
+
+    public ProgramStageWorkingListController( ProgramStageWorkingListDefinitionService workingListDefinitionService )
```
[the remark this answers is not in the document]
muilpp: this is also outdated, now the controller only extends the AbstractCrudController, there was no need for the service

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageWorkingListDefinitionService.java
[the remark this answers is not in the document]
muilpp: this is not there anymoer

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageQueryCriteria.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    private List<EventDataFilter> dataFilters;
+
+    /**
+     * Property to filter tracked entity instances based on tracked entity
+     * attribute values
+     */
+    private List<AttributeValueFilter> attributeValueFilters = new ArrayList<>();
+
+    @JsonProperty
```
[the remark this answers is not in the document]
muilpp: you're right, Json should be enough, they got it because of legacy reasons

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/DefaultProgramStageWorkingListDefinitionService.java
```diff
@@ -0,0 +1,257 @@
...
+        validateName( errors, programStageWorkingListDefinition );
+
+        ProgramStageQueryCriteria queryCriteria = programStageWorkingListDefinition.getProgramStageQueryCriteria();
+
+        if ( queryCriteria == null )
+        {
+            return errors;
+        }
+
+        validateDateFilterPeriods( errors, queryCriteria );
```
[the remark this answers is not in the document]
muilpp: This is also outdated, but still interesting. 
It might be a better design to validate during the creation of the object, the thing is that I think that's something we should agree as a team, because currently this is validated in the bundle hook objects, this one in particular is in ProgramStageWorkingListObjectBundleHook, as it is for many other objects. That's why I think we should agree on one of the two options and stick to it.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/ProgramStageWorkingListDefinitionControllerTest.java
```diff
@@ -0,0 +1,320 @@
...
+            "The working list id: " + workingListId2 + " is not present in the response" );
+    }
+
+    @Test
+    void shouldUpdateWorkingListDefinitionWhenUpdateRequested()
+    {
+        String workingListId = createWorkingListDefinition( "Test working list to update" );
+
+        String updatedName = "Updated working list";
+        assertStatus( HttpStatus.OK, PUT( "/programStageWorkingListDefinitions/" + workingListId,
```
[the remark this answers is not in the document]
muilpp: I referred to `ProgramStageControllerTest` to build this test, do you mean it's better to use the method `assertWebMessage`, like in `ProgramRuleControllerTest`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglistdefinition/ProgramStageQueryCriteria.java
```diff
@@ -0,0 +1,230 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+
+/**
+ * Represents the filtering/sorting criteria to be used when querying program
+ * stage working lists.
+ */
+@Setter
+public class ProgramStageQueryCriteria implements Serializable
```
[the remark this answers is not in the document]
muilpp: I agree with this. 
For collections, no problem, but.. for other objects, even enums like ouMode or eventStatus, I'm not sure if we can generate bugs if we assign them to a value by default. We would have to analyze and refactor the code, and even then still generate some bugs? I think that should be a separate task.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHook.java
```diff
@@ -0,0 +1,186 @@
...
+        {
+            errorReports.add( new ErrorReport( ProgramStageWorkingList.class, ErrorCode.E4064, ouMode ) );
+        }
+    }
+
+    private void validateDataFilters( List<ErrorReport> errorReports, List<EventDataFilter> dataFilters )
+    {
+        if ( !CollectionUtils.isEmpty( dataFilters ) )
+        {
+            dataFilters.forEach( avf -> {
```
teleivo, line 142: What does `avf` stand for? Wondering if that is a typo or intentional. EventDataFilter -> `evf`? Just `f` is enough as well I think 😄

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHookTest.java
```diff
@@ -0,0 +1,325 @@
...
+    void shouldFailWhenDatePeriodAbsoluteAndNoStartNorEndDateSpecifiedForEnrolledAt()
+    {
+        ProgramStageQueryCriteria queryCriteria = ProgramStageQueryCriteria.builder()
+            .enrolledAt( createDatePeriod( null, null, DatePeriodType.ABSOLUTE ) )
+            .build();
+        programStageWorkingList.setProgramStageQueryCriteria( queryCriteria );
+
+        List<ErrorReport> errorReports = workingListHook.validate( programStageWorkingList, bundle );
+
+        assertEquals( 1, errorReports.size() );
```
teleivo, line 136: I don't think we should use this pattern 😅 As explained this leads to error messages that leave out important context. You will get `expected 2 to be 1` (or similar). You will not know whether the one you expected is in or not and what other errors we might have gotten. Tracker assertions https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/AssertValidations.java try to provide that context.

I would also try to avoid asserting on the entire error message. If someone just wants to fix a typo it would be great not having to fix tests as well.

If there is not yet a helper for asserting `ErrorReport` we should create one. For example

```java
    @Test
    void test() {
        ErrorReport error1 = new ErrorReport( OrganisationUnit.class, ErrorCode.E1102, "foo" );
        ErrorReport error2 = new ErrorReport( Period.class, ErrorCode.E1101, "bar" );
        List<ErrorReport> errors = List.of(error1, error2);

        assertErrorReport(errors, ErrorCode.E1102, "wrong arg"); // passes with 'foo'
    }

    private void assertErrorReport(List<ErrorReport> actual, ErrorCode expectedCode, String expectedMessage) {
        assertFalse(actual.isEmpty(), expectedCode + " not found as there are no errors");
        assertTrue( actual.stream().anyMatch( err -> expectedCode == err.getErrorCode() &&
                        err.getMessage().contains( expectedMessage )),
                () -> String.format( "error report with code %s and and (partial) message '%s' not found in %s", expectedCode, expectedMessage, actual ));
    }
```

gives an error like

```
org.opentest4j.AssertionFailedError: error report with code E1102 and and (partial) message 'wrong arg' not found in [ErrorReport{message=Organisation unit not found or not accessible: `foo`, errorCode=E1102, mainKlass=class org.hisp.dhis.organisationunit.OrganisationUnit, errorKlass=null, value=null}, ErrorReport{message=Period is invalid: `bar`, errorCode=E1101, mainKlass=class org.hisp.dhis.period.Period, errorKlass=null, value=null}] ==> 
Expected :true
Actual   :false
```

If you absolutely want to make sure there is only x amount of errors you can add that assertion into the helper wrapping the assertions that belong together in an `assertAll`. This way we get all the context we need 😄

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/descriptors/ProgramStageWorkingListSchemaDescriptor.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.hisp.dhis.schema.descriptors;
+
+import org.hisp.dhis.programstageworkinglist.ProgramStageWorkingList;
+import org.hisp.dhis.schema.Schema;
+import org.hisp.dhis.schema.SchemaDescriptor;
+import org.hisp.dhis.security.Authority;
+import org.hisp.dhis.security.AuthorityType;
+
+import com.google.common.collect.Lists;
```
teleivo, line 36: use java lists see https://github.com/dhis2/wow-backend/blob/master/guides/java_coding_principles.md#java-coding-principles

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_12__Create_table_program_stage_working_list.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE if not exists programstageworkinglist (
+    programstageworkinglistid integer NOT NULL,
```
teleivo, line 2: should this maybe be a primary key https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/programstageworkinglist/hibernate/ProgramStageWorkingList.hbm.xml
```diff
@@ -0,0 +1,38 @@
...
+    <cache usage="read-write" />
+
+    <id name="id" column="programstageworkinglistid">
+      <generator class="native" />
+    </id>
+    &identifiableProperties;
+
+    <property name="name" column="name" not-null="true" length="230" />
+
+    <property name="description" />
```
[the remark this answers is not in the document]
larshelge: We haven't done this in the rest of the system. We use text in the database which is virtually unlimited ~2GB. We could add a limit but should probably be system wide/consistent.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_12__Create_table_program_stage_working_list.sql
larshelge: We use `constraint primary key` as part of the constraints further down, which is missing.

PR #12661, opened 2022-12-30
[WIP]fix: Significant query inefficiencies in the /api/analytics/events/query/ endpoint [2.40-DHIS2-14088]

d-bernat:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/DimensionalItemObjectCollection.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Container for dimension or filter based dimensional items.
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class DimensionalItemObjectCollection
```
larshelge, line 42: We already have a property `allItems` on the `DimensionalObject` object, and `DimensionalObject` already has a collection of `DimensionalItemObject`. Are you sure we need a new object here? Could we possibly use what we have instead of introducing a new object?

https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/DimensionalObject.java#L166

gnespolino (approved):

larshelge:
Closing due to inactivity, please open a new PR when ready :+1:

PR #12662, opened 2022-12-30
Dhis2 14417 tests

jason-p-pickering:
The added test fails on the current version of master because the resource tables timestamp is not updated.


jimgrace (approved):
Excellent additions. Thanks @jason-p-pickering

PR #12664, opened 2023-01-02
fix: Fix analytics test to use current year reference periods

enricocolasante:
Fix analytics test that was using static years and now that we are in 2023 it was not working anymore


jbee (approved):

maikelarabori (approved):
Thanks for fixing it @enricocolasante, looks good :+1: 

Cheers,
Maikel

PR #12683, opened 2023-01-05
fix: validate access of DashboardItem objects when update Dashboard

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14404

- In 2.39 and older versions, we added a loop hole to allow user to update a `Dashboard` even if that user doesn't have READ access to all `DashboardItem` associate objects such as `Visualization` or `Report`. https://github.com/dhis2/dhis2-core/pull/12453

- In 2.40 we don't allow that, instead this PR will throw an error for same use case.  So in order to update a `Dashboard` user need READ access to all objects associate with all `DashboardItem`.

### Test

1. Login with SuperUser.
2. Create a new UserA, user should not have `ALL` authority, but can read/update Dashboard.
3. Create a new dashboard `DashboardA`, add one Visualization to it.
4. In Visualization app, update sharing settings of Visualization in step 2 to make user UserA doesn't have METADATA_READ access to it.
5. Share `DashboardA` to UserA with METADATA_READ_WRITE permission.
6. Login with UserA.
7.  In Dashboard app, edit `DashboardA` , add one more Visualization to it.
8. Expect: an error is returned from server.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
@@ -245,6 +245,8 @@ public enum ErrorCode
 
     /* Metadata Validation (continued) */
     E4060( "Object could not be deleted: {0}" ),
+    E4061(
+        "DashboardItem `{0}` object reference `{1}` with id `{2}` not found or not accessible" ),
 
     /* SQL views */
     E4300( "SQL query is null" ),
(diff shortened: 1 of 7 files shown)


jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -245,6 +245,8 @@
 
     /* Metadata Validation (continued) */
     E4060( "Object could not be deleted: {0}" ),
+    E4061(
+        "DashboardItem `{0}` has invalid reference `{1}` with id `{2}`. You might not have READ permission to this object." ),
```
larshelge, line 249: To make it more consistent with existing error codes use something like:

"DashboardItem `{0}` object reference `{1}` with id `{2}` not found or not accessible"

vietnguyen: fixed

larshelge (approved):

PR #12699, opened 2023-01-06
feat: Do not run slow metadata checks by default

jason-p-pickering:
https://dhis2.atlassian.net/browse/DHIS2-14446


jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/DataIntegrityCheck.java
```diff
@@ -78,6 +78,9 @@ public final class DataIntegrityCheck implements Serializable
     @JsonProperty
     private final String issuesIdType;
 
+    @JsonProperty
+    private final Boolean isSlow;
```
jbee, line 82: I think this is better made a `boolean`

jimgrace (approved):
Looks reasonable to me

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/DataIntegrityController.java
```diff
@@ -84,6 +81,21 @@ public WebMessage runDataIntegrity(
...
+                .getDataIntegrityChecks() )
+            {
+                if ( !e.isSlow() )
+                {
+                    String name = e.getName();
+                    checks.add( name );
+                }
+            }
+        }
+
```
jbee, line 98: Sorry, gave you a bad pointer there. Remembered now that the logic is different. The controller and the job configuration just use the empty list to mean "all" (so far). The place where we should filter this is here https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java#L939, filter the returned set by the `isSlow` (please use streams :D)

jbee (changes requested):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
```diff
@@ -963,6 +963,18 @@ private Set<String> expandChecks( Set<String> names )
         return expanded;
     }
 
+    private Set<String> getDefaultChecks()
+    {
+        ensureConfigurationsAreLoaded();
+
+        return checksByName.entrySet()
```
jbee, line 971: Since you never need the key of the map you should directly stream the `values()`. This will also allow to `filter(not(DataIntegrityCheck::isSlow))` and `map(DataIntegrityCheck::getName)` like that. (`not` is on `Predicate`)

jason-p-pickering: Thanks for the tip. Fixed now.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/DataIntegrityController.java
jason-p-pickering: I think its in the right place now.

PR #12710, opened 2023-01-09
Add metadata check for data elements in datasets with different period types

jason-p-pickering:
Data elements should not belong to datasets with different period types.

This SQL based check replicates the Java based check.


jbee (approved):
Nice!

olavpo (approved):

PR #12666, opened 2023-01-02
fix: TEIs are not filterable by potentialDuplicate property (NTI) [ DHIS2-14362]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/12392

PR #12720, opened 2023-01-10
fix: TEIs are not filterable by potentialDuplicate property [DHIS2-14362] [2.39]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/12666

### DIFF
--- a/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
+++ b/dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.tracker.importer;
 
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-e2e-test/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
```diff
@@ -284,4 +292,52 @@ public void shouldReturnAscOrderedEventByTEIAttribute()
         List<String> events = response.extractList( "instances.event.flatten()" );
         assertEquals( List.of( "ZwwuwNp6gVd", "olfXZzSGacW" ), events, "Events are not in the correct order" );
     }
+
+    @Test
+    void getTeiByPotentialDuplicateParamNull()
```
muilpp, line 297: We agreed to use the test name convention: _shouldXXWhenYYY_, is there a reason not to use something like `shouldReturnPotentialDuplicatePropertyWhenParamNull` in this method? 

The same applies for the ones below.

lucaCambi77: yes, I know but that is before we agreed on the naming convention and in master, it is already merged that way. So I'll leave it like this for now

muilpp (approved):

PR #12724, opened 2023-01-10
feat: data integrity check results with start time

jbee:
### Summary
* adds a `startTime` to check results (so a duration can be computed)
* fixes `finishedTime` to be actually after the check is performed in some of the paths
* fixes `error` to always hold a message even if the exception message was `null`
* logs any exceptions thrown by a check as error

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/DataIntegrityDetails.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/DataIntegrityDetails.java
@@ -56,6 +56,9 @@ public class DataIntegrityDetails implements Serializable
     private final DataIntegrityCheck check;
 
     @JsonProperty
+    private final Date startTime;
+
+    @JsonProperty
     private final Date finishedTime;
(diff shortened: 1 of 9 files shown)


jason-p-pickering (commented):
Looks good. Tested locally and the changes do exactly what is advertised. I triggered a check which is known to throw an error, and got this response which includes the new field start time as well as the error message
```
    "program_indicators_with_invalid_expressions": {
        "name": "program_indicators_with_invalid_expressions",
        "displayName": "Program indicators with invalid expression",
        "section": "Program Indicators",
        "severity": "WARNING",
        "description": "Lists all program indicators that have an invalid expression",
        "recommendation": "Check that the expression evaluates to a number",
        "issuesIdType": "programIndicators",
        "isSlow": false,
        "startTime": "2023-01-11T06:12:56.436",
        "finishedTime": "2023-01-11T06:12:57.021",
        "error": "Check failed because an exception was thrown: null",
        "count": -1
    }
}

jason-p-pickering (approved):

larshelge (approved):

PR #12728, opened 2023-01-11
feat: [DHIS2-14300] Inform client that breaking the glass is permitted

superskip:
Makes `tracker/trackedEntities/<uid>?program=<uid>` emit a special error message when the specified program is protected AND `programOwner` lies in the user's {search scope} minus {capture scope}.

The reason for making this into a special case is that it helps clients figure out if breaking the glass is applicable or not.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntitiesSupportService.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntitiesSupportService.java
@@ -108,11 +108,21 @@ class TrackedEntitiesSupportService
 
             if ( !errors.isEmpty() )
             {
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntitiesSupportService.java
```diff
@@ -143,4 +130,33 @@ public TrackedEntityInstance getTrackedEntityInstance( String id, String pr, Lis
 
         return trackedEntityInstance;
     }
+
+    @SneakyThrows
+    private void testAccessToOwningOrgUnit( User user, TrackedEntityInstance trackedEntityInstance, Program program )
```
larshelge, line 135: Would it be good to add some additional unit tests for this logic?

PR #12741, opened 2023-01-12
fix: Locale part of cache key for analytics queries [DHIS2-14242]

larshelge:
Includes the database locale of the current user, if set, to the analytics cache key. This is to ensure that a user can get a cache hit from a previous request which includes metadata in a different locale/language than the user itself.


maikelarabori (approved):

maikelarabori:
Looks good, some tests failing though.

PR #12753, opened 2023-01-14
fix: Predictor with integer P.I.s DHIS2-14499

jimgrace:
See [DHIS2-14499](https://dhis2.atlassian.net/browse/DHIS2-14499). Some analytics objects such as some program indicators now return `Long` instead of `Double`. The expression parser expects all numeric values to be `Double`, so predictors were breaking.

The fix is to make sure all analytics values are processed in predictor expressions as `Double`.

[DHIS2-14499]: https://dhis2.atlassian.net/browse/DHIS2-14499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-reporting/src/test/java/org/hisp/dhis/predictor/PredictionAnalyticsDataFetcherTest.java
+++ b/dhis-2/dhis-services/dhis-service-reporting/src/test/java/org/hisp/dhis/predictor/PredictionAnalyticsDataFetcherTest.java
@@ -179,7 +179,7 @@ class PredictionAnalyticsDataFetcherTest
             .addValue( programIndicatorB.getUid() )
             .addValue( orgUnitA.getUid() )
             .addValue( aocC.getUid() )
-            .addValue( 20 ); // Note: boxed as Long, not Double
+            .addValue( 20L ); // Note: boxed as Long, not Double
 
         aocGrid.addRow()
             .addValue( periodB.getIsoDate() )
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

maikelarabori (approved):

dhis-2/dhis-services/dhis-service-reporting/src/test/java/org/hisp/dhis/predictor/PredictionAnalyticsDataFetcherTest.java
```diff
@@ -179,7 +179,7 @@ void testGetValues()
             .addValue( programIndicatorB.getUid() )
             .addValue( orgUnitA.getUid() )
             .addValue( aocC.getUid() )
-            .addValue( 20.0 );
+            .addValue( 20 ); // Note: boxed as Long, not Double
```
jbee, line 182: If this should be a `Long` it needs to be `20L`, `20` is an `Integer`. For the sake of the test this might not make a difference but the comment becomes confusing.

d-bernat (approved):

dhis-2/dhis-services/dhis-service-reporting/src/test/java/org/hisp/dhis/predictor/PredictionAnalyticsDataFetcherTest.java
jimgrace: Nice catch, fixed!

PR #12755, opened 2023-01-16
fix: P.I. custom aggregation func desc DHIS2-14500

jimgrace:
See [DHIS2-14500](https://dhis2.atlassian.net/browse/DHIS2-14500). When using a custom aggregation function in a program indicator expression, the code that validates it was getting a NPE.

The fix is to use default parameters instead of `null` inside `DefaultProgramIndicatorService.validate`.

[DHIS2-14500]: https://dhis2.atlassian.net/browse/DHIS2-14500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
@@ -66,6 +66,8 @@ import org.hisp.dhis.program.Program;
 @Builder( toBuilder = true )
 public class ExpressionParams
 {
+    public static final ExpressionParams DEFAULT_EXPRESSION_PARAMS;
+
     /**
      * Dummy data for sample periods, so in the absence of real sampled data the
      * parser will still traverse the contents of aggregation functions once for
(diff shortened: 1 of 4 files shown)


enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
```diff
@@ -66,6 +66,8 @@
 @Builder( toBuilder = true )
 public class ExpressionParams
 {
+    public static final ExpressionParams DEFAULT_EXPRESSION_PARAMS;
```
jbee, line 69: Are `ExpressionParams` truly immutable? I got the impression they have mutable state that changes during traversal. In that case a constant makes little sense and can cause different runs to affect each other.

stian-sandvold (approved):

d-bernat (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expression/ExpressionParams.java
jimgrace: Yes, `ExpressionParams` are truly immutable. Every field except one is final and there is no good reason for that -- I will fix that in another PR. Once the instance is created all the collections are never modified, only read. (It would be nice if one could declare a collection in an object like this as unmodifiable -- do you know any reasonable way to do this?)

`ExpressionState` on the other hand is mutable and holds the current parsing state, but `ExpressionParams` describes the parameters used to invoke expression processing and is immutable.

PR #12760, opened 2023-01-16
fix: data integrity test setup so it is execution order independent

jbee:
The issue with the test setup is that the endpoint cannot be tested in isolation as long as the same spring application context is used for multiple tests. Therefore the expectation here needed to be relaxed to just check that at least the check that the setup did run must be in the completed set but there may be other checks from other tests too.


vietnguyen (approved):

stian-sandvold (approved):
Good job!

Philip-Larsen-Donnelly (approved):

PR #12764, opened 2023-01-16
feat: Data Integrity - Support long lists of checks in the API [DHIS2-14506]

jason-p-pickering:
### Summary
Listing all the checks by name (their ID) in a request parameter could soon exceed URL length limits.
On the other hand checks should be sort of "human readable". 
Especially since their name is used as the base for the fallback display name.

To address the issue each check was also given a code. The code is not defined but derived from the name
by joining the first letters of each word in a checks name. Also the letters are upper-cased.
For example the check name `categories_no_options` has the code `CNO`. 
The resulting codes are also checked to be unique in tests.

In the API all input for names will also accept codes. Names and codes can be mixed freely.
Codes will be expanded/replaced with the corresponding name before checks are run in the service layer.

**Additional Changes**
- `*` expansion of name patterns will not include _slow_ checks
- POST request to trigger checks may use either the `checks` request parameter to list checks to perform or use a request body. If both are present the body takes precedence 
- Better standardization of the names of checks
- Cleaned up `details_id_type`which can be used to link to objects identified by the checks
- Added some `summary_id` and `details_id` which will be used for external reports

**Cleanup**
- The normalization of names only takes place in the service layer (no longer in controller)
- `@Nonnull` annotations were added to the `DataIntegrityService` to avoid null collection issues since controller collections can sometimes be null

### Automatic Testing
New tests were added to test the POST with body endpoints as well as testing using codes.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/DataIntegrityCheckType.java
```diff
@@ -92,7 +92,7 @@
     PROGRAM_RULE_ACTIONS_WITHOUT_DATA_OBJECT,
     PROGRAM_RULE_ACTIONS_WITHOUT_NOTIFICATION,
     PROGRAM_RULE_ACTIONS_WITHOUT_SECTION,
-    PROGRAM_RULE_ACTIONS_WITHOUT_STAGE;
+    PROGRAM_RULE_ACTIONS_WITHOUT_STAGE_ID;
```
jbee, line 95: Had to rename this to make it's code unique. I did clash with the one above. We will remove these later anyhow so this is just to make the uniqueness tests pass until these get removed.

jbee:
This waits for the changes to the YAML definitions to be added.

larshelge (approved):

PR #12777, opened 2023-01-18
refactor: Improve tracker program rule integration test [DHIS2-14408]

enricocolasante:
- Clean up program rule integration test in tracker
- Add more test cases to support following refactoring
- Add warning assertions in Assertions class
- Move all relevant test resources to programrule folder


stian-sandvold (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -91,6 +91,20 @@ public static void assertHasOnlyErrors( ValidationReport report, ValidationCode.
...
     }
 
+    /**
+     * assertHasWarnings asserts the report contains only given warnings in any
+     * order.
+     *
+     * @param report import report to be asserted on
+     * @param warnings expected warnings
+     */
+    public static void assertHasOnlyWarnings( ImportReport report, Warning... warnings )
```
muilpp, line 101: Does this method mean "has only warnings" as in it has no errors? or does it mean it has only the warnings passed as parameter?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleTest.java
```diff
@@ -265,129 +269,265 @@ void shouldNotImportWhenErrorIsTriggeredOnAllEntities()
...
-        programRuleActionService.addProgramRuleAction( programRuleActionShowError );
-        programRuleA.getProgramRuleActions().add( programRuleActionShowError );
-        programRuleService.updateProgramRule( programRuleA );
+        assignProgramRule();
+        systemSettingManager.saveSystemSetting( SettingKey.RULE_ENGINE_ASSIGN_OVERWRITE, true );
+        TrackerImportParams params = fromJson( "tracker/programrule/tei_enrollment_completed_event.json" );
+
+        ImportReport importReport = trackerImportService.importTracker( params );
+
+        assertNoErrors( importReport );
```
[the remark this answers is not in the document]
enricocolasante: This is something that I am not sure how to tackle.
I need to import something as a setup, and I need to be sure it was correctly imported before I can really test.
Without this assertion, I will probably have the next one failing but it will be quite difficult to understand what went wrong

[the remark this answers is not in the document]
muilpp: If I'm not mistaken, there are three tests in this class where this strategy is used, and in all of them, the same .json is imported. 
I think we could have a separate test to validate the import is correct, and then in the rest of the tests we assume it works. If not, since then the error message would not be clear, we could have a separate test class and have the json imported during set up.
I mean it, because I think we should aim to assert only one thing, and at the end of the test, asserting in the middle it usually means it's a validation not related to the test. And also in this case, we validate the same in three different tests. Do you agree?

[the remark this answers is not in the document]
enricocolasante: Moved to a new test class.
Now it seems much more cleaner
Thank you :)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -91,16 +91,32 @@ public static void assertHasOnlyErrors( ValidationReport report, ValidationCode.
         assertHasErrors( report, codes.length, codes );
     }
 
+    /**
+     * assertHasWarnings asserts the report contains only given warnings in any
```
teleivo, line 95: Please adapt the javadoc to `assertHasOnlyWarningsAndNoErrors` and its description. Also the overloaded method below.

muilpp (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -91,16 +91,32 @@ public static void assertHasOnlyErrors( ValidationReport report, ValidationCode.
...
+     * order.
+     *
+     * @param report import report to be asserted on
+     * @param warnings expected warnings
+     */
+    public static void assertHasOnlyWarningsAndNoErrors( ImportReport report, Warning... warnings )
+    {
+        assertAll(
```
teleivo, line 103: I think you can just delegate to the `assertHasOnlyWarningsAndNoErrors` taking a `ValidationReport` by calling `report.getValidationReport()`. I used the pattern in the other assertions.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -91,16 +91,32 @@ public static void assertHasOnlyErrors( ValidationReport report, ValidationCode.
...
+    }
+
     /**
      * assertHasWarnings asserts the report contains only given warnings in any
      * order.
      *
      * @param report validation report to be asserted on
      * @param warnings expected warnings
      */
-    public static void assertHasOnlyWarnings( ValidationReport report, Warning... warnings )
```
teleivo, line 101: I would like for us to keep the symmetry we had in the API with regards to error/warning assertions.

Like

```
assertHasOnlyErrors
assertHasOnlyWarnings
```

So if you want to create a new assertion based on the above like `assertHasOnlyWarningsAndNoErrors` I would still keep the lower level ones for error/warning.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -0,0 +1,181 @@
...
+        ProgramRuleVariable programRuleVariable = createProgramRuleVariableWithDataElement( 'A', program,
+            dataElement2 );
+        programRuleVariableService.addProgramRuleVariable( programRuleVariable );
+
+        injectAdminUser();
+
+        assignProgramRule();
+        TrackerImportParams params = fromJson( "tracker/programrule/tei_enrollment_completed_event.json" );
+
+        trackerImportService.importTracker( params );
```
teleivo, line 102: If we don't use/adjust params I would always inline the `fromJson` call. Our tests usually have enough lines of code to read through 😅 

```java
trackerImportService.importTracker( fromJson( "tracker/programrule/tei_enrollment_completed_event.json" ) );
```

So if you agree, please do so in all tests 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -0,0 +1,181 @@
...
+    @Test
+    void shouldImportWithWarningWhenDataElementWithSameValueIsAssignedByAssignRule()
+        throws IOException
+    {
+        TrackerImportParams params = fromJson( "tracker/programrule/event_update_datavalue_same_value.json" );
+        params.setImportStrategy( TrackerImportStrategy.CREATE_AND_UPDATE );
+
+        ImportReport importReport = trackerImportService.importTracker( params );
+
+        Assertions.assertHasOnlyWarningsAndNoErrors( importReport.getValidationReport(),
```
teleivo, line 114: I think we have variants of the assertions where we just pass in the ImportReport. If it does not exist please create it as the other assertions have it.

So we don't have to do the `importReport.getValidationReport()` over and over. Please use that one in all tests 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -0,0 +1,181 @@
...
+        ImportReport importReport = trackerImportService.importTracker( params );
+
+        assertHasOnlyErrors( importReport, E1307 );
+    }
+
+    @Test
+    void shouldImportWithWarningWhenDataElementWithDifferentValueIsAssignedByAssignRuleAndOverwriteKeyIsTrue()
+        throws IOException
+    {
+        systemSettingManager.saveSystemSetting( SettingKey.RULE_ENGINE_ASSIGN_OVERWRITE, true );
```
teleivo, line 136: Do we need to "cleanup" this system setting change? So we don't run into a flaky test in the future.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -151,7 +167,7 @@ public static void assertHasErrors( ValidationReport report, int count, Validati
      * assertHasWarnings asserts the report contains given count of warnings and
      * warnings that contain given code and that are linked to given tracker
      * object in any order. <em>NOTE:</em> prefer
-     * {@link #assertHasOnlyWarnings(ValidationReport, Warning...)}
+     * {@link #assertHasOnlyWarningsAndNoErrors(ValidationReport, Warning...)}
```
teleivo, line 170: something seems of in the javadoc 🤔 are there too many parameters in the `@link`? I assume

`{@link #assertHasOnlyWarningsAndNoErrors(ImportReport, Warning...)}`

is what I meant

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -210,9 +226,11 @@ public static void assertHasWarning( ValidationReport report, Warning warning )
     {
         assertTrue( report.hasWarnings(), "warning not found since report has no warnings" );
         assertTrue( report.hasWarning( w -> Objects.equals( warning.getWarningCode(), w.getWarningCode() ) &&
-            Objects.equals( warning.getUid(), w.getUid() ) ),
-            String.format( "warning with code %s for object %s not found in report with warning(s) %s",
-                warning.getWarningCode(), warning.getUid(), report.getWarnings() ) );
+            Objects.equals( warning.getUid(), w.getUid() ) &&
+            w.getWarningMessage().contains( warning.getWarningMessage() ) ),
```
teleivo, line 230: This also adds asymmetry to our API. We are now asserting on the warning message in `assertHasWarning` but not the error message in `assertHasError`.

teleivo (commented):
Nice, tests are pretty readable given the complexity of program rules 😁 

Please use a static import for all assertions 😄 
There is some asymmetry now in the API of our assertions which I'd like to straighten.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/programrule/ProgramRuleAssignActionTest.java
enricocolasante: Not really, this test class is extending TrackerTest that is extending SingleSetupIntegrationTestBase that is running tests inside a transaction. Enything changed inside a test is rollbacked when the test is done

PR #12778, opened 2023-01-18
fix: Allow ordering TEIs by attribute [DHIS2-13981] [2.39]

muilpp:
Fix order field supported values for /trackedEntityInstanceFilters endpoint


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -99,6 +103,7 @@
 import org.springframework.stereotype.Repository;
 
 import com.google.common.collect.Lists;
+import io.debezium.util.Strings;
```
teleivo, line 106: same here please use `String.join`

teleivo (changes requested):

teleivo (approved):

PR #12780, opened 2023-01-18
fix: Allow ordering TEIs by attribute [DHIS2-13981] [2.38]

muilpp:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityInstanceStore.java
```diff
@@ -97,6 +101,7 @@
 import org.springframework.stereotype.Repository;
 
 import com.google.common.collect.Lists;
+import io.debezium.util.Strings;
```
teleivo, line 104: I don't think we want to use this one. I guess you meant `String.join`. Please check the other PRs. I might have missed it 😅

teleivo (changes requested):

teleivo (approved):

PR #12786, opened 2023-01-19
fix: Add missing mock dependency to fix 4 failing tests

eternalgooner:
- I pulled the latest master branch and tried building the project
- The build failed at the dhis-service-tracker module step
- I investigated this module's failing tests and it seems that 2 test classes (4 tests) were missing a mock dependency (NPEs being thrown as class member not initialised)
- dhis-service-tracker module back passing now :)


larshelge:
Thanks for the contribution :+1:

PR #12791, opened 2023-01-20
fix: Validate query filters for item value type [DHIS2-12152]

larshelge:
Adds validation for query item filters according to the value type of the query item. Previously, e.g. text values could be supplied as filters for integer type data elements, leading to a SQL exception when trying to compare a integer type analytics database column against a text value.


maikelarabori (approved):

maikelarabori:
Looks good, some integration tests are failing.

d-bernat (approved):

PR #12795, opened 2023-01-20
feat: Update client errors API response [DHIS2-13733]

muilpp:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BadRequestException.java
```diff
@@ -0,0 +1,47 @@
...
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.common;
+
+/**
+ * Exception representing a bad request query.
```
teleivo, line 31:
```suggestion
 * Exception representing a bad request.
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ForbiddenException.java
```diff
@@ -0,0 +1,47 @@
...
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.common;
+
+/**
+ * Exception representing a forbidden query.
```
teleivo, line 31:
```suggestion
 * Exception representing a forbidden request.
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -136,16 +135,11 @@ protected void initBinder( WebDataBinder binder )
         this.enumClasses.forEach( c -> binder.registerCustomEditor( c, new ConvertEnum( c ) ) );
     }
 
-    @ExceptionHandler( org.hisp.dhis.feedback.BadRequestException.class )
```
teleivo, line 139: I don't think we should do that. AFAIK BadRequestException/ForbiddenException were recently introduced by @jbee. Is there any issue with us simply using BadRequestException/ForbiddenException from `org.hisp.dhis.feedback`?

We want to return HTTP status 400/403 by throwing an exception in our controllers. If I understand correctly it would also have the advantage of showing up in the OpenAPI spec.

teleivo (changes requested):
I am requesting changes just as a precaution so we don't merge this change before we resolved the CrudControllerAdvice changes.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
muilpp: You are right, for some reason I thought I created that method myself in a previous commit, I just rolled it back to the way it was. 

The thing about using `org.hisp.dhis.feedback.BadRequestException` is that we have to add the exception to the method signature, whilst using org.hisp.dhis.common.BadRequestException we don't. Since it's used in a lot of places, I thought we'd be better off that way, but if we favor OpenAPI, I'll change it.

larshelge (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
teleivo: AFAIK making it a checked exception was intentional as it allows us to extract that information and put it into the OpenAPI spec. Right @jbee ?

jbee: Making these checked was intentional. By forcing them to be listed in the `throws` they propagate to all the places that somehow might call that code. This makes sure all exceptions that go along with a specific HTTP status code are aggregated up to the controller endpoint method. 

By making them checked you cannot miss an intentional error status code an endpoint might have as the compiler forces you to either propagate or handle. This helps OpenAPI but I'd argue it also helps the programmer to understand directly from the method signatures what error they might cause, or if you don't want this to be exposed in the REST API that a try-catch at some level would be required. 

It also helps to stay true to a contract that previously was only encoded implicitly or by convention. For example the pre- and post- hook handlers should never intentionally throw something that not ends up as a `Conflict` status. Now this contract is clearly stated by those methods only throwing a `ConflictException`.  

Having said all that this does not mean you should never use unchecked exceptions. It just means failures that you want to propagate to the REST API should use these checked exceptions as they are part of the API. This includes the code that throws them. Classic example would be validation.

jbee (changes requested):
As @teleivo said please do not introduce another handful of exceptions that correspond to status codes. Use the ones specifically added for that purpose and propagate them in the `throws` clause of the methods. Yes, that might affect many methods but it only makes visible how far reaching consequences it has if these methods throw such an exception.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
```diff
@@ -195,4 +205,10 @@ public static QueryItem parseQueryItem( String item, Function<String, QueryItem>
         return queryItem;
     }
 
+    @FunctionalInterface
+    public interface CheckedFunction<T, R>
```
teleivo, line 209: @jbee can you please help us find a good home for this 😅 I think this could be helpful for others 😄

muilpp: in that case we should move it to another package too

jbee: it should be in the `dhis-api` module. I'd pick the `org.hisp.dhis.util` package.

jbee (changes requested):
Some cleanup left to fully use the existing exceptions in `feedback`.

I think the reason that the extra interface with checked exception is needed here is because the service does not use a clear sequence of phases each having one task:

1. parse (if needed)
2. validate
3. process

Instead these 3 are intertwined and happen at the same time in some loop for each item.
When these are split up you can also use streams again without issue because the throwing of an exception is never within a stream. It is either as a consequence of a stream resulting in "false" (parse/validate phase) or you are in process knowing everything is valid and no exception should be thrown. 

I am not saying you should change the code now. I understand that we also want to keep changes small. Just pointing this out for the future.

muilpp:
> Some cleanup left to fully use the existing exceptions in `feedback`.
> 
> I think the reason that the extra interface with checked exception is needed here is because the service does not use a clear sequence of phases each having one task:
> 
>     1. parse (if needed)
> 
>     2. validate
> 
>     3. process
> 
> 
> Instead these 3 are intertwined and happen at the same time in some loop for each item. When these are split up you can also use streams again without issue because the throwing of an exception is never within a stream. It is either as a consequence of a stream resulting in "false" (parse/validate phase) or you are in process knowing everything is valid and no exception should be thrown.
> 
> I am not saying you should change the code now. I understand that we also want to keep changes small. Just pointing this out for the future.

I agree this needs refactor, but I think it would be out of the scope of this ticket, I'll create a TECH ticket to tackle this in the future.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -136,18 +135,6 @@ protected void initBinder( WebDataBinder binder )
...
-    public WebMessage badRequestException( org.hisp.dhis.feedback.BadRequestException ex )
-    {
-        WebMessage message = badRequest( ex.getMessage(), ex.getCode() );
-        if ( !ex.getErrorReports().isEmpty() )
-        {
-            message.setResponse( new ErrorReportsWebMessageResponse( ex.getErrorReports() ) );
-        }
-        return message;
-    }
-
```
jbee, line 150: This needs to stay as the other handler below will not add the error reports like this does.

muilpp: Then I guess we need it for a ForbiddenException too, right?

jbee: The https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ForbiddenException.java cannot (yet) have error reports attached to it so currently it does not have this.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
muilpp: That's what I meant, is there a reason why ForbiddenException doesn't generate the report?

jbee: Oh, I added just what was needed to that point. If you want a `ForbiddenException` with a report you can add the field and a similar mapping to the `WebMessage` like I did for `BadRequestException`. But usually a forbidden does not need much of a detail that would come with a report. I guess that is why it wasn't needed so far.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/CheckedFunction.java
```diff
@@ -0,0 +1,37 @@
...
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.util;
+
+import org.hisp.dhis.feedback.BadRequestException;
+
+@FunctionalInterface
+public interface CheckedFunction<T, R>
```
teleivo, line 33: The prefix `Checked` feels a little too generic but I can't think of a better one 😅 I think we'll deal with it once we need another one of these for another Exception.

teleivo (approved):

PR #12807, opened 2023-01-23
fix: Scheduler - reliably remove jobs from schedule (2.38)

jbee:
backport from cherry-pick #12805 with manual merge conflict resolve.

Apart from simple conflicts due to changes on similar lines in `ErrorCode` and `JobConfiguration` the main cause of conflict was that 2.38 does not have the most recent `JobProgress` API with `FailurePolicy`. We opted to not backport this as many files were changed on the usage site of that API. So I had to remove some code related to the `FailurePolicy`.


larshelge:
Tests failing @jbee .

PR #12814, opened 2023-01-24
chore: remove sharing backward compatibility code

vietnguyen:


larshelge:
Some tests failing @vietnguyen .

PR #12817, opened 2023-01-24
fix: Check if dynamic property when sorting by attribute [DHIS2-14560]

muilpp:


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityInstanceService.java
```diff
@@ -237,32 +238,33 @@ public List<Long> getTrackedEntityInstanceIds( TrackedEntityInstanceQueryParams
...
+            .filter( DefaultTrackedEntityInstanceService::isDynamicColumn )
+            .map( attributeService::getTrackedEntityAttribute )
             .collect( Collectors.toList() );
 
         params.addAttributesIfNotExist( QueryItem.getQueryItems( sortAttributes ).stream()
             .filter( sAtt -> !params.getFilters().contains( sAtt ) ).collect( Collectors.toList() ) );
     }
 
+    public static boolean isDynamicColumn( String propName )
+    {
```
lucaCambi77, line 263: i think we can avoid static here?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/MvcTestConfig.java
```diff
@@ -232,6 +234,12 @@ public void addInterceptors( InterceptorRegistry registry )
         registry.addInterceptor( new RequestInfoInterceptor( requestInfoService ) );
     }
 
+    @Override
+    public void addFormatters( FormatterRegistry registry )
+    {
+        registry.addConverter( new StringToOrderCriteriaListConverter() );
+    }
```
lucaCambi77, line 241: do we need this right now? I don't see any web MVC test. Maybe we can add it when it is required. Maybe i am wrong

lucaCambi77 (approved):

PR #12819, opened 2023-01-24
refactor: Refactor assign rule action execution in rule engine step [DHIS2-14408]

enricocolasante:
- Create interface `AssignValueExecutor` to allow remove duplication of `isEqual` method
- Refactor `isEqual` test to parameterized test


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/AssignValueExecutor.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return true if the values are equal, false if not.
+     */
+    default boolean isEqual( String value1, String value2, ValueType valueType )
+    {
+        if ( valueType.isNumeric() )
+        {
+            return NumberUtils.isParsable( value1 ) && NumberUtils.isParsable( value2 ) &&
+                MathUtils.isEqual( Double.parseDouble( value1 ), Double.parseDouble( value2 ) );
+        }
+        else
```
teleivo, line 61: I would remove the else clause and simply return. Less verbose 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/AssignValueExecutor.java
```diff
@@ -0,0 +1,66 @@
...
+import org.hisp.dhis.common.ValueType;
+import org.hisp.dhis.system.util.MathUtils;
+
+/**
+ * This executor assigns a value to a field if it is empty, otherwise returns an
+ * error
+ *
+ * @Author Enrico Colasante
+ */
+public interface AssignValueExecutor<T> extends RuleActionExecutor<T>
```
teleivo, line 42: An alternative to an interface with a default method is that `boolean isEqual( String value1, String value2, ValueType valueType )` which is just a function is declared as such.

Not sure if there is value in making `isEqual` part of the RuleActionExecutor (hierarchy) 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/implementers/AssignValueExecutor.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param value1 the first value.
+     * @param value2 the second value.
+     * @param valueType the value type.
+     * @return true if the values are equal, false if not.
+     */
+    default boolean isEqual( String value1, String value2, ValueType valueType )
+    {
+        if ( valueType.isNumeric() )
+        {
+            return NumberUtils.isParsable( value1 ) && NumberUtils.isParsable( value2 ) &&
```
teleivo, line 58: Github mentions NumberFormatException that we are not catching. Can/should we trust the values to be perfectly parseable numbers?

enricocolasante: We actually don't and this is why we call the `isParsable` method from `NumberUtils`

teleivo (approved):

PR #12832, opened 2023-01-25
fix: Analytics event query validation in filter [DHIS2-14561]

maikelarabori:
Fixes related to the event query filter param.
Recently, we introduced a validation for the filters based on the related `ValueType`.

This was too restrictive as clients might want to use operators like `!=`,  `like`, `not like`, `not null`, etc. We started to face validation errors for common filters that had valid comparisons.

This PR aims to make the validation more "relaxed", so filters that are comparable are always permitted. This will allow a broader set of possible comparisons and at the same time disallow invalid comparisons for the most restrictive types.
The idea is to have a more balanced validation.

I also took some time for small code improvements/clean-up.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/util/ValidationUtilsTest.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/util/ValidationUtilsTest.java
@@ -27,6 +27,30 @@
  */
 package org.hisp.dhis.system.util;
 
+import static org.hisp.dhis.common.ValueType.BOOLEAN;
+import static org.hisp.dhis.common.ValueType.COORDINATE;
+import static org.hisp.dhis.common.ValueType.DATE;
+import static org.hisp.dhis.common.ValueType.DATETIME;
+import static org.hisp.dhis.common.ValueType.EMAIL;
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/util/ValidationUtilsTest.java
```diff
@@ -186,50 +212,313 @@ void testUsernameIsValid()
...
+    void testValueIsComparable()
+    {
+        testIntegerTypes();
+        testDoubleTypes();
+        testBooleanTypes();
+        testDateTimeTypes();
+        testStringTypes();
+    }
+
+    private void testIntegerTypes()
```
larshelge, line 234: I would say put test annotations on each method, makes it easier to identify which test fails.

larshelge:
Nice work, thanks for fixing my naive implementation.

dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/util/ValidationUtilsTest.java
maikelarabori: Done @larshelge

PR #12835, opened 2023-01-26
fix: Check dynamic properties sorting by attribute

stian-sandvold:


muilpp:
Will be merged in https://github.com/dhis2/dhis2-core/pull/12849

PR #12837, opened 2023-01-26
fix: Remove data values with null value from rule engine context [DHIS2-13979]

enricocolasante:
When preparing the context for rule engine, null values for data values were replaced by default value.
This was introducing a bug as `d2:hasValue()` was not working properly for data elements with null value


larshelge (approved):

zubaira (approved):

zubaira:
> When preparing the context for rule engine, null values for data values were replaced by default value. This was introducing a bug as `d2:hasValue()` was not working properly for data elements with null value

if the expression description is working fine then it all looks good to me.

enricocolasante:
> > When preparing the context for rule engine, null values for data values were replaced by default value. This was introducing a bug as `d2:hasValue()` was not working properly for data elements with null value
> 
> if the expression description is working fine then it all looks good to me.

Expression description is based on `ProgramRuleVariable`s and it is not influenced by the actual value present in the data value, so it will work fine

PR #12838, opened 2023-01-26
feat: OptionSets persistent description [DHIS2-14102]

jbee:


mortenoh (approved):

vietnguyen (approved):

vietnguyen:
Please add below method if it need to be translatable 

```
@Override
@JsonProperty
@JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
@Translatable( propertyName = "description", key = "DESCRIPTION" )
public String getDisplayDescription()
{
    return getTranslation( "DESCRIPTION", getDescription() );
}
```

jbee:
I didn't add such translation to any of the previous description fields I added and the ticket does not specifically ask for it. I think we wait until this is an explicit request and then maybe bulk add this to all descriptions that don't have it yet.

PR #12842, opened 2023-01-27
ci: automate local jib builds for mac arm architecture

teleivo:
This PR gives devs on a new Mac (with arm64 CPU architecture) the same developer experience as devs on devices with amd64 architecture.

This is done by

1. automatically building the Docker image locally using the appropriate CPU architecture when using `./dhis-2/build-dev.sh`
2. building and pushing multi-architecture Docker images for every PRs to `dhis2/core-pr:<PR-number>`. No need to add a label to the PR 🎉 

## How

1. activate appropriate maven jib profile based on CPU architecture in bash. Choosing this approach over automatically activating a maven profile based on the architecture. This is to not risk running both the jibBuild and jibBuildArmOnly profiles in our pipelines. There might be ways like using https://maven.apache.org/enforcer/enforcer-rules/requireActiveProfile.html to prevent that but they would require more work.

4. build and push the container images using Jib as we do on Jenkins

## Test Local Image

On a Mac with arm64 CPU architecture run

```sh
./dhis-2/build-dev.sh
docker inspect dhis2/core-dev:local -f "{{ .Architecture }}"
```

Output should be `arm64`.

You should then also be able to run this image using

```sh
docker compose up
```

## Test PR Image

Multi-arch images are built and pushed for every PR. An example for this PR [dhis2/core-pr:12842
](https://hub.docker.com/layers/dhis2/core-pr/12842/images/sha256-23630f6720e3877a3f8a67c4b6b2df3dbb16cc13eb52574d035d25d3cf7b7f6d?context=explore)

```sh
docker pull dhis2/core-pr:12842
docker inspect dhis2/core-pr:12842 -f "{{ .Architecture }}"
```

should show the architecture of your machine.

### DIFF
--- a/dhis-2/build-dev.sh
+++ b/dhis-2/build-dev.sh
@@ -1,23 +1,6 @@
 #!/usr/bin/env bash
 # Builds DHIS2 war and Docker image for development use
 
-#
-## bash environment
-#
-
-if test "$BASH" = "" || "$BASH" -uc "a=();true \"\${a[@]}\"" 2>/dev/null; then
-    # Bash 4.4, Zsh
-    set -euo pipefail
-else
-    # Bash 4.3 and older chokes on empty arrays with set -u.
-    set -eo pipefail
-fi
-shopt -s nullglob globstar
-
-#
-## script environment
-#
-
 D2CLUSTER="${1:-}"
 IMAGE=dhis2/core-dev
 TAG=local
@@ -34,9 +17,7 @@ if [[ "$ARCH" == *arm64* || "$ARCH" == *aarch64* ]]; then
   JIB_PROFILE="-P jibBuildArmOnly"
 fi
 
-print() {
-    echo -e "\033[1m$1\033[0m" 1>&2
-}
+echo "Building dhis2-core and Docker image..."
 
 print "Building dhis2-core and Docker image..."
 
@@ -46,9 +27,9 @@ mvn clean install --threads 2C -DskipTests -Dmaven.test.skip=true -f "${DIR}/dhi
 mvn -DskipTests -Dmaven.test.skip=true -f "${DIR}/dhis-web/dhis-web-portal/pom.xml" jib:dockerBuild $JIB_PROFILE
 
 if test -z "$D2CLUSTER"; then
-    print "No cluster name specified, skipping deploy"
+    echo "No cluster name specified, skipping deploy"
 else
-    print "Deploying to d2 cluster $D2CLUSTER..."
+    echo "Deploying to d2 cluster $D2CLUSTER..."
 
     d2 cluster up "$D2CLUSTER" --image $IMAGE:$TAG
     d2 cluster logs "$D2CLUSTER"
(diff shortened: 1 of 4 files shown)


radnov:
> figure out whether its a good idea to automatically activate the profile

@teleivo What if we add an extra profile `jibBuildAmd` with `activation` for amd64 based systems? This should result in building the correct arch if you don't specify a profile (based on one's system), but if you do specify `jibBuild` (like we do in CI) it should build both arches regardless of the system we build on, as the explicitly and implicitly activated profiles overlap. 🤔

radnov (commented):
> figure out whether its a good idea to automatically activate the profile

@teleivo What if we add an extra profile `jibBuildAmd` with `activation` for amd64 based systems? This should result in building the correct arch if you don't specify a profile (based on one's system), but if you do specify `jibBuild` (like we do in CI) it should build both arches regardless of the system we build on, as the explicitly and implicitly activated profiles overlap. 🤔

teleivo:
> 

The default https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#platform-object is linux amd64 in the jib maven plugin.

dhis-2/build-dev.sh
```diff
@@ -23,6 +24,16 @@ TAG=local
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 
+# Choosing this approach over automatically activating a maven profile based on the architecture.
+# This is to not risk running both the jibBuild and jibBuildArmOnly profiles in our pipelines.
+# There might be ways like using https://maven.apache.org/enforcer/enforcer-rules/requireActiveProfile.html
+# to prevent that but they would require more work.
+ARCH=$(mvn help:system | grep "os\.arch")
+JIB_PROFILE=
+if [[ "$ARCH" == *arm64* ]]; then
```
radnov, line 33: `os.arch` might be something like `aarch64` on some hosts. Someone with an M1 Mac should test this.

radnov (commented):
Looks good to me, just one comment about the `os.arch` check.

dhis-2/build-dev.sh
teleivo: > os.arch might be something like aarch64 on some hosts. Someone with an M1 Mac should test this.

We should and will definitely test this! 😄 

We are using `mvn` to emulate what we would do when activating profiles automatically via https://maven.apache.org/guides/introduction/introduction-to-profiles.html

I opted for it since I think it does a little "sanitizing" for us. I cannot find all the arch values (yet).

Maven profile code uses
https://github.com/codehaus-plexus/plexus-utils/blob/4f333f2691873e5ba024991a03390a4e773511e9/src/main/java/org/codehaus/plexus/util/Os.java#L103

For example for me I get

```sh
uname -m
x86_64

mvn help:system | grep "os\.arch"
os.arch=amd64
```

teleivo: Lets test for `arm64` and `aarch64` to be safe. Both refer to the same architecture and seem to be used interchangeably.

stian-sandvold (approved):

radnov (approved):

PR #12843, opened 2023-01-28
feat: API Router

mortenoh:
## Summary

- [x] Implement `Route` class and flyway migration
- [x] Implement `RouteController` for basic CRUD operations
- [x] Implement encryption of passwords/tokens
- [x] Implement `RouteService` which can decrypt single proxies
- [x] Implement runnable proxies in `RouteController`
- [x] Implement authorities for routes

[DHIS2-12193](https://dhis2.atlassian.net/browse/DHIS2-12193)

[DHIS2-12193]: https://dhis2.atlassian.net/browse/DHIS2-12193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
```diff
@@ -0,0 +1,74 @@
...
+@Accessors( chain = true )
+public class Route
+    extends BaseIdentifiableObject
+    implements MetadataObject
+{
+    @JsonProperty
+    private String description;
+
+    @JsonProperty( required = true )
+    private boolean enabled = true;
```
larshelge, line 61: I have a slight preference for booleans to be default false to match the Java default. Could we perhaps make this `disabled = false`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,160 @@
...
+    private final RouteStore routeStore;
+
+    private final ObjectMapper objectMapper;
+
+    @Qualifier( AES_128_STRING_ENCRYPTOR )
+    private final PBEStringCleanablePasswordEncryptor encryptor;
+
+    private static final RestTemplate restTemplate = new RestTemplate();
+
+    static
```
larshelge, line 74: Perhaps use @PostConstruct to let Spring manage the lifecycle? Probably does not make much difference.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,160 @@
...
+        restTemplate.setRequestFactory( requestFactory );
+    }
+
+    public Route getDecryptedById( @Nonnull String id )
+    {
+        Route route = routeStore.getByUidNoAcl( id );
+
+        if ( route == null )
+        {
+            return null;
```
larshelge, line 91: Do we want to throw an exception here, i.e. is this an invalid state/programming error or user error? Could silently return null hide problems and make it hard to debug?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        try
+        {
+            route = objectMapper.readValue( objectMapper.writeValueAsString( route ), Route.class );
+            decrypt( route );
```
larshelge, line 97: Can `decrypt` throw JsonProcessingException? If not, perhaps move `decrypt` out of the try/catch?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,160 @@
...
+        UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromHttpUrl( route.getUrl() )
+            .queryParams( queryParameters );
+
+        String body = StreamUtils.copyToString( request.getInputStream(), StandardCharsets.UTF_8 );
+
+        HttpEntity<String> entity = new HttpEntity<>( body, headers );
+
+        HttpMethod httpMethod = HttpMethod.resolve( request.getMethod() );
+
+        if ( httpMethod == null )
```
larshelge, line 131: Prefer `ObjectUtils.firstNonNull( httpMethod, HttpMethod.GET)` to avoid if statement.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,160 @@
...
+    private void decrypt( Route route )
+    {
+        Auth auth = route.getAuth();
+
+        if ( auth == null )
+        {
+            return;
+        }
+
+        if ( auth.getType().equals( "api-token" ) )
```
larshelge, line 149: Do we want some sort of constants/interface instead of inline strings?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/RouteObjectBundleHook.java
```diff
@@ -0,0 +1,95 @@
...
+        encrypt( route );
+    }
+
+    private void encrypt( Route route )
+    {
+        Auth auth = route.getAuth();
+
+        if ( auth == null )
+        {
+            return;
```
larshelge, line 73: Do we want to silently return null or throw exception?

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/descriptors/RouteSchemaDescriptor.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public Schema getSchema()
+    {
+        Schema schema = new Schema( Route.class, SINGULAR, PLURAL );
+        schema.setRelativeApiEndpoint( API_ENDPOINT );
+        schema.setOrder( 1200 );
+
+        schema.add( new Authority( AuthorityType.CREATE_PUBLIC, Lists.newArrayList( "F_ROUTE_PUBLIC_ADD" ) ) );
+        schema.add( new Authority( AuthorityType.CREATE_PRIVATE, Lists.newArrayList( "F_ROUTE_PRIVATE_ADD" ) ) );
```
larshelge, line 57: I am wondering if sharing, and particular private sharing, really makes sense here. Do we want to give access to routes only to certain users? I think creating private routes does not make much sense?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_14__Add_table_route.sql
```diff
@@ -0,0 +1,29 @@
+-- Add proxy table
```
larshelge, line 1: Rename proxy to route in comment.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/dbms/HibernateDbmsManager.java
```diff
@@ -333,6 +333,7 @@ public void emptyDatabase()
         emptyTable( "fileresource" );
         emptyTable( "trackedentitycomment" );
         emptyTable( "userinfo" );
+        emptyTable( "proxy" );
```
larshelge, line 336: Rename proxy to route.

dhis-2/dhis-support/dhis-support-hibernate/src/main/resources/org/hisp/dhis/usertype/UserTypes.hbm.xml
```diff
@@ -99,6 +99,10 @@
     <param name="clazz">java.util.ArrayList</param>
   </typedef>
 
+  <typedef class="org.hisp.dhis.hibernate.jsonb.type.JsonBinaryType" name="jbMap">
```
larshelge, line 102: Is Map a good candidate as JSONB entity? Should we have some wrapper around it?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+@OpenApi.Tags( "integration" )
+@RequiredArgsConstructor
+@RequestMapping( value = RouteSchemaDescriptor.API_ENDPOINT )
+@ApiVersion( { DhisApiVersion.DEFAULT, DhisApiVersion.ALL } )
+public class RouteController
+    extends AbstractCrudController<Route>
+{
+    private final RouteService routeService;
+
+    @RequestMapping( value = "/exec/{id}", method = { RequestMethod.GET, RequestMethod.POST } )
```
larshelge, line 66: I don't have a better option but what is `exec` referring to? Perhaps `route` is better.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @JsonProperty( required = true )
+    private String url;
+
+    @JsonProperty( required = true )
+    private Map<String, String> headers = new HashMap<>();
```
larshelge, line 67: Maybe create our own headers class. I seem to recall some bad experience from using Map as a Hibernate entity.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
mortenoh: Sure, that requires changes to my already commited PR.. but maybe it makes sense?

larshelge:
Nice work!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
larshelge: Right, I think we already had a Flyway migration fix this week, another won't hurt :-)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
mortenoh: It's thread safe so just did it like that for now, long term we might want `Route` specific TTLs etc, I can make it local if it makes sense

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
mortenoh: I don't really like throwing http messages from the core layer, that's why i return `null` == 404, I think if we want that we should do a larger refactor

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
mortenoh: Changed to `Objects.requireNonNullElse`

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/RouteObjectBundleHook.java
mortenoh: No need to report here, if there is no auth.. just means its a simple non-authorized http request, so no `auth` is fine

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/descriptors/RouteSchemaDescriptor.java
mortenoh: It might. But in the controller that actually exec the router, it will check for metadata first.. but then if you are in the auhorities list.. you can still exec it

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
mortenoh: Is `/api/routes/routes` better? it sounds a bit weird

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
mortenoh: We are using `HttpHeaders<>` at the upper levels, but i didn't want to add more spring deps to our core api modue.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
larshelge: Okay. Maybe move the creation of the HttpComponentsClientHttpRequestFactory to a separate method?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
larshelge: Right, I did not mean HTTP message, rather one of our existing exceptions, but fair enough.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
larshelge: Right, I was thinking we could create our own to encapsulate it a bit better, but fair enough.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
larshelge: Ah sorry. We already have routes from the schema descriptor.

larshelge: What if we add a _new_ controller for the actual "execution" of routes. It could be called something like `ApiGatewayController`, exposed at `/api/gateway`. I noticed that Jira/Atlassian uses that pattern in the their API. Then we define the gateway as a dedicated API, which uses/depends on the RouteService to route requests.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+    public ResponseEntity<String> exec( @PathVariable( "id" ) String id, @CurrentUser User user,
+        HttpServletRequest request )
+        throws IOException
+    {
+        Route route = routeService.getDecryptedById( id );
+
+        if ( route == null )
+        {
+            throw new HttpClientErrorException( HttpStatus.NOT_FOUND, "Route not found" );
```
larshelge, line 75: Include the ID of the route in the error message to make it clearer perhaps.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
```diff
@@ -0,0 +1,74 @@
...
+@Setter
+@EqualsAndHashCode( callSuper = true )
+@Accessors( chain = true )
+public class Route
+    extends BaseIdentifiableObject
+    implements MetadataObject
+{
+    @JsonProperty
+    private String description;
+
```
larshelge, line 59: Should we have a descriptive name for the route, which makes up a good API path? I.e. it is easier for a consumer to call `/api/routes/commodities` vs `/api/routes/bgqJyhGtf1`.

mortenoh: We can do that with the `code` field perhaps?

mortenoh: The `code` has been added as possible route name id.

dhis-2/dhis-support/dhis-support-hibernate/src/main/resources/org/hisp/dhis/usertype/UserTypes.hbm.xml
mortenoh: It's a very generic entity... I'm not sure, in the lower layers we are using `HttpHeaders` but don't want to use the web layer for this.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
mortenoh: We can discuss that later on. I think we should focus on getting it in, we definetely want a lot of input here from the FE side also, I think the important bits is that its clear and understandable what happens behind the endpoint

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -0,0 +1,162 @@
...
+        HttpHeaders headers = new HttpHeaders();
+        route.getHeaders().forEach( headers::add );
+
+        if ( route.getAuth() != null )
+        {
+            route.getAuth().apply( headers );
+        }
+
+        HttpHeaders queryParameters = new HttpHeaders();
+        request.getParameterMap().forEach( ( key, value ) -> queryParameters.addAll( key, Arrays.asList( value ) ) );
```
larshelge, line 127: I think it is usually god to _not_ use `Arrays.asList` directly. It returns some sort of nested class from inside the Arrays class, and changes to this list will be reflected in the original array. Better to use a method that gives you a new, regular ArrayList, like `ListUtils.newList`.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_5__add_evenhook_table.sql
```diff
@@ -5,6 +5,7 @@ create table if not exists eventhook (
     eventhookid int8 not null,
     uid varchar(11) null,
     code varchar(100) null,
+	disabled bool NOT NULL,
```
larshelge, line 8: Remove extra indentation, probably tabs instead of spaces.

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/route.hibernate/Route.hbm.xml
```diff
@@ -0,0 +1,43 @@
...
+        &identifiableProperties;
+
+        <property name="name" column="name" not-null="true" unique="true" length="230"/>
+
+        <many-to-one name="createdBy" class="org.hisp.dhis.user.User" column="userid"
+                     foreign-key="fk_route_userid_userinfoid"/>
+
+        <property name="description" type="text"/>
+
+        <property name="enabled" type="boolean"/>
```
jbee, line 23: wasn't this disabled in the class now?

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/descriptors/RouteSchemaDescriptor.java
```diff
@@ -0,0 +1,62 @@
...
+    public static final String API_ENDPOINT = "/" + PLURAL;
+
+    @Override
+    public Schema getSchema()
+    {
+        Schema schema = new Schema( Route.class, SINGULAR, PLURAL );
+        schema.setRelativeApiEndpoint( API_ENDPOINT );
+        schema.setOrder( 1200 );
+
+        schema.add( new Authority( AuthorityType.CREATE_PUBLIC, Lists.newArrayList( "F_ROUTE_PUBLIC_ADD" ) ) );
```
jbee, line 56: Even if it is copy&paste we should not keep using `Lists.newArrayList`. I know we said we update them all later but we should not get used to the bad habit of copy&paste code without bring it up to date after it gets pasted.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+    @RequestMapping( value = "/{id}/run", method = { RequestMethod.GET, RequestMethod.POST } )
+    public ResponseEntity<String> run( @PathVariable( "id" ) String id, @CurrentUser User user,
+        HttpServletRequest request )
+        throws IOException
+    {
+        Route route = routeService.getDecryptedRoute( id );
+
+        if ( route == null )
+        {
+            throw new HttpClientErrorException( HttpStatus.NOT_FOUND, String.format( "Route %s not found", id ) );
```
jbee, line 75: Please use `feedback.NotFoundException`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if ( !aclService.canRead( user, route ) && !user.hasAnyAuthority( route.getAuthorities() ) )
+        {
+            throw new HttpClientErrorException( HttpStatus.FORBIDDEN, "User not authorized" );
```
jbee, line 80: Please use `feedback.ForbiddenException`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        ResponseEntity<String> entity = routeService.exec( route, request );
+
+        if ( entity.getStatusCode().is4xxClientError() )
+        {
+            throw new HttpClientErrorException( entity.getStatusCode() );
```
jbee, line 87: I guess in line with some other discussions we would make these into `feedback.ConflictException` with a custom `ErrorCode` for this particular issue. Than you can have a message where the status code is added.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        else if ( entity.getStatusCode().is5xxServerError() )
+        {
+            throw new HttpServerErrorException( entity.getStatusCode() );
```
jbee, line 91: Similar as comment above.

jbee (changes requested):

netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/route.hibernate/Route.hbm.xml
mortenoh: We renamed from `disabled` to `enabled`.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_5__add_evenhook_table.sql
mortenoh: That just a diff tab.. there are no extra spaces in the file itself

PR #12851, opened 2023-01-30
fix: only add lock exceptions for org units in data capture hierarchy [DHIS2-13908]

jbee:
### Summary
Limits adding lock exceptions to users that have the affected organisation unit in their data capture organisation unit hierarchy.

### Automatic Testing
Added a controller test for the validation failed scenario.

### Manual Testing
Error Scenario
* Create a user with only the `F_DATASET_PUBLIC_ADD` authority
* switch to new user
* try to add a lock exception for an existing data set and some existing organisation unit
* check a 403, forbidden status us returned with a descriptive error message

Success Scenario
* add the target organisation unit or any of its parents to the test user's data capture set (the "normal organisation unit" set of the user)
* rerun the test scenario above and check the exception can now be added


larshelge (approved):

larshelge:
Nice spot and test :+1:

PR #12852, opened 2023-01-30
fix: cannot patch user.userGroups using api/users endpoint

vietnguyen:
backport https://github.com/dhis2/dhis2-core/pull/12841

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
@@ -65,6 +65,9 @@ import org.hisp.dhis.common.MergeMode;
 import org.hisp.dhis.common.Pager;
 import org.hisp.dhis.common.UserOrgUnitType;
 import org.hisp.dhis.commons.collection.CollectionUtils;
+import org.hisp.dhis.commons.jackson.jsonpatch.JsonPatch;
+import org.hisp.dhis.commons.jackson.jsonpatch.JsonPatchOperation;
+import org.hisp.dhis.commons.jackson.jsonpatch.operations.AddOperation;
 import org.hisp.dhis.dbms.DbmsManager;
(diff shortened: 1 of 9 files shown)


jbee (commented):
I think post hook is the wrong place to add this in. 
If I understand correctly each time a user is changed in a specific way something with the user groups needs to be updated as well. This sound like a object bundle hook problem so that other calls that might have such an effect would also properly update. I understand that it gets more complicated in the bundle hooks as we would have to find out from the difference between the old and the new user object that such a change was made but still that maybe is the more consistent way of doing this.

vietnguyen:
@jbee This is just to support the `Users` app to reduce number of requests.
 I think we don't want to support update `user.userGroups` from the `User` side of association in the metadata import service. This hack should only be fixed in `UserController`. 
When import both `Users` and `UserGroups` in one payload , the import service will first create the `User` then create `UserGroup` and update the association. 
Usually, people will import all `Users` in one payload. Then they import another payload with the `UserGroup` and their members list.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -817,14 +821,33 @@ protected void prePatchEntity( User oldEntity, User newEntity )
...
+        for ( JsonPatchOperation op : patch.getOperations() )
+        {
+            JsonPointer userGroups = op.getPath().matchProperty( "userGroups" );
+            if ( userGroups == null )
+            {
+                continue;
+            }
+
+            String opName = op.getOp();
+            if ( opName.equals( "add" ) )
```
larshelge, line 843: Could we use some sort of constant here?

vietnguyen: Fixed

jbee:
Tests fails for unrelated reasons - working on a backport that should fix it.

larshelge (approved):

PR #12858, opened 2023-01-31
fix: data value validation - compare date ranges as dates not timestamps [DHIS2-14591]

jbee:
Adjusting `endDate` to `23:59:59.999` did not work out. 100+ tests failed. Quick dive showed that there are many downstream implications to such a change. 

Therefore I took the more local approach to only fix the date comparison itself. 
When comparing dates without time one always has to decide in which time-zone the timestamp should be considered as date - means there is still room for semantic errors. I think the most reasonable is to compare the timestamps as dates in the system zone. 

This PR fixes the issues related to the data validation. There clearly are more but this at the moment causes test failures and needs fixing.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
@@ -27,9 +27,6 @@
  */
 package org.hisp.dhis.period;
 
-import static org.hisp.dhis.util.DateUtils.atEndOfDay;
-import static org.hisp.dhis.util.DateUtils.atStartOfDay;
-
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
```diff
@@ -425,4 +425,10 @@ public boolean isDefault()
     {
         return Objects.isNull( dateField );
     }
+
+    private Date atEndOfDay( Date date )
```
stian-sandvold, line 429: Do we know for sure all dates passed in here has 00:00:00 time? Would it make sense to set time to 00:00:00 then add the offset?

jbee: We do. I started doing this with java.time but then you have to consider time-zones and sort of rely on that all involved code has the same idea of time zone which is almost unlikely since timestamps are by definition UTC and system zone is unlike to be UTC. Therefore I just assumed whatever is passed as start is "zero time" so I hard added a milli less than a day. But the results were the same both ways. Lots of trouble. I think we have to give up on this way of fixing it.

larshelge (approved):

stian-sandvold (approved):

PR #12874, opened 2023-02-01
ci: add labels for revision/branch when building locally

teleivo:
since images by default will get name dhis2/core-dev:local adding labels help identify where an image came from

### DIFF
--- a/dhis-2/build-dev.sh
+++ b/dhis-2/build-dev.sh
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 # Builds DHIS2 war and Docker image for development use
 
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
 D2CLUSTER="${1:-}"
 IMAGE=dhis2/core-dev
 TAG=local
-
-DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+BUILD_REVISION=$(git --git-dir "$DIR/../.git" rev-parse HEAD)
+BUILD_BRANCH=$(git --git-dir "$DIR/../.git" branch --show-current)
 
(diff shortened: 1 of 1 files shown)


dhis-2/build-dev.sh
```diff
@@ -1,11 +1,13 @@
...
 
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
 D2CLUSTER="${1:-}"
 IMAGE=dhis2/core-dev
 TAG=local
-
-DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+BUILD_BRANCH=$(git --git-dir "$DIR/../.git" rev-parse HEAD)
+BUILD_REVISION=$(git --git-dir "$DIR/../.git" branch --show-current)
```
radnov, line 10: The `BUILD_BRANCH` and `BUILD_REVISION` variables are swapped, they should be the other way around.

Also, is the `--git-dir` part needed? I'm getting the same results without it, too.

radnov (changes requested):

dhis-2/build-dev.sh
teleivo: good catch 😅 

`--git-dir` is so we are safe if someone calls the script from outside this repo.

Try `cd`ing out and run the script again. You should see

```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

teleivo: its the equivalent of us adding `${DIR}` to the pom path in 

`mvn clean install --threads 2C -DskipTests -Dmaven.test.skip=true -f "${DIR}/pom.xml" -pl -dhis-web-embedded-jetty,-dhis-test-integration,-dhis-test-coverage`

radnov (approved):

stian-sandvold (approved):

PR #12875, opened 2023-02-01
fix: Param displayProperty:SHORTNAME doesn't return short name [DHIS2-13173]

d-bernat:
please see https://dhis2.atlassian.net/browse/DHIS2-13173

By conflict resolution and rebasing on master there are now some more commits inside, sorry for that. All commits not related to this story are about DataApproval commits

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
@@ -53,7 +53,6 @@ import static org.hisp.dhis.common.DimensionalObjectUtils.getDimensionFromParam;
 import static org.hisp.dhis.common.DimensionalObjectUtils.getDimensionItemsFromParam;
 import static org.hisp.dhis.common.DimensionalObjectUtils.getDimensionalItemIds;
 import static org.hisp.dhis.common.DimensionalObjectUtils.getItemsFromParam;
-import static org.hisp.dhis.common.DisplayProperty.NAME;
 import static org.hisp.dhis.common.IdScheme.UID;
 import static org.hisp.dhis.feedback.ErrorCode.E7125;
 import static org.hisp.dhis.util.ObjectUtils.firstNonNull;
@@ -237,7 +236,7 @@ public class DefaultDataQueryService
         List<OrganisationUnit> userOrgUnits, boolean allowNull, DisplayProperty displayProperty,
         IdScheme inputIdScheme )
     {
-        return getDimension( dimension, items, relativePeriodDate, displayProperty == null ? NAME : displayProperty,
+        return getDimension( dimension, items, relativePeriodDate, displayProperty,
             userOrgUnits, allowNull, inputIdScheme );
     }
 
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
@@ -126,7 +126,7 @@ public class DefaultDataQueryService
         if ( isNotEmpty( request.getFilter() ) )
         {
             params.addFilters( getDimensionalObjects( request.getFilter(), request.getRelativePeriodDate(),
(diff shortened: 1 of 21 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
```diff
@@ -234,9 +234,11 @@ public DimensionalObject getDimension( String dimension, List<String> items, Eve
 
     @Override
     public DimensionalObject getDimension( String dimension, List<String> items, Date relativePeriodDate,
-        List<OrganisationUnit> userOrgUnits, boolean allowNull, IdScheme inputIdScheme )
+        List<OrganisationUnit> userOrgUnits, boolean allowNull, DisplayProperty displayProperty,
+        IdScheme inputIdScheme )
     {
-        return getDimension( dimension, items, relativePeriodDate, NAME, userOrgUnits, allowNull, inputIdScheme );
+        return getDimension( dimension, items, relativePeriodDate, displayProperty == null ? NAME : displayProperty,
```
larshelge, line 240: Would be good to avoid this in-line check. We should probably handle null further down in the method calls.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventDataQueryService.java
```diff
@@ -378,7 +378,7 @@ public EventQueryParams getFromAnalyticalObject( EventAnalyticalObject object )
         for ( DimensionalObject dimension : ListUtils.union( object.getColumns(), object.getRows() ) )
         {
             DimensionalObject dimObj = dataQueryService.getDimension( dimension.getDimension(),
-                getDimensionalItemIds( dimension.getItems() ), date, null, true, idScheme );
+                getDimensionalItemIds( dimension.getItems() ), date, null, true, null, idScheme );
```
larshelge, line 381: Would be good to somehow avoid adding many null arguments, makes the code more error prone and harder to read.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datasetreport/jdbc/AnalyticsDataSetReportStore.java
```diff
@@ -158,7 +158,7 @@ public Map<String, Object> getAggregatedSubTotals( DataSet dataSet, List<Period>
                 if ( filters != null )
                 {
                     params.addFilters(
-                        dataQueryService.getDimensionalObjects( filters, null, null, IdScheme.UID ) );
+                        dataQueryService.getDimensionalObjects( filters, null, null, null, IdScheme.UID ) );
```
larshelge, line 161: Would be good to somehow avoid adding many null arguments, makes the code more error prone and harder to read.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventDataQueryService.java
d-bernat: Yes, I agree, we are planning to redesign the way how we are passing parameters down to the methods.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datasetreport/jdbc/AnalyticsDataSetReportStore.java
d-bernat: please see above

maikelarabori (approved):
Looks good @d-bernat. As @larshelge has pointed out, we should get rid of "null" arguments. We can discuss it later and find a better design for those cases.

gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultDataQueryService.java
d-bernat: Outdated due the requested changes to pass null down to the methods and handle it there

PR #12880, opened 2023-02-01
fix: remove deleted organisation unit from users [DHIS2-8953]

jbee:
### Summary
When deleting an organisation unit it has to be removed from all 3 of the member collections a user has.

### Automatic Testing
Added a test where a user had the deleted OU in all of the 3 member collections.

### Manual Testing
* create a new OU
* edit a user and select the created OU in all three OU trees (capture, search, data output)
* delete the OU
* deletion should be successful
* check the user, the OU should no longer be shown or selected in the trees


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserDeletionHandler.java
```diff
@@ -72,11 +74,21 @@ private void deleteUserRole( UserRole role )
 
     private void deleteOrganisationUnit( OrganisationUnit unit )
     {
-        for ( User user : unit.getUsers() )
+        for ( User user : userStore.getUsers( new UserQueryParams().addOrganisationUnit( unit ) ) )
```
jbee, line 77: FYI: test showed that `unit.getUsers()` would not work. I assume this is because the object is not fully initialized.

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserDeletionHandler.java
larshelge: Right, it depends where the incoming org unit is coming from. If it comes from outside the "transactional boundary" / not within a transaction/Hibernate session then the users will not be initialized. I think for most deletion handlers, the incoming object will be attached to a transaction/session. So I guess it would be good to understand why this is a problem here in particular instead of working around. Could be a good reason for all I know.

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/DhisConvenienceTest.java
```diff
@@ -2897,11 +2897,13 @@ protected User createAndAddUser( boolean superUserFlag, String userName, Organis
     }
 
     protected User createAndAddUser( boolean superUserFlag, String userName, OrganisationUnit orgUnit,
-        OrganisationUnit dataViewOrganisationUnits, String... auths )
+        OrganisationUnit dataViewOrganisationUnit, String... auths )
     {
-        User user = _createUserAndRole( superUserFlag, userName, newHashSet( orgUnit ),
-            dataViewOrganisationUnits != null ? newHashSet( dataViewOrganisationUnits ) : newHashSet( orgUnit ),
```
jbee, line 2903: OBS! Note the `newHashSet( orgUnit )` - `orgUnit` can be `null` here causing a set with a `null` element to be created.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/UserDeletionHandler.java
jbee: The instance came directly from test setup where it never was fully initialized by hibernate. Also there was another issue in the setup causing a `null` to be inserted in the set of units which I fixed now.

larshelge: Great!

PR #12882, opened 2023-02-01
fix: Return meaningful error message for wrong enum value in parameter [DHIS2-14182]

enricocolasante:
This PR is the first one in order to fix https://dhis2.atlassian.net/browse/DHIS2-14182
The issue was a wrong value of the `commitMode` parameter in the code (it is correct in the docs) so the user was trying to import and just run the validation, but the wrong value was discarded and the default to validate and commit was used instead.

The purpose here is to fix the wrong value and to add validation for all parameters in tracker importer.

First step to do so is to intercept `MethodArgumentTypeMismatchException`, that is thrown by Spring when a conversion from a String to a specific type in the parameters of an endpoint is not successful, and return a 400 (BAD_REQUEST) code.
For now I am only focusing on enums, perhaps we should then expand it to other types.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
@@ -194,7 +194,7 @@ public class CrudControllerAdvice
     {
         if ( ex.getRequiredType() == null )
         {
-            return defaultExceptionHandler( ex );
+            return handleBadRequest( ex );
         }
 
         if ( ex.getRequiredType().isEnum() )
@@ -206,7 +206,7 @@ public class CrudControllerAdvice
                 ex.getValue(), ex.getName(), validValues );
             return badRequest( errorMessage );
         }
-        return defaultExceptionHandler( ex );
+        return handleBadRequest( ex );
     }
 
     @ExceptionHandler( InvalidEnumValueException.class )
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
@@ -104,6 +104,7 @@ import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.HttpServerErrorException;
 import org.springframework.web.client.HttpStatusCodeException;
 import org.springframework.web.client.RestClientException;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportControllerTest.java
```diff
@@ -370,7 +373,7 @@ void verifyShouldThrowWhenJobReportNotFound()
         mockMvc.perform( get( ENDPOINT + "/jobs/" + uid + "/report" )
             .content( "{}" )
             .contentType( MediaType.APPLICATION_JSON )
-            .accept( MediaType.APPLICATION_JSON ) ).andExpect( status().isNotFound() )
+            .accept( MediaType.APPLICATION_JSON ) )
```
enricocolasante, line 376: After adding the CrudControllerAdvice class, this status is not reflecting the WebMessage status anymore.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -187,6 +188,27 @@ public WebMessage illegalQueryExceptionHandler( IllegalQueryException ex )
         return conflict( ex.getMessage(), ex.getErrorCode() );
     }
 
+    @ExceptionHandler( MethodArgumentTypeMismatchException.class )
+    @ResponseBody
+    public WebMessage methodArgumentTypeMismatchException( MethodArgumentTypeMismatchException ex )
+    {
+        if ( ex.getRequiredType() == null )
+        {
+            return defaultExceptionHandler( ex );
```
jbee, line 197: I think we should always make this a bad request. Any type conversion issue should mean the input was invalid. We can just go with a generic message in case we don't have a specific handling for the type.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportControllerTest.java
jbee: Can't really follow you. If the report does not exist I think we want to keep having this result in a 404.

enricocolasante: As we discussed, these type of tests are not returning the correct status as the `mockMvc` should set the WebMessage Advice to enhance the response with the correct status

jbee (approved):

larshelge:
Nice fix @enricocolasante!

PR #12891, opened 2023-02-02
fix: Add pagination to Potential Duplicate Get Query [TECH-644]

lucaCambi77:
https://dhis2.atlassian.net/browse/TECH-644

This issue was initially planned with the Deduplication feature. It adds a quick implementation for the Potential Duplicate pagination.
Actually, in the front-end (both old and new Capture) ) the Potential Duplicate table is not used at all, so in https://dhis2.atlassian.net/browse/DHIS2-14599 the front-end team will see how this can be further improved or adapted if required at some point. If not, it is worth considering deprecating the `/potentialDuplicates` endpoint also from the backend


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationController.java
```diff
@@ -159,18 +151,18 @@ public PotentialDuplicate postPotentialDuplicate(
 
     @PutMapping( value = "/{id}" )
     @ResponseStatus( value = HttpStatus.OK )
-    public void updatePotentialDuplicate( @PathVariable String id, @RequestParam( value = "status" ) String status )
+    public void updatePotentialDuplicate( @PathVariable String id,
+        @RequestParam( value = "status" ) DeduplicationStatus status )
         throws NotFoundException,
         BadRequestException
     {
         checkDeduplicationStatusRequestParam( status );
```
enricocolasante, line 159: This check will never throw an Exception. If an invalid value is passed for the DeduplicationStatus then Spring will throw a `MethodArgumentTypeMismatchException` that will be intercepted in `CrudControllerAdvice` and translated to a `BadRequestException`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationController.java
```diff
@@ -97,40 +92,37 @@
...
-        HttpServletResponse response )
+    public PagingWrapper<ObjectNode> getPotentialDuplicates(
+        PotentialDuplicateCriteria potentialDuplicateCriteria,
+        HttpServletResponse response,
+        @RequestParam( defaultValue = DEFAULT_FIELDS_PARAM ) List<String> fields )
         throws BadRequestException
     {
-        checkDeduplicationStatusRequestParam(
-            Optional.ofNullable( query.getStatus() ).map( Enum::name ).orElse( null ) );
+        checkDeduplicationStatusRequestParam( potentialDuplicateCriteria.getStatus() );
```
enricocolasante, line 105: This check will never throw an Exception. If an invalid value is passed for the DeduplicationStatus then Spring will throw a `BindingException` that should be intercepted in CrudControllerAdvice and translated to a BadRequestException. The interceptor for this exception is not implemented yet

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationController.java
```diff
@@ -229,11 +221,11 @@ private void checkDbAndRequestStatus( PotentialDuplicate potentialDuplicate,
                 + DeduplicationStatus.MERGED.name() );
     }
 
-    private void checkDeduplicationStatusRequestParam( String status )
+    private void checkDeduplicationStatusRequestParam( DeduplicationStatus status )
```
enricocolasante, line 224: This shouldn't be necessary then

enricocolasante (commented):
I think there is something odd in the Controller and the validation of the `DeduplicationStatus`.
Generally is good to have a test in `web-api-test` to have all the test cases to cover the different invalid parameters and the errors returned

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationController.java
lucaCambi77: yes this is old code so I need to review with your suggestion

lucaCambi77:
@enricocolasante I've updated MVC and added integration tests. There is should be still one test failing  related to https://dhis2.atlassian.net/browse/DHIS2-13928

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -0,0 +1,169 @@
...
+        PotentialDuplicate potentialDuplicate = potentialDuplicate( origin.getUid(), duplicate.getUid() );
+        assertStatus( HttpStatus.OK,
+            PUT( ENDPOINT + potentialDuplicate.getUid() + "?status=" + DeduplicationStatus.INVALID.name() ) );
+    }
+
+    @Test
+    void shouldThrowBindExceptionWhenGetPotentialDuplicateWithInvalidStatus()
+    {
+        assertStatus( HttpStatus.BAD_REQUEST,
+            GET( ENDPOINT + "?status=" + "invalidStatus" ) );
```
teleivo, line 131: @enricocolasante should we test these BindExceptions and MethodArgumentTypeMismatchExceptions in individual controllers? Or is it simply duplicating tests you already have elsewhere?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -0,0 +1,169 @@
...
+        PotentialDuplicate potentialDuplicate = new PotentialDuplicate( origin.getUid(), duplicate.getUid() );
+        assertStatus( HttpStatus.OK, POST( ENDPOINT, objectMapper.writeValueAsString( potentialDuplicate ) ) );
+    }
+
+    @Test
+    void shouldThrowPostPotentialDuplicateWhenMissingDuplicateTeiInPayload()
+        throws Exception
+    {
+        PotentialDuplicate potentialDuplicate = new PotentialDuplicate( origin.getUid(), null );
+        assertStatus( HttpStatus.BAD_REQUEST, POST( ENDPOINT, objectMapper.writeValueAsString( potentialDuplicate ) ) );
```
teleivo, line 89: We have multiple tests that are testing the same endpoint and method. Since we only assert on the status code there is a risk that multiple tests test the same case. Even if this is not the case now it could be after a refactoring and not be caught by the tests.

I wonder if there is a keyword specific to these cases that we can assert on in the body? Not asserting on the exact message.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/deduplication/hibernate/HibernatePotentialDuplicateStore.java
```diff
@@ -104,64 +109,79 @@ public HibernatePotentialDuplicateStore( SessionFactory sessionFactory, JdbcTemp
...
+        CriteriaQuery<Long> countCriteriaQuery = cb.createQuery( Long.class );
+        Root<PotentialDuplicate> root = countCriteriaQuery.from( PotentialDuplicate.class );
 
-            setStatusParameter( query.getStatus(), hibernateQuery );
+        countCriteriaQuery
+            .select( cb.count( root ) );
 
-            return hibernateQuery.getSingleResult().intValue();
-        } ).orElseGet( () -> {
+        Predicate[] predArray = getQueryPredicates( query, cb, root );
```
teleivo, line 122: I would inline vars that are directly used. Less to read 😄

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/deduplication/hibernate/HibernatePotentialDuplicateStore.java
```diff
@@ -104,64 +109,79 @@ public HibernatePotentialDuplicateStore( SessionFactory sessionFactory, JdbcTemp
...
+        if ( !query.getTeis().isEmpty() )
         {
-            hibernateQuery.setParameterList( "status", Collections.singletonList( status ) );
+            predicateList.add( builder.and(
+                builder.or(
+                    root.get( "original" ).in( query.getTeis() ),
+                    root.get( "duplicate" ).in( query.getTeis() ) ) ) );
         }
+
+        Predicate[] predicates = new Predicate[predicateList.size()];
```
teleivo, line 175: Is intellij flagging this for you? 🤔 

According to https://stackoverflow.com/questions/9572795/convert-list-to-array-in-java

`predicateList.toArray(new Predicate[0]);`

should be preferred

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/deduplication/DeduplicationServiceIntegrationTest.java
```diff
@@ -71,24 +73,23 @@ public void setUpTest()
...
-        assertEquals( 2, deduplicationService.getAllPotentialDuplicates().size() );
-        potentialDuplicateQuery.setStatus( DeduplicationStatus.OPEN );
-        assertEquals( 1, deduplicationService.getAllPotentialDuplicatesBy( potentialDuplicateQuery ).size() );
-        potentialDuplicateQuery.setStatus( DeduplicationStatus.INVALID );
-        assertEquals( 1, deduplicationService.getAllPotentialDuplicatesBy( potentialDuplicateQuery ).size() );
-        potentialDuplicateQuery.setStatus( DeduplicationStatus.ALL );
-        assertEquals( 2, deduplicationService.getAllPotentialDuplicatesBy( potentialDuplicateQuery ).size() );
+
+        criteria.setStatus( DeduplicationStatus.OPEN );
+        assertEquals( 1, deduplicationService.getPotentialDuplicates( criteria ).size() );
```
teleivo, line 88: Such assertions especially in integration tests can be frustrating once the test is flaky and only fails on CI. They just tell us `expected 1, got 3` if the sizes don't match. Ideally we also see whats in the collection if sizes don't match.

What do you think of adding and using

```java
    /**
     * Asserts that the given collection is of the expected size. Use {@link #assertIsEmpty(Collection)} instead of asserting on size 0.
     *
     * @param expectedSize the expected size of actual
     * @param actual the actual collection to check
     */
    public static <T> void assertSize(int expectedSize, Collection<T> actual) {
        assertNotEmpty(actual);
        assertEquals(expectedSize, actual.size(), () -> String.format("Expected collection to contain %d elements, got %d instead: %s", expectedSize, actual.size(), actual));
    }
```

here https://github.com/dhis2/dhis2-core/blob/ecf7d9990ec39aaeded74cd287633022c7769493/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java#L112

gives us a message like

```
org.opentest4j.AssertionFailedError: Expected collection to contain 6 elements, got 3 instead: [1, 2, 3] ==> 
Expected :6
Actual   :3
```

teleivo (approved):
Great 😄 
Some minor comments and a question for @enricocolasante

enricocolasante (approved):
Just a small comment on redundant tests

PR #12897, opened 2023-02-02
refactor: Simplify event orgunit sql query [TECH-1478]

abyot:


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -2272,92 +2276,102 @@ private List<ProgramStageInstance> sort( List<ProgramStageInstance> batch )
...
+
+        String orgUnitSql;
 
-        if ( params.getOrgUnit() != null && !params.isPathOrganisationUnitMode() )
+        switch ( orgUnitSelectionMode )
         {
-            mapSqlParameterSource.addValue( "organisationunitid", params.getOrgUnit()
-                .getId() );
+        case ALL:
+            orgUnitSql = null;
```
stian-sandvold, line 2298: Could we, in cases where we would now return null, return a valid SQL expression that neutralizes the overall query? That way we don't get a "AND ()", but we actually "break" the query. 

For example "AND (ou.organisationunitid IN (-1))". I am not sure if (null) is valid SQL, but even if it is, does it resolve to TRUE or FALSE in the sense of the "AND"? My worry is that (null) will be valid and treated as TRUE, meaning the query will still complete, but with ALL data (worst case)?

stian-sandvold (changes requested):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
abyot: We append the `AND` only when we have `orgUnitSql` - [otherwise we skip](https://github.com/dhis2/dhis2-core/blob/ab7c1fa941fa78ca493bd4ea6fbf98dc20d5bac3/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java#L1312).

We also have ouMode validation before reaching to this point.

stian-sandvold (approved):

lucaCambi77 (approved):

PR #12907, opened 2023-02-03
fix: Fix importMode values in tracker importer [DHIS2-14182]

enricocolasante:
ImportMode parameter in tracker importer docs had wrong values `[COMMIT, VALIDATION]` instead of `[COMMIT, VALIDATE]` as stated in the [docs](https://docs.dhis2.org/en/full/develop/dhis-core-version-master/developer-manual.html#request-parameters).
On top of that, request parameters were not properly validated and wrong values were just ignored and the default value was used insted. So when trying to use the dry-run, with `importMode=VALIDATION`, the importer was using the default value `COMMIT`.

The solution in this PR is meant to be temporary and it should be refactored when we have the `BindingException` handler in place. [DHIS2-13928](https://dhis2.atlassian.net/browse/DHIS2-13928)

[DHIS2-13928]: https://dhis2.atlassian.net/browse/DHIS2-13928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Documentation is fixed [here](https://github.com/dhis2/dhis2-docs/pull/1132)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataapproval/hibernate/HibernateDataApprovalStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataapproval/hibernate/HibernateDataApprovalStore.java
@@ -471,19 +471,16 @@ public class HibernateDataApprovalStore
                 ")";
         }
 
-        String readyBelowSubquery = "true"; // Ready below if this is the lowest
-                                           // (highest number) approval orgUnit
-                                           // level.
+        // Ready below if this is the lowest (highest number) approval level
+        String readyBelowSubquery = "true";
 
         if ( approvalLevelBelowOrgUnit != null )
         {
-            readyBelowSubquery = "not exists ( " + // Ready if nothing expected
-                                                  // below is
-                                                  // unapproved(/unaccepted)
+            // Ready if nothing expected below is unapproved(/unaccepted)
+            readyBelowSubquery = "not exists ( " +
                 "select 1 " +
-                "from organisationunit dao " + // Lower-level Data Approval
-                                                            // OrgUnit (DAO) where approval
-                                                            // is needed to be ready.
+                // Lower Data Approval OrgUnit (DAO) where approval is required
+                "from organisationunit dao " +
                 "where " + statementBuilder.position( "o.uid", "dao.path" ) + " = "
                 + pathPositionAtLevel( orgUnitLevel ) + " " +
                 "and dao.hierarchylevel = " + approvalLevelBelowOrgUnit.getOrgUnitLevel() + " " +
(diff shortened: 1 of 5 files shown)


teleivo (commented):
> ImportMode parameter in tracker importer had wrong values `[COMMIT, VALIDATE]` instead of `[COMMIT, VALIDATION]` as stated in the [docs](https://docs.dhis2.org/en/full/develop/dhis-core-version-master/developer-manual.html#request-parameters). On top of that, request parameters were not properly validated and wrong values were just ignored and the default value was used insted. So when trying to use the dry-run, with `importMode=VALIDATION`, the importer was using the default value `COMMIT`.
> 
> The solution in this PR is meant to be temporary and it should be refactored when we have the `BindingException` handler in place. [DHIS2-13928](https://dhis2.atlassian.net/browse/DHIS2-13928)

* api test shouldNotCommitWhenImportModeIsValidate is failing

The importMode for

* metadata is `COMMIT, VALIDATE`
* tracker is `COMMIT, VALIDATION`

Tracker is inconsistent on multiple accounts 😩 metadata vs tracker, verb vs noun.

Since such requests did not work until now we could choose the more consistent option of `COMMIT, VALIDATE`. So only update the docs and validate the enums. If we diverge in tracker it should be for a very good reason. This one does not seem to have one. I think there is use in users being able to transfer knowledge they have from metadata APIs to tracker 😄

enricocolasante:
> > ImportMode parameter in tracker importer had wrong values `[COMMIT, VALIDATE]` instead of `[COMMIT, VALIDATION]` as stated in the [docs](https://docs.dhis2.org/en/full/develop/dhis-core-version-master/developer-manual.html#request-parameters). On top of that, request parameters were not properly validated and wrong values were just ignored and the default value was used insted. So when trying to use the dry-run, with `importMode=VALIDATION`, the importer was using the default value `COMMIT`.
> > The solution in this PR is meant to be temporary and it should be refactored when we have the `BindingException` handler in place. [DHIS2-13928](https://dhis2.atlassian.net/browse/DHIS2-13928)
> 
> * api test shouldNotCommitWhenImportModeIsValidate is failing
> 
> The importMode for
> 
> * metadata is `COMMIT, VALIDATE`
> * tracker is `COMMIT, VALIDATION`
> 
> Tracker is inconsistent on multiple accounts weary metadata vs tracker, verb vs noun.
> 
> Since such requests did not work until now we could choose the more consistent option of `COMMIT, VALIDATE`. So only update the docs and validate the enums. If we diverge in tracker it should be for a very good reason. This one does not seem to have one. I think there is use in users being able to transfer knowledge they have from metadata APIs to tracker smile

Agreed

stian-sandvold (approved):

teleivo (approved):

PR #12911, opened 2023-02-04
feat: is( X in A, B, ... ) [DHIS2-14452]

jimgrace:
See [DHIS2-14452](https://dhis2.atlassian.net/browse/DHIS2-14452), especially [this comment](https://dhis2.atlassian.net/browse/DHIS2-14452?focusedCommentId=180081). This implements the syntax:

`is(X in 'A', 'B', ..., 'Z')`

which is equivalent to the following SQL logic (and in fact is translated into this for program indicators):

`X in ('A', 'B', ..., 'Z')`

[DHIS2-14452]: https://dhis2.atlassian.net/browse/DHIS2-14452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionIs.java
+++ b/dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionIs.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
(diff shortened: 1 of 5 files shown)


enricocolasante (approved):

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionIs.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public Object evaluateAllPaths( ExprContext ctx, CommonExpressionVisitor visitor )
+    {
+        List<Object> values = ctx.expr().stream()
+            .map( visitor::visit )
+            .collect( toList() );
+
+        Object arg0 = values.get( 0 );
+
+        for ( int i = 1; i < values.size(); i++ )
```
jbee, line 74: FYI: you can use streams with `.skip(1)` to find the match if you wanted

jbee (approved):

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionIs.java
jimgrace: Cool, I did not know that. Thanks!

Actually I think I might keep this the same even though the for loop is more clunky. I want the `evaluate` method to exit the loop as soon as it finds a match (not just for performance but there may be side-effects in evaluation of the missing value strategy). There may be some benefit in having the `evaluateAllPaths` logic look as similar as possible, although `.skip(1)` could have been used here as you say.

PR #12913, opened 2023-02-05
perf: Speed expressions via I18n cache [TECH-1504]

jimgrace:
@larshelge , @jbee, and @stian-sandvold: I am interested in what you think of the approach I've taken in this PR. The issue as described in [TECH-1504](https://dhis2.atlassian.net/browse/TECH-1504):

> Currently to prepare for every expression evaluation a call is made to `i18nManager.getI18n()` and the `I18n` object is stored in the `CommonExpressionVisitor` so it is available during the parsing of the expression. However it is currently only used to provide human-readable program variable names when constructing the description of a program indicator (and only for some program variables that have translations).
> 
> In a recent test of real-world heavy predictor use (many periods times many organisation units), it was found that about 44% of the predictor evaluation time was spent inside `i18nManager.getI18n()` to fetch `I18n` objects that were not needed. This task is to avoid making this call when it is not needed.
> 
> (The only thing that took longer than fetching the `I18n` values was the execution inside ANTLR code, which took slightly longer than getting the `I18n` object. When we switch to the new PEG parser to replace ANTLR this may have a similar performance benefit. Either of these gains by themself can nearly double expression performance by decreasing the time by 44% or so. But together they can increase performance by a factor of something like 7 or 8 times if they decrease the time by something like 85% to 88%.)

One approach would be to supply the `I18n` object only when parsing a program indicator expression to get its description, and maybe have a null value otherwise. But that introduces a dependency between supply and demand that would be broken should we decide to use `I18n` elsewhere in expression processing, and it means that the object reference will sometimes be null and sometimes not, which is not great.

This PR takes the approach that the `I18n` object can be "lazy" fetched only when needed. The challenge is that we can't just have `CommonExpressionVisitor` keep a reference to a `I18nManager` because the expression visitor is under `dhis-support` and doesn't know about services. But it can have a reference to `I18n` because it has access to `dhis-api`.

So I wrote `SingleValueCache` which delays getting, and then caches, a single value of any type. It be initialized inside a service that has access to `I18nManager` with code such as `new SingleValueCache<I18n>( i18nManager::getI18n )` and then the `I18n` instance can be retrieved when needed via `getI18nCache().get()`.

Let me know what you think about the "lazy" approach and the `SingleValueCache` implementation. I'm very open to suggestions if there is a better way.

P.S. SonarCloud doesn't like my `throw new RuntimeException( ex )` as we still do in some other parts of the code. Should we just ignore this or should I subclass `RuntimeException` as SonarLint suggests?

[TECH-1504]: https://dhis2.atlassian.net/browse/TECH-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbee (commented):
Generally if this improves the performance it should be worth doing. I suggest to have the implementation of your cache implement `Supplier` and only use `Supplier` type in the dependent code. The cache class itself is then just a cached supplier.

jimgrace:
@jbee Brilliant, thanks. In fact I looked around at `Supplier` stuff and found a Guava `Supplier` that does exactly what we need. So we don't even need the new class!!!

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/DefaultExpressionService.java
```diff
@@ -749,11 +750,13 @@ private <T extends IdentifiableObject> Map<String, T> getUidMap( Class<T> type,
      */
     private CommonExpressionVisitor newVisitor( ExpressionItemMethod itemMethod, ExpressionParams params )
     {
+        i18nManager.getI18n();
```
maikelarabori, line 753: Is this needed to force some initialization, or can be removed?

maikelarabori (approved):
It looks good @jimgrace. I just left a minor question.

Thanks

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/DefaultExpressionService.java
jimgrace: Oops, good catch @maikelarabori. This was a test statement I used to measure the performance difference with and without the fix, and I forgot to take it out. (It would have defeated the speed-up!) Taken out now. Thank you!!!

PR #12920, opened 2023-02-06
feat: OpenAPI YAML generation

jbee:
### Summary
* adds YAML support for OpenAPI spec.
* the `serverUrls` returned by the generated spec will use the same base URL as the server request used to generate
* the escaping of newlines is no longer done in `Descriptions` as JSON and YAML do them differently
* multi-line strings are handled differently to better support JSON and YAML differences

### Endpoints
Also adds new endpoints to make it easier to get a spec for a particular path:

    /api/{path}/openapi.json
    /api/{path}/openapi.yaml

For example to get a spec on the `/users` endpoint this becomes

    /api/users/openapi.json
    /api/users/openapi.yaml

Consequently to get a spec of all paths the endpoint is

    /api/openapi.json
    /api/openapi.yaml

The `/openapi` endpoints continue to provide explicit path and tag filter capability

    /api/openapi/openapi.json?tag=metadata
    /api/openapi/openapi.yaml?tag=metadata

### Manual Testing
Generating grouped JSON and YAML files and importing them in stoplight to see if errors are reported.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiAnalyse.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiAnalyse.java
@@ -337,7 +337,7 @@ final class ApiAnalyse
                 RequestBody a = p.getAnnotation( RequestBody.class );
                 Api.RequestBody requestBody = endpoint.getRequestBody()
                     .init( () -> new Api.RequestBody( p, a.required() ) );
-                Api.Schema type = analyseParamSchema( endpoint, p.getParameterizedType(), p.getType() );
+                Api.Schema type = analyseParamSchema( endpoint, p.getParameterizedType() );
                 consumes.forEach( mediaType -> requestBody.getConsumes().putIfAbsent( mediaType, type ) );
             }
             else if ( isParams( p ) )
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/JsonGenerator.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/JsonGenerator.java
@@ -27,11 +27,18 @@
  */
 package org.hisp.dhis.webapi.openapi;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.UnaryOperator;
 
+import lombok.Builder;
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
+
+import com.fasterxml.jackson.core.io.JsonStringEncoder;
 
 /**
  * A utility to generate JSON based on simple {@link Runnable} callbacks for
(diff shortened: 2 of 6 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiAnalyse.java
```diff
@@ -337,7 +337,7 @@ else if ( p.isAnnotationPresent( RequestBody.class ) )
                 RequestBody a = p.getAnnotation( RequestBody.class );
                 Api.RequestBody requestBody = endpoint.getRequestBody()
                     .init( () -> new Api.RequestBody( p, a.required() ) );
-                Api.Schema type = analyseParamSchema( endpoint, p.getParameterizedType(), p.getType() );
+                Api.Schema type = analyseParamSchema( endpoint, p.getParameterizedType() );
```
jbee, line 340: FYI: This is a bugfix: when the `p.getType()` was passed this was causing request bodies with parameterized types to not be fully mapped correctly.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/JsonGenerator.java
```diff
@@ -59,15 +70,81 @@
...
+            "", "", JsonGenerator::escapeYamlString,
+            "|-", "", true, JsonGenerator::escapeYamlText,
+            format -> format.toBuilder()
+                .itemIndent( "  " )
+                .newLineBeforeItem( true )
+                .newLineAfterArrayStart( false )
+                .newLineBeforeArrayEnd( false )
+                .newLineAfterObjectStart( false )
+                .newLineBeforeObjectEnd( false )
+                .build() );
```
claudemamo, line 103: Can we leverage Jackson's JsonNode so that we have an AST instead? Like this, we don't need to worry about YAML vs JSON, newlines, indentation, etc...

jbee: It is possible but not advisable. An intermediate representation was avoided specifically because of size/performance and the additional complexity. This generates YAML just fine. Involve jackson brings extra complexity and costs. Results might change based on version. Corner cases might not do what you want.

claudemamo: I can't understand how handling all the syntactic details of JSON and YAML brings less complexity than simply relying on JsonNode. It's not valid to argue that the JsonNode's behaviour might change between versions or that it might not fit some corner cases: we could make that argument for every dependency!

jbee: The main point is that we do not want to create a in-memory object tree representation for performance reasons 
and to the extend they being used here JSON and YAML are quite simple to generate. 

You are right that any dependency has that problem but no dependency does not so that is why I don't use one. Also going to a intermediate object graph hides subtle distinctions. For example, now we found that it makes sense to handle multi-line strings different from "simple" strings. This would have been lost in an object graph. And a transformer would have to work hard to go back to making some strings into a multi-line notation while keeping others in single line notation. Also which of the many multi-line notations would the transformer pick? 

I can't understand why you so desperately want to use the jackson hammer on this? The result is the same so why do you care as a consumer?

mortenoh (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/JsonGenerator.java
claudemamo: As a consumer I don't but as a code maintainer I do care.

larshelge (approved):

PR #12933, opened 2023-02-06
fix: category constraints are now added if needed [DHIS2-14241]

gnespolino:
### can only be merged once the release and control team has approved it

Categories on which to execute `getCanReadDimensionItems` should always be added when user is not superuser and no dimensionItems are present on them.


gnespolino:
Needs a fix. Closing.

PR #12936, opened 2023-02-07
feat: use Converter to parse fields request parameter DHIS2-13731

teleivo:
# Why

Clients control what fields they want in the response JSON via the query parameter `fields`. We use that knowledge in tracker to avoid fetching data from the DB that will not be returned in the JSON. Currently, parsing of the `fields` query parameter needs to be done twice as `FieldPath`'s cannot be passed into the field filtering service.

The current API of the `FieldFilterParser` is error-prone as devs need to know/remember how Spring handles request parameters containing characters like `[]`, and multiple `fields` parameters.

For example `@RequestParam List<String> fields` in a controller

will turn a request with `fields=attributes[attribute,value],deleted` into List

```java
0 = "attributes[attribute"
1 = "value]"
2 = "deleted"
```

Spring obviously cannot know the rules of our field filtering language. Spring splits nested fields in this case that belong to the same parent.

A request with `fields=attributes[attribute,value]&fields=deleted` results in

```java
0 = "attributes[attribute,value]"
1 = "deleted"
```

`FieldFilterParser` and its associated classes `FieldFilterParams` and `FieldFilterService` have been designed with Spring's behavior in mind.

The parser expects a singleton `Set` (not sure why) of all the comma-separated `fields` values. So a dev has to remember to join the elements split up by Spring back before parsing

```java
FieldFilterParser.parse( Collections.singleton( StringUtils.join( fields, "," ) ) );
```

Yes there is the `FieldFilterParams` helper to aid. Since its not the only way to use the parser its easy to forget this crucial step.

# What

* Spring Converter `FieldPathConverter` turns the query parameter `fields` into `List<FieldPath>`. The knowledge of how Spring handles request parameters with brackets `[]` is now contained here and in `FieldFilterParams`. `FieldFilterService` accepts `List<FieldPath>`, `String` or `FieldFilterParams`.

* FieldFilterParser only exposes

```java
public static List<FieldPath> parse( String fields )
public static List<FieldPath> parseWithPrefix( String fields, String prefix )
```

forcing you have to pass in the complete request parameter value `fields=attributes[attribute,value],deleted`.

# Future ideas

* I was not able to get rid of all `List<String> fields` query parameters. Some controllers check if a certain field filter is present in `fields` via string matching. We could at some point provide a `FieldPaths` wrapper of `List<FieldPath>` providing an API for finding such matches.
* Automatically filter responses of controllers which have a `@RequestParams List<FieldPath> fields`


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
```diff
@@ -76,7 +75,7 @@
 /**
  * @author Morten Olav Hansen
  */
-@Service
+@Service( "org.hisp.dhis.fieldfiltering.FieldFilterService" )
```
jbee, line 78: Do we need the name? We try to get away from explicit names.

jbee (approved):
I like it. 

Just using `String` and leave the splitting to the service would also have been fine IMO.
Having this more "build in" is both nice but also it is a form of magic complexity to deal with. From the PR I think it is worth trying if this is smoother.

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
teleivo: I thought this is our convention as per https://github.com/dhis2/wow-backend/blob/master/guides/spring_conventions.md#bean-id

Now that I re-read it I see this one is not implementing an interface 😅 I'll remove that change 👌🏻

mortenoh (approved):

PR #12945, opened 2023-02-08
Feature/im play

tonsV2:
This PR will make the pipeline deploy anything committed to master as an instance on IM.

The instance can be found here: https://play.im.prod.test.c.dhis2.org/dev

### DIFF
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -193,7 +193,7 @@ pipeline {
             }
             steps {
                 script {
-                    if (env.GIT_BRANCH == 'master') {
+                    if (env.GIT_BRANCH == 'feature/im-play') {
                         withCredentials([usernamePassword(credentialsId: 'dhis2-im-bot', passwordVariable: 'PASSWORD', usernameVariable: 'USER_EMAIL')]) {
                             dir('im-db-manager') {
                                 sparseCheckout('https://github.com/dhis2-sre/im-database-manager', 'master', '/scripts')
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -188,6 +188,7 @@ pipeline {
             environment {
                 HTTP = "http --check-status"
                 INSTANCE_HOST = "https://api.im.prod.test.c.dhis2.org"
+                INSTANCE_URL = "https://play.im.prod.test.c.dhis2.org/dev"
                 IMAGE_REPOSITORY = "core-dev"
                 IMAGE_TAG = "latest"
                 IMAGE_PULL_POLICY = "Always"
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -216,7 +216,7 @@ pipeline {
 
                                 dir('scripts') {
                                     echo 'Creating DHIS2 instance on IM...'
-                                    sh "(./findByName.sh play dev && ./update-dhis2.sh play dev) || ./deploy-dhis2.sh play dev"
+                                    sh "(./findByName.sh play dev && ./restart.sh play dev) || ./deploy-dhis2.sh play dev"
                                     waitFor.statusOk("${env.INSTANCE_URL}")
                                 }
                             }
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -29,6 +29,7 @@ pipeline {
     }
 
     stages {
+/*
         stage('Build') {
             steps {
                 implementIoBuildStarted(buildJob: true, buildName: "${STAGE_NAME}")
@@ -182,7 +183,7 @@ pipeline {
                 }
             }
         }
-
+*/
         stage('Update IM Play dev instance') {
             if (env.GIT_BRANCH == 'master') {
                 environment {
@@ -228,13 +229,13 @@ pipeline {
         always {
             archiveArtifacts artifacts: '**/target/surefire-reports/TEST-*.xml'
         }
-
+/*
         failure {
             script {
                 slack.sendFailureMessage()
             }
         }
-
+*/
         aborted {
             script {
                 slack.sendTimedoutMessage(30)
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -182,6 +182,41 @@ pipeline {
                 }
             }
         }
+
+        stage('Update IM Play dev instance') {
+            if (env.GIT_BRANCH == 'master') {
+                steps {
+                    script {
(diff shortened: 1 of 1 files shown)


jenkinsfiles/dev
```diff
@@ -182,6 +182,41 @@ pipeline {
...
+
+        stage('Update IM Play dev instance') {
+            if (env.GIT_BRANCH == 'master') {
+                steps {
+                    script {
+                        withCredentials([usernamePassword(credentialsId: 'e2e-im-user', passwordVariable: 'PASSWORD', usernameVariable: 'USER_EMAIL')]) {
+                            dir('im-db-manager') {
+                                sparseCheckout('https://github.com/dhis2-sre/im-database-manager', 'master', '/scripts')
+
+                                dir('/scripts') {
```
radnov, line 194: It should be just `scripts` here without the leading `/`.

jenkinsfiles/dev
```diff
@@ -182,6 +182,41 @@ pipeline {
...
+
+                                    sh '[ -n "$DATABASE_ID" ]'
+                                    echo "DATABASE_ID is ${env.DATABASE_ID}"
+                                }
+                            }
+
+                            dir('im-manager') {
+                                sparseCheckout('https://github.com/dhis2-sre/im-manager', 'master', '/scripts')
+
+                                dir('im-manager/scripts') {
```
radnov, line 209: It should be just `scripts` here without the leading `im-manager/`, as we're already in the `im-manager` dir at that point.

jenkinsfiles/dev
```diff
@@ -182,6 +182,41 @@ pipeline {
...
+                    script {
+                        withCredentials([usernamePassword(credentialsId: 'e2e-im-user', passwordVariable: 'PASSWORD', usernameVariable: 'USER_EMAIL')]) {
+                            dir('im-db-manager') {
+                                sparseCheckout('https://github.com/dhis2-sre/im-database-manager', 'master', '/scripts')
+
+                                dir('/scripts') {
+                                    env.DATABASE_ID = sh(
+                                            returnStdout: true,
+                                            // TODO: How do we avoid hard coding the name/version?
+                                            script: "./list.sh | jq -r '.[] | select(.Name == \"whoami\") | .Databases[] | select(.Name == \"Sierra Leone - 2.39.0.sql.gz\") | .ID'"
```
radnov, line 198: Name of the database should be something like “Sierra Leone - dev.sql.gz” (based on [the db manager seed script](https://github.com/dhis2-sre/im-database-manager/blob/master/scripts/seed.sh)); there’re no databases currently uploaded on prod, though.

jenkinsfiles/dev
```diff
@@ -182,6 +182,41 @@ pipeline {
...
+                                    echo "DATABASE_ID is ${env.DATABASE_ID}"
+                                }
+                            }
+
+                            dir('im-manager') {
+                                sparseCheckout('https://github.com/dhis2-sre/im-manager', 'master', '/scripts')
+
+                                dir('im-manager/scripts') {
+                                    echo 'Creating DHIS2 instance ...'
+                                    sh "INSTANCE_HOST=https://api.im.prod.test.c.dhis2.org IMAGE_REPOSITORY=core-dev IMAGE_TAG=latest IMAGE_PULL_POLICY=Always ./deploy-dhis2.sh play dev"
```
radnov, line 211: The environment variables can be declared within an `environment` block, like [here for example](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L64-L69)

jenkinsfiles/dev
```diff
@@ -182,6 +182,41 @@ pipeline {
                 }
             }
         }
+
+        stage('Update IM Play dev instance') {
+            if (env.GIT_BRANCH == 'master') {
+                steps {
+                    script {
+                        withCredentials([usernamePassword(credentialsId: 'e2e-im-user', passwordVariable: 'PASSWORD', usernameVariable: 'USER_EMAIL')]) {
```
radnov, line 190: Need to update that to `dhis2-im-bot` for the prod IM env.

radnov (changes requested):

jenkinsfiles/dev
tonsV2: How come it worked the other place? It's your code that I stole ;-)

jenkinsfiles/dev
tonsV2: Same as above

jenkinsfiles/dev
tonsV2: "Sierra Leone - 2.39.0.sql.gz" is the name of the database on the manager.

jenkinsfiles/dev
radnov: It didn't :D You managed to copy it while it still was WIP and not tested

jenkinsfiles/dev
```diff
@@ -182,6 +182,48 @@ pipeline {
...
+                HTTP = "http --check-status"
+                IM_HOST = "https://api.im.prod.test.c.dhis2.org"
+                INSTANCE_URL = "https://play.im.prod.test.c.dhis2.org/dev"
+                IMAGE_REPOSITORY = "core-dev"
+                IMAGE_TAG = "latest"
+                IMAGE_PULL_POLICY = "Always"
+            }
+            steps {
+                script {
+                    if (env.GIT_BRANCH == 'feature/im-play') {
```
radnov, line 197: Need to update this to `master`

radnov (changes requested):
Looks good to me, just one tiny but important detail noted.

jenkinsfiles/dev
tonsV2: Good point!

radnov (approved):

jenkinsfiles/dev
```diff
@@ -182,6 +182,48 @@ pipeline {
...
+                                }
+                            }
+
+                            dir('im-manager') {
+                                sparseCheckout('https://github.com/dhis2-sre/im-manager', 'master', '/scripts')
+
+                                dir('scripts') {
+                                    echo 'Creating DHIS2 instance on IM...'
+                                    sh "(./findByName.sh play dev && ./restart.sh play dev) || ./deploy-dhis2.sh play dev"
+                                    waitFor.statusOk("${env.INSTANCE_URL}")
```
teleivo, line 219: How long will this wait? And I assume it will result in a failed step if it times out, right?

teleivo (commented):
Nice 🤩 

How long does this stage take?

jenkinsfiles/dev
tonsV2: It'll now timeout after 5 minutes

teleivo (approved):

PR #12948, opened 2023-02-08
feat: GeoJSON import - validate and reduce coordinates to pairs [DHIS2-14634]

jbee:
### Summary
Validation of `coordinates`. Empty coordinates should result in validation error. Empty is `null`, `[]` or any level of array that in the end only contains `null`, `[]` as innermost element.

In addition the `coordinates` array is reduced to long/lat pairs in case there is a 3rd (or more) components, for example:

    [1]                  => [1]
    [1,2]                => [1,2]
    [1,2,3]              => [1,2]
    [[1,2,3],[4,5,6]]    => [[1,2],[4,5]]

### Automatic Testing
New unit test was added for low level coordinate handling in `CoordinateUtils`.
A controller test scenario was added for the validation and for the coordinate pair reduction.

### Manual Testing
* Import the example file from the ticket (not sure the target organisation units exist)
* Check the coordinates on the target and see that one long/lat was retained
* Check that the last entry with empty coordinates caused a validation error during import


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/geojson/CoordinatesUtils.java
```diff
@@ -0,0 +1,214 @@
...
+import org.hisp.dhis.jsontree.JsonArray;
+import org.hisp.dhis.jsontree.JsonObject;
+import org.hisp.dhis.jsontree.JsonValue;
+
+/**
+ * Utility to adjust GeoJSON coordinates of a geometry.
+ *
+ * @author Jan Bernitt
+ */
+@NoArgsConstructor( access = AccessLevel.PRIVATE )
```
larshelge, line 44: Nice way of avoiding a private constructor which will generate an IDE warning.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/geojson/CoordinatesUtils.java
```diff
@@ -0,0 +1,214 @@
...
+public final class CoordinatesUtils
+{
+    public static String geometryWithCoordinatePairs( JsonObject geometry )
+    {
+        if ( geometry.isUndefined() )
+            return geometry.node().getDeclaration();
```
larshelge, line 50: Add brackets even for one-liners. Yes is it not necessary but to make it consistent/follow code style.

larshelge (approved):

larshelge:
Looks good!

mortenoh (approved):

PR #12950, opened 2023-02-08
refactor: Attribute flag columns as JSONB [DHIS2-14606]

jbee:
### Summary
Replaces all boolean attribute flag columns with a single JSONB column that holds a set (array) of the `enum` names representing those former columns that were `true`.

Note that the getter/setter had to use `Set<String>` when directly mapped from `jsonb` column type without a custom mapper. 

### Preheat
The preheat was using dedicated methods to get attribute lists per target object class. One request for any, one for mandatory and one for unique attributes associated with a particular object types. So a preheat with _N_ object types would do `N * 3` DB requests. Instead we now do a single request for all attributes independent of number of object types imported (preheated). 

### Automatic Testing
A test already existed testing each attribute type. It was just changed to be a postgres integration controller test to test full circle.

### Manual Testing
* create an attribute
* edit the object types it can be used with and select a single one
* add a attribute value to one of the target object types

Do the above for each of the possible object types.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
@@ -29,10 +29,12 @@ package org.hisp.dhis.attribute;
 
 import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
 
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 import org.hisp.dhis.category.Category;
 import org.hisp.dhis.category.CategoryOption;
@@ -256,6 +258,18 @@ public class Attribute
         {
             objectTypes.remove( type );
         }
+    }
+
+    @JsonIgnore
+    public Set<String> getObjectTypes()
+    {
+        return objectTypes.stream().map( ObjectType::name ).collect( toSet() );
+    }
+
+    public void setObjectTypes( Set<String> objectTypes )
+    {
+        this.objectTypes.clear();
+        objectTypes.forEach( name -> this.objectTypes.add( ObjectType.valueOf( name ) ) );
     }
 
     @JsonProperty
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.attribute;
 
 import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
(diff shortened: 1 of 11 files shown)


netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
```diff
@@ -258,6 +260,18 @@ public void setAttribute( ObjectType type, boolean isAttribute )
...
 
+    @JsonIgnore
+    public Set<String> getObjectTypes()
+    {
+        return objectTypes.stream().map( ObjectType::name ).collect( toSet() );
+    }
+
+    public void setObjectTypes( Set<String> objectTypes )
+    {
+        this.objectTypes.clear();
```
larshelge, line 271: Is there some theoretical race condition potential here? I.e. two threads calling this non-synchronized method might override/clear the set.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/attribute/hibernate/HibernateAttributeStore.java
```diff
@@ -58,48 +53,4 @@ public HibernateAttributeStore( SessionFactory sessionFactory, JdbcTemplate jdbc
...
-        if ( !CLASS_ATTRIBUTE_MAP.containsKey( klass ) )
-        {
-            return new ArrayList<>();
-        }
-
-        CriteriaBuilder builder = getCriteriaBuilder();
-
-        return getList( builder, newJpaParameters()
-            .addPredicate( root -> builder.equal( root.get( CLASS_ATTRIBUTE_MAP.get( klass ) ), true ) )
-            .addPredicate( root -> builder.equal( root.get( "unique" ), true ) ) );
```
larshelge, line 103: Good to remove unused methods.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_13__Attribute_flags_as_jsonb.sql
```diff
@@ -0,0 +1,107 @@
+/* Converts numerous attribute boolean columns into a single jsonb column
+   with an array containing enum names of those columns that were true */
+
+/* create jsonb set column */
+alter table attribute add column if not exists objecttypes jsonb;
+
+/* make sure we have a function to create jsonb arrays without nulls */
+create or replace function jsonb_build_array_without_nulls(variadic anyarray)
```
larshelge, line 8: This is nice.

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/Attribute.java
jbee: The modified collection is not one that hibernate is aware of as we convert back and forth but you are right, this is a race condition should multiple threads use the same Attribute instance. Will think of a better way to do this.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_13__Attribute_flags_as_jsonb.sql
jbee: There is surprisingly little support for filtering nulls in the standard posgres functions. You can filter object members with null values but not arrays. And you cannot convert object keys to an array within an update. Some simple obvious operations do not yet exist in the standard functions, which is a bit surprising.

PR #12953, opened 2023-02-09
DO NOT MERGE: dependabot does not have access to secrets

teleivo:
this is a test if https://github.com/dhis2/dhis2-core/pull/12952
will work correctly on a fork. As we previously had the issue that forks do not have access to secrets.


teleivo:
I am seeing some weird interactions with the other non-fork branch 😂 closing for now

PR #12962, opened 2023-02-09
ci: don't login to Dockerhub master api tests

teleivo:
DOCKERHUB_PUSH env contains a string 'false' which is considered truthy by JS I assume JS is powering the action underneath. So true && 'false' resulted in 'false' in the if of the step. Since 'false' is truthy we did log in


radnov (approved):
🥴

PR #12971, opened 2023-02-09
fix: Return events if explicit fields provided [DHIS2-14720] [2.39.1]

muilpp:
Back port of https://github.com/dhis2/dhis2-core/pull/12961


muilpp:
this needs to be merged to 2.39.1.1 instead

PR #12973, opened 2023-02-09
ci: Remove timeout duration from slack message [skip ci]

radnov:
The overall pipeline timeout duration is 60m for the `stable` and `dev` pipelines and the timeout duration can vary based on multiple factors like inactivity (no new logs) for any of the stages/steps, as well as specific shorter timeouts like [here](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L219). 

i.e.  there are multiple factors that can change the final timeout duration and it's hard to predict, hence having a specific timeout duration in the message we send in Slack turns out to be inaccurate most of the time.

### DIFF
--- a/jenkinsfiles/canary
+++ b/jenkinsfiles/canary
@@ -107,7 +107,7 @@ pipeline {
 
         aborted {
             script {
-                slack.sendTimedoutMessage(30)
+                slack.sendTimedoutMessage()
             }
         }
     }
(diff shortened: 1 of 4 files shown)


teleivo:
Can you please explain why you remove the timeout? That's not clear to me 🙂

radnov:
> Can you please explain why you remove the timeout? That's not clear to me 🙂

I updated [the PR description](https://github.com/dhis2/dhis2-core/pull/12973#issue-1578314084), hopefully it makes more sense.

teleivo (approved):

PR #12976, opened 2023-02-09
chore: use feedback package web exceptions

jbee:
### Summary
Cleanup PR. When new checked exceptions in `feedback` were introduced there existed exceptions of similar or same name in web layer. These were not used because they could not be thrown by service layer. These were not replaced in the other PR to keep the change-set reasonable. This now cleans it up so just the newly introduced exceptions of that name are used.


vietnguyen (approved):

larshelge (approved):

larshelge:
Nice work! There is a conflict with a recently merged file/PR now @jbee .

PR #12984, opened 2023-02-10
fix: allow attribute UID in fields param for all overloaded methods

teleivo:
During https://github.com/dhis2/dhis2-core/pull/10790 (https://dhis2.atlassian.net/browse/DHIS2-13158) we added the ability to pass an attribute UID in the `fields` query parameter. There are multiple overloaded methods in `FieldFilterService` to achieve that only one of which was changed.


teleivo:
Closing in favor of https://github.com/dhis2/dhis2-core/pull/13001 which connects to the created jira ticket https://dhis2.atlassian.net/browse/DHIS2-14735.

PR #12989, opened 2023-02-10
ci: Disable concurrent builds for dev pipeline [skip ci]

radnov:
## Background

We have [2 triggers](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L10-L13) for the `dev` pipeline - one is a basic `cron` schedule (every 2 hours on weekdays), and the other one is polling GitHub for changes every 30 min. The polling is the "main" one, while the 2 hour schedule is only used for making sure that we also indirectly pick up any changes in the bundled apps, in case the main trigger hasn't picked up any new changes from `dhis2-core`. 

## Problem

However, this has the side-effect of sometimes triggering two concurrent builds (partially described in [DEVOPS-69](https://dhis2.atlassian.net/browse/DEVOPS-69)) which now can lead to issues with [the stage for (re-)deploying a DHIS2 instance with the Instance Manager](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L186). And in general, running two builds of the same commit concurrently is more often unneeded.

We don't have that problem with [the AWX stage](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L168), because jobs in AWX targeting the same host will be executed sequentially.

## Solution

We can disable concurrent builds with the [disableConcurrentBuilds()](https://www.jenkins.io/doc/book/pipeline/syntax/#available-options) option, which will put the builds in queue until the currently running one is complete.

[DEVOPS-69]: https://dhis2.atlassian.net/browse/DEVOPS-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonsV2 (approved):

teleivo (approved):
Makes sense!

PR #12994, opened 2023-02-12
feat: Predictor norm. dist. functions [DHIS2-14714]

jimgrace:


dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionNormDistAbstract.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public Object evaluate( ExprContext ctx, CommonExpressionVisitor visitor )
+    {
+        ExprContext expr0 = ctx.expr( 0 );
+
+        Double x = visitor.castDoubleVisit( expr0 );
+
+        // If the second argument is specified, use it as the mean of the normal
+        // distribution. Otherwise, compute the mean of the first argument.
+        Double mean = (ctx.expr().size() > 1)
```
maikelarabori, line 77: Do we always need to calculate/visit the `mean`? Otherwise, we could move it to line `92`, so it's computed only `stddev` is valid.

maikelarabori (approved):

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/function/FunctionNormDistAbstract.java
jimgrace: True @maikelarabori, good point! If stddev is invalid we don't have to compute the mean. However, (a) it will be very fast to compute the mean (everything's already in memory; no database or hibernate access needed), (b) an invalid stddev should be extremely rare, and (c) it might help anyone reading the code if the arguments are processed in order. For these reasons I think it's not worth trying to optimize performance in this case.

enricocolasante (approved):

PR #13001, opened 2023-02-13
fix: allow attribute UID in fields param for all FieldFilterService methods DHIS2-14735

teleivo:
During https://github.com/dhis2/dhis2-core/pull/10790 (https://dhis2.atlassian.net/browse/DHIS2-13158) we added the ability to pass an attribute UID in the `fields` query parameter. There are multiple methods in `FieldFilterService` to achieve that only one of which was changed.


jbee (approved):
I think this makes sense.

mortenoh (approved):

PR #13004, opened 2023-02-13
feat: non-null unique short name columns [DHIS2-14526]

jbee:
### Summary
Makes `shortName` a mandatory and unique property for the following object types:

* `ProgramTrackedEntityAttributeGroup`
* `DataSet`
* `OrganisationUnitGroup`
* `CategoryOption`
* `Constant`
* `DataElementGroup`

As a consequence any test setup creating or updating these objects needed to provide a `shortName`.

### Automatic Testing
Existing tests were adjusted.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
```diff
@@ -91,6 +91,7 @@ public CategoryOption()
     public CategoryOption( String name )
     {
         this.name = name;
+        this.shortName = name;
```
larshelge, line 94: I assume this is just for testing purposes @jbee ? This is a little bit risky perhaps as we might see unintended consequences. If it is for test purposes we might want to update `DhisConvenienceTest` methods. We need to coordinate with front-end/maintenance app I assume before merging this PR.

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v40/V2_40_16__Unique_non_null_short_names.java
```diff
@@ -0,0 +1,63 @@
...
+        throws Exception
+    {
+        copyUniqueValue( context, "program_attribute_group", "programtrackedentityattributegroupid", "name",
+            "shortname", 50 );
+        copyUniqueValue( context, "dataset", "datasetid", "name", "shortname", 50 );
+        copyUniqueValue( context, "orgunitgroup", "orgunitgroupid", "name", "shortname", 50 );
+        copyUniqueValue( context, "dataelementcategoryoption", "categoryoptionid", "name", "shortname", 50 );
+        copyUniqueValue( context, "constant", "constantid", "name", "shortname", 50 );
+        copyUniqueValue( context, "dataelementgroup", "dataelementgroupid", "name", "shortname", 50 );
+    }
```
larshelge, line 62: Looks good!

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
jbee: It is testing only but I can also update on the call side instead.

jbee: And the `default` option uses this. And lots of callers in tests.

netroms (approved):

PR #13007, opened 2023-02-13
fix: Check if dueDate param is null [DHIS2-14736]

muilpp:
There's an issue when updating events if the dueDate is not set in the payload. For some legacy reason we assumed it would always be set, in this PR we aim to solve this by checking if it's null first.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-14736

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
@@ -248,7 +248,7 @@ public class EventManager
             catch ( Exception e )
             {
                 handleFailure( workContext, importSummaries, events, IMPORT_ERROR_STRING, UPDATE );
-                log.error( "Failed to update events: " + e );
+                log.error( "Failed to update events: ", e );
             }
 
             final List<String> eventPersistenceFailedUids = importSummaries.getImportSummaries().stream()
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
@@ -248,6 +248,7 @@ public class EventManager
             catch ( Exception e )
             {
                 handleFailure( workContext, importSummaries, events, IMPORT_ERROR_STRING, UPDATE );
+                log.error( "Failed to update events: " + e );
             }
 
             final List<String> eventPersistenceFailedUids = importSummaries.getImportSummaries().stream()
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
```diff
@@ -248,6 +248,7 @@ public ImportSummaries updateEvents( final List<Event> events, final WorkContext
             catch ( Exception e )
             {
                 handleFailure( workContext, importSummaries, events, IMPORT_ERROR_STRING, UPDATE );
+                log.error( "Failed to update events: " + e );
```
teleivo, line 251: Please use https://www.javadoc.io/doc/org.slf4j/slf4j-api/1.7.25/org/slf4j/Logger.html#error(java.lang.String,%20java.lang.Throwable) instead of string concatenating the exception 😄

teleivo (commented):
Do I understand correctly that a `null` dueDate on the event is ok and will be persisted? 😄

muilpp:
> Do I understand correctly that a `null` dueDate on the event is ok and will be persisted? smile
As I understood, dueDate is not a mandatory parameter, so we'll check for null the same way we do for other date fields. Do you mean it should be mandatory?

teleivo:
> > Do I understand correctly that a `null` dueDate on the event is ok and will be persisted? smile
> 
> As I understood, dueDate is not a mandatory parameter, so we'll check for null the same way we do for other date fields. Do you mean it should be mandatory?

You are right, confirmed with @zubaira 😄

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
```diff
@@ -248,6 +248,7 @@ public ImportSummaries updateEvents( final List<Event> events, final WorkContext
             catch ( Exception e )
             {
                 handleFailure( workContext, importSummaries, events, IMPORT_ERROR_STRING, UPDATE );
```
teleivo, line 250: I just checked that `handleFailure` will add an error in the import summary.
Its called 3 times after catching an exception like this.

Why did you decide logging the exception? If it is valuable, should we do this in all 3 cases and thus move this into `handleFailure`?

muilpp: I added the log because in the summary we collect just a general error, but we are missing important information for us in regard to why did it fail. You are right we could add the same log to the `addEvents` method, to the `delete` method I think it's not really useful as it doesn't do as many things as the other two, would you agree?

teleivo (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/EventManager.java
muilpp: After a closer examination, the `addEvents` method is already catching a general Exception, so I don't think a log is useful there either, I'll leave it in the `updateEvents` method only, for now.

PR #13008, opened 2023-02-13
chore(WIP): don't mutate arguments in FieldPathHelper

teleivo:
- [ ] needs https://github.com/dhis2/dhis2-core/pull/13001

Mutating the `fields` parameter passed to the FieldFilterService is surprising and error prone. My expectation was that I could reuse the argument. FieldPathHelper is mutating it instead.


teleivo:
closing as I will include that change in a bigger PR related to tracker

PR #13012, opened 2023-02-13
fix: Remove hibernate validator dependency [DHIS2-13928]

enricocolasante:
`hibernate-validator` dependency was not really used in the system so it was removed. By consequence also `validation-api` can be removed.

Some tests needed to be fixed as having `hibernate-validator` in the classpath was hiding some errors. I didn't investigate in depth why, but some `not-null` values were set to null and tests were not working properly.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiToken.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiToken.java
@@ -36,6 +36,7 @@ import org.hisp.dhis.common.MetadataObject;
 import org.hisp.dhis.schema.PropertyType;
 import org.hisp.dhis.schema.annotation.Property;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
@@ -63,9 +64,7 @@ public class ApiToken extends BaseIdentifiableObject implements MetadataObject
 
     private List<ApiTokenAttribute> attributes = new ArrayList<>();
 
-    @JsonProperty
-    @JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
-    @Property( PropertyType.TEXT )
+    @JsonIgnore
     public String getKey()
     {
         return key;
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/mapping/MapController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/mapping/MapController.java
@@ -155,7 +155,7 @@ public class MapController
         Map newMap = deserializeJsonEntity( request );
         newMap.setUid( uid );
 
-        mergeService.merge( new MergeParams<>( newMap, map )
+        mergeService.merge( new MergeParams<>( map, newMap )
             .setMergeMode( params.getMergeMode() )
             .setSkipSharing( params.isSkipSharing() )
             .setSkipTranslation( params.isSkipTranslation() ) );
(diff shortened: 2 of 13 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/apikey/ApiToken.java
```diff
@@ -63,9 +64,7 @@ public ApiToken()
 
     private List<ApiTokenAttribute> attributes = new ArrayList<>();
 
-    @JsonProperty
-    @JacksonXmlProperty( namespace = DxfNamespaces.DXF_2_0 )
-    @Property( PropertyType.TEXT )
+    @JsonIgnore
```
enricocolasante, line 67: @netroms Please review carefully ApiToken and ApiTokenController.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/mapping/MapController.java
```diff
@@ -155,7 +155,7 @@ public WebMessage putJsonObject( @PathVariable String uid, @CurrentUser User cur
         Map newMap = deserializeJsonEntity( request );
         newMap.setUid( uid );
 
-        mergeService.merge( new MergeParams<>( newMap, map )
+        mergeService.merge( new MergeParams<>( map, newMap )
```
teleivo, line 158: how did the "wrong" order manifest itself? was that a bug? 😅

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/datavalue/DataSetValueQueryParams.java
```diff
@@ -57,7 +55,6 @@
 @AllArgsConstructor( access = AccessLevel.PRIVATE )
 public class DataSetValueQueryParams
 {
-    @NotBlank
```
teleivo, line 60: Just checking with @jbee 🤔 No mention of `NotBlank` in https://github.com/dhis2/wow-backend/blob/master/guides/preferred_patterns.md#use-nonnull-in-apis yet 😄 

Do we have a stance on this usage?

dhis-2/pom.xml
```diff
@@ -833,7 +830,6 @@
                 <RestrictImports>
                   <reason>Use javax.annotation.Nonnull and javax.annotation.CheckForNull instead</reason>
                   <bannedImports>
-                    <bannedImport>javax.validation.constraints.NotNull</bannedImport>
```
teleivo, line 836: Are you removing it since we removed the dependency that includes this class?
Just checking since I think this was added to guide us in https://github.com/dhis2/wow-backend/blob/master/guides/preferred_patterns.md#use-nonnull-in-apis

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/datavalue/DataSetValueQueryParams.java
jbee: I think we talked about `@Valid` which would be needed to give `@NotBlank` and other validation annotation an effect in spring layer. There was no agreement to use it.

dhis-2/pom.xml
jbee: I think we can remove it or keep it just to be safe. Both should work.

jbee (approved):
@netroms should have a look at the API token key using `@JsonIgnore` instead of overriding it with `null`. To me it looks good.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/mapping/MapController.java
enricocolasante: Yes it was a bug.
First parameter is used as a base and then all not null fields of the second object will be applied to the first one.
Using the wrong order was resulting in an object with some required fields as null, so persistence was failing.

dhis-2/pom.xml
enricocolasante: Removed because the dependency is not there anymore

netroms (approved):

PR #13028, opened 2023-02-14
feat: Add minor working list improvements [DHIS2-14367]

muilpp:
This is a continuation of the feature program stage working list, the [original PR](https://github.com/dhis2/dhis2-core/pull/12657) was merged a bit too son, in this one I'll tackle some pending minor improvements.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-14367

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHookTest.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHookTest.java
@@ -27,7 +27,16 @@
  */
 package org.hisp.dhis.dxf2.metadata.objectbundle.hooks;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.hisp.dhis.feedback.ErrorCode.E4062;
+import static org.hisp.dhis.feedback.ErrorCode.E4063;
+import static org.hisp.dhis.feedback.ErrorCode.E4064;
+import static org.hisp.dhis.feedback.ErrorCode.E4065;
+import static org.hisp.dhis.feedback.ErrorCode.E4066;
+import static org.hisp.dhis.feedback.ErrorCode.E4067;
+import static org.hisp.dhis.feedback.ErrorCode.E4068;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHookTest.java
```diff
@@ -149,10 +156,8 @@ void shouldFailWhenDatePeriodAbsoluteAndNoStartNorEndDateSpecifiedForEnrollmentO
 
         List<ErrorReport> errorReports = workingListHook.validate( programStageWorkingList, bundle );
 
-        assertEquals( 1, errorReports.size() );
-        assertEquals(
-            "Start date or end date have to be specified when date period type is set to ABSOLUTE for item `EnrollmentIncidentDate`",
-            errorReports.get( 0 ).getMessage() );
+        ErrorReport expectedErrorReport = new ErrorReport( this.getClass(), E4062, "EnrollmentIncidentDate" );
```
teleivo, line 159: Does the value of the first parameter `this.getClass()` matter to our test? If not we could set it to null. If that would fail because its accessed I would wrap that in a helper to highlight whats important like `error(E4062, "EnrollmentIncidentDate")`. Hiding the fact that we have to pass `this.getClass()`.

The signatures you chose in `assertErrorReport` taking an expected error hides a bit what you are asserting on and it forces us to repeat a bit what we do in our implementation.

For these reasons I rather use something like

`assertHasError(errors, Code code, String messageContains )`


So I don't have to deal with creating an error in the test and figuring out which fields I need to set on the error 😅 

Maybe others have a different view on this 😄 👌🏻

teleivo (approved):
Nice! Thank you for following up 😄 
I would write the assertion slightly different but I'll leave it up to you if you agree.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/test/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramStageWorkingListObjectBundleHookTest.java
muilpp: Makes sense to have the test as much decoupled as possible from the implementation.

PR #13048, opened 2023-02-16
fix: Handle BindException [DHIS2-13928]

enricocolasante:
Handling `BindException` in `CrudControllerAdvice` and return a BAD_REQUEST if any request parameter conversion fails (in the specific a `TypeMismatchException` is thrown ).

`BindException` is also handling validation errors managed through `hibernate-validator` framework that is not currently used in DHIS2.


jbee (approved):
Nice!

larshelge (approved):

larshelge:
Nice fix.

PR #13049, opened 2023-02-16
fix: Characters such as ":" (colon) ";" (semi-colon) and probably others not accepted in Tracker capture Attributes [ DHIS2-10215 ]

lucaCambi77:

Currently, if we use a filter in a query or for attributes and the filter has a value for example, like:project:x it will fail and eventually split at the first filter value column (:), missing the remaining part.
With this issue, we add a regex to check and split accordingly so we can now get any chars sequence after the operator. This applies to `TrackerTrackedEntityCriteriaMapper` and `TrackerEventCriteriaMapper`.

This issue also simplifies  the `TrackedEntityInstanceServiceTest` as it was way too verbose and complex to follow


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapper.java
```diff
@@ -124,7 +123,7 @@ public TrackedEntityInstanceQueryParams map( TrackerTrackedEntityCriteria criter
             orgUnits.addAll( user.getOrganisationUnits() );
         }
 
-        QueryFilter queryFilter = getQueryFilter( criteria.getQuery() );
+        QueryFilter queryFilter = pareQueryFilter( criteria.getQuery() );
```
enricocolasante, line 126: Typo

```suggestion
        QueryFilter queryFilter = parseQueryFilter( criteria.getQuery() );
```

enricocolasante (approved):

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
```diff
@@ -574,4 +574,51 @@ void testMappingOrderParamsGivenInvalidField()
...
+        assertEquals( "Query item or filter is invalid: " + TEA_1_UID + ":lke:value", exception.getMessage() );
+    }
+
+    @Test
+    void shouldCreateQueryFiltersWhenQueryHasOperatorAndValueWithDelimiter()
+        throws ForbiddenException,
+        BadRequestException
+    {
+        criteria.setQuery( "like:project:x" );
+        TrackedEntityInstanceQueryParams params = mapper.map( criteria );
```
teleivo, line 609: dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java has dedicated tests for `parseQueryItem`. Should we move some of these over there so that we do not duplicate tests? And maybe only keep one test in the Mappers to test the integration between `RequestParamUtils` and the mappers.

teleivo (approved):
Nice 😄 

I added some minor comments for javadocs and test organization.

PR #13029, opened 2023-02-14
fix: tracker does not return fields in accordance with field filtering DHIS2-14783

teleivo:
* Add `FieldFilterService.filterIncludes(Class rootClass, List<FieldPath> filter, String path)` to answer the question "is path `x.y.z` included in given `filter`". Reusing the same logic that determines which fields are included when applying the filter.
* Don't mutate the `fields` parameter passed to the FieldFilterService. It's surprising and error prone if the argument is reused.
* Adapt tracker controllers to use `filterIncludes`

PR #13051, opened 2023-02-16
fix: tracker does not return fields in accordance with field filtering DHIS2-14783

teleivo:
* Add `FieldFilterService.filterIncludes(Class rootClass, List<FieldPath> filter, String path)` to answer the question "is path `x.y.z` included in given `filter`". Reusing the same logic that determines which fields are included when applying the filter.
* Don't mutate the `fields` parameter passed to the FieldFilterService. It's surprising and error prone if the argument is reused.
* Adapt tracker controllers to use `filterIncludes`

# Open Question

`FieldPathHelper.applyExclusions` only excluded exact matches of exclusion field paths from given fieldPathMap. Children of excluded field paths remained in the fieldPathMap. This works since Jackson will ignore children without parents when serializing to JSON.

So

there is no need to remove children without parents from `List<FieldPath> filter` when

1. serializing

but there is a need for it when 

2. in `filterIncludes` when determining whether a field will be serialized or not. Here there is no data and so no involvement of Jackson.

Should the removal of such children be done

A. in `FieldPathHelper.applyExclusions` - con: not needed for serialization, pro: ties methods together that should rely on the same logic 
B. in `FieldPathHelper.filterIncludes`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
+++ b/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
@@ -132,30 +132,25 @@ public class FieldFilterService
     /**
      * Determines whether given path is included in the resulting ObjectNodes
      * after applying {@link #toObjectNode(Object, List)}. This obviously
-     * requires that the actual data contains such path.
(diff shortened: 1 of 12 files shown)


teleivo:
I put my jira issue in the wrong project. I thought I have to recreate a new PR from a new base branch matching the jira issue but jira seems to pick it up correctly anyway. So closing this one to keep the review on the original PR https://github.com/dhis2/dhis2-core/pull/13029

PR #13066, opened 2023-02-17
fix: Incorrect pager and data from dataElementOperands endpoint [DHIS2-14727]

d-bernat:
please see: https://dhis2.atlassian.net/browse/DHIS2-14727

Solution:
Line String deg = CollectionUtils.popStartsWith( filters, "dataElement.dataElementGroups.id:eq:" );
removes all related dataElementGroups from filter collection used later by calling calculatePaginationCountKey (List<String> filters).
The key returned by this function is used for caching of total count. The key is not unique when the filter collection is empty and pager does not work properly. Instead of modified content of filter collection the WebOption instance with unmodified filters is used.


jbee (changes requested):
I think this code was already hard to understand due to objects being manipulated in a mutable way that should have been treated as immutable. The `filters`, the `dataElementOperands` being reassigned. They act both as input and later as result. 

I think it is time to not make it worse. We should clean this up. If filters are changed because a group is split away this should result in a new instance of a changed filter set. Either use the original filter set to get your correct cache key or also include other inputs that are derived.

d-bernat:
> I think this code was already hard to understand due to objects being manipulated in a mutable way that should have been treated as immutable. The `filters`, the `dataElementOperands` being reassigned. They act both as input and later as result.
> 
> I think it is time to not make it worse. We should clean this up. If filters are changed because a group is split away this should result in a new instance of a changed filter set. Either use the original filter set to get your correct cache key or also include other inputs that are derived.

gnespolino:
> I think this code was already hard to understand due to objects being manipulated in a mutable way that should have been treated as immutable. The `filters`, the `dataElementOperands` being reassigned. They act both as input and later as result.
> 
> I think it is time to not make it worse. We should clean this up. If filters are changed because a group is split away this should result in a new instance of a changed filter set. Either use the original filter set to get your correct cache key or also include other inputs that are derived.



> I think this code was already hard to understand due to objects being manipulated in a mutable way that should have been treated as immutable. The `filters`, the `dataElementOperands` being reassigned. They act both as input and later as result.
> 
> I think it is time to not make it worse. We should clean this up. If filters are changed because a group is split away this should result in a new instance of a changed filter set. Either use the original filter set to get your correct cache key or also include other inputs that are derived.

This seems to be a small fix, I'm not sure one should address a refactor within the same ticket of this bug. I'd suggest to live it as it is, and create a proper ticket to make it better, with a proper description.

gnespolino (approved):

maikelarabori (approved):

maikelarabori:
I agree that we need to improve this code, but I believe it would be better done as part of a separate Jira.

jbee (approved):

PR #13067, opened 2023-02-17
fix: Pagination flag and unlimited

maikelarabori:
This PR fixes the pagination when analytics unlimited is enabled.
It needs to be taken into consideration when the Grid is being created.
The changes will improve the logic on top of a recent code we introduced for handling the `lastPage` flag in analytics events/enrollments.


gnespolino (approved):
good job!

d-bernat (approved):

PR #13071, opened 2023-02-18
fix: MAX/MIN_SUM_ORG_UNIT w/filter [DHIS2-14430]

jimgrace:
See [DHIS2-14430](https://dhis2.atlassian.net/browse/DHIS2-14430), especially the [February 14, 2023 comment from Blake Hansen](https://dhis2.atlassian.net/browse/DHIS2-14430?focusedCommentId=180557). The original commit for this feature did not work properly when periods were used in the filter.

This PR fixes the problem and further simplifies the subquery for MAX_SUM_ORG_UNG and MAX_SUM_ORG_UNG. The subquery generated by `getMinOrMaxValueSubquerySql` now fetches the same dimensions as the outer query with the addition of `OU`, so the subquery will find the min or max within an orgUnit and the main query will sum across orgUnits as appropriate. To do this, the `WHERE` clause from `getWhereClause` is brought into the subquery instead of being in the main query. This simplifies a number of things:

1. Periods are filtered in the subquery so, for example, if non-consecutive periods are chosen only values from those periods will be included, such as selecting only the first and last quarters of a year and finding the min/max monthly values within those quarters.
2. The min and max subquery can operate across calendar year partition boundaries, for example finding the maximum monthly value within a financial year.
3. Values needed to filter on in the `WHERE` clause don't have to be returned also by the subquery including time-related values like period and year which would have to be replaced by dummy values so the outer `WHERE` clause would find the values and accept them.
4. Tests for the earliest start date and latest end date (or for the periods) don't have to be hand-constructed within the subquery because they are added automatically when generating the `WHERE` clause.

At first I started refactoring the code by making the subquery more complex, hand-crafting period filters within the subquery `WHERE` clause, and assigning a dummy value to the `year` column so it would pass through to the outside `WHERE` clause, etc. In my next refactor I constructed a stripped-down `DataQueryParams` object that would contain only those time-related dimensions that needed filtering within the subquery, so I could call `getWhereClause` within the subquery with a simpler `params` that filtered only what was needed (which would then be filtered again by the main query `WHERE` clause). I discovered that I needed to put values like `aggregationType` into the stripped-down params or `getWhereClause` wouldn't work. This seemed fragile. Finally I realized that if I moved the whole `WHERE` clause into the subquery it actually simplified a lot as mentioned above, made the code smaller, and was less fragile.

The only awkwardness is that the `WHERE` clause now has to be added to the main query only if it is not already in the subquery. This is currently done with the test:

`            if ( !params.getAggregationType().isMinOrMaxInPeriodAggregationType() )`

I thought about creating a convenience method, something like `params.whereClauseInSubquery()` which would do the above test, but this is not really a property of the `DataQueryParams`; it is more an implementation strategy within `JdbcAnalyticsManager`. I thought also of creating an additional method within `JdbcAnalyticsManager` to just make this test, but in my opinion the awkwardness of the above test is less bad than creating an extra method just to hide it. I also feel that the benefits of moving the `WHERE` clause into the subquery outweigh the awkwardness of this test. But I'm open to considering other opinions. (I even tried moving this test to inside `getWhereClause` but quickly realized that if I did so it wouldn't run within the subquery!!!)

There might be benefits as well by refactoring `getFirstOrLastValueSubquerySql` in the future to move the `WHERE` clause into that subquery also. I chose not to do this now to keep the functionality of this PR more bounded and since it seemed a non-zero risk of breaking code that currently works.

I added some tests to `AnalyticsServiceTest` that succeed with this PR but failed without it.

[DHIS2-14430]: https://dhis2.atlassian.net/browse/DHIS2-14430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


maikelarabori (approved):

larshelge (approved):

larshelge:
Nice work and thanks for handling this @jimgrace . Approved the PR. There is one conflict now which needs to be resolved.

d-bernat (approved):

PR #13082, opened 2023-02-21
refactor: improve and simplify assertMapEquals

jimgrace:


jbee (approved):
I am happy with this but @teleivo should also be happy with the error messages the change generates. The explicit version was mostly build so one does get descriptive error messages.

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java
```diff
@@ -78,18 +78,17 @@ public static <E> void assertContainsOnly( Collection<E> expected, Collection<E>
...
-        {
-            assertEquals( e.getValue(), actual.get( e.getKey() ),
-                String.format( "Expected value not found in %s", actual ) );
-        }
-        for ( Map.Entry<K, V> e : actual.entrySet() )
-        {
-            assertEquals( e.getValue(), expected.get( e.getKey() ),
-                String.format( "Did not expect value in %s", actual ) );
-        }
+        assertContainsOnly( expected.entrySet(), actual.entrySet() );
```
teleivo, line 91: I just looked at the error messages using

```java
@Test
void foo() {
  assertMapEquals(
          Map.of("color", "red", "day", "monday", "month", "july"),
          Map.of("color", "red", "temperature", "28C")
  );
}
```

assertion message

before

```  
org.opentest4j.AssertionFailedError: Expected value not found in {temperature=28C, color=red} ==> 
Expected :monday
Actual   :null
```

after the change

```
org.opentest4j.AssertionFailedError: Expected [month=july, day=monday] to be in [color=red, temperature=28C] ==> 
Expected :true
Actual   :false

org.opentest4j.AssertionFailedError: Expected [temperature=28C] NOT to be in [color=red, temperature=28C] ==> 
Expected :true
Actual   :false

org.opentest4j.AssertionFailedError: Expected [color=red, month=july, day=monday] or actual [color=red, temperature=28C] contains unexpected duplicates ==> 
Expected :3
Actual   :2
```

I think the error message are clearer than before 😄 

One thing we might want to adjust is the last assertion message https://github.com/dhis2/dhis2-core/blob/c0eee70357cedf67fdaccea7ec37d9c8a22642ab/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java#L78

`Expected [color=red, month=july, day=monday] or actual [color=red, temperature=28C] contains unexpected duplicates`

We have the check for size to catch the case where the actual contains the expected values but either of them has a duplicated value. Since `assertContainsOnly( Collection<E> expected, Collection<E> actual )` deals with `Collection` the implementation might be one that does not allow duplicates. In that case using the word `duplicates` is misleading.

I think it would be good enough to adjust the wording. Maybe

```java
"Collection sizes mismatch. Expected %s, got %s."
```

This way we can figure out if there is a duplicate element we are missing. In case there are no duplicates the message can be considered redundant but it does not hurt.

teleivo (approved):
Nice 😄 
That shows the power of reusable assertions 😋

larshelge (approved):

PR #13098, opened 2023-02-22
fix: Remove ProgramTrackedEntityAttribute entity [DHIS2-14815]

larshelge:
Removes the `ProgramTrackedEntityAttribute` entity from the system including store, service, Hibernate mapping, controller. This entity is not in use.


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_19__Remove_table_trackedentityattributegroup.sql
```diff
@@ -0,0 +1,4 @@
+
+drop table programtrackedentityattributegroupmembers;
```
jbee, line 2: `if exists` to be safe?

jbee (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_19__Remove_table_trackedentityattributegroup.sql
larshelge: You are right.

PR #13099, opened 2023-02-22
feat: Use rule engine version using expression parser [TECH-1491]

enricocolasante:
Use new version of rule engine library that is not using ANTLR parser anymore.
Now ANTLR library is only used by `dhis-antlr-expression-parser` so we do not need to declare it as a dependency.

### DIFF
--- a/dhis-2/pom.xml
+++ b/dhis-2/pom.xml
@@ -64,7 +64,7 @@
     <!-- *Dependencies* -->
 
     <!-- DHIS2 Rule Engine -->
-    <dhis2-rule-engine.version>2.1.2</dhis2-rule-engine.version>
+    <dhis2-rule-engine.version>2.1.3-SNAPSHOT</dhis2-rule-engine.version>
 
     <!-- HISP Quick and Staxwax -->
     <dhis-hisp-quick.version>1.4.1</dhis-hisp-quick.version>
--- a/dhis-2/pom.xml
+++ b/dhis-2/pom.xml
@@ -64,7 +64,7 @@
     <!-- *Dependencies* -->
 
     <!-- DHIS2 Rule Engine -->
-    <dhis2-rule-engine.version>2.0.47</dhis2-rule-engine.version>
+    <dhis2-rule-engine.version>2.1.2</dhis2-rule-engine.version>
 
     <!-- HISP Quick and Staxwax -->
     <dhis-hisp-quick.version>1.4.1</dhis-hisp-quick.version>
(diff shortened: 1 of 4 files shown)


dhis-2/pom.xml
```diff
@@ -64,7 +64,7 @@
     <!-- *Dependencies* -->
 
     <!-- DHIS2 Rule Engine -->
-    <dhis2-rule-engine.version>2.0.47</dhis2-rule-engine.version>
+    <dhis2-rule-engine.version>2.1.2</dhis2-rule-engine.version>
```
enricocolasante, line 67:
```suggestion
    <dhis2-rule-engine.version>2.1.3-SNAPSHOT</dhis2-rule-engine.version>
```

larshelge (approved):

jimgrace:
@enricocolasante Note that several core classes still reference `antlr4-runtime` directly:

```
CommonExpressionVisitor
FunctionOrgUnitAncestor
FunctionOrgUnitDataSet
FunctionOrgUnitGroup
FunctionOrgUnitProgram
```
Also a number of core classes reference the ANTLR tool generated classes in `dhis-support-antlr-expression-parser / target / generated-sources`.

Does this mean we should still keep the `antlr4-runtime` reference in our maven build? To my surprises `mvn clean install` still works in the PR branch, perhaps because of the following?

`              <ignoredUsedUndeclaredDependency>org.antlr:antlr4-runtime</ignoredUsedUndeclaredDependency>`

jbee (approved):

jbee:
@jimgrace I think this was just about rule engine usage from core no longer being ANTLR dependent. The other usages from predictors and co still need it. Or am I missing what was change here?

enricocolasante:
> @enricocolasante Note that several core classes still reference `antlr4-runtime` directly:
> 
> ```
> CommonExpressionVisitor
> FunctionOrgUnitAncestor
> FunctionOrgUnitDataSet
> FunctionOrgUnitGroup
> FunctionOrgUnitProgram
> ```
> 
> Also a number of core classes reference the ANTLR tool generated classes in `dhis-support-antlr-expression-parser / target / generated-sources`.
> 
> Does this mean we should still keep the `antlr4-runtime` reference in our maven build? To my surprises `mvn clean install` still works in the PR branch, perhaps because of the following?
> 
> ` <ignoredUsedUndeclaredDependency>org.antlr:antlr4-runtime</ignoredUsedUndeclaredDependency>`

We generally prefer to declare transitive dependencies in our pom, it means that if we use a dependency that is imported by another dependency we still declare it, even if it will work without the declaration.
It is the case for ANTLR, as it will be not used anymore at some point in the future and ANTLR is very sensitive about versions and now it is only used by `dhis-antlr-expression-parser`, it can be removed and let it be managed by the dependency.

Now we can upgrade ANTLR in `dhis-antlr-expression-parser` without worrying about dhis-core consequences and dependabot is much happier

jimgrace:
@enricocolasante Great, thanks for the explanation. Yes, it is far better to not have to keep the `antlr-runtime` versions in sync between the two repos, which we have not always done very well!

PR #13102, opened 2023-02-22
fix: provide displayName for all data integrity checks [DHIS2-14770]

jbee:
### Manual Testing
* open scheduler app
* add new job of type "data integrity"
* select "only selected checks"
* select one of the checks 
* safe

Verify that the app does not crash when opening the selection for checks


Philip-Larsen-Donnelly (approved):
I think this fix is fine for avoiding the crash, but we may want to keep a ticket open to use the correct displayName eventually?

PR #13111, opened 2023-02-23
fix: error when request with filter userAttribute:eq:true

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14825

### DIFF
--- a/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/AbstractCrudControllerTest.java
+++ b/dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/AbstractCrudControllerTest.java
@@ -781,6 +781,24 @@ class AbstractCrudControllerTest extends DhisControllerConvenienceTest
 
     }
 
+    @Test
+    void testFieldFilterWithAttribute()
+    {
+        Attribute attribute = createAttribute( 'A' );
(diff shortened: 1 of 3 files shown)


jbee (approved):

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/AbstractCrudControllerTest.java
```diff
@@ -781,6 +781,19 @@ void testFieldsFilterWithAttributeValues()
...
+    void testFieldFilterWithAttribute()
+    {
+        Attribute attribute = createAttribute( 'A' );
+        attribute.setDataElementAttribute( true );
+        manager.save( attribute );
+
+        JsonList<JsonIdentifiableObject> response = GET(
+            "/attributes?fields=id,name&filter=dataElementAttribute:eq:true" )
+                .content().getList( "attributes", JsonIdentifiableObject.class );
+        assertEquals( attribute.getUid(), response.get( 0 ).getId() );
```
jbee, line 794: I think to be safe we need a test setup with another attribute created that is not of the type we want to filter out.

vietnguyen: added one more test case now

PR #13119, opened 2023-02-24
feat: Analytics query endpoint default time dimension and default sort order [DHIS2-14805]

d-bernat:
please see https://dhis2.atlassian.net/browse/DHIS2-14805 and https://dhis2.atlassian.net/browse/DHIS2-14811


maikelarabori (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/PeriodCriteriaUtils.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Helper class for period criteria
+ */
+public class PeriodCriteriaUtils
+{
+
+    private PeriodCriteriaUtils()
+    {
+        throw new IllegalStateException( "Utility class" );
+    }
```
gnespolino, line 48: we can add `@NoArgsConstructor(access=PRIVATE)` rather than this, to avoid boilerplate code

gnespolino (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/PeriodCriteriaUtils.java
d-bernat: sorry, it was on automerge, will be done in backport and next cases

PR #13120, opened 2023-02-24
fix: ignore api files exteral as redirect after login

netroms:


mortenoh (approved):

jbee (approved):

dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/security/config/DhisWebCommonsWebSecurityConfig.java
```diff
@@ -307,7 +307,9 @@ public DefaultAuthenticationSuccessHandler authenticationSuccessHandler()
         public MappedRedirectStrategy mappedRedirectStrategy()
         {
             MappedRedirectStrategy mappedRedirectStrategy = new MappedRedirectStrategy();
-            mappedRedirectStrategy.setRedirectMap( Map.of( "/dhis-web-commons-stream/ping.action", "/" ) );
+            Map<String, String> redirectMap = Map.of( "/dhis-web-commons-stream/ping.action", "/",
```
larshelge, line 310: Perhaps move those URLs to a static final Map at the top of the class to make it easy to find and add more URLs later?

PR #13133, opened 2023-02-24
fix: Characters such as ":" and ";" not accepted in Tracker capture Attributes [DHIS2-10215]

lucaCambi77:

**Solution**

1)  splitting and grouping by the regex delimiter + operator + delimiter and collect operators and values
2)  operator -> value pair can be created positionally

This is not covering the edge case of a value that is exactly `delimiter + operator + delimiter`

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
@@ -30,7 +30,9 @@ package org.hisp.dhis.webapi.controller.tracker.export;
 import static org.hisp.dhis.common.DimensionalObject.DIMENSION_NAME_SEP;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
```diff
@@ -68,22 +70,10 @@ private RequestParamUtils()
...
-    private static final String OPERATOR_VALUE_ITEM_FILTER_REG_EX = "(?i)(" + OPERATOR_GROUP + ")" +
-        DIMENSION_NAME_SEP + "(.(?i)(?!" + OPERATOR_GROUP + "))+";
-
-    private static final Pattern OPERATOR_VALUE_ITEM_FILTER_VALIDATION_PATTERN = Pattern
-        .compile(
-            "(.*)" + DIMENSION_NAME_SEP + OPERATOR_VALUE_ITEM_FILTER_REG_EX );
-
-    private static final Pattern OPERATOR_VALUE_ITEM_FILTER_PATTERN = Pattern
-        .compile( OPERATOR_VALUE_ITEM_FILTER_REG_EX );
+    private static final Pattern QUERY_FILTER_OPERATOR_PATTERN = Pattern
```
stian-sandvold, line 73: Could we keep the comment above to explain the RegEx?

stian-sandvold (approved):

PR #13140, opened 2023-02-25
feat: Data Element Group Predictions [DHIS2-14635]

jimgrace:
See [DHIS2-14635](https://dhis2.atlassian.net/browse/DHIS2-14635). This PR allows a prediction to be made for each data element in a data element group.

After discussing syntax alternatives with @jbee, we decided on a pre-processor for prediction expressions that could be adapted in the future for other uses as well (for predictors or other expression types). It involves very simple Java parsing of a header section of an expression before calling the main parser for the rest of the expression. Three classes do most of the work:

1. `PreprocessorExpression` does basic parsing for any type of expression having a preprocessor header.
2. `PredictorPreprocessorExpression` extends this to parse a predictor expression with a preprocessor header.
3. `PredictionPreprocessor` handles variable substitution and generates one transient predictor for each data element in the data element group. It also validates and gets an expression description for predictor expressions that might include a preprocessor header.

`DefaultExpressionService.getExpressionItemDescriptions` is added as a convenience for `PredictionPreprocessor` to generate expression descriptions. (The code was split out of `DefaultExpressionService.getExpressionDescription`.)

`DefaultPredictionService` was enhanced to call `PredictionPreprocessor` and to provide expression validation and description methods.

`PredictorController` was altered to call `DefaultPredictionService` for predictor expression validation and description.

Of special note is the transaction handling for `DefaultPredictionService.predictSimple`. This is tricky because of the way predictors use both Hibernate transactions and `BatchHandler`. Where possible, predicted values are output using BatchHandler, which does direct SQL INSERT statements outside of the Hibernate transaction. This is a tremendous speed boost. But sometimes predicted datavalues are written to a period that does not yet exist. When periods need to be created for these datavalues they are done so through Hibernate, but this is in the context of the Hibernate transaction so the BatchHandler cannot yet INSERT datavalues for these new periods. (It gets a missing foreign key record error.) To avoid this problem, data for any new periods that are created during the prediction run are entered through Hibernate instead of BatchHandler.

The trouble with having multiple predictions as part of the same transaction is this: If one prediction run needs to output datavalues for periods that do not yet exist, it will create those periods through Hibernate and output these datavalues also through Hibernate. However if a subsequent prediction is under the same transaction and needs to output datavalues to these same periods, it will find that the periods already exist. So it will try to output the datavalues via BatchHandler, which will fail because the periods don't yet exist outside of the continuing Hibernate transaction.

To avoid this problem, `DefaultPredictionService.predictSimple` is tagged with

`    @Transactional( propagation = Propagation.REQUIRES_NEW )`

so it will run in a separate transaction (while the outer transaction is suspended). The final challenge is that when invoking a method from within the same class, Spring transaction processing is not involved. This is overcome by `DefaultPredictionService.predict` getting the bean from Spring for the service and calling `predictSimple` using that. Thanks to [this Stack Overflow answer](https://stackoverflow.com/questions/37217075/spring-nested-transactions/48649581#48649581) for pointing me in the right direction!

With this fix, if one prediction run creates one or more new periods and writes new datavalues to these period through Hibernate, then ends the transaction, then the next prediction run will find these periods preexisting and can use BatchHandler. So it ends up being efficient for all subsequent prediction runs.

Even before this PR, this may have been a problem in running multiple predictors and predictor groups. This new code should fix that as well. I intend to backport this transactional processing to all prior supported versions. (The reason I didn't do this first as a fix and then do feature later is I didn't want to jeopardize this new feature making it through review in time for 2.40, which it has been promised for.)

The new `PredictionServiceTest.testPredictionPreprocessor` method tests for this condition by writing into a not-yet-existing period. It fails without the aforementioned transaction processing and succeeds with it.

[DHIS2-14635]: https://dhis2.atlassian.net/browse/DHIS2-14635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
```diff
@@ -0,0 +1,99 @@
...
+
+        if ( parts.size() != 4 )
+        {
+            throw new ParserException( format( "Predictor preprocessor expression should have four parts: '%s'",
+                expression ) );
+        }
+
+        // forEach
+
+        if ( !parts.get( 0 ).equals( "forEach" ) )
```
larshelge, line 69: Do we have a constant or central place for these keywords?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param expression the expression to parse for preprocessing.
+     */
+    public PredictorPreprocessorExpression( String expression )
+    {
+        super( expression );
+
+        if ( parts.size() != 4 )
+        {
+            throw new ParserException( format( "Predictor preprocessor expression should have four parts: '%s'",
```
larshelge, line 63: Do we want to use an `ErrorCode` for these exceptions to allow programmatic handling on the client-side?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PreprocessorExpression.java
```diff
@@ -0,0 +1,136 @@
...
+
+import org.hisp.dhis.antlr.ParserException;
+
+/**
+ * Parser for expression preprocessing.
+ *
+ * @author Jim Grace
+ */
+@Getter
+public abstract class PreprocessorExpression
```
larshelge, line 45: Should we possibly extend `ErrorCodeException`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PreprocessorExpression.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Validates a UID.
+     *
+     * @param uid the UID to validate.
+     * @param expression the full expression.
+     */
+    protected void validateUid( String uid, String expression )
+    {
+        if ( !UID_PATTERN.matcher( uid ).matches() )
```
jbee, line 129: Can we use `CodeGenerator.isValidUid` here?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
```diff
@@ -0,0 +1,99 @@
...
+import org.hisp.dhis.antlr.ParserException;
+
+/**
+ * Parser for predictor expression preprocessing.
+ *
+ * @author Jim Grace
+ */
+@Getter
+public class PredictorPreprocessorExpression
+    extends PreprocessorExpression
```
jbee, line 43: I don't think we should start with using inheritence here. I'd join these to into just `PredictorExpression`. That these can have a preprocessor part is a detail. I'm not sure we want to can share any of that logic with other usages should they arise. We should have a look at that if we get there.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/DefaultPredictionService.java
```diff
@@ -192,6 +199,44 @@ private List<Predictor> fetchPredictors( List<String> predictors, List<String> p
 
     @Override
     public void predict( Predictor predictor, Date startDate, Date endDate, PredictionSummary predictionSummary )
+    {
+        for ( Predictor p : preprocessor.preprocess( predictor ) )
+        {
+            // Trigger a new transaction when calling self (must be inside loop)
+            PredictionService self = applicationContext.getBean( PredictionService.class );
+
+            self.predictSimple( p, startDate, endDate, predictionSummary );
```
jbee, line 208: I know why but it screams bad design :D
Lets put a short comment here also that we need to go through the proxy for transactions.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionPreprocessor.java
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Gets the data element group for preprocessing.
+     */
+    private DataElementGroup getDeg( String deUid, String expression )
+    {
+        DataElementGroup deg = idObjectManager.get( DataElementGroup.class, deUid );
+
+        if ( deg == null )
+        {
+            throw new ParserException(
```
jbee, line 184: Any ID lookup should result in a `org.hisp.dhis.feedback.NotFoundException` if we don't want to catch it so it ends up as the response to the client.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionPreprocessor.java
```diff
@@ -0,0 +1,205 @@
...
+     * Gets one data element from the data element group.
+     */
+    private DataElement getOneDataElement( DataElementGroup deg, String expression )
+    {
+        Set<DataElement> dataElements = deg.getMembers();
+
+        if ( dataElements.isEmpty() )
+        {
+            throw new ParserException(
+                format( "DataElementGroup '%s' is empty when evaluating '%s'", deg.getName(), expression ) );
```
jbee, line 200: This I would use `org.hisp.dhis.feedback.ConflictException`. Again assuming this should end up as the response to the client. A "parse exception" should only be used to indicate that the syntax was wrong.

jbee (commented):
About the transaction handling: Could we identify all missing periods in advance (a pre-loop of sorts)? Could we then go and create them in an independent transaction that is complete once the "create periods" call is made? And after that contrinue adding data values always using the batch hanler as now all periods should exist? I'd think that would be less intertwined.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
jimgrace: Good point. I've now defined a constant for "forEach", but also moved all references to that string into the same class. Following @jbee's suggestion there is now a single `PredictorExpression` class that models predictor expressions both with and without needing preprocessing, so "forEach" doesn't have to to be known in different classes.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
jimgrace: Yes, that would be nice. As it stands now, all the parsing uses `ParserException` including many parsing errors that are generated from the [dhis2-antlr-expression-parser](https://github.com/dhis2/dhis2-antlr-expression-parser) repo that was written before the backend error codes were invented and doesn't have access to the backend error codes anyway. I'm hoping that when @jbee and I integrate the PEG parser into core we can find a more clever way to have error codes throughout the parsing errors including from another repo, something like a namespace or numeric range perhaps.

It's even worse than that. All the user sees in the Maintenance app for a bad expression is "Expression is not well-formed". They never see even these messages. When the ANTLR parsing code was released there wasn't even room in the WebAPI for an expression error message to be returned to the frontend. There seems to be now but it's always just the text for the `ExpressionValidationOutcome` enum and the only values used are `VALID` and `EXPRESSION_IS_NOT_WELL_FORMED`. We will need to overhaul this as well when the PEG parser is integrated so we can have meaningful code and meaningful text both go across the WebAPI.

I don't think it makes sense at this point to have these parsing exceptions different from the others; we need to overhaul them all.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PreprocessorExpression.java
jimgrace: Perhaps so. See previous comment.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PreprocessorExpression.java
jimgrace: Right you are! Done!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
jimgrace: OK, done that way. There is now just a single `PredictorExpression` class which holds both preprocessor predictor expressions and non-preprocessor predictor expressions, so it is much less awkward and easier to use.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/DefaultPredictionService.java
jimgrace: I've expanded the comment to talk about proxy as well as transaction. Yeah, it does seem a bit convoluted.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionPreprocessor.java
jimgrace: I tried that, and it got real messy with declaring the `throws` on each of the nested calls. All the other expression description processing generates runtime errors only and so this needed special handling. As a compromise I changed this to `IllegalStateException` for now which is also used elsewhere in expression description generation and is trapped well in the methods returning expression status.

Speaking of which, currently in the controllers for expression descriptions we first do the check to see if an expression is valid which in turn calls the expression description generation. Then if it is valid, we call the expression description generation again to get the expression. Hopefully when we integrate the PEG parser we can make this much less convoluted and just generate the expression or the exception in one go and not have to do things twice.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/predictor/PredictionPreprocessor.java
jimgrace: Also changed to `IllegalStateException` for now, and also hope we can revisit all of the exception generation when we integrate the PEG parser.

jimgrace:
@jbee commented:

> About the transaction handling: Could we identify all missing periods in advance (a pre-loop of sorts)? Could we then go and create them in an independent transaction that is complete once the "create periods" call is made? And after that contrinue adding data values always using the batch hanler as now all periods should exist? I'd think that would be less intertwined.

I like this idea a lot, now that I know it's possible to complete a transaction and keep going all within a service. What can be done is to find all the periods that _could_ be predicted into by a predictor and create them, then go back and make the predictions. This does not guarantee that all these periods _will_ be predicted into because it's possible that the predictor expression's `MissingValueStrategy` will cause no prediction to be written into a period. But if we create some periods in the database that have no datavalues I don't think this is a problem and it allows us to use the `BatchHandler` for all predictions, which performs better and requires less special-casing.

Another approach would be to bypass Hibernate for any new periods and create them directly through SQL. This would avoid altogether having to re-enter the controller through the Spring bean, and only periods having datavalues would have to be created.

I'm not sure which of these approaches I like better, but I like them both better than what we do now, and we should consider changing to one of them in the future.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorPreprocessorExpression.java
jimgrace: I've created constants for these keywords now. Also I've taken @jbee's suggestion and created a single PredictorExpression class for both predictor expressions requiring preprocessing and those not requiring preprocessing. This means that all the decisions about the "forEach" keyword are now in one class.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PreprocessorExpression.java
jimgrace: Right you are! Done!

maikelarabori (approved):

PR #13141, opened 2023-02-26
chore: deploy instance with TTL set to 10 years (approx.) [skip ci]

tonsV2:
The default TTL of an instance deployed using the instance manager is 48 hours. This PR will deploy an instance with a TTL of 10 years thus preventing the instance from being garbage collected soon than that

### DIFF
--- a/jenkinsfiles/dev
+++ b/jenkinsfiles/dev
@@ -194,6 +194,7 @@ pipeline {
                 IMAGE_PULL_POLICY = "Always"
                 FLYWAY_MIGRATE_OUT_OF_ORDER = "true"
                 FLYWAY_REPAIR_BEFORE_MIGRATION = "true"
+                INSTANCE_TTL = "315360000"
             }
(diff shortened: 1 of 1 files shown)


jenkinsfiles/dev
```diff
@@ -194,6 +194,7 @@ pipeline {
                 IMAGE_PULL_POLICY = "Always"
                 FLYWAY_MIGRATE_OUT_OF_ORDER = "true"
                 FLYWAY_REPAIR_BEFORE_MIGRATION = "true"
+                INSTANCE_TTL = "315360000"
```
radnov, line 197: 10 years seems a bit excessive 😄, but it shouldn't really matter.

radnov (approved):

jenkinsfiles/dev
tonsV2: Time flies :smile:

PR #13147, opened 2023-02-27
fix: Return an error when parameter is null [TECH-1502]

enricocolasante:
Convert parameters of type `TrackerIdSchemeParam` using a `PropertyEditor` instead of a `Converter`.
What I understood looking at Spring code is that when parsing parameters Spring will first try to use a `PropertyEditor` that is specific for the MVC layer and it make sense for us to throw an exception if the parameter is null.
Before we were using a `Converter` that is a global object used by Spring and even though in our case was not used anywhere else than the MVC layer, it was conceptually wrong to raise an exception for null values.

On the same logic, I changed the `PropertyEditor` for Enums to raise an exception if the value is empty.

When parsing the paramter Spring will try to use the `PropertyEditor` and it will catch an `IllegalArgumentException` or `ConversionException` and wrap it in a `TypeMismatchException` that we are already handling in `CrudControllerAdvice`

We may need to agree on naming conventions for this part, I didn't put a lot of thoughts in it yet

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
@@ -78,7 +78,7 @@ import org.hisp.dhis.webapi.controller.exception.MetadataImportConflictException
 import org.hisp.dhis.webapi.controller.exception.MetadataSyncException;
 import org.hisp.dhis.webapi.controller.exception.MetadataVersionException;
 import org.hisp.dhis.webapi.controller.exception.NotAuthenticatedException;
-import org.hisp.dhis.webapi.controller.tracker.imports.StringToTrackerIdSchemeParamConverter;
+import org.hisp.dhis.webapi.controller.tracker.imports.IdSchemeParamEditor;
 import org.hisp.dhis.webapi.security.apikey.ApiTokenAuthenticationException;
 import org.hisp.dhis.webapi.security.apikey.ApiTokenError;
 import org.springframework.beans.TypeMismatchException;
@@ -134,7 +134,7 @@ public class CrudControllerAdvice
         binder.registerCustomEditor( Date.class, new FromTextPropertyEditor( DateUtils::parseDate ) );
         binder.registerCustomEditor( IdentifiableProperty.class, new FromTextPropertyEditor( String::toUpperCase ) );
         this.enumClasses.forEach( c -> binder.registerCustomEditor( c, new ConvertEnum( c ) ) );
-        binder.registerCustomEditor( TrackerIdSchemeParam.class, new StringToTrackerIdSchemeParamConverter() );
+        binder.registerCustomEditor( TrackerIdSchemeParam.class, new IdSchemeParamEditor() );
     }
 
     @ExceptionHandler( org.hisp.dhis.feedback.BadRequestException.class )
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -248,6 +267,10 @@ private String getGenericFieldErrorMessage( String fieldType )
 
     private String getNotValidValueMessage( Object value, String field )
     {
+        if ( value == null || "".equals( value ) )
```
teleivo, line 270: You might want to follow GitHubs advice. See code scanning notice above 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/StringToTrackerIdSchemeParamConverter.java
```diff
@@ -27,22 +27,23 @@
...
+import java.beans.PropertyEditorSupport;
+
 import org.hisp.dhis.common.CodeGenerator;
 import org.hisp.dhis.tracker.TrackerIdScheme;
 import org.hisp.dhis.tracker.TrackerIdSchemeParam;
-import org.springframework.core.convert.converter.Converter;
 import org.springframework.util.StringUtils;
 
-public class StringToTrackerIdSchemeParamConverter implements Converter<String, TrackerIdSchemeParam>
+public class StringToTrackerIdSchemeParamConverter extends PropertyEditorSupport
```
teleivo, line 37: Even if we have no convention maybe call it `Editor` instead of `Converter` 🤷🏻

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
```diff
@@ -103,7 +104,7 @@ void shouldReturnABadRequestWhenInvalidValueForIntegerInBindingObjectIsPassedAsP
             .param( "integerNumber", "10.5" ) )
             .andExpect( content().string( containsString( "Bad Request" ) ) )
             .andExpect( content().string( containsString(
-                "Value 10.5 is not valid for parameter integerNumber. It should be of type Integer" ) ) );
+                "Value 10.5 is not valid for parameter integerNumber. For input string: \"10.5\"" ) ) );
```
teleivo, line 107: This suffix seems redundant to the beginning of the message 🤔 It would not help me as a user to know why `10.5` is not valid and what I should do about it. Where does this one come from?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/StringToTrackerIdSchemeParamConverter.java
enricocolasante: Done

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
enricocolasante: I was reluctant because I needed to cast to String but I followed github advice

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
enricocolasante: Maybe this needs some agreement on the types that we use as parameters.
The part of the message `It should be of type X` would make sense for some types but not for others that are just implementation details.
As now, we are just checking if the type is primitive and in that case telling the user to use that type, in all the other cases we are returning the message generated by Spring

teleivo (approved):

PR #13151, opened 2023-02-27
fix: Add pagination to Potential Duplicate Get Query [ 2.38 ]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/12891


lucaCambi77:
For now, we keep it only in 2.40 as it is not an active endpoint

PR #13171, opened 2023-02-28
fix: Org unit descendant check using path property [TECH-1511]

larshelge:
Introduces a faster way of performing the org unit "is descendant" check, and applies it to the analytics security manager. Instead of performing the check in a transactional method which fetches the org unit from the database, the check is done locally to the org unit by comparing the org unit ancestor `path` property directly with the ancestor identifiers. This saves a transaction and database lookup per check. In particular for maps and pivot tables with hundreds or thousands of org units, this yields a huge performance benefit.

The performance improvement only applies for users which do not have the `ALL` authority and do have at least one data view org unit.

YourKit CPU profiling:

analytics-is-descendant

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
@@ -613,7 +613,7 @@ public class OrganisationUnit
      * ancestor org units.
      *
      * @param ancestors the collection of ancestor org units.
-     * @return true if his org unit is a descendant of the ancestors.
+     * @return true if this org unit is a descendant of the ancestors.
      */
     public boolean isDescendant( Collection<OrganisationUnit> ancestors )
     {
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
@@ -621,9 +621,7 @@ public class OrganisationUnit
             return false;
         }
 
-        Set<String> ancestorUids = IdentifiableObjectUtils.getUidsAsSet( ancestors );
-
-        return ancestorUids.stream()
+        return IdentifiableObjectUtils.getUidsAsSet( ancestors ).stream()
             .anyMatch( uid -> path.contains( uid ) );
     }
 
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
```diff
@@ -605,6 +607,24 @@ public void updateParent( OrganisationUnit newParent )
...
+     */
+    public boolean isDescendant( Collection<OrganisationUnit> ancestors )
+    {
+        if ( isEmpty( ancestors ) )
+        {
+            return false;
+        }
+
+        return IdentifiableObjectUtils.getUidsAsSet( ancestors ).stream()
+            .anyMatch( uid -> StringUtils.contains( path, uid ) );
```
jbee, line 625: I think it is better to use 
```java
return ancestors.stream().map(OrganisationUnit::getUid).anyMatch(uid -> path.contains(uid));
```
This avoids building the set of UIDs and as `path` and `uid` should be non-null we can use `contains` directly.

Or if we assume this might be called with non persisted OUs
```java
return ancestors.stream()
  .map(OrganisationUnit::getUid)
  .filter(Objects::nonNull)
  .anyMatch(uid -> path.contains(uid));
```
This would highlight how this situation is handled. Also something we might want to mention in the javadoc.

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
larshelge: Agreed, I had refactored some code in the "previous" isDescendant method and missed that we create two streams here, good catch.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
```diff
@@ -605,6 +607,24 @@ public void updateParent( OrganisationUnit newParent )
         newParent.getChildren().add( this );
     }
 
+    /**
+     * Indicates whether this org unit is a descendant of any of the given
+     * ancestor org units.
+     *
+     * @param ancestors the collection of ancestor org units.
+     * @return true if his org unit is a descendant of the ancestors.
```
maikelarabori, line 616:
```suggestion
     * @return true if this org unit is a descendant of the ancestors.
```

maikelarabori (approved):

PR #13198, opened 2023-03-02
fix: correct eventDate in NTI [DHIS2-14836]

zubaira:
executionDate or eventDate is stored in the database as a timestamp. But when we fetch it from the database using JDBC connection then resultSet.getDate( "executionDate" ) trims the timestamp part from the date. For example `2023-03-23T12:23:00.000` will be returned as `2023-03-23`. This is fixed by using resultSet.getTimestamp


larshelge:
Could you add a description with some context on why this is the correct date @zubaira ?

zubaira:
> Could you add a description with some context on why this is the correct date @zubaira ?

executionDate or eventDate is stored in the database as a timestamp. But when we fetch it from the database using JDBC connection then resultSet.getDate( "executionDate" ) trims the timestamp part from the date. For example 2023-03-23T12:23:00.000 will be returned as 2023-03-23. This is fixed by using resultSet.getTimestamp

teleivo (approved):
Did you go through all date/timestamp related columns to check if we made the same mistake elsewhere?

stian-sandvold (approved):

jason-p-pickering:
It is a timestamp in the DB @zubaira , but at least in this instance (2.39.1.1) there is never any HH:MM:SS contained in any of the events exeuction date

executiondate | 2020-09-01 00:00:00
completeddate | 2020-09-01 05:12:13.371
created       | 2020-09-02 07:59:37.126
lastupdated   | 2020-09-02 07:59:37.18

They are always truncated to 00:00:00. I suppose this could be a client issue only sending the date, but is it possible that these may have been unintentionally migrated to dates during some flyway/upgrade process?

PR #13207, opened 2023-03-02
feat: filter apps api by current user authority [DHIS2-14750]

amcgee:
~Includes #13208, can be deduplicated if that is merged first~
Fixes [DHIS2-14750](https://dhis2.atlassian.net/browse/DHIS2-14750)

This filters the list of apps returned by `/api/apps` to only show those apps accessible to the current user.  In addition, it does the following:

1. Correct handling of App Manager authority - previously there was an App Manager authority (`M_dhis-web-maintenance-appmanager`) separate from the App Management App (`M_dhis-web-app-management`).  However the former no longer exists and cannot be assigned to users.  It was still being used in a few places to gate access to the full list of installed applications.  This change updates those gates to use the App Management App authority, which is also what allows a user to install apps.
2. Correct handling of new-style plugins when listing / searching for dashbaord items - previously this only supported apps of type DASHBAORD_WIDGET
3. Fixed /api/apps/menu API which includes all bundled applications (coming from the list we maintain in AppManager)

Related to #13152 

[DHIS2-14750]: https://dhis2.atlassian.net/browse/DHIS2-14750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppManager.java
```diff
@@ -240,4 +209,63 @@ Resource getAppResource( App app, String pageName )
...
+     *
+     * @return list of installed apps with given name
+     */
+    public static List<App> filterAppsByShortName( final String name, Collection<App> apps, final String operator )
+    {
+        return apps.stream()
+            .filter( app -> (("ilike".equalsIgnoreCase( operator )
+                && app.getShortName().toLowerCase().contains( name.toLowerCase() )) ||
+                ("eq".equalsIgnoreCase( operator ) && app.getShortName().equals( name ))) )
+            .collect( Collectors.toList() );
```
netroms, line 255: Very similar to the above method "filterAppsByName()", maybe break it down to a common method that use name, app property and operator as input.

netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/AppMenuManager.java
```diff
@@ -55,32 +47,25 @@
...
     private final I18nManager i18nManager;
 
-    private final UserService userService;
-
     private final LocaleManager localeManager;
 
-    private final List<WebModule> menuModules = new ArrayList<>();
+    private final List<WebModule> menuModules;
+
+    private final AppManager appManager;
```
netroms, line 58: This can be removed since it is not in use anymore.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
```diff
@@ -104,28 +102,22 @@
...
+    public @ResponseBody Map<String, List<WebModule>> getWebModules( HttpServletRequest request )
     {
         String contextPath = ContextUtils.getContextPath( request );
 
-        return Map.of( "modules", getAccessibleAppMenu( contextPath, username ) );
+        return Map.of( "modules", getAccessibleAppMenu( contextPath ) );
     }
 
-    public List<WebModule> getAccessibleAppMenu( String contextPath, String username )
+    public List<WebModule> getAccessibleAppMenu( String contextPath )
```
netroms, line 113: Method can be private.

KaiVandivier (approved):
Looks good to me! 🙂

PR #13208, opened 2023-03-02
fix: remove username request parameter in get modules servlet

netroms:


amcgee:
@netroms thanks for this!  I added a couple more commits to add on what you started - I've simplified the AppMenuManager logic by using the static list of bundled apps from AppManager.  This means the `/api/apps/menu` endpoint now works in both Jetty and Tomcat/Struts environments.  I also added a few handy helpers to `CurrentUserUtil` for getting and checking current user authorities, and I used these helpers instead of loading the entire current User object.  This, in combination with #13152 and #13207 will be very helpful to have in 2.40 so that we can start to migrate away from `getModules.action`!

PR #13210, opened 2023-03-02
feat: add Metadata package import changelog [DHIS2-14460]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14460
### Overview
- This PR introduce a new `MetadataPackage` object. 
- This object is created only when user import metadata payload which includes a `package` object like below
```
{
  "package": {
    "DHIS2Build": "5d136cb",
    "DHIS2Version": "2.36.6",
    "code": "CRVS_VE",
    "description": "Vital Events",
    "lastUpdated": "20220120T140039",
    "locale": "en",
    "name": "CRVS_VE_TKR_1.1.0_DHIS2.36.6-en",
    "type": "TKR",
    "version": "1.1.0"
  }
}
```
- A new api endpoint `api/metadataPackages` also added. This endpoint only support `GET` method as the `MetadataPackage` object can only be created when importing metadata via `api/metadata` endpoint.
- The `MetadataPackage` record will only be saved to database if the import process completed successfully. Otherwise the import error is returned and the package is not created.
- The metadata payload will also be saved to property `MetadataPackage.importFile` which is a reference to a `FileResource` in DHIS2. It will only be saved if the `MetadataPackage` is created successfully.
- If user import a metadata package with package.name existed and package version is the same then an error will be be thrown. In this case, if user need to increase the version of the metadata package before importing.


netroms (approved):

larshelge (changes requested):
Looks good. Let's discuss the data model a bit.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.40/V2_40_20__Add_table_metadata_package.sql
```diff
@@ -0,0 +1,25 @@
...
+    dhis2build VARCHAR(10) NOT NULL,
+    locale VARCHAR(10) NOT NULL,
+    importfile int8,
+    lastupdatedby int8,
+
+    CONSTRAINT metadatapackageid PRIMARY KEY (metadatapackageid),
+    CONSTRAINT fk_metadatapackage_fileresourceid FOREIGN KEY (importfile) REFERENCES fileresource (fileresourceid),
+    CONSTRAINT fk_metadatapackage_createdby_userinfoid FOREIGN KEY (createdby) REFERENCES userinfo(userinfoid),
+    CONSTRAINT fk_metadatapackage_lastupdateby_userinfoid FOREIGN KEY (lastupdatedby) REFERENCES userinfo(userinfoid) );
+
```
larshelge, line 21: Remove blank lines.

vietnguyen:
Need to finalize the specification for this feature. Close the PR for now.

PR #13212, opened 2023-03-03
fix: Createdbyuserinfo and lastupdatedbyuserinfo jsonb fields in programinstance are all blank (Old Tracker Capture)

lucaCambi77:
Set programinstance's lastUpdatedUserInfo when :

- creating enrollments

- adding or updating the Enrollment events. This is due to /events endpoint invoked even though we are on the enrollment page

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
@@ -607,6 +607,7 @@ public abstract class AbstractEnrollmentService
         }
 
         programInstance.setCreatedByUserInfo( UserInfoSnapshot.from( importOptions.getUser() ) );
+        programInstance.setLastUpdatedByUserInfo( UserInfoSnapshot.from( importOptions.getUser() ) );
 
         programInstanceService.addProgramInstance( programInstance, importOptions.getUser() );
 
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventStore.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventStore.java
@@ -34,6 +34,7 @@ import java.util.Set;
 import org.hisp.dhis.dxf2.events.report.EventRow;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.program.ProgramStageInstance;
+import org.hisp.dhis.program.UserInfoSnapshot;
 import org.hisp.dhis.user.User;
 
 /**
@@ -69,6 +70,8 @@ public interface EventStore
     List<EventRow> getEventRows( EventSearchParams params, List<OrganisationUnit> organisationUnits );
 
     int getEventCount( EventSearchParams params, List<OrganisationUnit> organisationUnits );
+
+    void updateEnrollmentsLastUpdatedUserInfo( List<String> enrollmentUids, UserInfoSnapshot userInfoSnapshot );
 
     /**
      * Delete list of given events to be removed. This operation also remove
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
@@ -2074,6 +2074,21 @@ public class JdbcEventStore implements EventStore
             PreparedStatement::execute );
     }
 
+    @Override
+    public void updateEnrollmentsLastUpdatedUserInfo( List<String> enrollmentUids, UserInfoSnapshot userInfoSnapshot )
+    {
+        List<List<String>> uidsPartitions = Lists.partition( Lists.newArrayList( enrollmentUids ), 20000 );
+
+        PGobject userInfoToJson = userInfoToJson( userInfoSnapshot, jsonMapper );
+
+        uidsPartitions.forEach( uids -> jdbcTemplate.execute(
+            "update programinstance set lastupdatedbyuserinfo = :lastupdatedbyuserinfo where uid IN (:uids)",
+            new MapSqlParameterSource()
+                .addValue( "lastupdatedbyuserinfo", userInfoToJson )
+                .addValue( "uids", uids ),
+            PreparedStatement::execute ) );
+    }
+
     private void bindEventParamsForInsert( PreparedStatement ps, ProgramStageInstance event )
         throws SQLException,
         JsonProcessingException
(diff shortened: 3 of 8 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/EventImportTest.java
```diff
@@ -244,67 +246,137 @@ protected void setUpTest()
...
-    void testAddEventWithDueDateForProgramWithoutRegistration()
+    void shouldSetEnrollmentLastUpdatedUserInfoWhenAddEventOnProgramWithRegistration()
+        throws IOException
     {
-        String eventUid = CodeGenerator.generateUid();
+        Enrollment enrollment = createEnrollment( programWithRegistration.getUid(),
+            trackedEntityInstanceMaleA.getTrackedEntityInstance() );
 
-        Enrollment enrollment = createEnrollment( programA.getUid(),
+        assertEquals( ImportStatus.SUCCESS,
```
muilpp, line 290: I understand the purpose of this assertion is to stop the execution of the test in case the enrollment is not imported correctly, but the goal of this test is to verify whether the field lastUpdatedByUser is updated or not, not that the enrollment or the event below are imported correctly. Should we take for granted that the import will work and there's already a test somewhere doing this validation? Or maybe create a new class and have this imports done during set up?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
```diff
@@ -607,6 +607,7 @@ private ImportSummary addEnrollment( Enrollment enrollment, ImportOptions import
         }
 
         programInstance.setCreatedByUserInfo( UserInfoSnapshot.from( importOptions.getUser() ) );
+        programInstance.setLastUpdatedByUserInfo( UserInfoSnapshot.from( importOptions.getUser() ) );
```
teleivo, line 610: What is the meaning of the `lastUpdated` columns? Is this method creating an enrollment?

I would not expect `lastUpdated` to be populated when something was never updated.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -2074,6 +2074,21 @@ public void updateTrackedEntityInstances( List<String> teiUids, User user )
             PreparedStatement::execute );
     }
 
+    @Override
+    public void updateEnrollmentsLastUpdatedUserInfo( List<String> enrollmentUids, UserInfoSnapshot userInfoSnapshot )
```
teleivo, line 2078: Does this only update `enrollments`? If so I would expect this to be in an  `EnrollmentStore` rather then `EventStore` 🤔 Unless that leads us to a cyclic dependency between enrollments <-> events. I am bit worried about the lack of structure when it comes to our module and package dependencies. That's why I am bringing this up.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
```diff
@@ -2074,6 +2074,21 @@ public void updateTrackedEntityInstances( List<String> teiUids, User user )
...
+    {
+        List<List<String>> uidsPartitions = Lists.partition( Lists.newArrayList( enrollmentUids ), 20000 );
```
teleivo, line 2080: How did we come up with this batch size? Are we using the same one elsewhere? A constant would help document that 😄

teleivo (commented):
This only affects old tracker right?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
stian-sandvold: At creation, we set lastUpdated = createdAt, although there have been no change from the original (as it doesn't exist any more) the lastUpdated also represent when it was last worked on, which is initially the same as createdAt. This looks good to me :)

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventStore.java
```diff
@@ -70,6 +71,8 @@ List<Event> getEvents( EventSearchParams params, List<OrganisationUnit> organisa
 
     int getEventCount( EventSearchParams params, List<OrganisationUnit> organisationUnits );
 
+    void updateEnrollmentsLastUpdatedUserInfo( List<String> enrollmentUids, UserInfoSnapshot userInfoSnapshot );
```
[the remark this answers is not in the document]
stian-sandvold: Good point, I think it would be great if you could add javadoc to this new method @lucaCambi77 

As to when to use javadoc, the short answer is always 🤔 ? Especially I think it's useful in interfaces. Even if things are straight forward, javadocs should be added. We should enforce this more :)

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java
stian-sandvold: @mortenoh

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/enrollment/AbstractEnrollmentService.java
stian-sandvold: Im not sure I see where we add the lastUpdatedBy object when just updating an enrollment? I see events -> enrollments, and create enrollment. Could you tell us where you covered the case of updating (Lets say attribute values of an enrollment) @lucaCambi77 ?

lucaCambi77: This is change is requested by Calle Hedberg. Every time enrollments get updated, the `lastUpdatedByUserInfo` should be set. From the enrollment page, we can add or update events and this will be a call to /events endpoint.
So I am wondering now if updating  the `lastUpdatedByUserInfo` for the `programinstance` table makes sense then if we update the events

stian-sandvold: So, to clarify, here are the cases we should update the lastUpdatedByUserInfo:

1. Create new enrollment
2. Update enrollment status, attributes, etc
3. Create event in enrollment
4. update event in enrollment

We only add createByUserInfo when creating the enrollment, which is already done.

Of those 4 cases over, I think this PR only covers 1, 3 and 4. So we are missing 1 case.

lucaCambi77:
https://github.com/dhis2/dhis2-core/pull/13236 has been created instead

PR #13214, opened 2023-03-03
fix: move login localeSelect handler

tomzemp:
Defines localeSelect event handler in login.js to avoid issues with content security policy directive (and hence reenable translation on login page)

**Before**

image

**After**
image


netroms (approved):

amcgee (approved):
👍

PR #13216, opened 2023-03-03
fix: Force collate C for column ounamehierarchy in analytics [DHIS2-14871]

maikelarabori:
In the Events/Enrollments API, the sorting order is not correctly respected when we sort by the column `ounamehierarchy`.

This happens in some instances, depending on the DB column (type, value, collate, location, etc.) and seems like a "common" issue.

In our case, the column `ounamehierarchy` recently showed this problem. This PR will ensure that this particular column is always exported as collate "C" in order to fix it. This was the decision adopted after a quick discussion based on pros and cons.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsSqlUtils.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsSqlUtils.java
@@ -241,21 +241,21 @@ public class AnalyticsSqlUtils
     }
 
     /**
-     * This method will simply prefix the given column with the collate
-     * function. ie: columnA -> collate "columnA"
+     * This method will simply prefix the given "collate" with the collate
+     * function. ie: Posix -> collate "Posix"
      *
      * The final statement is surrounded by blank spaces to make its usage safet
      * to the caller.
      *
-     * @param column the column to be "collated".
-     * @return the collate statement, or blank if the given column is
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsSqlUtils.java
```diff
@@ -237,4 +239,25 @@ public static String getCoalesce( List<String> fields, String defaultColumnName
...
+     * function. ie: columnA -> collate "columnA"
+     *
+     * The final statement is surrounded by blank spaces to make its usage safet
+     * to the caller.
+     *
+     * @param column the column to be "collated".
+     * @return the collate statement, or blank if the given column is
+     *         null/blank.
+     */
+    public static String getCollate( String column )
```
gnespolino, line 254: I would pass the enum itself here. Also column isn't the proper name for this parameter

gnespolino (approved):
Looks great, approved with a minor comment

d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsSqlUtils.java
maikelarabori: In this case, I preferred to keep it as String because this is a `util` class. Avoiding dependencies on custom types makes it more reusable I believe...

PR #13225, opened 2023-03-06
fix: broken oauth [DHIS2-7318]

netroms:
[DHIS2-7318](https://dhis2.atlassian.net/browse/DHIS2-7318)

[DHIS2-7318]: https://dhis2.atlassian.net/browse/DHIS2-7318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/spring2fa/TwoFactorAuthenticationProvider.java
```diff
@@ -103,22 +108,21 @@ public Authentication authenticate( Authentication auth )
...
-        validateTwoFactor( user, details );
+        validateTwoFactor( user, auth.getDetails() );
 
         return new UsernamePasswordAuthenticationToken( userService.createUserDetails( user ),
             result.getCredentials(),
             result.getAuthorities() );
     }
 
-    private void validateTwoFactor( User user, ForwardedIpAwareWebAuthenticationDetails details )
+    private void validateTwoFactor( User user, Object details )
```
jbee, line 118: Is there no common base class or interface you could use instead of `Object` ?

jbee (approved):

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/spring2fa/TwoFactorAuthenticationProvider.java
netroms: Unfortunately not, sometimes as with OAuth it is a map, other cases it is WebAuthenticationDetails

PR #13244, opened 2023-03-07
fix: map Dari to Persian Afghanistan (DHIS2-14873)

tomzemp:
We use the code [prs](https://iso639-3.sil.org/code/prs) throughout our apps for providing translations for Dari, the variant of Persian spoken in Afghanistan. 

However, we rely on java.util.locale for generating the names of the languages returned from api/locales/ui, and `prs` is not supported. This PR maps `prs` code to language: `fa`, country: `AF` . This then provides a name for the language `Persian (Afghanistan)`. 

Note: we have discussed this internally and while we would prefer to use the term `Dari` as it is the name preferred by the Afghan government, we think that the possible alternative of `Persian (Afghanistan)` is preferable to the existing approach of just showing the unmapped code `prs`.

There are more notes on the [ticket](https://dhis2.atlassian.net/jira/software/c/projects/DHIS2/issues/DHIS2-14873), but this PR represents a more direct way of addressing the bug.

**Before**
image

**After**
image
(my local build is in Norwegian, hence the Norwegian language names)

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/WebLocale.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/WebLocale.java
@@ -28,7 +28,6 @@
 package org.hisp.dhis.webapi.webdomain;
 
 import java.util.AbstractMap;
-import java.util.AbstractMap.SimpleEntry;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/WebLocale.java
```diff
@@ -43,12 +45,16 @@
 
     private String name;
 
+    private static Map<String, Locale> localeMap = Map.ofEntries(
```
larshelge, line 48: I think this can be written simply `Map.of( .. )`.

PR #13245, opened 2023-03-07
ci: Add Flyway repair flag to dev pipeline [skip ci]

radnov:
Enable the Flyway repair flag when resetting the play/dev WAR, in order to avoid having to manually reset the instance when there's a migration issue.


tonsV2 (approved):
Does it make sense to also add the out of order parameter?

radnov:
> Does it make sense to also add the out of order parameter?

The `flyway.migrate_out_of_order` option already defaults to `true` in [the dhis.conf template we use in AWX](https://github.com/Philip-Larsen-Donnelly/paldwx/blob/master/roles/dhis2.dhis2/templates/dhis.conf.j2#L77).

teleivo (approved):

PR #13253, opened 2023-03-08
fix: event pageSize is not respected with more than 1 category DHIS2-13135

teleivo:
## Main Bug Description

`/tracker/events?pageSize` is not respected for events with multiple category options (COs). If an event belongs to a program with a category combo (CC) with multiple categories the category option combo (COC) will be composed of multiple category options. One category option per category 😅 

The `pageSize` query parameter translates to an SQL limit. So for this endpoint the limit should mean number of events returned.

If the program has a category combination of 2 categories the COC will be composed of 2 COs.

Right now `pageSize=2` leads to a result set with one row per event * 2 (the number of COs)

```
   psi_uid   │  co_uid
═════════════╪═════════════
 AGu4QcxKJQ1 │ yMj2MnmNI8L
 AGu4QcxKJQ1 │ B3nxOazOO2G
```

So instead of getting 2 events we get 1. If you were to look at the query and extend the limit to see the next rows it would look like

```
   psi_uid   │  co_uid
═════════════╪═════════════
 AGu4QcxKJQ1 │ yMj2MnmNI8L
 AGu4QcxKJQ1 │ B3nxOazOO2G
 AmofQP7pKPL │ yMj2MnmNI8L
 AmofQP7pKPL │ C6nZpLKjEJr
```

The root cause is that we join on `dataelementcategoryoption` without any aggregation. This leads to the number of rows = number of events * number of categories. Right now the aggregation is done in Java code https://github.com/dhis2/dhis2-core/blob/20f4fb95e269fe6276ccc4d1ca88988f20cf5d64/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java#L486-L497
At this point the limit has already had its effect and we will not get all necessary events.

### Side bug

Discovered while adding tests for the issue reported via Jira.

* order by `occurredAt` uses enrollments `occurredAt` ([pi.incidentdate](https://github.com/dhis2/dhis2-core/blob/f1df7838dce14abf75d126018706826637602946/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java#L216)) instead of events `occurredAt` (psi.executiondate). dxf2 Event.eventDate is set from [psi.executiondate](https://github.com/dhis2/dhis2-core/blob/f1df7838dce14abf75d126018706826637602946/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/JdbcEventStore.java#L450) and mapped to tracker [occurredAt](https://github.com/dhis2/dhis2-core/blob/f1df7838dce14abf75d126018706826637602946/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/EventMapper.java#L45) from dxf2 Event.eventDate

## Constraints

* Pagination can only be correct if the limit is applied to the final result set. Meaning all possible joins, filters and ordering needs to be applied.
* Sharing needs to be respected. A user has access to an event if they are a super user or have access to every CO in the events COC. This means we need to aggregate as COCs in CCs with more than one category would otherwise lead to number of rows = number of events * number of categories.
* The query is very dynamic and has lots of permutations. JdbcEventStore > 2300 LoC.

## Fix

* Aggregate COs using `string_agg` in SQL instead of the result set mapper in the JdbcEventStore.
* Merge it with the aggregation of the option size count which is just the count of COs.
* Merge it with the ACL check. A user has access to an event only if the user has access to all COs of the events COC. Use the ACL check we already had but instead of returning an access flag for each events CO, aggregate over all COs using `bool_and`.

### Sample query

To make it easier to see what sub-query `coc_agg` is used in the end to make the ACL check and aggregate COs into a single string like `xwZ2u3WyQR0;M58XdOfhiJ7` and get the CO count

```sql
inner join (select coc.categoryoptioncomboid  as id,
                                  string_agg(co.uid, ';')    as co_uids,
                                  count(co.categoryoptionid) as co_count
                           from categoryoptioncombo coc
                                    inner join categoryoptioncombos_categoryoptions cocco
                                               on coc.categoryoptioncomboid = cocco.categoryoptioncomboid
                                    inner join dataelementcategoryoption co
                                               on cocco.categoryoptionid = co.categoryoptionid
                           group by coc.categoryoptioncomboid
                           having bool_and(case
                                               when (co.sharing ->> 'owner' is null or co.sharing ->> 'owner' = 'o1HMTIzBGo7') or
                                                    co.sharing ->> 'public' like '__r_____' or
                                                    co.sharing ->> 'public' is null or
                                                    (jsonb_has_user_id(co.sharing, 'o1HMTIzBGo7') = true and
                                                     jsonb_check_user_access(co.sharing, 'o1HMTIzBGo7', '__r_____') =
                                                     true) then true
                                               else false end) = True) as coc_agg
                          on coc_agg.id = psi.attributeoptioncomboid
```

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
@@ -36,11 +36,9 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
 import java.util.Objects;
 import java.util.function.Supplier;
 
-import org.hisp.dhis.dxf2.metadata.objectbundle.feedback.ObjectBundleValidationReport;
 import org.hisp.dhis.tracker.report.ImportReport;
 import org.hisp.dhis.tracker.report.Status;
 import org.hisp.dhis.tracker.report.ValidationReport;
@@ -266,16 +264,6 @@ public class Assertions
     {
         assertNotNull( report );
         assertFalse( report.hasErrors(), errorMessage( "Expected no validation errors, instead got:\n", report ) );
-    }
-
-    public static void assertNoErrors( ObjectBundleValidationReport report )
-    {
-        assertNotNull( report );
-        List<String> errors = new ArrayList<>();
-        report.forEachErrorReport( err -> {
-            errors.add( err.toString() );
-        } );
-        assertFalse( report.hasErrorReports(), String.format( "Expected no errors, instead got: %s\n", errors ) );
     }
 
     private static Supplier<String> errorMessage( String errorTitle, ValidationReport report )
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
@@ -270,7 +270,6 @@ public class Assertions
 
     public static void assertNoErrors( ObjectBundleValidationReport report )
     {
-        // TODO(ivo): refactor, is there a way to reuse the existing errorMessage()?
         assertNotNull( report );
         List<String> errors = new ArrayList<>();
         report.forEachErrorReport( err -> {
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
@@ -36,9 +36,11 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.function.Supplier;
 
+import org.hisp.dhis.dxf2.metadata.objectbundle.feedback.ObjectBundleValidationReport;
 import org.hisp.dhis.tracker.report.ImportReport;
 import org.hisp.dhis.tracker.report.Status;
 import org.hisp.dhis.tracker.report.ValidationReport;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -260,12 +255,29 @@ public static void assertNoErrorsAndNoWarnings( ImportReport report )
...
+            errorMessage( "Expected import with status OK, instead got:\n", report.getValidationReport() ) );
+    }
+
     public static void assertNoErrors( ValidationReport report )
     {
         assertNotNull( report );
         assertFalse( report.hasErrors(), errorMessage( "Expected no validation errors, instead got:\n", report ) );
     }
 
+    public static void assertNoErrors( ObjectBundleValidationReport report )
```
enricocolasante, line 271: Just thinking if this is the right place for this.
This is an assertion for metadata validation object, probably shouldn't be in tracker assertion.

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
teleivo: @jbee any opinion on a better place for this assertion?

Maybe in `org.hisp.dhis.dxf2.metadata.objectbundle.feedback.Assertions` in the dhis-test-integration module 🤔

jbee: Given the dependencies I guess it is hard to move it module wise. Maybe we can use a more general package? `org.hisp.dhis.feedback` ?

jbee (approved):
Looks like a good solution to me ;)

PR #13255, opened 2023-03-08
fix: fix misleading errors in relationships validator [DHIS2-14711]

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/LinkValidator.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        reporter.addErrorIf( () -> hasMoreThanOneReference( relationship.getFrom() ),
+            relationship, E4001, "from", relationship.getRelationship() );
+        reporter.addErrorIf( () -> hasMoreThanOneReference( relationship.getTo() ),
+            relationship, E4001, "to", relationship.getRelationship() );
+    }
+
+    private void validateRelationshipDoesNotLinkEntityToItself( Reporter reporter, Relationship relationship )
+    {
+        if ( relationshipItemValueType( relationship.getFrom() ) != null
```
enricocolasante, line 66: You are checking if the `from` item is not null here but in the previous validation you assumed that it is always not null (if it is null you can have a NPE in method `hasMoreThanOneReference`)

Maybe this is a discussion that we need to have. I am in favor to make assumption in these validators as long as you make sure that what you are making the assumption for was already validated and that this validator will not run if the assumption is false.
In this case, you can assume that `from` is not null if you make sure that this validator runs after `ConstraintValueTypeValidator`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ValidationUtils.java
```diff
@@ -45,19 +45,19 @@ private ValidationUtils()
         throw new IllegalStateException( "Utility class" );
     }
 
-    public static TrackerType relationshipItemValueType( RelationshipItem item )
+    public static String relationshipItemValueType( RelationshipItem item )
```
enricocolasante, line 48: Not sure why this is needed.
I would rather use the Enum itself to do any future comparison than get the name here and then check if 2 strings are equal. You even needed to use `equalsIgnoreCase` in `DataRelationsValidator`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/DataRelationsValidator.java
```diff
@@ -63,25 +62,26 @@ private void validateRelationshipReference( Reporter reporter, TrackerBundle bun
         RelationshipItem item )
     {
         Optional<String> uid = getUidFromRelationshipItem( item );
-        TrackerType trackerType = relationshipItemValueType( item );
+        String trackerType = relationshipItemValueType( item );
```
enricocolasante, line 65: If we agree that we can make assumptions, can we have the uid as a simple String? If this validator only run after another validator already checked that the uid is there, we can assume that

enricocolasante (changes requested):
There is at least one repeated validation and some order of validators could be changed so we can make some assumptions and simplify the code.

I would also consider having a `RelationshipValidatorTest` with a few test cases, to check how all the validators work together.

teleivo (commented):
I think we should chat about RelationshipValidator as I wonder whether we can capture what its supposed to do in a simpler way 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintValueTypeValidator.java
```diff
@@ -0,0 +1,73 @@
...
+                    relationshipType.getFromConstraint() );
+                validateValueTypeExists( reporter, relationship, "to", relationship.getTo(),
+                    relationshipType.getToConstraint() );
+            } );
+    }
+
+    private void validateValueTypeExists( Reporter reporter, Relationship relationship, String relSide,
+        RelationshipItem item, RelationshipConstraint constraint )
+    {
+        if ( relationshipItemValueType( item ) == null )
```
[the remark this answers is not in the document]
muilpp: The way I understood we want the validators to be, is we want them to run all the validations they contain in the simplest way possible. 
Having that in mind, the thing about adding this logic to the `MandatoryFieldsValidator`, is that in case `from` or `to` are not set, we would have to skip this validation, otherwise we would get a NPE. That's why, having to skip the validation, pushes me into thinking that this needs to be in a separate validator.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/DataRelationsValidator.java
muilpp: I think it's a good idea to assume the uid will always be present here. I'll move the LinkValidator so it's run before the DataRelationsValidator, this way we also assure there's only one uid per relationship item.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/MandatoryFieldsValidator.java
```diff
@@ -48,6 +54,15 @@ public void validate( Reporter reporter, TrackerBundle bundle,
         reporter.addErrorIfNull( relationship.getTo(), relationship, E1124, "to" );
         reporter.addErrorIf( () -> relationship.getRelationshipType().isBlank(), relationship, E1124,
             "relationshipType" );
+        reporter.addErrorIf(
```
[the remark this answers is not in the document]
muilpp: yes, I know we have the `MetaValidator` as kind of a convention, but does it not make more sense to keep this validation in here and delete the `MetaValidator?`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/ValidationCode.java
```diff
@@ -149,7 +149,7 @@
     E4001( "Relationship Item `{0}` for Relationship `{1}` is invalid: an Item can link only one Tracker entity." ),
     E4003( "There are duplicated relationships." ),
     E4004( "Missing required relationship property: 'relationshipType'." ),
-    E4005( "RelationShip: `{0}`, do not exist." ),
+    E4005( "RelationShip: `{0}`, does not exist." ),
```
enricocolasante, line 152: This code can actually be removed as we are not using it

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/Seq.java
```diff
@@ -108,6 +108,12 @@ public static <T> Seq<T> seq( Validator<T> v1, Validator<T> v2, Validator<T> v3,
         return new Seq<>( List.of( v1, v2, v3, v4, v5, v6, v7, v8 ) );
     }
 
+    public static <T> Seq<T> seq( Validator<T> v1, Validator<T> v2, Validator<T> v3, Validator<T> v4,
```
enricocolasante, line 111: Do you really need this?
Maybe it was needed at some point in the refactoring

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidatorTest.java
```diff
@@ -0,0 +1,329 @@
...
+        RelationshipConstraint relationshipConstraintTo = new RelationshipConstraint();
+        relationshipConstraintTo.setRelationshipEntity( to );
+
+        relType.setFromConstraint( relationshipConstraintFrom );
+        relType.setToConstraint( relationshipConstraintTo );
+
+        return relType;
+    }
+
+    private RelationshipType createRelTypeConstraint( RelationshipEntity from, RelationshipEntity to, String uid )
```
enricocolasante, line 286: Not used

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+        validator.validate( reporter, bundle, bundle );
+
+        assertEquals( 1, reporter.getErrors().size(), "Expected just one error when relationship links to itself, but "
+            + reporter.getErrors().size() + " found" );
+        assertHasError( reporter, relationship, E4000, "Relationship: `" + uid + "` cannot link to itself" );
+    }
+
+    @Test
+    void shouldFailWhenRelationshipLinksProgramsAndRelationshipTypeIsNotSet()
```
enricocolasante, line 152: This test name seems wrong to me. RelationType is set; the validation fails because the items are TEIs instead of enrollments

enricocolasante (approved):
Just a few minor comments

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidator.java
```diff
@@ -55,10 +54,9 @@ public RelationshipValidator()
                                 new ExistenceValidator(),
                                 new MandatoryFieldsValidator(),
                                 new MetaValidator(),
-                                new DataRelationsValidator(),
```
teleivo, line 58: Is the DataRelationsValidator unused now? I don't see the file removed in the diff.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/LinkValidator.java
```diff
@@ -0,0 +1,82 @@
...
+    public void validate( Reporter reporter, TrackerBundle bundle, Relationship relationship )
+    {
+        validateRelationshipItemsContainOnlyOneEntity( reporter, relationship );
+        validateRelationshipDoesNotLinkEntityToItself( reporter, relationship );
+    }
+
+    private void validateRelationshipItemsContainOnlyOneEntity( Reporter reporter,
+        Relationship relationship )
+    {
+        reporter.addErrorIf( () -> hasUnexpectedReferences( relationship.getFrom() ),
```
teleivo, line 56: `hasOnlyOneEntity` is more precise and is the same language you already use in the callers method name 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/DuplicationValidator.java
```diff
@@ -27,62 +27,36 @@
...
-        Relationship relationship,
-        RelationshipItem item )
+    @Override
+    public void validate( Reporter reporter, TrackerBundle bundle, Relationship relationship )
     {
-        Optional<String> uid = getUidFromRelationshipItem( item );
-        TrackerType trackerType = relationshipItemValueType( item );
-
-        if ( TRACKED_ENTITY.equals( trackerType ) )
+        if ( bundle.getPreheat().isDuplicate( relationship ) )
```
teleivo, line 46: nice 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidator.java
```diff
@@ -0,0 +1,178 @@
...
+        RelationshipType relationshipType = bundle.getPreheat()
+            .getRelationshipType( relationship.getRelationshipType() );
+
+        validateRelationshipConstraintEntity( reporter, bundle, relationship, "from", relationship.getFrom(),
+            relationshipType.getFromConstraint() );
+        validateRelationshipConstraintEntity( reporter, bundle, relationship, "to", relationship.getTo(),
+            relationshipType.getToConstraint() );
+    }
+
+    private void validateRelationshipConstraintEntity( Reporter reporter, TrackerBundle bundle,
```
teleivo, line 66:
```suggestion
    private void validateRelationshipConstraint( Reporter reporter, TrackerBundle bundle,
```

The `Entity` in the class and method name is not needed IMHO. We are validating that the constraint is met.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidator.java
```diff
@@ -0,0 +1,178 @@
...
+        Relationship relationship,
+        String relSide,
+        RelationshipItem item,
+        RelationshipConstraint constraint )
+    {
+        switch ( constraint.getRelationshipEntity() )
+        {
+        case TRACKED_ENTITY_INSTANCE:
+            validateTrackedEntityInstanceRelationship( item, reporter, relationship, relSide, bundle, constraint );
```
teleivo, line 75: Please keep the order of args the same with the Validator interface. `reporter, bundle, relationship, ...` Makes the code consistent 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidator.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return trackedEntity != null
+            ? Optional
+                .of( bundle.getPreheat().getIdSchemes().toMetadataIdentifier( trackedEntity.getTrackedEntityType() ) )
+            : Optional.empty();
+    }
+
+    private Optional<MetadataIdentifier> getTrackedEntityTypeFromTrackedEntityRef( TrackerBundle bundle, String uid )
+    {
+        final Optional<TrackedEntity> payloadTei = bundle.getTrackedEntities().stream()
```
teleivo, line 168: I think `TrackerBundle.findTrackedEntityByUid` does the same

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidator.java
```diff
@@ -0,0 +1,178 @@
...
+        else if ( !eventExist( bundle, item.getEvent() ) )
+        {
+            reporter.addError( relationship, E4012, TrackerType.EVENT.getName(), item.getEvent() );
+        }
+    }
+
+    private void validateTrackedEntityInstanceType( RelationshipItem item, Reporter reporter, Relationship relationship,
+        String relSide, TrackerBundle bundle, RelationshipConstraint constraint )
+    {
+        getRelationshipTypeUidFromTrackedEntity( bundle,
```
teleivo, line 137: formatting looks a bit odd 🤔 can the 2 lines not live on the same line?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+    void shouldFailWhenRelationshipUidDoesNotHaveCorrectFormat()
+    {
+        Relationship relationship = new Relationship();
+        relationship.setRelationship( "wrong uid" );
+        when( bundle.getRelationships() ).thenReturn( List.of( relationship ) );
+
+        validator.validate( reporter, bundle, bundle );
+
+        assertEquals( 1, reporter.getErrors().size(),
+            "Expected just one error when uid not correct, but " + reporter.getErrors().size() + " found" );
```
teleivo, line 95: We will get the actual size printed by `assertEquals` what we do not get is what errors we actually got. Printing `reporter.getErrors()` would be helpful 😄

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+        when( bundle.getRelationships() ).thenReturn( List.of( relationship ) );
+        when( preheat.getRelationshipType( relationship.getRelationshipType() ) )
+            .thenReturn( createTrackedEntityRelationship() );
+        when( bundle.getPreheat() ).thenReturn( preheat );
+
+        validator.validate( reporter, bundle, bundle );
+
+        assertEquals( 1, reporter.getErrors().size(), "Expected just one error when relationship links to itself, but "
+            + reporter.getErrors().size() + " found" );
+        assertHasError( reporter, relationship, E4000, "Relationship: `" + uid + "` cannot link to itself" );
```
teleivo, line 148: Are we testing the same thing we already tested in the `LinkValidatorTest`?

Is there any interesting "integration" scenario we should test? 🤔 The `RelationshipValidator` puts the individual validators together. It currently runs all of them in a seq so we cannot test that multiple errors are returned to the user within one relationship like when grouping Validators in `all`. I am not sure we need this test class if the individual Validators are well covered.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/DuplicationValidator.java
```diff
@@ -27,62 +27,36 @@
...
 import org.hisp.dhis.tracker.validation.Reporter;
 import org.hisp.dhis.tracker.validation.Validator;
-import org.hisp.dhis.tracker.validation.validator.ValidationUtils;
 
-/**
- * @author Morten Svanæs <msvanaes@dhis2.org>
- */
-class DataRelationsValidator
-    implements Validator<Relationship>
+public class DuplicationValidator implements Validator<Relationship>
```
teleivo, line 40: @enricocolasante I just noticed we have https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/preprocess/DuplicateRelationshipsPreProcessor.java

which runs before validation. So is there even any work for this Validator i.e. can we remove it?

enricocolasante: The validator and the preprocessor are doing 2 slightly different things.
The preprocessor is removing duplicated relationships from the payload.
The validator is making sure that we are not adding to the DB a relationship that is already there

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidator.java
muilpp: the `DataRelationsValidator` has been merged into `ConstraintEntityValidator`. Maybe it doesn't appear in the diff because I created it in this PR.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/LinkValidator.java
muilpp: yeah, you are right, but then I would have to negate it, that's why I was not convinced about it.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidatorTest.java
muilpp: My idea was to test it because of validating the order in which they are run. While it's true the individual test is the same, I wanted to see that since they are all run in seq, there's only one error at a time.

teleivo: I don't think we should duplicate all of these tests in the Validator's own unit test and here. If I break a Validator I will then have 2 failing tests that test exactly the same behavior. Is there one specific scenario that can only be tested in integration like this? If not I don't think we need these tests here which is ok 😄

muilpp: Other than to validate that they are run in sequence, I don't see any other reason to test it. But I don't think we should integration test it when it's something we can unit test.

teleivo: I did not mean with a DB by "integration". To me this test is about testing the integration of all relationship Validators. If there is no special scenario we can only test this way I would delete the tests as they are duplicates. I'll let @enricocolasante way in. If you both feel like they add more value than cost then I am ok 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/ConstraintEntityValidator.java
teleivo: No worries 😄 `RelationshipConstraint` defines what type from/to are (specifically the `RelationshipEntity`). It also defines that the trackedEntity needs to be of a specific tracked entity type which we validate in here 

`if ( !type.isEqualTo( constraint.getTrackedEntityType() ) )`

so to me this Validator is validating the constraint and hot just the `constraint.RelationshipEntity`.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/RelationshipValidatorTest.java
enricocolasante: I agree that the order is the only thing we are not testing in the validator-specific tests.

So maybe one case that should be tested is to make sure that if we have the same relationship for `from` and `to` but the relationship is invalid, we want the error to be returned to be why the relationship is invalid and not the fact that we don't allow auto relationships.

Or thinking the other way around, what would it happen if we switch the order of LinkValidator and ConstraintEntityValidator?

In any case I would review the tests that we have there, like `shouldFailWhenRelationshipUidDoesNotHaveCorrectFormat` does not add any value to me and I agree with Ivo will just result in a duplicate failure if something goes wrong.

But I wouldn't spend a lot of time on that as it can get really complicated soon, so if you can spot something meaningful that will fail if some order is not respected, go on on that, otherwise, we can remove it.

teleivo (approved):

PR #13263, opened 2023-03-09
fix(apphub): use appVersions api to get single appversion

Birkbjo:
Fixes https://dhis2.atlassian.net/browse/DHIS2-13226

Note this should not be merged until the `/appVersions` api is live on App Hub! 
Related: https://github.com/dhis2/app-hub/pull/661

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/DefaultAppHubService.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/DefaultAppHubService.java
@@ -42,7 +42,6 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Optional;
 
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -54,6 +53,7 @@ import org.hisp.dhis.external.conf.ConfigurationKey;
 import org.hisp.dhis.external.conf.DhisConfigurationProvider;
 import org.springframework.http.HttpMethod;
 import org.springframework.stereotype.Service;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
 /**
@@ -105,6 +105,16 @@ public class DefaultAppHubService implements AppHubService
     }
 
     @Override
+    public AppVersion getWebAppVersion(String versionId)
+    {
+        String appHubApiUrl = dhisConfigurationProvider.getProperty( ConfigurationKey.APPHUB_API_URL );
+        String appVersionEndpoint = "v2/appVersions";
+        String url = String.format( "%s/%s/%s", appHubApiUrl, appVersionEndpoint, versionId );
+
+        return restTemplate.getForObject( url, AppVersion.class );
+    }
+
+    @Override
     public AppStatus installAppFromAppHub( String id )
     {
         if ( id == null )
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/DefaultAppHubService.java
```diff
@@ -104,6 +105,18 @@ public List<WebApp> getAppHub()
         return Arrays.asList( apps );
     }
 
+    @Override
+    public AppVersion getWebAppVersion( String versionId )
+    {
+        validateUuid( versionId );
```
Birkbjo, line 111: Doesn't seem like this is enough to fix the "security vulnerability" reported. Do we need a regex-check, or how can we resolve it?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/AppHubUtils.java
```diff
@@ -86,6 +89,32 @@ public static void validateApiVersion( String apiVersion )
...
     }
 
+    /**
+     * Validates the UUID. Apphub uses UUID-v4 for all identifiers.
+     *
+     * @param uuid the uuid
+     * @throws IllegalQueryException if the UUID is invalid
+     */
+
+    public static void validateUuid( String uuid )
```
stian-sandvold, line 99: Usually we return the error codes from the service layer, for example, you can rewrite this to: "public static boolean validUuid( String uuid )"; Then return false instead of an exception

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/DefaultAppHubService.java
stian-sandvold: Is the exception from the validation actually thrown here?

See my comment about the validation util method; You could rewrite this to a simple if-else, that would completely let you ignore the real logic if the validation fails.

stian-sandvold (changes requested):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/DefaultAppHubService.java
Birkbjo: Well, validateUuid will throw if validation fails. So then it would propagate here and throw as well.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/apphub/AppHubUtils.java
larshelge: If there are different possible error states there is merit in either returning the specific `ErrorCode` or throw an exception, so that the caller can understand exactly what the issue is.

larshelge: The IllegalQueryException will be caught in the web layer, and we have a similar approach for analytics (see `DefaultQueryValidator`).

jbee:
I think most of the changes here are obsolete with https://github.com/dhis2/dhis2-core/pull/15214

jbee:
Also note that I used a Java `UUID` class to prevent user input for a app (version) ID to be anything but a valid UUID.

Birkbjo:
> I think most of the changes here are obsolete with #15214

Agreed, we can probably close this.

PR #13285, opened 2023-03-13
fix: Characters not accepted in Tracker capture Attributes [ DHIS2-10215 ]

lucaCambi77:

This PR should fix the issue of characters not being accepted while using `?filter` in a request for attributes in endpoints such as `tracker/trackedEntities`.
We initially thought of encoding the value of the filter to allow any chars and avoid having the problem of the filter value containing chars we use for the split.
A more straightforward solution can be to simplify the way we use the multi-operand filters in a way we only allow _comparison operators_ such as `gt | ge | lt | le` and with values that support only digits and dates chars such as:

- space (`//s`)
- digits (`//d`)
- time zone (`T`, `+`, `-` ) (not sure this is needed but it is a valid Date format)
- date separator (same as negative time zone) (`-`) 
- milliseconds (`.`)

If we don't fall in the case of a multi-operand, it will be a single `operator:value`. If not even that case, we'll throw a bad format exception for the filter.

Created https://dhis2.atlassian.net/browse/DHIS2-14926 to update the tracker docs accordingly

**Please note this only applies to the new tracker exporter endpoints**
If users want to use multi-operand for something it is not covered here, they can use multiple `filter` in the request.

We might want to align the metadata endpoints ( also using `?filter` ) in another issue and find the best way to do it


lucaCambi77:
@enricocolasante in the case you are mentioning, a user can use multiple `filter` in the request. I think it is better to have a working solution for ranges than something we are unsure of or forces us to do a refactor of the API

enricocolasante:
> @enricocolasante in the case you are mentioning, a user can use multiple `filter` in the request. I think it is better to have a working solution for ranges than something we are unsure of or forces us to do a refactor of the API

I agree. Just pointing out that this is a breaking change so it needs to be well-documented in the release notes

larshelge (approved):

PR #13309, opened 2023-03-14
fix: data value set import stric data element fallback [DHIS2-14693]

jbee:
### Summary
Modifies the strict data elements validation.
If a dataset is specified in the import the data element of values must be linked to the dataset.
If no dataset is specified in the import the data element of values must be linked to any of the target elements.

### Manual Testing
* Navigate to System Settings App
* Click Data Import
* Check “require data elements to be part of data set”
* Navigate to Import/Export App
* Export data:
  * Select Data export
  * Check Sierra Leone
  * Add ART monthly summary dataset
  * Click export data and save the file
* Import data:
  * Select Data Import
  * Upload the file you just downloaded
  * Select Start dry run

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetImportValidator.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetImportValidator.java
@@ -476,8 +476,13 @@ public class DataValueSetImportValidator
     private static void checkDataValueStrictDataElement( DataValueEntry dataValue, ImportContext context,
         DataSetContext dataSetContext, DataValueContext valueContext )
     {
-        if ( context.isStrictDataElements()
-            && !dataSetContext.getDataSetDataElements().contains( valueContext.getDataElement() ) )
+        if ( !context.isStrictDataElements() )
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetImportValidator.java
```diff
@@ -476,12 +467,20 @@ private static void checkDataValuePeriodType( DataValueEntry dataValue, ImportCo
...
     {
-        if ( context.isStrictDataElements()
-            && !dataSetContext.getDataSetDataElements().contains( valueContext.getDataElement() ) )
+        if ( !context.isStrictDataElements() )
+        {
+            return;
+        }
+        List<DataSet> targets = dataSetContext.getDataSet() != null
+            ? List.of( dataSetContext.getDataSet() )
+            : context.getValueContextDataSets().get( dataValue.getDataElement(),
```
jbee, line 476: FYI: I go through a cache to make sure this is not loading from DB query more than once per data element. This cache is also used with keys of shape `{de}-{ou}-{aoc}` but using just `{de}` cannot clash and the value type is the same for both.

netroms (approved):

PR #13340, opened 2023-03-16
fix: Users should be able to see EDIs by default in shared dashboards[DHI2S-14906]

d-bernat:
please see: https://dhis2.atlassian.net/browse/DHIS2-14906


maikelarabori (approved):
Looks good @d-bernat :+1:

gnespolino (approved):

PR #13354, opened 2023-03-17
fix: empty file from VisualizationDataController [DHIS2-14240]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14240
### Issue
- With a GET request to `api/visualizations/KQFhHEk5DQ7/data.csv` the server return an empty file. Although data is shown in Data Visualization App.
- The method [getDimensionalObject()](https://github.com/dhis2/dhis2-core/blob/5a6f43ee2c5cba85a5ed9180cbf455f365381a00/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java#L457-L457) at line 457 doesn't handle the dimension of type `DATA_ELEMENT_GROUP_SET`. This leads to Visualization grid rows is empty.
- In result, the backend service failed to map the valueMap to the visualization grid and return an empty file with just table header.

### Fix
- Add additional method to handle `DATA_ELEMENT_GROUP_SET`  and other similar dimension types.

### Test
- Send a GET request to `api/visualizations/KQFhHEk5DQ7/data.csv`, server will return below file 
[DS Cases by Diagnosis last year (4).csv](https://github.com/dhis2/dhis2-core/files/10999224/DS.Cases.by.Diagnosis.last.year.4.csv)

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CombinationGenerator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CombinationGenerator.java
@@ -30,6 +30,8 @@ package org.hisp.dhis.common;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.collections4.CollectionUtils;
+
 /**
  * @author Lars Helge Overland
  */
@@ -149,6 +151,11 @@ public class CombinationGenerator<T>
             int index = indexes[i];
 
             List<T> object = objects.get( i );
+
+            if ( CollectionUtils.isEmpty( object ) )
+            {
+                continue;
+            }
 
             current.add( object.get( index ) );
         }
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CombinationGenerator.java
```diff
@@ -150,6 +152,11 @@ private List<T> getCurrent()
 
             List<T> object = objects.get( i );
 
+            if ( CollectionUtils.isEmpty( object ) )
+            {
+                continue;
+            }
+
```
vietnguyen, line 159: This is to fix 500 error when Visualization's tableRows is empty.

mortenoh (approved):

larshelge (approved):

PR #13363, opened 2023-03-17
fix: use etag cache control instead of last-modified-by for app overrides [DHIS2-14957]

amcgee:
Jira issue: [DHIS2-14957](https://dhis2.atlassian.net/browse/DHIS2-14957)

This issue was causing some hard-to-diagnose issues when uninstalling bundled app overrides - the installed version had a last-modified date LATER than the bundled version (because it was installed later).  When uninstalling the override, the bundled version should have been restored.  However, a browser would still request an asset with a last-modified header LATER than the "latest" now on the server

The solution here is to use etag cache headers instead of last-modified.  The request should then use `If-None-Match` instead of `If-Modified-Since`.

~There is a potential issue with the way we're calculating the etag here, since we're naively using the version of the installed app which is not guaranteed to have identical file contents.  We could instead pre-calculate a hash for each file resource in the app or for each app bundle.~
We generate an etag header using an md5 hash of the lastModified datetime, so it should always be unique when a new version is installed (or uninstalled)

[DHIS2-14957]: https://dhis2.atlassian.net/browse/DHIS2-14957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/AppOverrideFilter.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/AppOverrideFilter.java
@@ -30,7 +30,6 @@ package org.hisp.dhis.webapi.filter;
 import static java.util.regex.Pattern.compile;
 
 import java.io.IOException;
-import java.util.Date;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


KaiVandivier (approved):
Tested out locally and I can confirm that this fixes the issue of uninstalled app overrides 👍

larshelge (approved):

PR #13365, opened 2023-03-19
fix: Three ownership analytics bugs [DHIS2-13373]

jimgrace:
This PR addresses the three problems found while investigating Brian O'Donnell's comment [here](https://dhis2.atlassian.net/browse/DHIS2-13373?focusedCommentId=181405) in [DHIS2-13373](https://dhis2.atlassian.net/browse/DHIS2-13373). The problems, as reported [here](https://dhis2.atlassian.net/browse/DHIS2-13373?focusedCommentId=181428) in that ticket, are:

> 1. When multiple TEI ownership changes happen within the same day, the ownership analytics table is not correctly built for that day. This may result in the owning organisation unit reported for a period ending on or after that day as one of the owning organisation units during that day and not the owning organisation unit as of the end of that day.
> 2. Incorrect owning organisation units are reported when the reporting period is not persisted in the database. Instead of reporting the true owning organisation unit for such a period, the event organisation unit (for event analytics), or the enrollment organization unit (for enrollment analytics) are reported instead.
> 3. When using an enrollment program indicator with non-default analytics period boundaries, reporting is allowed for a period that does not contain the enrollment. When this is the case, the owning organisation unit will be reported at the start or end of the period containing the enrollment rather than ownership at the start or end of period being reported on.

Problem 1 is fixed in `JdbcOwnershipWriter` by stripping the time part from the ownership end date/time so duplicate dates will be correctly merged into one output row, and by moving both the OU and ENDDATE into the merged row. JavaDoc comments are added to make the code intention clearer.

Problems 2 and 3 are both fixed by changing the way the `analytics_ownership_{programUid}` table is joined. Previously, the event or enrollment period was joined to the period in `_periodstructure` to get the period start or end date. This caused both problems because the enrollment period was being joined instead of the reporting period and because `_periodstructure` contains entries only for persisted periods.

Now, `DefaultEventQueryPlanner` requires that any analytics query for ownership will create one query for each period. This allows `OrgUnitTableJoiner` to join the `analytics_ownership_{programUid}` table based only on `params.getEarliestStartDate()` or `params.getLatestEndDate()` which greatly simplifies the table joining logic. `_periodstructure` is no longer used.

Test cases are added to `EventAnalyticsServiceTest` for multiple ownership changes in a day and an enrollment program indicator with non-default analytics period boundaries that tracks the ownership changes after the enrollment period. Also the test periods are not persisted because this is not needed for events, and a comment is added to that effect.

The test described by Brian O'Donnell in [DHIS2-13373](https://dhis2.atlassian.net/browse/DHIS2-13373) now works as expected.

P.S. I also fixed the 7 "New" [not] code smells in `DhisConvenienceTest` that SonarCloud was failing on just because the PR touched something unrelated in that class.

[DHIS2-13373]: https://dhis2.atlassian.net/browse/DHIS2-13373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-13373]: https://dhis2.atlassian.net/browse/DHIS2-13373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventQueryPlanner.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventQueryPlanner.java
@@ -267,8 +267,8 @@ public class DefaultEventQueryPlanner
     {
         List<EventQueryParams> queries = new ArrayList<>();
 
-        if ( (params.isFirstOrLastPeriodAggregationType() || params.useIndividualQuery()) &&
-            !params.getPeriods().isEmpty() )
+        if ( (params.isFirstOrLastPeriodAggregationType() || params.getOrgUnitField().getType().isOwnership() ||
+            params.useIndividualQuery()) && !params.getPeriods().isEmpty() )
         {
             for ( DimensionalItemObject period : params.getPeriods() )
             {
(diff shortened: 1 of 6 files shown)


maikelarabori (approved):

gnespolino (approved):
looks good

PR #13377, opened 2023-03-20
fix: Characters not accepted in Tracker capture Attributes [ DHIS2-10215 ] REVERT

lucaCambi77:

We are not going to add to the release https://github.com/dhis2/dhis2-core/pull/13285, so we need to revert the first partial fix we did for this issue


enricocolasante (approved):

enricocolasante:
You should also create a PR for patch/2.40.0 and get it RCB approved.
@stian-sandvold @Bekkalizer how should we proceed on that?

PR #13391, opened 2023-03-21
fix: use etag cache control instead of last-modified-by for app overrides [DHIS2-14957] (2.38)

KaiVandivier:
Backport of #13363 -- note that the AppOverrideFilter.java file is in a different location here than in 2.39


jbee:
@KaiVandivier I think the method lacks a `throws` declaration that might have been already present in newer versions.

jbee:
Closed in favour of #14627

PR #13392, opened 2023-03-21
fix: use etag cache control instead of last-modified-by for app overrides [DHIS2-14957] (2.37)

KaiVandivier:
Backport of #13363


larshelge:
Unit tests are failing @KaiVandivier , seems like an issue with the backport.

PR #13405, opened 2023-03-22
fix: Add null check in deletion handlers

larshelge:


dayures:
@larshelge could this be backported to version 40 of dhis2?

PR #13407, opened 2023-03-22
fix: createdAtClient for events can be written only on inserts, never on updates [DHIS2-14991]

gnespolino:
Background:
in Analytics we want `lastUpdated` (and `created`) time dimensions to have the following logic:
```
analytics.lastUpdated = psi.lastUpdatedAtClient!=null ? psi.lastUpdatedAtClient : psi.lastUpdated

analytics.created = psi.createdAtClient!=null ? psi.createdAtClient : psi.created
```
this is needed because we want the `lastupdated` and `created` to reflect when the user performed a given action, rather than when it was stored in the database.
While in most cases `lastUpdated` and `lastUpdatedAtClient` are the same, when data is entered through Android application those values can be quite different. A user might update an event and get it synched a considerable amount of time later.
To do so we need consistent values in `createdAtClient` and `lastUpdatedAtClient`.
After reviewing Tracker code, both for old and new importers, I found that `createdAtClient` can be overwritten by the value passed in the payload during update operations, while, by definition, it should always stick to the value passed (if any) during the insert operation.
This PR change this behavior to only allow writing `createdAtClient` once, i.e. during insert operations.


lucaCambi77 (dismissed):
If it is a change of behavior or a new feature, we need a test coverage and add tests where relevant. I think tests when we add or update events should be ok. Also, I am wondering what impact can have this, I mean outside of the Android usage that it seems relevant in this case?

gnespolino:
> If it is a change of behavior or a new feature, we need a test coverage and add tests where relevant. I think tests when we add or update events should be ok. Also, I am wondering what impact can have this, I mean outside of the Android usage that it seems relevant in this case?

I spoke with @enricocolasante before this PR and I still am not sure if this is something we want (well analytics does, not sure about the tracker philosophy regarding this `createdAtClient`).

This PR is to open up a discussion on which tracker team has the ownership, so at the moment, I'm not spending more time writing test that might not needed if this PR will be rejected.

In case it is accepted, I'll write some tests.

enricocolasante (approved):

maikelarabori (approved):

PR #13418, opened 2023-03-23
fix: Return translated orgunit name in event payload [DHIS2-14596]

abyot:
- replace` orgunit.name` (which is untranslated version) with `orgunit.displayname` (which is translated version)
- modify SQL query to include translation values based on user locale
- existence of user locale does not mean we have always translation value, so we coalesce translated and untranslated values, and pick the first non null.


lucaCambi77 (commented):
We need a description for a PR so it is easier to understand the changes through the code. Also, does this have test coverage? I can't see any test

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/AbstractEventService.java
```diff
@@ -623,7 +623,7 @@ public Event getEvent( ProgramStageInstance programStageInstance, boolean isSync
         if ( ou != null )
         {
             event.setOrgUnit( ou.getUid() );
-            event.setOrgUnitName( ou.getName() );
+            event.setOrgUnitName( ou.getDisplayName() );
         }
```
lucaCambi77, line 627: Why here do we add the `displayName` but later we do 
`coalesce(ou_displayname.value, ou.name) as ou_name`

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/trackedentity/store/TwoParamFunction.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.Getter;
+
+/**
+ * @author abyot
+ *
+ */
+@Getter
+@AllArgsConstructor
+public class TwoParamFunction implements QueryElement
+{
```
lucaCambi77, line 40: Do we need a @Getter here? Also,  the name `TwoParamFunction`  is not really clear to me. Either we create a function to support multiple params or we give a more meaningful name as in the case it is only about column and fallback i see

abyot: intention was to make it reusable for a function that takes two params. Can rename it to `TableColumnWithFallback`

PR #13428, opened 2023-03-24
fix: Support for lastUpdatedAtClient/createdAtClient in analytics [DHIS2-14981]

gnespolino:
### Do not merge it. NEEDS APPROVAL FROM RCB.
This PR aims to always provide consistent `lastUpdatedAtClient` and `createdAtClient`.
In particular this affects (in a positive way) the way a user can access data on when a given event was created/updated even when it happened using an Android device.
Before this PR, it could be possible that those information could get lost/overwritten.
Moreover Analytics can now expose a consistent `lastUpdated` and `created` time dimension, representing when a user created/updated a given event, rather than when it was written in the database (which could be different in case of android devices)


enricocolasante (approved):

maikelarabori (approved):

Philip-Larsen-Donnelly:
Approved by RCB - see DHIS2-14981

PR #13429, opened 2023-03-24
fix: Set builder default value for fields in Stats

enricocolasante:
We got a warning because of initializing fields while using the lombok Builder annotation at class level.
To solve this, we needed to add the annotation @Builder.Default to make sure lombok initializes the fields with the explicit value.


teleivo (approved):

muilpp:
very nice PR description! ;)

muilpp (approved):

PR #13436, opened 2023-03-25
ci: automatically link DHIS2 Jira issues [TECH-123]

amcgee:
This uses [dhis2/action-dhis2-jira-automation](https://github.com/dhis2/action-dhis2-jira-automation) (through a [reusable workflow](https://github.com/dhis2/workflows/blob/master/.github/workflows/link-jira-issues.yml)) to enforce GitHub-Jira linking rules, most notably:

- Enforce that a valid Jira issue is specified in the PR title, or `[NO JIRA]` is explicitly declared
- Check for an `APPROVED-<version>` label on all linked Jira issues when the base branch is `patch/<version>`, fail check if not approved
- Automatically create a comment on the PR with links to Jira, including the Jira issue summary as well

There's probably a lot more we could do with this, but this is a good start.  If we like it we can make it a required check for at least patch branches.  An easy and quite cool next step would be to automatically set fix versions and maybe also target versions (if missing) and to optionally auto-transition issues to testing once fix versions match target versions

See examples in [the action repository](https://github.com/dhis2/action-dhis2-jira-automation#the-action-in-action)

> NOTE: currently only works with public-facing Jira projects (notably TECH is not publicly accessible, maybe it should be?  Otherwise we can update the action to support Jira authentication with a PAT)

To test, edit the title of this PR ☝️ ...


Philip-Larsen-Donnelly (approved):
It would be great to get this into core and then roll-out to apps.
@stian-sandvold what do you think? 
Maybe the access to TECH issues is a blocker at the moment? ...but if we are moving away from the TECH project maybe we can coordinate this?

stian-sandvold (approved):

stian-sandvold:
> It would be great to get this into core and then roll-out to apps. @stian-sandvold what do you think? Maybe the access to TECH issues is a blocker at the moment? ...but if we are moving away from the TECH project maybe we can coordinate this?

Lets get it in and see. For now, let's ignore the TECH project, and force merge if neccesary

PR #13444, opened 2023-03-27
chore: log API token authentication events

netroms:


vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/AuthenticationLoggerListener.java
```diff
@@ -129,6 +136,15 @@ else if ( OAuth2AuthenticationToken.class.isAssignableFrom( event.getSource().ge
                 authName = user.getUsername();
             }
         }
+        else if ( ApiTokenAuthenticationToken.class.isAssignableFrom( event.getSource().getClass() ) )
```
jbee, line 139: Is't this basically an `instanceof` check?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAuthenticationToken.java
```diff
@@ -146,4 +146,23 @@ public Map<String, Serializable> getUserSettings()
     {
         return user.getUserSettings();
     }
+
+    @Override
+    public boolean equals( Object o )
+    {
+        if ( this == o )
+            return true;
+        if ( o == null || getClass() != o.getClass() )
```
jbee, line 155: Is there a good reason for this class to even allow subclasses? To me this looks like a class that might even be best done as `@Value`.

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/apikey/ApiTokenAuthenticationToken.java
netroms: I agree. I wanted to keep the scope of this PR as most to the logging issue. This equals came as a Sonar warning. I want to package more API token refactorings into a separate PR.

PR #13447, opened 2023-03-27
feat: ignore nested payloads on old patch endpoints

netroms:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractCrudController.java
```diff
@@ -1259,6 +1260,15 @@ protected Patch diff( HttpServletRequest request )
         {
             throw new BadRequestException( "Unknown payload format." );
         }
-        return patchService.diff( new PatchParams( mapper.readTree( request.getInputStream() ) ) );
+        JsonNode jsonNode = mapper.readTree( request.getInputStream() );
+        for ( JsonNode node : jsonNode )
+        {
+            if ( node.isContainerNode() )
+            {
+                throw new BadRequestException( "Patch payload must be operating on the endpoint object only." );
```
jbee, line 1268: Maybe it is more helpful and clear to the user if we rephrase this to point out that the value can only be a simple value?

jbee (approved):

vietnguyen (approved):

PR #13478, opened 2023-03-28
perf: prune periods with database statement [DHIS2-14996]

jbee:
Prunes unused periods using SQL instead of a code loop for performance reasons.

### Manual Testing
* count periods in database (`select count(*) from period;`)
* run the prune: `POST /api/maintenance/periodPruning`
* count periods again - in demo database this should remove some so count is now less

If no periods have been removed insert a period into the table before running the test (for demo DB this works): 
```sql
insert into period (periodid, periodtypeid, startdate, enddate) 
  values (4444, 11, '2029-01-01', '2029-12-31');
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
@@ -315,23 +315,22 @@ public class JdbcMaintenanceStore implements MaintenanceStore
     public void prunePeriods()
     {
         String sql = "with period_ids(periodid) as ("
-            + "  select distinct periodid from completedatasetregistration"
-            + "  union select distinct periodid from dataapproval"
-            + "  union select distinct periodid from dataapprovalaudit"
-            + "  union select distinct periodid from datainputperiod"
-            + "  union select distinct periodid from datainputperiod"
-            + "  union select distinct periodid from datavalue"
-            + "  union select distinct periodid from datavalueaudit"
-            + "  union select distinct periodid from eventchart_periods"
-            + "  union select distinct periodid from eventreport_periods"
-            + "  union select distinct periodid from eventvisualization_periods"
-            + "  union select distinct periodid from interpretation"
-            + "  union select distinct periodid from lockexception"
-            + "  union select distinct periodid from mapview_periods"
-            + "  union select distinct periodid from validationresult"
-            + "  union select distinct periodid from visualization_periods)"
+            + "    select distinct periodid from completedatasetregistration"
+            + "    union select distinct periodid from dataapproval"
+            + "    union select distinct periodid from dataapprovalaudit"
+            + "    union select distinct periodid from datainputperiod"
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -304,6 +304,30 @@ public int deleteSoftDeletedTrackedEntityInstances()
         return result;
     }
 
+    @Override
+    public void prunePeriods()
+    {
+        String sql = "with period_ids(periodid) as (\n"
+            + "  select distinct periodid from completedatasetregistration\n"
```
larshelge, line 311: Looks good. Just curious, should we remove the line break / `\n`? We don't typically use this is SQL statements, there _might_ some issues with different platforms, and even ORMs don't typically format the SQL they execute.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -304,6 +304,30 @@ public int deleteSoftDeletedTrackedEntityInstances()
         return result;
     }
 
+    @Override
+    public void prunePeriods()
```
larshelge, line 308: Consider adding a note in Javadoc that explains that this SQL query must be updated if new table with foreign key to the period table is added.

netroms (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -304,6 +304,36 @@ public int deleteSoftDeletedTrackedEntityInstances()
...
+            + "    union select distinct periodid from dataapproval"
+            + "    union select distinct periodid from dataapprovalaudit"
+            + "    union select distinct periodid from datainputperiod"
+            + "    union select distinct periodid from datavalue"
+            + "    union select distinct periodid from datavalueaudit"
+            + "    union select distinct periodid from eventchart_periods"
+            + "    union select distinct periodid from eventreport_periods"
+            + "    union select distinct periodid from eventvisualization_periods"
+            + "    union select distinct periodid from interpretation where periodid is not null"
+            + "    union select distinct periodid from lockexception where periodid is not null"
```
larshelge, line 328: Did you check if could move the null check to the main subselect, meaning `select periodid from period_ids where periodid is not null`?

larshelge (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
jbee: That works too.

PR #13414, opened 2023-03-23
chore: make new tracker enrollment web independent of dhis-service-dxf2 TECH-1525

teleivo:
repeating what we did in https://github.com/dhis2/dhis2-core/pull/13283 now for enrollment.

* add new tracker service for enrollments. Note that for enrollments we call out to ProgramInstanceService which we cannot simply replace with the ProgramInstanceStore like we did for relationships. We cannot because ProgramInstanceService contains more logic we will need to copy/transfer it later on.

NOTE: we are still in the 🚧 phase going with broad strokes through web/service/repo layers. We will come back to all these layers and have discussions on how we want to adapt them. First goal is for new tracker to become independent from dxf2. All code that maps from `dxf2` to `dhis-api` will be removed very soon 👷🏻 Please comment if you spot a bug but don't comment on its style 💅🏻 

# Next

* create a new tracker service for events
* then tracked entity
* remove dxf2 dependencies in new tracker relationship service
* make sure we have a new tracker store for every entity and do not need to rely on dxf2 or dhis-api stores

PR #13493, opened 2023-03-29
chore: make new tracker event web independent of dhis-service-dxf2 [TECH-1526]

enricocolasante:
Continuing with events what we did in https://github.com/dhis2/dhis2-core/pull/13283 for relationships and in https://github.com/dhis2/dhis2-core/pull/13414 for enrollments.

* add new tracker service for events
* add new JdbcEventStore. Differently from previous entities we also needed to move the store inside new tracker as it was implemented in dxf2.
* add new tracker csv event service

This PR is mostly moving code around, we rely on 2 tests to assure that the refactoring didn't break anything TrackerEventsExportControllerTest and EventExporterTest. These are the only 2 files to review carefully and check if something is missing.

# What is still missing for events (coming in next PRs)

* The new JdbcEventStore should only be concerned with new tracker names so occurredAt instead of eventDate and so on.
* Remove dependency from dxf2 in TrackerEventCriteriaMapper
* Remove dependency from dxf2 csv event service from TrackerImportController

NOTE: we are still in the 🚧 phase going with broad strokes through web/service/repo layers. We will come back to all these layers and have discussions on how we want to adapt them. First goal is for new tracker to become independent from dxf2. All code that maps from `dxf2` to `dhis-api` will be removed very soon 👷🏻 Please comment if you spot a bug but don't comment on its style 💅🏻 

# Next

* then tracked entity
* remove dxf2 dependencies in new tracker relationship service
* make sure we have a new tracker store for every entity and do not need to rely on dxf2 or dhis-api stores


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/event/EventService.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @author Morten Olav Hansen <mortenoh@gmail.com>
+ */
+public interface EventService
+{
+    // -------------------------------------------------------------------------
+    // READ
+    // -------------------------------------------------------------------------
+
+    @Transactional( readOnly = true )
```
teleivo, line 44: Does it make sense to have these in the interface? 🤔 I think not but if yes then they should be on all methods.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/event/EventUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import org.apache.commons.lang3.StringUtils;
+import org.hisp.dhis.program.UserInfoSnapshot;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@Slf4j
+public class EventUtils
+{
```
muilpp, line 41: I'm not sure if this falls into the category of "style", but if this is a utility class, it should have a private constructor to prevent its instantiation

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/EventMapperTest.java
```diff
@@ -52,7 +52,7 @@
     @Mock
     private Geometry geometry;
 
-    private EventMapper mapper = Mappers.getMapper( EventMapper.class );
+    private Dxf2EventMapper mapper = Mappers.getMapper( Dxf2EventMapper.class );
```
teleivo, line 55: This test should be renamed to Dxf2EventMapperTest as its testing the old mapper. Or adapted to test the new EventMapper, unless it does not bring any value.

teleivo (approved):
Nice 😄 

The new event service has method `public Grid getEventsGrid( EventSearchParams params )` can we remove that? Or are you going to remove it later on. For relationship and enrollment it worked to only keep the service methods that are called from the entities controller. Not sure this works for event as well.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/event/EventUtils.java
enricocolasante: Fixed along with other quick sonar issues

muilpp (approved):

PR #13503, opened 2023-03-31
fix: Enrollments POST API failure with caching issue

lucaCambi77:
Currently with a post to /enrollments, when we check a program to orgunit association, we create a cache for a program only once, so subsequent requests for a new orgunit will fail.
With this fix, we check for a program and program to orgunit in the cache. If not exists we do a database lookup and update the cache for that specific org unit

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.association.jdbc;
 
+import java.sql.Array;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
```diff
@@ -78,53 +79,44 @@ public SetValuedMap<String, String> getOrganisationUnitsAssociationsForCurrentUs
             } );
     }
 
-    public SetValuedMap<String, String> getOrganisationUnitsAssociations( Set<String> uids )
+    /**
+     * Look for a program - org Unit association in a Cache. If the association
```
enricocolasante, line 83: This JavaDoc would be better placed on the interface.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/association/jdbc/JdbcOrgUnitAssociationsStore.java
```diff
@@ -78,53 +79,35 @@ public SetValuedMap<String, String> getOrganisationUnitsAssociationsForCurrentUs
             } );
     }
 
-    public SetValuedMap<String, String> getOrganisationUnitsAssociations( Set<String> uids )
+    public boolean checkOrganisationUnitsAssociations( String program, String orgUnit )
```
muilpp, line 82: since this returns a boolean, wouldn't it be better to name it something like `isProgramAssociatedToOrgUnit?`

muilpp (approved):

PR #13513, opened 2023-04-04
fix: Missing assigned user mapping for events [TECH-1526]

enricocolasante:
Assigned user was retrieved correctly from the DB but never actually set to the event.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/event/JdbcEventStore.java
```diff
@@ -387,6 +387,7 @@ public List<ProgramStageInstance> getEvents( EventSearchParams params,
                     eventUser.setName( resultSet.getString( "user_assigned_name" ) );
                     eventUser.setFirstName( resultSet.getString( "user_assigned_first_name" ) );
                     eventUser.setSurname( resultSet.getString( "user_assigned_surname" ) );
+                    event.setAssignedUser( eventUser );
```
teleivo, line 390: Can you please add a test into TrackerEventsExportControllerTest 😄 The `JsonEvent.getAssignedUser` already exists for asserting 😋

teleivo (approved):

lucaCambi77 (approved):

PR #13514, opened 2023-04-04
chore: make new tracker trackedEntity web independent of dhis-service-dxf2 TECH-1531

teleivo:
repeating what we did in https://github.com/dhis2/dhis2-core/pull/13283 now for trackedEntity.

* add new tracker service for trackedEntity (`DefaultTrackedEntityService`)
* merge [TrackedEntitiesSupportService](https://github.com/dhis2/dhis2-core/blob/786df41e108dcd6a1a1f21df496c20f670628bd6/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackedEntitiesSupportService.java) into `DefaultTrackedEntityService` 🥳 
* return HTTP status 403 from `/tracker/trackedEntities/{id}` if ACL checks in DefaultTrackedEntityService fail (formerly HTTP status 401 returned in DefaultTrackedEntityService). Reason is that the user did provide valid credentials but is just not authorized => https://auth0.com/blog/forbidden-unauthorized-http-status-codes/
* throw NotFoundException (and ForbiddenException) in DefaultTrackedEntityService `getTrackedEntity(String uid, ...` as the tracked entity has to exist. We settled on this in a backend team meeting a while ago. If it has to exist name method `get...` and throw. If it might exist return `Optional`. If a collection is returned make sure its empty instead of null if nothing was found. Using checked exceptions from `feedback` also improves our OpenAPI spec 🥳  

NOTE: we are still in the 🚧 phase going with broad strokes through web/service/repo layers. We will come back to all these layers and have discussions on how we want to adapt them. First goal is for new tracker to become independent from dxf2. All code that maps from `dxf2` to `dhis-api` will be removed very soon 👷🏻 Please comment if you spot a bug but don't comment on its style 💅🏻

# Next PRs

- duplicate the [aggregate code](https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/aggregates) we use to fetch multiple tracked entities. We will duplicate it without the sync related code as sync is not yet implemented in new tracker. We will implement it soon 😄 

The next PR will then remove the mapping from dxf2 currently necessary in `DefaultTrackedEntityService` code using the aggregate.

- remove org.hisp.dhis.dxf2.events.event.csv.CsvEventService from TrackerTrackedEntitiesExportController
- remove new relationship service dependency on dxf2


enricocolasante (approved):
I just think that we may need one more PR later to remove dependency `CsvEventService` from `TrackerTrackedEntitiesExportController`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -0,0 +1,690 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Slf4j
+@Transactional( readOnly = true )
+@Service( "org.hisp.dhis.tracker.trackedentity.TrackedEntityService" )
+@RequiredArgsConstructor
+public class DefaultTrackedEntityService implements TrackedEntityService
+{
+    private final org.hisp.dhis.trackedentity.TrackedEntityInstanceService teiService;
```
muilpp, line 91: could this service, `TrackedEntityInstanceService`, be merged into `TrackedEntityService` that we implement in this class?

one contains the word instance and the other doesn't, but in the end both of them work with the same `TrackedEntityInstance` object.

teleivo: We might at a later point👌 We cannot right now as it's used by others as well. Let's take it step by step 😅

muilpp (approved):

PR #13538, opened 2023-04-05
feat: cleanup job for expired lock exceptions [DHIS2-7763]

jbee:
### Summary
Adds a `created` date to `LockException`s. 

Existing entries are initialised with "now". 

Also adds a job that runs nightly which removes lock exceptions older than 6 months. 
The job is added automatically on startup if it does not exist but it is also configurable by users.
The number of months before lock exceptions expire can be configured between 1-12 months. 

That means existing entries are first cleaned 6 months later. This should allow installations to smoothly transition with this new behaviour.

### Manual Testing
* check the existing lock exception count (e.g. `/api/lockExceptions?fields=id,period[name],organisationUnit[id,name],created,dataSet[id,name]` )
* goto scheduler app
* there should be a job already to cleanup lock exceptions, run it manually
* check existing lock exceptions again, normally none should have been removed as none is old enough
* also check `/api/scheduling/completed/LOCK_EXCEPTION_CLEANUP` , the stage summary tells how many were deleted by the job
* edit `created` date of some of the lock exceptions in the database, for example by changing the date 8 month backwards
```sql
update lockexception set created = now() - interval '8 month' where organisationunitid = 525;
```
* double check some rows in the database were affected
* run job again (either in scheduler app or `POST /api/jobConfigurations/OQ9KeLgqy20/execute`
* check `/api/scheduling/completed/LOCK_EXCEPTION_CLEANUP`  again, now some lock exceptions should have been deleted
* also check `/api/lockExceptions?fields=id,period[name],organisationUnit[id,name],created,dataSet[id,name]` again

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/dataset/hibernate/LockException.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/dataset/hibernate/LockException.hbm.xml
@@ -10,6 +10,8 @@
       <generator class="native" />
     </id>
 
+    <property name="created" type="timestamp" not-null="true" />
+
     <many-to-one name="organisationUnit" class="org.hisp.dhis.organisationunit.OrganisationUnit"
         foreign-key="fk_lockexception_organisationunitid" column="organisationunitid" />
 
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_3__Add_lock_exception_created_column.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_3__Add_lock_exception_created_column.sql
@@ -1,1 +1,1 @@
-alter table lockexception add column if not exists created timestamp without time zone NOT NULL default now();
+alter table lockexception add column if not exists created timestamp without time zone not null default now();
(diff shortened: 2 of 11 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_3__Add_lock_exception_created_column.sql
```diff
@@ -0,0 +1 @@
+alter table lockexception add column if not exists created timestamp without time zone NOT NULL default now();
```
larshelge, line 1: Lower-case "not null" just to make it look pretty.

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/dataset/hibernate/LockException.hbm.xml
```diff
@@ -10,6 +10,8 @@
       <generator class="native" />
     </id>
 
+    <property name="created" type="timestamp" not-null="true" />
```
larshelge, line 13: Sort of silly that we did not add created time field when this entity was introduced!

larshelge (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_3__Add_lock_exception_created_column.sql
jbee: done

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/dataset/hibernate/LockException.hbm.xml
jbee: Yes, I was surprised when I found there was no date directly or indirectly.

vietnguyen (approved):

PR #13540, opened 2023-04-06
chore: make new tracker trackedEntity web independent of dhis-service-dxf2 TECH-1531

teleivo:
follow up to https://github.com/dhis2/dhis2-core/pull/13514

* duplicate dxf2 aggregate code for fetching multiple TEIs now mapping to dhis-api directly

NOTE: we are still in the 🚧 phase going with broad strokes through web/service/repo layers. We will come back to all these layers and have discussions on how we want to adapt them. First goal is for new tracker to become independent from dxf2.

# Next PRs

- remove new relationship service dependency on dxf2. Ban dxf2 dependency from dhis-service-tracker module 🎉 
- remove org.hisp.dhis.dxf2.events.event.csv.CsvEventService from TrackerTrackedEntitiesExportController
- double-check any dxf2 tests that we might have forgotten to duplicate/adapt for new tracker
- https://dhis2.atlassian.net/browse/DHIS2-14752 I think I have a simply idea 😅


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/trackedentity/EnrollmentParams.java
```diff
@@ -0,0 +1,76 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.trackedentity;
+
+import lombok.Value;
+import lombok.With;
+
+@With
+@Value
+public class EnrollmentParams
```
enricocolasante, line 35: `EnrollmentParams` and `EventParams` seem duplication of the same classes in the respective packages.
Is this wanted or is it something we know and we will solve later?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/trackedentity/EnrollmentParams.java
teleivo: Good catch 😅 That was not intentional, even though these classes will look differently in the end 😋

muilpp (approved):
In the future, could we split these huge PRs into smaller ones?

PR #13546, opened 2023-04-07
chore: Clean up async calls in tracker importer [TECH-1532]

enricocolasante:
This PR is needed to prepare the ground for the implementation of the `SYNC` feature in the "new" tracker importer.
`SYNC` is the only feature missing to make "new" tracker equivalent to the old one.

Currently, in tracker importer there are 3 different jobs that can run, one is importer itself when is running in async mode, and the other 2 are side effects that are called at the end of the importer respectively to run notifications and to run notification program rules.
These jobs were overly complicated in my opinion as there were wrapped in a `Thread` class extending `SecurityContextRunnable` and instantiated as a Prototype by Spring.
Now, a job is created and the relevant task is executed.


teleivo (commented):
It is indeed easier to understand what is going on with your changes 😄 I think it would be useful to come up with less generic names as well

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/job/TrackerMessageManager.java
```diff
@@ -85,12 +78,9 @@ public void consume( TextMessage message )
         jobConfiguration.setUid( trackerMessage.getUid() );
         trackerImportParams.setJobConfiguration( jobConfiguration );
 
-        TrackerImportThread trackerImportThread = trackerImportThreadFactory.getObject();
-        trackerImportThread.setTrackerImportParams( trackerImportParams );
-
         SecurityContextHolder.getContext()
             .setAuthentication( AuthenticationSerializer.deserialize( trackerMessage.getAuthentication() ) );
 
-        taskExecutor.executeTask( trackerImportThread );
```
jbee, line 94: I think the thread class main purpose was to set the user during asnyc imports. I guess without it the current user is `null` which is equivalent to the `ALL` superuser. Have you tested this? I don't think this is what we would want. The user should be the user who triggered the import.

jbee (commented):
Looks good in general but I am not sure about removing the threads. I think they are important to run the asnyc import with the user that triggered the import.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/job/TrackerMessageManager.java
enricocolasante: mmm... maybe this is not a problem in this particular case as the importer is considering the user running it the one present in the params and it is validating all ownerships against it, but it could potentially still be a security issue if we can get some metadata that we shouldn't have access to.

Now my doubt is, is it safe that AsyncTaskExecutor is not carrying the SecurityContext? Even more that we assume that a `null` authentication object is equivalent to `ALL` superuser?

enricocolasante:
Closing PR as it needs a better look at the general issue of how a user should be passed to a job

PR #13548, opened 2023-04-09
fix: backward comp for update analytics dimension restrictions [DHIS2-14870]

netroms:
## Summary:
Fixes broken backward compatibility when using the “simulated” (old) UserCredentials object to save analytics dimension restrictions.
Before this fix, updates to the analytics dimension collections on the UserCredentials object would not be reflected in the real object when using PUT.

### Automatic test:
UserControllerTest#testRemoveCogCatDimFromUserCredentialsLegacyFormat()

Jira: [DHIS2-14870](https://dhis2.atlassian.net/browse/DHIS2-14870)

[DHIS2-14870]: https://dhis2.atlassian.net/browse/DHIS2-14870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
@@ -1586,6 +1586,22 @@ public class User
                 newUser.setUserRoles( userRoles );
             }
 
+            Set<Category> oldCatDimCon = oldUser.getCatDimensionConstraints();
+            Set<CategoryOptionGroupSet> oldCogDimCon = oldUser.getCogsDimensionConstraints();
+
+            Set<Category> newCatDimCon = newUserCredentialsRaw.getCatDimensionConstraints();
+            Set<CategoryOptionGroupSet> newCogDimCon = newUserCredentialsRaw.getCogsDimensionConstraints();
+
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -1586,6 +1586,22 @@ public static void populateUserCredentialsDtoCopyOnlyChanges( User oldUser, User
                 newUser.setUserRoles( userRoles );
             }
 
+            Set<Category> oldCatDimCon = oldUser.getCatDimensionConstraints();
+            Set<CategoryOptionGroupSet> oldCogDimCon = oldUser.getCogsDimensionConstraints();
+
+            Set<Category> newCatDimCon = newUserCredentialsRaw.getCatDimensionConstraints();
+            Set<CategoryOptionGroupSet> newCogDimCon = newUserCredentialsRaw.getCogsDimensionConstraints();
+
+            if ( oldCatDimCon != null && newCatDimCon != null && !oldCatDimCon.equals( newCatDimCon ) )
```
jbee, line 1595: You could use `Objects.equals` for null save equals comparison.

larshelge (approved):

PR #13549, opened 2023-04-10
feat: PI first/last aggregation type [DHIS2-14795]

jimgrace:
This PR implements the 4 aggregations types `FIRST` (sum org unit), `LAST` (sum org unit), `FIRST_AVERAGE_ORG_UNIT`, and `LAST_AVERAGE_ORG_UNIT` for program indicators with analytics type Event. See [DHIS2-14795](https://dhis2.atlassian.net/browse/DHIS2-14795). Most of the implementation is in `JdbcEventAnalyticsManager` and, secondarily, in `AbstractJdbcEventAnalyticsManager`.

The heart of the implementation is `JdbcEventAnalyticsManager.getFirstOrLastValueSubquerySql`, which now handles these 4 aggregation types for both PI and non-PI queries. Non-PI subqueries return only a limited number of columns. For PI queries this was a challenge because the subquery finds data from the first/last event while the outer query filters by the PI filter expression. (See the ticket for a discussion of this.) The subquery needs to return any column(s) that might be needed by the PI filter. The easiest way to do this is to have a PI subquery return all columns as well as the PI expression as the `value` column (`select *, [PI expression] as value`). This means that the subquery cannot override the period column with the selected period (for example, `cast('202303' as text) as "monthly"`), as the period column is already returned with all the other columns because of the `select *`. This overriding of the selected period is moved to the main query.

Note that before these changes there was already a problem with the subquery for non-PI use because the subquery would look over the 10 years up to the query period end but the main query would accept only rows where the `yearly` column matched the query period. This means that _last_ subqueries would return anything only if the latest data was in the year queried for. For _first_ subqueries it was even worse. If any data existed in the last 10 years preceding the query period, and before of the query period year, then the query would always return no data. The subquery would find the first data in the last 10 years, but the main query would reject it if it was not in the query period year. The following SQL was generated from the preexisting code by changing a tracker data element aggregation type to `FIRST` and querying that data element in the data visualizer. (It always returned no data because some data existed before 2023):

```
select sum(ax."JuTpJ2Ywq5b") as value, ax."monthly"
from (select "psi",
             "yearly",
             "JuTpJ2Ywq5b",
             cast('202303' as text)                                                           as "monthly",
             "uidlevel1",
             row_number() over (partition by ax."ou",ax."ao" order by ax."executiondate" asc) as pe_rank
      from analytics_event_qdkgabb5jlk as ax
      where ax."executiondate" >= '2013-03-31'
        and ax."executiondate" <= '2023-03-31'
        and ax."JuTpJ2Ywq5b" is not null) as ax
where (ax."monthly" in ('202303'))
  and ax."uidlevel1" in ('ImspTQPwCqd')
  and ax."yearly" in ('2023')
  and ax."pe_rank" = 1
group by ax."monthly"
limit 200001
```

This is now fixed by omitting the test for `yearly` in the main query when there is a first/last subquery.

The subquery now orders by both `executiondate` and `created`. The testers (see the ticket comments) noted that when only ordering by `executiondate`, and there were only two events and they had the same `executiondate` (because up until now it hasn't included the time component), the _first_ and _last_ subqueries would strangely enough both pick the same event. Adding `created` to the `order by` clause (with the same direction `asc` or `desc`) will break the tie in such cases and reliably return the earlier created event for _first_ and the later created event for _last_. This should also benefit non-PI queries when the first and last event happen on the same day (without the time component).

Other details of the implementation are discussed in the ticket. Also of particular note is that when there is a PI in the first/last subquery, the PI analytics boundaries are used to determine the query boundaries, rather than hard coding the boundaries from 10 years before the end of the period to the end of the period. This gives the PI designer complete freedom to define the boundaries.

I added PI aggregation test cases to `EventAnalyticsServiceTest` including these 4 aggregation types as well as `SUM` and `AVERAGE`.

[DHIS2-14795]: https://dhis2.atlassian.net/browse/DHIS2-14795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gnespolino (approved):
Good one

d-bernat (approved):

PR #13570, opened 2023-04-12
chore: add GitHub actions dependency cache

netroms:


.github/workflows/run-tests.yml
```diff
@@ -21,6 +21,13 @@ jobs:
           java-version: 11
           distribution: temurin
           cache: maven
```
teleivo, line 23: Maven dependencies are already cached see https://github.com/actions/setup-java#caching-packages-dependencies 😄

PR #13588, opened 2023-04-12
fix: Validate at least one relationship property set  [DHIS2-14711]

muilpp:
In the [previous PR](https://github.com/dhis2/dhis2-core/pull/13255) I missed one of the error messages that we show to the user when we shouldn't. 
What happened was that when 'from' and 'to' are present in the request but have a null value, the `LinkValidator` detects it as if we were trying to link an entity to itself. This PR fixes it while validating in the `MandatoryFieldsValidator`, not only that from and to are present, but also that one of the three properties (trackedEntity, enrollment or event) have a value.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-14711


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/MandatoryFieldsValidator.java
```diff
@@ -44,9 +45,18 @@
     public void validate( Reporter reporter, TrackerBundle bundle,
         Relationship relationship )
     {
-        reporter.addErrorIfNull( relationship.getFrom(), relationship, E1124, "from" );
-        reporter.addErrorIfNull( relationship.getTo(), relationship, E1124, "to" );
+        reporter.addErrorIf( () -> relationship.getFrom() == null || allItemFieldsAreNull( relationship.getFrom() ),
```
teleivo, line 48: I feel like we over-complicate some of these validations by trying to modularizing too much 🤔 I think there is a way we can validate more in one go. 

If we move the `MetaValidator` before the `MandatoryFieldsValidator` we know the RelationshipType exists. We can then use the constraints like in suggestion https://github.com/dhis2/dhis2-core/commit/19c843ff23e12ebedc9e7a543797fe011b6cec34
This way any validator coming after can rely on the fact that the relationship has its from and to set correctly. The error message is also accurate in letting the user know what needs to be set instead of what they provided.

I think this would work and make the overall relationship validation simpler. Unless I am missing something 😅 Let me know 😋

enricocolasante: `MetaValidator` before the `MandatoryFieldsValidator` is not going to work as if the relastionshipType is null it will return `Cannot find relationshipType` error while it should be `Missing required property`.

For me the issue is that we don't have it modularized enough.
The validation that you add is already happening in `LinkValidator`, the issue is that it is happening together with the auto link validation and they are in fact not independent.
What I would do maybe is to move `validateRelationshipItemsContainOnlyOneEntity` into `MetaValidator`

For me one concept that we need to keep in mind to make the whole validator easiest to understand is that we need to push "static" validation up, so validators that check that the object is correctly formed, and push down validators that check semantic of the object, so they can assume that the object is well formed

muilpp: You are right, I missed the `validateRelationshipItemsContainOnlyOneEntity` in `LinkValidator`.
But, why can't we move it to `MandatoryFieldsValidator` instead?

That way, I would already be moving this static validation up the ladder, but I'm not sure it belongs to the `MetadataValidator`, because in the end I don't check the values exists as I see every other `MetadataValidator` doing, I just check one of them is not null.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/MandatoryFieldsValidator.java
muilpp: @enricocolasante @teleivo , I updated the PR to make it a bit more clear what I meant.
I moved the method to validate the relationship items from the `LinkValidator` to the `MandatoryFieldsValidator`. In there we check the items 'from' and 'to' we receive are correctly set.
I think it's where it belongs, since `MetadataValidator` seems to be dealing with the preheat, and in this case we just want to make sure only one of the three properties of an Item has a value set. Moving it lower doesn't feel right either, because we don't go into the detail of the item's value, we just check it has a value.

Let me know what you think.

teleivo: Do the changes made by @muilpp reflect what you meant in your separation by schema/semantics? 🤔 

With these changes the error message is

`is invalid: an item must link exactly one of trackedEntity, enrollment, event.`

if someone sets for example `trackedEntity` and `enrollment` on an item. What we do not tell the user is which one they should set. Lets assume they then proceed to remove the wrong property. Then they would get that information in different error message produced by a different validator that property `trackedEntity` needs to be set on the item.

It should be clear what a user needs to do to make the import pass. If we cannot accomplish that by separating schema and semantics we should not stick to it but rather favor the user experience 😅

enricocolasante: I agree, we should provide all possible information to the user to fix the issue.
This PR is fixing what it was meant to fix, we were never returning all the information that we can.
To accomplish what you are saying and still fixing the issue of this PR, we should have this validation between the `MetaValidator` and the `LinkValidator` so we can assume that `relationshipType` is ok and we can get the relevant information from it.

In theory, if we want to remove some of the complexity that is caused by sticking to our validation structure in other entities, we could completely remove `MandatoryFieldsValidator` and validate that `relationshipType` is not null in `MetaValidator` and check that `to` and `from` are not null together with the only one validation.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/validation/validator/relationship/MandatoryFieldsValidatorTest.java
```diff
@@ -126,23 +129,109 @@ void verifyRelationshipValidationFailsOnMissingFrom()
 
         validator.validate( reporter, bundle, relationship );
 
-        assertMissingProperty( reporter, relationship, "from" );
+        assertHasError( reporter, relationship, E4001,
+            "Relationship item `from` for relationship `" + relationshipUid
```
teleivo, line 133: I would only assert on `"Relationship item from for relationship " + relationshipUid` so that you don't have to change every test if you want to tweak the message.

I think

`is invalid: an item must link exactly one of trackedEntity, enrollment, event.`

is missing a "to" before "link" in

`is invalid: an item must link to exactly one of trackedEntity, enrollment, event.`

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/validation/validator/relationship/MandatoryFieldsValidator.java
teleivo: > In theory, if we want to remove some of the complexity that is caused by sticking to our validation structure in other entities, we could completely remove MandatoryFieldsValidator

I think we should if its holding us back in providing a better user experience 😄

PR #13592, opened 2023-04-13
fix: Do not set default rounding for EDIs [DHIS2-15055]

maikelarabori:
Small fix to make EDIs behave like a default Indicator regarding rounding.

Basically, the end result value of an EDI should be rounded following the same behaviour as an Indicator that is created with default rounding (which means no rounding is set at creating time).

In order to achieve it, we simply leave the `decimals` property unset when converting the EDI to Indicator. The current flow will take care of the correct rounding approach (applying all the same existing rounding rules).

_This fix needs to go into the current frozen version 2.40.0, so I tried to make as less changes as possible._

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expressiondimensionitem/ExpressionDimensionItem.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/expressiondimensionitem/ExpressionDimensionItem.java
@@ -174,7 +174,7 @@ public class ExpressionDimensionItem
     }
 
     /**
-     * Converts the current object to {@Indicator}.
+     * Converts the current object to {@link Indicator}.
      *
      * @return the {@link Indicator} object.
      */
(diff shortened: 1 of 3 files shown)


d-bernat (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expressiondimensionitem/ExpressionDimensionItem.java
```diff
@@ -170,4 +172,28 @@ public DimensionItemType getDimensionItemType()
     {
         return DimensionItemType.EXPRESSION_DIMENSION_ITEM;
     }
+
+    /**
+     * Converts the current object to {@Indicator}.
```
jimgrace, line 177: Should be {@link Indicator}

jimgrace (approved):
Good idea to move this code into ExpressionDataItem.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/expressiondimensionitem/ExpressionDimensionItem.java
maikelarabori: Ah, good catch. Fixed!

PR #13605, opened 2023-04-13
fix: organisation unit special root based queries as sub-paths [DHIS2-13901]

jbee:
### Summary
Removes special parameters `includeChildren`, `includeDescendants` and `includeAncestors` as well as `level` from endpoint `/api/organisationUnits/{uid}`.

Instead these are used as dedicated paths (endpoints):
* `/api/organisationUnits/{uid}/children` (direct children)
* `/api/organisationUnits/{uid}/children?level=2` (children 2 levels down, that means all children of the direct children)
* `/api/organisationUnits/{uid}/descendants`
* `/api/organisationUnits/{uid}/ancestors`

This is **not** merely a change from a request parameter to a path. When this was solved as request parameter this was based on single object lookup with some special list lookup on top. This meant the result handling of the list items was the handling intended for single objects. Whereas the new path solution is using the same handling as the normal list `/api/organisationUnits` has. This is most notably in the default fields included which is just `id` and `displayName`  for a list whereas it is most of the properties for a single object.

### Cleanup
The special list for single object in the OU controller was the only reason the `getEntity` method was returning a list, not a single entity. With that now gone the method could change to just return the expected single object value.
In case the value does not exist a `NotFoundException` is thrown. This allowed to remove that check and throw code at most call sites. 

This is now also consistent with the existing `/api/organisationUnits/{uid}/parents` list which now is based on the same `getEntityList` implementation with special loading function.

### Compatibility
* With this behaviour propagating to all callers of `getEntity` this meant 2 callers for an update of a `User` would now throw `NotFoundException` instead of `ConflictException` when trying to update a user that does not exist.
* Special list now have same rendering as the usual list `/api/organisationUnits`. This means without the `fields` parameter they contain less properties then their former parameter counterparts would have returned.
* Special lists can now be used with `fields` and `filter` 

### Manual Testing
Use the 4 new API path described above and check they match the expectations. For example, check against the organisation tree shown in an app.

* `/api/organisationUnits/{uid}/children` (only direct children)
* `/api/organisationUnits/{uid}/children?level={n}` (children n level down, where 1 is direct children, 0 would the the unit itself )
* `/api/organisationUnits/{uid}/descendants` (the unit and all its children and grant-children - any depth)
* `/api/organisationUnits/{uid}/ancestors` (the unit and all its parents up to the root)
* `/api/organisationUnits/{uid}/parents` (only the parents of the unit but not the unit itself)

For example

* `/api/organisationUnits/Rp268JB6Ne4/descendants?fields=id,name,level`
* `/api/organisationUnits/Rp268JB6Ne4/parents?fields=id,name,level`
* `/api/organisationUnits/Rp268JB6Ne4/ancestors?fields=id,name,level`
* `/api/organisationUnits/ImspTQPwCqd/children?level=1`

### Automatic Testing
Added new controller test scenarios for the new endpoints.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ApiTokenController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ApiTokenController.java
@@ -30,7 +30,6 @@ package org.hisp.dhis.webapi.controller.security;
 import static org.hisp.dhis.dxf2.webmessage.WebMessageUtils.objectReport;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -58,7 +57,6 @@ import org.hisp.dhis.user.CurrentUser;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.webapi.controller.AbstractCrudController;
 import org.hisp.dhis.webapi.mvc.annotation.ApiVersion;
-import org.hisp.dhis.webapi.webdomain.WebOptions;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PostMapping;
@@ -80,15 +78,6 @@ public class ApiTokenController extends AbstractCrudController<ApiToken>
     private static final List<String> VALID_METHODS = List.of( "GET", "POST", "PATCH", "PUT", "DELETE" );
 
     private final ApiTokenService apiTokenService;
-
-    // Overwritten to get full access control on GET single object
-    @Override
-    protected List<ApiToken> getEntity( String uid, WebOptions options )
-    {
-        ArrayList<ApiToken> list = new ArrayList<>();
-        java.util.Optional.ofNullable( manager.get( ApiToken.class, uid ) ).ifPresent( list::add );
-        return list;
-    }
 
     @Override
     public void partialUpdateObject( String pvUid, Map<String, String> rpParameters,
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
@@ -49,6 +49,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nonnull;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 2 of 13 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ApiTokenController.java
```diff
@@ -81,15 +79,6 @@ public class ApiTokenController extends AbstractCrudController<ApiToken>
 
     private final ApiTokenService apiTokenService;
 
-    // Overwritten to get full access control on GET single object
```
jbee, line 84: @netroms I concluded that this comment (and the override) must be outdated since it is basically a copy of the overridden implementation. Or I fail to see how overriding the method with same implementation has any effect on access control. Please double check.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -653,37 +656,30 @@ public WebMessage putJsonObject( @PathVariable( "uid" ) String pvUid, @CurrentUs
...
         throws ConflictException,
-        ForbiddenException
+        ForbiddenException,
+        NotFoundException
     {
-        List<User> users = getEntity( userUid, NO_WEB_OPTIONS );
-
-        if ( users.isEmpty() )
-        {
-            throw new ConflictException( getEntityName() + " does not exist: " + userUid );
```
jbee, line 681: FYI: This is technically a change in API behaviour. Previously an update attempt on a non existing user would result in conflict whereas it now results in not found. Not found is consistent with any other update.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -653,37 +656,30 @@ public WebMessage putJsonObject( @PathVariable( "uid" ) String pvUid, @CurrentUs
...
-        ForbiddenException
+        ForbiddenException,
+        NotFoundException
     {
         User inputUser = renderService.fromJson( request.getInputStream(), getEntityClass() );
 
-        List<User> users = getEntity( pvUid, NO_WEB_OPTIONS );
-        if ( users.isEmpty() )
-        {
-            throw new ConflictException( getEntityName() + " does not exist: " + pvUid );
```
jbee, line 663: FYI: This is technically a change in API behaviour. Previously an update attempt on a non existing user would result in conflict whereas it now results in not found. Not found is consistent with any other update.

netroms (approved):

netroms (approved):

mortenoh (approved):

PR #13610, opened 2023-04-14
ci: remove explicit host ports for e2e docker compose [skip ci]

radnov:
Using explicit host ports for the E2E tests causes a port allocation error when we [run the tests in parallel on Jenkins](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/dev#L108). With only a container port specified (like `"8080"`), [a random port is chosen on the host](https://docs.docker.com/compose/compose-file/compose-file-v3/#ports).


teleivo (approved):
Sorry forgot that we are also using this on jenkins 😬 

I might add an override or so as locally its just so much nicer not having to check and adapt port every time I run the tests.

PR #13619, opened 2023-04-14
fix: Allow multiple filters for words range in attributes filtering [DHIS2-10215]

lucaCambi77:
Should complete the work for https://dhis2.atlassian.net/browse/DHIS2-10215

Allow a pair of multiple attribute filters in case it is a word range (`SW` and `EW` operators)

### DIFF
--- a/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
+++ b/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
@@ -356,6 +356,22 @@ class TrackerTrackedEntityCriteriaMapperTest
     }
 
     @Test
+    void shouldCreateQueryFiltersWhenAttributesHasMultipleFiltersForWordRange()
+        throws ForbiddenException,
+        BadRequestException
+    {
+        criteria.setFilter( Set.of( TEA_1_UID + ":sw:hello", TEA_1_UID + ":ew:world" ) );
+
+        List<QueryFilter> queryFilters = mapper.map( criteria ).getFilters().stream()
+            .flatMap( f -> f.getFilters().stream() )
+            .collect( Collectors.toList() );
+
+        assertContainsOnly( List.of(
+            new QueryFilter( QueryOperator.SW, "hello" ),
+            new QueryFilter( QueryOperator.EW, "world" ) ), queryFilters );
+    }
+
+    @Test
     void testFilterWhenMultipleTEAFiltersAreRepeated()
     {
         criteria.setFilter( Set.of( TEA_1_UID + ":gt:10", TEA_1_UID + ":lt:20",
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
```diff
@@ -355,6 +355,22 @@ void testFilterWhenTEAFilterIsRepeated()
             anyOf( containsString( "TvjwTPToKHO:GT:10" ), containsString( "TvjwTPToKHO:LT:20" ) ) );
     }
 
+    @Test
```
enricocolasante, line 358: I thought that the whole mechanism of allowing multiple filters was working only because we were not dealing with text.
What if the user send something like `:sw:project:x:ew:world`? Would it work correctly?

lucaCambi77: In that case, later on, while parsing, a filter with the operator `sw` and value  `project:x:ew:world` is created. We can't reliably say that the user is looking for a single or a multi-operand with text because of the way we split. So for simplicity, we create a single filter.
Finally, in the [documentation](https://github.com/dhis2/dhis2-docs/pull/1153) we are explaining how to use multi-operand with text.
If you see a better way to handle this, please go ahead. We discussed encoding the **filter value** but that would need a larger refactor

enricocolasante: I would suggest having a session in the design meeting about it.

The docs about this feature were really confusing as they were stating to use multiple filters to perform some actions while we were not allowing to have multiple filters.
It would be good to understand if that limitation made sense and what implies in this new version of filters.
Before the fix of this issue, we were allowing to have quite a few different text operators working together while now we are blocking all of them, except for the 2 in this PR. We should understand if this is the behavior that we want or not.
Fixing the filters to work well in all the cases is a big refactor as you said, maybe we should then think of not blocking attributes to be specified more than once

lucaCambi77: To summarize:
**Multiple operand:**
- ?filter=abcde:LT:5:GT:4
**Multiple filters:**
- ?filter=abdce:LT:5&filter=abcde:GT:4

As discussed in various meetings with DHSI2-10215 we are restricting multiple operands to only numeric and date cases plus multiple filters for text because in that case, multiple operands case is difficult to apply.

enricocolasante: Ah ok, I am sorry I understood it all wrong.
So I think we are ok with the code, we just need to be careful on the docs then.
We need to make sure that we highlight that only `SW` and `EW` word operators are allowed in a multiple filter. Before all operators were allowed in multiple operands (even if it wasn't make so much sense)

And we need to specify exactly what is the breaking change in the release notes.
The problematic part here is that clients that were using multiple operands with word operators will now experience a different behavior but the request will not break, as you answered before something like this `:sw:project:ew:world` will return just one operand while the user is expecting a range. I know we cannot do better than this but we need to explain this really carefully and clear.

enricocolasante (approved):
Ok on the code.
We will need some more work on the docs and release notes

lucaCambi77:
@enricocolasante i am going to close this. Ideally, if we encode the filter value as discussed, we can add the multi operand also for word range. Multi filter for the same uid still can be discussed but I don't see why we shouldn't remove the check we have in place now and allow it

PR #13624, opened 2023-04-16
fix: Predictions by COC fixes [DHIS2-9832]

jimgrace:
This PR fixes the functionality in [DHIS2-9832 Predictions by category option combination](https://dhis2.atlassian.net/browse/DHIS2-9832) to meet the needs of PEPFAR who requested the feature.

The original implementation of this feature selected input data elements whose catOptionCombos (based on the data element's category combo) were a subset (proper or improper) of the output data element's catOptionCombos. For these data elements it then precomputed a mapping between the input catOptionCombos and the output catOptionCombos. However, there were two problems with this approach:

1. PEPFAR sometimes needs to reduce the dimensions for the output data element, for example having an output data element that represents only positive HIV test results across age and sex dimensions, whereas the input data might have a positive/negative test result dimension in addition to age and sex.

2. Input data might not always have the same catOptionCombos as dictated by the data element's category combo because either (a) the category combo may have changed through time, or (b) the category combo can be overridden per dataset.

For these reasons the prediction disaggregation logic was rewritten. Now an input value is used only if each option in its catOptionCombo is compatible with the options in the output catOptionCombo. The test for compatibility is as follows: If the category to which the option belongs has any options which may be in the output, then the option belonging to such a category must be in the current output catOptionCombo. This is more easily understood with an example: Say that the input options are sex=f and age=50+. If there is only one output category sex, then this value will be considered only when the output catOptionCombo has sex=f. If there are two output categories sex and age, then this value will be considered only when the output catOptionCombo includes both sex=f and age=50+.

This rewrite of the predictor disaggregation algorithm is actually simpler than the original, so `PredictionDisaggregatorUtils` is no longer needed. Therefore, `PredictionDisaggregatorUtilsTest` is also not needed, and `PredictionDisaggregatorAbstractTest` is combined into `PredictionDisaggregatorTest`.

Other notable changes:

1. `MapMap` and `MapMapMap` have performance improvements in `putEntry`, `putEntries`, and `getValue`. In each case, they had been doing one more get or put than was necessary. This extra operation was actually causing a measurable performance difference because of an extra hashCode calculation on the key (even measurable when the key was a `Long`).

2. `Map4` is deleted because it was no longer used anywhere.

3. `PredictionDataConsolidator` performance was improved by changing the orgUnitsRemaining `Queue` type from `OrganisationUnit` to `Long` (storing, and therefore hashing, only the ID instead of the entire object).

4. `PredictionDataValueFetcher` performance was improved by changing `cocLookup` from a `CachingMap` (which looked up each new catOptionCombo one at a time) to a `Map` that holds all the catOptionCombos in the system (from a single database call). In a test PEPFAR instance with over 12,000 catOptionCombos this made a significant performance increase.

5. `CollectionUtils.mapOf` was generalized to handle any number of key/value pairs, since more pairs were needed in the unit tests.


dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
```diff
@@ -57,6 +57,11 @@
 {
     public static final String[] STRING_ARR = new String[0];
 
+    private CollectionUtils()
```
maikelarabori, line 60: Recently, we are given preference to (Lombok annotation)
```
@NoArgsConstructor( access = PRIVATE )
```
instead of declaring the private constructor programatically.

maikelarabori (approved):
Nice improvements :+1:

gnespolino (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/collection/CollectionUtils.java
jimgrace: Thanks, done!

PR #13626, opened 2023-04-16
chore: remove translation of legacy names in new tracker export criteria TECH-1514

teleivo:
We do not need to translate between old and new tracker field names anymore. This was needed when the criteria classes were used in old and new tracker. Old and new tracker now have their own criteria classes representing the request parameters and their own services.

If I understood correctly https://github.com/dhis2/dhis2-core/blob/a3457cc8b5552584caf3e0c4e917e1c83e6898da/dhis-2/dhis-api/src/main/java/org/hisp/dhis/webapi/controller/event/webrequest/PagingAndSortingCriteriaAdapter.java#L83-L88 `isLegacy` can be removed.


gnespolino (approved):
Good one!

enricocolasante (approved):

PR #13633, opened 2023-04-17
fix: Validate capture scope when program protected [DHIS2-15021]

muilpp:
Both the `/tracker/trackedEntities` and `/trackedEntityInstances` endpoints do not validate the provided org unit against the user's capture scope when the program is protected or closed (as stated [in the documentation](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-238/new-tracker.html?h=access+level+develop+2.38#webapi_nti_access_level)). It's always using the search scope instead.
This PR aims to fix it and align these two endpoints with `/tracker/events` and `/events`, by using the service `TrackerAccessManager` to run this particular validation.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-15021


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapper.java
```diff
@@ -216,7 +220,7 @@ private Set<DimensionalItemObject> getDuplicatedAttributes( List<QueryItem> attr
             .collect( Collectors.toSet() );
     }
 
-    private Set<OrganisationUnit> validateOrgUnits( Set<String> orgUnitIds, User user )
+    private Set<OrganisationUnit> validateOrgUnits( Set<String> orgUnitIds, User user, Program program )
```
teleivo, line 223: nit: I would put the user first to show we validate that this user has access to the rest of the parameters (entities).

teleivo (commented):
Makes sense to me 😄 

One thing to note is that `/tracker/trackedEntities/{uid}` does not use criteria and its mapper. Did you check that this `/tracker/trackedEntities/{uid}` also does the same validation?

muilpp:
@teleivo you are right, `/tracker/trackedEntities/{uid}` is a bit different, but it ends up using the `trackerOwnerShipManager` which in the end does the same validation.

teleivo (approved):

enricocolasante (approved):

PR #13648, opened 2023-04-18
fix: Validate against capture scope in /enrollments when program is closed or protected [DHIS2-15145]

muilpp:


muilpp:
This will be merged in [here](https://github.com/dhis2/dhis2-core/pull/13684)

PR #13659, opened 2023-04-19
fix: program rule to assign optionName instead of code [DHIS2-13541]

zubaira:
[DHIS2-13541](https://dhis2.atlassian.net/browse/DHIS2-13541)
To add support for using `optionName` instead of `optionCode` if `useCodeForOptionSet` is false

[DHIS2-13541]: https://dhis2.atlassian.net/browse/DHIS2-13541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR will be merged once we have release version for rule-engine


dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -551,4 +554,32 @@ private DataItem getDisplayName( ProgramRuleVariable prv )
             .valueType( getItemValueType( dataElement.getValueType() ) )
             .build();
     }
+
```
enricocolasante, line 557: This method is quite difficult to follow.

Could you add a guard clause like
```
if ( prv.getUseCodeForOptionSet() )
{
    return List.of();
}
```

- Do not use a builder for Option object
- Do not use Optional.ofNullable to check existence of something, put `prv.getDataElement().getOptionSet()` in the if like `if ( prv.hasDataElement() && prv.getDataElement().getOptionSet() )`
- Same for attributes

enricocolasante (changes requested):
I think we should have an integration test that will use the name instead of the code in some rules and assert that the correct value is showed

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEntityMapperService.java
zubaira: I can try that. but probably sonarcloud would complain

zubaira: resolved

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -551,4 +554,30 @@ private DataItem getDisplayName( ProgramRuleVariable prv )
...
+                .map( op -> new Option( op.getName(), op.getCode() ) )
+                .toList();
+        }
+        else if ( prv.hasTrackedEntityAttribute() && prv.getAttribute().hasOptionSet() )
+        {
+            return prv.getAttribute().getOptionSet().getOptions().stream()
+                .map( op -> new Option( op.getName(), op.getCode() ) )
+                .toList();
+        }
+        else
```
enricocolasante, line 578: Now you can remove this else and just have `return List.of();`

zubaira:
> I think we should have an integration test that will use the name instead of the code in some rules and assert that the correct value is showed

resolved

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -551,4 +554,28 @@ private DataItem getDisplayName( ProgramRuleVariable prv )
...
     }
+
+    private List<Option> getOptions( ProgramRuleVariable prv )
+    {
+        if ( prv.getUseCodeForOptionSet() )
+        {
+            return List.of();
+        }
+
+        // Option list should only be populated if rule variable has dataElement/attribute with associated option set
```
muilpp, line 565: I think the condition below is self-explanatory, do we need the comment above?

dhis-2/pom.xml
```diff
@@ -64,7 +64,7 @@
     <!-- *Dependencies* -->
 
     <!-- DHIS2 Rule Engine -->
-    <dhis2-rule-engine.version>2.1.3</dhis2-rule-engine.version>
+    <dhis2-rule-engine.version>2.1.6-SNAPSHOT</dhis2-rule-engine.version>
```
enricocolasante, line 67: Can you remove the snapshot before merging?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/programrule/engine/ProgramRuleEngineTest.java
```diff
@@ -458,6 +495,34 @@ void testAssignValueTypeAge()
...
+    void testAssignOptionSetName()
+    {
+        setUpRuleActionForOptionSetName();
+
+        Event programStageInstance = eventService.getEvent( "UID-PS13" );
+        List<RuleEffect> ruleEffects = programRuleEngine.evaluate( programStageInstance.getProgramInstance(),
+            programStageInstance, Sets.newHashSet(), List.of( programRuleToTestOptionSet ) );
+
+        assertNotNull( ruleEffects );
+        assertThat( ruleEffects.get( 0 ).ruleAction(), Matchers.instanceOf( RuleActionShowWarning.class ) );
```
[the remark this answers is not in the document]
enricocolasante: mmm...
The import can be banned, but if it is used like this then the ban gets fooled :-1:

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -554,4 +557,27 @@ private DataItem getDisplayName( ProgramRuleVariable prv )
...
     }
+
+    private List<Option> getOptions( ProgramRuleVariable prv )
+    {
+        if ( prv.getUseCodeForOptionSet() )
+        {
+            return List.of();
+        }
+
+        if ( prv.hasDataElement() && prv.getDataElement().hasOptionSet() )
```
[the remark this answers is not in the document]
enricocolasante: Yes, they cannot be both true at the same time

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/programrule/engine/ProgramRuleEngineTest.java
```diff
@@ -458,6 +493,34 @@ void testAssignValueTypeAge()
...
+    {
+        setUpRuleActionForOptionSetCode();
+
+        Event programStageInstance = eventService.getEvent( "UID-PS13" );
+        List<RuleEffect> ruleEffects = programRuleEngine.evaluate( programStageInstance.getProgramInstance(),
+            programStageInstance, Sets.newHashSet(), List.of( programRuleToTestOptionSet ) );
+
+        assertNotNull( ruleEffects );
+        assertInstanceOf( RuleActionShowError.class, ruleEffects.get( 0 ).ruleAction() );
+        assertThat( ruleEffects.get( 0 ).data(), Matchers.is( option1.getCode() ) );
```
teleivo, line 521: `assertEquals` should probably do as described in https://github.com/dhis2/wow-backend/blob/master/guides/testing.md#assertions

teleivo (approved):

PR #13660, opened 2023-04-19
feat: Programs are now validate against TET [DHIS2-15042]

gnespolino:
This PR adds a validation to check if all programs passed to Tei Analytics endpoints are defined for the given tracked entity type


maikelarabori (commented):
Looks good @gnespolino. I just think would be nice to have some unit and e2e tests.

Thanks

d-bernat (approved):

maikelarabori (approved):

PR #13673, opened 2023-04-20
fix: category option groups are consistently named categoryOptionGroups [BETA-52]

jbee:
The issue is caused by XML annotations considering the property to be named `categoryOptionGroups` while for Java and JSON it is just `groups`. During the schema building the XML annotations presence overrides the actual property name `groups` with `categoryOptionGroups`. This then causes a couple of issues one of which is that when the object is converted to JSON the JSON mapper is restricted to only map the explicitly listed properties. If that includes `categoryOptionGroups` it will always be empty since for Java and JSON the property is called `groups`.

To keep backwards compatibility the JSON name is also changed to `categoryOptionGroups`.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
@@ -303,8 +303,10 @@ public class CategoryOption
         this.categoryOptionCombos = categoryOptionCombos;
     }
 
-    @JsonProperty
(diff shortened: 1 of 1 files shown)


Philip-Larsen-Donnelly:
@jbee can you please document this breaking change in https://github.com/dhis2/dhis2-releases/blob/master/releases/2.40/README.md and check if there are any references in the docs that should be updated?

vietnguyen (approved):

larshelge (approved):

PR #13710, opened 2023-04-24
feat: add support for subpath wildcards in route urls [DHIS2-15098]

amcgee:
Fixes [DHIS2-15098](https://dhis2.atlassian.net/browse/DHIS2-15098) - support path wildcard suffix `/**`
Fixes [DHIS2-15191](https://dhis2.atlassian.net/browse/DHIS2-15191) - pass through non-2xx responses
Fixes [DHIS2-15197](https://dhis2.atlassian.net/browse/DHIS2-15197) - disable cookie store
Fixes [DHIS2-15198](https://dhis2.atlassian.net/browse/DHIS2-15198) - filter request and response headers to exclude transport-level metadata

This adds support for sub-paths in Route URLs.  Subpaths are only supported if the route is configured with a URL ending in `/**`, and can be accessed via the route's Run API like so - `/api/routes/ABC123/run/sub/path/xyz`, which if route `ABC123` is configured with the URL `https://google.com/**` would proxy the request to `https://google.com/sub/path/xyz`

Attempting to access a subpath for a route which doesn't have the `/**` path wildcard suffix will result in a 400 Bad Request error (I'm not sure if this should be a 404 instead...)

This has been approved for 2.40.0 if we want to support it there (since this makes the routes API quite a bit more useful), though with this implementation it would also be a non-breaking extension to add `/**` support in v 40.1

---
In addition, prevent the HTTP Client used for upstream requests from remembering cookies
Finally, forward headers to the upstream request (stripping sensitive ones) and pass through the original response from the upstream server.

In the future we will want to add X-Forwarded-* headers.

[DHIS2-15098]: https://dhis2.atlassian.net/browse/DHIS2-15098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-15197]: https://dhis2.atlassian.net/browse/DHIS2-15197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-15191]: https://dhis2.atlassian.net/browse/DHIS2-15191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-15198]: https://dhis2.atlassian.net/browse/DHIS2-15198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-api/src/test/java/org/hisp/dhis/route/RouteTest.java
+++ b/dhis-2/dhis-api/src/test/java/org/hisp/dhis/route/RouteTest.java
@@ -62,11 +62,11 @@ class RouteTest
     @Test
     void testAllowsSubpaths()
     {
-        Assertions.assertEquals( false, tldRoute.allowsSubpaths() );
-        Assertions.assertEquals( false, routeWithSubPath.allowsSubpaths() );
-        Assertions.assertEquals( false, routeWithDirectorySubPath.allowsSubpaths() );
-        Assertions.assertEquals( true, routeWithPathWildcard.allowsSubpaths() );
-        Assertions.assertEquals( true, routeWithSubPathAndPathWildcard.allowsSubpaths() );
+        Assertions.assertFalse( tldRoute.allowsSubpaths() );
+        Assertions.assertFalse( routeWithSubPath.allowsSubpaths() );
+        Assertions.assertFalse( routeWithDirectorySubPath.allowsSubpaths() );
+        Assertions.assertTrue( routeWithPathWildcard.allowsSubpaths() );
+        Assertions.assertTrue( routeWithSubPathAndPathWildcard.allowsSubpaths() );
     }
 
     @Test
--- a/dhis-2/dhis-api/src/test/java/org/hisp/dhis/route/RouteTest.java
+++ b/dhis-2/dhis-api/src/test/java/org/hisp/dhis/route/RouteTest.java
@@ -27,8 +27,11 @@
  */
 package org.hisp.dhis.route;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -142,9 +144,19 @@ public ResponseEntity<String> exec( Route route, User user, HttpServletRequest r
...
 
-        UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromHttpUrl( route.getUrl() )
+        UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromHttpUrl( route.getBaseUrl() )
             .queryParams( queryParameters );
 
+        if ( subPath.isPresent() )
+        {
+            if ( !route.allowsSubpaths() )
+            {
+                throw new IllegalArgumentException(
```
jbee, line 154: We try to use the exceptions that already map to a HTTP response directly, in this case use `org.hisp.dhis.feedback.BadRequestException`. Then propagate the `throws` on the methods to the controller level.

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
amcgee: @jbee thanks for the review! - most services seem to return IllegalArgumentException or similar which makes sense to me, as the service might be used differently in different controllers, can you expand on why we shouldn't keep concerns separated between service and controller here?  Happy to change it if needed, but curious why we're moving in that diretion

amcgee:
FYI I'm adding fixes for [DHIS2-15197](https://dhis2.atlassian.net/browse/DHIS2-15197) and [DHIS2-15191](https://dhis2.atlassian.net/browse/DHIS2-15191) to this PR as they came up while testing this.  [DHIS2-15198](https://dhis2.atlassian.net/browse/DHIS2-15198) will need to be further investigated

amcgee:
@mortenoh @jbee would you mind taking another quick look at this one?  I updated it to address some other issues found along the way.

dhis-2/dhis-api/src/test/java/org/hisp/dhis/route/RouteTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @AfterEach
+    void tearDown()
+    {
+    }
+
+    @Test
+    void testAllowsSubpaths()
+    {
+        Assertions.assertFalse( tldRoute.allowsSubpaths() );
```
mortenoh, line 65: Static imports

mortenoh (approved):

Philip-Larsen-Donnelly (approved):

PR #13712, opened 2023-04-24
fix: correct android settings app authority and namespace protection [DHIS2-15087]

amcgee:
The namespace used by the Android Settings App and the Android Capture App is `ANDROID_SETTINGS_APP` not `ANDROID_SETTING_APP`.  Additionally, the authority for the Android Settings App is `M_androidsettingsapp` not `M_Android_Setting` (see [api/authorities](https://play.dhis2.org/dev/api/authorities) or the user role authority selector below where you'll see duplication because one authority comes from this class and another from the installed app)

Screenshot 2023-04-24 at 16 36 14

The solution here is the simplest one available, which is to simply follow the authority and namespace used by the installed app.  Implementations which had assigned users the old authority (`M_Android_Setting`) will no longer have that authority assigned, but it didn't do anything useful before so I think that's fine.  The name displayed for the authority will also be a bit less nice (`androidsettingsapp app` instead of `Android Setting app`) but that is universally the case with generated app authority names, and we should fix it in the future by using the application's defined title instead of trying to infer it from the short name.  We can do that in a future change.

I haven't tested this yet, would be great if someone can confirm the fix with [the PR docker image](https://hub.docker.com/layers/dhis2/core-pr/13712) **before we merge**.

We want to get this in ASAP so that it can be included in 2.40.0

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/AndroidSettingsApp.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/AndroidSettingsApp.java
@@ -33,20 +33,20 @@ import org.hisp.dhis.datastore.DatastoreService;
 import org.springframework.stereotype.Component;
 
 /**
- * The main purpose (so far) of the {@link AndroidSettingApp} component is to
+ * The main purpose (so far) of the {@link AndroidSettingsApp} component is to
  * establish the protected {@link #NAMESPACE} in the {@link DatastoreService} so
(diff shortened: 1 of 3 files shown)


jbee (approved):

jbee:
Just needs fixing for the test `DatastoreControllerAndroidSettingsAppTest` where the name change also has to be applied.

amcgee:
> Just needs fixing for the test DatastoreControllerAndroidSettingsAppTest where the name change also has to be applied.
@jbee fixed, thanks

netroms (approved):

PR #13724, opened 2023-04-25
feat: Return reserved value for non-existent event repetitions [2.41-DHIS2-15171]

d-bernat:
Please see: https://dhis2.atlassian.net/browse/DHIS2-15171 and https://dhis2.atlassian.net/browse/DHIS2-15199

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueStatus.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueStatus.java
@@ -30,24 +30,34 @@ package org.hisp.dhis.common;
 /**
  * Enum for description of repeatable stage value
  */
-public enum RepeatableStageValueStatus
+public enum ValueStatus
 {
     //repeating of the repeatable stage does not exist
-    UNDEFINED,
+    NOT_DEFINED,
 
     //value is not set
-    UNSET,
+    NOT_SET,
 
     //value is set
     SET;
 
-    public static RepeatableStageValueStatus of( boolean isDefined, boolean isSet )
+    public static ValueStatus of( boolean isDefined, boolean isSet )
     {
         if ( !isDefined )
         {
-            return RepeatableStageValueStatus.UNDEFINED;
+            return ValueStatus.NOT_DEFINED;
         }
 
-        return isSet ? RepeatableStageValueStatus.SET : RepeatableStageValueStatus.UNSET;
+        return isSet ? ValueStatus.SET : ValueStatus.NOT_SET;
+    }
+
+    public static String toString( ValueStatus valueStatus )
+    {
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueStatus.java
```diff
@@ -30,24 +30,34 @@
...
         {
-            return RepeatableStageValueStatus.UNDEFINED;
+            return ValueStatus.NOT_DEFINED;
         }
 
-        return isSet ? RepeatableStageValueStatus.SET : RepeatableStageValueStatus.UNSET;
+        return isSet ? ValueStatus.SET : ValueStatus.NOT_SET;
+    }
+
+    public static String toString( ValueStatus valueStatus )
```
maikelarabori, line 54: Instead of associating the "abbreviations" here, would probably be better to associate them with the enums themselves. Like it's done in this enum, for example:
```
DataDimensionType.java
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -370,9 +370,8 @@ else if ( !params.hasPeriods() && params.hasFilterPeriods() )
 
             columns.add( columnAndAlias.asSql() );
 
-            // does repeatable stage exist?
-            if ( queryItem.hasProgramStage() && queryItem.getProgramStage().getRepeatable()
-                && queryItem.hasRepeatableStageParams() )
+            // asked for row context if allowed and needed
+            if ( rowContextAllowedAndNeeded( params, queryItem ) )
```
maikelarabori, line 374: Would be nice to have a e2e test to prevent DHIS2-15199 to happen in the future.

maikelarabori (approved):
Looks good @d-bernat. Left two comments for review.

Thanks!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueStatus.java
d-bernat: Done

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
d-bernat: Done

gnespolino (approved):

PR #13729, opened 2023-04-26
chore: Use assertInstanceOf instead of Hamcrest method [TECH-1349]

enricocolasante:


dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/common/ValueTypeMappingTest.java
```diff
@@ -50,63 +48,63 @@
...
         Object o = NUMERIC.convertSingle( "1" );
-        assertThat( o, is( instanceOf( BigInteger.class ) ) );
+        assertInstanceOf( BigInteger.class, o );
     }
 
     @SuppressWarnings( { "unchecked", "rawtypes" } )
     @Test
     void testConvertMultipleNumeric()
     {
         Object o = NUMERIC.convertMany( List.of( "1", "2", "3" ) );
```
teleivo, line 58: Can you please change the var type to `List<Object>` which `NUMERIC.convertMany` returns. So we don't need to cast later on. (apply in all tests)

teleivo (approved):
Nice 😄

jbee (approved):
Thank you for the cleanup :)

PR #13731, opened 2023-04-26
chore: rename ProgramStageInstance to Event TECH-1537

teleivo:
Tracker names evolved over the years. Some of this evolution is documented in https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker-deprecated.html\#webapi_tracker_migration

ProgramStageInstance has been renamed in parts of the code to Event years ago. Rename classes, methods, variables, ... related to ProgramStageInstance to Event


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Objects.java
```diff
@@ -87,7 +87,7 @@
     MAP( "map", Map.class ),
     DASHBOARD( "dashboard", Dashboard.class ),
     PROGRAM( "program", Program.class ),
-    PROGRAMSTAGEINSTANCE( "programStageInstance", ProgramStageInstance.class ),
+    EVENT( "event", Event.class ),
```
jbee, line 90: Hm... we have to double check that the name `PROGRAMSTAGEINSTANCE` isn't part of an generic API.

jbee (approved):
Thanks for taking this on. 
I am pretty sure this is fine. Just the `Objects` enum re-name might have consequences. We should check this.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Objects.java
teleivo: Its part of the `/system/objectCounts` and `/dataSummary`response.

We return the count per value
* https://github.com/dhis2/dhis2-core/blob/df3d184edafc38879a3f9a099daea580bcb477b0/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/ObjectCount.java#L50
* https://github.com/dhis2/dhis2-core/blob/df3d184edafc38879a3f9a099daea580bcb477b0/dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datastatistics/DefaultDataStatisticsService.java#L201

I reverted the changes here for the enum name and the `value`. So we can merge the changes and wait for @mortenoh opinion if it is ok to change these public facing mentions of `programStageInstance` as well 👌🏻

enricocolasante (approved):
A couple of things that I would double check and maybe add in next PRs:
- Some `ProgramStageInstance psi` where correctly renamed to `Event event` but there are still a few `Event psi` that don't seem to have conflicts with other variables in the scope and could be renamed to `event`
- `PROGRAM_STAGE_INSTANCE` value in `RelationshipType` enum could be renamed and it shouldn't have anything to do with SQL

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Objects.java
teleivo: Sample from `curl -u admin:district -H 'content-type: application/json' 'https://play.dhis2.org/dev/api/system/objectCounts'`

```json
{
  "indicator": 77,
  "visualization": 292,
  "period": 384,
  "programStageInstance": 423663,
  "organisationUnit": 1332,
  "validationRule": 37,
  "dataValue": 4937367,
  "dataElement": 1036,
  "program": 14,
  "organisationUnitGroup": 18,
  "trackedEntityInstance": 73124,
  "programInstance": 73152,
  "indicatorType": 5,
  "eventVisualization": 50,
  "user": 130,
  "map": 93,
  "dataSet": 26,
  "indicatorGroup": 17,
  "dataElementGroup": 84,
  "userGroup": 34,
  "dashboard": 27
}
```

PR #13749, opened 2023-04-27
fix: Predictor white space parsing [DHIS2-14635]

jimgrace:
This is in support of [DHIS2-14635 Predictions by Data Element Group](https://dhis2.atlassian.net/browse/DHIS2-14635).

This PR changes the predictor expression preprocessing so that any whitespace characters can be used to separate tokens (not just the space character), and so all whitespace characters are faithfully copied to the prediction description.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorExpression.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorExpression.java
@@ -85,9 +85,7 @@ public class PredictorExpression
 
     private static final String IN = "in";
 
-    private static final String PREPROCESSOR_SEPARATOR = " --> ";
-
-    private static final int PREPROCESSOR_SEPARATOR_LENGTH = PREPROCESSOR_SEPARATOR.length();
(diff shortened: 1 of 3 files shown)


maikelarabori (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/expression/PredictorExpression.java
```diff
@@ -166,6 +155,19 @@ public PredictorExpression( String expression )
                 "UID '%s' must start with a letter and contain 10 more letters and numbers in '%s'", degUid,
                 expression ) );
         }
+
+        // -->
```
larshelge, line 159: Perhaps remove this comment? Looks good otherwise!

jimgrace: @larshelge, thanks. This is a comment for one of the six expression parts. The fifth part (array index 4) is the separator '-->'. I've now numbered the comments for each part and said more about this comment so hopefully that makes it clearer.

PR #13754, opened 2023-04-28
chore: rename ProgramInstanceStore to EventStore TECH-1537

teleivo:
its methods. Replace programStageInstance with event.

follow up to https://github.com/dhis2/dhis2-core/pull/13753

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
@@ -87,44 +87,44 @@ public class EventTrackerConverterService
     }
 
     @Override
-    public List<org.hisp.dhis.tracker.imports.domain.Event> to( List<Event> programStageInstances )
+    public List<org.hisp.dhis.tracker.imports.domain.Event> to( List<Event> events )
     {
-        List<org.hisp.dhis.tracker.imports.domain.Event> events = new ArrayList<>();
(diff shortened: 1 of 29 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
```diff
@@ -168,18 +167,17 @@ public List<Event> from( TrackerPreheat preheat, List<org.hisp.dhis.tracker.impo
     @Override
     public Event fromForRuleEngine( TrackerPreheat preheat, org.hisp.dhis.tracker.imports.domain.Event event )
     {
-        Event psi = from( preheat, event, null );
+        Event result = from( preheat, event, null );
```
muilpp, line 170: just curious, why do we call it result instead of event in there?
I've seen it in a couple of other places too.

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
teleivo: There is already one event in scope. We have a lot of code that maps events from one type to another one. Like validation being done on the `org.hisp.dhis.tracker.imports.domain.Event` which represents potentially invalid user input using `org.hisp.dhis.program.Event` which is used to persist (comes from the preheat).

In such situation we need to decide how to name them. In methods/functions that take an `event` and return another type of `event` I chose `result` as it makes it clear that this is the event that will be returned.

PR #13757, opened 2023-04-28
feat: add support for subpath wildcards in route urls [DHIS2-15098] (2.40)

amcgee:
Backport of #13710 to 2.40


Philip-Larsen-Donnelly (approved):
Approved as backport

PR #13763, opened 2023-04-28
fix: TEI query fails when search text contains a comma [ DHIS2-9324 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-9324

Spring uses converters to parse request parameters. When request parameters as a String are converted to a Set, it will use by default a comma (,) as a delimiter and split the string.
This is not used in the code base and shouldn't be as default, so we add a custom converter to undo this default split with this PR.
This PR also put MVC converters in mvc/requestconverter folder so it is clear what are used for

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/WebMvcConfig.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/WebMvcConfig.java
@@ -56,6 +56,8 @@ import org.hisp.dhis.webapi.mvc.messageconverter.MetadataExportParamsMessageConv
 import org.hisp.dhis.webapi.mvc.messageconverter.StreamingJsonRootMessageConverter;
 import org.hisp.dhis.webapi.mvc.messageconverter.XmlMessageConverter;
 import org.hisp.dhis.webapi.mvc.messageconverter.XmlPathMappingJackson2XmlHttpMessageConverter;
+import org.hisp.dhis.webapi.mvc.requestconverter.StringToOrderCriteriaListConverter;
+import org.hisp.dhis.webapi.mvc.requestconverter.StringToSetConverter;
 import org.hisp.dhis.webapi.view.CustomPathExtensionContentNegotiationStrategy;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -223,6 +225,7 @@ public class WebMvcConfig extends DelegatingWebMvcConfiguration
     {
         registry.addConverter( new StringToOrderCriteriaListConverter() );
         registry.addConverter( new FieldPathConverter() );
+        registry.addConverter( new StringToSetConverter() );
     }
 
     @Primary
--- a/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/helpers/QueryParamsBuilder.java
+++ b/dhis-2/dhis-e2e-test/src/main/java/org/hisp/dhis/helpers/QueryParamsBuilder.java
@@ -63,19 +63,17 @@ public class QueryParamsBuilder
         String[] split = param.split( "=" );
         Optional<MutablePair<String, String>> pairOptional = getByKey( split[0] );
 
-        if ( pairOptional.isPresent() )
+        if ( pairOptional.isEmpty()
+            || pairOptional.get().getKey().equals( "filter" )
+            || pairOptional.get().getKey().equals( "attribute" ) )
+        {
+            queryParams.add( MutablePair.of( split[0], split[1] ) );
+        }
+        else
         {
             MutablePair<String, String> existingPair = pairOptional.get();
-
-            if ( !existingPair.getKey().equals( "filter" ) && !existingPair.getKey().equals( "attribute" ) )
-            {
-                existingPair.setRight( split[1] );
-            }
-
-            return this;
+            existingPair.setRight( split[1] );
         }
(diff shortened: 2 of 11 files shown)


gnespolino (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/WebMvcConfig.java
```diff
@@ -223,6 +225,7 @@ protected void addFormatters( FormatterRegistry registry )
     {
         registry.addConverter( new StringToOrderCriteriaListConverter() );
         registry.addConverter( new FieldPathConverter() );
+        registry.addConverter( new StringToSetConverter() );
```
jbee, line 228: I am pretty sure this would cause issues. For example the `fields` parameter expects a comma separated list  of inputs to result in a list/set of the fields listed. With the replacement now I guess it would result in a set with just a single entry of all the values. We also have other parameters like the `filter` parameter where we do custom split because split by `,` would result in the wrong list. In such cases the parameter is simply modelled as `String` in the controller method and the splitting is done afterwards in dedicated code. I don't see why such a solution would not also work well in your case. If you want a `Set<String>` with just one value of "everything" just model it as `String` in the API and construct the `Set.of` from that API parameter.

stian-sandvold: Just had a quick look at this, and I am not sure I quite follow/understand:

The core of the problem here, is that "attribute=<uid>:<op>:41,0" is split incorrectly, as we want "41,0" to the be search string, not 41 and 0 individually. It makes sense to me that we need to treat the 41,0 as a string, not a set.

As Jan mentions, we have several parameters that expects "," to separate items in a list of values, rather than being part of a single value (IE. fields, filter).

My understanding is that @jbee concern is that this change will alter the behaviour for those other parameters, since we are now changing how everything is converted, instead of just the /query?attribute param.

But now you are saying @lucaCambi77 that this change is not changing how it works? Can we remove that change and the problem is still fixed? If not, is there an alternative way to solve this that does not affect all controllers, but just the endpoint in question? (/Query)?

lucaCambi77: `attribute=::41,0` like any parameter request is a sequence of chars. Then Spring is able to convert this to an object by using default converters.
When can have more than one `?filter` o `attributes` in a request, so we want this to be a set object. Unfortunately, Spring has a default way to do this conversion by splitting at `,` a parameter. Hence attribute=::41,0 will become a set object 

- attribute=::41
- 0

This pr fixes this by using a converter so a string won't break. This is for every controller because I don't see a reason why we need this behavior to be in place.  This is a minimal change that is not affecting the way `?filter` or ?`attributes` work.
Notice that I don't think `?fields`  will use this because it is a List and not a Set. However, `?fields` has already a converter.
If you are worrying about what this breaks, please give me an example so I understand

jbee: Yes, @stian-sandvold is mostly right about my concerns. I missed that what we use elsewhere is `List<String>` and `Set<String>` is otherwise unused as you say @lucaCambi77 . So changing the behaviour for target type `Set<String>` would not effect existing parameters I still think changing its default to not split on `,` is not a good choice as we would expect exactly that behaviour should the target be a `List<String>` or even a `Collection<String>`. I think your use case of not wanting to split on `,` because it is part of a single value is the odd case that should not adjust the common case. Esepcially not if that than behaves inconsistent with similar cases thus be confusing for developers to have a reasonable expectation. Also since the "workaround" for your special case is very simple. Have your value be modelled as `String` parameter and wrap that into a `Set<String>` in the controller code.

lucaCambi77: @jbee 
`Have your value be modelled as String parameter and wrap that into a Set<String> in the controller code.`
With what logic should I then create a Set in the controller? Using a split char? And what if that split char is part of the filter value?

`I think your use case of not wanting to split on, because it is part of a single value is the odd case that should not adjust the common case. `
The odd case is that we rely on Spring to split on `,`. This doesn't make sense and shouldn't be the rule. Applying changes to single controllers as @stian-sandvold is saying will create an anti-pattern because we are moving the problem somewhere else. In my opinion, it should be the opposite, so given no rule on splitting, then every controller should have a specific rule if required. Looking at tests there aren't cases where we rely on Spring default split from String to Set, but in case we can just go there and add for that specific controller. 

The change is documented in the MVC configuration so if we want to apply this to List we can do it, there aren't infinite types of collections and the use case is limited.

Notice there is also another option which is converting from String to a custom type like we are doing for `?fields`. In that case, we could convert from String to a more complex object like `QueryItem` but there are multiple cases and also there is a logic that shouldn't be in a converter

lucaCambi77: if you have like in your example say `checks=a,b,c` and `a=thing1,thing2`, then you can't reliably split on `,` to get distinct values. Yeah, now I see also in the [documentation](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-239/new-tracker.html#common-request-parameters). Let's see in the backend meeting if we can find a solution

PR #13647, opened 2023-04-18
fix: users cannot change their avatar [DHIS2-14223]

netroms:
# Summary
Fixes issue where the user can not update/change their profile picture without having admin like F_USER_ADD authority.
This fix moves this functionality from the /users endpoint to the /me endpoint which it should have been in the first place, since it does not require any special authorities to use. 

### Manual test:
1. Upload a picture by using going to profile and chose upload picture. This will fail but the picture will be uploaded.
2. Use chrome inspector and look at the POST request to /fileResources, capture the result and get the fileResource ID.
3. Use the following curl to link the new uploaded file to the avatar:
```
curl -X PUT --location "http://localhost:8080/api/41/me" \
    -H "Content-Type: application/json;charset=UTF-8" \
    -d "{\"avatar\": {\"id\": \"FILE_RESOURCE_UID\"}}" \
    --basic --user USERNAME:PASSWORD
```

Jira: [DHIS2-14223](https://dhis2.atlassian.net/browse/DHIS2-14223)
Related Jira: [DHIS2-7791](https://dhis2.atlassian.net/browse/DHIS2-7791)

Front end Jira: [DHIS2-15148](https://dhis2.atlassian.net/browse/DHIS2-15148)

[DHIS2-14223]: https://dhis2.atlassian.net/browse/DHIS2-14223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-7791]: https://dhis2.atlassian.net/browse/DHIS2-7791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-15148]: https://dhis2.atlassian.net/browse/DHIS2-15148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13766, opened 2023-04-28
fix: users cannot change their avatar [DHIS2-14223] (2.40.0)

amcgee:
Backport of #13647 (patch/2.40.0)


Philip-Larsen-Donnelly (approved):
Approved as backport

PR #13768, opened 2023-04-28
fix: Validate user uid when creating new users

netroms:


jbee (approved):
I wonder if we have or should have something generic that makes sure all our objects with UID can only be created with valid UIDs.

PR #13779, opened 2023-05-02
fix: disable cookie management for web requests

mortenoh:
Disables cookie management for webhooks.


jbee (approved):

amcgee (approved):
Looks good thanks @mortenoh

PR #13780, opened 2023-05-02
fix: disable cookie management for web requests [DHIS2-15197]

mortenoh:
Disables cookie management for webhooks.


amcgee (approved):
LGTM, RCB approved

Philip-Larsen-Donnelly (approved):

PR #13787, opened 2023-05-02
fix: Misleading message about password requirements [DHIS2-15195]

netroms:
### Summary
This PR removes the password validation method in ValidationUtils.passwordIsValid(). This method was confusing and did not correctly validate the password according to the actual password rules we apply in the PasswordValidationService.validate().
Instead of using ValidationUtils.passwordIsValid(), we now use PasswordValidationService.validate() everywhere where passwords are validated.

The max password length is also changes according to the recommendations in the Jira case.  The new max password length is now 256.

Jira: [DHIS2-15195](https://dhis2.atlassian.net/browse/DHIS2-15195)

[DHIS2-15195]: https://dhis2.atlassian.net/browse/DHIS2-15195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/ValidationUtils.java
```diff
@@ -99,15 +95,11 @@
 
     private static final Pattern INTERNATIONAL_PHONE_PATTERN = Pattern.compile( "^\\+(?:[0-9].?){4,14}[0-9]$" );
 
-    public static final String NOT_VALID_VALUE_TYPE_CLASS = "not_valid_value_type_class";
-
     public static final String NOT_VALID_VALUE_TYPE_OPTION_CLASS = "not_valid_value_type_option_class";
 
-    public static final int MAX_PASSWORD_LENGTH = 35;
+    private static final Set<String> BOOL_FALSE_VARIANTS = Sets.newHashSet( "false", "False", "FALSE", "f", "F", "0" );
```
jbee, line 100: Use `Set.of`

jbee (approved):

mortenoh (approved):

PR #13792, opened 2023-05-03
fix: handle fixed delay jobs in scheduler list API and startup [DHIS2-15241] (2.39)

jbee:
backport of #13782


jbee:
Reopened: While queues and the controller for it does not exist the scheduler startup changes were backported so they need fixing in 2.39 too.

netroms (approved):

PR #13800, opened 2023-05-03
feat: Create endpoints and service for custom icons [DHIS2-15039]

muilpp:
In this PR I'm creating the endpoints, service and database scripts for the custom icon feature.

This feature allows clients to upload, update, delete and fetch icons, as well as persist metadata associated with them. These custom icons coexist with the standard ones, but have their own domain so we can distinguish them from other file resources.

I am using the `IconController` to expose the endpoints. This controller already exists, and I extended by adding endpoints to create, update and delete icons. I also extended the two endpoints that return icons, `getIcon(String)` and getAllIcons(Collection<String), so in case we are dealing a custom icon, two extra properties will be present: `fileResourceUid`, and `userUid`. In case it's a default icon, the response remains the same.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-15039


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/hibernate/HibernateCustomIconStore.java
```diff
@@ -0,0 +1,99 @@
...
+    @Override
+    @Transactional( readOnly = true )
+    public List<CustomIcon> getAllIcons()
+    {
+        return sessionFactory.getCurrentSession().createNativeQuery( "select * from customicon", CustomIcon.class )
+            .getResultList();
+    }
+
+    @Override
+    @Transactional( readOnly = true )
```
enricocolasante, line 76: Transactions should be initiated at service level
https://github.com/dhis2/wow-backend/blob/master/guides/transactions.md

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Test
+    void shouldGetAllKeywordsWhenRequested()
+    {
+        List<String> keywordList = getAllStandardIcons().values().stream()
+            .map( BaseIcon::getKeywords )
+            .flatMap( List::stream ).toList();
+
+        Assertions.assertEquals( keywordList.size() + keywords.size(), iconService.getKeywords().size(),
```
enricocolasante, line 118: Static import

enricocolasante (changes requested):
Double-check the @Transactional annotations

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconStore.java
```diff
@@ -0,0 +1,85 @@
...
+     * Gets all custom icons present in the database
+     *
+     * @return a list containing all icons
+     */
+    List<CustomIcon> getAllIcons();
+
+    /**
+     * Returns a list with all the custom icon keywords
+     *
+     * @return a list will all the custom icon keywords
```
enricocolasante, line 61: typo

```suggestion
     * @return a list with all the custom icon keywords
```

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -27,63 +27,226 @@
...
 
     @Override
-    public Collection<String> getKeywords()
+    @Transactional
+    public void deleteCustomIcon( String key )
+        throws BadRequestException,
+        NotFoundException
+    {
+        CustomIcon icon = validateCustomIconExists( key );
+        fileResourceService.getFileResource( icon.getFileResourceUid() ).setAssigned( false );
```
enricocolasante, line 203: Shouldn't we use private method `getFileResource`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -27,63 +27,226 @@
...
+
+    @Override
+    @Transactional
+    public void addCustomIcon( CustomIcon customIcon )
+        throws BadRequestException,
+        NotFoundException
+    {
+        validateIconExists( customIcon.getKey() );
+        FileResource fileResource = getFileResource( customIcon.getFileResourceUid() );
+        customIconStore.save( customIcon, fileResource.getId(), currentUserService.getCurrentUser().getId() );
```
enricocolasante, line 166: In my opinion `id` is a repository concept and we shouldn't have to deal with it in the service.
I would change the signature of the `save` method to either accept `FileResource` and `User` object and extract the id itself or to just get the `customIcon` and through the `uid` join with the relevant tables and get the needed `id`s

enricocolasante (commented):
If the client cannot really specify the user but we always need to use the current user, let's remove the `userUid` field from the DTO.

I would push the usage of `Id` only in the repository layer

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -0,0 +1,227 @@
...
+    void shouldGetIconsFilteredByKeywordWhenRequested()
+        throws BadRequestException,
+        NotFoundException
+    {
+        Optional<DefaultIcon> defaultIcon = getAllDefaultIcons().values().stream()
+            .filter( si -> si.getKeywords().length > 0 ).findAny();
+
+        if ( defaultIcon.isEmpty() )
+        {
+            return;
```
teleivo, line 132: Why would we not test getting icons by keyword if there is no default icon with keywords?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -0,0 +1,227 @@
...
+        iconService
+            .addCustomIcon(
+                new CustomIcon( "iconKeyC", "description", new String[] { "k6", "k7", "k8" }, fileResourceC.getUid(),
+                    currentUserService.getCurrentUser().getUid() ) );
+        FileResource fileResourceD = createAndPersistFileResource( 'D' );
+        iconService
+            .addCustomIcon( new CustomIcon( "iconKeyD", "description", new String[] { keyword }, fileResourceD.getUid(),
+                currentUserService.getCurrentUser().getUid() ) );
+
+        assertEquals( 1, iconService.getIcons( new String[] { "k4", "k5", "k6" } ).size(),
```
teleivo, line 152: Please use assertions like

`assertContainsOnly(List.of(iconB), iconService.getIcons( new String[] { "k4", "k5", "k6" } ))`

they are easier to read and give more context in case the test fails. You'll get the actual icons instead of just sizes. I think you'll need `@EqualsAndHashCode` on the Custom/DefaultIcons.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertNotNull( iconResource.get().getURL() );
+    }
+
+    @Test
+    void shouldFailWhenGettingIconDataOfNonDefaultIcon()
+    {
+        Exception exception = assertThrows( NotFoundException.class,
+            () -> iconService.getIconResource( "madeUpIconKey" ) );
+
+        String expectedMessage = "No default icon found with key madeUpIconKey.";
```
teleivo, line 183: nit: inline `expectedMessage`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -66,70 +77,43 @@
...
+        throws WebMessageException,
+        NotFoundException
     {
-        Optional<IconData> icon = iconService.getIcon( iconKey );
+        Icon icon = iconService.getIcon( iconKey );
 
-        if ( !icon.isPresent() )
+        if ( icon == null )
         {
             throw new WebMessageException(
```
teleivo, line 101: Why is this still needed? I thought `iconService.getIcon` is already throwing NotFoundException as per our convention 🤔

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -66,70 +77,43 @@
...
     public void getIconData( HttpServletResponse response, @PathVariable String iconKey )
         throws WebMessageException,
-        IOException
+        IOException,
+        NotFoundException
     {
         Optional<Resource> icon = iconService.getIconResource( iconKey );
 
-        if ( !icon.isPresent() )
+        if ( icon.isEmpty() )
```
teleivo, line 116: this should also not be necessary once `iconService.getIconResource` throws NotFoundException (see comment in service)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -140,4 +124,64 @@ public void getIconData( HttpServletResponse response, @PathVariable String icon
 
         StreamUtils.copyThenCloseInputStream( icon.get().getInputStream(), response.getOutputStream() );
     }
+
+    @GetMapping
+    public @ResponseBody List<IconResponse> getAllIcons( @RequestParam( required = false ) String[] keywords )
+    {
+        Collection<Icon> icons;
+
+        if ( keywords == null )
```
teleivo, line 133: Not sure if this is ever happening but can `keywords` be an empty array? If so should we also call `getIcons()` and get all icons?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconDto.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+public class IconDto
```
teleivo, line 39: Can we rename this to `IconBody` and make it package private? Its not super clear to me at first what this class is vs `IconResponse`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -42,7 +45,7 @@
      *
      * @return a collection of data about all the icons in the system
      */
-    Collection<IconData> getIcons();
+    Collection<Icon> getIcons();
 
     /**
      * Gets info about the icons in the system tagged with all the keywords in a
```
teleivo, line 51:
```suggestion
     * Gets icons tagged with all given keywords.
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
```diff
@@ -0,0 +1,138 @@
...
+        List<CustomIcon> customIcons = new ArrayList<>();
+
+        final String sql = """
+                    select c.key as iconkey, c.description as icondescription, c.keywords as keywords, f.uid as fileresourceuid, u.uid as useruid
+                    from customicon c join fileresource f on f.fileresourceid = c.fileresourceid
+                    join userinfo u on u.userinfoid = c.createdby
+                    where keywords @> string_to_array(?,',')
+                    """;
+
+        jdbcTemplate.query(sql, getRowCallbackHandler(customIcons), String.join(",", keywords));
```
teleivo, line 78: Have you tried https://www.baeldung.com/spring-jdbctemplate-in-list#NamedParameterJdbcTemplate so we do not have to join in Java and split again in SQL?

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    private CurrentUserService currentUserService;
+
+    @Autowired
+    private ContextService contextService;
+
+    @Test
+    void shouldCreateCustomIconWhenFileResourceExist()
+        throws IOException
+    {
+        String message = createIcon( createFileResource() );
```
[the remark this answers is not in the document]
muilpp: I understand it might be a bit less clear because of being inside a method, the thing is that we need this in each test in this class. that's why I created it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -51,28 +54,88 @@
...
+        throws NotFoundException;
 
     /**
      * Gets the icon with the correct key if one exists
      *
      * @param key key of the icon
      * @return the icon resource
+     * @throws NotFoundException if no default icon exists with the provided key
+     */
+    Optional<Resource> getIconResource( String key )
```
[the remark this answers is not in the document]
muilpp: legacy reasons, I didn't create nor change the signature of this method

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -27,63 +27,226 @@
...
 
     @Override
-    public Collection<String> getKeywords()
+    @Transactional
+    public void deleteCustomIcon( String key )
+        throws BadRequestException,
+        NotFoundException
+    {
+        CustomIcon icon = validateCustomIconExists( key );
+        getFileResource( icon.getFileResourceUid() ).setAssigned( false );
```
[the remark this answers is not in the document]
muilpp: Yes, there's a job that when run deletes the unassigned file resources.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconDto.java
muilpp: Fine by me, the thing is that I've seen in other controllers this convention of naming the incoming json class "XXXDTO" and the outgoing one "XXXResponse"

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconStore.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return the list of custom icons that contain all the keywords
+     */
+    List<CustomIcon> getIconsByKeywords( String[] keywords );
+
+    /**
+     * Gets all custom icons present in the database
+     *
+     * @return a list containing all icons
+     */
+    List<CustomIcon> getAllIcons();
```
teleivo, line 58: I would be consistent in the signatures across service and store

`List<CustomIcon> getAllIcons();` (Store), `Collection<Icon> getIcons();` (Service)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public List<CustomIcon> getIconsByKeywords(String[] keywords) {
+        final String sql = """
+                    select c.key as iconkey, c.description as icondescription, c.keywords as keywords, f.uid as fileresourceuid, u.uid as useruid
+                    from customicon c join fileresource f on f.fileresourceid = c.fileresourceid
+                    join userinfo u on u.userinfoid = c.createdby
+                    where keywords @> string_to_array(?,',')
+                    """;
+
+        return jdbcTemplate.query(sql, customIconRowMapper, String.join(",", keywords));
```
teleivo, line 73: I had a comment here which I cannot see anymore so I don't know what your answer was 🤷🏻 It is awkward to join into a string and have to split again 😅 

have you looked at
https://www.baeldung.com/spring-jdbctemplate-in-list#NamedParameterJdbcTemplate

The driver does have support for `String[]` to postgres `text[]`
https://access.crunchydata.com/documentation/pgjdbc/42.2.20/arrays.html
which would make the joining and splitting a waste. Not sure if `JdbcTemplate` allows us to pass the `String[]` into `NamedParameterJdbcTemplate`. Prepared statements (used underneath in the JdbcTemplate AFAIK) seem to support arrays.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -27,63 +27,217 @@
...
+    public Resource getIconResource( String key )
+        throws NotFoundException
     {
-        return icons.values();
+        if ( defaultIcons.containsKey( key ) )
+        {
+            return new ClassPathResource( String.format( "%s/%s.%s", ICON_PATH, key, DefaultIcon.Icons.SUFFIX ) );
+        }
+
+        throw new NotFoundException( String.format( "No default icon found with key %s.", key ) );
```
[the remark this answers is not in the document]
muilpp: No, but here we want to fetch the default icons only, I'll change the method name to make it a bit more clear.

dhis-2/dhis-web-api-test/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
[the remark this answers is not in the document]
teleivo: You could still use it in other tests but not hide it in here though 😄

PR #13801, opened 2023-05-03
fix: datastore update skip sharing if namespace ignores it [DHIS2-15087]

jbee:
When a datastore namespace does not use sharing the update has to bypass the sharing check on store level.


amcgee (approved):
LGTM thanks @jbee !

vietnguyen (approved):

PR #13812, opened 2023-05-04
fix: Icon/style in metadata item [2.41-DHIS2-10496_4]

d-bernat:
Please see: https://dhis2.atlassian.net/browse/DHIS2-10496


gnespolino (approved):
Good one

PR #13815, opened 2023-05-04
fix: correct android settings app authority and namespace protection [DHIS2-15087] (2.40.0)

amcgee:
Backport of #13712 

This is a merge-base branch for both 2.40 and patch/2.40.0, if it is updated to match the head of patch/2.40.0 it will need to be force-pushed again for the 2.40 merge.


Philip-Larsen-Donnelly (approved):
Approved as backport

PR #13823, opened 2023-05-05
chore: rename internals of /eventFilters TECH-1537

teleivo:
to use the name event instead of programStageInstance.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programstagefilter/DefaultEventFilterService.java
```diff
@@ -49,48 +49,44 @@
 @RequiredArgsConstructor
 @Service( "org.hisp.dhis.programstagefilter.ProgramStageInstanceFilterService" )
```
enricocolasante, line 50: Rename the service annotation too

enricocolasante (approved):

muilpp (approved):

PR #13830, opened 2023-05-05
chore: Rename pi occurrences [TECH-1547]

enricocolasante:
- Renaming aliases as `pi` or `pis` to `enrollment` and `enrollments`

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentStore.java
@@ -44,18 +44,18 @@ public interface EnrollmentStore
     String ID = EnrollmentStore.class.getName();
 
     /**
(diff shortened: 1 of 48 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentStore.java
```diff
@@ -44,18 +44,18 @@
     String ID = EnrollmentStore.class.getName();
 
     /**
-     * Count all enrollments by PI query params.
+     * Count all enrollments by enrollment query params.
      *
      * @param params EnrollmentQueryParams to use
-     * @return Count of matching PIs
+     * @return Count of matching Enrollments
```
teleivo, line 50:
```suggestion
     * @return Count of matching enrollments
```

I would consistently use lowercase unless unless its at the start of a sentence 🙂

teleivo (approved):

PR #13843, opened 2023-05-08
fix: Removes a wrong annotation in the security configuration

netroms:


Philip-Larsen-Donnelly (approved):
Needed to fix login on master

mortenoh (approved):

PR #13854, opened 2023-05-09
fix: Remove PAT tokens when user is deleted

netroms:
### Summary
Removes all PATs created by the user getting deleted.

## Automatic test
ApiTokenServiceImplTest#testShouldDeleteTokensWhenUserIsDeleted()

## Manual test
1. Login as a non admin user.
2. Create some PATs.
3. Login as admin user.
4. Delete the user from step 1.
5. Try to create a new PAT, with the same user as step 3 or any other user.


jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/apikey/HibernateApiTokenStore.java
```diff
@@ -55,16 +57,18 @@ public HibernateApiTokenStore( SessionFactory sessionFactory, JdbcTemplate jdbcT
     }
 
     @Override
-    public List<ApiToken> getAllOwning( User currentUser )
+    @Nonnull
+    public List<ApiToken> getAllOwning( @Nonnull User currentUser )
```
vietnguyen, line 61: Just a minor comment. This method could be renamed to `findByCreatedBy( User user )`

vietnguyen (approved):

PR #13856, opened 2023-05-09
feat: execute jobs as specific user [DHIS2-15166]

jbee:
### Summary
Adds a persisted  `executedBy` property to `JobConfiguration`.
It **can** be set to the user that should be running the `Job`.
If set, a `SecurityContext` for that user is wrapped around the `Job#execute` call.
No form of password check or any other actual authentication is performed to obtain the user's `SecurityContext` .
Instead 
* system jobs run without an user as superuser
* user created jobs can set any specific user or none to run as superuser
* programmatic created jobs can set any specific user or none to run as superuser

### Automatic Testing
There are tests using the `executeNow` method of the scheduling manager. 
In tests this always has done a synchronous `execute` on the job. 
Since this now will set and clear the `SecurityContext` the context present before the call needs to be restored afterwards as this is not actually running in a worker thread as it would in production.

### Manual Testing
Triggering data integrity manually will run the data integrity checks asynchronously using a job as the user that triggered it.
* `POST` to `/api/dataIntegrity?checks=categories_unique_category_combo` with user auth header (either no `Content-Type` header or `application/json` as value even for empty body)
* check `/api/scheduling/completed/DATA_INTEGRITY` - the root object in the array now has a `userId` property referring to the user that triggered the job (this ID is extracted from the security context so it means execution occurs as that user)

In contrast when running data integrity as a persisted job by creating a configuration in the scheduler app and running it manually the same result of that run `/api/scheduling/completed/DATA_INTEGRITY` does not have a `userId` as the job runs as superuser.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
```diff
@@ -464,6 +465,18 @@ static boolean hasTwoFactorSecretForApproval( User user )
...
+    /**
+     * Use this method instead of {@link #createUserDetails(User)} if no
+     * {@link User} instance is available or if the one available is not fully
+     * loaded or connected to a session.
+     *
+     * @see #createUserDetails(User)
+     * @param userUid UID of the {@link CurrentUserDetails} to create
+     * @return the implementation object
+     */
+    CurrentUserDetails createUserDetails( String userUid )
```
jbee, line 477: FYI: The reason I added this is because the `User` instance usually is not attached to a session in a worker thread which then fails when roles are accessed to compute the details. So it is better to load the user within this method so it is attached to the session opened by the service method.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/parameters/PredictorJobParameters.java
```diff
@@ -47,12 +50,20 @@
  */
 @Getter
 @Setter
+@Builder
 @NoArgsConstructor
+@AllArgsConstructor
 public class PredictorJobParameters implements JobParameters
 {
+    /**
+     * Today plus n days
```
jimgrace, line 59: Thanks for adding the comments! I suggest making this and the following comment:
```
    /**
     * Today plus n days (n can be negative)
     */
```
If n is positive it will be a future date and if n is negative it will be a past date. Both positive and negative offsets have valid use cases for both predictor start and predictor end. "Today plus n days" is technically correct. I just want the reader to know that n can be a negative number. I would not want to say "Today plus or minus n days" because that might give the false impression that a positive n selects a date range of both plus and minus from today. So that's why I suggest adding "(n can be negative)". If you like, "(can be negative)" also works, or anything else that informs the reader that n may be negative.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/parameters/PredictorJobParameters.java
```diff
@@ -47,12 +50,20 @@
...
+     */
     @JsonProperty
     private int relativeStart;
 
+    /**
+     * Today plus n days
```
jimgrace, line 65: Same as previous.

jimgrace (approved):
Looks good to me!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/parameters/PredictorJobParameters.java
jbee: Done

netroms (approved):

PR #13861, opened 2023-05-10
feat: allow for default property types in schema properties

mortenoh:
## Summary

This update introduces a new default property type for the `Property` annotation. Previously, the default type was `TEXT`, which resulted in the property type being overridden whenever the `Property` annotation was used. With this update, you will need to explicitly set the overridden property type; otherwise, it will be automatically detected.


david-mackessy:
I'm not familiar with how this all works yet, but is there any way to add a test to confirm the new behaviour?

jbee (approved):

vietnguyen (approved):

PR #13865, opened 2023-05-10
chore: remove unused Json/Jackson annotations TECH-1552

teleivo:
tracker import/export JSON is defined using other types which have the JSON annotations. Remove all JSON related annotations from tracker types in dhis-api instead of renaming them to the names trackedEntity, enrollment and event.

Use lombok Getters/Setters.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Enrollment.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Enrollment.java
@@ -34,6 +34,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 
+import lombok.Getter;
+import lombok.Setter;
+
 import org.hisp.dhis.audit.AuditAttribute;
 import org.hisp.dhis.audit.AuditScope;
 import org.hisp.dhis.audit.Auditable;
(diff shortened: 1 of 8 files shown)


jbee (commented):
Perfect opportunity to also use `@Getter` (and `@Setter` where needed) ;)

teleivo:
> Perfect opportunity to also use `@Getter` (and `@Setter` where needed) ;)

😂 oh the luck of touching old files 🤦🏻

teleivo:
> To ensure we are not breaking anything, did you check if we are serializing these entities to be audited? Or does anyone know if we are doing that? I hope there are tests for that

@abyot do you know?

jbee (approved):
I don't know about the audit but otherwise it looks good.

teleivo:
From @mortenoh 
> regarding the tracker hibernate objects, the one thing to be aware off.. now that we are using jsonb more and more, we do actually use the jackson annotations on hibernate objects even if they are not used on the outside web-api, for example event.dataValues
This is also true in general for the current audit system, they are all using the actual hibernate objects, so we can not remove any annotations that would invalidate the json object (potentially we should use the tracker DTOs instead in the audit, but we currently do not) 

This change might not be safe. As I am not confident that it will not break auditing I will close it for now.

PR #13874, opened 2023-05-11
feat: Add displayName to WebLocale and use user-selected language [DHIS2-14907]

david-mackessy:
### Summary
**Previous behaviour:**
The `api/locales/ui` and `api/locales/db` endpoints both returned data which was formatted in the server language.

**New behaviour:**
Both endpoints now return data in a formatted language based on this precedence:
1. User UI language is set
2. Server UI language is set
3. English is used as fallback (this hasn't changed, it was always the default).

A new **_displayName_** field has also been added to `WebLocale`. `WebLocale` now returns data represented in 3 fields:
- **_locale_**: e.g. en, fr, zh
- **_name_**: the locale in its own language e.g. English, français, 中国人
- **_displayName_**: the language represented in the user-selected language e.g. English, French, Chinese

Cleaned up `WebLocale` class safely as it's only used in this functionality.

### Automated testing
Web API tests added to cover scenarios
- confirming new format
- confirming new format is returned in the correct language precedence (1. user, 2. system, 3. default)
- changing user language and confirming it takes effect in a new call

### Manual testing
Endpoints updated:
- `api/locales/ui`
- `api/locales/db`

Methods updated:
- `GET`

Expected behaviour:
- new payload should be in the format similar to this:
```
[
    {
        "locale": "en",
        "name": "English",
        "displayName": "anglais"
    },
    {
        "locale": "ar",
        "name": "العربية",
        "displayName": "arabe"
    }
]
```
- the **_locale_** field should always be in the same format e.g. en, fr, es
- the **_name_** field should always be the locale in its own language
- the **_displayName_** field should be in the language based on this precedence when set:
1. user UI language
4. system UI language
5. default English as fallback


jbee (approved):
Good work! 
Tests and description looking good as well.

vietnguyen (approved):

PR #13900, opened 2023-05-11
fix: Add missing pageCount in tracker responses [DHIS2-13649]

abyot:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportController.java
```diff
@@ -134,10 +134,12 @@ PagingWrapper<ObjectNode> getTrackedEntities( TrackedEntityCriteria criteria,
                 ? (long) trackedEntityService.getTrackedEntityCount( queryParams, true, true )
                 : null;
 
+            int pageCount = (int) Math.ceil( count / (double) queryParams.getPageSizeWithDefault() );
```
teleivo, line 137: It might be useful to let the `PagingWrapper` do that for us. Everywhere we use the wrapper we would want the `pageCount` to be `count/pageSize` right?

gnespolino (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportController.java
abyot: done.

teleivo (commented):
Can you please have a look at `/tracker/relationships` which has the same bug 😅

abyot:
> Can you please have a look at `/tracker/relationships` which has the same bug sweat_smile

this is now fixed :)

teleivo (approved):

PR #13932, opened 2023-05-16
fix: TrackedEntityAttribute store entityIntsance to trackedEntity renaming TECH-1552

lucaCambi77:
rename `entityInstance` to `trackedEntity` in `HibernateTrackedEntityAttributeStore`

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/TrackedEntityServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/TrackedEntityServiceTest.java
@@ -151,7 +151,7 @@ class TrackedEntityServiceTest extends TransactionalIntegrationTest
 
     private TrackedEntityInstance teiFemaleA;
 
-    private TrackedEntityAttribute uniqueIdAttribute;
+    private TrackedEntityAttribute generatedAttribute;
 
     private TrackedEntityAttribute trackedEntityAttributeB;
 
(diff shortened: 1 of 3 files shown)


enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/TrackedEntityServiceTest.java
```diff
@@ -502,6 +502,34 @@ void testSavePerson()
         assertEquals( ImportStatus.SUCCESS, importSummary.getStatus() );
     }
 
+    @Test
+    void shouldCreateTeiWhenValidateUniqueAttribute()
```
teleivo, line 506: Can you write this test in https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityServiceTest.java ?

All `deprecated/tracker` code will be removed (one day). Tests of code that will not be removed with old tracker should thus be written in test classes outside of any deprecated package.

teleivo (commented):
Thanks a lot for the fix 😄 
If possible the test should be put in a non deprecated package.

teleivo (commented):
Thanks a lot for the fix 😄 
If possible the test should be put in a non deprecated package.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/TrackedEntityServiceTest.java
lucaCambi77: ok @teleivo this looks difficult to bring in the test class you are mentioning. I've made a similar test for the entity attribute store

lucaCambi77: sorry the test is not correct yet, it needs some adjustment

lucaCambi77: it should be ok now, the test is to check attribute  value uniqueness otherwise it makes only sense in the old tracker importer

teleivo (approved):
Thank you @lucaCambi77 !

PR #13933, opened 2023-05-16
chore: improve OpenAPI spec for /tracker/enrollments TECH-1546

teleivo:
# Why

Our docs

https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-240/tracker.html

are often outdated as its further away from the code. Once we have documented our APIs using OpenAPI specs we can remove the current documentation from the tracker docs. The docs at `docs.dhis2.org` will get a link to our OpenAPI spec and become more high-level than right now.

# What

Describe our API using an OpenAPI spec. Starting with `/tracker/enrollments` as its one of the simpler ones.

* Add markdown to add descriptions to the query parameters. 

* Describe query parameters and properties that are UIDs

* Make getEnrollments public as only public methods are discovered by the `ApiAnalyse.analyseController` right now.

## Things to improve later on

* Some properties/query params are marked as required even though they have a default. Not sure why that is 🤷🏻 
* Find a way to describe metadata properties which can be one of the supported idSchemes. Like `String Enrollment.program` which can be a UID, NAME, CODE or ATTRIBUTE. At least during import and in theory (according to our docs, at least for /tracker/events 😂) also during export.
* Make controller methods package-private once they are discovered by the `ApiAnalyse.analyseController`. Jan is looking into this 😄
* Make `PagingAndSortingCriteriaAdapter` property descriptions reusable. Jan is working on making them globally available.
* Should we describe all the properties of TrackedEntity, Enrollment, Event, Relationship? I think it would be helpful so we all share the same understanding of what they represent.
* Update query parameters that represent a collection of UIDs that are currently separated by `;` instead of `,`. Once we use `,` and `List<String>` we should annotate with `@OpenApi.Property( { UID[].class, Some.class } )`


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
```diff
@@ -82,8 +86,24 @@
 
     private final EnrollmentFieldsParamMapper fieldsMapper;
 
+    @Value
+    @OpenApi.Property
+    @OpenApi.Shared( value = false )
+    private static class ObjectListResponse
```
jbee, line 92: I guess the point of such a generic response class which uses the `EntityType` mechanism for "generics" is to be reused in more than one controller. So if this is needed and useful this should be declared outside of the enrolment specific controller. Otherwise this could simply use `List<Enrollment>` instead to be less complex and do the same.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
teleivo: You are right and I am going to reuse it. I just did not want to think about were to put it just yet 😂 I can go back to the `List<Enrollment>` and make it generic when I reuse it 👌🏻

jbee: Its fine to keep it as is until you reuse it, just wanted to make sure you didn't misunderstand the concept.

dhis-2/dhis-web-api/src/main/resources/openapi/org.hisp.dhis.webapi.controller.tracker.export.enrollment.EnrollmentsExportController.openapi.md
```diff
@@ -0,0 +1,90 @@
...
+
+Get enrollments with given UID(s).
+
+### Parameter: `followUp`
+
+Get enrollments with given follow-up status of the instance for the given program.
+
+### Parameter: `includeDeleted`
+
+Get soft-deleted enrollments by specifying `includeDeleted=true`. Soft-deleted enrollments are excluded by default.
```
enricocolasante, line 27: Just to be consistent with other parameters

```suggestion
Get soft-deleted enrollments by specifying `includeDeleted=true`. Defaults to `false`, meaning that by default soft-deleted enrollments are excluded.
```

enricocolasante (approved):

PR #13945, opened 2023-05-17
chore: improve OpenAPI spec for /tracker TECH-1546

teleivo:
# Why

Our docs

https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-240/tracker.html

are often outdated as they are further away from the code. Once we have documented our APIs using OpenAPI specs we can remove the current documentation from the tracker docs. The docs at `docs.dhis2.org` will get a link to our OpenAPI spec and become more high-level than right now.

# What

Describe our `/tracker` APIs using an OpenAPI spec

* Add markdown to add descriptions to the query parameters. Some are just placeholder we will fill in in a later PR.

* Describe query parameters and properties that are UIDs

* Make controller methods consistently package private now that they are discovered by the `ApiAnalyse.analyseController` 🎉 

## Things to improve later on

* Some properties/query params are marked as required even though they have a default. Not sure why that is 🤷🏻 
* Find a way to describe metadata properties which can be one of the supported idSchemes. Like `String Enrollment.program` which can be a UID, NAME, CODE or ATTRIBUTE. At least during import and in theory (according to our docs, at least for /tracker/events 😂) also during export.
* Should we describe all the properties of TrackedEntity, Enrollment, Event, Relationship? I think it would be helpful so we all share the same understanding of what they represent.
* Update query parameters that represent a collection of UIDs that are currently separated by `;` instead of `,`. Once we use `,` and `List<String>` we should annotate with `@OpenApi.Property( { UID[].class, Some.class } )`
* Describe JSON/CSV endpoints separately as they do not always support the same query paramters. We will need to use a workaround as OpenAPI does not take media types into account. Spring also has some limitations which will require another workaround 😅

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/OpenApiExport.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/OpenApiExport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2022, University of Oslo
+ * Copyright (c) 2004-2023, University of Oslo
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -25,54 +25,36 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package org.hisp.dhis.webapi.controller.tracker.view;
+package org.hisp.dhis.webapi.controller.tracker.export;
 
-import java.time.Instant;
+import java.util.List;
 
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.Value;
 
 import org.hisp.dhis.common.OpenApi;
-import org.hisp.dhis.common.UID;
-import org.hisp.dhis.dataelement.DataElement;
-
-import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
- * @author Morten Olav Hansen <mortenoh@gmail.com>
+ * OpenAPI specifications used across tracker export endpoints.
  */
-@OpenApi.Shared
-@Data
-@Builder
-@NoArgsConstructor
-@AllArgsConstructor
-public class DataValue
+public class OpenApiExport
 {
-    @JsonProperty
-    private Instant createdAt;
+    private OpenApiExport()
+    {
+        throw new IllegalStateException( "Utility class" );
+    }
 
-    @JsonProperty
-    private Instant updatedAt;
+    @Value
+    @OpenApi.Property
+    @OpenApi.Shared( value = false )
+    public static class ListResponse
+    {
+        Integer page = 1;
 
-    @JsonProperty
-    private String storedBy;
+        Integer pageSize = org.hisp.dhis.common.Pager.DEFAULT_PAGE_SIZE;
 
-    @JsonProperty
-    private boolean providedElsewhere;
+        Long total;
 
-    @OpenApi.Property( { UID.class, DataElement.class } )
-    @JsonProperty
-    @Builder.Default
-    private String dataElement = "";
-
-    @JsonProperty
-    private String value;
-
-    @JsonProperty
-    private User createdBy;
-
-    @JsonProperty
-    private User updatedBy;
+        @OpenApi.Property( value = OpenApi.EntityType[].class )
+        List<Object> instances;
+    }
 }
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.webapi.controller.tracker.export.event;
 
+import static org.hisp.dhis.common.OpenApi.Response.Status;
 import static org.hisp.dhis.webapi.controller.tracker.ControllerSupport.RESOURCE_PATH;
 import static org.hisp.dhis.webapi.controller.tracker.export.event.RequestParams.DEFAULT_FIELDS_PARAM;
 import static org.hisp.dhis.webapi.utils.ContextUtils.CONTENT_TYPE_CSV;
@@ -60,6 +61,7 @@ import org.hisp.dhis.tracker.export.event.Events;
 import org.hisp.dhis.webapi.common.UID;
 import org.hisp.dhis.webapi.controller.event.webrequest.PagingWrapper;
 import org.hisp.dhis.webapi.controller.tracker.export.CsvService;
+import org.hisp.dhis.webapi.controller.tracker.export.OpenApiExport;
 import org.hisp.dhis.webapi.controller.tracker.view.Event;
 import org.hisp.dhis.webapi.mvc.annotation.ApiVersion;
 import org.hisp.dhis.webapi.utils.ContextUtils;
@@ -74,6 +76,7 @@ import org.springframework.web.bind.annotation.RestController;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+@OpenApi.EntityType( Event.class )
 @OpenApi.Tags( "tracker" )
 @RestController
 @RequestMapping( value = RESOURCE_PATH + "/" + EventsExportController.EVENTS )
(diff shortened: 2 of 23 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportController.java
```diff
@@ -70,6 +73,7 @@
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
 
+@OpenApi.EntityType( Relationship.class )
 @OpenApi.Tags( "tracker" )
 @RestController
 @RequestMapping( produces = APPLICATION_JSON_VALUE, value = RESOURCE_PATH + "/"
```
jbee, line 79: I was wondering if the `value` here should end with a `/` since the usual approach is that the method level path part starts with a `/`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
```diff
@@ -48,6 +49,7 @@
  *
  * @author Giuseppe Nespolino
  */
+@OpenApi.Property
 @Data
 @NoArgsConstructor
 class RequestParams extends PagingAndSortingCriteriaAdapter
```
jbee, line 55: I noticed that there are multiple classes with the same simple class name. We should make that unique.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportController.java
```diff
@@ -110,6 +115,7 @@
 
     private final TrackedEntityFieldsParamMapper fieldsMapper;
 
+    @OpenApi.Response( status = OpenApi.Response.Status.OK, value = OpenApiExport.ListResponse.class )
```
jbee, line 118: I think importing `Status` as type might read nicer. If that clashes with another enum of the same name I even would go as far as importing the `OK` constant directly as there cannot be misunderstand here if we shorten the full type to just simple type name or just the constant name.

jbee (commented):
Some details.

I changed the markdown naming conventions in my PR #13934 - you don't have to update this now but we have to make sure not to forget about aligning this.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
teleivo: Why is that necessary?

These classes are package private so a good name can be reused without adding prefixes like we did before. The context of `enrollment`, `trackedentity`, ... is clear from the package so repeating that is redundant.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportController.java
teleivo: did it in the other controllers, just forgot here 😅

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/OpenApiExport.java
```diff
@@ -0,0 +1,60 @@
...
+public class OpenApiExport
+{
+    private OpenApiExport()
+    {
+        throw new IllegalStateException( "Utility class" );
+    }
+
+    @Value
+    @OpenApi.Property
+    @OpenApi.Shared( value = false )
```
teleivo, line 48: @jbee should this be shared true or false 🤔 Conceptually I would like `ListResponse` to be shared for the same Entity like `Enrollment`, ... but since this is the generic `ListResponse` I am not sure it should be shared.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
```diff
@@ -165,10 +170,11 @@ private boolean areAllEnrollmentsInvalid( RequestParams requestParams, EventSear
...
 
-    @GetMapping( "{uid}" )
-    public ResponseEntity<ObjectNode> getEvent(
-        @PathVariable String uid,
-        @RequestParam( defaultValue = DEFAULT_FIELDS_PARAM ) List<FieldPath> fields )
+    @OpenApi.Response( OpenApi.EntityType.class )
+    @GetMapping( "/{uid}" )
+    public ResponseEntity<ObjectNode> getEventByUid(
+        @OpenApi.Param( { UID.class, Event.class } ) @PathVariable String uid,
+        @OpenApi.Param( name = "fields", value = String[].class ) @RequestParam( defaultValue = DEFAULT_FIELDS_PARAM ) List<FieldPath> fields )
```
teleivo, line 177: @jbee is the `name = "fields"` required if the param is already `fields`?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/RequestParams.java
```diff
@@ -35,42 +35,59 @@
...
     private String trackedEntity;
 
+    @OpenApi.Property( { UID.class, OrganisationUnit.class } )
     private String orgUnit;
 
     private OrganisationUnitSelectionMode ouMode;
 
     private AssignedUserSelectionMode assignedUserMode;
 
+    @OpenApi.Property( { UID[].class, User.class } )
```
teleivo, line 90: @enricocolasante can you please check if I am correct in using this `User` class. As there is User, UserInfo and some snapshot variants 😅

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/RequestParams.java
```diff
@@ -115,9 +135,11 @@ class RequestParams extends PagingAndSortingCriteriaAdapter
 
     private Set<String> filterAttributes = new HashSet<>();
 
+    @OpenApi.Property( { UID[].class, Enrollment.class } )
     private Set<String> enrollments = new HashSet<>();
 
     private IdSchemes idSchemes = new IdSchemes();
 
+    @OpenApi.Property( name = "fields", value = String[].class )
```
teleivo, line 143: @jbee same here I am wondering whether the `name = "fields"` is necessary 🤔

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportController.java
teleivo: We discussed this and this seems to be correct 😄 The top level mapping should not end in `/` as it allows requests like `/enrollments` and `/enrollments/` to be processed by `getEnrollments`. The endpoints returning single entities have the `/` prefix in their mapping `/{uid}`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/RequestParams.java
```diff
@@ -35,42 +35,59 @@
...
     private String program;
 
+    @OpenApi.Property( { UID.class, ProgramStage.class } )
     private String programStage;
 
     private ProgramStatus programStatus;
 
     private Boolean followUp;
 
+    @OpenApi.Property( { UID.class, TrackedEntity.class } )
```
teleivo, line 80: Another question is what type to use here:

For metadata its simple as we only have one so I am using `org.hisp.dhis.program.Program`.

For tracker data its not so clear as we have the `org.hisp.dhis.webapi.controller.tracker.view.TrackedEntity;` representing the JSON we accept and return and the internal `org.hisp.dhis.trackedentity.TrackedEntity` used to persist.

@jbee and @enricocolasante which would you choose? I think I should stick to the JSON related ones in `org.hisp.dhis.webapi.controller.tracker.view.`

enricocolasante: I would say the view one as it is the one that the client interacts with

enricocolasante (approved):

PR #13951, opened 2023-05-17
chore: refactor password and token generators

netroms:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CodeGenerator.java
```diff
@@ -285,7 +291,116 @@ public static boolean isValidSHA256HexFormat( String s )
...
+     * The password will contain at least one digit, one special character, one
+     * uppercase letter and one lowercase letter.
+     * <p>
+     * The password will not contain more than 2 consecutive letters to avoid
+     * generating words.
+     *
+     * @param passwordSize the size of the password.
+     * @return a random password.
+     */
+    public static char[] generateValidPassword( int passwordSize )
```
jbee, line 317: I am wondering if this is just used in tests or also part of production. Since the first 4/5 characters are off a fix type there is little variation for short passwords. Which would be fine in tests but if we generate short ones in some role in production this might be to small of a space.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/helpers/PasswordGenerator.java
```diff
@@ -0,0 +1,116 @@
...
+     * The password will contain at least one digit, one special character, one
+     * uppercase letter and one lowercase letter.
+     * <p>
+     * The password will not contain more than 2 consecutive letters to avoid
+     * generating words.
+     *
+     * @param passwordSize the size of the password.
+     * @return a random password.
+     */
+    public static String generateValidPassword( int passwordSize )
```
jbee, line 75: This code seems to be a duplicate of the new code in the `CodeGenerator`. We should reuse it in one way or another.

jbee (approved):

david-mackessy (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CodeGenerator.java
netroms: Minimum default length for passwords is 8, and this method it's only used for invites. It also uses a lot of special chars (after 4th char), so entropy is pretty high then. So someone would need to lower the default password length to 4 and do a lot of invites for this to be an issue.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/helpers/PasswordGenerator.java
netroms: Not quite sure how to do this. We would need to publish it like json-tree, right now. We could potentially convert e2e tests into a module like the rest, then this would be trivial to share. This probably would mean we need to change the CI pipelines in GitHub and Travis.

PR #13961, opened 2023-05-18
chore: Create UID type in web [TECH-1565]

enricocolasante:
This is the first try to create a `UID` type and use it in the web layer.
`UID` type held an alphanumeric string of 11 characters and it is not possible to instantiate and invalid `UID`.
For now, it only lives in web layer. Maybe later we can think of having it as a proper type that is used all across the system.

I didn't use the pre-existent class `org.hisp.dhis.common.UID` as this class is used for OpenAPI purposes and to make it work I should have changed all the uses that are already present in the code. Later we can slowly change all the string fields annotated as `UID` for OpenAPI to `UID` type.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
@@ -271,6 +271,16 @@ class EnrollmentsExportControllerTest extends DhisControllerConvenienceTest
                 .getMessage() );
     }
 
+    @Test
+    void shouldReturnABadRequestErrorWhenInvalidUidIsPassedInThePath()
+    {
+        assertEquals(
+            "Value invalidUid is not valid for parameter uid. UID must be an alphanumeric string of 11 characters",
+            GET( "/tracker/enrollments/invalidUid" )
+                .error( HttpStatus.BAD_REQUEST )
+                .getMessage() );
+    }
+
     private Event event()
     {
         Event event = new Event( enrollment, programStage,
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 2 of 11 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
```diff
@@ -0,0 +1,57 @@
...
+package org.hisp.dhis.webapi.common;
+
+import lombok.Getter;
+
+import org.hisp.dhis.common.CodeGenerator;
+
+/**
+ * UID represents an alphanumeric string of 11 characters.
+ */
+@Getter
```
enricocolasante, line 37: I was even thinking to remove the getter annotation and just have a `toString` method that will return the value

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
```diff
@@ -271,6 +271,16 @@ void getEnrollmentByIdNotFound()
                 .getMessage() );
     }
 
+    @Test
+    void shouldReturnABadRequestErrorWhenInvalidUidIsPassedInThePath()
+    {
+        assertEquals(
+            "Value invalidUid is not valid for parameter uid. UID must be an alphanumeric string of 11 characters",
```
teleivo, line 278:
```suggestion
            "Value 'invalidUid' is not a valid UID. UID must be an alphanumeric string of 11 characters.",
```

I would quote the invalid value just in case it can contain whitespace. In that case its easier to spot.

I would also just test this once in a test like the EnumBindingTest. We can then rely on the mechanism in all controllers.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
teleivo: That would be surprising to me at least. So I would not remove the getter.

The class is immutable (even with the getter) which it should be 😄 

I would add `@EqualsAndHashCode` since we use it lots in maps and sets.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
```diff
@@ -0,0 +1,57 @@
...
+public final class UID
+{
+    private static final String VALID_UID_FORMAT = "UID must be an alphanumeric string of 11 characters";
```
teleivo, line 40: How specific do we want to be? Because there is more to a String (`CodeGenerator.CODE_PATTERN`) for it to be a valid UID.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private final String value;
```
teleivo, line 42: If I think of the idScheme `attribute`, `attribute` has a `UID` while its `value` has a different meaning. Maybe it does not matter `identifier` could be an alternative name. I am ok with how it is though and let others weigh in 😄

teleivo (commented):
Nice 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
enricocolasante: I just added that it starts with a letter.
I think it make sense to be really specific here to help client fix the issue.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/common/UIDBindingTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.web.bind.annotation.ResponseBody;
+
+class UIDBindingTest
+{
+
+    private static final String VALID_UID = "bRNvL6NMQXb";
+
+    private static final String INVALID_UID = "invalidUid";
+
+    private static final String ERROR_MESSAGE = "Value 'invalidUid' is not valid for parameter uid. UID must be an alphanumeric string of 11 characters";
```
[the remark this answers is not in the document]
enricocolasante: I did manage to add that information in the error message :grimacing:

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -267,11 +273,19 @@ private String getGenericFieldErrorMessage( String fieldType )
...
     {
-        if ( value == null || (value instanceof String && ((String) value).isEmpty()) )
+        return getNotValidValueMessage( value, field, false );
+    }
+
+    private String getNotValidValueMessage( Object value, String field, boolean isPathVariable )
+    {
+        if ( value == null || (value instanceof String stringValue && stringValue.isEmpty()) )
         {
             return MessageFormat.format( "{0} cannot be empty.", field );
```
teleivo, line 283: can `field` be empty as well? 🤯

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
muilpp: We will need to get the value doing `UID.value`, to me the distinction between the values of `UID` and `attribute` will be clear enough.

muilpp (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
```diff
@@ -147,13 +147,14 @@ public PagingWrapper<ObjectNode> getEnrollments( RequestParams requestParams )
     @OpenApi.Response( OpenApi.EntityType.class )
     @GetMapping( value = "/{uid}" )
     public ResponseEntity<ObjectNode> getEnrollmentByUid(
-        @OpenApi.Param( { UID.class, Enrollment.class } ) @PathVariable String uid,
```
teleivo, line 150: I think you should only remove `UID.class` once you make changes to the `OpenApi` spec generation. I assume the `OpenAPI` spec is not as detailed as before when removing `UID.class`.

enricocolasante: Make sense. I will put it back and we can improve that part later.

teleivo (approved):

PR #13969, opened 2023-05-19
fix: add READ access validation for all references in MetadataImportService [DHIS2-11811]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-11811
#### Summary
- User tries to import a metadata payload with some objects that have inaccessible references. For example: a `Program` with an inaccessible `DataElement` and this `DataElement` exists in database.
- 500 Error with message`duplicate key value violates unique constraint` is thrown for the **inaccessible reference objects** which already exist in database.

### Issue
- Currently the Preheat query applies a `METADATA_READ` sharing check for all importing objects and their references. Hence, inaccessible objects won’t be available in the Preheat map in later steps of the importing process.
- There is `ReferenceCheck` which validates if a reference object exists, but it relies on the Preheat map. In order words, error `invalid reference` will be returned if the User doesn’t have `METADATA_READ` access. However, that error is bypassed if the reference object is included in the importing payload. In this case, the import service considers the reference object doesn’t exist in the database and will try to import it. This leads to a `duplicate key value violates unique constraint` error as the object actually exists, just not accessible for currentUser.

### Fix
- Add a `skipSharing=true` flag to Preheat query. So the PreheatMap will contains all importing objects.
- In `ReferenceCheck` add additional logics to check if current User has `METADATA_READ` access to all reference objects.   Including references of `EmbeddedObject`.

### Test
- Added a new test class `MetadataImportIntegrationTest`
- Manual test can be done using same test data in `MetadataImportIntegrationTest`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/ReferencesCheck.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/ReferencesCheck.java
@@ -77,8 +77,7 @@ public class ReferencesCheck implements ValidationCheck
 
         for ( IdentifiableObject object : joinObjects( persistedObjects, nonPersistedObjects ) )
         {
-            List<PreheatErrorReport> errorReports = checkReferences( object, bundle.getPreheat(),
-                bundle.getPreheatIdentifier(), bundle.isSkipSharing(), ctx );
+            List<PreheatErrorReport> errorReports = checkReferences( bundle, object, ctx );
 
             if ( !errorReports.isEmpty() && object != null )
             {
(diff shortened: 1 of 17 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/ReferencesCheck.java
```diff
@@ -247,8 +283,50 @@ private void checkSharing( IdentifiableObject object, Preheat preheat,
...
+        }
+
+        return new PreheatErrorReport( identifier, object.getClass(), errorCode, property, args );
+    }
+
+    private boolean isShareable( ValidationContext context, Class<?> klass )
+    {
+        Schema schema = context.getSchemaService().getDynamicSchema( klass );
+        return schema != null && schema.isShareable();
+    }
```
jbee, line 307: The `AclService` already has two `isShareable` methods. I think we can add this one, unless it is complicated to get access to the `AclService` here.

jbee (approved):
Looks good!

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/ReferencesCheck.java
vietnguyen: right, i forgot about that. updated now.

PR #13979, opened 2023-05-22
chore: upgrade eclipse formatter to 4.27

netroms:
### Summary
* Upgrades the Eclipse CDT formatter used by the Maven spotless plugin, from 4.9 to 4.27
* This is needed to support JDK17+ features like multiline strings.
* The upgrade triggers quite a bit of cleaning up, most fixes seems related to the previous formatter version not applying the rules correctly everywhere.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -1666,12 +1666,12 @@ public class JdbcEventStore implements EventStore
 
         return switch ( orgUnitSelectionMode )
         {
-        case ALL -> null;
-        case CHILDREN -> getChildrenOrgUnitsPath( params, ouTable );
-        case DESCENDANTS -> getDescendantOrgUnitsPath( params, ouTable );
-        case CAPTURE -> getCaptureOrgUnitsPath( user, ouTable );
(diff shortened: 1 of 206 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1664,13 +1664,14 @@ private String getOrgUnitSql( User user, EventSearchParams params, String ouTabl
...
         }
 
-        return switch (orgUnitSelectionMode) {
-            case ALL -> null;
-            case CHILDREN -> getChildrenOrgUnitsPath(params, ouTable);
-            case DESCENDANTS -> getDescendantOrgUnitsPath(params, ouTable);
-            case CAPTURE -> getCaptureOrgUnitsPath(user, ouTable);
-            case SELECTED -> getSelectedOrgUnitsPath(params, ouTable);
-            default -> getAccessibleOrgUnitsPath(params, user, ouTable);
+        return switch ( orgUnitSelectionMode )
```
david-mackessy, line 1667: all of the switch statements\expressions in this PR look like they are missing indentation. The `case`s and `default` should be indented once from the method curly brackets to indicate that they are enclosed/part of that method. This would make it consistent with how other formatted methods look.

jbee (approved):
Not sure about the switch indent. Is this a change or did we never indent them? I am happy with both but I think we should not change it without discussing it in the backend group.

PR #13981, opened 2023-05-22
fix: escape attribute filter value when it contains comma or colon [ DHIS2-15298 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-15298

Reference:
- comma is the separator for multiple filters `filter1,filter2...`
- colon is the separator for the filter composition `<uid>:<operator>:<value>`
- multi-operator filter is uid:operator:value:operator:value, for example, `abcdef123:lt:10:gt:5`

**An Idea for a possible solution:**

A client can now use a **filter value** with a comma `,` or a colon `:` by escaping those chars with the `backslash`.
A client can be any application invoking a rest URL.

### Step to fix:

- Convert a filter mapped as `Set<String>` to a `String`
- Create a list of filters splitting by `/,`
- Create operator and value splitting by `/:`

This is now applied to `/tracker/trackedEntities` and `tracker/events` endpoints

**Note this doesn't have an impact on the API. Multiple ?filter= will still be mapped as a comma-separated list due to Spring conversion**
 
### Algorithm to split:
We can use a negative lookahead to avoid splitting for a separator.
This works nicely for the colon. However, for the comma, we can have a corner case.  Suppose we have this filter:
`w75KJ2mc4zz:like:some,escaped:value/,A75KJ2mc4zz:like:some_value`
Our regex will fail because we have / as the final char of the filter and it will fail to split.
To solve this, the algorithm will first remove all slashes that are not escaping, so as actual chars of the filter. Then, we will reconstruct the filter positionally,
So our API invocation will be:
`w75KJ2mc4zz:like:some/,escaped/:value//,A75KJ2mc4zz:like:some_value`
In the first stage the filter will become:
`w75KJ2mc4zz:like:some/,escaped/:value,A75KJ2mc4zz:like:some_value`
and keep track of the removed slash. We can now safely split by comma. Finally, we'll add back the removed slashes

Open point:
Shall we use the multiple operators only for ranges (words, digits, and dates)? Due to a general cleanup, now I have restored it for all the operators


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
```diff
@@ -63,11 +63,20 @@ class RequestParams extends PagingAndSortingCriteriaAdapter
...
+    /**
+     * Query filter for attributes
+     */
     private String query;
 
-    private Set<String> attribute = new HashSet<>();
+    /**
+     * Comma separated list of attribute UIDs
+     */
+    private String attribute;
```
jbee, line 74: What you can do in such a case to have this better described in OpenAPI is to annotate with 
```java
@OpenApi.Property( UID[].class, TrackerAttribute.class )
```
I assume this is about `TrackerAttribute`s. 
This would then document this input as being a "array" of tracker attribute UIDs.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
enricocolasante: This is not really a collection of UIDs, right?
This is a collection of filters, that will be formed by a UID, operator and value, right?

I wouldn't fix it here as we have other stories to tackle the TrackedEntity controller but just want to be sure that we are on the same page. If this was a collection of UIDs we wouldn't need all the escaping process that is the core of this PR

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
lucaCambi77: Not sure about this, [docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-239/new-tracker.html#tracked-entities-get-apitrackertrackedentities) says this is a 
- Comma separated values of attribute UIDs
but in fact, looking at the code, we can actually apply filters

PR #13984, opened 2023-05-22
fix: OpenAPI embedded types, parameter required, UIDs

jbee:
Several OpenAPI fixes:
* `boolean` type fields are not considered required by default
* parameters with `defaultValue()` in the annotation are not required
* both `UID` classes can be used for annotations like `@OpenApi.Param( { UID.class, Enrollment.class } )`
* `EmbeddedObject`s are not referred to via `Ref` type in the OpenAPI document
* `id` property of `IdentifiableObject` refers to its UID in JSON, not its `long` id


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/OpenApiGenerator.java
```diff
@@ -227,6 +228,8 @@ private static void addSimpleType( Class<?> source, Consumer<SimpleType.SimpleTy
         addSimpleType( URL.class, schema -> schema.type( "string" ).format( "url" ).nullable( true ) );
         addSimpleType( UUID.class, schema -> schema.type( "string" ).format( "uuid" ).nullable( true )
             .pattern( "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$" ) );
+        addSimpleType( UID.class, schema -> schema.type( "string" ).format( "uid" ).nullable( true )
+            .pattern( "^[a-zA-Z][0-9a-zA-Z]{10}$" ) );
```
teleivo, line 232: Not sure if it makes sense is feasible to reuse https://github.com/dhis2/dhis2-core/blob/4f528e745971be32740df144db827713caf2cb13/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CodeGenerator.java#L71 as that pattern is used in validating UIDs for example in the new UID type.

teleivo (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/OpenApiGenerator.java
jbee: I think we could do that.

teleivo (approved):

larshelge (approved):

david-mackessy (approved):

PR #14004, opened 2023-05-23
chore: use new UID type instead of String in tracker request params TECH-1565

teleivo:
to make use of the validation we now have for UIDs 🤩 

* added `toString` to UID so we simply print the UID value
* create factory for UID from anything that has a UID (UidObject)

NOTE: the mappers will be tackled soon in https://dhis2.atlassian.net/browse/TECH-1534 The code that had to change now `validation` methods will move into services. We can then figure out a way to remove the duplication/integrate better with the new UID type.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
@@ -37,6 +37,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
 import org.hisp.dhis.common.CodeGenerator;
+import org.hisp.dhis.common.UidObject;
 
 /**
  * UID represents an alphanumeric string of 11 characters starting with a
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
```diff
@@ -59,11 +60,22 @@ private UID( String value )
         this.value = value;
     }
 
+    @Override
+    public String toString()
+    {
+        return value;
```
jbee, line 66: I guess you could also use lombok `@ToString` as well as `@EqualsAndHashCode`

jbee (approved):

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/common/UID.java
teleivo: We are using `@EqualsAndHashCode` 👌🏻 

I tried to use @ToString but it will print `UID(value=JW6BrFd0HLu)` while I prefer `JW6BrFd0HLu`. So that we only insert the value in error messages instead of the value with its internal representation.

PR #14014, opened 2023-05-24
feat: OpenAPI user specified shared parameter base name

jbee:


enricocolasante (approved):
If this get merged than do we still need https://github.com/dhis2/dhis2-core/pull/14013?

teleivo (approved):

PR #14038, opened 2023-05-25
chore: document deprecated /tracker/relationships?tei parameter TECH-1546

teleivo:
We renamed the `tei` field to the new name 'trackedEntity' a while ago. 2 setters provided the old `tei` (`setTei`) and new query parameter `trackedEntity` (`setTrackedEntity`). Our OpenAPI generation does not provide the same mechanism as Spring. We therefore need to add the deprecated field so we can document it in our spec. We will be able to remove it in 2.42.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RequestParams.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RequestParams.java
@@ -52,7 +52,16 @@ class RequestParams extends PagingAndSortingCriteriaAdapter
 {
     static final String DEFAULT_FIELDS_PARAM = "relationship,relationshipType,from[trackedEntity[trackedEntity],enrollment[enrollment],event[event]],to[trackedEntity[trackedEntity],enrollment[enrollment],event[event]]";
 
+    /**
+     * @deprecated use {@link #trackedEntity} instead
+     */
+    @Deprecated( since = "2.41" )
     @OpenApi.Property( { UID.class, TrackedEntity.class } )
+    @Setter
+    private UID tei;
+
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RequestParams.java
```diff
@@ -74,11 +74,13 @@ class RequestParams extends PagingAndSortingCriteriaAdapter
     @Setter
     private List<FieldPath> fields = FieldFilterParser.parse( DEFAULT_FIELDS_PARAM );
 
+    /**
+     * @deprecated use {@link #setTrackedEntity(UID)} instead
+     */
+    @OpenApi.Property( name = "tei", value = { UID.class, TrackedEntity.class } )
+    @Deprecated( since = "2.41", forRemoval = true )
```
jbee, line 81: I made field annotations take precedence. If I see this correct you already have an annotation on the field, just that you do not set the name in that one.

teleivo: What we need is to document query parameters

1. `trackedEntity`
2. `tei` (deprecated)

separately 😄 We are only using a field for 1. but have setters for both. This is how Spring is able to support our use case.

Is there a way for us to document them separately? With or without changing our code?

teleivo: I went the straightforward way adding another field. The solution before was neat but we'll remove the deprecated field in a later release anyway 👌🏻

enricocolasante (approved):

jbee (approved):

PR #14048, opened 2023-05-26
chore: Harmonize enrollment controller [TECH-1564]

enricocolasante:
Goal of this PR is to create `EnrollmentOperationParams` object for the service layer.

**Detailed changes:**
- Change `trackedEntityUid` field in `EnrollmentQueryParams` to be consistent with other fields.
- Create `EnrollmentOperationParams` object that is meant to contain parameters for the service layer, it is created by converting `RequestParams` and it will be converted to `EnrollmentQueryParams`.
- Separate validations by layer. Move to service layer validations that are accessing DB. (ie. Checking if some metadata exists)

**Next PRs**
- Improving/Rename `EnrollmentParams`
- Pagination

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -212,11 +212,6 @@ public class DefaultEnrollmentService implements org.hisp.dhis.tracker.export.en
 
         EnrollmentQueryParams queryParams = opeationParamsMapper.map( params );
 
-        if ( !params.isPaging() && !params.isSkipPaging() )
-        {
-            params.setDefaultPaging();
-        }
-
         List<Enrollment> enrollmentList = new ArrayList<>(
             enrollmentService.getEnrollments( queryParams ) );
         if ( !params.isSkipPaging() )
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -45,6 +45,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.hisp.dhis.common.Pager;
 import org.hisp.dhis.common.SlimPager;
+import org.hisp.dhis.feedback.BadRequestException;
 import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.program.Enrollment;
@@ -77,6 +78,8 @@ public class DefaultEnrollmentService implements org.hisp.dhis.tracker.export.en
     private final CurrentUserService currentUserService;
 
     private final TrackerAccessManager trackerAccessManager;
+
+    private final EnrollmentOperationParamsMapper opeationParamsMapper;
 
     @Override
     public Enrollment getEnrollment( String uid, EnrollmentParams params )
@@ -201,10 +204,13 @@ public class DefaultEnrollmentService implements org.hisp.dhis.tracker.export.en
     }
 
     @Override
-    public Enrollments getEnrollments( EnrollmentQueryParams params )
-        throws ForbiddenException
+    public Enrollments getEnrollments( EnrollmentOperationParams params )
+        throws ForbiddenException,
+        BadRequestException
     {
         Enrollments enrollments = new Enrollments();
+
+        EnrollmentQueryParams queryParams = opeationParamsMapper.map( params );
 
         if ( !params.isPaging() && !params.isSkipPaging() )
         {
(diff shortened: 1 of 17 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParams.java
```diff
@@ -0,0 +1,179 @@
...
+import lombok.Builder;
+import lombok.Getter;
+
+import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.program.ProgramStatus;
+import org.hisp.dhis.webapi.controller.event.mapper.OrderParam;
+
+@Getter
+@Builder
+public class EnrollmentOperationParams
```
muilpp, line 44: if we use the builder pattern, should we create a private constructor to make sure we just use one way of initializing the class?
also, I'm not sure the builder default values would be set for the default constructor.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapper.java
```diff
@@ -0,0 +1,188 @@
...
+        params.setTotalPages( operationParams.isTotalPages() );
+        params.setSkipPaging( toBooleanDefaultIfNull( operationParams.isSkipPaging(), false ) );
+        params.setIncludeDeleted( operationParams.isIncludeDeleted() );
+        params.setUser( user );
+        params.setOrder( operationParams.getOrder() );
+
+        return params;
+    }
+
+    private Program validateProgram( String uid )
```
muilpp, line 109: will these four validations remain in this mapper?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -201,36 +204,39 @@ private Set<TrackedEntityAttributeValue> getTrackedEntityAttributeValues( User u
...
+        EnrollmentQueryParams queryParams = opeationParamsMapper.map( params );
+
         if ( !params.isPaging() && !params.isSkipPaging() )
         {
             params.setDefaultPaging();
         }
 
         List<Enrollment> enrollmentList = new ArrayList<>(
-            enrollmentService.getEnrollments( params ) );
+            enrollmentService.getEnrollments( queryParams ) );
```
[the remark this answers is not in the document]
enricocolasante: Yes, the goal here is to make all the changes in the controller and in the tracker service, without changing the signature of the service-core services.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapper.java
enricocolasante: For now I would say yes. If then we feel the need to extract them to a validator, we can go that way.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -282,7 +288,7 @@ private List<Enrollment> getEnrollments( Iterable<Enrollment> enrollments )
             if ( enrollment != null && trackerOwnershipAccessManager
                 .hasAccess( user, enrollment.getTrackedEntity(), enrollment.getProgram() ) )
             {
-                enrollmentList.add( getEnrollment( enrollment, EnrollmentParams.FALSE ) );
+                enrollmentList.add( getEnrollment( enrollment, params ) );
```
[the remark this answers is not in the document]
enricocolasante: I think so, I am not sure why this was hardcoded to `EnrollmentParams.FALSE` because when getting enrollments, then the `getEnrollment` method is called and it should return the `fields` requested by the client

muilpp (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
```diff
@@ -133,7 +131,7 @@ PagingWrapper<ObjectNode> getEnrollments( RequestParams requestParams )
 
     @OpenApi.Response( OpenApi.EntityType.class )
     @GetMapping( value = "/{uid}" )
-    public ResponseEntity<ObjectNode> getEnrollmentByUid(
```
teleivo, line 136: why can we remove this now? Did you check that the OpenAPI spec still looks well?

The class level `@OpenApi.EntityType( Enrollment.class )` with method level `@OpenApi.Response( OpenApi.EntityType.class )` set the response type since the OpenAPI generation would not know what `ResponseEntity<ObjectNode>` is.

enricocolasante: I haven't check and this change wasn't suppose to be here. Reverted

PR #14068, opened 2023-05-29
fix: Use search scope if no program provided [DHIS2-15021]

muilpp:
As stated by @Bekkalizer, when no program is specified in the request, we would want to give the user access to tracked entity type attributes and the tracked entity itself based on the search org unit. We would however not want to expose tracked entity attributes or enrollment data from protected or closed programs based on this same rule - these would have to be in capture scope. 

ticket: https://dhis2.atlassian.net/browse/DHIS2-15021

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManagerTest.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManagerTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManagerTest.java
```diff
@@ -0,0 +1,101 @@
...
+class DefaultTrackerAccessManagerTest
+{
+
+    @Mock
+    private OrganisationUnitService organisationUnitService;
+
+    @InjectMocks
+    private DefaultTrackerAccessManager trackerAccessManager;
+
+    @Test
```
enricocolasante, line 56: This test can be really important to understand how access control works.
I wouldn't spare any case to test even if it can feel a little bit redundant, 
I would add tests like
`shouldNotHaveAccessWhenProgramOpenAndSearchAccessNotAvailable`
`shouldHaveAccessWhenProgramNullAndSearchAccessAvailable`
`shouldNotHaveAccessWhenProgramClosedAndCaptureAccessNotAvailable`
`shouldHaveAccessWhenProgramProtectedAndCaptureAccessAvailable`

teleivo (approved):

enricocolasante (approved):

PR #14097, opened 2023-05-31
fix: EDI - parsing error for numeric expression [2.41-DHIS2-15332]

d-bernat:
Please see https://dhis2.atlassian.net/browse/DHIS2-15332.


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/expressiondimensionitem/ExpressionDimensionItemService.java
```diff
@@ -101,6 +101,11 @@ else if ( uids.length > 0 )
      */
     public boolean isValidExpressionItems( String expression )
     {
+        if ( expression.chars().allMatch( Character::isDigit ) )
```
gnespolino, line 104: There's a StringUtils.isNumeric helper in apache commons, we might want to use that

gnespolino (approved):

PR #14099, opened 2023-05-31
docs: document /tracker/... fields, filter and filterAttributes parameters TECH-1568

teleivo:
I took the operators for `filter` from https://github.com/dhis2/dhis2-core/blob/07a892eaeab83eaa95778047cb898da9cc53f492/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java#L71-L86

Maybe we can find a way to add `@OpenApi.Property` with some custom type to tie the `filter` to the valid operators declared in the code instead of copying them into the description.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
+++ b/dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
@@ -131,3 +131,73 @@ Supported properties are `assignedUser`, `assignedUserDisplayName`, `attributeOp
 `completedBy`, `createdAt`, `createdBy`, `deleted`, `enrolledAt`, `enrollment`, `enrollmentStatus`, `event`, `followup`,
 `occurredAt`, `orgUnit`, `orgUnitName`, `program`, `programStage`, `scheduleAt`, `status`, `storedBy`, `trackedEntity`,
 `updatedAt`, `updatedBy`.
+
+### `*.parameter.EventRequestParams.fields`
+
+Get only the specified fields in the JSON response. This query parameter allows you to remove unnecessary fields from
+the JSON response and in some cases decrease the response time. Refer to
+https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_metadata_field_filter
+for how to use it.
+
+NOTE: this query parameter has no effect on a CSV response!
+
+### `*.parameter.EventRequestParams.filter`
+
+`<filter1>[,<filter2>...]`
+
+Get events matching given filters on data values. A filter is a colon separated data element UID with operator and value
+pairs. Example: `filter=H9IlTX2X6SL:sw:A` with operator starts with `sw` followed by a value. Special characters
+like `+` need to be percent-encoded so `%2B` instead of `+`. Multiple operator/value pairs for the same data element
+like `filter=AuPLng5hLbE:gt:438901703:lt:448901704` are allowed. Repeating the same data element UID is not allowed.
+Operator and values are case-insensitive. A user needs metadata read access to the data element and data read access to
+the program (if the program is without registration) or the program stage (if the program is with registration).
+
+Valid operators are:
+
+- `EQ` - equal to
+- `IEQ` - equal to
+- `GE` - greater than or equal to
+- `GT` - greater than
+- `LE` - less than or equal to
+- `LT` - less than
+- `NE` - not equal to
+- `NEQ` - not equal to
+- `NIEQ` - not equal to
+- `IN` - equal to one of the multiple values separated by semicolon ";"
+- `ILIKE` - is like (case-insensitive)
+- `LIKE` - like (free text match)
+- `NILIKE` - not like
+- `NLIKE` - not like
+- `SW` - starts with
+- `EW` - ends with
+
+### `*.parameter.EventRequestParams.filterAttributes`
+
+`<filter1>[,<filter2>...]`
+
+Get events matching given filters on tracked entity attributes. A filter is a colon separated attribute UID with
+optional operator and value pairs. Example: `filter=H9IlTX2X6SL:sw:A` with operator starts with `sw` followed by a
+value. Special characters like `+` need to be percent-encoded so `%2B` instead of `+`. Multiple operator/value pairs for
+the same attribute like `filter=AuPLng5hLbE:gt:438901703:lt:448901704` are allowed. Repeating the same attribute UID is
+not allowed. Operator and values are case-insensitive. A user needs metadata read access to the attribute and data
+read access to the program (if the program is without registration) or to the program stage (if the program is with
+registration).
+
+Valid operators are:
+
+- `EQ` - equal to
+- `IEQ` - equal to
+- `GE` - greater than or equal to
+- `GT` - greater than
+- `LE` - less than or equal to
+- `LT` - less than
+- `NE` - not equal to
+- `NEQ` - not equal to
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
```diff
@@ -131,3 +131,71 @@ Supported properties are `assignedUser`, `assignedUserDisplayName`, `attributeOp
...
+for how to use it.
+
+NOTE: this query parameter has no effect on a CSV response!
+
+### `*.parameter.EventRequestParams.filter`
+
+`<filter1>[,<filter2>...]`
+
+Get events matching given filters on data values. A filter is a colon separated data element UID with optional
+operator and value pairs. Example: `filter=H9IlTX2X6SL:sw:A` with operator starts with `sw` followed by a value. Special
```
enricocolasante, line 149: What does it mean `optional` operator and value pairs?
Can we just filter on a data value? If yes, what is the expected result?

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
```diff
@@ -131,3 +131,71 @@ Supported properties are `assignedUser`, `assignedUserDisplayName`, `attributeOp
...
+characters like `+` need to be percent-encoded so `%2B` instead of `+`. Multiple operator/value pairs for the same
+data element like `filter=AuPLng5hLbE:gt:438901703:lt:448901704` are allowed. Repeating the same data element UID is not
+allowed. Operator and values are case-insensitive. A user needs access to the attribute to filter on it. A user needs
```
enricocolasante, line 152: Filter are on data values, right? I am not sure I understand why you are talking about attribute here. Did you mean data element or am I missing something?

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
```diff
@@ -131,3 +131,71 @@ Supported properties are `assignedUser`, `assignedUserDisplayName`, `attributeOp
...
+- `NIEQ` - not equal to
+- `IN` - equal to one of the multiple values separated by semicolon ";"
+- `ILIKE` - is like (case-insensitive)
+- `LIKE` - like (free text match)
+- `NILIKE` - not like
+- `NLIKE` - not like
+- `SW` - starts with
+- `EW` - ends with
+
+### `*.parameter.TrackedEntityRequestParams.filterAttributes`
```
enricocolasante, line 174:
```suggestion
### `*.parameter.EventRequestParams.filterAttributes`
```

enricocolasante (commented):
Just a few questions on filters

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
teleivo: its optional but I don't know what is expected. its weird is what I see 😂 maybe someone else knows? at least when I tried it did not seem to mean, return events that have a data value with this data element.

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
teleivo: This should be data value or data element. Do you know which one the user needs to have access to and what access?

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
Bekkalizer: I didn't know that the operator and value pairs was optiona. It sounds sensible to allow this as a syntax for filtering on non-null values - but is there any precedence on this in the code or other endpoints?

teleivo: @enricocolasante @Bekkalizer the operator is optional but it does not work in a way that makes sense to me.

Here I get events with a data value of data element `DX4LVYeP7bw` equal to 50

```
curl --silent -u system:System123 -H 'content-type: application/json' 'https://play.dhis2.org/dev/api/tracker/events?orgUnit=DiszpKrYNg8&page=1&pageSize=1&fields=event,dataValues&filter=DX4LVYeP7bw:EQ:50' | jq
{
  "page": 1,
  "pageSize": 1,
  "instances": [
    {
      "event": "SObENdEf76z",
      "dataValues": [
        {
          "createdAt": "2019-08-21T13:27:26.629",
          "updatedAt": "2019-09-27T00:01:48.287",
          "storedBy": "system",
          "providedElsewhere": false,
          "dataElement": "DX4LVYeP7bw",
          "value": "50"
        },
        {
          "createdAt": "2019-08-21T13:27:26.629",
          "updatedAt": "2019-09-27T00:01:48.287",
          "storedBy": "system",
          "providedElsewhere": false,
          "dataElement": "k0rev4WSffi",
          "value": "20"
        }
      ]
    }
  ]
}
```

the same request without operator

```
curl --silent -u system:System123 -H 'content-type: application/json' 'https://play.dhis2.org/dev/api/tracker/events?orgUnit=DiszpKrYNg8&page=1&pageSize=1&fields=event,dataValues&filter=DX4LVYeP7bw' | jq    {
  "page": 1,
  "pageSize": 1,
  "instances": [
    {
      "event": "rBjxtO8npTb",
      "dataValues": []
    }
  ]
}
```

teleivo: So if we think optional operators could have a use we would need to define that first and implement it accordingly. Should we thus hide the fact that the operator is optional from the docs for now? Then figure out how it should behave, implement and document it.

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
teleivo: @Bekkalizer @enricocolasante Do you know if the user needs to have access to data values and/or data elements and what access?

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
Bekkalizer: Agreed - I think that is a good approach.

dhis-2/dhis-web-api/src/main/resources/openapi/EventsExportController.md
teleivo: @Bekkalizer @enricocolasante do you know if the user needs to have access to data values and/or data elements and what access? I would love to get this merged and this is the only missing piece 😄

abyot: since you are on export, user needs to have `Permission.DATA_READ` access at program stage level. We don't have data sharing at data element level.

teleivo: @abyot and what access is needed to filter on attributes?

abyot: like data elements, we don't have data sharing at attribute level. We instead rely on data sharing at program level. so you just check data read access at program level.

by the way for both data element and attribute we have metadata sharing and a user needs to have metadata read access to get data read access.

teleivo: @abyot please let me know if the changed sentence is correct 😄

zubaira (approved):

PR #14100, opened 2023-05-31
feat: Add copy functionality for Program [DHIS2-15279]

david-mackessy:
# Summary
New endpoint added to allow copying of an existing Program:
`/api/programs/{uid}/copy`

The endpoint will accept optional params which can be used in the copy process. The only param that we initially support is the `prefix` param which is used to prefix the `Program` and `ProgramStage` `name` property only. e.g.  
`/api/programs/{uid}/copy?prefix=test`  

## Copying
Each class has a mix of shallow and deep copying.  
### Deep copy
The classes that require deep copying are:
- `Program`
- `ProgramSection`
- `ProgramTrackedEntityAttribute`
- `ProgramIndicator`
- `ProgramRuleVariable`
- `ProgramStage`
- `ProgramStageSection`
- `ProgramStageDataElement`
- `Enrollment`  

Each of these classes have a new `copyOf` static method, are copied as a new instance and get new unique property values where appropriate e.g.
- `id`
- `uid`

### Shallow copy
Every other property value gets shallow copied:
- collection -> new instance of the same collection
- primitive -> same value
- object -> same reference  

### Copy v Clone  
The `copy` approach was preferred to over the Java built-in `clone` approach. Some reasons why:
- Copy
    - Full control over what is/isn't copied
    - Lower chance of corrupt state as always starting from a fresh new instance
    - clear visibilty over explicit shallow & deep copies
- Clone
    - Higher chance of having corrupt state as once an object is cloned, it may have some state which it shouldn't (e.g. uid)
    - More ceremony involved:
        - implementing interfaces
        - catching exceptions
        - casting
    - New fields added would be shallow-cloned by default which may not be what we want

## Hibernate
There are no Hibernate changes implemented.

# Testing

## Automated
- Unit tests added for new code
- Service tests
- Controller tests  

Unit tests have been added, with a comment, in each impacted API class to assert the number of fields matches its current field count. If new fields are added, the test will fail & consideration should given to how that new field should be copied.

## Manual
Tested manually using Postman

#### Copy an existing `Program`  
Program can be created first:
- Use the JSON payload at `dhis-2/dhis-test-web-api/src/test/resources/program/create_program.json`
  - This payload includes all items that are deep copied from `Program`
  - It can be updated to include any of its other properties while testing 
- `POST` request to `api/metadata`

Then to copy:
Request: `POST` -> `/api/programs/{originalUid}/copy`  
Response: 
```json
{
    "httpStatus": "Created",
    "httpStatusCode": 201,
    "status": "OK",
    "message": "Program created: '{copyUid}'"
}
```  
Verify new `Program` details with call:  
Request: `GET` -> `/api/programs/{copyUid}`  
</br>

#### Copy a non existing `Program`  
Request: `POST` request to `/api/programs/{uid}/copy`  
Response: 
```json
{
    "httpStatus": "Not Found",
    "httpStatusCode": 404,
    "status": "ERROR",
    "message": "Program with id {uid} could not be found.",
    "errorCode": "E1005"
}
```
</br>

#### Copy an existing `Program` which has a name of length 228  
The default prefix if none provided is `Copy of ` which is 8 chars long.
Request: `POST` -> `/api/programs/{uid}/copy`  
Response: 
```json
{
    "httpStatus": "Conflict",
    "httpStatusCode": 409,
    "status": "ERROR",
    "message": "ERROR: value too long for type character varying(230)",
    "errorCode": "E1004"
}
```


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/copy/CopyService.java
```diff
@@ -0,0 +1,107 @@
...
+import org.hisp.dhis.program.Enrollment;
+import org.hisp.dhis.program.EnrollmentService;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.program.ProgramCopyTuple;
+import org.hisp.dhis.program.ProgramEnrollmentsTuple;
+import org.hisp.dhis.program.ProgramService;
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.stereotype.Service;
+
+@Service
```
vietnguyen, line 49: Do we need `@Transactional` for this service ?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/copy/CopyService.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private EnrollmentService enrollmentService;
+
+    public CopyService( ProgramService programService, EnrollmentService enrollmentService )
+    {
+        this.programService = programService;
+        this.enrollmentService = enrollmentService;
+    }
+
+    public String copyProgramFromUid( String uid, Map<String, String> copyOptions )
```
vietnguyen, line 62: I think we need `@Transactional` for this method since it does both get and save entity to database.

david-mackessy: yep, nice catch, thanks. The underlying services this method calls do have their own transactional methods for saving but it should be wrapped in one transaction (Program + Enrollments). I'll add this.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
```diff
@@ -904,4 +915,67 @@ public void setAccessLevel( AccessLevel accessLevel )
...
+        copy.setSkipOffline( original.isSkipOffline() );
+        copy.setUseFirstStageDuringRegistration( original.getUseFirstStageDuringRegistration() );
+        copy.setCategoryCombo( original.getCategoryCombo() );
+        copy.setDataEntryForm( original.getDataEntryForm() );
+        copy.setExpiryPeriodType( original.getExpiryPeriodType() );
+        copy.setNotificationTemplates( newSetFromObjectOrEmpty( original.getNotificationTemplates() ) );
+        copy.setOrganisationUnits( newSetFromObjectOrEmpty( original.getOrganisationUnits() ) );
+        copy.setProgramAttributes( newListFromObjectOrEmpty( original.getProgramAttributes() ) );
+        copy.setProgramIndicators( newSetFromObjectOrEmpty( original.getProgramIndicators() ) );
+        copy.setProgramRuleVariables( newSetFromObjectOrEmpty( original.getProgramRuleVariables() ) );
```
vietnguyen, line 961: I think we need to create new entities for those `ProgramRuleVariable`s because we have one-to-many assocation between `Program` and `ProgramRuleVariable`. 
Same for other properties such as `ProgramSection`, `ProgramIndicator` , `ProgramTrackedEntityAttribute`.

david-mackessy: Yep, I see your point. It does look like they are needed as deep copies also. I wonder was there any specific reason why these were not included in the original Jira issue. @larshelge does this sound right to you also?
So these would be the extra deep copies required:
- `Set<ProgramSection> programSections`
- `List<ProgramTrackedEntityAttribute> programAttributes`
- `Set<ProgramIndicator> programIndicators`
- `Set<ProgramRuleVariable> programRuleVariables`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/ObjectUtils.java
```diff
@@ -193,4 +195,34 @@ public static <T, U extends RuntimeException> T throwIfNull( T object, Supplier<
...
+     * Util method that always returns a new Set, either instantiated from a
+     * non-null Set passed as an argument, or if a null arg is passed then
+     * returning an empty Set. This helps reduce possible NullPointerExceptions
+     * when trying to instantiate a Set with a null value.
+     *
+     * @param set
+     * @return
+     * @param <T>
+     */
+    public static <T> Set<T> newSetFromObjectOrEmpty( Set<T> set )
```
jbee, line 209: There is a `List.copyOf` and `Set.copyOf` in the JDK. Is there a problem with immutable collections? If we want a mutable version of these I think we should also name them `copyOf`. It would also be good for such utility methods if you add the `@CheckForNull` and `@Nonnull` annotations to the signature especially if the method motivation is just about the null problem.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/StreamUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * the actual stream of elements if non-null. This will help reduce
+     * NullPointExceptions.
+     *
+     * @param collection of any type
+     * @return a stream of elements (of Type T) from the collection if the
+     *         collection is not null or, return an empty stream if the
+     *         collection is null.
+     * @param <T> is the Type of element
+     */
+    public static <T> Stream<T> nullSafeCollectionToStream( Collection<T> collection )
```
jbee, line 57: `streamOf`` with nullness annotations should do it and result in more readable code. Not sure if someone mentioned for you that we have the convention that generally we do assume any collection type to be nun-null. If you have code that makes different assumptions it should definitely use the nullness annotation.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
```diff
@@ -0,0 +1,236 @@
...
+class ProgramControllerTest extends DhisControllerConvenienceTest
+{
+    public static final String PS1_UID = "ps1abcdes9i";
+
+    public static final String PS2_UID = "ps2abcdes9i";
+
+    private OrganisationUnit orgUnit;
+
+    @BeforeEach
+    public void testSetup()
```
jbee, line 76: Would be nice to also do the setup using the API if it not gave you trouble for some reason. seeing the API calls allows to easily replay the test manually.

jbee (approved):
Looks good overall. Nice you wrote lots of tests. I did not have time to review the hibernate mapping changes too careful. Also the formatting changes made this very hard to see the actual changes.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCopyTuple.java
```diff
@@ -0,0 +1,38 @@
...
+package org.hisp.dhis.program;
+
+/**
+ * record used to hold 2 instances of Program
+ *
+ * @author David Mackessy
+ * @param copy
+ * @param original
+ */
+public record ProgramCopyTuple(Program copy, Program original) {
```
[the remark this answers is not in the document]
david-mackessy: I'll move move this into Program. It's used in the API and Services modules.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/ObjectUtils.java
david-mackessy: `List/Set.copyOf` throws a `NPE` if any element is `null` or if the original collection is `null`. I can add those annotations and rename as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/StreamUtils.java
david-mackessy: I can rename this. I haven't heard that convention from anyone yet about assuming any collection is not null. That's a very big assumption but happy to go along with what we all do. I have seen that we usually instantiate an empty collection when declaring it as a class property, is that the main/only way we base our assumption for this?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
david-mackessy: do you mean create all of those objects each as a readable json string & post them to their relevant endpoints? It's much easier to work with the Java API than trying to figure out the valid JSON needed for each of those objects.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/StreamUtils.java
jbee: This differs from case to case. Also there is much code not up to speed with this. Just remember this for any changes or new code to code so that collections are never null unless it is a special case e.g. results from third party code and code that handles such cases to adapt to our assumptions. We also have the convention to assume that any collection received may be immutable. Again this is not true for private methods that know their caller but a public method receiving a collection should assume it cannot be changed.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
jbee: Yes, that is what I mean and yes it is easier to use the Java API. It is up to you what feels best suited but in general we try to keep controller level tests just using the REST API if this does not make things overly complicated. As I said this makes it much easier to reply and vary a test manually.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramController.java
```diff
@@ -125,6 +138,28 @@ public ResponseEntity<MetadataExportParams> getProgramWithDependencies( @PathVar
...
+    {
+        Program programCopy = null;
+        try
+        {
+            programCopy = copyService.copyProgramFromUid( uid, copyOptions );
+        }
+        catch ( DataIntegrityViolationException dive )
+        {
+            throw new ConflictException(
+                Objects.requireNonNull( dive.getRootCause() ).getMessage() );
```
david-mackessy, line 157: With the use of 'requireNonNull' here it provides a nicer, more detailed message if an error occurs. e.g. when the Program name is too long during a copy, the error message would look like this `value too long for type character varying(230)`
Without the 'requireNonNull', we can't get the root cause and the error would look like this e.g. `could not execute statement; SQL [n/a]; nested exception is org.hibernate.exception.DataException: could not execute statement`

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
david-mackessy: The setup has been updated to use the endpoint calls for clearer visibility of what's being done.
I found that using the Java API & saving using the manager was much simpler & more reliable with what data was in the DB too for the assertions. Finding out the sequence of API calls needed was very time-consuming, maybe that's just because Program has a deeply nested structure with differing types of relationships

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/Program.hbm.xml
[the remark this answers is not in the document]
david-mackessy: no Hibernate changes. Only formatting.

vietnguyen (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/webapi/json/domain/JsonProgramStage.java
```diff
@@ -0,0 +1,68 @@
...
+package org.hisp.dhis.webapi.json.domain;
+
+import org.hisp.dhis.jsontree.JsonArray;
+import org.hisp.dhis.jsontree.JsonList;
+import org.hisp.dhis.jsontree.JsonObject;
+
+/**
+ * @author David Mackessy
+ */
+public interface JsonProgramStage extends JsonObject
```
jbee, line 37: There is a `JsonIdentifiableObject` you can extend to not have to do all of the base methods

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramController.java
david-mackessy: the `dive` exception variable can throw a NPE from the call to `getRootCause()`. Sonar will not allow a call to `getRootCause()` without the `requireNonNull`.
So we can either keep as is with the more detailed message or remove the `requireNonNull` and have the message which does not give any idea why the error occurred.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
```diff
@@ -904,4 +913,86 @@ public void setAccessLevel( AccessLevel accessLevel )
...
+            .toList();
+    }
+
+    public static Set<ProgramIndicator> copyProgramIndicators( Program copy,
+        Set<ProgramIndicator> original )
+    {
+        return StreamUtils.streamOf( original ).filter( Objects::nonNull )
+            .map( programAttr -> ProgramIndicator.copyOf( programAttr, copy ) )
+            .collect( toSet() );
+    }
```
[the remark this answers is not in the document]
david-mackessy: This looks good. I'll update. When you say base class, do you mean `Program` or `BaseIdentifiableObject` or something else?
They could be put in a `CopyUtils` class?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramController.java
jbee: In other words the `getRootCause()` can return `null`? We should not express the expectation it is never null when it can be. That is confusing. If it can be null just assign it to a local variable and check if it is and if so generate a message from the `dive` exception, e.g. use its `getMessage()` in case of `null`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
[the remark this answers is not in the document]
jbee: `BaseIdentifiableObject` should work fine.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/StreamUtils.java
```diff
@@ -0,0 +1,65 @@
...
+     * NullPointExceptions.
+     *
+     * @param collection of any type
+     * @return a stream of elements (of Type T) from the collection if the
+     *         collection is not null or, return an empty stream if the
+     *         collection is null.
+     * @param <T> is the Type of element
+     */
+    @Nonnull
+    public static <T> Stream<T> streamOf( Collection<T> collection )
```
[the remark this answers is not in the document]
david-mackessy: yep, this class can be deleted now.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleVariable.java
```diff
@@ -239,4 +241,53 @@ public void setValueType( ValueType valueType )
...
+     *        an original and a copy.
+     */
+    private void setProgramStageFromMapping( ProgramRuleVariable copy, ProgramRuleVariable original,
+        Map<String, Program.ProgramStageTuple> stageMappings )
+    {
+        ProgramStage originalStage = original.getProgramStage();
+        if ( originalStage != null && stageMappings.containsKey( original.getProgramStage().getUid() ) )
+        {
+            ProgramStage copyStage = stageMappings.get( originalStage.getUid() ).copy();
+            copy.setProgramStage( copyStage );
```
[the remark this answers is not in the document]
david-mackessy: I've actually already updated this which complies with your comment. Next commit will have it.

jbee (approved):
Looks great!
I think it is time to get this merged. Any further improvements can be done in separate PR(s).

PR #14124, opened 2023-06-02
fix: always flush/close batch handlers in DVS import [DHIS2-15362]

jbee:
To ensure the batch handlers for data values and data value audits are always flushed they are created outside the main try-catch construct and flushed both in the success path and the exception path. This can theoretically cause `flush()` being called twice in case the exception is thrown after the flush in the success path but before the return. This has no further consequence as calling flush multiple times will be considered as a NOOP. 

To allow following the code better the main notifier start/finish calls where also pulled to the main try-catch. 

In contrast to before the the batch handler for data value audit is always created but `init()` is only called when audit is enabled (init here means establishing the connection/start the TX). Calling `flush()` on an not yet open connection again is a NOOP.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
@@ -114,6 +114,7 @@ import org.hisp.dhis.user.CurrentUserService;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.util.DateUtils;
 import org.hisp.dhis.util.ObjectUtils;
+import org.hisp.quick.BatchHandler;
 import org.hisp.quick.BatchHandlerFactory;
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -626,14 +627,31 @@ public ImportSummary importDataValueSetPdf( InputStream in, ImportOptions option
...
+                .addJobSummary( id, notificationLevel, summary, ImportSummary.class );
+
+            return summary;
         }
         catch ( Exception ex )
         {
+            dvBatch.flush();
+            dvaBatch.flush();
             log.error( DebugUtils.getStackTrace( ex ) );
-            notifier.clear( id ).notify( id, ERROR, "Process failed: " + ex.getMessage(), true );
```
jbee, line 636: Note that I removed the `clear( id )` part here as I think this wasn't intentionally clearing the messages added so far in case of an error.

david-mackessy (approved):

vietnguyen (approved):

PR #14140, opened 2023-06-05
fix: always flush/close batch handlers in DVS import [DHIS2-15362] (2.40)

jbee:
Cherry-pick backport of #14124


jason-p-pickering (approved):
I did not test explicitly against 2.40, but based on my results in 2.39, I believe that this solves the issue in 2.40 as well.

PR #14141, opened 2023-06-05
fix: always flush/close batch handlers in DVS import [DHIS2-15362] (2.39)

jbee:
Cherry-pick backport of #14124

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
@@ -115,6 +115,7 @@ import org.hisp.dhis.user.CurrentUserService;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.util.DateUtils;
 import org.hisp.dhis.util.ObjectUtils;
+import org.hisp.quick.BatchHandler;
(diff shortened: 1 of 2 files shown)


jason-p-pickering:
I tested this against a copy of a production database which was affected by this issue. A locust test with 30 clients was created to send a small payload (3 values) to the dataValueSet endpoint. In this case, the user did not have explicit write access to the dataset, which will result in an aborted import. 

The payload which was used is below: 

```
        {
            "dataSet": "K4Lx3AcgpsF",
            "completeDate": "2023-05-31",
            "period": "2023W21",
            "orgUnit": "XGvMbsXhqY9",
            "dataValues": [
                {
                    "dataElement": "CDc0ANqWjMC",
                    "categoryOptionCombo": "jJ8GtngnAdM",
                    "value": "1"
                },
                {
                    "dataElement": "db6j50oRry4",
                    "categoryOptionCombo": "jJ8GtngnAdM",
                    "value": "2"
                },
                {
                    "dataElement": "whVe8mmkU85",
                    "categoryOptionCombo": "jJ8GtngnAdM",
                    "value": "3"
                }
            ]
        }
```

This branch was built and tested using the locust test (left side of the graph) and against a stock 2.39 build (on the right side of the graph). The graph shows total database connections. 

image


As can be seen from the left side of the graph, the total database connections increase slightly while the Locust test is running, but remain stable. On the right side of the graph (stock 2.39) the database connections quickly reach the maximum (80) and at that point, the server locks up. 

So, based on my testing, this is a pass. Thanks for the fix @jbee !

david-mackessy (approved):

PR #14144, opened 2023-06-05
fix: always flush/close batch handlers in DVS import [DHIS2-15362]

jbee:
Cherry-pick backport of  #14124

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
@@ -115,6 +115,7 @@ import org.hisp.dhis.user.CurrentUserService;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.util.DateUtils;
 import org.hisp.dhis.util.ObjectUtils;
+import org.hisp.quick.BatchHandler;
 import org.hisp.quick.BatchHandlerFactory;
 import org.hisp.staxwax.factory.XMLFactory;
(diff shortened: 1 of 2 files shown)


jason-p-pickering (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -629,14 +630,33 @@ public ImportSummary importDataValueSetPdf( InputStream in, ImportOptions option
...
+            NotificationLevel notificationLevel = options.getNotificationLevel( INFO );
+            notifier.notify( id, notificationLevel, "Import done", true )
+                .addJobSummary( id, notificationLevel, summary, ImportSummary.class );
+
+            return summary;
         }
         catch ( Exception ex )
         {
+            dvBatch.flush();
+            dvaBatch.flush();
```
david-mackessy, line 657: Would it be worth logging that the batch handlers are being flushed? (in the try and/or catch)
This would help when reading logs and not having to inspect the code.
Exposing app behaviour in the logs usually helps understand what the app is doing.

jbee: Yes, in hindsight I learned that some more logging would be good here. As I said in the meeting the logging situation of the import needs general improvement. I make a new ticket for specifically working on the logging.

jbee: Created https://dhis2.atlassian.net/browse/DHIS2-15396

PR #14166, opened 2023-06-07
fix: use utf-8 encoding for email sender service

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14901

- The `EmailMessageSender` use `org.apache.commons.mail.HtmlEmail` which takes in a plain text message and a html message. This PR encode both messages with `StandardCharsets.UTF_8`.

### Test
- Use endpoint `api/email/notification?recipients=test@test.com&message={message}subject={subject}` for manually testing. Also need a SMTP server,  I use [this free tool](https://www.wpoven.com/tools/free-smtp-server-for-testing).
- Test steps as below:

1. Set up email server configurations in System Settings > Email.
2. Send a POST request to endpoint `api/email/notification?recipients=test@test.com&message={message}subject={subject}`. The message value must contain Arabic characters, or any other language.
3. Server return 200 with message "Email sent".
4. Check the email sent by DHIS2 in the test SMTP server inbox and verify if the email content has "???" characters.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/message/EmailMessageSender.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/message/EmailMessageSender.java
@@ -29,6 +29,7 @@ package org.hisp.dhis.message;
 
 import static java.util.Collections.singleton;
 
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -291,6 +292,7 @@ public class EmailMessageSender
         throws EmailException
     {
         HtmlEmail email = new HtmlEmail();
+        email.setCharset( StandardCharsets.UTF_8.toString() );
         email.setHostName( hostName );
         email.setFrom( sender, getEmailName() );
         email.setSmtpPort( port );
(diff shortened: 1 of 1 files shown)


mortenoh (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/message/EmailMessageSender.java
```diff
@@ -306,12 +308,17 @@ private HtmlEmail getHtmlEmail( String hostName, int port, String username, Stri
...
+        String content = sender == null ? text
             : (text + LB + LB + sender.getName() + LB
-                + (sender.getOrganisationUnitsName() != null ? (sender.getOrganisationUnitsName() + LB)
-                    : StringUtils.EMPTY)
-                + (sender.getEmail() != null ? (sender.getEmail() + LB) : StringUtils.EMPTY)
-                + (sender.getPhoneNumber() != null ? (sender.getPhoneNumber() + LB) : StringUtils.EMPTY));
+                + getNullSafe( sender.getOrganisationUnitsName() )
+                + getNullSafe( sender.getEmail() )
+                + getNullSafe( sender.getPhoneNumber() ));
+        return new String( content.getBytes( StandardCharsets.UTF_8 ), StandardCharsets.UTF_8 );
```
jbee, line 316: I am not quite following the idea behind this recreation of a `String`. First it encodes the `content` in UTF-8 to then recreate a string instance from the UTF-8 bytes. That should give us the original `content` value. Is this some form of indirect validation?

david-mackessy:
I'm guessing this is fixing being able to see certain characters? Adding a test to confirm that's now possible would be beneficial.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/message/EmailMessageSender.java
vietnguyen: The idea is to convert a non-unicode text to Unicode. 
If the original text is already in UTF-8 then it should return same value.

vietnguyen:
> I'm guessing this is fixing being able to see certain characters? Adding a test to confirm that's now possible would be beneficial.

Yes,  i thought about that. But this service use `org.apache.commons.mail.HtmlEmail` which doesn't expose a `getContent()` method. It takes in a string and modifies it base on the email settings, charset encoding, etc. 
Then we can only call a `email.send()` method and get a status back. There is no existing test for this `EmailMessageSender` so I think it's a bit out-of-scope for this fix. 
I believe best way to test this is to manually test the sent email.

david-mackessy (approved):

PR #14176, opened 2023-06-08
fix: use read-write JDBC template for SQL views [DHIS2-15108]

jbee:
Same fix as https://github.com/dhis2/dhis2-core/commit/1a7dbfba5e2eea97827aa9e739979c89dbc09cd4 but for 2.40.0


amcgee:
Thanks @jbee !  @Philip-Larsen-Donnelly I think we wanted this on a custom branch based on `patch/2.40.0` instead of on the branch itself..?

PR #14180, opened 2023-06-08
chore: Harmonize event controller [DHIS2-1570]

muilpp:
Goal of this PR is to create the `EventOperationParams` object for the service layer.

Detailed changes:
    Create EventOperationParams object that is meant to contain parameters for the service layer, it is created by converting RequestParams and it will be converted to EventSearchParams.
    Separate validations by layer. Move to service layer validations that are accessing DB. (ie. Checking if some metadata exists)
    
Ticket:  https://dhis2.atlassian.net/browse/TECH-1570


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -0,0 +1,321 @@
...
+    {
+        if ( attributeOptionCombo != null && !user.isSuper()
+            && !aclService.canDataRead( user, attributeOptionCombo ) )
+        {
+            throw new ForbiddenException(
+                "User has no access to attribute category option combo: " + attributeOptionCombo.getUid() );
+        }
+    }
+
+    private void validateOperationParams( EventOperationParams params, User user, Program program )
```
teleivo, line 241: To be consistent with the other methods this should rather be called `validateOrgUnitSelectionMode` as this is what its concerned with.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -0,0 +1,321 @@
...
+        {
+            String violation = getOuModeViolation( params, user, program );
+            if ( violation != null )
+            {
+                throw new BadRequestException( violation );
+            }
+        }
+    }
+
+    private String getOuModeViolation( EventOperationParams params, User user, Program program )
```
teleivo, line 254:
```suggestion
    private String getOrgUnitModeViolation( EventOperationParams params, User user, Program program )
```

we should consistently use `orgUnit` everywhere 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventSearchParams.java
```diff
@@ -319,7 +317,7 @@ public List<QueryItem> getDataElementsAndFilters()
         return items;
     }
 
-    public EventSearchParams addDataElements( List<QueryItem> des )
+    public EventSearchParams addDataElements( Set<QueryItem> des )
```
teleivo, line 320: Does order matter? If so we should not make this change.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/CategoryOptionComboService.java
```diff
@@ -49,7 +49,7 @@
  * @author Lars Helge Overland
  */
 @Component
-class CategoryOptionComboService
+public class CategoryOptionComboService
```
teleivo, line 52: Does it need to be public?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -0,0 +1,321 @@
...
+
+    //For now this maps to EventSearchParams. We should create a new EventQueryParams class that should be used in the persistence layer
+    @Transactional( readOnly = true )
+    public EventSearchParams map( EventOperationParams operationParams )
+        throws BadRequestException,
+        ForbiddenException
+    {
+        User user = currentUserService.getCurrentUser();
+
+        Program program = validateProgram( operationParams.getProgramUid() );
```
teleivo, line 89: I have missed this in reviewing @enricocolasante changes to the enrollment controller. By moving these validations from the service into the mapper we cannot use them when validating operations for single entities. `/tracker/trackedEntites/{te}?program={uid}` is such a case.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -0,0 +1,321 @@
...
+    private final AclService aclService;
+
+    private final CategoryOptionComboService categoryOptionComboService;
+
+    private final TrackerAccessManager trackerAccessManager;
+
+    private final CurrentUserService currentUserService;
+
+    //For now this maps to EventSearchParams. We should create a new EventQueryParams class that should be used in the persistence layer
+    @Transactional( readOnly = true )
```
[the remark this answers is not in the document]
muilpp: I would assume the same. I've checked every service and all the methods running a query against the database do it in a transaction except `CurrentUserService.getCurrentUser`, is there a reason for that?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -0,0 +1,321 @@
...
+
+    private final AclService aclService;
+
+    private final CategoryOptionComboService categoryOptionComboService;
+
+    private final TrackerAccessManager trackerAccessManager;
+
+    private final CurrentUserService currentUserService;
+
+    //For now this maps to EventSearchParams. We should create a new EventQueryParams class that should be used in the persistence layer
```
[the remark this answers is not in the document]
muilpp: This is out of the scope of this PR, but I guess we'll tackle this naming issue when we work on the persistence layer

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: I can't find where right now, but I remember reading somewhere that single entities requests are out of the scope of this refactor, are they not?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventSearchParams.java
muilpp: The data elements is now actually a `LinkedHashSet,` so it should maintain the order. In the last commit, I removed the setter to not override the value, so now using the method `addDataElements`, where we just add elements to the existing collection should be enough.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -108,9 +112,18 @@ class EventExporterTest extends TrackerTest
 
     private Program program;
 
-    final Function<EventSearchParams, List<String>> eventsFunction = ( params ) -> eventService.getEvents( params )
-        .getEvents()
-        .stream().map( Event::getUid ).collect( Collectors.toList() );
+    final Function<EventOperationParams, List<String>> eventsFunction = ( params ) -> {
+        try
+        {
+            return eventService.getEvents( params )
```
enricocolasante, line 118: Maybe it is not part of this PR but this try/catch here seems dangerous.
We should re-evaluate this test when we tackle the service part.
Maybe add a comment here just saying to double check this later.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -324,40 +311,31 @@ void testExportEventsWhenFilteringByDataElementsWithProgramTypeFilter(
...
+            .build();
 
         List<String> events = eventFunction.apply( params );
 
         assertContainsOnly( List.of( "pTzf9KYMk72" ), events );
     }
 
-    @ParameterizedTest
+    //This test is commented because it doesn't seem like we don't offer the feature to filter by multiple values with the like operator
+    //@ParameterizedTest
```
enricocolasante, line 329: We should sort this out before the PR is merged, anyway is always better to @Disable the test and put the explanation in the description

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -901,15 +842,17 @@ void testEnrollmentFilterAttributesWithMultipleFiltersOnDifferentAttributes()
         assertContainsOnly( List.of( "dUE514NMOlo" ), trackedEntities );
     }
 
-    @Test
+    //@Test
+    //This test is commented because it doesn't seem like we don't offer the feature to filter by multiple values with the like operator
```
enricocolasante, line 846: Same here @Disable

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -345,19 +329,15 @@ void testExportEventsWhenFilteringByDataElementsEqual( Function<EventSearchParam
...
-
         DataElement datael00001 = dataElement( "DATAEL00001" );
 
-        params.setDataElements( Set.of(
-            new QueryItem( datael00001, QueryOperator.IN, "value00001;value00002", datael00001.getValueType(),
-                null,
-                null ) ) );
+        EventOperationParams params = EventOperationParams.builder().orgUnitUid( orgUnit.getUid() )
+            .enrollments( Set.of( "nxP7UnKhomJ", "TvctPPhpD8z" ) ).programStageUid( programStage.getUid() )
+            .filters( datael00001.getUid() + ":like:%value00001%;%value00002%" )
```
enricocolasante, line 339:
```suggestion
            .filters( datael00001.getUid() + ":in:value00001;value00002" )
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -902,14 +847,16 @@ void testEnrollmentFilterAttributesWithMultipleFiltersOnDifferentAttributes()
...
     {
-        EventSearchParams params = new EventSearchParams();
-        params.setOrgUnit( orgUnit );
-
         QueryItem item = queryItem( "toUpdate000", QueryOperator.LIKE, "day" );
         item.addFilter( new QueryFilter( QueryOperator.LIKE, "in" ) );
-        params.addFilterAttributes( item );
+
+        EventOperationParams params = EventOperationParams.builder().orgUnitUid( orgUnit.getUid() )
+            .filterAttributes( "toUpdate000:like:%day%:like:%in%" ).build();
```
enricocolasante, line 859:
```suggestion
            .filterAttributes( "toUpdate000:like:day:like:in" ).build();
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -902,14 +847,16 @@ void testEnrollmentFilterAttributesWithMultipleFiltersOnDifferentAttributes()
     }
 
     @Test
+    @Disabled( "This test is disabled because it doesn't look like we offer the feature to filter by multiple values with the like operator" )
```
enricocolasante, line 850:
```suggestion
   
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -345,19 +329,15 @@ void testExportEventsWhenFilteringByDataElementsEqual( Function<EventSearchParam
 
     @ParameterizedTest
     @MethodSource( "getEventsFunctions" )
-    void testExportEventsWhenFilteringByDataElementsIn( Function<EventSearchParams, List<String>> eventFunction )
+    @Disabled( "This test is disabled because it doesn't look like we offer the feature to filter by multiple values with the like operator" )
```
enricocolasante, line 332:
```suggestion
   
```

enricocolasante (commented):
Disabled tests can now be fixed and they should work

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
lucaCambi77: `getEventsFunctions()` probably was in place to parametrize tests with multiple functions. I think you can get rid of it and use and helper method.

lucaCambi77:
@muilpp overall looks good to me although if we are going to do something similar in the next future, let's break it down into smaller issues otherwise it is difficult to review. I am not sure about the name `EventOperationParams` or better what is the meaning of `Operation`?. I see this is later converted in the EventSearchParams so i am not really convinced by using `Operation`.  In any case, now I see in `EventOperationParams` there are unused methods we can clean?  Also, I think we can get rid of `getEventsFunctions()`  in `EventExporterTest` and use a helper method. In the controller `EventRequestParamsMapper` variable could be `requestParamMapper`? Also, we can think about removing the prefix Event in class names where applicable?

muilpp:
@lucaCambi77 the naming of the params and the mapper is a new convention we want to follow to harmonize the tracker team controllers. The idea is to have for each layer (web, service and data), a class storing the value of the parameters and the names of each in case of events, would be `EventRequestParams`, `EventOperationParams` and `EventQueryParams`. It is already working for `EnrollmentExportController`. If I'm not mistaken, this was discussed in the backend meeting not too long ago.

lucaCambi77 (approved):

PR #14184, opened 2023-06-08
docs: Document program and includeAllAttributes params in TE Controller [TECH-1568]

enricocolasante:
In this PR we are just documenting how the API is working today. We may need to check again if this behavior makes sense.

Attributes are linked to a TE but they are divided between the ones that are linked directly to a TE and the ones that are specific to a `Program`.

`IncludeAllAttributes` param will show all the attributes TE and Program ones, but only if `program` is not specified.
If `program` is specified all the TE attributes and the Program ones linked to the specified Program are shown.
In the case that a `program` is specified then `includeAllAttributes` is just ignored.

**Open issues**
- `IncludeAllAttributes` seems to not work fine when attributes are specified in `fields`
- `program` is filtering the entity in the response and it is manipulating the fields in the response at the same time.
- Should we have a validation for `program` and `includeAllAttributes` used together or should we make them work independently?

### DIFF
--- a/dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
+++ b/dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
@@ -10,6 +10,9 @@ Get a tracked entity with given UID.
 
 ### `getTrackedEntityByUid.parameter.program`
 
+Get tracked entities enrolled in the specified `program`.
+When a `program` is specified, all the Tracked Entity attributes and the Program specific ones are included.
+
 ## Common for all endpoints
 
 ### `*.parameter.TrackedEntityRequestParams.query`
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
```diff
@@ -10,6 +10,9 @@ Get a tracked entity with given UID.
 
 ### `getTrackedEntityByUid.parameter.program`
 
+Get tracked entities enrolled in the specified `program`.
+When a `program` is specified, all the Tracked Entity attributes and the Program specific ones are included.
```
teleivo, line 14: I think we should be more explicit than "are included". The question I have is where are they included? This leads to the one thing I would like to know: should the TEI only have tracked entity type attributes in the response? while the enrollment should only have program attribute values? (@stian-sandvold @Bekkalizer)

The implementation is really concerning to me:
`Enrollment.trackedEntity.trackedEntityAttributeValues` should only contain the attribute values for program attributes AFAIK. On the other hand the `Enrollment.trackedEntity` is the enrolled TEI. So the TEI might have its own attribute values for tracked entity type attributes. So by setting the `Enrollment.trackedEntity.trackedEntityAttributeValues` to a subset we risk of removing the attribute values of the TEI (when hibernate is used) 😬 To me there seems to be a high risk that TEI and enrollment will show the wrong attributes.

dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
```diff
@@ -10,6 +10,9 @@ Get a tracked entity with given UID.
 
 ### `getTrackedEntityByUid.parameter.program`
```
teleivo, line 11: You will need to duplicate this text in `### *.parameter.TrackedEntityRequestParams.program` as its also a param for the collections endpoint.

teleivo (commented):
Thank you for investigating this 😅 

I would stick to English casing so `Program` would only be written this way at the beginning of a sentence.

PR #14193, opened 2023-06-09
feat: Rename programstageinstance DB table to event [TECH-1542]

lucaCambi77:
https://dhis2.atlassian.net/browse/TECH-1542

### Overall approach
Always starts from DB and then applies changes to the application code

### **First Step:**
- Rename `programstageinstance` to `event` with SQL migration script. This is nicely done by Postgres which keeps foreign keys and indexes.
- Rename `programstageinstance` to `event` in XML hibernate mapping XML.

Now the application starts (also tests using the database). With https://github.com/dhis2/dhis2-core/pull/14167 we introduced hibernate DDL validation, so we can reliably say that we are aligned with the database although this validation is not probably always 100% working so we need to double-check. Foreign keys are mapped with the key name and not with the `programstageinstance` table name.

Now `programstageinstance` table is renamed to `event`. We expect to see failing raw SQL statements (JDBC statements aka native queries). 

### **Next Steps: Iteration from DB to application code**

Iterate over changes from DB to application code.

**In these steps, we'll work on code related to each team and also the common parts:**
We'll iterate similarly to the first step, doing iteration from DB to application code.

This will involve:
- rename `programstageinstance` to `event` for tables and columns  with the prefix ?
- rename `programstageinstanceid` to `eventid` 

Also here, for DB, it seems this step is nicely done by Postgres which renames also fk and indexes references. So we might also want to rename tables and columns in the db that are prefixing `programstageinstance`?

For application code, this can be found in the following:

- dhis-service-administration
- dhis-service-analytics
- dhis-service-core
- dhis-service-dxf2 (old_tracker)
- dhis-service-tracker

- dhis-support-hibernate (HibernateDbmsManager class)

### Finalize

Finalizing with tasks not strictly mandatory we can group like:

- rename `psi` and  `programstageinstance` java variables and in SQL 
This has probably already been done in other tasks. There are some psi alias left in queries but I don't think it is worth renaming it as prone to error?
I'd say to avoid renaming the constraints name and sequence name. Although `programstageinstance` will remain in the name, renaming it is not bringing much value?

- Renaming the migration to a more sensible progressive number?

Reference for queries to find more about the `programstageinstance` in tables and columns in notes:

Notes:
https://docs.google.com/document/d/1fBFP1dMegSzobPCocZw2TfJG0l9rGi5wGXq2sy5Rr3E/edit#


enricocolasante (approved):
Nice job!
Not in this PR, but I would think about renaming `psi` occurrences in tracker module at some point

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/Event.hbm.xml
```diff
@@ -51,8 +51,8 @@
     <many-to-one name="organisationUnit" class="org.hisp.dhis.organisationunit.OrganisationUnit" column="organisationunitid"
       foreign-key="fk_programstageinstance_organisationunitid" index="programstageinstance_organisationunitid" />
 
-    <list name="messageConversations" table="programstageinstance_messageconversation">
-      <key column="programstageinstanceid" />
+    <list name="messageConversations" table="event_messageconversation">
```
teleivo, line 54: Is `event_messageconversation` going against our convention of all lowercase with no separators like `eventcomments`? If so should we remove the `_` as we already rename the table?

lucaCambi77: this table and also `programinstance_messageconversation` I understand we'll be removed soon because no longer in use so we can leave it like this

maikelarabori (approved):
Locally, it seems that table export, event reports, line listing and data visualizer are working fine. If by any chance, we find anything missing we can rename and fix it later.

Cheers

PR #14255, opened 2023-06-16
fix: sortOrder gaps in Category.categorycombos [DHIS2-15351]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-15351
### Summary
- After importing the payload attached in jira, User get `null` values in `categoryCombos` array from `api/categories/{id}` as below.
```json
"categoryCombos": [
    null,
    {
      "id": "TIAbMD7ETV6"
    },
    null,
    {
      "id": "FEs7WKVk9B1"
    }
  ],

```

### Issue
- The issue caused by `sort_order` gaps in table `categorycombos_categories`. Hibernate will inject `null` item to a List if there are gaps in `list-index` or `sort_order` column.
- Another issue is that the table `categorycombos_categories`  has only one `sort_order` column but both `CategoryCombo.categories` and `Category.categoryCombos` collections are mapped with this index column. This causes the list-index to be updated incorrectly. 
- In Maintenance app > Category page, we don't show `CategoryCombo` list,  so I think we can **remove** the `list-index` from `categoryCombos` property in `Category.hbm.xml` file. This will **solve** the reported issue.
- In hibernate xml mapping, a `<list>` tag  can be replaced by`<bag>` which is basically a list without `list-index`. It also has `order-by` attribute which we can use later if needed.

### Test
- Added unit test
- Manual test can be done using `categories.json` file attached in jira. Just import the file then send GET request to any `Category` in the payload, verify if there is `null` in `categoryCombos` array.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/category/hibernate/Category.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/category/hibernate/Category.hbm.xml
@@ -39,13 +39,12 @@
                     foreign-key="fk_category_categoryoptionid"/>
     </list>
 
-    <list name="categoryCombos" table="categorycombos_categories" inverse="true">
+    <bag name="categoryCombos" table="categorycombos_categories" inverse="true">
       <cache usage="read-write" />
       <key column="categoryid" foreign-key="fk_categorycombo_categoryid" />
-      <list-index column="sort_order" base="1" />
       <many-to-many class="org.hisp.dhis.category.CategoryCombo" column="categorycomboid"
         foreign-key="fk_categorycombos_categories_categorycomboid" />
-    </list>
+    </bag>
 
     <property name="dataDimension" column="datadimension" not-null="true" />
 
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MetadataImportExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MetadataImportExportControllerTest.java
@@ -30,6 +30,7 @@ package org.hisp.dhis.webapi.controller;
 import static org.hisp.dhis.web.WebClient.Body;
 import static org.hisp.dhis.web.WebClient.ContentType;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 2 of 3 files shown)


jbee (approved):
I like this solution. We should think about if we pro-actively apply something similar to other places where we currently rely on the `list_index`.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/category/hibernate/Category.hbm.xml
```diff
@@ -39,13 +39,12 @@
                     foreign-key="fk_category_categoryoptionid"/>
     </list>
 
-    <list name="categoryCombos" table="categorycombos_categories" inverse="true">
+    <bag name="categoryCombos" table="categorycombos_categories" inverse="true">
```
larshelge, line 42: Just curious, are we sure we want to use the more exotic `bag` collection type? Do we use it other places? Feels a bit arbitrary if we only use it here. Are we handling sort order gaps consistently other places, as this might happen to all lists I assume?

vietnguyen: This is a _hibernate mapping error_ of `Category` and `CategoryCombo` association. It's different from the `OptionSet.opion` or any other `<list>` sortOrder issue that i have fixed before. 
In details:
- We have a many-to-many association between `Category` and `CategoryCombo` and we wanted `list-index` to be implemented on both sides. Means that  both `Category.categoryCombos`  and `CategoryCombo.categories` have `list-index`. 
- However the issue is that in table `categorycombos_categories` **there is only one `sort_oder` column used by two `<list-index>`**. Hence, whatever sortOrders are saved correctly for one side of this association should be incorrect in the other side.
- Now, we can fix this by adding one more `sortOrder` column to that table. But if you have a look at Maintenance App > Category edit page, there is no `CategoryCombo` list. Which means the `Category.categoryCombos` sortOrder is not in used. I guess it has been used by a Struts App long time ago. Therefore instead of fixing a feature that is not in used, I think it's better to remove it.
- Why this issue only happen from 2.39 ? The reason is from 2.39 we moved from our custom `NodeDeserialiser` to `JacksonDeserializer`.  The `NodeDeserialiser` may have excluded the `null` items from the collection when converting the list of Objects to `org.hisp.dhis.node.Node`.
- After this, we can create a jira issue and go over all of remaining `<list>` mapping to verify if the `sortOrder` is in used. If not we can convert them to `<bag>` to avoid  unintended issues. 
- To expand the issue further, currently there are two ways of mapping `list-index` in our codes.
1. **Implicit**: there is no `sortOrder` property in Java Object model. So the order values are implicitly generated and saved by hibernate base on the Java List indexes.
2. **Explicit**: there is a `sortOrder` property in Java Object model. User manually set the order value to the object and hibernate will save it to database. 
Until now, approach (1) has not caused any error beside the issue fixed by this PR, but as mentioned, it's a many-to-many mapping error. While approach (2) has caused many issues for `OptionSet.option`.

PR #14256, opened 2023-06-16
feat: remove Debezium based cache invalidation

netroms:
# Summary

* Removes Debezium based cache invalidation strategy from the code base, Redis based strategy is now the default.
* Refactors some of the Redis based strategy code.
* Adds a testable message publisher in order to be able to test the Hibernate event listeners and the message publisher.
* Adds mock tests to the cache invalidation event listener.
* Adds the "dhis-support-cache-invalidation" to the "dhis-web-api", which is needed in order for the cache invalidation module to be included in the server.


dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/CacheInvalidationEnabledCondition.java
```diff
@@ -35,22 +36,27 @@
...
+            return true;
+        }
+
+        if ( SystemUtils.isTestRun( context.getEnvironment().getActiveProfiles() ) )
         {
             return false;
         }
 
-        return getConfiguration().isEnabled( ConfigurationKey.DEBEZIUM_ENABLED );
+        return getConfiguration().isEnabled( ConfigurationKey.REDIS_CACHE_INVALIDATION_ENABLED );
```
jbee, line 54: Do we have to think about transition for installations that already used DEBEZIUM?

david-mackessy (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/SystemUtils.java
```diff
@@ -51,6 +51,11 @@ public static boolean isEnableCacheInTest( String[] profiles )
         return Arrays.asList( profiles ).contains( "cache-test" );
     }
 
+    public static boolean isCacheInvalidationInTest( String[] profiles )
+    {
+        return Arrays.asList( profiles ).contains( "cache-invalidation-test" );
```
jbee, line 56: Usually `Arrays.asList` can not be replaced by `List.of` unless you worry that the profiles array might include a `null` value.

jbee (approved):
I am not following all of it on a higher level but on the code level it looks good.

PR #14281, opened 2023-06-19
fix: test that OpenAPI JSON document is semantically valid [DHIS2-15432]

jbee:
The generated OpenAPI document could become semantically invalid. This means it is valid JSON but values in the tree are not valid with regards to the OpenAPI spec. Mostly this happens with paths that do not conform to the expectations.

To prevent this from happening silently the validity is checked using swagger validation. 

The fixes needed were:
* ignore any controller defined in test sources using `@OpenApi.Ignore`
* make sure paths start with a `/` ( `TeiAnalyticsController`)

### DIFF
--- a/dhis-2/dhis-test-web-api/pom.xml
+++ b/dhis-2/dhis-test-web-api/pom.xml
@@ -455,6 +455,12 @@
       <artifactId>log4j-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.swagger.parser.v3</groupId>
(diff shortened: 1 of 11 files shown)


jbee:
@gnespolino Added you since you might know the controller were I changed the paths definitions and if that possibly is covered by tests.

gnespolino:
> @gnespolino Added you since you might know the controller were I changed the paths definitions and if that possibly is covered by tests.

I'll check it, thank you!

larshelge (approved):

PR #14302, opened 2023-06-20
feat: replace ANTLR parser for validation rule with PEG parser

jbee:
WIP


jbee:
This is too old for sure and has to be restarted at some point

PR #14315, opened 2023-06-21
feat: add support for programId in dimension endpoint [DHIS2-15427]

gnespolino:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/EventAnalyticsController.java
```diff
@@ -393,22 +394,33 @@ public void getQueryHtmlCss(
...
             EventAnalyticsPrefixStrategy.of( programStageId ) );
 
         return dimensionFilteringAndPagingService.pageAndFilter( dimResponse, dimensionsCriteria, fields );
     }
 
+    private void validateRequest( String programId, String programStageId )
+    {
+        if ( StringUtils.isBlank( programId ) && StringUtils.isBlank( programStageId ) )
+        {
+            throw new IllegalArgumentException( "Either programId or programStageId must be specified" );
```
d-bernat, line 420: The exception message does not follow the logic.  I think both of them must not be blank.

d-bernat (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/EventAnalyticsController.java
gnespolino: No, in this case the caller can pass program, program stage or both.
The invocation is unbelievable when both are blank

maikelarabori (commented):
Hey @gnespolino, left two comments that would be nice to get reviewed 👍 Other than that it looks good.

maikelarabori (approved):

PR #14318, opened 2023-06-21
feat: Program Copy - Add permission check [DHIS2-15279]

david-mackessy:
# Summary
- Add permission check to the `Program` Copy endpoint
`AclService` check added to check if user has write access to the `Program` being copied.
- It seems that the required `AuthorityType`s to copy a `Program` are:
  - F_PROGRAM_PUBLIC_ADD
  - F_PROGRAM_INDICATOR_PUBLIC_ADD

Different responses can be returned depending on the access the User has:
- no read access -> NOT_FOUND response
- no write access -> FORBIDDEN response. 

This is because there is a get done first to get the Program by its UID (read access required)
- no exception is thrown in this operation. Only a null object is returned from the service. So it's hard to determine if the object is genuinely not found or if it's a permission issue.

Then when saving the write access is checked, which can throw a permission error.

# Test
- Add service test when `User` has no write access.
- Add web tests to test numerous scenarios with different auth scenarios.
- 3 Integration tests were required for testing when no public access on the `Program` but sharing has been updated to include a specific `User`. (jsonb function doesn't work with H2)
    - user with read only access
    - user with write only access
    - user with read & write access

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.webapi.controller;
 
 import static org.hisp.dhis.feedback.ErrorCode.E1005;
+import static org.hisp.dhis.web.WebClientUtils.assertStatus;
 import static org.hisp.dhis.webapi.utils.TestUtils.getMatchingGroupFromPattern;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -38,6 +39,7 @@ import java.util.stream.Collectors;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.jsontree.JsonList;
 import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
+import org.hisp.dhis.user.User;
 import org.hisp.dhis.web.HttpStatus;
 import org.hisp.dhis.web.WebClient;
 import org.hisp.dhis.webapi.DhisControllerConvenienceTest;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
```diff
@@ -253,4 +255,60 @@ void testCopyProgramWithInvalidUid()
...
+        User userWithNoAuthorities = switchToNewUser( "test1" );
+        Set<String> authorities = userWithNoAuthorities.getUserRoles().stream()
+            .flatMap( role -> role.getAuthorities().stream() )
+            .collect( Collectors.toSet() );
+        assertEquals( 0, authorities.size() );
+
+        JsonWebMessage response = POST( "/programs/%s/copy".formatted( PROGRAM_UID ) )
+            .content( HttpStatus.FORBIDDEN )
+            .as( JsonWebMessage.class );
+        System.out.println( response );
```
jbee, line 271: don't forget to remove ;)

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerTest.java
```diff
@@ -253,4 +255,60 @@ void testCopyProgramWithInvalidUid()
         assertEquals( "Program with id invalid could not be found.", response.getMessage() );
         assertEquals( E1005, response.getErrorCode() );
     }
+
+    @Test
+    void testCopyProgramWithUserWithNoAuthorities()
+    {
+        User userWithNoAuthorities = switchToNewUser( "test1" );
+        Set<String> authorities = userWithNoAuthorities.getUserRoles().stream()
```
jbee, line 263: There is `getAllAuthorities()` on `User`, but maybe your implementation is nicer than the one currently in there :D

jbee (commented):
Looks good!

larshelge (approved):

jbee (approved):

PR #14328, opened 2023-06-22
fix: Prevent NPE in 'ou' ownership export [DHIS2-15467]

maikelarabori:
This PR fixes an NPE that happens when the table `programownershiphistory` contains rows with null `organisationunitid`.

Two preventions were added: one at querying time, so we avoid rows without `organisationunitid`, and another at the code level.

A small change was also applied: in the SQL function `exists` I gave preference to `select 1`, as null is also evaluated to `true`. It's just another preventive change.


d-bernat (approved):

jimgrace (changes requested):
Excellent, thank you!

jimgrace (approved):

PR #14329, opened 2023-06-22
chore: Refactor to User class to simplify some 'get' and 'has' methods

david-mackessy:
# Summary
Small refactor for the `User` class
- simplify some utility methods on the User class
- add tests for refactored methods
- add tests for methods calling refactored methods

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.user;
 
+import static org.apache.commons.collections4.CollectionUtils.emptyIfNull;
 import static org.springframework.beans.BeanUtils.copyProperties;
 
 import java.beans.IntrospectionException;
@@ -36,12 +37,12 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -298,16 +299,9 @@ public class User
      */
     public Set<String> getAllAuthorities()
     {
-        Set<String> authorities = new HashSet<>();
-
-        for ( UserRole group : userRoles )
-        {
-            authorities.addAll( group.getAuthorities() );
-        }
-
-        authorities = Collections.unmodifiableSet( authorities );
-
-        return authorities;
+        return emptyIfNull( userRoles ).stream()
+            .flatMap( role -> emptyIfNull( role.getAuthorities() ).stream() )
+            .collect( Collectors.toUnmodifiableSet() );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -298,16 +299,9 @@ public String getUserRoleNames()
...
-
-        for ( UserRole group : userRoles )
-        {
-            authorities.addAll( group.getAuthorities() );
-        }
-
-        authorities = Collections.unmodifiableSet( authorities );
-
-        return authorities;
+        return emptyIfNull( userRoles ).stream()
```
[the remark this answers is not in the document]
jbee: I think it is good to be safe here but I would do it differently for avoiding unnecessary idle work to find out that nothing needs done which is what you did know from the start if the set is null. So I'd do it like this:
```java
return userRoles == null ? Set.of() : ...
```
Within the `filter` however I'd keep it as is just for being safe and clearly and easily stated.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -1024,27 +1003,14 @@ public boolean isAuthorized( Authorities auth )
...
-        for ( UserGroup group : groups )
-        {
-            if ( group != null && group.getManagedGroups() != null && !group.getManagedGroups().isEmpty() )
-            {
-                return true;
-            }
-        }
-
-        return false;
+        return !getManagedGroups().isEmpty();
```
jbee, line 1013: This is correct but maybe we want to avoid building that collection. I think `Stream` with `anyMatch` should work nicely.

dhis-2/dhis-api/src/test/java/org/hisp/dhis/user/UserTest.java
```diff
@@ -125,4 +126,120 @@ void getAllAuthoritiesChanged()
...
+
+    @Test
+    void testGetAllRestrictions()
+    {
+        final User user = new User();
+        UserRole userRole = new UserRole();
+        userRole.setRestrictions( Set.of( "R1", "R2" ) );
+        user.setUserRoles( Set.of( userRole ) );
+        Set<String> restrictions = user.getAllRestrictions();
+        assertEquals( 2, restrictions.size() );
```
jbee, line 194: Better to assert the actual expected set.

jbee (commented):
Looks nice. Some details to iron out.

Great you added tests as well. I am not sure how good the test coverage already was for scenarios where items get aggregated from more than one source, like multiple groups with authorities. If no such exist it would be good to add one for each method that tests that scenario as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -1024,27 +1006,15 @@ public boolean isAuthorized( Authorities auth )
...
-        for ( UserGroup group : groups )
-        {
-            if ( group != null && group.getManagedGroups() != null && !group.getManagedGroups().isEmpty() )
-            {
-                return true;
-            }
-        }
-
-        return false;
+        return getManagedGroups().stream().anyMatch( Objects::nonNull );
```
jbee, line 1017: You misunderstood my intention in the previous comment, so the idea was not to call `getManagedGroups()` to avoid building intermediate collection(s), instead you would do basically the same first part of the code, but then use `anyMatch` to see if there is something to find.
```java
        return groups == null ? false
            : groups.stream()
                .anyMatch(group -> group.getManagedGroups() != null && !group.getManagedGroups().isEmpty);
```
Similar with `hasAuthorities()` ;)

jbee (commented):
Almost :D

jbee (approved):

netroms (approved):

PR #14330, opened 2023-06-22
chore: click house datasource [DHIS2-15456]

gnespolino:
Sample config:

```
clickhouse.connection.url = jdbc:clickhouse://localhost:8123/my_database
clickhouse.connection.username = username
clickhouse.connection.password = password
```

add `CLICKHOUSE_ENABLED=true` to system env to enable clickhouse datasource and jdbcTemplate


larshelge:
Just curious, would it be more future proof and less technology specific to name these properties in a more abstract way? I.e. we could target a general "analytics database" concept, and have a type property indicate the database technology in use? If we would like to extend later to another analytics database like AlloyDB, we can just add another enum type. 

Example properties:

```
analytics.db = clickhouse
analytics.connection.url = jdbc:clickhouse://localhost:8123/my_database
analytics.connection.username = username
analytics.connection.password = password
```

For other types, we could do:

```
analytics.db = alloydb
analytics.connection.url = jdbc:postgresql://localhost:5432/my_database
analytics.connection.username = username
analytics.connection.password = password
```

I think this is a more natural and extensible approval. Let me know your thoughts @gnespolino @maikelarabori .

PR #14360, opened 2023-06-26
fix: Skip repeatable events validation for program events [DHIS2-15471]

enricocolasante:


stian-sandvold (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EventImportValidationTest.java
```diff
@@ -207,6 +207,24 @@ void testNonRepeatableProgramStage()
         assertHasOnlyErrors( importReport, ValidationCode.E1039 );
     }
 
+    @Test
+    void testNonRepeatableProgramStageForProgramEvent()
```
muilpp, line 211: do we need to assert in two different parts of the test? could we not split it and have each assertion in a separate test?

enricocolasante: The second assertion depends on the first part to be successfully imported.
To correctly implement this we would need to create a new test class where we have a setup with the correct state for this test case, but for the purpose of this bug fix that is gong to be backported in multiple places, I would prefer to not change too many things

muilpp (approved):

PR #14361, opened 2023-06-26
fix: Analytics /events takes default relative period for analytics [2.41-DHIS2-15468]

d-bernat:
This fix addresses the issue related to `pe` and `filter` in analytics.
Currently, the `pe` in the `filter` is being ignored, and the default `pe` defined in system settings is being used in every request.


gnespolino (approved):

maikelarabori (approved):

Philip-Larsen-Donnelly (approved):
great! Let's get this merged to master and tested!

PR #14371, opened 2023-06-26
fix: Analytics /events takes default relative period for analytics [DHIS2-15468] 

maikelarabori:
_**[Backport from master/2.41]**_
This fix addresses the issue related to `pe` and `filter` in analytics.
Currently, the `pe` in the `filter` is being ignored, and the default `pe` defined in system settings is being used in every request.


Philip-Larsen-Donnelly (approved):
Approved based on testing on the dev version.

gnespolino (approved):

PR #14382, opened 2023-06-27
feat: deprecate and remove old generic PATCH method in API

netroms:
# Summary
* Removes the old PATCH code in the AbstractCrudController


david-mackessy:
What's the history of these `PATCH` endpoints?
- I see that 2 are being removed
  - `/{uid}`
  - `/{uid}/property`
- and there is only 1 similar endpoint remaining
  - `/{uid}`
- Is the old `/{uid}/property` not used at all any more?
- were they ever marked as deprecated to give notice of removal, if relevant?

jbee (approved):

vietnguyen (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/web/WebClient.java
```diff
@@ -152,13 +152,10 @@ default HttpResponse PATCH(String url, String body) {
         HttpMethod.PATCH, url, ContentType("application/json-patch+json"), Body(body));
   }
 
-  default HttpResponse PATCH_OLD(String url, String body) {
```
jbee, line 155: The naming `PATCH_OLD` maybe was a bit misleading. Because this is a general client so this isn't in any way connected to the CRUD old vs. new patch. We may have endpoints that do a `PATCH` HTTP method and this client should have a way to call them. So in hindsight we should have name this `PATCH_JSON` or something to make clear that this is sending a patch with a JSON body if there is one.

jbee (approved):

PR #14386, opened 2023-06-27
fix: Prevent NPE in 'ou' ownership export [DHIS2-15467]

maikelarabori:
**_[Backport from master/2.41]_**

This PR fixes an NPE that happens when the table `programownershiphistory` contains rows with null `organisationunitid`.

Two preventions were added: one at querying time, so we avoid rows without `organisationunitid`, and another at the code level.

A small change was also applied: in the SQL function `exists` I preferred `select 1`, as we only care about the number of rows returned. It's just another preventive change (we don't need to be concerned about the column name in the future).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcOwnershipAnalyticsTableManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcOwnershipAnalyticsTableManager.java
@@ -29,6 +29,7 @@ package org.hisp.dhis.analytics.table;
 
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.StringUtils.SPACE;
 import static org.hisp.dhis.analytics.ColumnDataType.CHARACTER_11;
 import static org.hisp.dhis.analytics.ColumnDataType.DATE;
 import static org.hisp.dhis.analytics.ColumnNotNullConstraint.NOT_NULL;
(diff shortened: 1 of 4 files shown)


gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcOwnershipAnalyticsTableManager.java
```diff
@@ -212,20 +213,25 @@ private String getInputSql( Program program )
...
             "union " +
-            "select o.trackedentityinstanceid, '" + TEI_OWN_TABLE_ID
-            + "' as startdate, null as enddate, o.organisationunitid " +
+            "select o.trackedentityinstanceid, '" + TEI_OWN_TABLE_ID +
+            "' as startdate, null as enddate, o.organisationunitid " +
             "from trackedentityprogramowner o " +
-            "where o.programid=" + program.getId() + " " +
-            "and exists (select programid from programownershiphistory p " +
+            "where o.programid=" + program.getId() + SPACE +
+            "and exists (select 1 from programownershiphistory p " +
```
Philip-Larsen-Donnelly, line 231: @maikelarabori I'm curious why something has changed from "programid" to "1"?

maikel-reddot: Hey @Philip-Larsen-Donnelly, that's just a convention for the `EXISTS` function, as what we select does not really matter. The function takes into consideration only the number of rows returned. My original comment is confusing, now that I read it again.
More details here:
https://www.postgresql.org/docs/current/functions-subquery.html#FUNCTIONS-SUBQUERY-EXISTS

maikelarabori: Hey @Philip-Larsen-Donnelly, that's just a convention for the `EXISTS` function, as what we select does not really matter. The function takes into consideration only the number of rows returned. My original comment is confusing, now that I read it again.
More details here:
https://www.postgresql.org/docs/current/functions-subquery.html#FUNCTIONS-SUBQUERY-EXISTS

PR #14400, opened 2023-06-28
fix: escape attribute filter value when it contains comma or colon [ DHIS2-15298 ] [ 2.40 ]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/13981

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
```diff
@@ -158,62 +139,154 @@ private static Stream<String> parseUidString( String input )
...
      */
-    public static List<QueryItem> parseAttributeQueryItems( Set<String> items,
+
+    public static List<QueryItem> parseAttributeQueryItems( String filterItem,
         Map<String, TrackedEntityAttribute> attributes )
         throws BadRequestException
     {
+        if ( StringUtils.isEmpty( filterItem ) )
+        {
+            return List.of();
```
enricocolasante, line 161: This is causing an issue in the exporter endpoint as this collection shouldn't be immutable in this context, in the store `QueryItem`s can be added to it to make the order work.
When trying to order on an attribute, the server returns a 500 error.

In master PR we found the issue but it is going to be solved in the controller refactoring, here it needs to be changed.

enricocolasante (changes requested):

enricocolasante (approved):

PR #14401, opened 2023-06-28
fix: Append prog. stage name for items with same name [DHIS2-15501]

maikelarabori:
These changes are related to the download/export of a Line Listing report.
Currently, we don't export the name of the program stage when we have two elements with the same name and different program stages. This leads to confusion, as the user doesn't know which element belongs to which program stage.

This fix will append also the program stage name, to the exported column, when there are elements with the same name and different program stages.


gnespolino (approved):

maikelarabori:
A new PR was created to deal with the new code formatting.

PR #14402, opened 2023-06-28
fix: get all available org units depending on ou mode [DHIS2-15394]

muilpp:
When a request uses the ouMode `DESCENDANTS` or `CHILDREN`, we don't check whether the user has access to any of the children org units related to the one in the request. More on this [here](https://docs.google.com/document/d/18C_NfbvoH0Dt4tySlsBMWsOJpTS0E_2TLtYCZVyIjNw/edit?pli=1). 

To fix this, there are two steps:
1. New validation in the service layer in case the mode is one of the two mentioned above. It gets all the lower level related org units and if any of them is the same as the capture scope/search scope org unit, the user is granted access.
2. Added a clause in the sql query, where we filter by the particular org unit(s) the user has access to.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-15394


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -237,19 +247,128 @@ private OrganisationUnit validateOrgUnit(String orgUnitUid) throws BadRequestExc
...
+      Program program,
+      User user,
+      OrganisationUnit orgUnit,
+      OrganisationUnitSelectionMode orgUnitMode,
+      Function<String, List<OrganisationUnit>> orgUnitDescendants) {
+
+    if (orgUnitMode == null) {
+      if (orgUnit != null) {
+        return getSelectedOrgUnits(user, program, orgUnit);
+      }
```
enricocolasante, line 298: Can we organize the if to make it clear that they are guard clauses even if we repeat some condition?

```suggestion
    if (orgUnitMode == null && orgUnit != null) {
        return getSelectedOrgUnits(user, program, orgUnit);
      }
      
      if (orgUnitMode == null && orgUnit == null) {
        return getAccessibleOrgUnits(user, program, orgUnit);
      }
      
```

Like this it is at a first sight what are the conditions

If the previous comment is implemented we do not even need these guard clauses

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -157,10 +160,17 @@ public EventSearchParams map(EventOperationParams operationParams)
...
+                program,
+                user,
+                orgUnit,
+                operationParams.getOrgUnitSelectionMode(),
+                organisationUnitService::getOrganisationUnitWithChildren))
         .setTrackedEntity(trackedEntity)
         .setProgramStatus(operationParams.getProgramStatus())
         .setFollowUp(operationParams.getFollowUp())
-        .setOrgUnitSelectionMode(operationParams.getOrgUnitSelectionMode())
+        .setOrgUnitSelectionMode(getOrgUnitMode(orgUnit, operationParams.getOrgUnitSelectionMode()))
```
enricocolasante, line 173: If we extract this orgUnitMode to a variable and use it in the validateUser method we can remove the guard clauses there because we are sure that orgUnitMode is not null.

Doing this remove the duplication of logic that now is present in getOrgUnitMode method and validateUser method

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
```diff
@@ -0,0 +1,439 @@
...
+    program = programStage.getProgram();
+
+    // to test that events are only returned if the user has read access to ALL COs of an events COC
+    CategoryOption categoryOption = get(CategoryOption.class, "yMj2MnmNI8L");
+    categoryOption.getSharing().setOwner("o1HMTIzBGo7");
+    manager.update(categoryOption);
+
+    manager.flush();
+
+    injectSecurityContext(userService.getUser("M5zQapPyTZI"));
```
enricocolasante, line 92: Do you need this inject then?
I think the setUp method will override this anyway

enricocolasante (approved):
Just a couple of style comments

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: I rather not have the `validateUser` method return a list of org units, but I think you are right to extract this in a new variable, I'll create a new method for that.

enricocolasante: Sorry, that was confusing. I didn't mean validateUser method, I meant getUserAccessibleOrgUnits method

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -156,11 +162,13 @@ public EventSearchParams map(EventOperationParams operationParams)
     return searchParams
         .setProgram(program)
         .setProgramStage(programStage)
-        .setOrgUnit(orgUnit)
+        .setOrgUnit(requestedOrgUnit)
```
teleivo, line 165: Why do we need to keep track of both the `requestedOrgUnit` and the `accessibleOrgUnit`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -196,6 +204,27 @@ public EventSearchParams map(EventOperationParams operationParams)
         .setIncludeDeleted(operationParams.isIncludeDeleted());
   }
 
+  private List<OrganisationUnit> validateAccessibleOrgUnits(
```
teleivo, line 207: nit: I would put this after `validateRequestedOrgUnit` so its in call order. I would also put `user` first as we are validating the user has access to the orgUnit, ... We are doing that in the ACL related services with `hasAccess(User user, ...)`/canAccess(User user, ...)` 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -236,22 +265,115 @@ private OrganisationUnit validateOrgUnit(String orgUnitUid) throws BadRequestExc
...
+   *
+   * @param program the program the user wants to access to
+   * @param user the user to check the access of
+   * @param orgUnit parent org unit to get descendants/children of
+   * @param orgUnitDescendants function to retrieve org units, in case ou mode is descendants
+   * @return a list containing the user accessible organisation units
+   */
+  private List<OrganisationUnit> getUserAccessibleOrgUnits(
+      Program program,
+      User user,
```
teleivo, line 293: nit: put `user` (then `orgUnit`, `orgUnitMode`) first as described above 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -236,22 +265,115 @@ private OrganisationUnit validateOrgUnit(String orgUnitUid) throws BadRequestExc
...
+      OrganisationUnit orgUnit,
+      OrganisationUnitSelectionMode orgUnitMode,
+      Function<String, List<OrganisationUnit>> orgUnitDescendants) {
+
+    if (orgUnitMode == null && orgUnit != null) {
```
teleivo, line 298: Is this behavior documented?
How can this be achieved by a user? AFAIK DESCENDANTS is the default, would `orgUnitMode` be null when a request with `ouMode=` is made?

It might make it more readable if we set the `orgUnitMode` to SELECTED/ACCESSIBLE ideally before calling this method (your `getOrgUnitMode()` has the logic we need I think). We would then only need to return the switch expression.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -236,22 +265,115 @@ private OrganisationUnit validateOrgUnit(String orgUnitUid) throws BadRequestExc
...
+   * there's a match, it means the user org unit is at the same level or above the supplied org
+   * unit.
+   *
+   * @param availableOrgUnits the org units to check if the user has access to
+   * @param program the program the user wants to access to
+   * @param user the user to check the access of
+   * @return a list with the org units the user has access to
+   */
+  private List<OrganisationUnit> getAccessibleDescendants(
+      List<OrganisationUnit> availableOrgUnits, Program program, User user) {
```
teleivo, line 347: nit: put user first

teleivo, line 347: nit: `availableOrgUnits` does not add value IMHO

```suggestion
      User user, Program program, List<OrganisationUnit> orgUnits) {
```

I switched order to keep user first and program second as this method seems almost like an overloaded one of

`private List<OrganisationUnit> getAccessibleOrgUnits(User user, Program program) {`

so we stay consistent

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1570,87 +1572,40 @@ private void setAccessiblePrograms(User user, EventSearchParams params) {
...
+              + "."
+              + PATH_LIKE
+              + " '"
+              + orgUnit.getPath()
+              + "%' "
+              + " and "
+              + ouTable
+              + "."
+              + "hierarchylevel = "
+              + orgUnit.getLevel());
```
teleivo, line 1590: do we not need to +1 anymore as we already got the children into `params.getAccessibleOrgUnits()`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1025,6 +1021,15 @@ private StringBuilder getFromWhereClause(
     return fromBuilder;
   }
 
+  private String getOrgUnitSql(EventSearchParams params, String ouTable) {
+    return switch (params.getOrgUnitSelectionMode()) {
+      case SELECTED -> getSelectedOrgUnitPath(params.getAccessibleOrgUnits(), ouTable);
```
teleivo, line 1026: your TrackedEntityQueryParams javadocs say

```
   * Prepares the organisation units of the given parameters to simplify querying. Mode ACCESSIBLE
is converted to DESCENDANTS for organisation units linked to the search scope of the given
   * user. Mode CAPTURE is converted to DESCENDANTS too, but using organisation units linked to the
   * user's capture scope, and mode CHILDREN is converted to SELECTED for organisation units
   * including all their children. Mode can be DESCENDANTS, SELECTED, ALL only after invoking this
   * method.
```

why are we dealing with modes other than `DESCENDANTS, SELECTED, ALL` here?

teleivo (commented):
Thanks a lot @muilpp for digging into ACL 😅 

Most of my comments are nitpicks and questions for me to understand our ACL better. Let's huddle and walk through the PR together 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: I'm not sure this is documented, but it's the conclusion I got after analyzing the code. There's [this ticket](https://dhis2.atlassian.net/browse/TECH-1588) to harmonize the endpoints when it comes to setting a default org unit mode

DESCENDANTS is set as the default but Spring is not picking it up (@enricocolasante knows more about this). Anyway, it does not make sense that DESCENDANTS is the default, but this will be tackled in the aforementioned ticket.

Regarding the change you requested, I think it makes sense to use the `getOrgUnitMode` method and have a clean switch expression in this one only.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: Exactly. Also, the +1 was forcing us to get the children only, but we also want to get the parent (the org unit in the request).

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: The javadocs refer to the `/trackedEntities` endpoint, and the `getOrgUnitSql` is meant for events only. I did that change in the docs while I was analyzing the ACL logic, maybe I should have done it in a separate ticket, but since it was such a small change...

teleivo (approved):

PR #14406, opened 2023-06-28
fix: show database spatialSupport in the system info endpoint (#13564)

netroms:
* fix: show database spatialSupport in the system info endpoint

(cherry picked from commit 7398f29fa03875bfbf42c3db73fbd63b1988a1b2)


dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/database/DatabaseInfoTest.java
```diff
@@ -66,7 +65,6 @@ void clearSensitiveInfo()
         assertNull( databaseInfo.getUrl() );
         assertNull( databaseInfo.getPassword() );
         assertNull( databaseInfo.getDatabaseVersion() );
-        assertFalse( databaseInfo.isSpatialSupport() );
```
david-mackessy, line 69: could the assertion be kept to confirm the value isn't cleared anymore?

jbee (approved):

PR #14414, opened 2023-06-29
fix: Long text does not display in pivot tables in data visualizer app [DHIS2-14188]

maikelarabori:
**_[Backport from master/2.41]_**

This fixes the aggregation type for type text types.
 Pivot tables created in the Data Visualizer app fail to display fields of “Long Text” type.


gnespolino (approved):

maikelarabori:
Should not go into 2.37.10, RCB.

PR #14415, opened 2023-06-29
fix: Update DataValueSet import endpoint to ignore dates supplied [DHIS2-8084]

david-mackessy:
# Summary
## Issue
When importing `dataValue`s that contain date values for `created` and `lastUpdated`, these values can override existing values in the database.

It looks like this was originally implemented as expected behaviour. There was even an [existing test](https://github.com/dhis2/dhis2-core/blob/8a96c2e3cce99e9fea558ee38da0cf06b3b164a4/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/datavalueset/DataValueSetServiceIntegrationTest.java#L440) confirming that overriding dates was possible during an import.

## Expected behaviour
The Platform backend team spoke about this and agreed that this is not the behaviour that we would expect for importing.
We would expect:
- `created`
  - date is only updated once on creation
  - uses the timestamp at the time of creation
  - can’t be changed after
- `lastUpdated`
  - date is updated on creation
  - uses the timestamp at the time of creation
  - date is updated after each value change, using the timestamp at the time of update
- dates supplied to the import endpoint will not be used

Reported as a bug but it is really a feature/change. Will keep as a bug for simplicity and implement the change as originally requested in this issue.

## API info
- endpoint:
  - `POST` `/api/dataValueSets`
  - `POST` `/api/dataValueSets?async=false` (to get the import summary response straight away)
  - `GET` `/api/dataValueSets?orgUnit={orgUnitId}&startDate=1922-01-01&endDate=2024-01-30&dataElement={dataElementId}` (to check values from database after import)
- endpoint accepts 5 different types of content formats
  - `JSON`
  - `XML`
  - `ADX`
  - `CSV`
  - `PDF`

## Change
- 4 of these formats (`JSON`, `XML`, `CSV`, `PDF`) get processed by the same service (`DefaultDataValueSetService`) and are channelled through the same processing point, so only 1 change was necessary
- the ADX format is processed by the `DefaultAdxDataService` and this did not require any changes as it already had the expected behaviour

# Tests
## Automated
- Added tests to cover the new behaviour in `DefaultDataValueSetService`
- Added tests for `DefaultAdxDataService` to explicitly define what behaviour we expect from that service method
- Removed 1 existing test which checked for overriding dates

## Manual
- Can be tested manually using the steps and attached xml file in [Jira ticket - DHIS2-8084](https://dhis2.atlassian.net/browse/DHIS2-8084). The `dataElement` and `orgUnit` need to exist in the database.
- Can use the `POST` endpoints detailed above
- Can also use this `JSON` payload, which needs the `dataElement` and `orgUnit` present in the database also
``` json
{
    "dataValues": [
        {
            "dataElement": "DElement100",
            "period": "201910",
            "orgUnit": "Orgunit1000",
            "categoryOptionCombo": "HllvX50cXC0",
            "attributeOptionCombo": "HllvX50cXC0",
            "value": "23",
            "storedBy": "system",
            "lastUpdated": "1987-12-21T23:59:38.376+0000",
            "created": "1987-12-21T23:59:38.376+0000",
            "followUp": "false"
        }
    ]
}
```


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -1212,8 +1212,8 @@ private DataValue createDataValue( DataValueEntry dataValue, ImportContext conte
         internalValue.setAttributeOptionCombo( valueContext.getAttrOptionCombo() );
         internalValue.setValue( trimToNull( value ) );
         internalValue.setStoredBy( context.getStoredBy( dataValue ) );
-        internalValue.setCreated( dataValue.hasCreated() ? parseDate( dataValue.getCreated() ) : now );
-        internalValue.setLastUpdated( dataValue.hasLastUpdated() ? parseDate( dataValue.getLastUpdated() ) : now );
+        internalValue.setCreated( now );
```
jbee, line 1215: I understand that removing the `dataValue` as a source leaves us with just using `now` all the time but I wonder if that would then carry one the `now` for existing values. I checked the code and in `DefaultDataValueSetService#960` we are doing 
```java
context.getDataValueBatchHandler().updateObject( internalValue );
```
Shouldn't that mean the `created` of the existing value gets effectively overridden with the `now` of the import?

jbee: missed that :)

jbee (approved):

PR #14429, opened 2023-06-29
fix: do not try to add indexes to views in SQL script

netroms:
# Summary
The "V2_41_10__Add_missing_uid_indexes_all_tables.sql" script would not check if tables scanned for missing UID indexes were a view and would incorrectly try to add an index to it. This would fail since this is not possible.
This fix adds a check to filter out tables starting with "_view".


jbee (approved):

david-mackessy (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_10__Add_missing_uid_indexes_all_tables.sql
```diff
@@ -98,7 +101,7 @@ BEGIN
     FOR rec IN (
         SELECT * FROM get_tables_missing_uid_index()
     ) LOOP
-            IF rec.table_name <> 'audit' THEN
+            IF rec.table_name <> 'audit' AND rec.table_name NOT LIKE '_view%' THEN
```
larshelge, line 104: This approach is actually not quite reliable, as there might be custom views created directly in the database (not by DHIS 2 SQL views) with any name.

To get the table type (table or view) you can join with the `tables` table and check the `table_type` column, something like this:

```
select c.table_schema, c.table_name, t.table_type 
from information_schema.columns c
inner join information_schema.tables t on c.table_schema = t.table_schema and c.table_name = t.table_name 
where c.column_name = 'uid'
and t.table_type = 'BASE TABLE';
```

jbee (approved):

david-mackessy (approved):

PR #14462, opened 2023-07-03
fix: Update User API when updating UserGroups to update lastUpdatedBy [DHIS2-15357]

david-mackessy:
# Summary
The APP UI uses the following endpoint to add/remove a `User` to a `UserGroup`
- `PUT /api/users/{id}` - with `User` body 

## Existing behaviour
When adding/removing a `User` to/from a `UserGroup` using this API the `lastUpdatedBy` property is not being updated in the updated `UserGroup`

## New Behaviour
A change in the `DefaultUserGroupService` will now ensure that the `UserGroup` property `lastUpdatedBy` gets updated when called through this API.

## Note
Here are several other APIs (probably more) that can update a `UserGroup` - all of which already work as expected (`lastUpdatedBy` updating correctly). These have been manually tested during this bug analysis.
- `User` API
  - Add
    - `POST /api/users/{id}/userGroups/{id}`
  - Remove
    - `DELETE /api/users/{id}/userGroups/{id}`
- `UserGroup` API
  - Add
    - `POST /api/userGroups/{id}/users/{id}`
    - `PUT /api/userGroups/{id}` - with `UserGroup` body
  - Remove
    - `DELETE /api/userGroups/{id}/userGroups/{id}`
    - `PUT /api/userGroups/{id}` - with `UserGroup` body

# Testing
## Updated API
- Automated tests added testing adding/removing `User` to/from `UserGroup` and checking the `lastUpdatedBy` property
- Manually tested locally

## Existing APIs
- tested manually to confirm behaviour

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserGroupService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserGroupService.java
@@ -208,6 +208,7 @@ public class DefaultUserGroupService
             if ( !updates.contains( userGroup ) && canAddOrRemoveMember( userGroup.getUid(), currentUser ) )
             {
                 userGroup.removeUser( user );
+                userGroupStore.update( userGroup, currentUser );
             }
         }
 
@@ -216,7 +217,7 @@ public class DefaultUserGroupService
             if ( canAddOrRemoveMember( userGroup.getUid(), currentUser ) )
             {
                 userGroup.addUser( user );
-                userGroupStore.updateNoAcl( userGroup );
+                userGroupStore.update( userGroup, currentUser );
             }
         }
     }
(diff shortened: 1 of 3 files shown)


netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserGroupService.java
```diff
@@ -216,7 +217,7 @@ public void updateUserGroups( User user, @Nonnull Collection<String> uids, User
             if ( canAddOrRemoveMember( userGroup.getUid(), currentUser ) )
             {
                 userGroup.addUser( user );
-                userGroupStore.updateNoAcl( userGroup );
```
jbee, line 219: I seems wrong that the no ACL variant would behave differently other than not making the ACL checks. If that is the case maybe we should talk about changing that and making it behave the same.

jbee (approved):

PR #14463, opened 2023-07-03
chore: Harmonize Tracked Entity Controller[TECH-1571]

enricocolasante:
This PR creates a `TrackedEntityOperationParams` object that is used in the service layer and all the needed mappers and tests.
Same work was done [here](https://github.com/dhis2/dhis2-core/pull/14180) and [here](https://github.com/dhis2/dhis2-core/pull/14048) for events and enrollments

**Next Steps**

- Use a `TrackedEntities` object to return result from the server to hide some of the pagination complexity? (It is done this way in other entities)
- Remove `includeDeleted` from `TrackedEntityParams`
- Correclty set default values in mapper and not in `RequestParams` as Spring binding doesn't work that way


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -0,0 +1,301 @@
...
+        throws BadRequestException,
+        ForbiddenException
+    {
+        Program program = validateProgram( operationParams.getProgramUid() );
+        ProgramStage programStage = validateProgramStage( operationParams, program );
+        TrackedEntityType trackedEntityType = validateTrackedEntityType( operationParams.getTrackedEntityTypeUid() );
+
+        User user = operationParams.getUser();
+        Set<OrganisationUnit> orgUnits = validateOrgUnits( user, operationParams.getOrganisationUnits(), program );
+        if ( operationParams.getOrganisationUnitMode() == OrganisationUnitSelectionMode.CAPTURE && user != null )
```
muilpp, line 105: could this not be done in the method `validateOrgUnits` or in a new one?
also, ACCESSIBLE is very similar to CAPTURE mode, what happens in that case?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/RequestParams.java
```diff
@@ -98,7 +98,7 @@ class RequestParams extends PagingAndSortingCriteriaAdapter
      * Selection mode for the specified organisation units, default is
      * ACCESSIBLE.
      */
-    private OrganisationUnitSelectionMode ouMode;
+    private OrganisationUnitSelectionMode ouMode = OrganisationUnitSelectionMode.DESCENDANTS;
```
muilpp, line 101: Why do we assume the default is DESCENDANTS?

enricocolasante: That is a good question but I am leaving the answer to the ACL epic.
I just left everything as it was before in this PR

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
enricocolasante: Moved inside the `validateOrgUnits` method.
I guess this is not working correctly 100%, anyway, I will leave any functional change to this part for the ACL epic.

muilpp (approved):

zubaira (approved):

PR #14500, opened 2023-07-06
chore: format pom.xml using spotless TECH-1583

teleivo:
* fix warnings like http is not secure by replacing it with an https url
  * replaced `http://maven.apache.org/xsd/maven-V4_0_0.xsd` with `http://maven.apache.org/xsd/maven-4.0.0.xsd` (which you get when creating a new maven module with IntelliJ; without the warning that http is insecure)


jbee (approved):
Looks good to me :)

enricocolasante (approved):

PR #14506, opened 2023-07-06
fix: Return Redis Notifications sorted by time [DHIS2-14838]

david-mackessy:
# Summary
## Problem when using Redis
The following App UI pages have their action buttons disabled permanently after initiating calls to the API:
- `Data Administration` -> `Resource Tables` (clicking `Generate tables` button)
  - API: `POST /api/resourceTables`
- `Data Administration` -> `Analytics Tables` (clicking `Start export` button)
  - API: `POST /api/resourceTables/analytics`

The reason this is a problem for the UI is because the UI seems to poll for `Notification`s, wanting to know when the process is completed so it can reenable the button.

The UI knows the process has completed when it detects a `Notification` marked as `completed: true` at the top of the returned list.

When using Redis, the `Notification`s are returned with the completed `Notification` not at the top of the list.

## Cause
The cause of this behaviour is due to the sequence of a few steps in the `RedisNotifier` - detailed in the Old behaviour section below

## Old behaviour
When running DHIS2 with Redis, the `RedisNotifier` was returning `Notification`s in an unexpected order.  
Processing steps:
1. get data from Redis (`Set<String>`)
2. sort (this is not the sorting we intended as it's still in `String` form)
3. map `String` to `Notification`
4. add `Notification` to `Collection`
5. return `Collection<Notification>`

## New behaviour
Processing steps:
1. get data from Redis (`Set<String>`)
2. map `String` to `Notification`
3. add `Notification` to `Collection`
4. sort (`Notification` compares by time which is the desired format)
5. return `Collection<Notification>`

## Note
The use of a `PriorityQueue` was investigated to leverage its automatic natural order sorting (to avoid an explicit call to sort) but the results were erratic when testing. Using `List` and then sorting was the easiest and safest solution while keeping the interface contract intact and not having to interfere with the `InMemoryNotifier` which is by far the most-used implementation.

# Testing

## Automatic
Test added to confirm new expected behaviour of ordered elements from `RedisNotifier`

## Manual
Manual testing requires the Server to be configured to use Redis

### APP UI
- Visiting these 2 pages and trigger their processes
  - `Data Administration` -> `Resource Tables` (clicking `Generate tables` button)
    - The `Generate tables` button should be reenabled after the process completes
  - `Data Administration` -> `Analytics Tables` (clicking `Start export` button)
    - The `Start export` button should be reenabled after the process completes

### API
- Trigger Resource Tables generation
  - `POST /api/resourceTables`
- Trigger Analytics Tables export
  - `POST /api/resourceTables/analytics`

Check progress/completion by calling the endpoints below - the `Notification` marked with `completed: true` should be the first in the list once the process is completed.
- `GET /api/system/tasks/ANALYTICS_TABLE`
- `GET /api/system/tasks/ANALYTICS_TABLE/{id}`
- `GET /api/system/tasks/RESOURCE_TABLE`
- `GET /api/system/tasks/RESOURCE_TABLE/{id}`


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/leader/election/RedisLeaderManager.java
```diff
@@ -67,12 +67,13 @@ public class RedisLeaderManager implements LeaderManager {
   public RedisLeaderManager(
       Long timeToLiveMinutes,
       StringRedisTemplate redisTemplate,
-      DhisConfigurationProvider dhisConfigurationProbider) {
-    this.nodeId = dhisConfigurationProbider.getProperty(ConfigurationKey.NODE_ID);
+      DhisConfigurationProvider dhisConfigurationProvider) {
```
david-mackessy, line 70: fixed typo

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/leader/election/RedisLeaderManager.java
```diff
@@ -67,12 +67,13 @@ public class RedisLeaderManager implements LeaderManager {
...
+    this.nodeId = dhisConfigurationProvider.getProperty(ConfigurationKey.NODE_ID);
     this.nodeUuid = UUID.randomUUID().toString();
     log.info(
-        "Setting up redis based leader manager with NodeUuid:%s and NodeID:%s",
-        this.nodeUuid, this.nodeId);
+        "Setting up redis based leader manager with NodeUuid:{} and NodeID:{}",
```
david-mackessy, line 74: fixed log params

jbee (approved):

larshelge (approved):

PR #14512, opened 2023-07-07
chore: allow turning off formatter for java via spotless DHIS2-15518

teleivo:
google java formatter cannot be disabled. spotless allows us to disable any formatter https://github.com/diffplug/spotless/tree/main/plugin-maven\#spotlessoff-and-spotlesson Use the same pattern we used before (the one recognized by the Eclipse formatter) also for disabling the pom.xml formatter

We should only turn off the formatter when absolutely necessary. Like when creating a DSL that represents a tree that would otherwise get flattened or complex SQL statements. Javadocs should use the appropriate html tags to avoid unwanted formatting.


enricocolasante (approved):

jbee (approved):
Thanks again!

PR #14553, opened 2023-07-11
chore: Remove includeDeleted from field params [TECH-1571]

enricocolasante:
This PR removed redundant `includeDeleted` filed from `TrackedEntityParams`.
A failing test was disabled and it will be fixed when harmonizing the relationship controller.

**Next Steps**

- Use a `TrackedEntities` object to return result from the server to hide some of the pagination complexity? (It is done this way in other entities)
- Correclty set default values in mapper and not in `RequestParams` as Spring binding doesn't work that way

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
@@ -99,6 +99,7 @@ import org.springframework.beans.factory.annotation.Autowired;
  * @author Luciano Fiandesio
  */
 class TrackedEntityServiceTest extends IntegrationTestBase {
+
   @Autowired protected UserService _userService;
 
   @Autowired private TrackedEntityService trackedEntityService;
@@ -152,12 +153,6 @@ class TrackedEntityServiceTest extends IntegrationTestBase {
   private Relationship relationshipB;
 
   private Relationship relationshipC;
-
-  private static List<String> uids(Collection<? extends BaseIdentifiableObject> trackedEntities) {
-    return trackedEntities.stream()
-        .map(BaseIdentifiableObject::getUid)
-        .collect(Collectors.toList());
-  }
 
   @Override
   protected void setUpTest() throws Exception {
@@ -628,6 +623,7 @@ class TrackedEntityServiceTest extends IntegrationTestBase {
   }
 
   @Test
+  @Disabled("IncludeDeleted param is not working when TE has a deleted relationship")
   void shouldIncludeDeletedEnrollmentAndEvents()
       throws ForbiddenException, NotFoundException, BadRequestException {
     TrackedEntityOperationParams operationParams =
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1017,7 +1015,7 @@ void shouldReturnTrackedEntityRelationshipsWithTei2Event()
         () -> assertEquals(eventA.getUid(), actual.getTo().getEvent().getUid()));
   }
 
-  private static List<String> uids(Collection<? extends BaseIdentifiableObject> trackedEntities) {
```
teleivo, line 1020: I think it can stay static as it needs nothing from the instance right 🤔

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1017,7 +1015,7 @@ void shouldReturnTrackedEntityRelationshipsWithTei2Event()
...
+  private List<String> uids(Collection<? extends BaseIdentifiableObject> trackedEntities) {
```
teleivo, line 1018: Maybe we should rename this to something like

```suggestion
  private List<String> uids(Collection<? extends BaseIdentifiableObject> identifiableObjects) {
```

as its generic and not trackedEntities specific

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportController.java
```diff
@@ -200,7 +200,10 @@ ResponseEntity<ObjectNode> getTrackedEntityByUid(
     TrackedEntity trackedEntity =
         TRACKED_ENTITY_MAPPER.from(
             trackedEntityService.getTrackedEntity(
-                uid.getValue(), program == null ? null : program.getValue(), trackedEntityParams));
+                uid.getValue(),
+                program == null ? null : program.getValue(),
+                trackedEntityParams,
+                false));
```
teleivo, line 206: Can you please make sure we have a test in TrackedEntitiesExportControllerTest that tests users get a 404 if they request a soft-deleted trackedEntity 😄

teleivo (commented):
Nice, long overdue 🥳 
I am just missing a test for the 404 on soft-deleted trackedEntities (see comment) for ✅

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerTest.java
```diff
@@ -390,6 +397,14 @@ void getTrackedEntityByIdNotFound() {
         GET("/tracker/trackedEntities/Hq3Kc6HK4OZ").error(HttpStatus.NOT_FOUND).getMessage());
   }
 
+  @Test
+  void shouldReturnNotFoundWhenGettingASoftDeletedTrackedEntityById() {
+    String uid = softDeletedTrackedEntity.getUid();
+    assertEquals(
```
teleivo, line 403: IMHO there is no need to assert on the message. A 404 says it all 😄 Asserting on the exact message always makes a test have to change with the implementation.

teleivo (approved):

lucaCambi77 (approved):

PR #14555, opened 2023-07-11
fix: provide displayName for all data integrity checks [DHIS2-14770] (2.37.10)

jbee:
backport of #13102


Philip-Larsen-Donnelly (approved):
Approved for 2.37.10. Thanks @jbee!
We should push this into `2.37` too, to make sure it is there if there are any "End of Support" builds later.

PR #14565, opened 2023-07-12
feat: use hibernate bag tag to map java List [DHIS2-15509]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-15509

### Summary
- This PR adds a new mechanism for supporting map `java.util.List` to `<bag>` in `hbm.xml` file with or without `sortOrder` property. More details in jira [DHIS2-15509](https://dhis2.atlassian.net/browse/DHIS2-15509).
- A new interface `SortableObject` is added which should be implemented by Object mapped to a List that need to be sorted.
- Update `Option` to implement `SortableObject` interface as a POC. `OptionSet.hbm.xml` uses `<bag>` with `sortOrder` for  mapping `List<Option>`.
- Update `Category.hbm.xml` for mapping `categoryCombos` without `sortOrder` using `<bag>`.

### Test
- Several tests added to `OptionControllerTest` and `MetadataImportExportControllerTest` for testing import `OptionSet.options` with and without `sortOrder`.

[DHIS2-15509]: https://dhis2.atlassian.net/browse/DHIS2-15509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
@@ -27,15 +27,12 @@
  */
 package org.hisp.dhis.dxf2.metadata.objectbundle.hooks;
 
-import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
-
 import lombok.AllArgsConstructor;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.hisp.dhis.analytics.SortOrder;
 import org.hisp.dhis.attribute.Attribute;
 import org.hisp.dhis.attribute.AttributeValue;
 import org.hisp.dhis.common.BaseIdentifiableObject;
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
```diff
@@ -73,6 +78,58 @@ public void preCreate(IdentifiableObject identifiableObject, ObjectBundle bundle
...
+                  sortableObject.setSortOrder(i);
+                  bundle.getPreheat().put(bundle.getPreheatIdentifier(), preheatedItem);
+                }
+              }
+              bundle.getPreheat().put(bundle.getPreheatIdentifier(), identifiableObject);
+            });
+  }
+
+  /**
+   * Fina all properties of given Schema class that are of type List which contains SortableObjects
```
david-mackessy, line 119: typo ('Fina' -> 'Find')

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
```diff
@@ -73,6 +78,58 @@ public void preCreate(IdentifiableObject identifiableObject, ObjectBundle bundle
...
+
+  /**
+   * This method loops through all sortable List of given object and sets the sortOrder value for
+   * each item in the List if it is not already set.
+   *
+   * @param identifiableObject Object to set sortOrder on
+   * @param bundle {@link ObjectBundle}
+   * @param schema Schema of given object
+   */
+  private void handleSortOrder(
```
david-mackessy, line 92: Seeing as `BaseNameableObject` now implements `SortableObject`, does that now mean that some objects will now have a sort order whether they need/want it? (I'm guessing that not all `BaseNameableObject`s need sort order). Just trying to understand the bigger picture for this.
So some objects will have a sortOder:
- which is used in the UI
- which is purely used for use in the DB?

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
vietnguyen: I didn't make `BaseNameableObject` to implement `SortableObject`, only `Option` implement it. 

```
public class Option extends BaseNameableObject implements MetadataObject, SortableObject {
```

david-mackessy: sorry, missed that :)

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
```diff
@@ -73,6 +78,58 @@ public void preCreate(IdentifiableObject identifiableObject, ObjectBundle bundle
...
+              for (int i = 0; i < collection.size(); i++) {
+                IdentifiableObject item = collection.get(i);
+                IdentifiableObject preheatedItem =
+                    bundle.getPreheat().get(bundle.getPreheatIdentifier(), item);
+                if (preheatedItem == null) {
+                  continue;
+                }
+                SortableObject sortableObject = (SortableObject) preheatedItem;
+                if (sortableObject.getSortOrder() == null) {
+                  sortableObject.setSortOrder(i);
```
jbee, line 110: I wonder if we should make this a bit smarter. Since we only set the order for objects which did not have one, we might not want to set it too `i` but to previous item's value + 1?

jbee (approved):
Looks good as a starting point. Not sure we want to set the order to the index in the sequence or the previous item's order + 1.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/IdentifiableObjectBundleHook.java
vietnguyen: Thought about this, i think it doesn't matter where those `null` item are put by the backend, User will eventually need to rearrange it so I made the code to be as simple as possible. 
Anyway, updated the code and added the index variable. We can re-evaluate this logic later.

PR #14567, opened 2023-07-12
feat: BadSqlGrammarException specific handlers [DHIS2-15184]

gnespolino:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/errorhandling/analytics/SqlStateHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.hisp.dhis.dxf2.webmessage.WebMessage;
+import org.hisp.dhis.dxf2.webmessage.WebMessageUtils;
+import org.springframework.jdbc.BadSqlGrammarException;
+
+public abstract class SqlStateHandler {
+
+  public static final SqlStateHandler DEFAULT_STATE_HANDLER = new SqlStateHandler() {};
+
+  public WebMessage handle(BadSqlGrammarException ex) {
```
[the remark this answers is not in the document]
gnespolino: some `SqlStateHandler` implementation might use the exception to further customize the error message

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -541,6 +541,7 @@ public Grid getAggregatedEventData(EventQueryParams params, Grid grid, int maxLi
       }
     } catch (BadSqlGrammarException ex) {
       log.info(AnalyticsUtils.ERR_MSG_TABLE_NOT_EXISTING, ex);
+      throw ex;
```
maikelarabori, line 544: I'm wondering if we couldn't just throw `new QueryRuntimeException(ErrorMessage errorMessage);`
In the error message, we could use the `ErrorCode` and define a message like:
`Bad SQL grammar. Error code "SQL_STATE_CODE"`. We could even have an Enum to hold the known SQL state codes (like the ones you have defined "42P01" and "42", and a generic one if a non-supported code is returned).

Note that QueryRuntimeException does not have a constructor for `ErrorMessage`, but I believe it could be added "_easily_".

In this way, we can return any SQL state of `BadSqlGrammarException`, and don't need to add any of the new classes below... The current flow will take care of everything. What do you think?

maikelarabori (commented):
Hey @gnespolino, I left a comment. Please take a look. Maybe we can simplify it a bit.

Thx!

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsUtils.java
```diff
@@ -980,6 +984,28 @@ public static void throwIllegalQueryEx(ErrorCode errorCode, Object... args) {
     throw new IllegalQueryException(new ErrorMessage(errorCode, args));
   }
 
+  /**
+   * Throws an {@link QueryRuntimeException} based on the given {@link BadSqlGrammarException}.
+   *
+   * @param ex the {@link BadSqlGrammarException}.
+   */
+  public static void throwQueryRuntimeException(BadSqlGrammarException ex) {
+
```
maikelarabori, line 993: I think this extra line can be removed

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/EventAnalyticsControllerTest.java
```diff
@@ -70,6 +71,8 @@ void setUp() {
   }
 
   @Test
+  @Disabled(
+      "generated query will fail on H2 database and we are now propagating BadSqlGrammarException since DHIS2-15184")
   void testGetQueryJson() {
```
maikelarabori, line 76: Now that we return a conflict message, we could assert the error message, instead of disabling it.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/EnrollmentAnalyticsControllerTest.java
```diff
@@ -70,6 +71,8 @@ void setUp() {
   }
 
   @Test
+  @Disabled(
+      "generated query will fail on H2 database and we are now propagating BadSqlGrammarException since DHIS2-15184")
   void testGetQueryJson() {
```
maikelarabori, line 76: Now that we return a conflict message, we could assert the error message, instead of disabling it.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -146,6 +146,22 @@ protected void initBinder(WebDataBinder binder) {
     binder.registerCustomEditor(UID.class, new UIDParamEditor());
   }
 
+  @ExceptionHandler
+  @ResponseBody
+  public WebMessage queryRuntimeException(QueryRuntimeException ex) {
```
maikelarabori, line 151: We could potentially encapsulate this code somewhere else.
Let's discuss the possibilities.

maikelarabori (commented):
Looks good @gnespolino, I left a few minor comments.

d-bernat (approved):

maikelarabori (approved):

PR #14570, opened 2023-07-12
fix: leader only job as job type property [DHIS2-15288]

jbee:
TBD

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
@@ -442,13 +442,9 @@ public class JobConfiguration extends BaseIdentifiableObject implements Secondar
   }
 
   @JacksonXmlProperty
-  @JsonProperty
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
   public boolean isLeaderOnlyJob() {
-    return leaderOnlyJob;
-  }
-
-  public void setLeaderOnlyJob(boolean leaderOnlyJob) {
-    this.leaderOnlyJob = leaderOnlyJob;
+    return !jobType.isRunOnAllNodes();
   }
 
   public boolean isInMemoryJob() {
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/scheduling/JobSchedulerController.java
```diff
@@ -106,7 +105,6 @@ public List<SchedulerEntry> getQueueableJobs(@RequestParam(required = false) Str
             : config -> !name.equals(config.getQueueName());
     return jobConfigurationService.getAllJobConfigurations().stream()
         .filter(JobConfiguration::isConfigurable)
-        .filter(not(JobConfiguration::isLeaderOnlyJob))
```
jbee, line 109: FYI: This logic was backwards but since user created job configurations before this change were created with `leaderOnlyJob=false` (which was wrong) this had to filter like this. But then I figured this entire filter does not make sense to have given we already filter  `isConfigurable` which means only user created jobs will be possible to select.

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/scheduling/SchedulingManagerTest.java
```diff
@@ -109,6 +109,9 @@ void setUp() {
     when(cacheProvider.createRunningJobsInfoCache()).thenReturn(new TestCache<>());
     when(cacheProvider.createCompletedJobsInfoCache()).thenReturn(new TestCache<>());
 
+    LeaderManager leaderManager = mock(LeaderManager.class);
+    when(leaderManager.isLeader()).thenReturn(true);
```
jbee, line 113: FYI: This illustrates quite well the change of user created job now being "leader only" as well. So to correct the test setup for this we now need to pose as leader to have them executed.

netroms (approved):

david-mackessy (approved):

PR #14581, opened 2023-07-13
feat: Add new 'ieq' operator for case insensitive exact text filter [DHIS2-12167]

david-mackessy:
# Summary
New feature request to add new operator `ieq` which performs a case insensitive exact string match

## Change
New operator `ieq` added which allows filtering for case insensitive exact string match.
- Implemented using ilike exact match
- only works for string type

Some Sonar changes implemented also
- merge switch cases
- simplify operator used (use `<` instead of `!>=`)

Scenario
`Program` with name `My Program`
| Filter query | Match |
| ----------- | -------- |
| filter=name:ieq:my program | true |
| filter=name:ieq:MY ProGram | true |
| filter=name:ieq:MY PROGRAM | true |
| filter=name:ieq:mmy program | false |
| filter=name:ieq:my programm | false |
| filter=name:ieq:mmy programm | false |

# Testing
## Automated
Tests added to check
- filtering for exact match
  - all lowercase
  - all uppercase
  - mixed case
- extra char at start
- extra char at end
- calling base endpoints e.g. `/organisationUnits?filter=name:ieq:l0`
- calling metadata endpoint e.g. `/api/metadata?programs=true&filter=name:ieq:test program`
- query params parsed correctly
- error when performed on non string property

## Manual
- check existing `dataElement`s with `GET /api/dataElements`
- filter by using new operator on any existing dataElement `GET /api/dataElements?filter=name:ieq:{data element name}`
  - passing in the data element name with different cases should not affect the response
  - the use of upper or lowercase will not be taken into consideration

Testing on non string values should return an error
- `GET /api/metadata?programs=true&filter=version:ieq:5`
```json
{
    "httpStatus": "Conflict",
    "httpStatusCode": 409,
    "status": "ERROR",
    "message": "Value `5` of type `Integer` is not supported by this operator."
}
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/QueryUtils.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/QueryUtils.java
@@ -55,6 +55,7 @@ import org.hisp.dhis.webapi.controller.event.webrequest.OrderCriteria;
 public final class QueryUtils {
 
   private static final String UNABLE_TO_PARSE = "Unable to parse `";
+
   private QueryUtils() {
     throw new UnsupportedOperationException("util");
   }
@@ -147,7 +148,7 @@ public final class QueryUtils {
     }
 
     throw new QueryParserException(
-            UNABLE_TO_PARSE + value + "` to `" + klass.getSimpleName() + "`.");
+        UNABLE_TO_PARSE + value + "` to `" + klass.getSimpleName() + "`.");
   }
 
   /**
@@ -159,14 +160,14 @@ public final class QueryUtils {
   @SuppressWarnings({"unchecked", "rawtypes"})
   public static <T> T getEnumValue(Class<?> klass, String value) {
     Optional<? extends Enum<?>> enumValue =
-            Enums.getIfPresent((Class<? extends Enum>) klass, value).toJavaUtil();
+        Enums.getIfPresent((Class<? extends Enum>) klass, value).toJavaUtil();
 
     if (enumValue.isPresent()) {
       return (T) enumValue.get();
     }
     Object[] possibleValues = klass.getEnumConstants();
     throw new QueryParserException(
-            UNABLE_TO_PARSE
+        UNABLE_TO_PARSE
             + value
             + "` as `"
             + klass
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/QueryUtils.java
```diff
@@ -54,6 +53,9 @@
  * @author Morten Olav Hansen <mortenoh@gmail.com>
  */
 public final class QueryUtils {
+
+  private static final String UNABLE_TO_PARSE = "Unable to parse `";
```
jbee, line 57: This is bordering on making the code less readable. It is good to use constants for "magic numbers" and similar working strings but when it comes to error messages, fields names and such repetition isn't necessarily the issue. We tend to keep inline strings for such as it helps to see what the code is doing or producing without having to jump somewhere else and then put it together in you mind. 
I think in this instance it is fine to use a constant just wanted to give you the gist of how we mostly do things even if tools might warn about duplicates.

jbee (commented):
Looks good!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/QueryUtils.java
david-mackessy: could easily argue that it's more readable & reduces mental processing required
- with the constant, hovering over it will show its value (no need to jump anywhere)
- because it's used in many other spots, and because you now know the value, you don't now have to manually read each string to make sure they all match.
- without the constant, you have to read every character in every string to know it's value

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/QueryUtils.java
jbee: I meant to click approve :D Yes in this case constant work. In cases where json is composed or multiple string parts are combined, like SQL building or complex error message building you cannot hover all at the same time, and even if you would need you mental capacity to substitute the constants with their values instead of using it to reason about the entire value put together.

vietnguyen (approved):

PR #14592, opened 2023-07-14
chore: Hide pagination in service [TECH-1591]

enricocolasante:
The goal of this PR is to harmonize `TrackedEntityController` with the controllers of other entities, it is not to be intended as a solution for pagination as it will be tackled in some other epic.

The idea is to make the service return a wrapping object that holds the info about the list of tracked entities and the pagination information.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntities.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntities.java
@@ -27,31 +27,23 @@
  */
 package org.hisp.dhis.tracker.export.trackedentity;
 
-import org.hisp.dhis.feedback.BadRequestException;
-import org.hisp.dhis.feedback.ForbiddenException;
-import org.hisp.dhis.feedback.NotFoundException;
+import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntities.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.RequiredArgsConstructor;
+import org.hisp.dhis.common.Pager;
+import org.hisp.dhis.trackedentity.TrackedEntity;
+
+@RequiredArgsConstructor(staticName = "of")
+@Getter
+@EqualsAndHashCode
+public class TrackedEntities {
+
+  public static final TrackedEntities EMPTY = new TrackedEntities(List.of(), null);
```
teleivo, line 42: If the goal is to align with enrollments/events can you please keep it closer to https://github.com/dhis2/dhis2-core/blob/a7796d77149c8e5ff8093dea17f3eaed80f2a4e3/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/Enrollments.java#L42

or change the other 2 entity wrappers 😄 For example constructor vs factory, consistent field name. I would stick to `trackedEntities` instead of `trackedEntityList` even if sonar does not like it.

So that we will have an easier time refactoring them.

teleivo (approved):

lucaCambi77 (approved):

PR #14613, opened 2023-07-17

enricocolasante:
Relationship is the missing entity controller to harmonize and it is the most different from the others, all the validations and the mapping is happening in the controller.
I will need to create some intermediate PRs to avoid creating just a huge one, so some parts of the code will not be completed at some stage.

**What it is in this PR**
- Create a `RelationshipRequestMapper` to map `RequestParams` to `RelationshipOperationParams` and validate input without accessing the DB
- Create a placeholder for `RelationshipOperationParams`. The shape of this param object will become clear in next PRs
- Temporary mapping `RelationshipOperationParams` to `LegacyRequestParams`.

**What is next**


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParams.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private boolean skipPaging;
+
+  private String trackedEntity;
+
+  private String enrollment;
+
+  private String event;
+
+  // -------------------------------------------------------------------------
```
teleivo, line 58: I don't find such comments useful IMHO. I am in favor of not adding + removing such separators.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
```diff
@@ -0,0 +1,83 @@
...
+ * LegacyRequestParams} to {@link RelationshipOperationParams} which is used to fetch events from
+ * the DB.
+ */
+@Component
+@RequiredArgsConstructor
+class RelationshipRequestParamsMapper {
+
+  public RelationshipOperationParams map(RequestParams requestParams) throws BadRequestException {
+    UID trackedEntity =
+        validateDeprecatedUidParameter(
```
teleivo, line 50: I think you need to rebase on latest master where this is just called `validateDeprecatedParameter` 🤔

PR #14626, opened 2023-07-18
feat: support JPA annotation mapping for object model

vietnguyen:
https://dhis2.atlassian.net/browse/TECH-1517

### Scope

As [mentioned here ](https://github.com/hibernate/hibernate-orm/blob/6.0/migration-guide.adoc#deprecation-of-hbmxml-mappings) , hbm.xml mapping format is considered deprecated and will no longer supported beyond 6.x. 

**The goal of this PR is to support both Hibernate xml and JPA annotation mapping for object model in DHIS2. It doesn't upgrade Hibernate to version 6.x** 

### Tasks

- [x] Make sure `org.hibernate.Criteria` is completely replaced by JPA query api.
- [x] Add javax.persistence.EntityManagerFactory  bean to HibernateConfig so all DAO classes can use EntityManager instead of hibernate SessionFactory.
- [x] Support both hibernate *.hbm.xml mapping files and JPA annotation mapping.  Will slowly migrate all hbm files to annotation.
- [x] Replace `SessionFactory` with `EntityManager`.
- [x] Find a way to inject `EntityManager` through constructor instead of field injection with `@PersistenceContext`, this is needed for unit tests with Mockito. Reference https://github.com/spring-projects/spring-data-jpa/commit/7b5425322b73314efd836fd8616a09e8d24c0c1b

### Next PRs
- [ ] Turn off property `hibernate.allow_update_outside_transaction` as JPA doesn't allow calling `session.update` or `session.flush` outside of a transaction. This requires refactoring in places where we explicitly call `session.update` or `session.flush` without `@Transactional` wrapper.
- [ ] `session.flush()` inside a transaction should be carefully tested in metadata import service. May need to be removed and let the objects to be managed by JPA PersistenContext.
- [ ] Split the `BaseIdentifiableObject` into multiple base objects in order to support entity inheritance mapping for only required properties. Such as `BaseShareableIdObject`, `BaseIdObject`, `BaseAttributeObject`, etc..
- [ ] Replace `save`/`update` with `persist`/`merge`


dhis-2/dhis-api/src/main/java/org/hisp/dhis/reservedvalue/ReservedValueStore.java
```diff
@@ -42,8 +42,6 @@ public interface ReservedValueStore extends GenericStore<ReservedValue> {
   List<ReservedValue> getAvailableValues(
       ReservedValue reservedValue, List<String> values, String ownerObject);
 
-  List<ReservedValue> reserveValuesJpa(ReservedValue reservedValue, List<String> values);
-
```
vietnguyen, line 46: This is a code clean up done by Luca from tracker team. He helped to fix one of the unit test failed after using entityManager.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/reservedvalue/hibernate/HibernateReservedValueStore.java
```diff
@@ -71,11 +70,14 @@ public HibernateReservedValueStore(
   @Override
   public List<ReservedValue> getAvailableValues(
       ReservedValue reservedValue, List<String> values, String ownerObject) {
-    List<String> availableValues = getIfAvailable(reservedValue, values, ownerObject);
+    if (isEmpty(values) || !reservedValue.getOwnerObject().equals(ownerObject)) {
+      return List.of();
+    }
+    List<String> availableValues = getIfAvailable(reservedValue, values);
```
vietnguyen, line 76: code clean up done by Luca.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/TranslationWebApiTest.java
```diff
@@ -1,104 +0,0 @@
...
-        mvc.perform(
-                get("/dataElements/" + dataElementA.getUid() + "?locale=" + locale.getLanguage())
-                    .session(session)
-                    .contentType(TestUtils.APPLICATION_JSON_UTF8))
-            .andReturn();
-    ObjectMapper mapper = new ObjectMapper();
-    JsonNode node = mapper.readTree(result.getResponse().getContentAsString());
-    assertEquals(valueToCheck, node.get("displayName").asText());
-  }
-}
```
vietnguyen, line 104: Remove this test class as it is obsolete.  The translations api are covered in AbstractCrudControllerTest.replaceTranslationsOk() and some other tests in that class.

dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/interceptor/ContextInterceptor.java
```diff
@@ -49,11 +53,7 @@ public class ContextInterceptor implements Interceptor {
 
   private static final String KEY_CURRENT_KEY = "keyCurrentKey";
 
-  private DatabaseInfoProvider databaseInfoProvider;
-
-  public void setDatabaseInfoProvider(DatabaseInfoProvider databaseInfoProvider) {
-    this.databaseInfoProvider = databaseInfoProvider;
-  }
+  private final DatabaseInfoProvider databaseInfoProvider;
```
vietnguyen, line 56: This is to fix a circular dependency between DataBaseInfoProvider and DhisConfigurationProvider

jbee (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/DefaultObjectBundleService.java
```diff
@@ -78,7 +79,7 @@ public class DefaultObjectBundleService implements ObjectBundleService {
   private final CurrentUserService currentUserService;
   private final PreheatService preheatService;
   private final SchemaService schemaService;
-  private final SessionFactory sessionFactory;
+  @PersistenceContext private EntityManager entityManager;
```
david-mackessy, line 82: does this get instantiated correctly given that it's not marked as `final`? Which is what the `RequiredArgsConstructor` uses for the class constructor.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/HibernateConfig.java
```diff
@@ -67,62 +81,121 @@ public HibernateConfigurationProvider hibernateConfigurationProvider(
...
+
+  /**
+   * Loads all the hibernate mapping files from the classpath
+   *
+   * @return Array of Strings representing the mapping files
+   */
+  private String[] loadResources() {
+    try {
+      PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+      Resource[] resources = resolver.getResources("classpath*:org/hisp/dhis/**/*.hbm.xml");
```
david-mackessy, line 179: Could this file check be more targeted? It looks like it's going to look through every package on the classpath when we know exactly where all `hbm.xml` files are located.

david-mackessy (approved):
nice work Viet

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/DefaultObjectBundleService.java
vietnguyen: nice catch. This was when i haven't found away to inject using constructor. Fixed now.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/HibernateConfig.java
vietnguyen: Not sure how to do this. Will check later and update on next PR.

PR #14628, opened 2023-07-18
fix: add timestamp to Event's executionDate, created, lastUpdated and dueDate

zubaira:
[DHIS2-14836](https://dhis2.atlassian.net/browse/DHIS2-14836)

[DHIS2-14836]: https://dhis2.atlassian.net/browse/DHIS2-14836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -101,6 +103,9 @@ class EventExporterTest extends TrackerTest {
 
   private TrackedEntity trackedEntity;
 
+  private String occurredAtTimeStamp = "2019-01-25T12:10:38.100";
```
teleivo, line 106: I would inline them in the assertion as these values are only used in one test. Also in here its hard to know to what entity they belong to as there is no context.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -245,6 +250,54 @@ void testExportEventsWithLastUpdateDates() throws ForbiddenException, BadRequest
...
+  @Test
+  void testExportEventsWithDatesIncludingTimeStamp()
+      throws ForbiddenException, BadRequestException {
+    Date date = new Date();
+    EventOperationParams params =
+        EventOperationParams.builder()
+            .orgUnitUid(orgUnit.getUid())
+            .enrollments(Set.of("nxP7UnKhomJ"))
+            .programStageUid(programStage.getUid())
+            .events(Set.of("pTzf9KYMk72"))
```
teleivo, line 262: If you are only after getting the `event` `pTzf9KYMk72` I would set `.events(Set.of("pTzf9KYMk72"))` on the builder. Setting more then that makes me question whether you are testing querying by dates but I don't think these are correct 
 The `pTzf9KYMk72` `"updatedAt": "2019-01-28T12:10:38.109"` which is not within the range you specify 🤔

teleivo (commented):
The issue says

| When sending an event to the server with eventDate, the date stored is different without hours minutes and seconds.

I think that is incorrect, at least from looking at your changes. We store it with time but only export the fields as dates instead of datetime. Right?

zubaira:
We fetch date and that is why timestamp is 00:00:00.

teleivo:
> We fetch date and that is why timestamp is 00:00:00.

right, so the issue description 

> When sending an event to the server with eventDate, the date stored is different without hours minutes and seconds.

is incorrect 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -354,19 +354,19 @@ public List<Event> getEvents(
             event.setAttributeOptionCombo(coc);
 
             event.setStoredBy(resultSet.getString("psi_storedby"));
-            event.setDueDate(resultSet.getDate("psi_duedate"));
-            event.setExecutionDate(resultSet.getDate("psi_executiondate"));
-            event.setCreated(resultSet.getDate("psi_created"));
```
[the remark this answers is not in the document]
zubaira: well if you see the comments. I have written down all the end points where are correcting fetching dates with timestamp. I will also take a look at legacy tracker endpoints to make sure behavior is consistent.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
zubaira: resolved

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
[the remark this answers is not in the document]
zubaira: @teleivo
to my understanding `eventDate` is when event actually happened/occurred. Now there are two possibilities 
1. event happened right at the time of enrollment
2. event happened sometime before or after the enrollment ( in this case recording timestamp would be difficult )

in the second case, the event will be stored without a timestamp, so `eventDate` does not necessarily have a timestamp stored with it.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -317,6 +321,52 @@ void testExportEventsWithLastUpdateDates() throws ForbiddenException, BadRequest
...
+        () ->
+            assertEquals(
+                "2019-01-25T12:10:38.100",
+                DateUtils.getIso8601NoTz(event.getExecutionDate()),
+                () ->
+                    String.format(
+                        "Expected %s to be in %s",
+                        event.getExecutionDate(), "2019-01-25T12:10:38.100")),
+        () ->
+            assertEquals(
```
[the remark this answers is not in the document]
zubaira: setting the same date and timestamp is not something that is being tested here. The test is to make sure 
the date is fetched with the timestamp and also making sure it's not 00:00:00.

`2019-01-25T12:00:00:00` This is also the date with the timestamp but the value of the timestamp is being lost. That is exactly what is being tested here. Its not just matching of JSON fixture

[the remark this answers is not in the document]
teleivo: You are right you did not set the timestamps of `occurredAt` and `scheduledAt` to the same value. I misread that 😅 

Can you please also assert that the `completedDate` has a timestamp set.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -317,6 +322,58 @@ void testExportEventsWithLastUpdateDates() throws ForbiddenException, BadRequest
...
+                "2019-01-28T12:32:38.100",
+                DateUtils.getIso8601NoTz(event.getDueDate()),
+                () ->
+                    String.format(
+                        "Expected %s to be in %s", event.getDueDate(), "2019-01-28T12:32:38.100")),
+        () ->
```
teleivo, line 369: Why did you not assert  `completedDate` in the same way you asserted the other two fields? So

```suggestion
        () ->
            assertEquals(
                "2019-01-25T12:10:38.100",
                DateUtils.getIso8601NoTz(event.getCompletedDate()),
                () ->
                    String.format(
                        "Expected %s to be in %s",
                        event.getCompletedDate(), "2019-01-28T22:17:43.000")),
```

And adding a time to the `dhis-2/dhis-test-integration/src/test/resources/tracker/event_and_enrollment.json` `"completedAt": "2019-01-28T22:17:43.000"` 😅

zubaira: > Why did you not assert `completedDate` in the same way you asserted the other two fields? So
> 
> And adding a time to the `dhis-2/dhis-test-integration/src/test/resources/tracker/event_and_enrollment.json` `"completedAt": "2019-01-28T22:17:43.000"` 😅

https://github.com/dhis2/dhis2-core/pull/14628#discussion_r1272153931

zubaira: checking if it has a timestamp and checking if it has exactly the same timestamp are two different things

zubaira: moreover, it sets `completedDate` when it processes this event so an exact comparison would be difficult in this case.

teleivo: What is setting the `completedDate`?

Please remove any timestamps that are automatically set from the JSON test fixture. Otherwise, its really hard to review the PR and understand the test in the future 😅

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/DateUtils.java
```diff
@@ -562,6 +562,24 @@ public static boolean dateTimeIsValid(final String dateTimeString) {
     }
   }
 
+  public static boolean hasTimeStamp(final String dateTimeString) {
```
teleivo, line 565: Since this is only ever used in tests I would rather write these as an assertion like `assertHasTimestamp` as we do in https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java

zubaira: well, I have just included and so it can be used in the future. It can be added in assertion separately but having it in DateUtils perfectly make sense to me.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
zubaira: the service layer is setting `completedDate` at the time of changing the event status to `COMPLETED`

zubaira: having a timestamp in completedDate is important and this is exactly what I am checking in the test using `hasTimeStamp`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
```diff
@@ -293,6 +303,23 @@ public static void assertSlimPager(int pageNumber, int pageSize, boolean isLast,
...
 
+  public static void assertHasTimeStamp(Date date) {
+    assertTrue(
+        hasTimeStamp(date),
+        String.format("Supported format is %s but found %s", DATE_WITH_TIMESTAMP_PATTERN, date));
+  }
+
+  private static boolean hasTimeStamp(Date date) {
+    try {
+
```
teleivo, line 314: nit: remove this newline

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/DateUtils.java
teleivo: I thought we agreed on removing this one, right? 😅 as we are not using this in our production code

zubaira: yes.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
zubaira: shouldnt it be taken care by our formatter ?

teleivo (approved):
Nice 🥳

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/Assertions.java
teleivo: it seems to be ok with some extra newlines 😁

stian-sandvold (approved):

PR #14629, opened 2023-07-18

enricocolasante:
Continuing from https://github.com/dhis2/dhis2-core/pull/14613

**What it is in this PR**
- Completely remove `LegacyRequestParams`
- Use `TrackerType` instead of `Class<?>` to identify the type of entity

**What is next**


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
```diff
@@ -28,56 +28,73 @@
...
- * the DB.
+ * Maps operation parameters from {@link RelationshipsExportController} stored in {@link
+ * RequestParams} to {@link RelationshipOperationParams} which is used to fetch relationships from
+ * the service.
  */
 @Component
 @RequiredArgsConstructor
 class RelationshipRequestParamsMapper {
 
+  private static boolean hasMoreThanOneParamNotNull(UID trackedEntity, UID enrollment, UID event) {
```
teleivo, line 55: Since its not related to params we can make it more generic like 

```suggestion
  private static boolean hasMoreThanOneNonNull(Object... values) {
```

I would put this function to the bottom as `map` should be the highlight of this class 😁

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParams.java
```diff
@@ -31,6 +31,7 @@
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
+import org.hisp.dhis.tracker.imports.TrackerType;
```
teleivo, line 34: Are we adding a dependency from export to import or did it already exist? I mean the Class might also do as a discriminator. I am just worried we open the gates before thinking more about structure.

Not sure `TrackerType` is worth that, just yet. At least from the changes I am not sure 🤔

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
```diff
@@ -28,56 +28,73 @@
...
+                    trackedEntity, requestParams.getEnrollment(), requestParams.getEvent())
+                .getValue())
         .page(requestParams.getPage())
         .pageSize(requestParams.getPageSize())
         .totalPages(requestParams.isTotalPages())
         .skipPaging(toBooleanDefaultIfNull(requestParams.isSkipPaging(), false))
         .build();
   }
+
+  private TrackerType getTrackerType(UID trackedEntity, UID enrollment, UID event) {
```
teleivo, line 90: Do we have something like this somewhere? Or is the `UID` type the difference. This looks too familiar 😁

enricocolasante: This looks familiar because we are doing something similar in some validator, but there we are extracting the information from `RelationshipItem`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParams.java
enricocolasante: I saw that and I actually forgot to move it outside of `imports` and under `tracker` package.
For me, `TrackerType` should be part of `tracker` package, it is not an import or export concept

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
enricocolasante: Formatter is moving static methods at the top of the class.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
```diff
@@ -129,6 +130,8 @@ public class TrackerBundle {
   private Map<TrackerType, Map<String, TrackerImportStrategy>> resolvedStrategyMap =
       initStrategyMap();
 
+  @Builder.Default @JsonIgnore private Set<String> updatedTeis = new HashSet<>();
```
teleivo, line 133: its odd that this is happening 😬 we did apply the formatting on the entire codebase, so why does it do that now for you?

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
enricocolasante: Could it be because of this?
https://github.com/dhis2/dhis2-core/pull/14624

teleivo: I don't think so because if I run `mvn spotless:apply -f ./dhis-2/pom.xml` there are no changes. Are you formatting via the IDE? If so how (is it configured)?

enricocolasante: I was formatting the whole file in the `Action on Save `. Changing that to `Change lines` solved the issue

lucaCambi77 (approved):

PR #14630, opened 2023-07-18
feat: Allow SQL Views read or write option [DHIS2-15108]

david-mackessy:
# Summary
- pre `v2.40`, `write`s were allowed in SQL Views
  - `write`s in this context means the underlying query implementation
    - e.g. A `User` adds a new SQL View with the query `select createatable();`
    - this query is acceptable and passes the SQL View query validation (seen as a select, `read`)
    - the query calls a function in the DB, which in turn creates a temp table (`write`s required)
- `v2.40` As part of [DHIS2-14718], a `readOnly = true` property was added to the Transactions for SQL Views
  - this meant that any SQL Views that were using writes were now not supported

## Change
- Decision made to keep SQL Views feature enabled in `read only` mode by default after discussions with @michael-markevich and @jbee
  - no impact to existing users using SQL Views with `read`s
  - new configuration key `SYSTEM_SQL_VIEW_WRITE_ENABLED` used to determine which service method should be used
    - default is `write`s disabled
    - when set to `on`, `write`s allowed
    - this applies an extra layer of security if SQL Views require `write`s (compared to pre `v2.40`)
    - update `dhis.conf` with the following setting to enable
      - `system.sql_view_write_enabled=on`

# Testing
## Automated
- tests added checking the correct control flow based on new config property value
  - underlying logic not changed

Tried to add an integration test with a new base test class without `@Transactional` but it started getting messy with sessions etc. (existing base class wraps test in `write` transaction).
End-to-end test looked too complicated to add to & an existing DB function would be required as part of the test.

## Manual
Tested locally by
- adding function `createatable` to database - steps detailed in Jira
- add SQL view
- `POST /api/sqlViews`
```json
{
  "sqlQuery": "select createatable();",
  "description": "SQL View requiring write permission",
  "type": "VIEW",
  "name": "view requiring write permission",
  "cacheStrategy": "NO_CACHE"
}
```
- execute view
  - `POST /api/sqlViews/{id}/execute`

- get SQL View data
  - `GET /api/sqlViews/{id}/data`

- successful response 
```json
{
    "pager": {
        "page": 1,
        "total": 1,
        "pageSize": 50,
        "pageCount": 1
    },
    "listGrid": {
        "title": "write required view",
        "subtitle": "write required view",
        "headers": [
            {
                "name": "createatable",
                "column": "createatable",
                "type": "java.lang.String",
                "hidden": false,
                "meta": false
            }
        ],
        "metaData": {},
        "rowContext": {},
        "height": 1,
        "rows": [
            [
                "1"
            ]
        ],
        "width": 1,
        "headerWidth": 1
    }
}
```
- failed response (read only mode but SQL View needs write permission)
```json
{
    "httpStatus": "Internal Server Error",
    "httpStatusCode": 500,
    "status": "ERROR",
    "message": "ERROR: cannot execute CREATE TABLE in a read-only transaction\n  Where: SQL statement \"create temporary table temp1\n  (value integer)\n  on commit drop\"\nPL/pgSQL function createatable() line 3 at SQL statement"
}
```

[DHIS2-14718]: https://dhis2.atlassian.net/browse/DHIS2-14718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbee (approved):
Looks good!

vietnguyen (approved):

PR #14035, opened 2023-05-25
fix: use correct name for createdAt column [DHIS2-15325]

zubaira:

PR #14631, opened 2023-07-18
fix: /tracker/trackedEntities?order=createdAt DHIS2-15325 DHIS2-15325

teleivo:
which was ordering by tracked entity id instead of created column.

- Lowercase column names in OrderColumn mapper as we use lowercase for our SQL.
- Fix tests which asserted using the wrong order or assumed we can set the lastupdated date. lastupdated is automatically set by HibernateGenericStore. So to assert we can order by that column we can only update the entity in some order and expect that to be returned

https://dhis2.atlassian.net/browse/DHIS2-15325
(replaces https://github.com/dhis2/dhis2-core/pull/14035)

More bugs of that sort might exist. We are looking through some issues in https://docs.google.com/document/d/1Y5GNNoRKaNv3ObiDJ2rUKyTSJ0IFNYWZFzYWJofJeWo/edit

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityQueryParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityQueryParams.java
@@ -973,14 +973,15 @@ public class TrackedEntityQueryParams {
         "trackedEntity",
         "uid",
         MAIN_QUERY_ALIAS), // Ordering by id is the same as ordering by created date
-    CREATED(CREATED_ID, "trackedentityinstanceid", MAIN_QUERY_ALIAS),
-    CREATED_AT("createdAt", "trackedentityinstanceid", MAIN_QUERY_ALIAS),
-    CREATED_AT_CLIENT("createdAtClient", "createdAtClient", MAIN_QUERY_ALIAS),
-    UPDATED_AT("updatedAt", "lastUpdated", MAIN_QUERY_ALIAS),
-    UPDATED_AT_CLIENT("updatedAtClient", "lastUpdatedAtClient", MAIN_QUERY_ALIAS),
+    // TODO(tracker): remove with old tracker
+    CREATED("created", CREATED_ID, MAIN_QUERY_ALIAS),
+    CREATED_AT("createdAt", CREATED_ID, MAIN_QUERY_ALIAS),
+    CREATED_AT_CLIENT("createdAtClient", "createdatclient", MAIN_QUERY_ALIAS),
+    UPDATED_AT("updatedAt", "lastupdated", MAIN_QUERY_ALIAS),
+    UPDATED_AT_CLIENT("updatedAtClient", "lastupdatedatclient", MAIN_QUERY_ALIAS),
     ENROLLED_AT(
         "enrolledAt",
-        "enrollmentDate",
+        "enrollmentdate",
         PROGRAM_INSTANCE_ALIAS), // this works only for the new endpoint
     // ORGUNIT_NAME( "orgUnitName", MAIN_QUERY_ALIAS+".organisationUnit.name" ),
     INACTIVE(INACTIVE_ID, "inactive", MAIN_QUERY_ALIAS);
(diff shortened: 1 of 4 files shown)


teleivo:
@enricocolasante @muilpp @zubaira please squint hard on the dates/order so we get it right 😁

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -973,14 +973,14 @@ public enum OrderColumn {
         "trackedEntity",
         "uid",
         MAIN_QUERY_ALIAS), // Ordering by id is the same as ordering by created date
-    CREATED(CREATED_ID, "trackedentityinstanceid", MAIN_QUERY_ALIAS),
```
teleivo, line 976: @enricocolasante I realized that this class might also be used by old tracker 🤦🏻 I assume that `order=created:asc` in old tracker should be equivalent to `order=createdAt:asc` in new tracker, right?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -973,14 +973,14 @@ public enum OrderColumn {
...
-    CREATED_AT("createdAt", "trackedentityinstanceid", MAIN_QUERY_ALIAS),
+    CREATED("created", "trackedentityinstanceid", MAIN_QUERY_ALIAS),
+    CREATED_AT("createdAt", CREATED_ID, MAIN_QUERY_ALIAS),
     CREATED_AT_CLIENT("createdAtClient", "createdAtClient", MAIN_QUERY_ALIAS),
```
enricocolasante, line 978:
```suggestion
    CREATED_AT_CLIENT("createdAtClient", "createdatclient", MAIN_QUERY_ALIAS),
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityQueryParams.java
enricocolasante: That is true.
I don't understand where is the bug and what is the fix though.
I am missing something
We are just changing the order column from `trackedentityinstanceid` to `created`, right?
Why was not working before?
Why should be working now?

teleivo: https://github.com/dhis2/dhis2-core/pull/14631/files#diff-6e6269e3c03b721125e0fee07a0b409b9aa0808e106a27f541c2c235ccef4a0aR977 is the bug

`order=createdAt` is mapped to SQL `order by trackedentityinstanceid` instead of `order by created`

Now the bug was originally only reported for new tracker by @zubaira but it also affects old tracker where 

`order=created` is mapped to SQL `order by trackedentityinstanceid` instead of `order by created`

> We are just changing the order column from trackedentityinstanceid to created, right?

No, its the other way round. `CREATED(CREATED_ID, "trackedentityinstanceid", MAIN_QUERY_ALIAS),` is order property, SQL column. Using the `CREATED_ID` which is supposed to represent the column name might have mislead you into thinking that 🤷🏻 

Does that answer your questions?

I will add a fix for old tracker in this PR as well.

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -335,29 +335,164 @@ void shouldOrderEntitiesByIdWhenParamCreatedAtSupplied() {
...
+    TrackedEntityQueryParams params = new TrackedEntityQueryParams();
+
+    params.setOrgUnits(Set.of(organisationUnit));
+    params.setOrders(List.of(new OrderParam("created", SortDirection.DESC)));
 
     List<Long> teiIdList = entityInstanceService.getTrackedEntityIds(params, true, true);
 
     assertEquals(
         List.of(
+            entityInstanceD1.getId(),
```
zubaira, line 370: just a comment.
I think readability can be improved if we store ascOrdered list of tei ids and then use it in the assertion 
`assertEquals( ascOrdered, teiIdList )`

zubaira (approved):

PR #14633, opened 2023-07-18
refactor: Group common attrs in Event/Vis. [DHIS2-13618]

maikelarabori:
This PR moves common duplicated attributes and methods to the parent class in order to avoid all duplication.


gnespolino (approved):
Good one ✂️

d-bernat (approved):

PR #14635, opened 2023-07-18
feat: support @OpenApi.Property on classes to eliminate need of havin…

claudemamo:
…g to annotate each field


claudemamo:
> I think i implemented this already like this: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/Property.java#L94-L96

`org.hisp.dhis.webapi.openapi.PropertyTest` fails when I remove my changes: the type remains `AProperty` instead of `String`.

jbee (approved):

PR #14664, opened 2023-07-20
fix: Harmonize ACL depending on org unit mode [DHIS-15394] [2.40]

muilpp:
Back port of: https://github.com/dhis2/dhis2-core/pull/14402


dhis-2/dhis-test-e2e/README.md
```diff
@@ -23,6 +32,25 @@ mvn test \
...
+Run Tomcat and when it's up and running, run the e2e test(s) you are interested in.
+
+#### Troubleshooting
+
+You might get errors while importing data during the test setup, it's quite possibly related to the
+database you are using.
+
+Until we find a better solution, and only if the error code is `E5002`, you can skip the reference
+check in `ReferencesCheck.checkReference` by replacing the following
+condition `if (!(isUserReference && skipSharing))` to `if (false)`
```
enricocolasante, line 52: I wouldn't advise this in a README file.
If we put it here, it will stay here for ever.

teleivo: I run tests against the dockerized setup so the DB is empty. Is the issue not that you might have data in your DB that is causing the problems? So the suggestion should rather be that you use a clean DB which is generally a good idea when running tests.

dhis-2/dhis-test-e2e/README.md
```diff
@@ -23,6 +32,16 @@ mvn test \
     -Dtest.track_called_endpoints=true
 ```
 
+### From IntelliJ
+
+Make sure the property instance.url in `dhis-2/dhis-test-e2e/src/main/resources/config.properties`
```
teleivo, line 37: This is independent of tomcat 😄 `instance.url` needs to point to DHIS2.

Maybe we need a higher level description at the start of how the e2e tests work. As there are 2 independent aspects 😅 
* how you run DHIS2 (docker, tomcat, embedded jetty?)
* how you run the e2e tests (from intellij, from CLI using maven, from within docker)

muilpp: I'll open a new PR for the readme, thanks!

enricocolasante (approved):

PR #14674, opened 2023-07-21
fix: use an index variable to set sortOrder

vietnguyen:
Continue on https://github.com/dhis2/dhis2-core/pull/14565. Use an index variable to set the sortOrder instead of the for loop index.


jbee (commented):
This wasn't what I had in mind when I commented on the previous PR. 

I make an example. If the items are given as this sequence (with sort order) `[2,5,-,8,-,9]` then the loop would make use of the sort order that is defined, so the outcome is `[2,5,6,8,9,9]` (it uses the previous sort order value +1). This way this would maintain the sequence order given except when the gap is not allowing for it, like in the case of `9` where now the order might swap or not.  

But I am also not sure this really is much of an improvement. Maybe it would be even better to always use bigger sort orders than those already occurring? Again, `[2,5,-,8,-,9]` would then become `[2,5,10,8,11,9]`. This would make sure items with a sort order stay in that order, items without stay in order in relation to each other but are always after the items that had a sort order value.

stian-sandvold (approved):

vietnguyen:
Close this PR for now, will revisit after done with jpa annotation.

PR #14677, opened 2023-07-21

enricocolasante:
Continuing from https://github.com/dhis2/dhis2-core/pull/14629

In this PR we moved all the service logic that was present in the Controller into the service. In the process we removed the direct dependency on `dhis-service-core` services and we are instead using `dhis-service-tracker` services.
This added ACL capabilities and it lead to set the `id` for entities in order to continue to use correctly `RelationshipStore` methods.

**What it is in this PR**

**What it is NOT in this PR**
- Any kind of ACL fixes. The code was moved in order to maintain the logic of permissions as it was before.
- Any change to the repository layer.

**What is next**

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -102,6 +102,7 @@ public class DefaultEnrollmentService
     }
 
     Enrollment result = new Enrollment();
+    result.setId(enrollment.getId());
     result.setUid(enrollment.getUid());
 
     if (enrollment.getTrackedEntity() != null) {
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
@@ -27,17 +27,15 @@
  */
 package org.hisp.dhis.tracker.export.relationship;
 
-import static org.hisp.dhis.tracker.export.relationship.RelationshipQueryParams.EMPTY;
-
 import lombok.RequiredArgsConstructor;
-import org.hisp.dhis.feedback.ForbiddenException;
+import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.feedback.NotFoundException;
-import org.hisp.dhis.tracker.export.enrollment.EnrollmentParams;
-import org.hisp.dhis.tracker.export.enrollment.EnrollmentService;
-import org.hisp.dhis.tracker.export.event.EventParams;
-import org.hisp.dhis.tracker.export.event.EventService;
-import org.hisp.dhis.tracker.export.trackedentity.TrackedEntityParams;
-import org.hisp.dhis.tracker.export.trackedentity.TrackedEntityService;
+import org.hisp.dhis.program.Enrollment;
+import org.hisp.dhis.program.EnrollmentService;
+import org.hisp.dhis.program.Event;
+import org.hisp.dhis.program.EventService;
+import org.hisp.dhis.trackedentity.TrackedEntity;
+import org.hisp.dhis.trackedentity.TrackedEntityService;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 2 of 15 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -102,6 +102,7 @@ public Enrollment getEnrollment(
     }
 
     Enrollment result = new Enrollment();
+    result.setId(enrollment.getId());
```
teleivo, line 105: Why do we need to set these?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -0,0 +1,74 @@
...
+  private final EnrollmentService enrollmentService;
+
+  private final EventService eventService;
+
+  @Transactional(readOnly = true)
+  public RelationshipQueryParams map(RelationshipOperationParams params) throws NotFoundException {
+    try {
+      return new RelationshipQueryParams(
+          switch (params.getType()) {
+            case TRACKED_ENTITY -> trackedEntityService.getTrackedEntity(
```
teleivo, line 63: I find it odd that we fetch in the mapper, we do not do that in the other entities, right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -0,0 +1,74 @@
...
+                params.getIdentifier(), TrackedEntityParams.TRUE, true);
+            case ENROLLMENT -> enrollmentService.getEnrollment(
+                params.getIdentifier(), EnrollmentParams.TRUE, true);
+            case EVENT -> eventService.getEvent(params.getIdentifier(), EventParams.TRUE);
+            case RELATIONSHIP -> null;
```
teleivo, line 68: Should we throw here instead?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -0,0 +1,74 @@
...
+          });
+    } catch (ForbiddenException ex) {
+      return EMPTY;
```
teleivo, line 71: Why are you catching this?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
```diff
@@ -76,10 +75,7 @@ public RelationshipOperationParams map(RequestParams requestParams) throws BadRe
             ObjectUtils.firstNonNull(
                     trackedEntity, requestParams.getEnrollment(), requestParams.getEvent())
                 .getValue())
-        .page(requestParams.getPage())
-        .pageSize(requestParams.getPageSize())
-        .totalPages(requestParams.isTotalPages())
-        .skipPaging(toBooleanDefaultIfNull(requestParams.isSkipPaging(), false))
+        .pagingAndSortingCriteriaAdapter(requestParams)
```
teleivo, line 78: Are you passing it for simplicity? We are using page, pageSize, ... in the other entities, right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
enricocolasante: Because the store is working with Hibernate objects and if we do not set the id then it will think that the entities are new ones and it will fail.
Not a fan of this solution but I really do not want to change the repository layer now

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
enricocolasante: Yes, will do

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
enricocolasante: To maintain the same behaviour as we have now.
This check will be repeated when checking the access for the relationship itself and the inaccessible ones will just be filtered out.
I am in a discussion with @JoakimSM and @Bekkalizer to check if this behaviour make sense, but anyway it will be tackled in a future PR

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipRequestParamsMapper.java
enricocolasante: Some here, I am just preparing the params as they are expected by the repository layer.
But maybe here something a little bit better can be done and I could create the adapter again when we are mapping operationParams into queryParams, so I have the part in the controller that is clean and I do not need to change the repository part.
I will look into that

teleivo (approved):

muilpp (approved):

PR #14701, opened 2023-07-24

enricocolasante:
Continuing from https://github.com/dhis2/dhis2-core/pull/14677

Client retrieves relationships specifying one end of the relationships, if the user has no access to such entity a `FORBIDDEN` error is returned.
Access control is done on both ends of the relationship. If the other end (the one not specified by the client) is not accessible, it is filtered out.
@Bekkalizer @JoakimSM do you agree with the behavior?

**What it is in this PR**
- Check access for entities used to filter the relationships.

**What is next**

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
@@ -27,8 +27,10 @@
  */
 package org.hisp.dhis.tracker.export.relationship;
 
+import java.util.List;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.common.IdentifiableObject;
+import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.EnrollmentService;
@@ -36,6 +38,9 @@ import org.hisp.dhis.program.Event;
 import org.hisp.dhis.program.EventService;
 import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.trackedentity.TrackedEntityService;
+import org.hisp.dhis.trackedentity.TrackerAccessManager;
+import org.hisp.dhis.user.CurrentUserService;
+import org.hisp.dhis.user.User;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -53,14 +58,21 @@ class RelationshipOperationParamsMapper {
 
   private final EventService eventService;
 
+  private final TrackerAccessManager accessManager;
+
+  private final CurrentUserService currentUserService;
+
   @Transactional(readOnly = true)
-  public RelationshipQueryParams map(RelationshipOperationParams params) throws NotFoundException {
+  public RelationshipQueryParams map(RelationshipOperationParams params)
+      throws NotFoundException, ForbiddenException {
+
+    User user = currentUserService.getCurrentUser();
 
     IdentifiableObject entity =
         switch (params.getType()) {
-          case TRACKED_ENTITY -> validateTrackedEntity(params.getIdentifier());
-          case ENROLLMENT -> validateEnrollment(params.getIdentifier());
-          case EVENT -> validateEvent(params.getIdentifier());
+          case TRACKED_ENTITY -> validateTrackedEntity(user, params.getIdentifier());
+          case ENROLLMENT -> validateEnrollment(user, params.getIdentifier());
+          case EVENT -> validateEvent(user, params.getIdentifier());
           case RELATIONSHIP -> throw new IllegalArgumentException("Unsupported type");
         };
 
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -73,30 +85,47 @@ public RelationshipQueryParams map(RelationshipOperationParams params) throws No
...
+  private TrackedEntity validateTrackedEntity(User user, String uid)
+      throws NotFoundException, ForbiddenException {
     TrackedEntity trackedEntity = trackedEntityService.getTrackedEntity(uid);
     if (trackedEntity == null) {
       throw new NotFoundException(TrackedEntity.class, uid);
     }
 
+    List<String> errors = accessManager.canRead(user, trackedEntity);
+    if (!errors.isEmpty()) {
+      throw new ForbiddenException("User has no access to TrackedEntity: " + uid);
```
teleivo, line 97: I would prefer if we would write sentences like

```suggestion
      throw new ForbiddenException("User has no access to tracked entity: " + uid);
```

and `throw new ForbiddenException("User has no access to enrollment: " + uid);`. We sometimes [write this way](https://github.com/dhis2/dhis2-core/blob/d848f6875c5d49f5697010e2ff0ba7dba07d2bae/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java#L279).

We not consistent yet. First, we should settle on one of these ways of writing 😅

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapperTest.java
```diff
@@ -112,8 +129,19 @@ void shouldThrowNotFoundExceptionWhenTrackedEntityIsNotFound() {
   }
 
   @Test
-  void shouldMapEnrollmentWhenAEnrollmentIsPassed() throws NotFoundException {
+  void shouldThrowForbiddenExceptionWhenUserHasNoAccessToTrackedEntity() {
```
teleivo, line 132: nit: and also up to you, I don't spell out the type of exception as that is apparent from the assertion. So I would name it `shouldThrow...` or `shouldFailToMap...`

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
```diff
@@ -588,37 +588,41 @@ void getRelationshipsByTrackedEntityRelationshipsNoAccessToRelationshipItemTo()
   }
 
   @Test
-  void getRelationshipsByTrackedEntityRelationshipsNoAccessToBothRelationshipItems() {
+  void shouldReturnForbiddenWhenGetRelationshipsWithBothEndsNotAccessible() {
```
teleivo, line 591: Now that you've added the logic + unit tests, do we need the `shouldReturnForbiddenWhenGetRelationshipsWithBothEndsNotAccessible` and `shouldReturnForbiddenWhenGetRelationshipsByNotAccessibleTrackedEntity`?

I am worried we have too many overlapping tests => many failing for the same reason 😅

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapperTest.java
enricocolasante: Done

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
enricocolasante: I agree, we can remove some overlapping tests.
I will remove `shouldReturnForbiddenWhenGetRelationshipsWithBothEndsNotAccessible`.
I still think that `shouldReturnForbiddenWhenGetRelationshipsByNotAccessibleTrackedEntity` make sense to be there to show the different behavior from `getRelationshipsByTrackedEntityRelationshipsNoAccessToRelationshipItemTo`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
enricocolasante: Yeah, I do not like to write errors. Check if you like the fix, following the same pattern as `NotFoundException`

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
teleivo: is `shouldReturnForbiddenWhenGetRelationshipsByNotAccessibleTrackedEntity` then specific to the `from` if so its name should probably be symmetrical to `getRelationshipsByTrackedEntityRelationshipsNoAccessToRelationshipItemTo`

muilpp (approved):

PR #14713, opened 2023-07-25
fix: Fix enrollment pagination [DHIS2-15616]

enricocolasante:
The issue was that `EnrollmentOperationParamsMapper` was always mapping pagination fields to default values.

Null values are checked and default values are assigned if necessary ONLY in the `RequestParamsMapper`.
In lower levels there shouldn't be any default value assignment, the mapping should just carry on the fields from higher levels.


teleivo (approved):
Thank you 😅

muilpp (approved):

PR #14716, opened 2023-07-25
feat: DB based job scheduling [DHIS2-15027]

jbee:
### Summary
A job scheduler implementation purely based on database state of `JobConfiguration`s.

Removes code that...
* would run tasks using a `ScheduledExecutorService` (the timing and selection was a in-memory process)
* would synchronize job execution state in the cluster using caches
* would run ad-hoc async tasks in manual executor with some use of job scheduling (not all usages)
* would track job progress state in memory

Adds code that...
* synchronizes job execution via database transition in the form of SQL update statements, one for each transition
* tracks job execution state in a JSONB column in `JobConfiguration`
* runs a endless scheduling loop every 20 seconds
* "remembers" input for import jobs as `FileResource`  

#### Semantic changes for `JobConfiguration`s from a (API/app) users point of view
* all jobs a persisted (no "in memory" special case)
* all jobs a only started by the current leader (no not "leader only" special case)
* all jobs can be CRON or delay based or run once ASAP (`JobType` is no longer hard coding the scheduling type)
* jobs can be cancelled by flipping the `cancel` column to true in the DB (does not matter how it is done)
* jobs are aware of their last progress in form of progress JSON and the `lastAlive` timestamp (which are updated indirectly when a job makes progress using the `JobProgress` API)
* job progress is tracked once per job configuration. During a run the progress reflects the current run, after the run it reflects the last run until the next run starts

#### API extensions
* `/jobConfigurations/due?seconds=10`: lists all jobs that should trigger in the next 10 seconds
* `/jobConfigurations/stale?seconds=1000`: lists all running jobs that have not made any progress in the last 1000s
* `/jobConfigurations/{uid}/cancel`: request cancellation for the job with UID
* `/jobConfigurations/{uid}/progress`: get the progress information of the job with UID

#### Scheduling Loop
The basics of the implemented scheduling solution.
The `JobScheduler` component is running a loop every 20 seconds in a separate thread using a normal `ScheduledExecutorService`. The loop checks if jobs in the DB need to run "soonish" based on their fields (scheduling type, cron expression, delay) and if so it delegates the run to a worker thread. For workers a normal `ExecutorService` with a cached thread pool is used. This is important to not prevent executing at times of high parallelism but auto-scale down at times of low parallelism. 

#### Scheduling and Leader Node
Only the leader node performs scheduling. The leader election is an integrated step in the loop so any leader switch is in sync with the scheduling itself. The jobs will not have to be slip in two groups. Any job will always just run on the current leader. If a leader switches in the middle of a job run the job run completes normally.

#### Bootstrapping on Startup
When the server starts it keeps running the `SchedulerStart` `StartupRoutine`. There only two things happen.
1. The scheduler loop is started (the thread running it every 20 seconds is spawned)
2. the `HeartbeatJob` is created (if not exists)

All subsequent tasks that are important to do regularly are part of the `HeartbeatJob` which runs with each loop invocation. This ensures the actions performed are traceable and have proper error handling to ensure isolation of failures so other subsequent but independent steps will run.

#### Heartbeat and Spawning System Threads
The  `HeartbeatJob` does perform tasks around management of  `JobConfiguration`s
* transfer the effect of cancellation from DB state to the node running the job
* spawn default (system) jobs if they do not exist (so now they can be recovered/reset by deleting the row in DB at any point in time)
* flip `jobState` to `DISABLED` when `enabled` is `false` (like an acknowledge, the disabled effect is linked to `enabled` flag)
* remove any done "run once" jobs after some time (system setting, includs their `FileResource` if it exists)
* reset(reschedule) jobs that are considered state (hanging) after some time (system setting)

#### Progress and Alive Tracking
Job progress is now tracked as part of the `JobConfiguration`. This happens indirectly when the code uses the `JobPorgress` API and at least 10 seconds have past since the last update the current timestamp and progress state is updated in the DB. This way stale jobs can be identified as well as any node can read the same progress from DB. This means the visible progress is potentially 1 step behind the actual progress.

#### Job Queuing 
Not to be confused with job queues. Job queues is linking several configurations to a sequence that triggers when the 1. in that sequence is triggered. Job queuing is the ability to of the system to issue many "run ASAP" jobs of the same type while another job of that type is running and thereby is blocking the immediate execution of the newly issued jobs. So these jobs "queue" until they get selected to run. Which of the queued jobs will be the next one to run out of all jobs that are ready is not predictable. This might seem an odd semantic at first but since ad-hoc jobs start from a web request there is no reasonable order to begin with. 

### Automatic Testing
Existing tests were adjusted. During spring integration testing the `executeNow` will execute the job synchronously in the main thread.

New tests were added mainly to check that the `JobConfiguration` entry cannot easily be edited so it gets out of its current state. Most of the fields of `JobConfiguration` are read-only and some even excluded from DB updates via hibernate as they never change after an entry was created.

### Manual Testing
Things to test...

* create jobs in the scheduler app
* run jobs manually in the scheduler app (check they return to "scheduled" state)
* delete jobs in the scheduler app
* edit jobs in the scheduler app
* disable/enable a job in the scheduler app
* test the continuous analytics job as it is DELAY based with the above 4 cases
* run data integrity checks (uses new run once jobs)
* set the `jobsCleanupAfterMinutes` to 1 minute and check the done data integrity job is deleted
* run a async data value set import (used new run once jobs with file resource)
* set the `jobsCleanupAfterMinutes` to 1 minute and check the done import job is deleted as well as its file resource is marked as "assigned" `false` (ready for cleanup)
* start a server with a completely empty `jobconfiguration` table, check it starts with no issue and spawns the system jobs
* delete any of the system jobs in the DB while the server is running, check it re-spawns after 20 seconds
* use the new endpoints listed in the beginning
* ...
* scenarios with cancellation 
* scenarios with stopping the server in the middle of a long running job, then starting it again and check the state is recovered to scheduled after some time (`jobsRescheduleAfterMinutes` system setting, default is 10 minutes)
* scenarios with job queues


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
```diff
@@ -103,101 +109,110 @@ public class JobConfiguration extends BaseIdentifiableObject implements Secondar
...
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
   private Date lastExecuted;
 
-  private String lastRuntimeExecution;
+  /**
+   * When the job execution finished most recently (only null when a job never finished running
+   * yet). Can be before {@link #lastExecuted} while a new run is still in progress.
+   */
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+  private Date lastFinished;
```
david-mackessy, line 132: Do we always use the `Date` type even for new fields, to keep things consistent? Do we have any plans to use/switch to the newer `LocalDate` ?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
```diff
@@ -103,101 +109,110 @@ public class JobConfiguration extends BaseIdentifiableObject implements Secondar
...
+  public SchedulingType getSchedulingType() {
+    if (schedulingType != null) return schedulingType;
+    if (cronExpression != null) return SchedulingType.CRON;
+    if (delay != null) return SchedulingType.FIXED_DELAY;
+    return SchedulingType.ONCE_ASAP;
+  }
+
+  public JobStatus getJobStatus() {
+    if (jobStatus != null) return jobStatus;
+    if (getSchedulingType() == SchedulingType.ONCE_ASAP) return JobStatus.NOT_STARTED;
```
david-mackessy, line 214: these checks seem like they could be avoided if a correct `jobStatus` is always set on creation depending on its type. Or is there some difficulty with that approach?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
```diff
@@ -364,141 +289,89 @@ public JobParameters getJobParameters() {
...
+        new SimpleTriggerContext(Clock.fixed(since, ZoneId.systemDefault()));
+    Date next = new CronTrigger(cronExpression).nextExecutionTime(context);
+    return next == null ? null : next.toInstant();
   }
 
-  public Integer getQueuePosition() {
-    return queuePosition;
+  private Instant nextDelayExecutionTime(@Nonnull Instant since) {
+    if (delay == null || delay <= 0) return null;
+    // always want to run delay after last start, right ways when never started
```
david-mackessy, line 362: this section of the comment `right ways when never started` doesn't read well & I don't know what it's supposed to mean.
Should it say `right away when never started` ?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
```diff
@@ -110,11 +113,26 @@ public interface JobProgress {
    * Flow Control API:
    */
 
+  /**
+   * OBS! Should only be called after the task is complete and no further progress is tracked.
+   *
+   * @return true, if all processes in this tracking object were successful.
+   */
+  default boolean isSuccessful() {
```
david-mackessy, line 121: is having these default methods increasing the risk of bugs? It relies on the implementer overriding these default method and providing their own logic (they don't have to and could call these methods, giving unexpected results). Would it not be better to force the implementer to override by leaving it abstract?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
```diff
@@ -51,148 +62,126 @@
...
+  GEOJSON_IMPORT(),
+  EVENT_IMPORT(),
+  ENROLLMENT_IMPORT(),
+  TEI_IMPORT(),
+  GML_IMPORT(),
+
+  /*
+  System Jobs
+  */
+  HEARTBEAT(every(20, "DHIS2rocks1", "Heartbeat")),
```
david-mackessy, line 116: providing hard-coded UIDs seems a bit strange here when we usually generate them everywhere else. Unless there's a specific reason for this?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultAuthenticationService.java
```diff
@@ -41,19 +45,24 @@ public class DefaultAuthenticationService implements AuthenticationService {
 
   @Override
   public void obtainAuthentication(String userId) throws NotFoundException {
-    if (userId == null) {
-      clearAuthentication();
-      return;
+    Object username = "System";
+    Collection<? extends GrantedAuthority> authorities = List.of(new SimpleGrantedAuthority("ALL"));
```
david-mackessy, line 49: It feels bad mixing these 2 scenarios like this (valid userId v system call). At best it works as expected. At worst, something happens with the `userId` and the current user has full access it seems, albeit with a different username and no password.

david-mackessy:
nice work, seems like a nice improvement on previous implementation 👍

david-mackessy (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
jbee: Good question. I used `Date` for consistency. Within persisted objects we have used `Date` for timestamps this far. I started using `Instant` in other new code or some of the `java.time` class. We should certainly move towards them. Good talking point for our agenda.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
jbee: Yes, this solution was chosen because of the different code paths. In particular the generic code that creates objects via reflection from user input. To not impose any new requirements on the user input while having the reasonable behaviour should this not be given explicitly this was the best solution that keeps the logic in one place.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
jbee: Oh yes, that is a typo that you guessed correct.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
david-mackessy: maybe we can chat about this when giving the demo next week, it's still a bit unclear for me :)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultAuthenticationService.java
jbee: Yes, I don't like this either. Any better ideas?

PR #14721, opened 2023-07-26
chore: map order values in controller TECH-1569

teleivo:
to dhis-api field names so store does not need to know naming used in controller views


teleivo:
close in favor of https://github.com/dhis2/dhis2-core/pull/14728 to match the correct jira issue

PR #14726, opened 2023-07-26
fix: Adapt old api to org unit mode acl standard [DHIS2-1604]

muilpp:
Back port to old API of: https://github.com/dhis2/dhis2-core/pull/14402

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventUtils.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventUtils.java
@@ -27,11 +27,26 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
+import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ACCESSIBLE;
+import static org.hisp.dhis.common.OrganisationUnitSelectionMode.SELECTED;
+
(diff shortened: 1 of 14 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventUtils.java
```diff
@@ -50,4 +64,136 @@ public static UserInfoSnapshot jsonToUserInfo(String userInfoAsString, ObjectMap
       throw new IllegalArgumentException(e);
     }
   }
+
+  public static List<OrganisationUnit> validateAccessibleOrgUnits(
```
teleivo, line 68: I don't want us to make concessions in new tracker code due to changes needed for old tracker 😅 Not a fan of utils 😂 

Can we either
* duplicate the code (which is our strategy so far) so that we stay independent
* move this in an existing ACL related class?

enricocolasante: I agree, I would leave this code in `EventOperationParamsMapper` and create a new `EventUtils` or something similar, in the deprecated tracker

teleivo (approved):

PR #14728, opened 2023-07-26
chore: implement /tracker/event?order consistently with other endpoints TECH-1606

teleivo:
Note that this is the first in many order related PRs. We will have more possibilities in refactoring some of the methods/reuse them in RequestParamUtils, ...

## dhis-web-api

- Validate (EventsExportController): all orderable fields can be ordered by our service  
- Validate (EventRequestParamsMapper): the fields in the users order request parameter can be ordered by   
- Validate (EventRequestParamsMapper): UIDs are valid. /tracker/events allows ordering by data elements and attributes by passing UIDs in the order request parameter   
- Validate (EventRequestParamsMapper): the fields in the users order request parameter are not repeated like order=createdAt,createdAt:desc order=createdAt,createdAt   
- Map (EventMapper): from org.hisp.dhis.webapi.controller.tracker.view models representing the JSON API (known to users) to org.hisp.dhis (dhis-api) models used to persist into the DB like org.hisp.dhis.program.Event   

## Store

- Map: from org.hisp.dhis (dhis-api) models used to persist into to DB columns.

## Additional changes

* Added tests for OrderCriteriaTest to understand its current behavior. We might want to change it in a later issue. Notice that an invalid sort direction like `foo` is turned into `asc`. We should never hide any such issues but fail instead.

### Service

- Validate: attributes, data elements exist
- Validate: ACL attributes, data elements (access denied if user does not have it instead of filtering the attribute from the order components)

### Store

- Validate: fields can be ordered by (maybe only for now as a last safety net; until we know for sure no one else than tracker can directly use our stores).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
@@ -198,6 +198,11 @@ public class DefaultEventService implements EventService {
     return Events.of(eventList, pager);
   }
 
+  @Override
+  public Set<String> getOrderableFields() {
+    return eventStore.getOrderableFields();
+  }
+
   /**
    * This method will apply the logic related to the parameter 'totalPages=false'. This works in
    * conjunction with the method: {@link EventStore#getEvents(EventSearchParams,
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventMapper.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventMapper.java
@@ -27,6 +27,9 @@
  */
 package org.hisp.dhis.webapi.controller.tracker.export.event;
 
+import static java.util.Map.entry;
+
+import java.util.Map;
 import org.hisp.dhis.program.Event;
 import org.hisp.dhis.webapi.controller.tracker.export.DataValueMapper;
 import org.hisp.dhis.webapi.controller.tracker.export.NoteMapper;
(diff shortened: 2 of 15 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventMapper.java
```diff
@@ -48,6 +51,42 @@
...
+          entry("enrollment", "enrollment.uid"),
+          entry("enrollmentStatus", "enrollment.status"),
+          entry("event", "uid"),
+          entry("followup", "enrollment.followup"),
+          entry("occurredAt", "executionDate"),
+          entry("orgUnit", "organisationUnit.uid"),
+          entry("orgUnitName", "organisationUnit.name"),
+          entry(
+              "program",
+              "enrollment.program.uid"), // TODO(ivo) this is what we do in the export. I assume we
```
teleivo, line 79: @enricocolasante do you have an opinion on whether we should use

`"enrollment.program.uid"` or `"program.uid"`

muilpp (approved):

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
```diff
@@ -409,25 +411,45 @@ void shouldMapOrderParameterToOrderCriteriaWhenFieldsAreSortable() throws BadReq
...
-            new OrderParam("scheduledAt", SortDirection.ASC)),
+            new OrderParam("created", SortDirection.ASC),
+            new OrderParam("programStage.uid", SortDirection.DESC),
+            new OrderParam("dueDate", SortDirection.ASC)),
         params.getOrders());
   }
 
   @Test
-  void shouldThrowWhenOrderParameterContainsUnsupportedField() {
+  void shouldThrowWhenOrderParameterContainsInvalidOrderComponents() {
```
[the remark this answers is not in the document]
teleivo: Yes 👌🏻 which one do you prefer?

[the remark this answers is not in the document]
muilpp: I prefer `shouldFail` :)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -198,6 +198,11 @@ public Events getEvents(EventOperationParams operationParams)
     return Events.of(eventList, pager);
   }
 
+  @Override
+  public Set<String> canOrderEventsByFields() {
```
enricocolasante, line 202: Would something like `getOrderableFields` be more descriptive?
At first glance, I thought that this method was returning a boolean

enricocolasante (commented):
After your question about the program, I realized that we should check if everything works fine when we are ordering program events that do not have an enrollment

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
teleivo: It was indeed a predicate before 😅 I did want to keep 'Events' in the the method name as only getEvents can be ordered. Not sure if this is important though. What do you think?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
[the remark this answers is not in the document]
teleivo: I will do this in one of the follow up PRs 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventMapper.java
teleivo: I changed the todo to `TODO(TECH-1606)`.
I will keep this as it is for now and investigate this separately.

Maybe this info tells you whats going on:
* "enrollment.program.uid" turns to DB alias "p_uid"
* which refers to the uid in program table which we get via joins in  
https://github.com/dhis2/dhis2-core/blob/b157e811a066a41d716cfc7d3d6ae3ae6408e276/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java#L784-L787

enricocolasante (approved):

PR #14737, opened 2023-07-27
chore: Order objects in tracker import report [DHIS2-15596]

enricocolasante:
Entities imported using tracker endpoint were presented out of order in the response.
It is essential to preserve the order because the client is allowed to not specify the `UID` of the entities in case they are being created and then it is the server's responsibility to create those `UID`s.

The request payload can present the entities in a flattened or nested way (or any mix of the two), the order that is preserved is exactly the order in which the entities appear in the payload.

*Side fixes*
- `Index` field is not present anymore in `bundleReport.typeReportMap."TRACKER_TYPE".objectsReport` as the order of the entities is just the one in the response
- `validationMode=SKIP` is now working when calling the import endpoint. Before, that option was skipping validation and persistence.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Enrollment.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Enrollment.java
@@ -83,10 +83,15 @@ public class Enrollment implements TrackerDto {
   @JsonProperty private User updatedBy;
 
   @JsonProperty int index;
+
   @JsonProperty private Geometry geometry;
+
   @JsonProperty @Builder.Default private List<Event> events = new ArrayList<>();
+
   @JsonProperty @Builder.Default private List<Relationship> relationships = new ArrayList<>();
+
   @JsonProperty @Builder.Default private List<Attribute> attributes = new ArrayList<>();
+
   @JsonProperty @Builder.Default private List<Note> notes = new ArrayList<>();
 
   @Override
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Enrollment.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Enrollment.java
@@ -82,14 +82,11 @@ public class Enrollment implements TrackerDto {
 
   @JsonProperty private User updatedBy;
 
+  @JsonProperty int index;
   @JsonProperty private Geometry geometry;
-
   @JsonProperty @Builder.Default private List<Event> events = new ArrayList<>();
-
   @JsonProperty @Builder.Default private List<Relationship> relationships = new ArrayList<>();
-
   @JsonProperty @Builder.Default private List<Attribute> attributes = new ArrayList<>();
-
   @JsonProperty @Builder.Default private List<Note> notes = new ArrayList<>();
 
   @Override
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/TrackerDto.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/TrackerDto.java
@@ -33,7 +33,10 @@ import org.hisp.dhis.tracker.TrackerType;
  * @author Morten Svanæs <msvanaes@dhis2.org>
  */
 public interface TrackerDto {
+
   String getUid();
 
   TrackerType getTrackerType();
+
+  int getIndex();
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/report/Entity.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/report/Entity.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.tracker.imports.report;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.ArrayList;
 import java.util.List;
@@ -39,13 +40,10 @@ import org.hisp.dhis.tracker.TrackerType;
  */
 @Data
 public class Entity {
-  /** Type of object this {@link Entity} represents. */
   @JsonProperty private final TrackerType trackerType;
 
-  /** Index into list. */
-  @JsonProperty private Integer index;
+  @JsonIgnore private int index;
 
-  /** UID of entity (if object is id object). */
   @JsonProperty private String uid;
 
   private List<Error> errorReports = new ArrayList<>();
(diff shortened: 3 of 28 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/report/TrackerTypeReport.java
```diff
@@ -92,20 +86,16 @@ public boolean isEmpty() {
...
 
   public void addEntity(Entity entity) {
-    this.entityReport.put(entity.getIndex(), entity);
+    this.entityReport.add(entity);
   }
 
   private List<Error> getErrorReports() {
     List<Error> errorReports = new ArrayList<>();
-    entityReport.values().forEach(entity -> errorReports.addAll(entity.getErrorReports()));
+    entityReport.forEach(entity -> errorReports.addAll(entity.getErrorReports()));
```
teleivo, line 94: I think we could also do

```suggestion
  return entityReport.values().stream().flatMap(e -> e.getErrorReports().stream()).toList();
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/DefaultValidationService.java
```diff
@@ -70,7 +71,13 @@ private ValidationResult validate(TrackerBundle bundle, Validator<TrackerBundle>
           "Skipping validation for metadata import by user '"
               + bundle.getUsername()
               + "'. Not recommended.");
-      return Result.empty();
+      return new Result(
```
teleivo, line 74: That was the bug you mentioned right? We did skip validation but also removed all the entities to be persisted 😬

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportReportTest.java
```diff
@@ -0,0 +1,134 @@
...
+                    + "&validationMode=SKIP"
+                    + "&atomicMode=OBJECT"
+                    + "&skipRuleEngine=true",
+                WebClient.Body("tracker/import_flatten_payload.json"))
+            .content(HttpStatus.OK)
+            .as(JsonImportReport.class);
+
+    assertReportEntities(
+        List.of("IybbQIQt6te", "daMwzsKN3te", "FRM97UKN8te"),
+        importReport,
```
teleivo, line 103: nit: the uids you expect are tracked entity ones I assume right? So might be bit clearer to move the tracker type parameter second so its closer to the expected uids 😄

teleivo (commented):
dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportReportTest.java is super valuable in understanding what this is all about 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/report/Entity.java
```diff
@@ -39,13 +40,10 @@
  */
 @Data
 public class Entity {
-  /** Type of object this {@link Entity} represents. */
   @JsonProperty private final TrackerType trackerType;
 
-  /** Index into list. */
-  @JsonProperty private Integer index;
+  @JsonIgnore private int index;
```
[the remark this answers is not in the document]
enricocolasante: Yes, we need to put it in the release note also explaining that this index is being removed because was not providing any valuable information

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/DefaultValidationService.java
enricocolasante: Yes, that is it. Skip validation was actually skipping everything

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportReportTest.java
```diff
@@ -0,0 +1,134 @@
...
+        TrackerType.TRACKED_ENTITY);
+    assertReportEntities(
+        List.of("IybbQIQt6en", "daMwzsKN3en"), importReport, TrackerType.ENROLLMENT);
```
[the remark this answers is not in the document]
enricocolasante: I actually find it very counterintuitive to have to stick to random UIDs. If it was for me I would have all UIDs as static variables so I don't have to deal with real UIDs.
But we can discuss more on this tomorrow

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportReportTest.java
enricocolasante: :+1:

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Enrollment.java
```diff
@@ -82,14 +82,11 @@ public class Enrollment implements TrackerDto {
 
   @JsonProperty private User updatedBy;
 
+  @JsonProperty int index;
```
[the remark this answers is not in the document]
enricocolasante: Yes, these classes are not exposed through the API but they are still serialized/deserialized for the async process

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/TrackerDto.java
```diff
@@ -33,7 +33,10 @@
  * @author Morten Svanæs <msvanaes@dhis2.org>
  */
 public interface TrackerDto {
+
   String getUid();
 
   TrackerType getTrackerType();
+
+  int getIndex();
```
teleivo, line 41: I wonder if we should give this a more descriptive name 🤔? As the index is actually the index into the import "bundle" or so.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/report/TrackerTypeReport.java
```diff
@@ -65,17 +64,12 @@ public TrackerTypeReport(
...
-        this.entityReport.put(entity.getIndex(), entity);
-      }
-    }
+    this.entityReport = entityReport;
   }
 
   @JsonProperty("objectReports")
   public List<Entity> getEntityReport() {
-    return new ArrayList<>(entityReport.values());
+    return entityReport.stream().sorted(Comparator.comparingInt(Entity::getIndex)).toList();
```
teleivo, line 72: Do you think its worth adding a unit test for this? I know we have the integration test, just wondering if that would be more obvious to spot where the error might come from.

teleivo (approved):
We agreed to unblock Simona with the implementation in this PR. This adds tests allowing us to change the implementation.

We will then figure out if we can adapt the implementation so that the entities do not have to be aware of the index. As if we compare our report to a collection its elements do not know they are in a collection and not at what index.

muilpp (approved):

PR #14738, opened 2023-07-27
fix: Avoid org unit and org unit mode in same request [TECH-1585]

muilpp:
Requests defining an org unit mode ACCESSIBLE or CAPTURE, should fail if they also define an org unit.

That's because these two modes get the org units from the user's profile, so any org unit provided will not be used to filter the results. In addition, the org unit will be validated, and if the user has no access to it, an exception will be raised, which doesn't make much sense, since this org unit plays no role in filtering the results.

Ticket: https://dhis2.atlassian.net/browse/TECH-1585


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -97,6 +99,8 @@ public EventOperationParams map(RequestParams requestParams) throws BadRequestEx
 
     validateUpdateDurationParams(requestParams);
 
+    validateOrgUnitParams(requestParams.getOrgUnit(), orgUnitMode);
```
enricocolasante, line 102: Maybe not part of this PR but should we make `orgUnit` field behave as the other endpoints.
`Event` is the only one that accepts one `orgUnit`, does it make sense?

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
teleivo: Very good point! Not sure how much work it would be to allow that. It would at least be a backwards compatible change

muilpp: I've seen that `trackedEntities` and `enrollments` accept a list of org units uids and event expects a single uid. I don't know the reason behind though, so I can't tell if it makes sense. What do you think? Should we refactor it to accept a list of uids too?

teleivo: I am in favor of all endpoints accepting a list in `orgUnits` request parameter but I don't know either if it makes sense to implement that for events. 🤷🏻

muilpp: Me too, @enricocolasante does it make sense from the user perspective to accept multiple org units for events?
In case it does, we can use the opportunity and edit the docs, because based on them, enrollments does not support a list of org units: https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html?h=enrollments+develop#enrollments-get-apitrackerenrollments

We should have it written as in tracked entities: https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html?h=master+develop#tracked-entities-get-apitrackertrackedentities

teleivo: OpenAPI spec is the source of truth from now on: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/resources/openapi/EnrollmentsExportController.md#parameterenrollmentrequestparamsorgunits 

We need to work on automaticall deploying the OpenAPI spec and linking it in https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html instead of the tables in there.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParamsMapperTest.java
```diff
@@ -198,4 +204,66 @@ void testMappingOrderParamsNoOrder() throws BadRequestException {
...
+
+  @Test
+  void shouldFailWhenOrgUnitSuppliedAndOrgUnitModeAccessible() {
+    RequestParams requestParams = new RequestParams();
+    requestParams.setOrgUnits(Set.of(UID.of(ORG_UNIT_1_UID)));
+    requestParams.setOrgUnitMode(ACCESSIBLE);
+
+    Exception exception = assertThrows(BadRequestException.class, () -> mapper.map(requestParams));
+
+    assertEquals(
```
teleivo, line 216: You could use `assertStartsWith` or `assertContains` to assert on parts of the message to be sure it is actually the case you wanted to test. This way you might still have a bit more flexibility in changing the message without always having to change the test.

If we move the function to RequestParamUtils we can also test it there or at least most of the cases so we don't have to repeat all the tests.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParamsMapper.java
```diff
@@ -90,4 +94,16 @@ public EnrollmentOperationParams map(RequestParams requestParams) throws BadRequ
...
   }
+
+  private void validateOrgUnitParams(Set<UID> orgUnits, OrganisationUnitSelectionMode orgUnitMode)
+      throws BadRequestException {
+    if (orgUnits != null
+        && !orgUnits.isEmpty()
+        && (orgUnitMode == ACCESSIBLE || orgUnitMode == CAPTURE)) {
+      throw new BadRequestException(
+          String.format(
+              "Org unit mode %s cannot be used with an org unit specified. Please remove the org unit and try again.",
```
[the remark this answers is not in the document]
muilpp: You are right, in the new API `orgUnits` can't be null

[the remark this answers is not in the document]
muilpp: I'm fine with the message for the new API, but for the old one, I think
`"ouMode %s cannot be used with orgUnits. Please remove the ou parameter and try again.",`
would be better, do you agree?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtilsTest.java
```diff
@@ -162,4 +175,40 @@ private TrackedEntityAttribute trackedEntityAttribute(String uid) {
...
+    Exception exception =
+        assertThrows(
+            BadRequestException.class,
+            () -> validateOrgUnitParams(Set.of(UID.of(orgUnit.getUid())), CAPTURE));
+
+    assertStartsWith("orgUnitMode CAPTURE cannot be used with orgUnits.", exception.getMessage());
+  }
+
+  @Test
+  void shouldMapOrgUnitModeWhenOrgUnitSuppliedAndOrgUnitModeSelected() {
```
teleivo, line 201: maybe use `shouldPass` to oppose the `shouldFail` 😄 

```suggestion
  void shouldPassWhenOrgUnitSuppliedAndOrgUnitModeSelected() {
```

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
```diff
@@ -452,4 +457,61 @@ void shouldThrowWhenOrderParameterContainsRepeatedOrderComponents() {
         () -> assertContains("enrolledAt", exception.getMessage()),
         () -> assertContains("zGlzbfreTOH", exception.getMessage()));
   }
+
+  @Test
+  void shouldFailWhenOrgUnitSuppliedAndOrgUnitModeAccessible() {
```
teleivo, line 462: Do we need all these tests in all the controllers even though they call a tested function from RequestParamUtils that is tested in RequestParamUtilsTest? Imagine how many tests would fail for the same reason if someone would add a bug in the validation function 😅

muilpp: Right, but for now I'll leave them in the old API and in the events mapper of the new one, until we refactor the events controller to accept a semi colon separated value as the input of `orgUnit.`

teleivo (approved):

PR #14784, opened 2023-07-31
chore: make it hard to misuse the EventSearchParams API TECH-1606

teleivo:
`/tracker/events` is the endpoint with the most "order features" of all tracker exporter endpoints. Users can order by event fields, data elements and tracked entities as in `/tracker/events?order=createdAt:desc,TvjwTPToKHO,scheduledAt,OBzmpRP6YUh:asc`.

Its essential that we preserve the user provided order in whatever data structure we use to represent the order.

Challenges:
* You can see from the example request that a UID might be a data element or a tracked entity attribute. In case we do not find a UID as one of the above we cannot be sure what the user meant.
* Certain field names are considered valid UIDs. For example: `completedAt, scheduledAt, completedBy, orgUnitName, lastUpdated, completedBy` are all valid UIDs. We thus need to be careful with checks like `isValidUid`.
* The requirement to preserve order and support 3 different types that do not all have something useful in common in a generic collection such as a `List` is hard (see typesafe heterogeneous container).

## Goal

The goal of this PR is to implement most tasks in https://dhis2.atlassian.net/browse/TECH-1606. In addition we want to ["make it easy to do the right thing and almost impossible to do the wrong thing"](https://www.infoq.com/articles/API-Design-Joshua-Bloch/).

Ordering by attributes (and data elements) turned from 

```java
EventOperationParams params =
        .filterAttributes("toUpdate000")
        .attributeOrders(List.of(OrderCriteria.of("toUpdate000", SortDirection.DESC)))
        .orders(List.of(new OrderParam("toUpdate000", SortDirection.DESC)))
        .build();
```

to

```java
EventOperationParams params =
        .orderBy(UID.of("toUpdate000"), SortDirection.DESC)
        .build();
```

EventOperationParams(EventSearchParams) will make sure to update internal structures necessary for the JdbcEventStore to create the query. Users do not have to remember that.

## Types

This is to illustrate how the types/information gets richer from the request to the store. Note that this is a pseudo notation as there is no union.

1. A request like
`order=createdAt:desc,TvjwTPToKHO,scheduledAt,OBzmpRP6YUh:asc` starts out as => `List<OrderCriteria<String>>`
2. EventRequestParamsMapper => `List<Order<String|UID>>`
3. EventOperationParamsMapper => `List<Order<String|DataElement|TrackedEntityAttribute>>`

Each stage makes sure that the types at runtime are one of the supported ones.

#### Filter

Filtering by attributes like

```java
    .attributeFilters("toUpdate000:eq:summer day")
```

turns into join

```sql
INNER JOIN trackedentityattributevalue "toUpdate000"
            ON "toUpdate000".trackedentityid = TEI.trackedentityid
INNER JOIN trackedentityattribute "toUpdate000ATT"
            ON "toUpdate000".trackedentityattributeid =
                "toUpdate000ATT".trackedentityattributeid AND
                "toUpdate000ATT".UID = 'toUpdate000' AND
                lower("toUpdate000".value) = 'summer day'
```

#### Order

Ordering by attributes

```java
    .orderBy(UID.of("toUpdate000"), SortDirection.ASC)
```

turns into

```sql
"toUpdate000".value AS toUpdate000_value,
```

in the select and join (as when filtering)

```sql
INNER JOIN trackedentityattributevalue "toUpdate000"
            ON "toUpdate000".trackedentityid = TEI.trackedentityid
INNER JOIN trackedentityattribute "toUpdate000ATT"
            ON "toUpdate000".trackedentityattributeid =
                "toUpdate000ATT".trackedentityattributeid AND
                "toUpdate000ATT".UID = 'toUpdate000'
```

An `order by toUpdate000_value ASC` is added to the event subquery. This is so that pagination is correct as the pagination using limit/offset is applied in the event subquery. Additionally, we add another `order by toUpdate000_value ASC` in the outer query as we cannot rely on ordering by simply ordering the subquery.

### Data elements

Data elements have to be treated differently than attributes as they are stored as JSONB in the event table itself.

An example JSONB

```json
{
    "D7m8vpzxHDJ": {"value": "P+", "created": "2014-11-02T22:55:17.849", "storedBy": null,
"lastUpdated": "2014-11-02T22:55:17.849", "providedElsewhere": false},
    "HmkXnHJxcD1": {"value": "TAI", "created": "2014-11-02T22:55:19.225", "storedBy": null, "lastUpdated":
    "2014-11-02T22:55:19.225", "providedElsewhere": false}
}
```

#### Filter

Filtering by data elements like

```java
  .dataElementFilters("DATAEL00001:like:%val%")
```

turns into select

```sql
lower(psi.eventdatavalues #>> '{DATAEL00001, value}') as DATAEL00001
```

and where clause in the event subquery

```sql
where lower(psi.eventdatavalues #>> '{DATAEL00001, value}') like :parameter_2
```

#### Ordering by data element

```java
    .orderBy(UID.of("DATAEL00006"), SortDirection.ASC)
```

turns into select

```sql
cast(psi.eventdatavalues #>> '{DATAEL00006, value}' as numeric) as DATAEL00006
```

in the event subquery.

Note that the cast is due to this data element being of a numeric value type.

An order by clause is added in the events subquery and the outer query

```sql
order by DATAEL00006 DESC
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/Order.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/Order.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker;
+
+import lombok.Value;
+import org.hisp.dhis.dataelement.DataElement;
+import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
+import org.hisp.dhis.webapi.controller.event.mapper.SortDirection;
+
+/**
+ * Tracker exporter APIs allow ordering by different types. For example events can be ordered by
+ * field names, {@link DataElement} and {@link TrackedEntityAttribute}. It is crucial for the order
+ * values to stay in one collection as their order needs to be kept as provided by the user. We
+ * cannot come up with a type-safe type that captures the above order features and that can be used
+ * in a generic collection such as a List (see typesafe heterogeneous container). We therefore use
+ * {@link Order} with a field of type {@link Object}. We get compile time type safety via such as
+ * {@link org.hisp.dhis.tracker.export.event.EventSearchParams#orderBy(DataElement, SortDirection)}.
+ * This allows us to advocate the types that can be ordered by while storing the order in a single
+ * {@code List} of {@link Order}. Runtime type checks are then used to ensure users (and developers)
+ * get meaningful error messages in case they order by unsupported fields or types.
+ */
+@Value
+public class Order {
+  Object field;
+  SortDirection direction;
+}
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/Order.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/Order.java
@@ -38,11 +38,12 @@ import org.hisp.dhis.webapi.controller.event.mapper.SortDirection;
  * values to stay in one collection as their order needs to be kept as provided by the user. We
  * cannot come up with a type-safe type that captures the above order features and that can be used
  * in a generic collection such as a List (see typesafe heterogeneous container). We therefore use
- * {@link Order} with a field of type {@link Object}. We get compile time type safety via such as
- * {@link org.hisp.dhis.tracker.export.event.EventSearchParams#orderBy(DataElement, SortDirection)}.
- * This allows us to advocate the types that can be ordered by while storing the order in a single
- * {@code List} of {@link Order}. Runtime type checks are then used to ensure users (and developers)
- * get meaningful error messages in case they order by unsupported fields or types.
+ * {@link Order} with a field of type {@link Object}. We get compile time type safety via methods
+ * such as {@link org.hisp.dhis.tracker.export.event.EventSearchParams#orderBy(DataElement,
+ * SortDirection)}. This allows us to advocate the types that can be ordered by while storing the
+ * order in a single {@code List} of {@link Order}. Runtime type checks are then used to ensure
+ * users (and developers) get meaningful error messages in case they order by unsupported fields or
+ * types.
  */
 @Value
 public class Order {
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParams.java
@@ -41,11 +41,12 @@ import org.hisp.dhis.category.CategoryOptionCombo;
 import org.hisp.dhis.common.AssignedUserSelectionMode;
 import org.hisp.dhis.common.IdSchemes;
 import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.common.UID;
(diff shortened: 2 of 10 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParams.java
```diff
@@ -41,11 +41,12 @@
...
 import org.hisp.dhis.common.IdSchemes;
 import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.common.UID;
 import org.hisp.dhis.event.EventStatus;
 import org.hisp.dhis.program.ProgramStatus;
 import org.hisp.dhis.program.ProgramType;
-import org.hisp.dhis.webapi.controller.event.mapper.OrderParam;
-import org.hisp.dhis.webapi.controller.event.webrequest.OrderCriteria;
+import org.hisp.dhis.tracker.Order;
+import org.hisp.dhis.webapi.controller.event.mapper.SortDirection;
```
teleivo, line 49: Note that this class is in the `dhis-api` module. This is the class we rely on so far. There are multiple classes of this sort in the code (`org.hisp.dhis.query`, `GistQuery.Order`, ...). We could discuss whether there is a way to come up with a standardized class for ordering in DHIS2 in the backend meeting.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/Order.java
```diff
@@ -0,0 +1,51 @@
...
+import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
+import org.hisp.dhis.webapi.controller.event.mapper.SortDirection;
+
+/**
+ * Tracker exporter APIs allow ordering by different types. For example events can be ordered by
+ * field names, {@link DataElement} and {@link TrackedEntityAttribute}. It is crucial for the order
+ * values to stay in one collection as their order needs to be kept as provided by the user. We
+ * cannot come up with a type-safe type that captures the above order features and that can be used
+ * in a generic collection such as a List (see typesafe heterogeneous container). We therefore use
+ * {@link Order} with a field of type {@link Object}. We get compile time type safety via such as
```
enricocolasante, line 41:
```suggestion
 * {@link Order} with a field of type {@link Object}. We get compile time type safety via methods such as
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -143,28 +137,17 @@ public EventSearchParams map(EventOperationParams operationParams)
...
+    EventSearchParams searchParams = new EventSearchParams();
+    List<QueryItem> dataElementFilters =
+        parseDataElementQueryItems(
+            operationParams.getDataElementFilters(), this::dataElementToQueryItem);
+    mapDataElementFilters(searchParams, dataElementFilters);
 
-    List<QueryItem> filters =
-        parseDataElementQueryItems(operationParams.getFilters(), this::dataElementToQueryItem);
+    List<QueryItem> attributeFilters = parseFilterAttributes(operationParams.getAttributeFilters());
+    validateFilterAttributes(attributeFilters);
```
enricocolasante, line 147:
```suggestion
    validateAttributeFilters(attributeFilters);
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -357,41 +334,13 @@ private boolean userCanSearchOrgUnitModeALL(User user) {
         || user.isAuthorized(Authorities.F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS.name());
   }
 
-  private List<QueryItem> parseFilterAttributes(
-      String filterAttributes, List<OrderParam> attributeOrderParams) throws BadRequestException {
+  private List<QueryItem> parseFilterAttributes(String filterAttributes)
```
enricocolasante, line 337:
```suggestion
  private List<QueryItem> parseAttributeFilter(String attributeFilters)
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -587,4 +501,59 @@ private static List<OrganisationUnit> getAccessibleDescendants(
...
+    for (QueryItem item : dataElementFilters) {
+      for (QueryFilter filter : item.getFilters()) {
+        params.addDataElementFilter((DataElement) item.getItem(), filter);
+      }
+    }
+  }
+
+  private void mapAttributeFilters(EventSearchParams params, List<QueryItem> attributeFilters) {
+    for (QueryItem item : attributeFilters) {
+      if (item.getFilters().isEmpty()) {
```
enricocolasante, line 515: This difference in behavior between attributes and data elements is because we support filtering on attributes with no value, meaning to check if the attribute is present, but we do not support the same for data element?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -587,4 +501,59 @@ private static List<OrganisationUnit> getAccessibleDescendants(
...
+        if (de != null) {
+          params.orderBy(de, order.getDirection());
+          continue;
+        }
+
+        TrackedEntityAttribute tea =
+            trackedEntityAttributeService.getTrackedEntityAttribute(uid.getValue());
+        if (tea == null) {
+          throw new BadRequestException(
+              "Cannot order by '"
```
enricocolasante, line 545: Do you think should provide here as well the list of supported fields for ordering?
At this point we cannot really assume that they were trying to ordering on a UID, so the error message could be the same of the one that we throw from the `RequestParamsMapper`.

It could be really strange that they add or remove a char from a `field` and they get completely different error messages

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
teleivo: Yes 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
teleivo: I would like to but there are two issues that make it hard
1. these validations are in the (service) mapper. Here we do not have access to the supported field names. The service and store both have access. We cannot add a dependency on the event service here as this would cause a cycle. Adding a dependency on the store here feels odd. Adding direct access to the JdbcEventStore implementation specific orderable fields does not feel right as well.
2. the error message at this point would still be different than the error message from the web layer. This is because we have already translated field names known to users (web view) to the ones that are known to the store (dhis-api).

teleivo: The case you are describing can only happen if the field name they pass in `order` is also a valid UID. I am not sure how often that could happen. I could try to solve this puzzle 😂 

We have documented all the orderable fields in OpenAPI. So in that particular case the answer can be found in the docs.

enricocolasante (approved):

muilpp (approved):

PR #14786, opened 2023-07-31
chore: Add test for default value in binding object [TECH-1592]

enricocolasante:
Weirdly binding of default values in binding object done by Spring MVC seems to work fine.
I am just adding a Controller test to verify that.

### DIFF
--- a/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
+++ b/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
@@ -143,6 +143,14 @@ class SpringBindingTest {
         .andExpect(content().string(containsString("3")));
   }
 
+  @Test
+  void shouldReturnPassedValueWhenDefaultParameterIsPassed() throws Exception {
+    mockMvc
+        .perform(get(ENDPOINT + "/default").param("defaultValue", "0"))
+        .andExpect(content().string(containsString("OK")))
+        .andExpect(content().string(containsString("0")));
+  }
+
   @Controller
   private class BindingController extends CrudControllerAdvice {
     @GetMapping(value = ENDPOINT)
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
```diff
@@ -135,6 +135,14 @@ void shouldReturnABadRequestWhenInvalidValueForBooleanInBindingObjectIsPassedAsP
                         "Value 'INVALID' is not valid for parameter booleanValue. Invalid boolean value [INVALID]")));
   }
 
+  @Test
+  void shouldReturnADefaultValueWhenDefaultParameterIsNotPassed() throws Exception {
```
muilpp, line 139: could we also add a test to verify that the default value is not used if the parameter has a value?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
```diff
@@ -135,6 +135,22 @@ void shouldReturnABadRequestWhenInvalidValueForBooleanInBindingObjectIsPassedAsP
...
+    mockMvc
+        .perform(get(ENDPOINT + "/default").param("booleanValue", "true"))
```
teleivo, line 141: Why do you pass `booleanValue` `true`? You are not asserting on anything it has an influence, right? If that is the case I would remove it.

teleivo (commented):
Could it be that there was a bug in Spring?

Or is it that the test setup is different from our production code (web config, ...) so that you are not seeing the same behavior?

Just wondering as you sounded puzzled as to why you cannot see it anymore 😬

enricocolasante:
> Could it be that there was a bug in Spring?
> 
> Or is it that the test setup is different from our production code (web config, ...) so that you are not seeing the same behavior?
> 
> Just wondering as you sounded puzzled as to why you cannot see it anymore grimacing

I have no idea.
First thing I was trying to go back and check exactly that, if we changed Spring version or any related library, but going back in time and I couldn't reproduce locally the issue.

I was always testing it manually and if Marc wasn't there as well I would start to think that I got it all wrong and there was never an issue in the first place

PR #14787, opened 2023-07-31
Tech 1585 org unit ou mode in request

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -306,8 +305,12 @@ private String getOrgUnitModeViolation(EventOperationParams params, User user, P
       case ALL -> userCanSearchOrgUnitModeALL(user)
           ? null
           : "Current user is not authorized to query across all organisation units";
```
enricocolasante, line 307: This validation altogether couldn't be moved to EventRequestParamsMapper?
For most of the modes it is just checking existence or not of orgUnits, for ALL mode we are checking the user and it shouldn't access the DB

muilpp: `ACCESSIBLE` and `CAPTURE` need the user to validate the org unit mode, and I wanted to keep all the unit mode validations together, that's the reason why they are in this mapper now.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -285,9 +284,7 @@ private String getOrgUnitModeViolation(EventOperationParams params, User user, P
...
       case ACCESSIBLE -> getAccessibleScopeValidation(user, program);
       case CAPTURE -> getCaptureScopeValidation(user);
-      case CHILDREN, SELECTED, DESCENDANTS -> params.getOrgUnitUid() == null
-          ? "Organisation unit is required for orgUnitMode: " + params.getOrgUnitMode()
-          : null;
+      default -> null;
```
enricocolasante, line 287: Shouldn't we have here now the validation for `ALL` mode?

userCanSearchOuModeALL(user)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -225,15 +231,23 @@ private void validateUpdateDurationParams(RequestParams requestParams)
     }
   }
 
-  // TODO Use RequestParamUtils.validateOrgUnitParams as soon /events accepts a list of org units
-  // UIDs
-  private void validateOrgUnitParams(UID orgUnit, OrganisationUnitSelectionMode orgUnitMode)
-      throws BadRequestException {
-    if (orgUnit != null && (orgUnitMode == ACCESSIBLE || orgUnitMode == CAPTURE)) {
+  private void validateOrgUnitMode(RequestParams params) throws BadRequestException {
+    if ((params.getOrgUnitMode().equals(ACCESSIBLE) || params.getOrgUnitMode().equals(CAPTURE))
```
enricocolasante, line 235: `params.getOrgUnitMode()` could be null at this point?
If yes, we could have an NPE here

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -112,6 +112,7 @@ public EventSearchParams map(EventOperationParams operationParams)
     ProgramStage programStage = validateProgramStage(operationParams.getProgramStageUid());
     OrganisationUnit requestedOrgUnit = validateRequestedOrgUnit(operationParams.getOrgUnitUid());
 
+    validateOrgUnitMode(operationParams, user, program);
     OrganisationUnitSelectionMode orgUnitMode =
```
enricocolasante, line 116: This default assignment of orgUnitMode in case something is null shouldn't be called before validation of the orgUnitMode?

What if the default assignment become `ACCESSIBLE` but the user has no orgUnits in search scope?

In my opinion we could move `OrganisationUnitSelectionMode orgUnitMode = getOrgUnitMode(requestedOrgUnit, operationParams.getOrgUnitMode());` in `RequestParamsMapper`, that is generally the place where we assign default values anyway

enricocolasante (commented):
A couple of doubts before :heavy_check_mark:

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
```diff
@@ -483,6 +481,8 @@ void shouldFailWhenOrgUnitSuppliedAndOrgUnitModeAccessible() {
     requestParams.setOrgUnit(UID.of(orgUnit.getUid()));
     requestParams.setOrgUnitMode(ACCESSIBLE);
 
+    when(organisationUnitService.getOrganisationUnit(orgUnit.getUid())).thenReturn(orgUnit);
```
teleivo, line 484: nit: I think we put the parameter to the method/function we test close to the call of the tested method. So the mock would go above the `requestParams` statements.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
```diff
@@ -495,42 +495,40 @@ void shouldFailWhenOrgUnitSuppliedAndOrgUnitModeCapture() {
...
     RequestParams requestParams = new RequestParams();
-    requestParams.setOrgUnit(UID.of(orgUnit.getUid()));
-    requestParams.setOrgUnitMode(SELECTED);
+    requestParams.setOrgUnitMode(OrganisationUnitSelectionMode.SELECTED);
 
-    EventOperationParams eventOperationParams = mapper.map(requestParams);
+    Exception exception = assertThrows(BadRequestException.class, () -> mapper.map(requestParams));
 
-    assertEquals(SELECTED, eventOperationParams.getOrgUnitMode());
+    assertStartsWith("orgUnit is required for orgUnitMode: SELECTED", exception.getMessage());
```
teleivo, line 512: Might be worth making one parameterized test for these 3 cases. Something like

```java
@ParameterizedTest
@EnumSource(value = OrgUnitMode.class, names = {"SELECTED", "DESCENDANTS", "CHILDREN"})
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: that's in the line 282, I can't put it here because CHILDREN, SELECTED and DESCENDANTS would be affected.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: Nice catch!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
muilpp: there's a condition on this method call, it's just called if orgUnitMode is not null.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
muilpp: my memory is really bad lately, so I could have easily have forgotten it, when did we decide this?

teleivo: I don't think we made a decision. Its just what I see us doing in all the mapper tests AFAIK 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -112,14 +112,15 @@ public EventSearchParams map(EventOperationParams operationParams)
     ProgramStage programStage = validateProgramStage(operationParams.getProgramStageUid());
     OrganisationUnit requestedOrgUnit = validateRequestedOrgUnit(operationParams.getOrgUnitUid());
 
-    validateOrgUnitMode(operationParams, user, program);
-    OrganisationUnitSelectionMode orgUnitMode =
-        getOrgUnitMode(requestedOrgUnit, operationParams.getOrgUnitMode());
+    if (operationParams.getOrgUnitMode() != null) {
```
teleivo, line 115: So far most validation functions get the `parameter` passed in regardless of whether its null or not. They then start with a guard clause exiting when there is nothing to validate. This particular function might already do a null check.

This makes the code in the main mapper function more readable IMHO.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -133,22 +133,26 @@ class EventOperationParamsMapperTest {
...
 
   @Test
   void shouldFailWithForbiddenExceptionWhenUserHasNoAccessToProgramStage() {
     programStage = new ProgramStage();
     programStage.setUid("PlZSBEN7iZd");
     EventOperationParams eventOperationParams =
-        EventOperationParams.builder().programStageUid(programStage.getUid()).build();
+        EventOperationParams.builder()
+            .programStageUid(programStage.getUid())
+            .orgUnitMode(ACCESSIBLE)
```
teleivo, line 154: Do we need to set this as it cannot be null and this test is testing the mapper. So the mapping in the web has not happened which would set the default?

Would it still make sense to add a default to the builder?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
```diff
@@ -369,11 +373,37 @@ void shouldReturnSelectedOrgUnitEventsWhenNoOuModeSpecifiedAndUserHasAccessToOrg
                     + " instead"));
   }
 
+  @Test
+  void shouldReturnAccessibleOrgUnitEventsWhenNoOuModeNorOrgUnitSpecifiedAndUserHasAccessToOrgUnit()
```
teleivo, line 377: Can this name be simplified to

```suggestion
  void shouldReturnAccessibleOrgUnitEvents()
```

the exact params are visible from the builder

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
```diff
@@ -369,11 +373,37 @@ void shouldReturnSelectedOrgUnitEventsWhenNoOuModeSpecifiedAndUserHasAccessToOrg
...
+
   @Test
   void shouldFailWhenNoOuModeSpecifiedAndUserHasNoAccessToOrgUnit() {
     injectSecurityContext(userService.getUser("FIgVWzUCkpw"));
     EventOperationParams params =
-        EventOperationParams.builder().programUid("pcxIanBWlSY").orgUnitUid("DiszpKrYNg8").build();
+        EventOperationParams.builder()
+            .programUid("pcxIanBWlSY")
+            .orgUnitUid("DiszpKrYNg8")
+            .orgUnitMode(SELECTED)
```
teleivo, line 405: The test name is 

`shouldFailWhenNoOuModeSpecifiedAndUserHasNoAccessToOrgUnit`

should probably be something like

`shouldFailWhenUserHasNoAccessToSelectedOrgUnit`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
```diff
@@ -384,7 +414,8 @@ void shouldFailWhenNoOuModeSpecifiedAndUserHasNoAccessToOrgUnit() {
   void shouldReturnAccessibleOrgUnitEventsWhenNoOrgUnitSpecifiedNorOuModeSpecified()
       throws ForbiddenException, BadRequestException {
     injectSecurityContext(userService.getUser("FIgVWzUCkpw"));
-    EventOperationParams params = EventOperationParams.builder().programUid("pcxIanBWlSY").build();
+    EventOperationParams params =
+        EventOperationParams.builder().programUid("pcxIanBWlSY").orgUnitMode(ACCESSIBLE).build();
```
teleivo, line 418: The test is called

`shouldFailWhenNoOuModeSpecifiedAndUserHasNoAccessToOrgUnit`

and should probably be called

`shouldFailWhenUserHasNoAccessToOrgUnit` or `shouldFailWhenUserHasNoAccessToOrgUnitGivenOrgUnitModeAccessible`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -68,11 +68,16 @@ class EventRequestParamsMapper {
...
+            requestParams.getOrgUnit(),
+            validateDeprecatedParameter(
+                "ouMode",
+                requestParams.getOuMode(),
+                "orgUnitMode",
+                requestParams.getOrgUnitMode()));
 
-    if (requestParams.getOrgUnitMode() != null) {
-      validateOrgUnitMode(requestParams);
+    if (orgUnitMode != null) {
```
teleivo, line 79: Same comment as in EventOperationParamsMapper `validateOrgUnitMode`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -231,23 +236,39 @@ private void validateUpdateDurationParams(RequestParams requestParams)
     }
   }
 
-  private void validateOrgUnitMode(RequestParams params) throws BadRequestException {
-    if ((params.getOrgUnitMode().equals(ACCESSIBLE) || params.getOrgUnitMode().equals(CAPTURE))
-        && params.getOrgUnit() != null) {
+  private void validateOrgUnitMode(UID orgUnit, OrganisationUnitSelectionMode orgUnitMode)
+      throws BadRequestException {
+    if ((orgUnitMode.equals(ACCESSIBLE) || orgUnitMode.equals(CAPTURE)) && orgUnit != null) {
```
teleivo, line 241: We should compare enums using `==` instead of `.equals` as `==` is null-safe. Even if you already checked that the mode cannot be null its a safer way to compare enums in general.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: Right. This check will be done in the web layer from now on, so here won't be necessary anymore because it can't be null at this point.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
muilpp: Yes, that's the reason why I need to set the org unit mode in each particular test.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
muilpp: I guess that's just my personal taste, but I like having the important details of the test in the title, so just by reading it I know what to expect.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
muilpp: Yeah. Anyway, with the latest changes this test is not useful anymore.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
muilpp: Same as above, this test will be deleted, it has no value anymore.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
teleivo: Yes probably subjective but to me the essence is getting the accessible org units 😅  Reapeating most params seems to redundant/verbose IMHO

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -189,15 +198,39 @@ private void validateUpdateDurationParams(RequestParams requestParams)
...
+    if ((orgUnitMode == ACCESSIBLE || orgUnitMode == CAPTURE) && orgUnit != null) {
       throw new BadRequestException(
           String.format(
               "orgUnitMode %s cannot be used with orgUnits. Please remove the orgUnit parameter and try again.",
               orgUnitMode));
     }
+
+    if ((orgUnitMode.equals(CHILDREN)
+            || orgUnitMode.equals(SELECTED)
+            || orgUnitMode.equals(DESCENDANTS))
```
enricocolasante, line 212:
```suggestion
    if ((orgUnitMode == CHILDREN
            || orgUnitMode == SELECTED
            || orgUnitMode == DESCENDANTS)
```

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -253,28 +251,25 @@ private void validateAttributeOptionCombo(CategoryOptionCombo attributeOptionCom
...
+  private void validateOrgUnitMode(
+      OrganisationUnitSelectionMode orgUnitMode, User user, Program program)
       throws BadRequestException {
-    if (params.getOrgUnitMode() != null) {
-      String violation = getOrgUnitModeViolation(params, user, program);
-      if (violation != null) {
-        throw new BadRequestException(violation);
-      }
+
+    String violation = getOrgUnitModeViolation(orgUnitMode, user, program);
```
teleivo, line 258: Can we inline this method `getOrgUnitModeViolation`? I think its only used once

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -303,6 +305,7 @@ void shouldMapAttributeFilters() throws BadRequestException, ForbiddenException
                     List.of(new QueryFilter(QueryOperator.EQ, "2")),
                     TEA_2_UID,
                     List.of(new QueryFilter(QueryOperator.LIKE, "foo"))))
+            .orgUnitMode(SELECTED)
```
teleivo, line 308: How can it be selected when there is no orgUnit?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -686,14 +685,8 @@ void shouldMapSearchScopeOrgUnitWhenOuModeAccessible()
     user.setOrganisationUnits(Set.of(searchScopeOrgUnit));
 
     when(currentUserService.getCurrentUser()).thenReturn(user);
-    when(organisationUnitService.getOrganisationUnit(orgUnit.getUid())).thenReturn(orgUnit);
 
-    EventOperationParams requestParams =
-        EventOperationParams.builder()
-            .programUid(program.getUid())
-            .orgUnitUid(orgUnit.getUid())
-            .orgUnitMode(ACCESSIBLE)
```
teleivo, line 695: Do we default to mode accessible? You removed it and the test says `shouldMapSearchScopeOrgUnitWhenOuModeAccessible`

Might have been my mistake at some point 🤷🏻 I see the use of `requestParams` name while this class should be `operationparams`. We could replace all with `params` in all the mapper tests then there is no confusion 😅

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -746,6 +739,7 @@ void shouldMapRequestedOrgUnitAsSelectedWhenOrgUnitProvidedAndNoOrgUnitModeProvi
         EventOperationParams.builder()
             .programUid(program.getUid())
             .orgUnitUid(orgUnit.getUid())
+            .orgUnitMode(SELECTED)
```
teleivo, line 742: The test name does not make sense I think

`shouldMapRequestedOrgUnitAsSelectedWhenOrgUnitProvidedAndNoOrgUnitModeProvided`

you do provide a mode now

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: It's true it's just used once, I just followed this mapper's pattern, above there are the methods `validateProgram`, `validateProgramStage` and `validateRequestedOrgUnit`, all of them also just used once.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
muilpp: nice one, this one should be accessible

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
muilpp: Yes, in this test case we default to ACCESSIBLE.
Good idea to not use requestParams.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
muilpp: Right, it's another test that has no value after the latest commits

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
teleivo: I mean inline method `getOrgUnitModeViolation`. You would still have `validateOrgUnitMode`

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -131,22 +134,25 @@ class EventOperationParamsMapperTest {
...
     orgUnit.setChildren(
         Set.of(
             createOrgUnit("captureScopeChild", "captureScopeChildUid"),
             createOrgUnit("searchScopeChild", "searchScopeChildUid")));
+
+    user = new User();
+    user.setOrganisationUnits(Set.of(orgUnit));
+    when(currentUserService.getCurrentUser()).thenReturn(user);
+
+    eventBuilder = eventBuilder.orgUnitMode(ACCESSIBLE);
```
teleivo, line 147: Can you please add a comment why its set to accessible. I am sure I won't know anymore next week 😅

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
teleivo: Not sure why we set the mode in the setup. It is clearer if its set here IMHO

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
```diff
@@ -107,6 +111,9 @@ void setUp() {
     // needed as some tests are run using another user (injectSecurityContext) while most tests
     // expect to be run by admin
     injectAdminUser();
+
+    operationParamsBuilder = EventOperationParams.builder();
```
teleivo, line 115: This test class is to test ordering/pagination of all enitites (right now its TE, EV).

So this should then probably be `eventOperationParamsBuilder`. We can shorten this to `eventParamsBuilder` IMHO

teleivo (approved):

PR #14801, opened 2023-08-01
chore: no need for QueryItem in handling filters TECH-1606

teleivo:
We found that our stores don't profit from using QueryItem. In addition, our APIs become simpler. We should therefore remove the use of QueryItem entirely.

Filters (attribute and data element ones) are now parsed in the web layer. This is the approach we are establishing for all parameters. The further down you go in our layers (web, service, repo) the more structured, easy to use the data should be. As we should take advantage of the knowledge we gain and the validations we've already done.

`filter=uid:op:value` transitions from `String` to `Map<String,List<QueryFilter>>` in the web and then `Map<TrackedEntityAttribute|DataElement,List<QueryFilter>>` in the service layer. (note this is pseudo-code).

`parseFilters` is replacing the tracked entity attribute, data element specific parse functions (currently in `OperationParamsUtils`). I had to make a few helpers public in `OperationParamsUtils` that will be moved to `RequestParamUtils` once the same changes are also made to `/tracker/trackedEntities?order=`.

* update OpenAPI spec on `order`, `filter` and  `filterAttributes`

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
@@ -32,10 +32,13 @@ import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ACCESSIBLE;
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.CAPTURE;
 import static org.hisp.dhis.tracker.export.enrollment.EnrollmentOperationParams.DEFAULT_PAGE;
 import static org.hisp.dhis.tracker.export.enrollment.EnrollmentOperationParams.DEFAULT_PAGE_SIZE;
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -88,6 +92,11 @@ public EventOperationParams map(RequestParams requestParams) throws BadRequestEx
             "event", requestParams.getEvent(), "events", requestParams.getEvents());
 
     validateFilter(requestParams.getFilter(), eventUids);
+    Map<String, List<QueryFilter>> dataElementFilters = new HashMap<>();
+    parseFilters(dataElementFilters, requestParams.getFilter());
```
enricocolasante, line 96: Is there a particular reason why we are populating the Map as a side effect instead of returning it?

enricocolasante (approved):

PR #14820, opened 2023-08-02
fix: drop foreign key [ TECH-1597 ]

lucaCambi77:
following https://github.com/dhis2/dhis2-core/pull/14803
We need to drop a foreign key before dropping `trackedentitymobilesetting`. This foreign key is in Sierra Leone but does not appear in the metadata


enricocolasante (approved):
I just wonder how the API tests were successful before?
The startup of the server for the tests should have failed

teleivo (approved):

PR #14827, opened 2023-08-03
fix: Fix broken footer URL link sent in email [DHIS2-13380]

david-mackessy:
back port of #14817


dhis-2/dhis-web-embedded-jetty/pom.xml
```diff
@@ -89,10 +89,6 @@
       <artifactId>log4j-slf4j-impl</artifactId>
       <scope>compile</scope>
     </dependency>
-    <dependency>
-      <groupId>org.springframework</groupId>
-      <artifactId>spring-orm</artifactId>
-    </dependency>
```
david-mackessy, line 95: Both of the removed deps in this pom were already removed in latest master #13710  & 2.40 #13757
Not sure why they weren't back ported. Maven build fails because these declared deps are not used.

vietnguyen (approved):

mortenoh (approved):

PR #14829, opened 2023-08-03
chore: simplify new tracked entities store TECH-1611

teleivo:
New tracker does not support grid queries. Once we remove the grid query code from the new tracker store we can simplify the SQL query to get TE ids.

The grid query returns data as row/column format using `Map<String,String>`.

https://github.com/dhis2/dhis2-core/blob/98be48af4bb6fe2b83cd2a45fdbd53b271fd861c/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/hibernate/HibernateTrackedEntityStore.java#L227-L255

In case there is any attribute present in a `filter` and maybe `attribute` and `order` (thus join) it would add an aggregation for `string_agg(TEA.uid || ':' || TEAV.value, ';@//@;') AS tea_values`. This aggregation required us to group by every column we wanted to return. This is obviously not needed for new tracker queries.

* Removed unused code.
* Removed createdAt/updatedAt from test fixture as they are automatically set by our code and have no effect other than to cause confusion.

I added additional tests for
* ordering by field name and attribute together with filtering
* ordering by the same attribute we also filter on
* pagination (which shows that we recently introduced a bug we will fix in https://dhis2.atlassian.net/browse/TECH-1601 see comment).

If you are curious you can read the following on the SQL we generate in the store.

## Details on the SQL query

The SQL query for getting tracked entity ids has at least these similarities to the one we do for events:
* inner sub-query to get TEs in correct order and paginated (limit/offset)
* order also specified on outer select as we cannot rely on ordering of the sub-query

#### Order by

```java
TrackedEntityOperationParams params =
    TrackedEntityOperationParams.builder()
        .orders(List.of(OrderCriteria.of("toUpdate000", SortDirection.ASC)))
```

turns into select in sub-query

```sql
"toUpdate000".value AS "toUpdate000"
```

adds a **LEFT** join (opposite to INNER join in events ordering by attributes). So here we do not filter out TEs when ordering by attributes they don't have.

```sql
LEFT JOIN trackedentityattributevalue AS "toUpdate000"
            ON "toUpdate000".trackedentityid = TE.trackedentityid AND
            "toUpdate000".trackedentityattributeid = 67
```

We add an additional left join on the trackedentityattribute table

```sql
LEFT JOIN trackedentityattributevalue TEAV ON TEAV.trackedentityid = TE.trackedentityid AND
                                            TEAV.trackedentityattributeid IN (67)
LEFT JOIN trackedentityattribute TEA
        ON TEA.trackedentityattributeid = TEAV.trackedentityattributeid
```

The sub-query gets an order by

```sql
ORDER BY "toUpdate000".value ASC
```

as well as the outer query

```sql
ORDER BY TE."toUpdate000" ASC
```

Previously we also added a

```sql
GROUP BY TE.trackedentityid, TE.uid, TE.created, TE.lastupdated, TE.ou, TE.ouname, TET.uid,
         TE.potentialduplicate, TE.inactive, TE."toUpdate000"
```

due to the grid aggregation of TEAV. Note that even though we in new tracker did not make a grid query the aggregation and group by were still present.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
@@ -702,7 +702,7 @@ public class DefaultTrackedEntityService implements TrackedEntityService {
 
   private void checkIfMaxTeiLimitIsReached(TrackedEntityQueryParams params, int maxTeiLimit) {
     if (maxTeiLimit > 0) {
-      int teCount = trackedEntityStore.getTrackedEntityCountForGridWithMaxTeiLimit(params);
+      int teCount = trackedEntityStore.getTrackedEntityCountForWithMaxTeiLimit(params);
 
       if (teCount > maxTeiLimit) {
         throw new IllegalQueryException("maxteicountreached");
@@ -726,7 +726,7 @@ public class DefaultTrackedEntityService implements TrackedEntityService {
 
     // using countForGrid here to leverage the better performant rewritten
     // sql query
-    return trackedEntityStore.getTrackedEntityCountForGrid(params);
+    return trackedEntityStore.getTrackedEntityCount(params);
   }
 
   /**
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
@@ -95,6 +95,7 @@ import org.springframework.stereotype.Repository;
 class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<TrackedEntity>
     implements TrackedEntityStore {
 
+  private static final String DEFAULT_ORDER = "TE.trackedentityid ASC";
   private static final String OFFSET = "OFFSET";
 
   private static final String LIMIT = "LIMIT";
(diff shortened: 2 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -702,7 +702,7 @@ private boolean isProgramMinAttributesViolated(TrackedEntityQueryParams params)
 
   private void checkIfMaxTeiLimitIsReached(TrackedEntityQueryParams params, int maxTeiLimit) {
     if (maxTeiLimit > 0) {
-      int teCount = trackedEntityStore.getTrackedEntityCountForGridWithMaxTeiLimit(params);
+      int teCount = trackedEntityStore.getTrackedEntityCountForWithMaxTeiLimit(params);
```
enricocolasante, line 705: Not part of this PR but as we are here, can we rename `Tei` to something else in the method names and the variables?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -95,6 +95,7 @@
 class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<TrackedEntity>
     implements TrackedEntityStore {
 
+  private static final String DEFAULT_ORDER = "TE.trackedentityid ASC";
```
enricocolasante, line 98: Didn't we agree that default order would be `DESC`?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
teleivo: Yes 😄 I am first adding tests and trying to change as little as needed. The next PRs will then implement whats missing.

muilpp (approved):

PR #14840, opened 2023-08-06
chore: create new tracker relationship store TECH-1535

teleivo:
this decouples us from old tracker allowing us to implement ordering and more as in other entities.


enricocolasante (approved):

teleivo:
> It won't be linked to a specific Jira issue
You won't have a pull request link in Jira. You will still be able to link the change from git to jira and thus why it was made.

> New relationship store seems similar to the older one with the only difference of having only 3 methods. Like we extracted 3 methods from the previous one. So what would happen with the existing one? are we keeping it or removing it?

That is the nature of copying. We need to detach ourselves from old tracker in order to make considerable changes. If we do not we only risk breaking old tracker code and even other teams code that might rely on the existing service/store. We will analyze what to do with the old services/stores. I am for: removing unused code in old servies/stores, moving code that is only used by old tracker into dxf2 packages, then figuring out if we need to support the remaining methods in new tracker 😄

PR #14859, opened 2023-08-07
chore: Refactor org unit modes ACL in service layer [TECH-1603]

muilpp:
First PR to harmonize the ACL validation in `/trackedEntities`, mostly focused on the service layer.

- Creation of the OperationsParamsValidation to validate the org unit mode. For now only used by tracked entities, but eventually enrollments and events should use it too.
- Rename orgUnits to accessibleOrgUnits in `TrackedEntityQueryParams`
- Fix tests broken by this refactor

Next PRs will focus on the web and persistance layers

Ticket: https://dhis2.atlassian.net/browse/TECH-1603

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.export;
+
+import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ACCESSIBLE;
+import static org.hisp.dhis.common.OrganisationUnitSelectionMode.CAPTURE;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.feedback.BadRequestException;
+import org.hisp.dhis.feedback.ForbiddenException;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.security.Authorities;
+import org.hisp.dhis.trackedentity.TrackerAccessManager;
+import org.hisp.dhis.user.User;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class OperationsParamsValidation {
+
+  public static void validateOrgUnitMode(
+      OrganisationUnitSelectionMode orgUnitMode, User user, Program program)
+      throws BadRequestException {
+
+    String violation =
+        switch (orgUnitMode) {
+          case ALL -> userCanSearchOrgUnitModeALL(user)
+              ? null
+              : "Current user is not authorized to query across all organisation units";
+          case ACCESSIBLE -> getAccessibleScopeValidation(user, program);
+          case CAPTURE -> getCaptureScopeValidation(user);
+          default -> null;
+        };
+
+    if (violation != null) {
+      throw new BadRequestException(violation);
+    }
+  }
+
+  private static String getCaptureScopeValidation(User user) {
+    String violation = null;
+
+    if (user == null) {
+      violation = "User is required for orgUnitMode: " + CAPTURE;
+    } else if (user.getOrganisationUnits().isEmpty()) {
+      violation = "User needs to be assigned data capture orgunits";
+    }
+
+    return violation;
+  }
+
+  private static String getAccessibleScopeValidation(User user, Program program) {
+    String violation;
+
+    if (user == null) {
+      return "User is required for orgUnitMode: " + OrganisationUnitSelectionMode.ACCESSIBLE;
+    }
+
+    // TODO If program null use search scope?
+    if (program == null || program.isClosed() || program.isProtected()) {
+      violation =
+          user.getOrganisationUnits().isEmpty()
+              ? "User needs to be assigned data capture orgunits"
+              : null;
+    } else {
+      violation =
+          user.getTeiSearchOrganisationUnitsWithFallback().isEmpty()
+              ? "User needs to be assigned either TE search, data view or data capture org units"
+              : null;
+    }
+
+    return violation;
+  }
+
+  private static boolean userCanSearchOrgUnitModeALL(User user) {
+    if (user == null) {
+      return false;
+    }
+
+    return user.isSuper()
+        || user.isAuthorized(Authorities.F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS.name());
+  }
+
+  /**
+   * Returns a list of all the org units the user has access to
+   *
+   * @param user the user to check the access of
+   * @param orgUnits parent org units to get descendants/children of
+   * @param orgUnitDescendants function to retrieve org units, in case ou mode is descendants
+   * @param program the program the user wants to access to
+   * @return a list containing the user accessible organisation units
+   * @throws ForbiddenException if the user has no access to any of the provided org units
+   */
+  public static Set<OrganisationUnit> validateAccessibleOrgUnits(
+      User user,
+      Set<OrganisationUnit> orgUnits,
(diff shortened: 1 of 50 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
```diff
@@ -0,0 +1,222 @@
...
+      return orgUnits.stream()
+          .filter(
+              availableOrgUnit ->
+                  user.getOrganisationUnits().stream()
+                      .anyMatch(
+                          captureScopeOrgUnit ->
+                              availableOrgUnit.getPath().contains(captureScopeOrgUnit.getPath())))
+          .collect(Collectors.toSet());
+    } else {
+      return orgUnits.stream()
```
teleivo, line 208: nit: since we return in the else I would remove the else and return instead (intellij will likely show you that cleanup option)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
```diff
@@ -0,0 +1,222 @@
...
+   */
+  private static Set<OrganisationUnit> getAccessibleDescendants(
+      User user, Program program, List<OrganisationUnit> orgUnits) {
+    if (orgUnits.isEmpty()) {
+      return Collections.emptySet();
+    }
+
+    if (isProgramAccessRestricted(program)) {
+      return orgUnits.stream()
+          .filter(
```
teleivo, line 200: Maybe there is a way to leverage that only `user.getOrganisationUnits()` and `user.getTeiSearchOrganisationUnits()` are different in the `filter` for the case of `isProgramAccessRestricted(program)` being true or false.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -286,7 +286,7 @@ private String getAccessibleScopeValidation(User user, Program program) {
       return "User is required for orgUnitMode: " + OrganisationUnitSelectionMode.ACCESSIBLE;
     }
 
-    if (program == null || program.isClosed() || program.isProtected()) {
+    if (program != null && (program.isClosed() || program.isProtected())) {
```
teleivo, line 289: was that wrong?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -196,19 +190,19 @@ public TrackedEntityQueryParams addOrganisationUnit(OrganisationUnit unit) {
    */
   public void handleOrganisationUnits() {
```
teleivo, line 191: Is there a way to get rid of this method? Does not have to be in this PR of course 😅 

When working on `order` there was something similar in I think it was events query params. The reason it existed was a poor API design. If some fields/collections need to be kept in sync or in a correct state space then the class itself should take care of that. We should not require clients to know that you need to initialize the class, set what you want and then call some `handleYYYY` method that then makes sure the fields are in a correct state.

For events you needed to know that if you want to order by an attribute you also add it to the attributes and filters collection of EventQueryParams 😓 Things like that need to be internal and happen on calls to in that case `orderBy(tea,asc)`.

Here it might be on adding an orgUnit 🤷🏻

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
```diff
@@ -0,0 +1,222 @@
...
+import org.hisp.dhis.feedback.BadRequestException;
+import org.hisp.dhis.feedback.ForbiddenException;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.security.Authorities;
+import org.hisp.dhis.trackedentity.TrackerAccessManager;
+import org.hisp.dhis.user.User;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class OperationsParamsValidation {
```
enricocolasante, line 52: Shouldn't this be a Validator more than Validation?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -322,7 +316,7 @@ public boolean hasFilters() {
 
   /** Indicates whether these parameters specify any organisation units. */
   public boolean hasOrganisationUnits() {
```
teleivo, line 318: Should we rename that as well?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
```diff
@@ -235,6 +239,7 @@ void shouldOrderTrackedEntitiesByCreatedAsc()
             .trackedEntityUids(Set.of("QS6w44flWAf", "dUE514NMOlo"))
             .trackedEntityTypeUid(trackedEntityType.getUid())
             .orderBy("created", SortDirection.ASC)
+            .user(importUser)
```
teleivo, line 242: nit: Please leave `orderBy` as the last call before build. That is what this test is about so should be easy to spot 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
```diff
@@ -0,0 +1,222 @@
...
+
+  public static void validateOrgUnitMode(
+      OrganisationUnitSelectionMode orgUnitMode, User user, Program program)
+      throws BadRequestException {
+
+    String violation =
```
enricocolasante, line 58: Maybe here I would think to remove the violation assignment and just directly throw the exception.
It seems to me that this kind of logic makes more sense when different cases and more of them can happen at the same. If you just throw you simplify the whole chain of validation calls

enricocolasante (approved):
Just a couple of style suggestion

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidation.java
```diff
@@ -0,0 +1,222 @@
...
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.security.Authorities;
+import org.hisp.dhis.trackedentity.TrackerAccessManager;
+import org.hisp.dhis.user.User;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class OperationsParamsValidation {
+
+  public static void validateOrgUnitMode(
```
teleivo, line 54: Javadocs might help us remember what the big picture of this is in the future 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: in case the program is null, we need to use the search scope by default

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
muilpp: The problem I have with this is that I want to break down this ticket into multiple PRs. 
I started with this one where I refactor the service, and to make sure it works as expected, I need to leave this ugly piece of code here. When I start working on the store, I'll remove it.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -0,0 +1,214 @@
...
+public class OperationsParamsValidator {
+
+  /**
+   * Validates the user is authorized and/or has the necessary configuration set up in case the org
+   * unit mode is ALL, ACCESSIBLE or CAPTURE. If the mode used is none of these three, no validation
+   * will be run
+   *
+   * @param orgUnitMode the {@link OrganisationUnitSelectionMode orgUnitMode} used in the current
+   *     case
+   * @param user
```
teleivo, line 61: please remove any parameters that you do not describe. Intellij will annoy us with linting warnings otherwise and we already have to many 😢

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -0,0 +1,214 @@
...
+   * @param program
+   * @throws BadRequestException if a validation error occurs for any of the three aforementioned
+   *     modes
+   */
+  public static void validateOrgUnitMode(
+      OrganisationUnitSelectionMode orgUnitMode, User user, Program program)
+      throws BadRequestException {
+
```
teleivo, line 69: nit: I don't add newlines at the start of functions. Would be great if we settle on one style. Unfortunately, this is not decided for us by the formatter 😂

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -0,0 +1,214 @@
...
+    switch (orgUnitMode) {
+      case ALL -> validateUserCanSearchOrgUnitModeALL(user);
+      case ACCESSIBLE -> validateAccessibleScope(user, program);
+      case CAPTURE -> validateCaptureScope(user);
+    }
+  }
+
+  private static void validateCaptureScope(User user) throws BadRequestException {
```
teleivo, line 77: nit: when we order our functions in call order it makes it easier to read the code from top to bottom 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    Set<OrganisationUnit> userOrgUnits =
+        isProgramAccessRestricted(program)
+            ? user.getOrganisationUnits()
+            : user.getTeiSearchOrganisationUnits();
+
+    return orgUnits.stream()
+        .filter(
+            availableOrgUnit ->
```
teleivo, line 204: nit: try to use shorter var names in lambdas like `o` for an orgUnit. Personally I find lambdas harder to read with var names I would use in a non-lambda form.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
muilpp: I would normally agree, but in this case I think writing a descriptive name helps understand what this lambda does. 
It iterates two different list of org units, and in my opinion, having a short var name would make it difficult to understand the list we are dealing with each time.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
muilpp: It's actually the formatter adding this new line.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
teleivo: The name of the collection var should give away what they are IMHO. Not sure if I remember correctly but maybe it was in the other PR that you had similar code with two different lambdas that were actually doing exactly the same but used different names for the lambda var. That itself obscures the lambda and prevents us from extracting generic functions 😅

PR #14866, opened 2023-08-08
chore: implement /tracker/relationships?order consistently with other endpoints TECH-1619

teleivo:
* validate field names are supported and not repeated in web layer
* validate fields advocated by web are actually supported by service/store
* map to List<Order> and map field names to Java field names (dhis-api org.hisp.dhis.relationship.Relationship class) in web layer
* stores do not need to handle order differently as we use hibernate to
  map from Java field names to DB columns
* add order and pagination tests
* hide classes if possible. Devs should depend on our services not stores and not any particular implementation
* order by primary key desc by default
* document order in OpenAPI spec
* only expose from our service whats used outside of the package (currently the only consumer is our exporter endpoint)
* service should have operation params in the signature, while stores should have the query params in their signature

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
@@ -34,7 +34,7 @@ import static org.hisp.dhis.common.SlimPager.FIRST_PAGE;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
+import java.util.Set;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.common.Pager;
 import org.hisp.dhis.common.SlimPager;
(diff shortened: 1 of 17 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -89,8 +88,8 @@ public Relationships getRelationships(RelationshipOperationParams params)
     return Relationships.of(relationships, pager);
   }
 
-  public int countRelationships(RelationshipQueryParams queryParams) {
-
+  private int countRelationships(RelationshipQueryParams queryParams) {
+    // TODO(ivo) should we pass something other than null?
```
teleivo, line 92: @enricocolasante do you have an opinion?

enricocolasante: I think for now it is ok. I would worry about this when we tackle the harmonization of the stores.

enricocolasante (approved):

muilpp (approved):

PR #14868, opened 2023-08-08
chore: Validate org unit mode in web layer [TECH-1603]

muilpp:
Second PR to harmonize the ACL validation in /trackedEntities, focusing now on the web layer.

Next PRs will focus on the persistance layers

Ticket: https://dhis2.atlassian.net/browse/TECH-1603


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Set;
+import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.common.UID;
+import org.hisp.dhis.feedback.BadRequestException;
+
+public class RequestParamsValidator {
+
+  /**
+   * Validates that no org unit is present if the ou mode is ACCESSIBLE or CAPTURE. If it is, an
```
enricocolasante, line 44: I would also specify that it will assign a default value if null

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParamsMapper.java
```diff
@@ -69,7 +69,7 @@ public EnrollmentOperationParams map(RequestParams requestParams) throws BadRequ
         validateDeprecatedParameter(
             "ouMode", requestParams.getOuMode(), "orgUnitMode", requestParams.getOrgUnitMode());
 
-    validateOrgUnitParams(orgUnits, orgUnitMode);
+    validateOrgUnitMode(orgUnits, orgUnitMode);
```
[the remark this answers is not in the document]
muilpp: Yes, but that's for enrollments. 
In this PR I'm focusing on tracked entities. I'd rather do that when we work on that endpoint, to avoid making unnecessary mistakes.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamUtils.java
```diff
@@ -153,24 +150,6 @@ private static Stream<String> parseUidString(String input) {
...
 
-  /**
-   * Validates that no org unit is present if the ou mode is ACCESSIBLE or CAPTURE. If it is, an
-   * exception will be thrown.
-   *
-   * @param orgUnits
-   * @param orgUnitMode
-   * @throws BadRequestException
-   */
-  public static void validateOrgUnitParams(
```
teleivo, line 164: I think you can rename this entire file to RequestParamsValidator as it only contains functions that validate request parameters 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
```diff
@@ -0,0 +1,114 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class RequestParamsValidator {
+
+  /**
+   * Validates that no org unit is present if the ou mode is ACCESSIBLE or CAPTURE. If it is, an
+   * exception will be thrown. If the org unit mode is not defined, SELECTED will be used by default
+   * if an org unit is present. Otherwise, ACCESSIBLE will be the default.
+   *
+   * @param orgUnits list of org units to be validated
+   * @param orgUnitMode
```
teleivo, line 52: Please check if intellij shows a warning if there is no description for this parameter. I think it does. We have tons of those 😬 I think you can remove the params that are obvious IMHO. I think that also removes the warning.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
```diff
@@ -0,0 +1,114 @@
...
+   * @throws BadRequestException
+   */
+  public static OrganisationUnitSelectionMode validateOrgUnitMode(
+      Set<UID> orgUnits, OrganisationUnitSelectionMode orgUnitMode) throws BadRequestException {
+
+    if (orgUnitMode == null) {
+      orgUnitMode = orgUnits.isEmpty() ? ACCESSIBLE : SELECTED;
```
teleivo, line 59: Can we return here? 🤔 If we can we should as then I don't have to read on if I am just interested in this default behavior.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
```diff
@@ -0,0 +1,114 @@
...
+   * an exception will be thrown. If the org unit mode is not defined, SELECTED will be used by
+   * default if an org unit is present. Otherwise, ACCESSIBLE will be the default.
+   *
+   * @param orgUnit the org unit to validate
+   * @param orgUnitMode
+   * @return
+   * @throws BadRequestException
+   */
+  public static OrganisationUnitSelectionMode validateOrgUnitMode(
+      UID orgUnit, OrganisationUnitSelectionMode orgUnitMode) throws BadRequestException {
```
teleivo, line 91: Can we reuse the above method by calling it from here with `validateOrgUnitMode(Set.of(orgUnit), mode)`?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParamsMapper.java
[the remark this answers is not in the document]
teleivo: You did make the change for events though 🤔 even though the PR description says trackedEntities 🤷🏻

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
muilpp: we can't because we need to validate if an orgUnit is present or not, depending on the resulting orgUnitMode of this condition.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
muilpp: I thought about it, the thing is that orgUnit could be null at this point, so I'd need to prevent an NPE and call it with an empty set in that case. 
Instead of doing that, and given that this method will be deleted as soon as /events accept a list of org units, I decided to leave it like it is.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParamsMapper.java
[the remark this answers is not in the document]
muilpp: Well yes, but the validations are the same.
What I don't want to do is change the value assigned to the `orgUnitMode`, because that could lead to errors I don't want to deal with in this PR.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
teleivo: Is `orgUnits.isEmpty()` not checking if an orgUnit is present?

So if `orgUnits.isEmpty()` the mode is ACCESSIBLE and there is no need to validate further and if `!orgUnits.isEmpty()` its SELECTED and there is no need to validate further.

muilpp: You are right, I misunderstood the first time.

teleivo (approved):

PR #14873, opened 2023-08-09
fix: Fix Dimensions Pager info [DHIS2-15613]

david-mackessy:
# Summary
`Dimension`s `Pager` information was broken. The following fields were not working/present:
- `total` (always `0`)
- `pageCount` (always `0`)
- `nextPage` (always missing)
- `prevPage` (always missing)

## Old behaviour
`Dimension`s `Pager` information always looked like this with the `total` and `pageCount` fields always being `0` and `nextPage` and `prevPage` fields always missing.
```json
"pager": {
        "page": 1,
        "total": 0,
        "pageSize": 13,
        "pageCount": 0
    }
```

## New behaviour
`Dimension`s `Pager` information now working as expected e.g.
```json
"pager": {
        "page": 2,
        "total": 52,
        "pageSize": 13,
        "nextPage": "http://localhost:9090/api/dimensions?page=3&pageSize=13",
        "prevPage": "http://localhost:9090/api/dimensions?pageSize=13",
        "pageCount": 4
    }
```

## Cause
- the `AbstractFullReadOnlyController` was handling the `GET` call for `Dimension`s
  - in that process, it looks for an `InternalHibernateGenericStore` of which there is none for `BaseDimensionalObject`
  - when no store is found, it impacts the paging part of the process
- The `Dimension`s call retrieves 5 different types of `BaseDimensionalObject` objects
  - `CategoryOptionGroupSet`
  - `DataElementCategory`
  - `DataElementGroupSet`
  - `OptionGroupSet`
  - `OrganisationUnitGroupSet`

## Change
- `DimensionController` now handles the `GET` call for `Dimensions` so we can control the different paging requirements
- added `PaginationUtils` method

## Notes
- I experimented updating the `AbstractFullReadOnlyController` to handle the bug but felt it was too risky given the possible large impact area affected
- it also makes sense to have any custom handling in its own controller

# Testing

## Automated
- Dimension web call tests added
- pagination utils tests added
- added pager links assertion method

## Manual
- Tested locally with 52 dimensions
- tested different request param options e.g.
  - page = 2
  - pageSize = 2
  - paging = false
  - fields = id

To view dimensions use
`GET /api/dimensions`
Response should look something like this:

```json
{
    "pager": {
        "page": 1,
        "total": 52,
        "pageSize": 50,
        "nextPage": "http://localhost:9090/api/dimensions?page=2",
        "pageCount": 2
    },
    "dimensions": [
        {
            "displayName": "Rural and Urban",
            "id": "EC40NXmsTVu"
        }
      ]
}
```
The `total`, `pageCount`, `nextPage` & `prevPage` fields should now be working as expected.
Calls with different query params look like this:
- show page 2
`GET /api/dimensions?page=2`

- request a specific page size
`GET /api/dimensions?pageSize=10`

- request no paging
`GET /api/dimensions?paging=false`

- request specific fields
`GET /api/dimensions?fields=id`

These calls can all be used together in different combinations e.g.
`GET /api/dimensions?page=2&pageSize10&fields=id`


netroms (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dimension/DimensionController.java
```diff
@@ -254,4 +279,37 @@ public ResponseEntity<JsonRoot> getDimensionsForDataSet(
...
+    List<String> fields = Lists.newArrayList(contextService.getParameterValues("fields"));
+    List<String> filters = Lists.newArrayList(contextService.getParameterValues("filter"));
+
+    if (fields.isEmpty()) {
+      fields.addAll(Preset.defaultPreset().getFields());
+    }
+
+    WebOptions options = new WebOptions(rpParameters);
+
+    if (!aclService.canRead(currentUser, getEntityClass())) {
```
vietnguyen, line 306: I think this will always return `true` as the `DimensionalObject` interface is returned from ` getEntityClass()` and this interface doesn't have and sharing setting.

`DimensionService.getAllDimensions()` already applied ACL check for each queries so only readable objects will be returned.

vietnguyen (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dimension/DimensionController.java
david-mackessy: cheers, I'll remove this check. `Schema` does return `null` for `DimensionalObject` so it does always evaluate to `true`.

PR #14902, opened 2023-08-16
feat: use icons/{uid}/icon to download custom icon

zubaira:
[DHIS2-15409](https://dhis2.atlassian.net/browse/DHIS2-15409)

[DHIS2-15409]: https://dhis2.atlassian.net/browse/DHIS2-15409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -88,12 +100,18 @@ public class IconController {
...
+  }
 
-    response.setHeader("Cache-Control", CacheControl.maxAge(TTL, TimeUnit.DAYS).getHeaderValue());
-    response.setContentType(MediaType.SVG_UTF_8.toString());
+  @GetMapping(value = "/{uid}/icon")
+  public void getCustomIconData(@PathVariable String uid, HttpServletResponse response)
+      throws NotFoundException, WebMessageException, IOException {
 
-    StreamUtils.copyThenCloseInputStream(icon.getInputStream(), response.getOutputStream());
+    if (iconService.iconExists(uid)) {
```
teleivo, line 110: This code is confusing to me 😅 How is it that if an `iconExists` it is a default icon and otherwise a custom icon? This is not obvious from an `iconService` that to me is returning `icons` 🤷🏻 Maybe @muilpp can clarify

Even more confusing is that
* the method is called `getCustomIconData` but also returns default icons.
* and `getIconData` only `downloadDefaultIcon` with a path of `"/{iconKey}/icon.svg"`. How does that path convey that to a user? When the path `"/{uid}/icon"` returns default and custom icons 😅

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -143,4 +161,42 @@ public void getIconData(HttpServletResponse response, @PathVariable String iconK
     webMessage.setMessage(String.format("Icon %s deleted", iconKey));
     return webMessage;
   }
+
+  private void downloadDefaultIcon(String defaultIcon, HttpServletResponse response)
```
teleivo, line 165: These methods are new and I don't see any new tests for it. I assume it is not tested but should be.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -143,4 +161,42 @@ public void getIconData(HttpServletResponse response, @PathVariable String iconK
...
+      throws IOException, NotFoundException {
+    Resource icon = iconService.getDefaultIconResource(defaultIcon);
+
+    response.setHeader("Cache-Control", CacheControl.maxAge(TTL, TimeUnit.DAYS).getHeaderValue());
+    response.setContentType(MediaType.SVG_UTF_8.toString());
```
teleivo, line 170: `downloadDefaultIcon` is used in `"/{iconKey}/icon.svg"` and `"/{uid}/icon"`. In the latter we also return custom icons. I just read about `png` format being used for custom icons in the issue description. Is it ok for users that the custom icons will be an png while default icons will be a svg in `"/{uid}/icon"`? I know we set the `content-type`. Its more about the usability of this feature.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: default icons are bundled icons identified by a key. The requirement is to have a single endpoint for both 
custom and default icons. So if the key exists it means it's a default icon otherwise find it in the custom icon list.
As far as `getCustomIconData` name is concerned it can be changed `getCustomOrDefaultIConData`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: @stian-sandvold

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
muilpp: @zubaira, if I understand this piece of code correctly, you expect `iconService.iconExists(uid)` to return true only if a default icon exists with that given uid, but [the implementation of this method](https://github.com/dhis2/dhis2-core/blob/DHIS2-15409/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java#L134) checks both default icons and custom icons, meaning that if any of these two kind of icons has that uid, it will return true.

zubaira: good catch @muilpp. Fixed it

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
stian-sandvold: It's a good question @teleivo and I was hoping to merge this and leave the responsibility of interpreting the type of image to the client. png is just one supported type for custom icons; So there will always be some need to figure out the type.

The idea here was to serve _any_ icon through the same endpoint as well, so I think this is a natural problem clients will have to deal with to be able to fetch anything from the same endpoint.




I'm happy to hear any suggestions on alternative solutions :D Also, I am hoping to have this merged soon, so we can have android and web start working on supporting this new endpoint in their apps (If this is how we end up doing it)

stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -88,12 +102,20 @@ public class IconController {
...
   public void getIconData(HttpServletResponse response, @PathVariable String iconKey)
       throws IOException, NotFoundException {
-    Resource icon = iconService.getDefaultIconResource(iconKey);
+    downloadDefaultIcon(iconKey, response);
+  }
 
-    response.setHeader("Cache-Control", CacheControl.maxAge(TTL, TimeUnit.DAYS).getHeaderValue());
-    response.setContentType(MediaType.SVG_UTF_8.toString());
+  @GetMapping(value = "/{key}/icon")
+  public void getCustomIconData(@PathVariable String key, HttpServletResponse response)
```
teleivo, line 109: The method names are confusing to me at least: there are default icons that come with DHIS2 and custom icons users can upload. This method serves both but is called `getCustomIconData`. The `/{iconKey}/icon.svg` endpoint is handled by `getIconData` but only returns default icons.

Is `/{iconKey}/icon.svg` supposed to only return default icons? Can custom icons also be stored as svgs? If `/{iconKey}/icon.svg` is only returning default icons the method should be renamed to something like `getDefaultIconData`.

`/{key}/icon` returns both types of icons while the method is called `getCustomIconData`. `getIconData` might be clearer 😅 .

muilpp: I agree with Ivo in that if the method returns both types of icons, it would make more sense to rename it to `getIconData`.

Since we have this new GET endpoint to download icons, seems like a smart decision to group both default and custom icons in here, but then, is it still useful to expose `/{iconKey}/icon.svg`?

JoakimSM (commented):
Should work for web as is 👍But, the href property returned from the icons endpoint is incorrect for custom icons. It is still being computed using the fileResourceId, giving a 404 if you try to use the url directly.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
stian-sandvold: I agree, we can rename the existing to "getCoreIcon" or something like that, to specify the kind of icon it returns, while the new could be "getIcon", since it returns any icon.

We should keep the iconKey/icon.svg endpoint to be backwards compatible, but mark it as deprecated.

teleivo: @stian-sandvold did `iconKey/icon.svg` exist before? I thought this custom icon was all new 🤔

zubaira:
> Should work for web as is 👍But, the href property returned from the icons endpoint is incorrect for custom icons. It is still being computed using the fileResourceId, giving a 404 if you try to use the url directly.

@JoakimSM ....fixed !!

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: fixed

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
stian-sandvold: Custom icon is all new, but the iconKey/icon.svg is how we serve the bundled icons :)

PR #14904, opened 2023-08-16
fix: Misleading message about password requirements

netroms:


david-mackessy (approved):
is there a test with a password of length greater than 60? couldn't see one while browsing the PR

netroms:
> is there a test with a password of length greater than 60? couldn't see one while browsing the PR

Good observation, added one now.

PR #14946, opened 2023-08-23
feat: user datastore query API [DHIS2-15723]

jbee:
### Summary
Updates the `/api/userDataStore` to support the same `fields` and `filter` parameters as `/api/dataStore` API. 

In this context the other methods in the endpoint were lined up so they behave equally in both APIs (following how `/dataStore` did behave). 

Also renamed some methods to line up better and used the feedback exceptions.

### Automatic Testing
New tests were added based on the existing tests for the `/dataStore` API just now for the `/userDataStore` API.
Existing tests were adjusted to expect the common messages.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractDatastoreController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractDatastoreController.java
@@ -46,7 +46,7 @@ import org.hisp.dhis.datastore.DatastoreFields;
 import org.hisp.dhis.datastore.DatastoreQuery;
 import org.hisp.dhis.datastore.DatastoreQuery.Field;
 import org.hisp.dhis.dxf2.webmessage.WebMessageUtils;
-import org.hisp.dhis.feedback.BadRequestException;
+import org.hisp.dhis.feedback.ConflictException;
 import org.hisp.dhis.webapi.JsonWriter;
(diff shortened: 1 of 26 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractDatastoreController.java
```diff
@@ -0,0 +1,127 @@
...
+        response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+        Throwable cause = ex.getCause();
+        String msg =
+            "Unknown error when running the query: "
+                + (cause != null && ex.getMessage().contains("could not extract ResultSet")
+                    ? cause.getMessage()
+                    : ex.getMessage());
+        if (cause != null
+            && cause.getMessage().contains("cannot cast type ")
+            && cause.getMessage().contains(" to double precision")) {
```
david-mackessy, line 112: going above and beyond to output a good error message which helps, nice.

david-mackessy (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractDatastoreController.java
jbee: Haha, I moved that from the `DatastoreController` without noticing it. I did good when I was implementing it originally :)

vietnguyen (approved):

PR #14969, opened 2023-08-28
feat: system setting for CRON delayed execution [DHIS2-15747]

jbee:
### Summary 
Introduces a new system setting for the maximum number of hours that a CRON based job will execute after the intended time of the day when delayed. After that time of the day the execution for the day is skipped and the next execution will be on the intended time of the next day (or next execution according to the CRON). 

It is important to note that the `nextExectionTime` property as computed by `SchedulerEntry` in the controller will also make use of the system setting so it states the correct next execution for the current system setting. The `getNextExecutionTime()` method in `JobConfiguration` will not be correct in case of a missed execution case that leads to a skip as it has no access to the system setting and therefore assumes maximum delay of a day. 

### Automatic Testing
Added a unit test to test the data-time logic


netroms (approved):

vietnguyen (approved):

david-mackessy:
- Are there scenarios you're aware of where this will be really helpful? 
- is there a default value for the max cron delay if not set?

PR #14977, opened 2023-08-28
fix: remove previous version of an app when installing [DHIS2-15766]

jbee:
### Summary
When installing an app we remove other version of the app if installed. 
We only have to consider jclouds as all installed apps are managed there.

I reversed the priority given to merge so that apps in jcloud always win over local apps to allow overriding a bundled app. This is validated anyhow when installing so if a user manages to do so the installed version should win I guess.

I also removed maintaining a list/map of reserved namespaces as this becomes a problem to maintain consistently and mostly this was not really used. The only place this was used is in validation to find if an app already uses a namespace. For this we could just iterate the cache.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -431,7 +442,10 @@ public void deleteApp(App app) {
       blobStore.removeBlob(config.container, resource.getName());
     }
 
-    reservedNamespaces.remove(app.getActivities().getDhis().getNamespace(), app);
+    String namespace = app.getActivities().getDhis().getNamespace();
+    if (namespace != null) {
```
jbee, line 446: FYI: Not sure why nobody ran into this NPE before because most apps do not have a namespace and this then would cause a NPE when trying to remove a `null` key from the map. Maybe this just "worked" because this was the last step so even if it failed with a NPE - at that point the deletion itself was already performed so the effect was still there just with an error message.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -328,14 +322,12 @@ public void reloadApps() {
         .filter(app -> !exists(app.getKey()))
         .forEach(this::installApp);
 
-    jCloudsAppStorageService.discoverInstalledApps().values().stream()
-        .filter(app -> !exists(app.getKey()))
-        .forEach(this::installApp);
+    jCloudsAppStorageService.discoverInstalledApps().values().forEach(this::installApp);
```
jbee, line 325: FYI: I effectively removed the filtering to only install what isn't already installed so that jclouds would win over local store.

Philip-Larsen-Donnelly:
I tested this locally with docker, but I still get multiple versions of apps in the filestore as I install them:

```
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:31 app-management_100.2.23
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:31 app-management_100.2.25
drwxr-x---  4 nobody nogroup 4096 Aug 28 17:33 dashboard_100.0.0
drwxr-x---  5 nobody nogroup 4096 Aug 28 17:34 maps_100.1.5
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:35 reports_100.0.18
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:35 reports_100.0.23
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:35 reports_100.0.6
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:32 reports_100.0.7
drwxr-x---  3 nobody nogroup 4096 Aug 28 17:35 reports_100.0.8
```

(I checked I was using the correct build [531a734])

Philip-Larsen-Donnelly:
Note: the first time this "fix" runs on an instance, there may be several old versions of a given app that should be removed, but the comments in this PR seem to imply we are only trying to remove one previous version.

david-mackessy:
I know we're dealing with the file system here but is there any way to use mocks and get some meaningful tests to confirm the new behaviour?

jbee:
I will have a think about what we could do to take care of all existing versions.

jbee:
@Philip-Larsen-Donnelly I updated it now and it deletes all versions of the same app. But it takes quite some time before the folders are actually removed. Not sure what is going on there. It feels quite funky when using the app and for some time after you clicked install and maybe navigate to the list of custom apps the old version is still showing and the folders also do exist (not the files in those folders actually) and then the app shows the "successful" message and the version changes and the folders are gone.

Philip-Larsen-Donnelly:
Hi @jbee 

Thanks for the update. I've been testing this a bit today and it does take a long time, but seems to get up to date eventually.

I was trying to understand why jclouds was so slow, so I played around a bit myself and was considering the following alteration to the deleteApp method

1. blobStore.deleteDirectory is really fast (instant) on the local filestore, and has no impact on S3, so we could run it before looping through the files. It is deprecated, unfortunately - so perhaps that excludes it from consideration?

2. Due to the slow speed, it's possible that a system crash, during delete, could result in a corrupt app artifact being loaded on next startup. I wonder if we could avoid this by simply deleting the manifest.webapp file before looping over the rest of the "resources"?

e.g.

```
 @Override
  public void deleteApp(App app) {
    log.info("Deleting app " + app.getName());

    // Delete all files related to app
    // fast but deprecated (works for local filestore):
    blobStore.deleteDirectory(config.container, app.getFolderName());
    
    // slower but works for S3:
    // delete the manifest file first in case the system crashes during deletion
    // and the manifest file is not deleted, resulting in an app that can't be installed
    blobStore.removeBlob(config.container, app.getFolderName()+"/manifest.webapp");
    for (StorageMetadata resource :
        blobStore.list(config.container, prefix(app.getFolderName()).recursive())) {
      log.debug("Deleting app file: " + resource.getName());

      blobStore.removeBlob(config.container, resource.getName());
    }
    log.info("Deleted app " + app.getName());
  }
```

jbee:
@Philip-Larsen-Donnelly Sure, thanks for the suggestion.

Philip-Larsen-Donnelly (approved):

stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
david-mackessy: The use of HashMap allows nulls so this was handled by the library I think. It never would've propagated. I see it's been updated to use ConcurrentHashMap which would throw a NPE.

PR #14991, opened 2023-08-30
fix: solve issue where controller methods falling under the same URL …

claudemamo:
…path and have matching HTTP methods are merged despite having different query params (see https://github.com/dhis2/notes/discussions/343)

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiFinalise.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiFinalise.java
@@ -32,6 +32,7 @@ import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toSet;
 
 import java.lang.reflect.Member;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.EnumMap;
 import java.util.HashMap;
@@ -40,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.function.Function;
 import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiFinalise.java
```diff
@@ -473,24 +475,106 @@ private static String getSharedNameForDeclaringType(Api.Parameter p) {
...
+                                      .add(e)));
+
+              for (Map.Entry<RequestMethod, List<Api.Endpoint>> methodAndEndpoints :
+                  endpointByMethod.entrySet()) {
+                toPathsAndOperations(methodAndEndpoints, path);
+              }
             });
   }
 
+  private void toPathsAndOperations(
```
jbee, line 501: Better have this return a map with the `String` fragment as a key, and the `Api.Endpoint` as a value and then fill  the map (that part)
```java
api.getEndpoints()
          .computeIfAbsent(path, s -> new EnumMap<>(RequestMethod.class))
          .put(methodAndEndpoints.getKey(), nonFragmentEndpoint);
```
at the caller. That should make it unnecessary to get the `path` involved and splits this into a function producing a value based on its input and than the caller using that to again do what it said, the group and merge. This way it becomes layers of nesting steps. Like this we need to read both sides to understand the code.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiFinalise.java
```diff
@@ -473,24 +475,106 @@ private static String getSharedNameForDeclaringType(Api.Parameter p) {
...
+        return methodAndEndpoints.getValue().get(i);
+      }
+    }
+
+    return methodAndEndpoints.getValue().get(0);
+  }
+
+  private AbstractMap.SimpleEntry<String, EnumMap<RequestMethod, Api.Endpoint>> fragmentise(
+      RequestMethod requestMethod, Api.Endpoint endpoint, String path) {
+    EnumMap<RequestMethod, Api.Endpoint> requestMethodObjectEnumMap =
```
jbee, line 543: `requestMethodObjectEnumMap` is a bit unnecessary long and hard to read. we do not have to repeat information in a name that is apparent from the type. Use names to add information the types do not provide. If we don't have more to say make it as short as possible without being confusing. E.g. here `endpoints` would be fine. Also feel free to use `Map.of(x,y)` even for places where we could use a `EnumMap`. It is just so much more readable and for a single entry map there is no difference in cost.

jbee (changes requested):
I wanted to understand and check the algorithm used for the fragments but this was really hard to read and took me a long time to review. So I felt compelled to leave more comments then usual on how to improve the readability. This being one of the key algorithms in the OpenAPI processing it also is important this code is readable.

The algorithm should do what we talked about :+1: ... I think :D

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiFinalise.java
```diff
@@ -473,24 +475,106 @@ private static String getSharedNameForDeclaringType(Api.Parameter p) {
...
+        new EnumMap<>(RequestMethod.class);
+    requestMethodObjectEnumMap.put(requestMethod, endpoint);
+
+    return new AbstractMap.SimpleEntry<>(
+        path + "#" + endpoint.getName(), requestMethodObjectEnumMap);
+  }
+
+  private boolean isMergeable(Api.Endpoint anEndpoint, Api.Endpoint anotherEndpoint) {
+    return anEndpoint.getParameters().equals(anotherEndpoint.getParameters());
```
jbee, line 552: Have you tested that this can be true at all? I mean is `equals` implemented in a way that this has a chance of being true if they have same list of parameters? What about order?

claudemamo: Yes, equals is true if parameter A matches parameter B in terms of the signature. Wrt the order, all endpoint parameters should be ordered the same way given that the Map implementation is a TreeMap.

PR #14992, opened 2023-08-30
fix: Validate org unit against search scope in events [TECH-1630]

muilpp:
`/events` needs to always validate any requested org unit against the search scope, regardless of the program access level.
It will be in the persistence layer where the capture scope will be used if the program is closed or protected.

Ticket: https://dhis2.atlassian.net/browse/TECH-1630


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -200,17 +186,20 @@ private OrganisationUnit validateRequestedOrgUnit(String orgUnitUid) throws BadR
...
     }
-
     return orgUnit;
   }
 
-  private void validateUser(User user, Program program, ProgramStage programStage)
+  private void validateUser(
+      User user, Program program, ProgramStage programStage, OrganisationUnit requestedOrgUnit)
       throws ForbiddenException {
+    if (user == null) {
```
enricocolasante, line 199: I am confused by this check.
When we have a null user doesn't mean that we are performing the operation as a super user? In other parts of the system that is true, or at least was true.
It seems to me that it is not responsibility of this mapper to check if a user is logged in or not.

muilpp: The way I understood this, the user is a superuser if the condition below is true, which checks that the authorities the user has correspond to the one of a superuser. If the user is null, it will have no authorities and for that reason, it should not be considered a superuser. Can you point me to these other parts of the code where we assume a user is a superuser if null?

jbee: Yes, this goes back to the scheduler not having a user when running as admin. I think I fixed this now in 2.41 but it will not get backported so ATM it feels best to keep the `null` check.

muilpp: @jbee if this is fixed in 2.41, is it safe to assume a null user is not a superuser anymore? this way we can get rid of some unnecessary logic going forward.

jbee: I have not tested it and I am unsure about the `User` vs. `CurrentUserDetails`. I do make sure there are `CurrentUserDetails` present when running as admin but I do not use an actual `User`. But that may be the better way to do it. I have to talk to @netroms about this after I am back from vacation.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
```diff
@@ -423,6 +426,41 @@ void shouldReturnNoEventsGivenUserHasNoAccess() throws ForbiddenException, BadRe
...
+    injectSecurityContext(userService.getUser("lPaILkLkgOM"));
+
+    EventOperationParams params =
+        EventOperationParams.builder().orgUnitUid("uoNW0E3xXUy").orgUnitMode(ALL).build();
+
+    List<Event> events = eventService.getEvents(params).getEvents();
+
+    assertFalse(
+        events.isEmpty(),
+        "Expected to find events when ou mode ALL no program specified and no org unit provided");
```
teleivo, line 458:
```suggestion
        "Expected to find events when ou mode ALL no program specified and org unit provided");
```

Does it make sense to accept mode ALL with an orgUnit? Because in a RequestParamValidator you reject a request like that with `orgUnitMode %s cannot be used with orgUnits. Please remove the orgUnit parameter and try again.`

Or is it due to the layers and that this layer is testing via the service (operation params)? So the web layer adds the orgUnit to the params even for mode ALL. What orgUnit will that be then?

teleivo (commented):
A general observation on our SQL is that I would not write capitalized SQL queries. I am not sure that we have a style guide for that but I think that is an outdated style 😅 increasing the RSI risk 😂

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/AclEventExporterTest.java
muilpp: You are right that the API will reject a request containing an org unit if the mode is ALL.

However, it is my understanding that we have jobs running as admin calling the service directly, thus bypassing the the web layer validations. 
For that reason I added a test to validate that even if an org unit is used with mode ALL, this org unit will be ignored when retrieving the results from the database.

I changed the test name to `shouldIgnoreRequestedOrgUnitAndReturnAllEventsWhenOrgUnitModeAllAndOrgUnitProvided`, maybe it's clearer this way.

teleivo (approved):

enricocolasante (approved):

PR #14993, opened 2023-08-30
fix: Update PUT in data stores to allow create or update [DHIS2-15746]

david-mackessy:
# Summary
Clients of the `dataStore` and `userDataStore` APIs sometimes had to make 1-2 calls to create or update an entry.

Examples from JIRA:
Currently if the client is not aware of whether or not an entry exists they have to do 1-2 calls. Either they assume it does not exist and try to `POST` it, which might fail when it exists which then requires a `PUT` call to update the entry. Or the client does a `PUT` first assuming it does exist. If that fails as it does not exist a `POST` is needed to create it.

## Change
The `PUT` mapping now behaves like the intended standard as set out in the [internet standard definition docs](https://www.rfc-editor.org/rfc/rfc9110.html#name-put)

new behaviour scenarios:
- `PUT` call when entry does not exist -> create new entry attempted (`201` status when created)
- `PUT` call when entry does exist -> update entry attempted (`200` status when updated)

APIs updated:
- `dataStore`
- `userDataStore`

This requires the `PUT` API to be able to receive a new `encrypt` boolean request parameter, as this is required in the `create` flow.

# Testing
## Automated
- new tests added to cover new behaviour
- some existing tests updated to reflect new behaviour

## Manual
To test the new behaviour, make calls to these endpoints, trying with existing and new keys:
- `PUT /api/dataStore/{namespace}/{key}` with JSON body
- `PUT /api/userDataStore/{namespace}/{key}` with JSON body

To check entries use:
- `GET /api/dataStore/{namespace}/{key}` 
- `GET /api/userDataStore/{namespace}/{key}`


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatastoreControllerTest.java
```diff
@@ -473,6 +477,78 @@ void testDeleteKeyJsonValue_ProtectedNamespaceWithSharing() {
...
+  void testPutEntry_EntryDoesNotExistAndIsCreated() {
+    assertStatus(HttpStatus.CREATED, PUT("/dataStore/pets/emu", "{\"name\":\"harry\"}"));
+    JsonDatastoreValue emu = GET("/dataStore/pets/emu").content().as(JsonDatastoreValue.class);
+    assertEquals("harry", emu.getString("name").string());
+  }
+
+  @Test
+  void testPutEntry_EntryAlreadyExistsAndIsUpdated() {
+    assertStatus(HttpStatus.CREATED, PUT("/dataStore/pets/emu", "{\"name\":\"harry\"}"));
+    assertStatus(HttpStatus.OK, PUT("/dataStore/pets/emu", "{\"name\":\"barry\"}"));
```
jbee, line 490: Nothing wrong with adding a bit of humor. Just that `harry` and `barry` is maybe not the best choice of values to show clearly what the difference is as they are very similar.

jbee (approved):
Nice!

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatastoreControllerTest.java
david-mackessy: 😄 yep, i'll update it

netroms (approved):

PR #15025, opened 2023-09-01
fix: Check for case-insensitive username duplicates

netroms:


david-mackessy:
so this checks for usernames with the same name (ignoring case).
does that mean we allowed previous usernames to exist with different cases? (e.g. John & joHN)
are we just accepting that previous state and only preventing from this point on? (no cleanup or warnings to users?)

vietnguyen (approved):

PR #15038, opened 2023-09-04
fix: Use search scope in /events requests [TECH-1630] [2.40]

muilpp:
`/events` need to use the search scope to validate the org unit in any incoming requests, regardless of the program access level.
This means that we will consider a request to be valid as long as the org unit in the request is in the user's search scope.
If the program is restricted (protected or closed), we will make sure to use the capture scope when we run the sql query, and worst case we would return an empty collection when no org units are present in the user capture scope, but we would not return an exception, as we did so far.

Ticket: https://dhis2.atlassian.net/browse/TECH-1630


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/event/EventQueryParams.java
```diff
@@ -115,9 +115,9 @@ public class EventQueryParams {
   // TODO Default set to DESCENDANTS to replicate master, but this will need to be fixed in
   // https://dhis2.atlassian.net/browse/TECH-1588
   private OrganisationUnitSelectionMode orgUnitSelectionMode =
-      OrganisationUnitSelectionMode.DESCENDANTS;
+      OrganisationUnitSelectionMode.ACCESSIBLE;
```
enricocolasante, line 118: Could you update the TODO here?

enricocolasante (approved):

PR #15039, opened 2023-09-04
fix: Export analytics crash due to invalid date and period [DHIS2-15573]

maikelarabori:
This fix is related to the analytics offset dates problem that happens during the Analytics Export process.

After internal discussion, we decided to add a new variable in "System Settings" representing a single number/offset to be used in the export process.

So, let’s suppose that the user/admin sets the offset value to “5”, and we are in the year 2023.

It means that analytics will accept dates from 2018 (inclusive) to 2028 (inclusive). Which translates to: [2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025, 2026, 2027, 2028].

If dates out of that year’s range are found the system sends back a warning message and also aborts the export process (through the job progress component).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
@@ -194,7 +194,7 @@ public class DefaultResourceTableService implements ResourceTableService {
   /**
    * This method checks if any of the year in the given list is within the offset defined in system
    * settings. The constant where the offset is defined can be seen at {@link
-   * SettingKey.ANALYTICS_PERIOD_YEARS_OFFSET}.
+   * SettingKey.ANALYTICS_MAX_PERIOD_YEARS_OFFSET}.
    *
    * <p>Based on the current year YYYY and the defined offset X. This method allows a range of X
    * years in the past and X years in the future. Including also the current year YYYY. So, for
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
```diff
@@ -181,9 +184,48 @@ public void generateDataElementTable() {
...
+                  .filter(year -> year >= minRangeAllowed && year <= maxRangeAllowed)
+                  .toList()
+              + ".";
+      errorMessage +=
+          "\n Years are out of range found: "
+              + yearsToCheck.stream()
+                  .filter(year -> year < minRangeAllowed || year > maxRangeAllowed)
+                  .toList()
+              + ".";
+      throw new RuntimeException(errorMessage);
```
jbee, line 227: Maybe throw a more specific exception. For example `ConflictException` in `feedback` package.

dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
```diff
@@ -264,6 +264,9 @@ public enum SettingKey {
   ANALYTICS_CACHE_TTL_MODE(
       "keyAnalyticsCacheTtlMode", AnalyticsCacheTtlMode.FIXED, AnalyticsCacheTtlMode.class),
 
+  /** The offset of years used during period generation during the analytics export process. */
+  ANALYTICS_PERIOD_YEARS_OFFSET("keyAnalyticsPeriodYearsOffset", 22, Integer.class),
```
jbee, line 268: Maybe add a "max" into the name to be more clear?

jbee (approved):
Looks good to me, some suggestions.

david-mackessy:
looks good - when updating the docs with the new system setting it would be good to have some info about the possible effects of changing the default value.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/resourcetable/DefaultResourceTableService.java
maikelarabori: I wanted to avoid a checked exception to not change the interface and all other implementations and have to handle it in other places. But let me double-check it.

dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
maikelarabori: Done.

PR #15040, opened 2023-09-04
fix: prevent a non-ALL user admin to update an ALL admin

netroms:


stian-sandvold (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -618,49 +619,44 @@ public boolean isAccountExpired(User user) {
...
     }
 
-    // Validate user role
+    // Validate if the current user can create/modify users with the ALL authority
+    User userToChange = userStore.get(user.getId());
+    if (!currentUser.isSuper() && userToChange != null && userToChange.getUserRoles() != null) {
+      Set<UserRole> roles = userToChange.getUserRoles();
+      for (UserRole role : roles) {
+        Set<String> authorities = role.getAuthorities();
+        if (authorities.contains(Authorities.ALL.name())) {
```
david-mackessy, line 636: There's a helper method in `User` to get all authorities directly without having to get the roles first - `getAllAuthorities`
Should it break out of the loop if it finds the `ALL` authority? There's probably no point checking any more if we know it has the `ALL` authority.

david-mackessy (approved):

PR #15067, opened 2023-09-06
fix: Add OptionSet/Options into metadata [DHIS2-15775]

maikelarabori:
Historically, `OptionSets` allow the creation of `Option` codes that are not unique across the system.
`Option` codes are unique only within their respective `OptionSet`.

This might become problematic when users define the same `Option` code in different `OptionSets`. Some parts of the application that rely purely on the `Option` code cannot know which `OptionSet` it belongs to. This is a problem in the `Metadata` object, currently used in analytics requests.

So, instead of returning only the `Option`, we will also return the `OptionSet`. This will allow consumers to find out the `OptionSet` a `Option` belongs to, if needed.

The new object returned will look like:
```
{
  "metaData": {
    "items": {
      "optionset1id": {
        "uid": "option set id 1",
        "name": "option set name 1",
        "options": [
          {
            "uid": "optionid1",
            "code": "PASSIVE"
          },
          {
            "uid": "optionid2",
            "code": "REACTIVE"
          }
        ]
      },
      "optionset2id": {
        "uid": "option set id 2",
        "name": "option set name 2",
        "options": [
          {
            "uid": "optionid3",
            "code": "PASSIVE"
          }
        ]
      },
...
}
```

Example of request:
```
http://localhost:8080/dhis/api/41/analytics/events/query/IpHINAT79UW?dimension=ou:USER_ORGUNIT,A03MvHHogjR.ebaJjqltK5N:IN:1;2&headers=ouname,A03MvHHogjR.ebaJjqltK5N&totalPages=false&displayProperty=NAME&outputType=EVENT&pageSize=10&page=1&includeMetadataDetails=true&stage=A03MvHHogjR&filter=pe:LAST_12_MONTHS&relativePeriodDate=2022-10-01
```


d-bernat (approved):

gnespolino (approved):
LGTM

PR #15071, opened 2023-09-07
fix:Remove orgUnitName from objects in tracker exporters [TECH-1633]

enricocolasante:
`OrgUnitName` was the only data linked to metadata that we were exposing in tracker exporters.
This PR is removing it from the API.


muilpp (approved):

teleivo (approved):
Can we do this without changing old tracker?

PR #15072, opened 2023-09-07
fix: prevent a non-ALL role admin to assign ALL authority to role

netroms:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -687,6 +685,27 @@ private void validateUserRoles(User user, User currentUser, List<ErrorReport> er
     }
   }
 
+  @Override
+  public List<ErrorReport> validateUserRoleCreateOrUpdate(UserRole role, User currentUser) {
```
vietnguyen, line 689: Do we need `  @Transactional(readOnly = true)` for this method ?

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -687,6 +685,27 @@ private void validateUserRoles(User user, User currentUser, List<ErrorReport> er
...
+
+    if (currentUser == null || role == null) {
+      return errors;
+    }
+
+    if (!currentUser.isSuper() && role.isSuper()) {
+      errors.add(new ErrorReport(UserRole.class, ErrorCode.E3032, currentUser.getUsername()));
+    }
+
+    UserRole userRoleBefore = userRoleStore.get(role.getId());
```
david-mackessy, line 701: i don't follow why we go to the `userRoleStore` to get the same `userRole` that we already seem to have. Is there a specific reason for this?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/SecurityCheck.java
```diff
@@ -138,6 +139,17 @@ private <T extends IdentifiableObject> void runValidationCheck(
         }
       }
 
+      if (object instanceof UserRole) {
```
david-mackessy, line 142: we could use the `instanceof` pattern match and remove the redundant cast e.g.
`object instanceof UserRole ur`

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/validation/SecurityCheck.java
netroms: This PR needs to be backported to Java 11 branches, so for convenience, it's better to refactor this later.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: This is done to compare with the original unmodified role, the incoming role could be mutated.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: Possibly, I added it to follow the same pattern as the other validation methods.

PR #15085, opened 2023-09-11
fix: Non data dimension org. unit, analytics 500 error [DHIS2-14923]

maikelarabori:
The respective endpoint is returning `500 internal server` error for calls to the org. unit analytics API (`api/orgUnitAnalytics`) when the query references an org. unit group set that is not set to be a data dimension.

It happens that when the org. unit group set is not defined as `data dimension`, it is not exported as analytics data, and hence cannot be queried. The column for this org. unit group set will not exist in the analytics table, and any query referencing this column will fail. In this case, with `500 error`.

This fix will return a friendly user message if users try to query an org. unit group set that is not defined as a data dimension.

Also, I applied some small code improvements.

**NOTE**: It's also recommended to exclude org. unit sets not that are not defined as data dimensions. This should be done by the client app, probably filtering out the ones that are not data dimensions. ie:
```
http://localhost:8080/dhis/api/organisationUnitGroupSets?paging=false&fields=id,displayName,organisationUnitGroups[id,displayName]&filter=dataDimension:eq:true
```


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/TableInfoReader.java
```diff
@@ -0,0 +1,100 @@
...
+        .filter(c -> !columnsInTable.contains(c))
+        .collect(toUnmodifiableSet());
+  }
+
+  /**
+   * Object representing the information related to the meta information of the respective table.
+   */
+  @Data
+  @AllArgsConstructor
+  public class TableInfo {
```
gnespolino, line 96: should be static

maikelarabori: Done

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/TableInfoReader.java
```diff
@@ -0,0 +1,100 @@
...
+   * @throws IllegalArgumentException if the argument 'tableName' is null or blank.
+   */
+  public TableInfo getInfo(@Nonnull String tableName) {
+    hasText(tableName, "Param 'tableName' cannot be null/blank");
+
+    String sql =
+        "select column_name"
+            + " from information_schema.columns"
+            + " where table_schema = 'public'"
+            + " and table_name = ?";
```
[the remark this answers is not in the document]
maikelarabori: Done

d-bernat (approved):

PR #15130, opened 2023-09-13
fix: Remove parameters from TrackedEntity exporter endpoint [DHIS2-15665]

enricocolasante:
Following up on https://github.com/dhis2/dhis2-core/pull/15109

This PR removes `query` parameter from TrackedEntity exporter endpoint.

Next PR will just update OpenAPI specs and review `HibernateTrackedEntityStore` to check if there is something to simplify, JavaDocs to update.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationParamUtils.java
```diff
@@ -138,35 +137,6 @@ else if (filters.length == 4) {
...
-  private static QueryFilter operatorValueQueryFilter(String[] operatorValue, String filter)
-      throws BadRequestException {
-    if (null == operatorValue || operatorValue.length < 2) {
-      throw new BadRequestException("Query item or filter is invalid: " + filter);
-    }
-
-    return operatorValueQueryFilter(operatorValue[0], operatorValue[1], filter);
-  }
-
   public static QueryFilter operatorValueQueryFilter(String operator, String value, String filter)
```
teleivo, line 140: When I worked on harmonizing order in /events I moved parsing of filters from the service layer (if I remember correctly) to the web layer https://github.com/dhis2/dhis2-core/blob/6b058254b83cfab4f92a3f26e9b5a11bbe95199f/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java#L251 I wanted to do the same for /trackedEntities but could not as the interaction between order, attributes and filters was so complex. I mention this so we don't forget to look at the changes for events and how to do the same for trackedEntities 😅

teleivo (approved):

PR #15140, opened 2023-09-14
fix: revert acl check on user group update (#15131)

netroms:
### Summary
Reverts change to do a full acl check during user group updates. This will not work when an user admin is doing a user update via a PUT operation and the user don't have access to modify user groups. There also exist a special authority allowing a user admin to add and remove group members without having full group modify access.
This is also makes sure the groups only are updated when they actually change, ie. in this instance, the membership count changes.
This PR also do some minor refactoring, mostly just removing some dead code/unused public methods in the user group service and remove a duplicate constant.

## New automatic test
UserControllerTest#updateUserHasAccessToUpdateGroups()

Jira: [DHIS2-15787](https://dhis2.atlassian.net/browse/DHIS2-15787)

[DHIS2-15787]: https://dhis2.atlassian.net/browse/DHIS2-15787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Philip-Larsen-Donnelly (approved):
Approved as backport

mortenoh (approved):

PR #15143, opened 2023-09-14
fix: /analytics API doesn't return the hierarchy for OUs that don't have data[2.41-DHIS2-15791]

d-bernat:
please see https://dhis2.atlassian.net/browse/DHIS2-15791

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/handler/MetadataHandler.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/handler/MetadataHandler.java
@@ -135,7 +135,7 @@ public class MetadataHandler {
 
       if (params.isShowHierarchy()) {
         Map<Object, List<?>> ancestorMap =
-            activeOrgUnits.stream()
+            organisationUnits.stream()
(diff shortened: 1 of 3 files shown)


gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/handler/MetadataHandler.java
```diff
@@ -135,7 +135,7 @@ public void addMetaData(DataQueryParams params, Grid grid) {
 
       if (params.isShowHierarchy()) {
         Map<Object, List<?>> ancestorMap =
-            activeOrgUnits.stream()
```
maikelarabori, line 138: I'm wondering if would be safer to do it only for download/export endpoints. I'm afraid we are changing the existing behavior for the `isShowHierarchy` flag.

maikelarabori: Basically, we could check if the request originates from download endpoints.

d-bernat: done

maikelarabori (approved):

PR #15149, opened 2023-09-14
fix: prevent a user change access to a higher organisation unit

netroms:
### Summary
Makes sure an admin user can not change a users's organisation unit to a higher level than what the admin user is assigned to.

### Automatic tests
UserControllerTest#testChangeOrgUnitLevelGivesAccessError()


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/UserControllerTest.java
```diff
@@ -232,6 +236,103 @@ private JsonImportSummary updateRolesNonAllAdmin() {
...
+                    + "{'op':'add','path':'/organisationUnits','value':[{'id':'"
+                    + orgA.getUid()
+                    + "'}]},"
+                    + "{'op':'add','path':'/dataViewOrganisationUnits','value':[{'id':'"
+                    + orgA.getUid()
+                    + "'}]},"
+                    + "{'op':'add','path':'/teiSearchOrganisationUnits','value':[{'id':'"
+                    + orgA.getUid()
+                    + "'}]}"
+                    + "]")
```
jbee, line 319: You can use `"""` blocks with `.formatted` to insert the UIDs to avoid putting together long string constants

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -618,25 +625,44 @@ public boolean isAccountExpired(User user) {
...
     validateUserGroups(user, currentUser, errors);
 
+    validateOrgUnit(user.getOrganisationUnits(), currentUser, errors);
+    validateOrgUnit(user.getDataViewOrganisationUnits(), currentUser, errors);
+    validateOrgUnit(user.getTeiSearchOrganisationUnits(), currentUser, errors);
+
     return errors;
   }
 
+  private void validateOrgUnit(
```
jbee, line 651: I think either the method needs a more intention carrying name or a comment to explain what we try to prevent with this check. From the name alone as it is this is very hard to reverse engineer.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/UserControllerTest.java
larshelge: Just curious, what does `.formatted` refer to / how do you use that @jbee ?

jbee: [`String.formatted` ](https://stackoverflow.com/questions/70995023/whats-the-difference-between-string-format-and-str-formatted-in-java)

jbee (approved):

PR #15170, opened 2023-09-15
fix: Harmonize filter field in TrackedEntity exporter [DHIS2-15665]

enricocolasante:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -570,26 +573,6 @@ void shouldReturnEmptyCollectionIfGivenFilterDoesNotMatch()
...
-            .orgUnitMode(DESCENDANTS)
-            .trackedEntityTypeUid(trackedEntityTypeA.getUid())
-            .filters(teaA.getUid())
-            .user(user)
-            .build();
-
-    List<TrackedEntity> trackedEntities =
-        trackedEntityService.getTrackedEntities(operationParams).getTrackedEntities();
-
-    // A filter with only a UID does not cause any join on attributes, so it does not actually
```
teleivo, line 588: Is that test removed because a filter with only a UID does actually cause an inner join? So this was a bug?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -570,26 +573,6 @@ void shouldReturnEmptyCollectionIfGivenFilterDoesNotMatch()
...
-    // filter TEs from the response in contrast to /tracker/events?filterAttribute=uid
-    assertContainsOnly(List.of(trackedEntityA, trackedEntityB), trackedEntities);
-  }
-
   @Test
   void shouldReturnTrackedEntitiesEvenIfTrackedEntityDoesNotHaveGivenAttribute()
```
teleivo, line 577: These operation params do not use any attribute UID, is the test name accurate?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParamsMapperTest.java
```diff
@@ -287,4 +298,119 @@ void shouldFailGivenInvalidOrderFieldName() {
         () -> assertStartsWith("order parameter is invalid", exception.getMessage()),
         () -> assertContains("unsupportedProperty1", exception.getMessage()));
   }
+
+  @Test
+  void shouldCreateCriteriaFiltersWithFirstOperatorWhenMultipleValidOperandAreNotValid()
```
teleivo, line 303: I am not sure what the word `Criteria` refers to 🤔 

I would move most of these tests into `RequestParamsValidatorTest` and only keep one or two in here that demonstrate we parse the `filter` parameter into the expected data structure.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -638,10 +630,7 @@ public TrackedEntityQueryParams orderBy(String field, SortDirection direction) {
    */
   public TrackedEntityQueryParams orderBy(TrackedEntityAttribute tea, SortDirection direction) {
     this.order.add(new Order(tea, direction));
-    this.addFiltersIfNotExist(
-        QueryItem.getQueryItems(List.of(tea)).stream()
-            .filter(sAtt -> !this.getFilters().contains(sAtt))
-            .toList());
+    this.filters.putIfAbsent(tea, new ArrayList<>());
```
teleivo, line 633: > To make it work properly and differentiate the first 2 cases, we need to represent the presence of the filter with no values as a null. So there will be an entry in the filters map with the attribute UID and a list of filters with one null element.

These many cases we have to deal with make it complex 😅 We want an API that is safe to use. I wonder if we now depend on order of calling `filterBy`, `orderBy`.

Do we get the same behavior when calling
* `orderBy(uid1, DESC)` and then `filterBy(uid1)` or
* `filterBy(uid1)` and then `orderBy(uid1, DESC)`

if not I think we should make it so it does not matter so its safer to use.

teleivo (commented):
I was hoping we can remove code from `OperationParamUtils` related to parsing like `parseAttributeQueryItems` and only have it in `RequestParamsValidator` 😄 

A list of filters with one null element is odd to me. One alternative of mapping to null is maybe less odd but more dangerous. I am wondering what your pros/cons were when making the decision 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -348,7 +347,7 @@ public List<Long> getTrackedEntityIds(
     if (!params.hasProgram()) {
       Collection<TrackedEntityAttribute> attributes =
           trackedEntityAttributeService.getTrackedEntityAttributesDisplayInListNoProgram();
-      params.addFiltersIfNotExist(QueryItem.getQueryItems(attributes));
+      attributes.forEach(params::filterBy);
```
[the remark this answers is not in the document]
enricocolasante: To be honest, this whole `if` block doesn't make sense to me.
I will try to find out if it can be removed.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
enricocolasante: It was just a duplicated test of `shouldReturnTrackedEntitiesIfTheyHaveGivenAttributeFilteredUsingOnlyUID` test

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
enricocolasante: I think this can be removed as well.
It was probably useful before removing `attributes` parameter

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParamsMapperTest.java
enricocolasante: Will do that.
These are just moved tests from `TrackedEntityOperationParamsMapperTest`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
enricocolasante: Absolutely, I think this can be problematic, I will check.
I remember that this was one of my concerns but then we went to all-devs-week and I completely forgot to check :grimacing:

enricocolasante:
> I was hoping we can remove code from `OperationParamUtils` related to parsing like `parseAttributeQueryItems` and only have it in `RequestParamsValidator` 😄
> 
> A list of filters with one null element is odd to me. One alternative of mapping to null is maybe less odd but more dangerous. I am wondering what your pros/cons were when making the decision 😅

`OperationParamUtils` is gone :grimacing: 
For the list with null element, it is not ideal, it was the easiest way that we thought of without changing `QueryFilter` code.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
```diff
@@ -308,6 +285,88 @@ void shouldFailParsingFiltersWithMissingValueAndTrailingColon() {
     assertEquals("Query item or filter is invalid: " + TEA_1_UID + ":lt:", exception.getMessage());
   }
 
+  @Test
+  void shouldParseFiltersWithFirstOperatorWhenMultipleValidOperandAreNotValid()
```
teleivo, line 222: The test name ends in `NotValid` and am wondering why we don't throw a BadRequestException 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -340,27 +340,16 @@ public TrackedEntities getTrackedEntities(TrackedEntityOperationParams operation
...
     if (!params.hasProgram()) {
       Collection<TrackedEntityAttribute> attributes =
           trackedEntityAttributeService.getTrackedEntityAttributesDisplayInListNoProgram();
       attributes.forEach(params::filterBy);
     }
 
     decideAccess(params);
-
-    // AccessValidation should be skipped only and only if it is internal
-    // service that runs the task (for example sync job)
```
teleivo, line 357: Why don't we need the `if`s anymore?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -614,6 +614,9 @@ public TrackedEntityQueryParams filterBy(TrackedEntityAttribute tea, QueryFilter
 
   public TrackedEntityQueryParams filterBy(TrackedEntityAttribute tea) {
     this.filters.putIfAbsent(tea, Collections.singletonList(null));
+    if (this.filters.get(tea).isEmpty()) {
```
teleivo, line 617: I think this warrants a comment as we won't remember 😅 

Also I think we need to document the special role of the singleton `null` somewhere. You describe it in the PR but maybe it should be part of the javadoc for the `filters` field.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
```diff
@@ -1103,6 +1103,28 @@ void shouldOrderEventsByDataElementAndNotFilterOutEventsWithoutThatDataElement()
     assertEquals(List.of("pTzf9KYMk72", "D9PbzJY8bJM"), events);
   }
 
+  @Test
+  void shouldOrderEventsByDataElementAndFilterOutEventsWithoutThatDataElement()
```
teleivo, line 1107: This test shows up as new but looks a lot like https://github.com/dhis2/dhis2-core/blob/5e446f580c5c99087ce0b753e58e5aa0b3ed44fa/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java#L1003C8-L1023 :thinking:

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
enricocolasante: This method was only called in one place and the checked flags were statically passed as false

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
enricocolasante: Yes, removing it.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -362,12 +362,16 @@ private String getFromSubQuery(TrackedEntityQueryParams params, boolean isCountQ
             .append(" FROM trackedentity " + MAIN_QUERY_ALIAS + " ")
 
             // INNER JOIN on constraints
+            // filter={uid1} => Map.of(tea, List.of(null))
+            // filter={uid1}:eq:2 => Map.of(tea, List.of(new QueryFilter(EQ,"in")))
```
teleivo, line 366: I assume its

```suggestion
            // filter={uid1}:eq:2 => Map.of(tea, List.of(new QueryFilter(EQ,"2")))
```

right? Same for the comment below if that one is correct

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -606,12 +606,24 @@ public User getUser() {
...
+   * that consist of an operator and a value.
+   */
   public TrackedEntityQueryParams filterBy(TrackedEntityAttribute tea, QueryFilter filter) {
     this.filters.putIfAbsent(tea, new ArrayList<>());
     this.filters.get(tea).add(filter);
     return this;
   }
 
+  /**
+   * Filter out Tracked Entity that have no value for the given tracked entity attribute {@code
```
teleivo, line 620:
```suggestion
   * Filter out any tracked entity that have no value for the given tracked entity attribute {@code
```

teleivo (approved):

teleivo:
> `OperationParamUtils` is gone 😬 
That's alright! The code is now in a more appropriate place 😄 

> For the list with null element, it is not ideal, it was the easiest way that we thought of without changing `QueryFilter` code.

As long as this never leaks out of the QueryParams its alright! I don't want to know about the meaning of a null element when creating or consuming the QueryParams. I would thus put the javadoc on the field itself. I would also want us to provide getters so that the store does not need to know about it. Is that currently true? I just saw you added comments with the `null` element in the store. That should stay an implementation detail of the QueryParams.

PR #15223, opened 2023-09-25
chore: move handling of total count and pagination to store TECH-1638

teleivo:
Having the logic of fetching `pageSize + 1` events in the store and the removal of the extra event in the service is error prone. This is now done transparently in the store.

Total count is expensive and should thus be discouraged. Do not expose this as part of the service as its not needed right now.

As our store are concerned with generating SQL we should try to adopt the appropriate language for that domain. Our methods are often named `getYYYQuery` while they do not actually generate a SQL query. For example `getOrderQuery(params)` actually generates https://www.postgresql.org/docs/current/queries-limit.html limit and offset clauses.

Remove any pagination related fields and methods from `EventQueryParams` as these are fully represented by the `PageParams`.

Opt for not mutating the `List<Event>` when removing the `pageSize + 1` event in ` getPage`.


enricocolasante (approved):
Just a style suggestion in the store

PR #15233, opened 2023-09-27
chore: remove pageSize+1 logic as we never expose SlimPager TECH-1638

teleivo:
this logic was added by mistake to events and enrollments and then propagated to the other entities when we refactored pagination. The `isLastPage` from `SlimPager` was never returned from new tracker endpoints.

`curl -s -u admin:district https://play.dhis2.org/dev/api/tracker/events\?\&page\=2\&pageSize\=4\&totalPages\=true\&order\=programStage:asc\&fields\=event | jq`

```json
{
  "page": 2,
  "total": 373603,
  "pageCount": 93401,
  "pageSize": 4,
  "instances": [
    {
      "event": "RGWzjuz7eTt"
    },
    {
      "event": "Yzx5VJzvQmJ"
    },
    {
      "event": "dOeZnvyjdsW"
    },
    {
      "event": "uh0sdUZhnWZ"
    }
  ]
}
```

`curl -s -u admin:district https://play.dhis2.org/dev/api/tracker/events\?\&page\=2\&pageSize\=4\&totalPages\=false\&order\=programStage:asc\&fields\=event | jq`

```json
{
  "page": 2,
  "pageSize": 4,
  "instances": [
    {
      "event": "Yzx5VJzvQmJ"
    },
    {
      "event": "dOeZnvyjdsW"
    },
    {
      "event": "uh0sdUZhnWZ"
    },
    {
      "event": "nfYuLdN48gD"
    }
  ]
}
```

The commons [Pager](https://github.com/dhis2/dhis2-core/blob/1afc9c8e2e00517f620e23f90d404a88071f21d6/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Pager.java#L43) was likely built with metadata endpoints in mind. In metadata endpoints totals are always returned. This does not match the tracker endpoints. So right now since we are using the common `Pager` we pass total `0` when we do not fetch totals. When returning the pagination fields in JSON we rely on [PagingWrapper](https://github.com/dhis2/dhis2-core/blob/1afc9c8e2e00517f620e23f90d404a88071f21d6/dhis-2/dhis-api/src/main/java/org/hisp/dhis/webapi/controller/event/webrequest/PagingWrapper.java#L79) which uses `Integer`. Important is that we set it to null in case the user did not want the total counts. Otherwise we would return `0`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -28,7 +28,6 @@
 package org.hisp.dhis.tracker.export.event;
 
 import static java.util.Map.entry;
-import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
 import static org.hisp.dhis.common.ValueType.NUMERIC_TYPES;
 import static org.hisp.dhis.system.util.SqlUtils.castToNumber;
 import static org.hisp.dhis.system.util.SqlUtils.lower;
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):
:+1: for the PR
We should just remember this logic when we discuss pagination with other teams and how it should be all over the system as this `lastPage` flag can be the middle ground between always returning the total count to the client and not passing any information about pagination

teleivo:
> We should just remember this logic when we discuss pagination with other teams and how it should be all over the system as this `lastPage` flag can be the middle ground between always returning the total count to the client and not passing any information about pagination

definitely! `isLastPage` could also easily be represented using the `nextPage` in the common Pager which is returned at least by metadata and gist. If there is no `nextPage` you are on the last one 😄

PR #15236, opened 2023-09-27
feat: Add system setting for parallel jobs in analytics export [DHIS2-15880]

larshelge:
This PR adds a system setting for explicitly setting the number of parallel threads/jobs to use for the analytics table export. The number of jobs is currently inferred from the available processors on the application server. This is useful when you want to experiment with the number of parallel jobs to measure performance.


gnespolino (approved):
LGTM

maikelarabori (approved):

PR #15239, opened 2023-09-27
fix: Harmonize filter field in TrackedEntity exporter [DHIS2-15665]

enricocolasante:
The goal of this PR is to simplify the management of filters and orders in `TrackedEntityQueryParams` and make it transparent for the consumer how to use the API.

Instead of having the store deal with the complexity of filters, we are exposing 3 different ways to get orders and filters information:
- `getAttributes()`: Returns attributes that are either ordered by or present in any filter.
- `getLeftJoinAttributes()`: Returns attributes that are only ordered by and not present in any filter.
- `getFilters()`: Returns all the filters specified by the client.

The first 2 methods are hiding the complexity around the interactions between filters and orders. In particular the fact that an attribute that is present in an order will be translated to a `LEFT JOIN` in the store, only and only if it is not present in any filter.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
@@ -37,7 +37,6 @@ import static org.hisp.dhis.common.Pager.DEFAULT_PAGE_SIZE;
 import static org.hisp.dhis.common.SlimPager.FIRST_PAGE;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -341,12 +340,6 @@ class DefaultTrackedEntityService implements TrackedEntityService {
   }
 
   public List<Long> getTrackedEntityIds(TrackedEntityQueryParams params) {
-    if (!params.hasProgram()) {
-      Collection<TrackedEntityAttribute> attributes =
-          trackedEntityAttributeService.getTrackedEntityAttributesDisplayInListNoProgram();
-      attributes.forEach(params::filterBy);
-    }
-
     decideAccess(params);
     validate(params);
     validateSearchScope(params);
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
@@ -558,7 +558,7 @@ class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<Tracked
   private String getFromSubQueryJoinOrderByAttributes(TrackedEntityQueryParams params) {
     StringBuilder joinOrderAttributes = new StringBuilder();
 
-    for (TrackedEntityAttribute orderAttribute : params.getAttributeInOrderButNotInFilter()) {
+    for (TrackedEntityAttribute orderAttribute : params.getLeftJoinAttributes()) {
 
       joinOrderAttributes
           .append(" LEFT JOIN trackedentityattributevalue AS ")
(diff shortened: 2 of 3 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -341,12 +340,6 @@ public TrackedEntities getTrackedEntities(TrackedEntityOperationParams operation
   }
 
   public List<Long> getTrackedEntityIds(TrackedEntityQueryParams params) {
-    if (!params.hasProgram()) {
```
enricocolasante, line 344: @stian-sandvold @Bekkalizer @JoakimSM that check is implementing a logic that is not documented in any place and it is potentially harmful in the sense that some TE will not be retrieved based on a filter that is not specified by the client.

If you cannot see a strong reason to maintain this here, I am happy to remove it

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -987,12 +956,7 @@ private String getQueryOrderBy(TrackedEntityQueryParams params, boolean innerOrd
       return "ORDER BY " + StringUtils.join(orderFields, ',') + SPACE;
     }
 
-    if (params.getFilters().entrySet().stream()
-        .noneMatch(qi -> !qi.getValue().isEmpty() && qi.getKey().isUnique())) {
-      return "ORDER BY " + DEFAULT_ORDER + " ";
-    }
-
-    return "";
+    return "ORDER BY " + DEFAULT_ORDER + " ";
```
enricocolasante, line 959: @stian-sandvold Do you know why at some point add the possibility to create a query without default order if there is a filter on an attribute that is unique?

It seems to me that it is safe and it is not impacting performance just to maintain that order by clause

teleivo (approved):
Nice simplification of the query params 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
stian-sandvold: I am not sure why this is here right now; I am sure it made sense at some point, but as far as I can tell now; This just means the default sorting order is not always the same, which is not great.

I think always using default, if nothing was explicitly ordered in the params, is the way to go.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
stian-sandvold, line 1: We're not introducing a TEA with null values. What kind of impact does that have on the query overall?

enricocolasante: I am not sure I understand what you mean but here is what we are doing in different cases:

- filter={uid1} => inner join on uid1 without filtering on value
- filter={uid1}:eq:2 => inner join on uid1 and filter value equals 2
- order={uid1}:desc&filter={uid1} => inner join on uid1 without filtering on value\
- order={uid1}:desc&filter={uid1}:eq:2 => inner join on uid1 and filter value equals 2
- order={uid1}:desc =>left join on uid1 without filtering on values (we need information about the attribute in order to sort but we cannot filter out results if the attribute has no value)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
enricocolasante: I am not really sure that this code is doing this kind of validation.
What this code is doing is adding a filter based on a flag present in the attribute.
This means that if there are 2 attributes, A and B where A has `DisplayInListNoProgram` flag as true and B has `DisplayInListNoProgram` as false, then if a client is not providing a program, only the TEs that have a value for attribute A will be returned.

I am not sure how `DisplayInListNoProgram` is set and exactly what it means. But anyway this check here is not preventing searching on any attribute or filtering the attributes that are returned.

PR #15252, opened 2023-09-28
fix: query generation consider multiple periods for single value queries [DHIS2-14971]

gnespolino:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/TimeFieldSqlRenderer.java
```diff
@@ -214,4 +229,40 @@ protected abstract String getColumnName(
    * @return the collection of {@link TimeField}
    */
   protected abstract Set<TimeField> getAllowedTimeFields();
+
+  protected String getSqlForAllPeriods(String alias, List<DimensionalItemObject> periods) {
```
maikelarabori, line 233: Would be nice to have a Javadoc here. Maybe showing some examples of possible SQL snippets output.

maikelarabori (approved):
Looks good. Left one minor comment. Also, would be nice to have some e2e tests as discussed. Thanks!

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/TimeFieldSqlRenderer.java
gnespolino: done!

d-bernat (approved):

PR #15284, opened 2023-10-02
fix: Note clean up [DHIS2-15806]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-15806

Notes can't be deleted by a user. Apparently, the only way to delete notes, and their association with events and enrollments, can be done by using the `MaintenanceStore`.
Therefore we can delete the `NoteDeletionHandler`. In addition, we do some cleanup of the `NoteService` and remove unused code.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/note/NoteService.java
```diff
@@ -36,38 +36,16 @@ public interface NoteService {
   /**
    * Adds an {@link Note}
    *
-   * @param comment The to TrackedEntityComment add.
+   * @param note The Note to add.
    * @return A generated unique id of the added {@link Note}.
```
enricocolasante, line 40: The method return type is void now

enricocolasante (approved):

zubaira (approved):

PR #15286, opened 2023-10-02
chore: update docker-compose.yml

michael-markevich:
Based on the security assessment results (https://dhis2.atlassian.net/browse/SEC-48), I suggest updating the docker-compose.yml to move the composer configuration to the production setup. This includes using the official Postgis repository, an additional health check and security options, and a clean setup without a demo database and debugging.

The current docker-compose.yml can still be used for development purposes.


tonsV2:
I would recommend creating a copy of the original docker compose file and naming it docker-compose.production.yml and make that the target of your updates.

Having both docker compose files in this PR make sense as we want to apply some of these changes to the original as well.

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
 
 services:
   web:
-    image: "${DHIS2_IMAGE:-dhis2/core-dev:local}"
+    image: "${DHIS2_IMAGE:-dhis2/core}"
```
tonsV2, line 5: A tag  should be used here so we're not just pulling latest.

Perhaps use an environment variable, so we don't have to update the docker compose file, if we want to run a different version.

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
...
+    restart: unless-stopped
     ports:
-      - 127.0.0.1:8080:8080 # DHIS2
-      - 127.0.0.1:8081:8081 # Debugger: connect using commandline flag -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:8081
-      - 127.0.0.1:9010:9010 # JMX port (for example for VisualVM)
+      - 127.0.0.1:${DHIS2_PORT:-8080}:8080
```
tonsV2, line 8: We don't want to use 127.0.0.1 in production :smile:

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
...
-              -Dcom.sun.management.jmxremote \
-              -Dcom.sun.management.jmxremote.port=9010 \
-              -Dcom.sun.management.jmxremote.local.only=false \
-              -Dcom.sun.management.jmxremote.authenticate=false \
-              -Dcom.sun.management.jmxremote.ssl=false"
+      JAVA_OPTS: "-Dlog4j2.configurationFile=/opt/dhis2/log4j2.xml"
     depends_on:
       db:
         condition: service_healthy
+    healthcheck:
```
tonsV2, line 18: Nice! We should use this in both the docker compose files :+1:

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
...
+      start_period: 120s
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    security_opt:
+      - no-new-privileges:true
 
   db:
-    image: ghcr.io/baosystems/postgis:12-3.3
+    image: postgis/postgis
```
tonsV2, line 28: Why use this image?

I would recommend [this](https://hub.docker.com/r/bitnami/postgresql) one which is also what we're using on Kubernetes.

Regardless of which image we're using. A tag should be present so we're not just using latest.

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
...
+    restart: unless-stopped
     ports:
       - 127.0.0.1:5432:5432
```
tonsV2, line 31: Strip 127.0.0.1? Unless I'm missing something? :smile:

docker-compose.yml
```diff
@@ -2,32 +2,35 @@ version: "3.8"
...
     volumes:
-      - db-dump:/docker-entrypoint-initdb.d/
+      - postgresql:/var/lib/postgresql
     environment:
       POSTGRES_USER: dhis
```
tonsV2, line 35: I would recommend a stronger password here and the usage of a .env file.

The pattern we're using on the IM projects is to have a committed .env.example file in the root of the project which serves as an example. The readme and the file itself can contain some instructions if needed.

docker-compose.yml
```diff
@@ -36,21 +39,14 @@ services:
...
-
-  db-dump:
-    image: busybox # busybox wget version does not have --no-clobber, so we need to do the [ -f ] test
-    command: sh -c '[ -f dump.sql.gz ] && echo "dump.sql.gz exists" || wget --output-document dump.sql.gz $$DHIS2_DB_DUMP_URL' # only download file if it does not exist
-    environment:
-      DHIS2_DB_DUMP_URL: "${DHIS2_DB_DUMP_URL:-https://databases.dhis2.org/sierra-leone/dev/dhis2-db-sierra-leone.sql.gz}"
-    working_dir: /opt/dump
-    volumes:
-      - db-dump:/opt/dump
+    security_opt:
```
tonsV2, line 45: This should be in both of our docker compose files.

tonsV2 (changes requested):
It might be beyond the scope of this PR but at some point we should probably supply some guide lines for log aggregation and backup.

https://docs.docker.com/storage/volumes/#back-up-restore-or-migrate-data-volumes
https://docs.docker.com/compose/compose-file/compose-file-v3/#logging

docker-compose.yml
radnov: Just a quick note here - the `postgis/postgis` image won't suffice, as it's only built for `linux/amd64` and we want this Docker compose file to also be usable by developers with the new Mac M* chips. 

The `ghcr.io/baosystems/postgis` one is built for `linux/arm64` as well and it provides some convenience, as we don't need to install `postgis` separately (like we do [here for the dhis2-db stack in the IM](https://github.com/dhis2-sre/im-manager/blob/master/stacks/dhis2-db/seed.sh#L19)), but it might actually be better to streamline this for consistency.  👍

radnov: @michael-markevich Right, I should have been a bit more clear - the `postgis/postgis` image won't suffice for the developer docker compose file.

Still, it might be worth having the image we recommend for production use be the same one we use in the Instance Manager, as it has been tested the most. Technically, `postgis/postgis` is also a custom image, the "official" one is https://hub.docker.com/_/postgres.

PR #15303, opened 2023-10-03
feat: Migrate CompleteDataSetRegistration async task to use new Job setup [DHIS2-15745]

david-mackessy:
# Summary
Migrate the custom `CompleteDataSetRegistration` async import execution to use the new Job setup, bringing it in line with how all other jobs are run.

## Changes
- added new `JobConfiguration` type `COMPLETE_DATA_SET_REGISTRATION_IMPORT`
- added new `JobType` check to include use of Notifier (if setup)
- added `@Transactional` annotation to all service methods that require write permission
- removed the `Notifier` dependency from service (along with its logs) as the new Job framework will automatically call the `Notifier` if it's setup/required
- refactored existing async task to implement Job
- use `MimeType` type instead of `String` for format type (JSON/XML)

# Testing
## Automated
- added end-to-end test which tests the async job execution for completing a dataset registration

## Manual
- to view completed data sets use `GET /api/completeDataSetRegistrations?dataSet={dataSetUid}&orgUnit={orgUnitUid}&period={period}`
- a valid dataset with an org unit must exist before completing
- to complete a data set registration using the new behaviour use:
  - `POST /api/completeDataSetRegistrations?async=true` with body
  
``` json
{
    "completeDataSetRegistrations": [
        {
            "dataSet": "VTdjfLXXmoi",
            "period": "202008",
            "organisationUnit": "z77LiPqLMoq"
        }
    ]
}
```

  - for XML use body

``` xml
<completeDataSetRegistrations xmlns="http://dhis2.org/schema/dxf/2.0">
    <completeDataSetRegistration period="202007" dataSet="VTdjfLXXmoi"
    organisationUnit="y77LiPqLMoq"/>
</completeDataSetRegistrations>
```


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CompleteDataSetRegistrationController.java
```diff
@@ -98,26 +93,30 @@
...
+  private final CurrentUserService currentUserService;
 
-  @Autowired private InputUtils inputUtils;
+  private final InputUtils inputUtils;
 
-  @Autowired private DefaultCompleteDataSetRegistrationExchangeService registrationExchangeService;
+  private final CompleteDataSetRegistrationExchangeService registrationExchangeService;
 
-  @Autowired private AsyncTaskExecutor taskExecutor;
+  private final AsyncTaskExecutor taskExecutor;
```
jbee, line 114: This field should now be unused and should be removed.

jbee (approved):
Looks good. 
I see you opted to not record any detail stages within the service but run the service method call as a stage. This is fine as it it the level of detail we had before. We can always add more later if we want to.

Just a minor cleanup comment.

david-mackessy:
> Looks good. I see you opted to not record any detail stages within the service but run the service method call as a stage. This is fine as it it the level of detail we had before. We can always add more later if we want to.
> 
> Just a minor cleanup comment.

@jbee yeah I kept the logging detail the same, so whether run sync/async it will look similar

PR #15307, opened 2023-10-03
fix: TrackedEntity Get Ids : Missing group by enrolledat field [DHIS2-15658 patch/2.40]

lucaCambi77:
see https://github.com/dhis2/dhis2-core/pull/15302


enricocolasante (changes requested):
patch/2.40.0 has already been released.
Unless this is a critical bug and someone from the release team (eg. Phil) has specifically asked to create a patch for some minor version, we should just fix it in the version branch 2.40

lucaCambi77:
close for now as we don't need that

PR #15331, opened 2023-10-05
fix: validate search scope when no program specified [TECH-1630]

muilpp:
When using the org unit mode `ACCESSIBLE` on a non restricted program, we were not filtering by the org units in the users' search scope. Fixed by adding a clause to make sure every result is in the user scope.

ticket: https://dhis2.atlassian.net/browse/TECH-1630


enricocolasante (approved):

gnespolino (commented):
The code looks good, but one more tracker review would be desirable for checking if the functional requirement is okay.

PR #15336, opened 2023-10-06
fix: Duplicated program stage field in /tracker/events select [DHIS2-15899]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-15899

In _getEventSelectQuery()_ method the field _ps.uid as ps_uid_ is duplicated. Add e2e test

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
@@ -474,7 +474,7 @@ public class TrackerExportTests extends TrackerNtiApiTest {
             .extractList("instances.programStage.flatten()");
 
     assertEquals(
-        List.of("nlXNK4b7LVr", "PaOOjwLVW23"),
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/importer/TrackerExportTests.java
```diff
@@ -456,6 +457,26 @@ public void shouldReturnAscOrderedEventByTEIAttribute() {
...
+        .validateSuccessfulImport();
+
+    List<String> actualPsList =
+        trackerActions
+            .get("events?order=programStage&fields=programStage&program=f1AyMswryyQ")
+            .validateStatus(200)
+            .extractList("instances.programStage.flatten()");
+
+    List<String> orderedPsList = new ArrayList<>(actualPsList);
+    orderedPsList.sort(Collections.reverseOrder());
```
enricocolasante, line 475: I cannot understand this assertion.
It seems it is asserting that a collection and its reverse should be the same.
I think the expected PsList here could be hardcoded to make it clear what is happening

muilpp (approved):

PR #15348, opened 2023-10-09
fix: use a dedicated DataSource for jdbcTemplates [DHIS2-15597]

gnespolino:
When we mix Hibernate and Spring JdbcTemplate code in the same flow, we end up having the same connection/transaction used for both.
While I can't see any advantage of doing so (since we can always propagate an exception in jdbcTemplate to have the hibernate session rollback), we have indeed issues when we want to gracefully handle jdbcTemplate exception without a rollback.
This is impossible in the current scenario due to the fact that springs detects any exception thrown during the flow, even if they are gracefully caught, and marks the transaction as rollback.
To avoid this side effect, we need to separate the data source used by spring and the one used by hibernate (just analytics).
In this PR we create an analyticsJdbcTemplate dedicated datasource, and we also add a flag to detect if multiple analytics queries are running.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
+++ b/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
@@ -33,6 +33,7 @@ import java.sql.SQLException;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import javax.sql.DataSource;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import net.ttddyy.dsproxy.listener.MethodExecutionContext;
 import net.ttddyy.dsproxy.listener.logging.DefaultQueryLogEntryCreator;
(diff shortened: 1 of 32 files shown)


d-bernat (approved):

maikelarabori (approved):

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
```diff
@@ -102,8 +95,8 @@ public JdbcTemplate readOnlyJdbcTemplate(@Qualifier("dataSource") DataSource dat
     return jdbcTemplate;
   }
 
-  @Bean("actualDataSource")
-  public DataSource actualDataSource(
+  public static DataSource getActualDataSource(
+      DhisConfigurationProvider dhisConfig,
```
vietnguyen, line 99: I'm not sure if we want to have separate connection pool for each dataSource for same database.
There would be confusion to the connection pool settings in `dhis.conf` file. 
For example `CONNECTION_POOL_MAX_SIZE` or `CONNECTION_POOL_MAX_IDLE_TIME` , etc.

gnespolino: yes those values should be then fine tuned

maikelarabori (approved):

PR #15356, opened 2023-10-10
fix: Validate search scope in tracked entities requests [TECH-1631]

muilpp:
Some changes were made in `/trackedEntities` based on wrong assumptions, this PR will revert these assumptions, which mainly are:

1. Have the service layer only validate the requested org unit is in the search scope, even if the program is restricted.
2. Also in the service layer, remove the logic where we fetch accessible org units. This is not needed anymore, since the filtering of `trackedEntities` will be done in the persistence layer (`DefaultTrackedEntityStore.getOwnedTeisPartitioned`)
3. In the persistence layer, remove the logic where we build the query depending on the list of accessible org units. Here, we need to revert this logic to how it was before, and leave it like that, since it will already filter by search or capture scope depending on the program access level. This is in `HibernateTrackedEntityStore.getFromSubQueryJoinOrgUnitConditions`

Ticket: https://dhis2.atlassian.net/browse/TECH-1631

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
@@ -77,7 +77,7 @@ public class TrackedEntityQueryParams {
    * Organisation units for which instances in the response were registered at. Is related to the
    * specified OrganisationUnitMode.
    */
-  private Set<OrganisationUnit> accessibleOrgUnits = new HashSet<>();
+  private Set<OrganisationUnit> orgUnits = new HashSet<>();
 
   /** Program for which instances in the response must be enrolled in. */
   private Program program;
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -186,19 +186,19 @@ public TrackedEntityQueryParams() {}
    */
   public void handleOrganisationUnits() {
```
enricocolasante, line 187: Maybe not part of this PR but this method could be moved to `TrackedEntityOperationParamsMapper` so we don't need to change orgUnits in `TrackedEntityQueryParams`?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
muilpp: I agree. It should be the same with these three [validation methods](https://github.com/dhis2/dhis2-core/pull/15356/files#diff-069b27403f4f267ecb5e14378a2e9c32d339e92acb494e688b821df93de33acbR343), they should probably be moved to the mapper as well.

In this PR I'm rolling back some of the changes I did, and since this is already hard enough, I didn't want to add more complexity to it by refactoring the operations layer. I'll create a new ticket to deal with these validations.

lucaCambi77 (approved):

PR #15384, opened 2023-10-12
feat: Add createdAtClient to relationships [DHIS2-15666]

enricocolasante:
Adding `createdAtClient` field to relationship entity in tracker (not adding it in deprecated tracker) and adding the possibility to order by this field.
`createdAtClient` was also added to the list of default fields to be returned when exporting a relationship.

**Changes required to support new field**
- Flyway script to add the column in the relationship table
- Add the field in Hibernate mapping for relationship
- Update the relationship object in all the layers and the relative mappers
- Update aggregate importer getting relationships inside a TE
- Update all the tests that were importing/exporting relationships
- Added tests for the ordering by the new field

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/RelationshipTrackerConverterService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/RelationshipTrackerConverterService.java
@@ -120,16 +120,17 @@ public class RelationshipTrackerConverterService
     org.hisp.dhis.relationship.RelationshipItem toItem =
         new org.hisp.dhis.relationship.RelationshipItem();
 
+    Date now = new Date();
     if (toRelationship == null) {
-      Date now = new Date();
 
       toRelationship = new org.hisp.dhis.relationship.Relationship();
       toRelationship.setUid(fromRelationship.getRelationship());
       toRelationship.setCreated(now);
-      toRelationship.setLastUpdated(now);
     }
+    toRelationship.setLastUpdated(now);
 
     toRelationship.setRelationshipType(relationshipType);
+    toRelationship.setCreatedAtClient(DateUtils.fromInstant(fromRelationship.getCreatedAtClient()));
 
     if (fromRelationship.getRelationship() != null) {
       toRelationship.setUid(fromRelationship.getRelationship());
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_29__Add_createdatclient_to_relationship.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_29__Add_createdatclient_to_relationship.sql
@@ -1,3 +1,1 @@
 alter table relationship add column if not exists "createdatclient" timestamp;
-
-update relationship set createdatclient = created where createdatclient IS NULL;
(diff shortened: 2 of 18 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/RelationshipTrackerConverterService.java
```diff
@@ -120,16 +120,17 @@ private org.hisp.dhis.relationship.Relationship from(
...
 
       toRelationship = new org.hisp.dhis.relationship.Relationship();
       toRelationship.setUid(fromRelationship.getRelationship());
       toRelationship.setCreated(now);
-      toRelationship.setLastUpdated(now);
     }
+    toRelationship.setLastUpdated(now);
 
     toRelationship.setRelationshipType(relationshipType);
+    toRelationship.setCreatedAtClient(DateUtils.fromInstant(fromRelationship.getCreatedAtClient()));
```
muilpp, line 133: Maybe it's just Github, but the format here seems a bit off.
While we are at it, why do we not use the `Date now` variable to set the creation date here?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
```diff
@@ -245,11 +246,28 @@ void getRelationshipsByEvent() {
...
     JsonObject relationship = assertFirstRelationship(r, relationships);
-    assertHasOnlyMembers(relationship, "relationship", "relationshipType", "from", "to");
+    assertHasOnlyMembers(
+        relationship, "relationship", "relationshipType", "createdAtClient", "from", "to");
     assertHasOnlyUid(from.getUid(), "event", relationship.getObject("from"));
     assertHasOnlyUid(to.getUid(), "trackedEntity", relationship.getObject("to"));
   }
 
+  @Test
+  void getRelationshipsByEventWithAllFields() {
```
muilpp, line 256: are we validating anything related to the field `createdAtClient` in this test?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_29__Add_createdatclient_to_relationship.sql
```diff
@@ -0,0 +1,3 @@
+alter table relationship add column if not exists "createdatclient" timestamp;
+
+update relationship set createdatclient = created where createdatclient IS NULL;
```
muilpp, line 3: why do we set it to 'created'?

muilpp (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
enricocolasante: Yes, we are asserting the relationship and in there we are checking that `createdAtClient` is the expected one

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_29__Add_createdatclient_to_relationship.sql
enricocolasante, line 3:
```suggestion
```

vgarciabnz: I think it will be better to leave it null. Sometimes this field is used to tell if the data is coming from Android or not (usually is is null for web, not null for Android)

PR #15389, opened 2023-10-12
refactor: Rename table dataelementcategoryoption to categoryoption [TECH-1636]

abyot:
Remove prefix dataelement from: 

- table names dataelementcategoryoption and dataelementcategory
- all sql statements


jbee (approved):
I assume this change has been approved in the context of a larger effort to update legacy names?
In any case we have to add an item to the release notes for the next version to point out the name change.

abyot:
> I assume this change has been approved in the context of a larger effort to update legacy names? In any case we have to add an item to the release notes for the next version to point out the name change.

Yes, this is part of the larger effort ... especially removing the prefix `dataelement` from category and related models.

enricocolasante (approved):

PR #15399, opened 2023-10-13
fix: Remove redundant count query for TE [DHIS2-15915]

enricocolasante:


enricocolasante:
This seems to be a feature, not a bug.
More discussion is needed if a fix is going to be released

PR #15403, opened 2023-10-14
feat: periodOffset in subExpression [DHIS2-15874]

jimgrace:
Scott tells me that the main requirement for using periodOffset within subexpressions is for data quality. For example, assume that a value for a given aggregate data element is expected to be entered at every facility every month. For example, the following indicator numerator expression will return a 1 for every facility where a value has been entered for each of 12 months ending with the current month, and a 0 otherwise. These numbers are summed up to whatever higher orgUnit level is used for reporting (for example, how many facilities in each district reported every month for the last 12 months). 

```
subExpression(if(#{uid}.aggregationType(COUNT)+#{uid}.aggregationType(COUNT).periodOffset(-1)+
#{uid}.aggregationType(COUNT).periodOffset(-2)+#{uid}.aggregationType(COUNT).periodOffset(-3)+
#{uid}.aggregationType(COUNT).periodOffset(-4)+#{uid}.aggregationType(COUNT).periodOffset(-5)+
#{uid}.aggregationType(COUNT).periodOffset(-6)+#{uid}.aggregationType(COUNT).periodOffset(-7)+
#{uid}.aggregationType(COUNT).periodOffset(-8)+#{uid}.aggregationType(COUNT).periodOffset(-9)+
#{uid}.aggregationType(COUNT).periodOffset(-10)+#{uid}.aggregationType(COUNT).periodOffset(-11)==12,1,0))
```
or, more simply, because the .aggregationType(COUNT) can be applied to a group of items:

```
subExpression(if((#{uid}+#{uid}.periodOffset(-1)+
#{uid}.periodOffset(-2)+#{uid}.periodOffset(-3)+
#{uid}.periodOffset(-4)+#{uid}.periodOffset(-5)+
#{uid}.periodOffset(-6)+#{uid}.periodOffset(-7)+
#{uid}.periodOffset(-8)+#{uid}.periodOffset(-9)+
#{uid}.periodOffset(-10)+#{uid}.periodOffset(-11)).aggregationType(COUNT)==12,1,0))
```

Scott says that the above query might be run for 12 months (each of the 12 months needing to reference data from that month plus the 11 preceding months).

This PR supports such a use case in a single SQL query. It constructs an inline table that maps between the period found in the database and the offset period desired for reporting. So, for example, a single database value for July can be used as the current month value for July, and a periodOffset(-1) value for August, a periodOffset(-2) value for September, etc.

See the initial JavaDoc comments in JdbcSubexpressionQueryGenerator, copied here, for simplified SQL code that demonstrates how a periodOffset would work in when evaluated for the two months 202309 and 202310:

```
 * select uidlevel2, monthly, 'subExpreUid' as dx, sum("A2VfEfPflHV" + "A2VfEfPflHV_minus_1") as value
 * from (select uidlevel2,
 *              shift.reportperiod as monthly,
 *              sum(case when dx = 'A2VfEfPflHV' and shift.delta = 0 then value else null end) as "A2VfEfPflHV",
 *              sum(case when dx = 'A2VfEfPflHV' and shift.delta = -1 then value else null end) as "A2VfEfPflHV_minus_1"
 *       from analytics
 *       join (values(-1,'202309','202308'),(-1,'202310','202309'),
 *                   (0,'202309','202309'),(0,'202310','202310'))
 *              as shift (delta, reportperiod, dataperiod) on dataperiod = monthly
 *       where monthly in ('202308', '202309', '202310')
 *       and dx in ('A2VfEfPflHV') // (greatly improves performance)
 *       group by uidlevel2, monthly, ou) as ax
 * where "A2VfEfPflHV" is not null
 * group by uidlevel2, monthly;
```


maikelarabori (approved):
Nice that you found a way to test the expressions end-to-end.
Cheers!

gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/JdbcAnalyticsManager.java
```diff
@@ -180,6 +182,10 @@ public Future<Map<String, Object>> getAggregatedDataValues(
         params = queryPlanner.assignPartitionsFromQueryPeriods(params, tableType);
       }
 
+      if (params.hasSubexpressions() && params.getSubexpression().hasPeriodOffsets()) {
```
jbee, line 185: Makes me wonder if this now introduces the feature of sub-expressions to the particular expression language used here or if it always had subexpressions already.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
```diff
@@ -0,0 +1,188 @@
...
+ * data period '202309' (Sep 2023).
+ *
+ * @author Jim Grace
+ */
+public class SubexpressionPeriodOffsetUtils {
+  private SubexpressionPeriodOffsetUtils() {
+    throw new UnsupportedOperationException("util");
+  }
+
+  protected static final String SHIFT = "shift";
```
jbee, line 59: `protected` does not make sense here, should be either `private`, package visibility or `public`.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
```diff
@@ -0,0 +1,188 @@
...
+    for (Integer delta : periodOffsets) {
+      for (Period reportPeriod : reportPeriods) {
+        Period dataPeriod = shiftPeriod(reportPeriod, delta);
+        sb.append("(")
+            .append(delta)
+            .append(",'")
+            .append(reportPeriod.getIsoDate())
+            .append("','")
+            .append(dataPeriod.getIsoDate())
+            .append("'),");
```
jbee, line 99: This is much less readable then using a template and inserting the values. Is this really performance critical to build a SQL query so we would need to use a `StringBuilder`?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
```diff
@@ -0,0 +1,188 @@
...
+        .append(" (")
+        .append(DELTA)
+        .append(", ")
+        .append(REPORTPERIOD)
+        .append(", ")
+        .append(DATAPERIOD)
+        .append(") on ")
+        .append(DATAPERIOD)
+        .append(" = ")
+        .append(quote(params.getPeriodType().toLowerCase()));
```
jbee, line 115: Same as above

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/ExpressionItemWithSql.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.hisp.dhis.parser.expression.antlr.ExpressionParser.ExprContext;
+
+/**
+ * A parsed item from an ANTLR expression, using the evaluate method to generate SQL (or to pass
+ * through the call to generate SQL.)
+ *
+ * @author Jim Grace
+ */
+public interface ExpressionItemWithSql extends ExpressionItem {
```
jbee, line 38: Makes we worried :D Sounds like we partially evaluate the tree and then generate some SQL but then remember this in this object? Question is: do we couple the expression SQL generation with this code or is this simply remembering the result of the SQL generated from the expression? And if so is this how it always works or just in some cases?

jbee (commented):
I also wonder if this optimized query also works with arbitrary queries where the subexpression is embedded into as well as arbitrary other math within the subexpression itself or is this kind of detecting a special case and if that is found a special optimized SQL is generated?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/JdbcAnalyticsManager.java
jimgrace: I'm not sure I understand the question. We had subExpressions before (for indicator expressions only). They just didn't support periodOffset until this PR.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
jimgrace: Good point, thanks. Changed.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
jimgrace: Good suggestion, thanks. Performance is not as important here as readability. Changed.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/SubexpressionPeriodOffsetUtils.java
jimgrace: Same. Changed.

jimgrace:
> I also wonder if this optimized query also works with arbitrary queries where the subexpression is embedded into as well as arbitrary other math within the subexpression itself or is this kind of detecting a special case and if that is found a special optimized SQL is generated?

@jbee, Subexpressions work for arbitrary math including parentheses for grouping and common functions like `if`. The math operators and common functions in a subexpression are translated into SQL using the same code that translates these elements from a program indicator expression into SQL.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/JdbcAnalyticsManager.java
jbee: Is this change now also for indicator expressions or some other type of expression so we basically adding the feature of subExpressions to that language. That was my question.

dhis-2/dhis-support/dhis-support-expression-parser/src/main/java/org/hisp/dhis/parser/expression/ExpressionItemWithSql.java
jbee: I don't think visiting multiple times can be avoided. It is simply a necessity of the substitution process. I just wanted to understand if we make this process more complicated or not. From what I get now we have not changed the process but just improved the code executing it which is always good :+1:

jbee (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/JdbcAnalyticsManager.java
jimgrace: Ah, thanks. subExpressions are still only for indicator expressions.

PR #15414, opened 2023-10-16
fix: Read replica database config not detected [DHIS2-15963]

larshelge:
Read replica databases configured in dhis.conf were no longer picked up. The cause was a refactor were `DefaultReadOnlyDataSourceManager` went from being a Spring managed bean to a regular class. This class had an init method managed by Spring `InitializingBean` which no longer would be called.

* Refactors `ReadOnlyDataSourceManager` so that the init method  is called in the main constructor.
* Adds some light refactoring of `ReadOnlyDataSourceManager` to make the code more testable.
* Renames `DefaultReadOnlyDataSourceManager` to `ReadOnlyDataSourceManager`.
* Adds basic testing in `ReadOnlyDataSourceManagerTest`.


dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DefaultReadOnlyDataSourceManager.java
```diff
@@ -1,167 +0,0 @@
...
-
-  /** State holder for the resolved read only data source. */
-  private DataSource internalReadOnlyDataSource;
-
-  /** State holder for explicitly defined read only data sources. */
-  private List<DataSource> internalReadOnlyInstanceList;
-
-  @Override
-  public void afterPropertiesSet() {
-    List<DataSource> ds = getReadOnlyDataSources();
```
larshelge, line 86: This was the source of the problem, where afterPropertiesSet was not called as this class is no longer a Spring-managed bean.

maikelarabori (approved):

PR #15431, opened 2023-10-18
fix: Align all GET missing key responses [DHIS2-10643]

david-mackessy:
# Summary
- This PR brings all `GET` requests for missing/confidential system setting keys in line with each other 
- This [previous PR](#13632) implemented the change for `JSON` requests only, on the API `/api/systemSettings/{key}`
- This work implements the same approach for the:
  - `text` request format on the API `/api/systemSettings/{key}`
  - `JSON` requests param API `/api/systemSettings?key={key}`

## Previous behaviour
  - missing/confidential keys returned an empty string response with `200` status
## New behaviour
  - missing/confidential keys return a `404` status with message `Setting does not exist or is marked as confidential`

## Note
The `html` content type response has been removed as part of this work:
- not sure why it was ever added in the [original PR](#4781)
- we have no existing tests for `html` content type
- can't see why a setting key value would be returned in `html` format
- [docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/settings-and-configuration.html?h=#webapi_system_settings) only mention `JSON` and `plain text` types also

# Testing
## Automated
- multiple tests added covering all valid & invalid `GET` scenarios

## Manual
- to get a system setting key use the APIs
  - `GET` `/api/systemSettings/{key}`
  - `GET` `/api/systemSettings?key={key1},{key2}`
- to get all system settings call:
  - `GET` `/api/systemSettings`

# Docs
- updated [docs PR](https://github.com/dhis2/dhis2-docs/pull/1244)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/SystemSettingController.java
```diff
@@ -191,28 +194,32 @@ public WebMessage setSystemSettingV29(@RequestBody Map<String, Object> settings)
...
 
   @GetMapping(value = "/{key}", produces = ContextUtils.CONTENT_TYPE_TEXT)
-  public @ResponseBody Serializable getSystemSettingOrTranslationAsPlainText(
+  public @ResponseBody ResponseEntity<String> getSystemSettingOrTranslationAsPlainText(
       @PathVariable("key") String key,
       @RequestParam(value = "locale", required = false) String locale,
       HttpServletResponse response,
       @CurrentUser User currentUser) {
+    if (!settingExistsAndIsNotConfidential(key)) {
+      return ResponseEntity.status(404).body(SETTING_DOESNT_EXIST_OR_CONFIDENTIAL);
```
jbee, line 203: Could we also use the `NotFoundException` instead to make this more visible for OpenAPI? Then you should also be able to use `String` as return type.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/SystemSettingController.java
david-mackessy: I did try that, to try to keep them all the same but the server bombs out with a 500. This was the only way I could get a 404 with an error message back to the client, which keeps the behaviour as close as possible.

vietnguyen (approved):

PR #15433, opened 2023-10-18
ci: add workflows to deploy/destroy IM instance on PR

radnov:
This PR enables deploying (and destroying) a DHIS2 instance (with the [Instance Manager](https://github.com/dhis2-sre/im-manager)) via the `deploy` label.

The Deploy workflow will: 
* wait for the "Run api tests" workflow, as this is where we build the PR docker images and it probably doesn't make sense to create an instance before/without the API tests passing
* deploy an instance via the IM, using the image built for the current PR
* wait for the instance to be ready
* trigger analytics generation
* add a comment in the PR with the URL of the instance

The Destroy workflow will clean up the deployed instance when closing/merging the PR.

### DIFF
--- a/.github/workflows/deploy-instance.yml
+++ b/.github/workflows/deploy-instance.yml
@@ -1,0 +1,53 @@
+name: Deploy instance
+
+on:
+  pull_request:
+    types: [opened, reopened, labeled, synchronize]
+
+# Cancel previous runs of the same workflow and PR number or branch/tag
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  deploy-instance:
+    if: contains(github.event.pull_request.labels.*.name, 'deploy')
+    runs-on: ubuntu-latest
+    env:
+      INSTANCE_NAME: pr-${{ github.event.pull_request.number }}
+    steps:
+      - name: Wait for API tests
+        uses: lewagon/wait-on-check-action@v1.3.1
+        with:
+          ref: ${{ github.head_ref }}
+          check-name: 'api-test'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/checkout@v4
+        with:
+          repository: dhis2-sre/im-manager
+          sparse-checkout: scripts/instances
+
+      - name: Install HTTPie
+        run: python -m pip install httpie
+
+      - name: Deploy DHIS2 instance
+        working-directory: scripts/instances
+        env:
+          HTTP: https --check-status
+          USER_EMAIL: ${{ secrets.IM_BOT_EMAIL }}
+          PASSWORD: ${{ secrets.IM_BOT_PASSWORD }}
+          IM_HOST: 'https://api.im.dhis2.org'
+          IMAGE_REPOSITORY: core-pr
+          IMAGE_TAG: ${{ github.event.pull_request.number }}
+          IMAGE_PULL_POLICY: Always
+          DATABASE_ID: sl-sierra-leone-dev-sql-gz
+        run: ./findByName.sh dev $INSTANCE_NAME && ./restart.sh dev $INSTANCE_NAME || ./deploy-dhis2.sh dev $INSTANCE_NAME
+
+      - name: Wait for instance
+        run: timeout 300 bash -c 'while [[ "$(curl -fsSL -o /dev/null -w %{http_code} https://dev.im.dhis2.org/$INSTANCE_NAME)" != "200" ]]; do sleep 5; done'
+
+      - name: Comment instance URL
+        uses: actions-cool/maintain-one-comment@v3
+        with:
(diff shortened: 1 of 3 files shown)


Philip-Larsen-Donnelly (approved):
LGTM

.github/workflows/deploy-instance.yml
```diff
@@ -0,0 +1,54 @@
...
+          IMAGE_REPOSITORY: core-pr
+          IMAGE_TAG: ${{ github.event.pull_request.number }}
+          IMAGE_PULL_POLICY: Always
+          DATABASE_ID: sl-sierra-leone-dev-sql-gz
+        run: ./findByName.sh dev $INSTANCE_NAME && ./restart.sh dev $INSTANCE_NAME || ./deploy-dhis2.sh dev $INSTANCE_NAME
+
+      - name: Wait for instance
+        run: timeout 300 bash -c 'while [[ "$(curl -fsSL -o /dev/null -w %{http_code} https://dev.im.dhis2.org/$INSTANCE_NAME)" != "200" ]]; do sleep 5; done'
+
+      - name: Comment instance URL
```
tonsV2, line 50: Do we want a comment on every build? Rhetorical :smile:

tonsV2 (approved):

.github/workflows/deploy-instance.yml
radnov: The [actions-cool/maintain-one-comment](https://github.com/actions-cool/maintain-one-comment) I'm using will keep just one comment 👍

amcgee:
Looks good!  Minor potential future improvement - we always add a comment to the PR (whether or not it has a label) that says `Add the "deploy" label to create a preview deployment of this Pull Request`.  That way the comment will always be at the top, right under the description, and when it's updated to link to the instance it will be easier to find than if it's a comment buried further down the thread

amcgee (approved):

stian-sandvold (approved):

.github/workflows/deploy-instance.yml
```diff
@@ -0,0 +1,57 @@
...
+          IMAGE_TAG: ${{ github.event.pull_request.number }}
+          IMAGE_PULL_POLICY: Always
+          DATABASE_ID: sl-sierra-leone-dev-sql-gz
+        run: ./findByName.sh dev $INSTANCE_NAME && ./restart.sh dev $INSTANCE_NAME || ./deploy-dhis2.sh dev $INSTANCE_NAME
+
+      - name: Wait for instance
+        run: timeout 300 bash -c 'while [[ "$(curl -fsSL -o /dev/null -w %{http_code} $INSTANCE_HOST/$INSTANCE_NAME)" != "200" ]]; do sleep 5; done'
+
+      - name: Generate analytics
+        run: curl -X POST -u 'admin:district' "$INSTANCE_HOST/$INSTANCE_NAME/api/resourceTables/analytics"
```
gnespolino, line 52: could we add
`-d 'executeTei=true'`
to this command, to allow the generation of Cross Program Linelist analytics data, which wouldn't be generated otherwise?

radnov: Sure, yeah. I'm just not sure if it could turn out to be a problem for any use case?

david-mackessy:
This is a great option to have, nice job @radnov 
One question that comes to mind, what happens when new code is pushed to an open PR? (e.g. addressing comments)
Does it recognise that a new image has been built and re-deploy the latest image?

radnov:
@david-mackessy Yep, the image will be built again and I've set the pull policy to [`Always`](https://github.com/dhis2/dhis2-core/blob/add-im-workflows/.github/workflows/deploy-instance.yml#L44), so that we get any changes re-deployed.

david-mackessy:
> @david-mackessy Yep, the image will be built again and I've set the pull policy to [`Always`](https://github.com/dhis2/dhis2-core/blob/add-im-workflows/.github/workflows/deploy-instance.yml#L44), so that we get any changes re-deployed.

beautiful 👍

.github/workflows/deploy-instance.yml
Philip-Larsen-Donnelly, line 52: It just occurred to me that we should probably use secrets for the SL login (we might want to use something other than admin:district).

Also, I presume this will fail on fork PRs?

gnespolino: It shouldn't, it's just an extra data export which is probably the fastest among all others.
It will "just" affect the disk space used.
BTW the PR is already good enough to me and what we're discussing might be addressed in another place and in other PRs.
I think we could make all analytics conditional on label like
- `deploy` just deploy the branch
- `deploy-analytics` deploy and runs analytics
- `deploy-analytics-cpl` deploy, runs analytics with `exportTei=true` flag

not sure if we could do that

radnov: We could, but let's try to "keep it simple" and go with generating analytics with the extra Cross Program Linelist data. If any issues come out of it, we can always fix it.

radnov: Good point about the secrets! And yes, with secrets this will fail on fork PRs, which is okay I suppose?

PR #15443, opened 2023-10-19
fix: revert acl check on user group update

netroms:
Original PR: https://github.com/dhis2/dhis2-core/pull/15131
JIRA: https://dhis2.atlassian.net/browse/DHIS2-15787


david-mackessy:
if this is a back port, can you link the PR it's a back port from (like we usually do) so we can see why this back port is being reverted (ideally the original PR gives the reason for reverting).

david-mackessy:
i can see the original PR in the commits - ignore my previous comment

david-mackessy (approved):

jbee (approved):

PR #15465, opened 2023-10-23
fix: /tracker/events?order=<attributeUID> filters out events without attribute [ DHIS2-15679 ]

lucaCambi77:
[DHIS2-15679](https://dhis2.atlassian.net/browse/DHIS2-15679)

Apply the same logic as in the `HibernateTrackedEntityStore` to:
- not exclude (use a LEFT JOIN) in the result  the attributes we use only to order
- use an INNER JOIN otherwise (attributes or attributes with filter) to filter out if they are not present

This PR also does a clean of the `HibernateTrackedEntityStore` to remove redundant left outer joins as apparently they were used to add fields to the select that no longer exist.

Notably, we remove in the `EventQueryParams#orderBy` method the add to the attribute list so it works the same as in the `TrackedEntityQueryParams`. As explained in the issue, for `dataelements` it works differently because it uses a JSONB column

[DHIS2-15679]: https://dhis2.atlassian.net/browse/DHIS2-15679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


muilpp (approved):

enricocolasante (changes requested):
It seems to me that the case `/tracker/events?attribute=<attributeUid>` is not working properly anymore

lucaCambi77:
@enricocolasante can you make an example of the failing expectations for that? I don't see any `/events?attribute` parameter option but maybe I am missing something

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventQueryParams.java
```diff
@@ -450,6 +459,17 @@ public Map<TrackedEntityAttribute, List<QueryFilter>> getAttributes() {
...
   }
 
+  /** Returns attributes that are only ordered by and not present in any filter. */
+  public Set<TrackedEntityAttribute> leftJoinAttributes() {
+    return SetUtils.difference(getOrderAttributes(), filterableAttributes().keySet());
+  }
+
+  public Map<TrackedEntityAttribute, List<QueryFilter>> filterableAttributes() {
+    return attributes.entrySet().stream()
+        .filter(a -> !a.getValue().isEmpty())
```
enricocolasante, line 469: If we filter by an attribute and we do not add any filter on the value this should still translate to a `INNER JOIN`.
Do we have a test like `shouldFilterEventsByAttributeWithNoValuesAndFilterOutEventsWithATrackedEntityWithoutThatAttribute`

enricocolasante (changes requested):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -61,11 +61,12 @@
 
 @ToString
 public class TrackedEntityQueryParams {
-  /**
-   * Each attribute will affect the final SQL query. Some attributes are filtered on, while
-   * attributes added via {@link #orderBy(TrackedEntityAttribute, SortDirection)} will be ordered
-   * by.
-   */
+
+  public static final int DEFAULT_PAGE = 1;
```
enricocolasante, line 65: All this page info must be removed.
It was removed in a previous PR, probably got lost in the merge

enricocolasante (approved):
Just remove the page info in TrackedEntityQueryParams

PR #15474, opened 2023-10-24
fix: Include time when filtering in tracker exporter [DHIS2-16019]

enricocolasante:


muilpp (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -555,6 +556,64 @@ void shouldReturnTrackedEntityIfGivenFilterMatches()
     assertContainsOnly(List.of(trackedEntityA), trackedEntities);
   }
 
+  @Test
+  void shouldReturnTrackedEntityIfTEWasUpdatedAfterPassedDateAndTime()
+      throws ForbiddenException, NotFoundException, BadRequestException {
+    Date oneHourBeforeLastUpdated =
+        Date.from(
+            trackedEntityA
```
lucaCambi77, line 564: I am not following how this date is supposed to test all changes we are applying. Can we make a test where we actually have a date match? Or maybe there is and i do not see it

enricocolasante: We are just testing that the filter on `updatedAt` is working.
We try to get all the TE that were updated in the last hour and we want to assert that they are present.
I am not sure we should assert on the date as the next test is doing the opposite and filter on something that will return no TEs and so there we will not have any date to assert

lucaCambi77: ok but i am a bit surprised that we made these changes and there are no tests failing. So maybe there is no test coverage or it is not fine-grained to time. I see many where conditions are changed now so maybe we could think of some test coverage for that. But maybe i am wrong

enricocolasante: Yeah, you are right.
There are no failures just because this change is not a breaking one, so any case working before is still working but we need to cover more cases.
I am adding the new test cases for all the changes

zubaira (approved):

PR #15494, opened 2023-10-26
fix: only exclude actual 'run once' jobs from scheduler main list [DHIS2-16074]

jbee:
### Summary
The filter applied to the scheduler main list was too unspecific and would also filter out disabled jobs that were removed from a queue and therefore would not have a CRON expression or delay time. 

### Manual Testing
* create some jobs
* create a queue with the new jobs
* delete the queue 
* check the jobs get listed for `/api/scheduler` as disabled

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
@@ -341,7 +341,10 @@ public class JobConfiguration extends BaseIdentifiableObject implements Secondar
   }
 
   public boolean isRunOnce() {
-    return cronExpression == null && delay == null && queueName == null;
+    return schedulingType == SchedulingType.ONCE_ASAP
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
```diff
@@ -341,7 +341,10 @@ public boolean isUsedInQueue() {
   }
 
   public boolean isRunOnce() {
-    return cronExpression == null && delay == null && queueName == null;
+    return schedulingType == SchedulingType.ONCE_ASAP
+        && cronExpression == null
+        && delay == null
+        && queueName == null;
```
jbee, line 347: FYI: the reason to check all the fields is so we do not exclude scheduled jobs which got executed manually so their `schedulingType` shortly switches to `ONCE_ASAP` as well but they still have a `cronExpression.

david-mackessy (approved):

PR #15518, opened 2023-10-27
fix: Make capture mode return enrollments in capture scope [TECH-1635]

muilpp:
Capture mode in /enrollments returns enrollments in search scope, which should not happen.
With this fix, we return only enrollments in the user capture scope.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -28,7 +28,9 @@
 package org.hisp.dhis.tracker.export.enrollment;
 
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ACCESSIBLE;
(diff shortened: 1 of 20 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -204,6 +206,9 @@ public List<Enrollment> getEnrollments(EnrollmentOperationParams params)
         && queryParams.isOrganisationUnitMode(OrganisationUnitSelectionMode.ACCESSIBLE)) {
       queryParams.setOrganisationUnits(user.getTeiSearchOrganisationUnitsWithFallback());
       queryParams.setOrganisationUnitMode(OrganisationUnitSelectionMode.DESCENDANTS);
+    } else if (user != null && queryParams.isOrganisationUnitMode(CAPTURE)) {
```
enricocolasante, line 209: We should apply the same fix to the next method that gets a Page of enrollments

enricocolasante (changes requested):

enricocolasante (approved):

PR #15521, opened 2023-10-28
ci:deploy nightly and stable builds on im-play

tonsV2:
The goal of this PR is to deploy our nightly and stable builds on IM.

In addition to the above, the PR also includes a fix for dev. Previously, the branch wasn't included in the health check. So even if dev-2.40 was being the deploy, dev was the instance being checked.

### DIFF
--- a/jenkinsfiles/canary
+++ b/jenkinsfiles/canary
@@ -41,6 +41,61 @@ pipeline {
                 }
             }
         }
+
+        stage('Update IM Play dev instance') {
+            environment {
+                HTTP = "http --check-status"
+                IM_REPO_URL = "https://github.com/dhis2-sre/im-manager"
+                IM_HOST = "https://api.im.dhis2.org"
+                INSTANCE_URL = "https://play.im.dhis2.org/nightly"
+                IMAGE_REPOSITORY = "core-canary"
+                IMAGE_PULL_POLICY = "Always"
+                FLYWAY_MIGRATE_OUT_OF_ORDER = "true"
+                FLYWAY_REPAIR_BEFORE_MIGRATION = "true"
+                INSTANCE_TTL = "315360000"
+                STARTUP_PROBE_FAILURE_THRESHOLD = "50"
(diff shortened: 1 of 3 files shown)


jenkinsfiles/canary
```diff
@@ -41,6 +41,61 @@ pipeline {
                 }
             }
         }
+
+        stage('Update IM Play dev instance') {
```
Philip-Larsen-Donnelly, line 45: This could be "canary" instances instead of "play" instances of we're being pedantic 🙂

Philip-Larsen-Donnelly (approved):

jenkinsfiles/canary
tonsV2: I'll just remove the dev since the context is already given from the pipeline. I would it concise, not pedantic :slightly_smiling_face:  Thanks :pray:

jenkinsfiles/stable
```diff
@@ -123,6 +123,70 @@ pipeline {
...
+                STARTUP_PROBE_FAILURE_THRESHOLD = "50"
+                LIVENESS_PROBE_TIMEOUT_SECONDS = "3"
+                READINESS_PROBE_TIMEOUT_SECONDS = "3"
+            }
+
+            steps {
+                script {
+                    if(env.DOCKER_IMAGE_TAG.endsWith("rc")) {
+                        catchError(buildResult: 'SUCCESS', stageResult: 'NOT_BUILT') {
+                            error "Skipping..." // Force an error that we will catch to set the stageResult
```
radnov, line 146: If you want to skip this stage when the `env.DOCKER_IMAGE_TAG.endsWith("rc")`, you can use the [`when` directive](https://www.jenkins.io/doc/book/pipeline/syntax/#when).

```
when {
                expression { env.DOCKER_IMAGE_TAG.endsWith("rc") }
}
```

radnov (changes requested):
Left one comment that I think should be addressed, otherwise looks good 👍

radnov (approved):

PR #15527, opened 2023-10-29
fix: event analytics multi-stage DEs [DHIS2-15873]

jimgrace:
See [DHIS2-15873](https://dhis2.atlassian.net/browse/DHIS2-15873).

### The problem

In a program indicator expression, a data element is notated as `#{programStage.dataElement}`, which should refer to the specified data element only within the specified program stage.

For enrollment analytics queries this has long been supported (maybe since they were invented?) A SQL subquery is generated to reference the specified data element in the specified program stage for the enrollment.

However, for event analytics queries this not been supported (maybe since they were invented?) The SQL generated just refers to the data element column with no reference to the program stage. This causes the behavior reported in the ticket: When a data element is assigned to more than one program stage, the `#{programStage.dataElement}` syntax finds all the values for the specified data element in every stage where it exists. When the SQL is generated for an event analytics query, the specified program stage is ignored.

### The fix

It would not be appropriate for event analytics queries to use the same SQL subquery that is used for enrollment analytics queries. This is because the data element value should be returned on an event-by-event basis but only for the specified program stage. The enrollment SQL subquery would find the value for the enrollment, not for the event.

Instead, a case statement is generated that will return the data element value only if the program stage is the one specified. So, for example, the old code would just reference the event column, translating `#{programStage.dataElement}` into the SQL column name:

`    "dataElement"`

The fix is to translate `#{programStage.dataElement}` into the following SQL case statement :

`    case when ax."ps" = 'programStage' then "dataElement" else null end`

This returns the data element value for the specified program stage, and a null value for all other program stages (as if the data element column had no value).

### Possible future work

This PR provides a minimal fix for the problem identified in the ticket. To minimize risk, further code changes were not made, though they could be in the future. In particular:

In `AbstractStatementBuilder` the methods `getProgramIndicatorDataValueSelectSql`, `getProgramIndicatorEventColumnSql`, `getProgramIndicatorEventColumnSql` (overload), `getProgramIndicatorEventInEnrollmentSelectSql`, and `getProgramIndicatorEventInEnrollmentSelectSql` (overload) all call each other and do very similar things. This might be reevaluated in the future to see if we should retain all these methods because of the requirements of the callers, or simplify them (possibly changing the callers) to a simpler set of methods. This analysis and possible change was not done at this time to minimize the risk of fixing the bug.

Note that the SQL subquery for `#{programStage.dataElement}` in event analytics also includes a `reportingStartDate` and/or `reportingEndDate`.  It is my guess that these do not need to be added to the new case statement for  `#{programStage.dataElement}` in an event analytics query because these boundary conditions are already handled in the main query. However, I did not research this assumption because it is outside the scope of this ticket. Again, I made only the minimal change required to fix the bug.

[DHIS2-15873]: https://dhis2.atlassian.net/browse/DHIS2-15873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbee (approved):

maikelarabori (approved):
Hey @jimgrace, after taking a look at the failing e2e, it seems that the changes are safe and the tests needed to be adapted as agreed (I hope they pass also on the server).

Cheers,
Maikel

PR #15534, opened 2023-10-30
feat: add created and lastUpdated fields in Icon object [DHIS2-16045]

zubaira:
PR will add three parameters in Icon metadata.

- created
- lastupdated

[DHIS2-16045](https://dhis2.atlassian.net/browse/DHIS2-16045)
[DHIS2-16043](https://dhis2.atlassian.net/browse/DHIS2-16043)

[DHIS2-16045]: https://dhis2.atlassian.net/browse/DHIS2-16045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-16043]: https://dhis2.atlassian.net/browse/DHIS2-16043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_32__Add_created_lastupdated_column_in_customicon.sql
```diff
@@ -0,0 +1,3 @@
+alter table customicon add column if not exists "created" timestamp;
+alter table customicon add column if not exists "lastupdated" timestamp;
+alter table customicon add column if not exists "custom" BOOLEAN;
```
enricocolasante, line 3: Do we really need this information on the DB?
This value will always be true, there is no use case where this will be false.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconResponse.java
```diff
@@ -56,10 +56,14 @@ public class IconResponse {
   @JsonProperty("href")
   private String reference;
 
-  public IconResponse(String key, String description, String[] keywords, String reference) {
+  @JsonProperty private boolean custom;
+
+  public IconResponse(
```
enricocolasante, line 61: We should expose `created` and `lastUpdated` as well, shouldn't we?

enricocolasante, line 61: We should expose `created` and `lastUpdated` as well, shouldn't we?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
```diff
@@ -554,18 +746,24 @@ public enum Icons {
 
     @Getter private final String[] keywords;
 
-    Icons(String key, String description, String[] keywords) {
+    @Getter private final boolean custom;
+
+    Icons(String key, String description, String[] keywords, boolean custom) {
```
enricocolasante, line 751: `custom` shouldn't be part of the constructor and `isCustom` should always return false

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIcon.java
```diff
@@ -49,4 +49,36 @@ public class CustomIcon implements Icon {
...
 
   private String createdByUserUid;
+
+  private boolean custom;
+
+  private Date created;
+
+  private Date lastUpdated;
+
+  public CustomIcon(
```
enricocolasante, line 59: `custom` shouldn't be part of the constructor and `isCustom` should always return true

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIcon.java
```diff
@@ -49,4 +49,36 @@ public class CustomIcon implements Icon {
   private String fileResourceUid;
 
   private String createdByUserUid;
+
+  private boolean custom;
```
enricocolasante, line 53: This shouldn't have a setter defined by Lombok

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIcon.java
enricocolasante, line 59: `custom` shouldn't be part of the constructor and `isCustom` should always return true

enricocolasante (changes requested):
`custom` flag should only be exposed but not saved in the DB
`created` and `lastUpdated` should be exposed in the response

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
```diff
@@ -29,520 +29,712 @@
...
+
+  private Date created;
+
+  private Date lastUpdated;
+
+  public DefaultIcon(String key, String description, String[] keywords, boolean custom) {
+    this.key = key;
+    this.description = description;
+    this.keywords = keywords;
+    this.custom = custom;
```
muilpp, line 54: similarly to the comment in `customIcon`, I think we custom should always return false in this case.
the way we have it now we could have an instance of `defaultIcon` with the custom flag set to true.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -110,12 +110,15 @@ public interface IconService {
    * @param key the key of the icon to update
    * @param description the new icons description
    * @param keywords the new icons keywords
+   * @param isCustom Icon is CustomIcon or default.
```
muilpp, line 113: can we follow the same structure when writing comments? 
for instance, in the lines above, Icon is written in lower case and CustomIcon is written in camel case but default is not.
If we follow the previous pattern, it should be something like this:
`* @param isCustom icon is custom or default`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -110,12 +110,15 @@ public interface IconService {
...
    * @throws BadRequestException when icon key is not specified
    * @throws NotFoundException when no icon with the provided key exists
    */
-  void updateCustomIcon(String key, String description, String[] keywords)
+  void updateCustomIcon(String key, String description, String[] keywords, boolean isCustom)
```
muilpp, line 117: do we need to update the custom property if we already know the icon is custom?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -164,9 +164,22 @@ public void updateCustomIcon(String key, String description, String[] keywords)
       icon.setKeywords(keywords);
     }
 
+    icon.setCustom(isCustom);
```
muilpp, line 167: same as above, won't this always be true? 
afaik, a custom icon will always stay a custom icon.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconMapper.java
```diff
@@ -60,13 +60,15 @@ public IconResponse from(Icon icon) {
           icon.getKeywords(),
           ci.getFileResourceUid(),
           ci.getCreatedByUserUid(),
-          getCustomIconReference(ci.getKey()));
+          getCustomIconReference(ci.getKey()),
+          icon.isCustom());
```
muilpp, line 64: here we know the icon is custom, can't we just return true here and false in the else block below?
this way we would save all the logic in the service and in the persistence layer

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconResponse.java
zubaira: yes

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
zubaira: that is applicable now but if we need to save default icon's information in database then we need this flag.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
zubaira: resolved

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_36__Add_created_lastupdated_column_in_customicon.sql
```diff
@@ -0,0 +1,3 @@
+alter table customicon add column if not exists "created" timestamp;
+alter table customicon add column if not exists "lastupdated" timestamp;
+alter table customicon add column if not exists "custom" BOOLEAN;
```
enricocolasante, line 3: This flag shouldn't be in the DB for now.
We are not storing default icons in the DB at the moment.
When we will decide to do so we can add this flag.
Having this flag now and propagate it in the different layers is error-prone and quite confusing

enricocolasante (changes requested):
The main issue here is around the custom flag.
We shouldn't over-engineering this and we shouldn't store the flag in the DB for now, we can return static true/false in the domain objects based on the type (custom/default)

muilpp (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
enricocolasante, line 113: This param is not there anymore

enricocolasante (approved):

PR #15547, opened 2023-10-31
fix: Async CSV Event import [DHIS2-12345]

david-mackessy:
# Summary
- Issue reported when async importing `CSV` `Event`s through Import/Export app
- UI would 'hang' waiting for the import summary to show as completed (never completing)
- Async import failing silently
- this is on the API `/api/events`

## Cause
- The async import was failing due to lazy init errors in Hibernate
- This occurred when trying to fetch some nested elements `dataElement -> optionSet -> options`

## Fix
- Bind a Hibernate session to the new thread executing the async task (this behaviour was used in previous async jobs before migrating to the newer Job setup which does not have this issue)

# Testing
## Automated
- added a new async CSV import e2e test

### UI
- follow video instructions on [Jira ticket ](https://dhis2.atlassian.net/browse/DHIS2-12345)

### API
1.
- `POST` `/api/events?skipFirst=true&async=true&dryRun=false&dataElementIdScheme=UID&orgUnitIdScheme=UID&eventIdScheme=UID&idScheme=UID&payloadFormat=csv`
- with sample csv file attached in Jira ticket (`events_sl_239.csv`)
2.
- `GET` `/api/system/tasks/EVENT_IMPORT/{jobId}`
- there should be a task element with a value of `"completed": true`
3.
- `GET` `/api/system/taskSummaries/EVENT_IMPORT/{jobId}`
- to get the task summary


jbee (approved):
As always great you covered it with a test.

mortenoh (approved):

PR #15549, opened 2023-10-31
fix: Return all visible events when no program specified [TECH-1663]

muilpp:
When no program present in `/events`, we need to return everything visible to the user according to its search scope and capture scope, that means:
- events in open or audited programs in search scope
- events in open or audited programs in capture scope
- events in protected or closed programs in capture scope

The changes are applied to both the new and old API.
Ticket: https://dhis2.atlassian.net/browse/TECH-1663


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1112,60 +1107,64 @@ private String createDescendantsSql(
...
+          AND + CUSTOM_ORG_UNIT_PATH_LIKE_MATCH_QUERY + customChildrenQuery);
     }
 
-    return ORG_UNIT_PATH_LIKE_MATCH_QUERY
-        + " AND (ou.hierarchylevel = "
-        + params.getOrgUnit().getHierarchyLevel()
-        + " OR ou.hierarchylevel = "
-        + (params.getOrgUnit().getHierarchyLevel() + 1)
-        + " ) ";
+    mapSqlParameterSource.addValue(COLUMN_USER_UID, user.getUid());
```
enricocolasante, line 1135: Would it make sense to move 
`mapSqlParameterSource.addValue(COLUMN_ORG_UNIT_PATH, params.getOrgUnit().getPath());
 mapSqlParameterSource.addValue(COLUMN_USER_UID, user.getUid());`

to `getOrgUnitSql` method and remove the addValue operation from the single methods?
It seems to me that it would simplify the code quite a bit even if we will end up with a value in the parameter source that is not used

enricocolasante (approved):
Just a style suggestion

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: The problem with that would be that modes `CAPTURE` and `ACCESSIBLE` don't have an orgUnit in the request, so calling `params.getOrgUnit().getPath()` in `getOrgUnitSql` would result in a NPE for these two cases.

Also, there's the scheduler running tasks with a null user. Since it's a system user, we assume it will use the mode `ALL`, but in that case calling `user.getUid()` would fail too.

PR #15554, opened 2023-11-01
fix: identify classes with an Integer ID field [DHIS2-16092]

netroms:
# Summary
- Issue reported that sometimes there is a WARN in the log complaining about wrong identifiers being sent to the Hibernate Redis cache invalidation when the identifier is an Integer instead of a Long.
- Added type check on the class id field and cast/parse to Integer if field is an Integer.
- To avoid excessive superclass discovery mechanisms to infer correct class for the ID field, a best effort to check for an Integer ID field on just the incoming class is done, since classes with an Integer ID field does not extend any superclasses.

## Manual test for bug/issue 

1. Start servers with a Redis server instance running already, use following dhis.conf lines to activate Redis cache invalidation:
```
redis.host = localhost
redis.port = 6379
redis.cache.invalidation.enabled = on 
```
2. Start at least two DHIS2 instances with the config above.
3. Log in to a server and go to the message inbox, like: [https://play.dhis2.org/dev/dhis-web-messaging](https://play.dhis2.org/dev/dhis-web-messaging/)
4. Compose a new "Feedback message" use any subject and body.
5. When pressing send, observe a WARN in the log looking like this:
```
WARN 2023-10-20T15:32:00,166 Fetching new entity failed, failed to execute get query! entityId=86556, entityClass=class org.hisp.dhis.message.UserMessage (BaseCacheEvictionService.java [lettuce-epollEventLoop-7-2])
org.hibernate.TypeMismatchException: Provided id of the wrong type for class org.hisp.dhis.message.UserMessage. Expected: class java.lang.Integer, got class java.lang.Long
at org.hibernate.event.internal.DefaultLoadEventListener.checkIdClass(DefaultLoadEventListener.java:155) ~[hibernate-core-5.4.28.Final.jar:5.4.28.Final]
at
```
## Automatic tests
N/A

## Jira
[DHIS2-16092](https://dhis2.atlassian.net/browse/DHIS2-16092)

[DHIS2-16092]: https://dhis2.atlassian.net/browse/DHIS2-16092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/CacheInvalidationListener.java
+++ b/dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/CacheInvalidationListener.java
@@ -29,6 +29,7 @@ package org.hisp.dhis.cacheinvalidation.redis;
 
 import io.lettuce.core.pubsub.RedisPubSubListener;
 import java.io.Serializable;
+import java.lang.reflect.Field;
 import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
 import org.hibernate.SessionFactory;
@@ -40,7 +41,6 @@ import org.hisp.dhis.common.IdentifiableObjectManager;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.dataset.CompleteDataSetRegistration;
 import org.hisp.dhis.dataset.DataSet;
-import org.hisp.dhis.datastatistics.DataStatisticsEvent;
 import org.hisp.dhis.datavalue.DataValue;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.period.Period;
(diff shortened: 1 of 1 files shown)


mortenoh (approved):

vietnguyen (approved):

dhis-2/dhis-support/dhis-support-cache-invalidation/src/main/java/org/hisp/dhis/cacheinvalidation/redis/CacheInvalidationListener.java
```diff
@@ -158,9 +158,18 @@ private Serializable getEntityId(String message) throws ClassNotFoundException {
...
+        Field idField = entityClass.getDeclaredField("id");
+        Class<?> idType = idField.getType();
+        if (idType == int.class) {
+          return Integer.parseInt(idPart);
+        }
+      } catch (NoSuchFieldException e) {
+        // Ignore this exception, as it's expected unless it's a class with an int ID.
+      }
+      // In most cases the ID will be a long.
       return Long.parseLong(idPart);
```
david-mackessy, line 173: It looks like if there is no id field and the NoSuchFieldException is thrown that we still try parse the `idPart` as a long. Is that what we want to try here?
Some tests for this logic would be beneficial I think with some many conditional branches in this method.

netroms: My thinking here is that if there is an ID field on the base class it will be an Integer. If you do a search for "int id;" you will see the result is only classes that has the ID field on the base class. So when it's not an int field, it will be a Long, mostly via the BaseIdentifiableObject. Making logic to catch every single case here is not the intention.

PR #15568, opened 2023-11-02
fix: (2.38) Enrolment notification in new capture app[DHIS2-15724] (#15327)

zubaira:


enricocolasante (changes requested):
Transactional was not present in 2.38.
If notifications are not working in 2.38 then there must be another issue

zubaira:
already supported

PR #15576, opened 2023-11-02
fix: Migrate event status VISITED to ACTIVE [TECH-608]

abyot:


enricocolasante (changes requested):
We should have at least a test verifying the change of the PR.

`VISITED` status is present in a lot of different places in the system, it would be good to have a more organic solution for this migration.
In my head `VISITED` should be a status that makes sense only for the client, then in our service layer and DB layer, we shouldn't even know that such status exists.
Any if that check if the status is `VISITED` is a potential source of errors and this checks should be grouped just in one layer of the system

abyot:
> We should have at least a test verifying the change of the PR.
> 
> `VISITED` status is present in a lot of different places in the system, it would be good to have a more organic solution for this migration. In my head `VISITED` should be a status that makes sense only for the client, then in our service layer and DB layer, we shouldn't even know that such status exists. Any if that check if the status is `VISITED` is a potential source of errors and this checks should be grouped just in one layer of the system

This is a minor issue with a deprecated API. Otherwise, we have a preprocessor and associated test that converts all incoming `VISITED` status to `ACTIVE` in the new tracker API.

enricocolasante (approved):

zubaira (approved):

PR #15579, opened 2023-11-02
feat: remove program prefix from TEAs in CPL [DHIS2-16067]

gnespolino:


maikelarabori (commented):
Looks good 👍 
Just a comment: would be nice to have a few e2e for CPL that send requests without the `program`.

maikelarabori (approved):

d-bernat (approved):

PR #15597, opened 2023-11-03
chore: Make async tracker importer use scheduler

enricocolasante:
Created `TrackerJobImport` for async calls of Tracker Importer.
`TrackerImportController` endpoints are converting the incoming request to a `Body` object that adds some logic to be able to process the import. In particular tracker importer allows the client to send entities to be created without a UID and the `BodyConverter` creates all missing UIDs in the payload.
So `TrackerObjects` is the object that is serialized/deserialized and saved as a resource to be used by the `Scheduler`

The big amount of changed files is caused by the change of some key signatures in the importer and this was necessary because in the previous implementation there was no good distinction between data and parameters.

Other changes:
- Removing timingStats from `ImportReport`, they are going to be present in `JobProgress`


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportService.java
```diff
@@ -27,20 +27,34 @@
...
+import org.hisp.dhis.scheduling.NoopJobProgress;
+import org.hisp.dhis.tracker.imports.domain.TrackerObjects;
 import org.hisp.dhis.tracker.imports.report.ImportReport;
 
 /**
  * @author Morten Olav Hansen <mortenoh@gmail.com>
  */
 public interface TrackerImportService {
+  /**
+   * TO FINISH
```
david-mackessy, line 40: is this a TODO? if it is can TODO be used so it is easily trackable?

david-mackessy:
Should there be an actual TrackerJob of some sort that implements the `org.hisp.dhis.scheduling.Job` interface somewhere? AFAIK that needs to be done for the Scheduler to pick them up and call `execute` on.

enricocolasante:
> Should there be an actual TrackerJob of some sort that implements the `org.hisp.dhis.scheduling.Job` interface somewhere? AFAIK that needs to be done for the `JobScheduler` to pick them up and call `execute` on.

Exactly and this is done in `org.hisp.dhis.webapi.controller.tracker.imports.TrackerImportJob`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportService.java
enricocolasante: Forgot about that :)

david-mackessy (approved):
nice job and it's covered with e2e tests. Great.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -78,144 +68,76 @@ public class DefaultTrackerImportService implements TrackerImportService {
...
 
-      if (exitOnError(validationReport, params)) {
-        return buildReportAndNotify(params, validationReport, opsTimer, bundleSize);
-      }
+    jobProgress.startingStage("Running PreHeat");
+    TrackerBundle trackerBundle =
+        jobProgress.runStage(() -> trackerBundleService.create(params, trackerObjects, user));
 
-      PersistenceReport persistenceReport = commit(params, opsTimer, trackerBundle);
+    Map<TrackerType, Integer> bundleSize = calculatePayloadSize(trackerBundle);
```
jbee, line 83: I tend to put even such fast and quite safe steps in a stage just in case there is a exception happening it gets handled more gracefully. But it will also work if you don't do that.

jbee (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
enricocolasante: Done

PR #15609, opened 2023-11-06
fix: Async CSV Event import fix [DHIS2-12345]

david-mackessy:
back port of #15547 

- requested to back port to this patch branch for proposed patch release
- patch/2.38.5 is currently frozen
- Platform team decided this fix was important enough to go into 2.38.5 rather that the intended 2.38.6


Philip-Larsen-Donnelly (approved):
Approved as backport

PR #15610, opened 2023-11-06
fix: Skip user scope validation if mode ALL [TECH-1634]

muilpp:
When the requests is run by a superuser, we can skip the org unit scope validation both in the service and in the sql query, because this user needs to have access to all entities in the system regardless of the scope.

Ticket: https://dhis2.atlassian.net/browse/TECH-1634


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapperTest.java
```diff
@@ -203,6 +207,59 @@ void shouldThrowExceptionWhenOrgUnitNotInScope() {
...
+
+    EnrollmentQueryParams queryParams = mapper.map(operationParams);
+    assertContainsOnly(
+        operationParams.getOrgUnitUids(),
+        queryParams.getOrganisationUnits().stream().map(BaseIdentifiableObject::getUid).toList());
+    assertEquals(operationParams.getOrgUnitMode(), queryParams.getOrganisationUnitMode());
+  }
+
+  @Test
+  void shouldFailWhenOrgUnitNotInScopeAndUserHasSearchInAllAuthority() {
```
enricocolasante, line 233: I am confused. What is `SearchInAll` authority adding to the user capabilities?

muilpp: The way I understood it, `SearchInAll` allows the user to run requests with the org unit mode `ALL`, but can't skip user scope validations like a superuser can. That means a user with this authority still needs to have the requested org unit in its search scope in order to see results in the response.

muilpp: I know, for a user who has this authority, I think `ALL`, `ACCESSIBLE` and `DESCENDANTS` in practice are the same thing.

enricocolasante (approved):

PR #15614, opened 2023-11-07
fix: handle versions in subpaths

mortenoh:
## Summary

Fixes an issue with subpath enabling and versions (we only checked for without version), this fix fixes that and assumes first that a version exists (as its what we usually recommend)


jenniferarnesen (commented):
Thanks for the quick work! Would it be possible to add a unit test for it?

jbee (approved):
A comment with examples of the cases would be nice.

mortenoh:
> A comment with examples of the cases would be nice.

Yeah, so for the `Router` we have this `subpath` ability, so if your url ends with `/**` it means subpaths are enabled, so if you go `api/router/your-router/run/a/b/c` the `a/b/c` will be added to the original router url, and this was working fine.

The problem was that the check was a bit naive, so if you had `/api/41/router/your-router/a/b/c` instead, it would not recognize if and think it was a non subpath enabled route.

netroms (approved):

PR #15616, opened 2023-11-07
fix(dataentry-metadata): always include default catcombo

Birkbjo:
Issue raised by JM. 
https://dhis2.atlassian.net/browse/DHIS2-16129

The `dataEntry` app would crash if no available data-elements (across available datasets) included the default catcombo. And the default was still used as the `attribute` `categoryCombo` on the dataSet. This is because the `default` was deliberately removed from `dataSetCategoryCombos` to avoid duplicate defaults, with the false assumption that this would always be included in the `dataElementCategoryCombos`. 

There's probably a better way to do this, so please give feedback! I would prefer not to have to iterate through the list - and instead manipulate the set before the conversion to a sorted list. But I didn't want to make more changes than necessary. 

```java
    if(dataElementCategoryCombos.contains(defaultCategoryCombo)) {
        dataSetCategoryCombos.remove(defaultCategoryCombo);
    }
```
Would also work, but I would prefer the structure of the default-catcombo to be consistent (field filters differ between `dataElementCatCombos` and `dataSetCatCombos`.


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultDataSetMetadataExportService.java
```diff
@@ -159,6 +159,10 @@ public ObjectNode getDataSetMetadata() {
     List<OptionSet> optionSets = sortById(getOptionSets(dataElements));
 
     dataSetCategoryCombos.remove(defaultCategoryCombo);
```
Birkbjo, line 161: This is the reason for the bug - we're removing the default `catCombo` from dataSet. So if this is used as the `attribute` `catCombo`, but not in any `dataElement` - it's not included.

david-mackessy:
Hi @Birkbjo, after having a quick look at the code and the PR comments, I think I grasp the issue.

I see that there are 2 Lists of `CategoryCombo`s in play here:
- `dataElementCategoryCombos`
- `dataSetCategoryCombos`
And it seems that the default `CategoryCombo` may or may not be present in one/either.

I see you would prefer to have it contained in a specific list if possible.
We could just always add it to the set you want it in explicitly after the first Set of `CategoryCombo`s is retrieved for each.
e.g.

```
Set<CategoryCombo> categoryCombos = flatMapToSet(dataElements, DataElement::getCategoryCombos);
categoryCombos.add(defaultCategoryCombo);
```
If it's already present it just ignores the operation.
Then just keep the removal of the duplicate (as is now) in the list you don't want it in.

Adding a test or two to prove the bug exists would help:
- keeping it from reappearing
- help other devs if they wanted to propose an alternative solution (knowing the tests pass after their change)

PR #15624, opened 2023-11-08
ci:remove javadoc and move IM stage to the end

tonsV2:


radnov (commented):
I think it's better to "disable" [the Generate Javadocs and Sync Javadocs stages](https://github.com/dhis2/dhis2-core/pull/15624/files#diff-a1d2e6c7c735f6ceb48974c9a41a7fa8aafbf865ef13e1a08d15e95513dd79d3L129-L153) by commenting them out with some message for the future, instead of completely removing them from the pipeline.

radnov (approved):

tonsV2:
> I think it's better to "disable" [the Generate Javadocs and Sync Javadocs stages](https://github.com/dhis2/dhis2-core/pull/15624/files#diff-a1d2e6c7c735f6ceb48974c9a41a7fa8aafbf865ef13e1a08d15e95513dd79d3L129-L153) by commenting them out with some message for the future, instead of completely removing them from the pipeline.

As mentioned on Slack. There's a great chance we'll never enable this again and even if we'll have it in the git history

Philip-Larsen-Donnelly (approved):

PR #15626, opened 2023-11-08
fix: tracker/events not returning event relationships [DHIS2-16117]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16117

- Populate `includeRelationships` parameter if `relationships` is included in the list of fields in the `EventRequestParamsMapper`
- Adds e2e and unit tests


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -144,7 +149,10 @@ public EventOperationParams map(RequestParams requestParams) throws BadRequestEx
             .attributeFilters(attributeFilters)
             .events(UID.toValueSet(eventUids))
             .enrollments(UID.toValueSet(requestParams.getEnrollments()))
-            .includeDeleted(requestParams.isIncludeDeleted());
+            .includeDeleted(requestParams.isIncludeDeleted())
+            .includeRelationships(
```
enricocolasante, line 153: I think we should follow the same strategy as the other entities and have `EventParams` in `EventOperationParams` instead of `includeRelationships`.
`EventParams` is already correctly populated in `EventFieldsParamMapper`, we can use it here instead of using directly `FieldFilterService`

muilpp (approved):

PR #15629, opened 2023-11-08
fix: data_elements should be repeated for different stages [DHIS2-13779]

gnespolino:


maikelarabori (approved):
Looks good. It would be nice to have a unit test or an e2e test. 👍

gnespolino:
> Looks good. It would be nice to have a unit test or an e2e test. 👍

there's already a (failing) unit test, I'm fixing it to include new cases

d-bernat (approved):

PR #15630, opened 2023-11-08
fix: run continuous jobs immediately [DHIS2-15276]

jbee:
The issue is that during tests when a test e.g. performs an import the actual import is just some milliseconds. But the first import would execute synchronous with the 20sec scheduler loop. So it is done shortly after the loop started to run. The queue would be empty again as the next test is waiting for the previous one to finish which requires finishing the import. So the next test runs again shortly after the loop cycle started, it then adds its import but at that point since the queue was empty the delay of almost 20 seconds has to be "payed" again essentially causing each test that has an import to take 20 seconds which causes overall execution time of the e2e tests go from 15min to 2h. 

To prevent this test issue the continuous jobs are directly added to the queue and also will spawn a worker if no queue was present (and worked) at that time. This should remove the 20 second delay for this type of jobs. The cost is that now there is a concurrent part in the scheduler where both the scheduling loop thread and async web worker threads from `executeNow` will add to the queue and potentially spawn worker.

As the accidental spawning of more than one worker is no issue this does not need to be synchronized the hard way.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobRunner.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobRunner.java
@@ -40,7 +40,27 @@ package org.hisp.dhis.scheduling;
  */
 public interface JobRunner {
 
+  /**
+   * During testing the scheduler might not be active in which case this is false. Otherwise, this
+   * should always be true in a production environment.
+   *
+   * @return true, if the scheduler is running a scheduling loop cycle, otherwise false
+   */
   boolean isScheduling();
 
+  /**
+   * Runs a job if it should now run according to its {@link SchedulingType} and related information
+   * like the CRON expression or the delay time.
+   *
+   * @param config the job to check and potentially run
+   */
+  void runIfDue(JobConfiguration config);
+
+  /**
+   * Manually runs a job. OBS! This bypasses any actual checking if the job is due to run. When this
+   * is called the job will run.
+   *
+   * @param config The job to run.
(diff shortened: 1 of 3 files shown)


enricocolasante (approved):

netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobRunner.java
```diff
@@ -40,7 +40,27 @@
...
+  /**
+   * Runs a job if it should now run according to its {@link SchedulingType} and related information
+   * like the CRON expression or the delay time.
+   *
+   * @param config the job to check and potentially run
+   */
+  void runIfDue(JobConfiguration config);
+
+  /**
+   * Manually runs a job. OBS! This bypasses any actual checking if the job is due to run. When this
```
david-mackessy, line 60: what does OBS stand for? seeing it in the code a good bit and have no idea.

jbee: OBS! is how you say "oops" in Sweden and I think Norway as well. Here it is often used to catch the eye to say: pay attention. I thought it is more international but maybe not. Maybe use the English form? Oops?

david-mackessy: ahhh ok. Sounds like the English alternative to 'N.B.' (which is a Latin acronym 😄). Now that I know, it's fine.

PR #15636, opened 2023-11-09
feat: query API for job configurations with errors [DHIS2-15276]

jbee:
Adds an API endpoint `/api/jobConfigurations/errors` to find and extract those `JobConfiguration`s with particular errors.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfigurationErrorParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfigurationErrorParams.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobRunErrorsParams.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @OpenApi.Ignore @CheckForNull private UID job;
+
+  /** The user that ran the job */
+  @OpenApi.Property({UID.class, User.class})
+  @CheckForNull
+  private UID user;
+
+  /** The earliest date the job ran that should be included */
+  @CheckForNull private Date from;
```
david-mackessy, line 62: Can we use `LocalDate` or `LocalDateTime` for new models seeing as we want to move to use these eventually? Or is there something restricting us from doing this?

david-mackessy (commented):
adding a new API without tests?! If something breaks in the most simple use case, how will we know about it?

jbee:
> adding a new API without tests?! If something breaks in the most simple use case, how will we know about it?

You got me :) Have to add some

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobRunErrorsParams.java
jbee: I think I should try at least, just did this out of a (bad) habit

jbee: We don't have setup a flexible parsing for `LocalDate` in the same way we have for `Date` so for now I keep it `Date`. At some point we should make an attempt to switch dates in APIs over while supporting the parsing flexibility.

vietnguyen (approved):

PR #15640, opened 2023-11-10
fix: Close database connection after test statement.

jason-p-pickering:
Use the try-resource pattern to automatically close the database connection during startup.


david-mackessy (approved):
nice

jbee (approved):

PR #15641, opened 2023-11-10
fix: Include AOC during event csv export [DHIS2-16107]

abyot:


muilpp:
is there a reason why there's no test to validate that the AOC is present in the response?

muilpp (commented):
is there a reason why there's no test to validate the AOC is present in the response?

abyot:
> is there a reason why there's no test to validate the AOC is present in the response?

the issue is more of a typo than basic functionality. not sure why the AOC was not included in the response. I see nothing special with the AOC, it is just another property ... then we need to test for the presence of each property.

muilpp:
> > is there a reason why there's no test to validate the AOC is present in the response?
> 
> the issue is more of a typo than basic functionality. not sure why the AOC was not included in the response. I see nothing special with the AOC, it is just another property ... then we need to test for the presence of each property.

I agree there's nothing special, but I think results of the mapper are a key thing to test. In my mind, we should have a unit test to validate the mapping returns all the properties we expect. In this case, it's probably not worth it for the old API, but in the new one we could add it on the existing test `CsvEventServiceTest#testReadEventsFromFileWithHeader`. @enricocolasante what do you think about it?

enricocolasante:
> > > is there a reason why there's no test to validate the AOC is present in the response?
> > 
> > 
> > the issue is more of a typo than basic functionality. not sure why the AOC was not included in the response. I see nothing special with the AOC, it is just another property ... then we need to test for the presence of each property.
> 
> I agree there's nothing special, but I think results of the mapper are a key thing to test. In my mind, we should have a unit test to validate the mapping returns all the properties we expect. In this case, it's probably not worth it for the old API, but in the new one we could add it on the existing test `CsvEventServiceTest#testReadEventsFromFileWithHeader`. @enricocolasante what do you think about it?

I completely agree.
We can think about adding tests for deprecated code, if it is too difficult or if it doesn't make sense, but if we change anything in not deprecated code, we must have a test.

enricocolasante (approved):

larshelge (approved):

PR #15648, opened 2023-11-10
chore: Rename executiondate to occurreddate in DB [TECH-1615]

enricocolasante:
Following on https://github.com/dhis2/dhis2-core/pull/15642

This PR aims to replace all occurrences of executionDate in the DB and repository layer to occurredDate.
The general rule that we are using is that date fields are suffixed with At in the API and then suffixed with Date in the service and the DB layer.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/params/dimension/DimensionParam.java
```diff
@@ -241,7 +241,7 @@ public enum StaticDimension implements TeiHeaderProvider {
     ENROLLMENTDATE(DATETIME, DimensionParamObjectType.PERIOD),
     ENDDATE(DATETIME, DimensionParamObjectType.PERIOD),
     INCIDENTDATE(DATETIME, DimensionParamObjectType.PERIOD),
-    EXECUTIONDATE(DATETIME, DimensionParamObjectType.PERIOD),
+    occurreddate(DATETIME, DimensionParamObjectType.PERIOD),
```
enricocolasante, line 244:
```suggestion
    EXECUTIONDATE(DATETIME, DimensionParamObjectType.PERIOD),
```

gnespolino (approved):
LGTM, api-e2e runs fine, so it should be good to go

lucaCambi77 (approved):

PR #15652, opened 2023-11-11
chore: Rename endDate to completedDate in service layer [TECH-1665]

enricocolasante:
`completedAt` field in `Enrollment` was internally translated to `endDate`.
This PR aims to replace all occurrences of `endDate` in the service layer to `completedDate`.
The general rule that we are using is that date fields are suffixed with `At` in the API and then suffixed with `Date` in the service and the DB layer.

**Next PRs**
- Rename `occurredAt` in service layer
- Rename `completedAt` in DB layer
- Rename `occurredAt` in DB layer

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/Enrollment.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/Enrollment.hbm.xml
@@ -32,7 +32,7 @@
 
     <property name="enrollmentDate" column="enrollmentdate" not-null="true"/>
 
-    <property name="getCompletedDate" column="enddate"/>
+    <property name="completedDate" column="enddate"/>
 
     <property name="followup" column="followup"/>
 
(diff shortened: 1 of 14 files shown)


muilpp:
just out of curiosity, is there a plan to name all these fields the same regardless of the layer they live in?
for instance, will we choose between `completedAt` or `completedDate` and use either of them everywhere?

muilpp (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/Enrollment.hbm.xml
```diff
@@ -32,7 +32,7 @@
 
     <property name="enrollmentDate" column="enrollmentdate" not-null="true"/>
 
-    <property name="endDate" column="enddate"/>
+    <property name="completedDate" column="enddate"/>
```
zubaira, line 35: enrollmentDate is mapped to enrollmentdate column. Can we do the same for completedDate?

zubaira (approved):

PR #15661, opened 2023-11-13
Use try-with-resources to close BatchHandler

jason-p-pickering:
This is a continuation of https://dhis2.atlassian.net/browse/DHIS2-15362 which seemed not to fully solve the issue with leaked DB connections. BatchHanlder from the quick package is now `AutoCloseable` and can thus be used in a try-with-resources block. This should help to ensure that all DB connections are completely closed.


jbee (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcOwnershipAnalyticsTableManager.java
```diff
@@ -178,30 +178,32 @@ private void populateTableInternal(AnalyticsTablePartition partition, String sql
...
+            queryRowCount.getAndIncrement();
+          });
+
+      log.info(
+          "OwnershipAnalytics query row count was {} for {}",
+          queryRowCount,
+          partition.getTempTableName());
+      batchHandler.flush();
+    } catch (Exception e) {
+      log.error(e.toString());
```
jbee, line 205: I think better use `error("Failed ...???", ex)` so that the stack trace is available in the logs

jbee (approved):

david-mackessy (approved):

PR #15686, opened 2023-11-14
fix: wait until file resource content is available

jbee:
`FileResource` content is stored asynchronously which makes it difficult to work with immediately after calling "save". 
Therefore a sleep-waiting was added.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/fileresource/FileResourceStorageStatus.java
```diff
@@ -33,6 +33,5 @@
 public enum FileResourceStorageStatus {
   NONE, // No content stored
   PENDING, // In transit to store, not available
-  FAILED, // Storing the resource failed
```
jbee, line 36: FYI: This was never used anywhere so it seemed just confusing to have

PR #15695, opened 2023-11-15
fix: return only UserLookup instances from linkedAccount end-point

netroms:
# Summary

- Changes the linkedAccount to return UserLookup instances instead of full User objects to limit the output to only needed fields.
- Extends the UserLookup to include role IDs and group IDs, if it is populated, which we do manually in the linkedAccount end-point.

Jira: [DHIS2-15701](https://dhis2.atlassian.net/browse/DHIS2-15701)

[DHIS2-15701]: https://dhis2.atlassian.net/browse/DHIS2-15701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mortenoh (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AccountController.java
```diff
@@ -511,8 +515,27 @@ public ResponseEntity<Map<String, String>> updatePassword(
...
+              linkedUserAccounts.get(i).getUserRoles().stream()
+                  .map(UserRole::getUid)
+                  .collect(Collectors.toSet()));
+      userLookups
+          .get(i)
+          .setGroups(
+              linkedUserAccounts.get(i).getGroups().stream()
+                  .map(UserGroup::getUid)
+                  .collect(Collectors.toSet()));
+    }
```
vietnguyen, line 536: Would be great if this can be pushed to service layer so we can have unit test or even better to the store layer with custom query to improve performance. Because `getUserRoles()` and `getGroups()` still trigger lazy select queries to database the performance is going to be affected if we have large collection. Another option is that the method `HibernateUserStore..getLinkedUserAccounts(currentUser)` could have `FetchType.EAGER`.

If we want to keep current solution then it seems there are some unnecessary for loops. Maybe we can rewrite as below:
```

UserLookups userLookups = new UserLookups();
linkedUserAccounts.foreach( u -> userLookups.add(UserLookUp.fromUser(user)));
```

Then  in `UserLookup.fromUser(user)`  we can set `roles` and `groups`
```
lookup.setDisplayName(displayName);
lookup.setRoles(user.getUserRoles().map(UserRole::getUid));
lookup.setGroups(user.getGroups().map(UserGroup::getUid));
```

david-mackessy:
This is a change in API behaviour without adding/updating any tests. How would we know if something is broken and roles and groups are not being returned as expected?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AccountController.java
jbee: Good suggestion :+1:

netroms: This is just a very quick way to make this work for linkedAccounts, which a special feature only in use by PEPFAR right now. I also doubt there will be many users returned as linked accounts.
The problem with setting roles and groups in the fromUser method is that we don't want to show roles and groups in the regular use case for UserInfo.

netroms: This is just a very quick way to make this work for linkedAccounts, which a special feature only in use by PEPFAR right now. I also doubt there will be many users returned as linked accounts.
The problem with setting roles and groups in the fromUser method is that we don't want to show roles and groups in the regular use case for UserInfo.

netroms: I guess we can make a special LinkedUser DTO object instead, to maybe make this cleaner.

vietnguyen: @netroms ok, then I think simply call `setRoles()` and `setGroups()` inside `UserLoopkup.fromUser(user)` instead of using the for loop would make this cleaner.

netroms: That would not work in the other use cases when we don't want to have the roles and groups included.

vietnguyen (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AccountController.java
netroms: Ideally, we might just make a special SQL query to only return username, name, and role IDs and group IDs. Could be nice in other cases too.

netroms:
> This is a change in API behaviour without adding/updating any tests. How would we know if something is broken and roles and groups are not being returned as expected? Or do existing tests cover it? (if so, calling it out helps reviewers)

There are no tests for this feature yet.

jimgrace (approved):

benguaraldi (approved):
Looks good to me!

FWIW, in our use case, we sometimes have people with upwards of 30 linked accounts.

jbee (approved):

PR #15727, opened 2023-11-16
feat: Refactoring of CurrentUserService and introduction of UserDetails and more

netroms:
# Summary
This PR includes several big changes.

1. Removal of the `CurrentUserService` and introduction of `UserDetails` (renamed from `CurrentUserDetails`), as an immutable DTO of the persisted User object. The `UserDetails` DTO purpose is to represent a “hydrated” and detached from persistence version of the User object during the user's “session”.

2. Removal of `getCurrentUserGroupInfo()` as a method for keeping up-to-date state on `User`/`UserGroup` relationship. This is no longer encouraged and developers should rather rely on hydrated and up to date UserDetail representation of the User and it's relationships. This however is not currently (in this PR) 100% finished as we need to invalidate logged-in users the same way we invalidate users when role membership etc. change. For the future, we will want to update/rebuild the in memory representation (`UserDetails`) on such changes. For now, we need to keep `getCurrentUserGroupInfo()` in one place. See: `InternalHibernateGenericStoreImpl.getSharingPredicates()`

3. Changing the `HibernateIdentifiableObjectStore` save/update methods to not accept NULL `User`/`UserDetails` parameters. This is only part of an ongoing refactoring to remove code relying on a NULL user object parameters to implicitly refer to a superuser.

4. Almost all tests now have a proper admin user during test setup. This still need more refactoring to be more consistent through all tests. We want to have fewer choices on making users during test setup. 

5. Removal of deprecated Spring Security configuration methods as a preparation for Spring Security 6, which removes the deprecated code.

6. Introduction of `SystemUser` for internal use. This should be used when there are async or internal jobs doing work. Note for further refactoring that we should remove usage of Spring's `@async` for async jobs, and rather rely on the job scheduler for such work.  

7. Moved all `SecurityService` methods into `UserService` to avoid cyclic dependency issues. Needs further refactoring to clean up unused methods etc.

### Issues that requires further work/PRs:
1. Refactor rest of the User object input parameter use into `UserDetails`. In places where this is still a problem, the usage of `UserDetails.fromUser()` is an indication that we have to convert back to a User object from `UserDetails`.
2. Make tests setup of user related objects more consistent, fewer ways to do the same thing.
3. Log out users when their `UserGroup` change, to refresh their `UserDetails`, see point 2 above.
4. Upgrade to Spring Security 6
5. Remove all `@async` usage and convert to jobs via job scheduler.
6. Refactor redundant and not used methods, that were moved from `SecurityService` into `UserService`.
7. To fix remaining cyclic dependency issues, a search for use of `@lazy` annotations is a good way to start.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectStore.java
```diff
@@ -430,26 +391,26 @@ T getByUniqueAttributeValue(
    * Look up list objects which have property createdBy or lastUpdatedBy linked to given {@link
    * User}
    *
-   * @param user the {@link User} for filtering
+   * @param userDetails the {@link User} for filtering
```
david-mackessy, line 394: javadoc @link param should reflect the change to use the `UserDetails` class also, as that is what will be used for filtering not the `User` class.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectStore.java
```diff
@@ -430,26 +391,26 @@ T getByUniqueAttributeValue(
...
    * @return List of objects found.
    */
-  List<T> findByUser(@Nonnull User user);
+  List<T> findByUser(@Nonnull UserDetails userDetails);
 
   /**
    * Look up list objects which have property lastUpdatedBy linked to given {@link User}
    *
-   * @param user the {@link User} for filtering
+   * @param userDetails the {@link User} for filtering
```
david-mackessy, line 402: javadoc link param should reflect the change to use the `UserDetails` class also, as that is what will be used for filtering not the `User` class.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectStore.java
```diff
@@ -430,26 +391,26 @@ T getByUniqueAttributeValue(
...
    * @return List of objects found.
    */
-  List<T> findByLastUpdatedBy(@Nonnull User user);
+  List<T> findByLastUpdatedBy(@Nonnull UserDetails userDetails);
 
   /**
    * Look up list objects which have property createdBy linked to given {@link User}
    *
-   * @param user the {@link User} for filtering
+   * @param userDetails the {@link User} for filtering
```
david-mackessy, line 410: javadoc link param should reflect the change to use the `UserDetails` class also, as that is what will be used for filtering not the `User` class.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElementStore.java
```diff
@@ -106,5 +106,13 @@ public interface DataElementStore extends GenericDimensionalObjectStore<DataElem
    */
   List<DataElement> getDataElementsByAggregationLevel(int aggregationLevel);
 
+  /**
+   * Returns all DataElements which the user hav access to.
```
david-mackessy, line 110: typo:
`hav` -> `has`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
```diff
@@ -557,6 +557,7 @@ public void updateParent(OrganisationUnit newParent) {
    * @return true if this org unit is a descendant of the ancestors.
    */
   public boolean isDescendant(Collection<OrganisationUnit> ancestors) {
+    // TODO: MAS optimize to use Set of UIDs from UserDetails as input instead
```
david-mackessy, line 560: just flagging in case this was not purposely left as a todo (in case you intended to be part of this PR). If intentional then fine.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnitService.java
```diff
@@ -310,7 +304,7 @@ List<OrganisationUnit> getOrganisationUnitsAtLevels(
   Long getOrganisationUnitHierarchyMemberCount(
       OrganisationUnit parent, Object member, String collectionName);
 
-  OrganisationUnitDataSetAssociationSet getOrganisationUnitDataSetAssociationSet(Integer maxlevels);
+  OrganisationUnitDataSetAssociationSet getOrganisationUnitDataSetAssociationSet(User user);
```
david-mackessy, line 307: should this be `UserDetails`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnitService.java
```diff
@@ -504,29 +430,11 @@ List<OrganisationUnit> getOrganisationUnitByCoordinate(
    *   <li>Return 1 as fall back.
    * </ul>
    */
-  int getOfflineOrganisationUnitLevels();
+  int getOfflineOrganisationUnitLevels(User user);
```
david-mackessy, line 433: should this be `UserDetails`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/SystemUser.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public Set<String> getUserOrgUnitIds() {
+    return Set.of();
+  }
+
+  @Override
+  public boolean hasAnyAuthority(Collection<String> auths) {
+    return false;
```
david-mackessy, line 127: should this be true?
if this method's purpose is to check if a user has a specific auth from a set of auths (which i imagine is to decide whether access is granted on something or not) then should this be true? (ALL auth trumps everything)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/SystemUser.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public boolean isAuthorized(String auth) {
+    return false;
```
david-mackessy, line 132: same with this (as above)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
```diff
@@ -55,6 +58,20 @@ public interface UserService {
 
   String TWO_FACTOR_AUTH_REQUIRED_RESTRICTION_NAME = "R_ENABLE_2FA";
 
+  String RESTORE_PATH = "/dhis-web-commons/security/";
+
+  String TBD_NAME = "(TBD)";
```
david-mackessy, line 63: what's this used for?

dhis-2/dhis-services/dhis-service-acl/src/main/java/org/hisp/dhis/security/acl/DefaultAclService.java
```diff
@@ -413,17 +482,23 @@ public Class<? extends IdentifiableObject> classForType(String type) {
...
+        : getAccess(object, userDetails, HibernateProxyUtils.getRealClass(object));
   }
 
   @Override
   public <T extends IdentifiableObject> Access getAccess(
-      T object, User user, Class<? extends T> objType) {
-    if (user == null || user.isSuper()) {
+      T object, UserDetails userDetails, Class<? extends T> objType) {
+
+    if (userDetails == null || isSuper(userDetails)) {
```
david-mackessy, line 501: is this similar to the old pattern we're trying to remove? (treating null user as super, now treating null userDetails as super?)

dhis-2/dhis-services/dhis-service-acl/src/main/java/org/hisp/dhis/security/acl/DefaultAclService.java
```diff
@@ -530,99 +618,97 @@ public <T extends IdentifiableObject> List<ErrorReport> verifySharing(T object,
...
-              object.getClass(), ErrorCode.E3001, user.getUsername(), object.getClass()));
+          new ErrorReport(object.getClass(), ErrorCode.E3001, userDetails, object.getClass()));
     }
 
     return errorReports;
   }
 
-  private boolean haveOverrideAuthority(User user) {
+  private boolean haveOverrideAuthority(UserDetails user) {
+    return user == null || user.isSuper();
```
david-mackessy, line 670: is this similar to the old pattern we're trying to remove? (treating null user as super, now treating null userDetails as super?)

dhis-2/dhis-services/dhis-service-acl/src/main/java/org/hisp/dhis/security/acl/DefaultAclService.java
```diff
@@ -530,99 +618,97 @@ public <T extends IdentifiableObject> List<ErrorReport> verifySharing(T object,
...
+  }
+
+  private boolean isSuper(UserDetails user) {
     return user == null || user.isSuper();
```
david-mackessy, line 674: same here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1036,6 +1027,16 @@ public <T extends IdentifiableObject> List<T> getAllNoAcl(@Nonnull Class<T> type
     return store.getAllNoAcl();
   }
 
+  @Override
+  public void persist(Object userAdmin) {
+    entityManager.persist(userAdmin);
```
david-mackessy, line 1032: with no context, this var name seems odd. Is this an object that being persisted? Why is it called userAdmin?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
```diff
@@ -109,47 +112,53 @@ public void save(@Nonnull T object) {
...
+    }
+    save(object, currentUserDetails, clearSharing);
   }
 
-  private void save(T object, User user, boolean clearSharing) {
-    String username = user != null ? user.getUsername() : "system-process";
+  @Override
+  public void save(@Nonnull T object, @Nonnull UserDetails userDetails, boolean clearSharing) {
+    checkNotNull(object);
+    checkNotNull(userDetails);
```
david-mackessy, line 132: if these objects are marked as @Nonnull params do we need these extra checks?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
```diff
@@ -188,16 +199,33 @@ public void update(@Nonnull T object, @CheckForNull User user) {
...
+    object.setAutoFields();
+
+    // TODO: MAS: id=0 should not be necessary, only needed for tests
+    // TODO: MAS: Replace all usage of id=0 with use of a SystemUser instance
+    // issues with transaction isolation
+    if (userDetails.getId() != 0L && !(userDetails instanceof SystemUser)) {
+
+      // See: https://www.baeldung.com/jpa-entity-manager-get-reference
+      User user = entityManager.getReference(User.class, userDetails.getId());
+      object.setLastUpdatedBy(user);
```
david-mackessy, line 215: does this mean that `lastUpdatedBy` will never show as system user? just wondering for my own understanding.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/InternalHibernateGenericStoreImpl.java
```diff
@@ -268,66 +293,81 @@ public List<Function<Root<T>, Predicate>> getDataSharingPredicates(
...
+    CurrentUserGroupInfo currentUserGroupInfo = new CurrentUserGroupInfo();
+
+    if (CollectionUtils.isEmpty(results)) {
+      currentUserGroupInfo.setUserUID(userUID);
+      return currentUserGroupInfo;
+    }
+
+    for (Object[] result : results) {
+      if (currentUserGroupInfo.getUserUID() == null) {
+        currentUserGroupInfo.setUserUID(result[0].toString());
```
david-mackessy, line 363: is it possible that result[0] could be null?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/hibernate/HibernateDataElementStore.java
```diff
@@ -145,9 +142,32 @@ public List<DataElement> getDataElementsByAggregationLevel(int aggregationLevel)
...
+      throw runtimeException;
+    }
+
+    List<Function<Root<DataElement>, Predicate>> sharingPredicates =
+        getSharingPredicates(
+            builder,
+            user.getUid(),
+            currentUserGroupInfo.getUserGroupUIDs(),
+            AclService.LIKE_READ_METADATA);
+
```
david-mackessy, line 167: would it make sense for the new code here (145 - 167) to be its own method so it doesn't muddy this method which gets data elements?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -66,10 +67,10 @@ public DefaultAggregateAccessManager(AclService aclService, CacheProvider cacheP
   // ---------------------------------------------------------------------
 
   @Override
-  public List<String> canRead(User user, DataValue dataValue) {
+  public List<String> canRead(UserDetails currentUser, DataValue dataValue) {
     List<String> errors = new ArrayList<>();
 
-    if (user == null || user.isSuper()) {
+    if (currentUser == null || currentUser.isSuper()) {
```
david-mackessy, line 73: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -98,46 +99,46 @@ public List<String> canRead(User user, DataValue dataValue) {
   }
 
   @Override
-  public List<String> canWrite(User user, DataSet dataSet) {
+  public List<String> canWrite(UserDetails userDetails, DataSet dataSet) {
     List<String> errors = new ArrayList<>();
 
-    if (user == null || user.isSuper()) {
+    if (userDetails == null || userDetails.isSuper()) {
```
david-mackessy, line 105: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -98,46 +99,46 @@ public List<String> canRead(User user, DataValue dataValue) {
...
     return errors;
   }
 
   @Override
-  public List<String> canRead(User user, DataSet dataSet) {
+  public List<String> canRead(UserDetails currentUser, DataSet dataSet) {
     List<String> errors = new ArrayList<>();
 
-    if (user == null || user.isSuper()) {
+    if (currentUser == null || currentUser.isSuper()) {
```
david-mackessy, line 120: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -98,46 +99,46 @@ public List<String> canRead(User user, DataValue dataValue) {
...
     }
 
     return errors;
   }
 
   @Override
-  public List<String> canWrite(User user, CategoryOptionCombo optionCombo) {
-    if (user == null || user.isSuper()) {
+  public List<String> canWrite(UserDetails userDetails, CategoryOptionCombo optionCombo) {
+    if (userDetails == null || userDetails.isSuper()) {
```
david-mackessy, line 133: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -146,25 +147,25 @@ public List<String> canWrite(User user, CategoryOptionCombo optionCombo) {
...
+    return canDataWriteCocCache.get(cacheKey, key -> canWrite(currentUser, optionCombo));
   }
 
   @Override
-  public List<String> canRead(User user, CategoryOptionCombo optionCombo) {
+  public List<String> canRead(UserDetails currentUser, CategoryOptionCombo optionCombo) {
     List<String> errors = new ArrayList<>();
 
-    if (user == null || user.isSuper()) {
+    if (currentUser == null || currentUser.isSuper()) {
```
david-mackessy, line 160: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultAggregateAccessManager.java
```diff
@@ -173,10 +174,10 @@ public List<String> canRead(User user, CategoryOptionCombo optionCombo) {
   }
 
   @Override
-  public List<String> canWrite(User user, DataElementOperand dataElementOperand) {
+  public List<String> canWrite(UserDetails currentUser, DataElementOperand dataElementOperand) {
     List<String> errors = new ArrayList<>();
 
-    if (user == null || user.isSuper()) {
+    if (currentUser == null || currentUser.isSuper()) {
```
david-mackessy, line 180: same point here about treating null user details as super

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/deduplication/hibernate/HibernatePotentialDuplicateStore.java
```diff
@@ -291,15 +283,22 @@ public void moveEnrollments(
...
+            .toList();
 
     enrollmentList.forEach(duplicate.getEnrollments()::remove);
 
+    User currentUser =
+        CurrentUserUtil.getCurrentUserDetails() == null
+            ? null
+            : entityManager.getReference(
+                User.class, CurrentUserUtil.getCurrentUserDetails().getId());
+
```
david-mackessy, line 295: is the ternary check needed here?
would `entityManager.getReference(User.class, CurrentUserUtil.getCurrentUserDetails().getId())` suffice? returning the valid User or null if not found?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramService.java
```diff
@@ -130,12 +128,13 @@ public List<Program> getProgramsByDataEntryForm(DataEntryForm dataEntryForm) {
   @Override
   @Transactional(readOnly = true)
   public List<Program> getCurrentUserPrograms() {
-    User user = currentUserService.getCurrentUser();
-    if (user == null || user.isSuper()) {
+    UserDetails currentUserDetails = CurrentUserUtil.getCurrentUserDetails();
+
+    if (currentUserDetails == null || currentUserDetails.isSuper()) {
```
david-mackessy, line 133: same point here about null user details treated as super user

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobConfigurationService.java
```diff
@@ -94,16 +95,20 @@ public class DefaultJobConfigurationService implements JobConfigurationService {
   private final FileResourceService fileResourceService;
   private final SystemSettingManager systemSettings;
 
+  // TODO: MAS: investigate use of propagation req. new and failing tests
   @Override
-  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  //    @Transactional(propagation = Propagation.REQUIRES_NEW, noRollbackFor =
+  //        Exception.class)
+  @Transactional
```
david-mackessy, line 102: i know this was brought up at the meeting this morning but i thought that we were leaving the requires new in the production code & trying to find a fix in the tests instead?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/jwt/Dhis2JwtAuthenticationManagerResolver.java
```diff
@@ -80,8 +80,9 @@
 @Component
 public class Dhis2JwtAuthenticationManagerResolver
     implements AuthenticationManagerResolver<HttpServletRequest> {
-  @Autowired private UserService userService;
+  //  @Autowired private UserService userService;
```
david-mackessy, line 83: can this be removed if not needed?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/ldap/authentication/CustomLdapAuthenticationProvider.java
```diff
@@ -33,22 +33,23 @@
 import org.springframework.security.ldap.authentication.LdapAuthenticationProvider;
 import org.springframework.security.ldap.authentication.LdapAuthenticator;
 import org.springframework.security.ldap.userdetails.LdapAuthoritiesPopulator;
-import org.springframework.stereotype.Component;
 
 /**
  * @author Viet Nguyen <viet@dhis2.org>
  */
-@Component
```
david-mackessy, line 41: has the removal of this as a spring component been checked? (where is was originally injected elsewhere in the code)

jbee (approved):
* commented code in `SpringSecurityActionAccessResolver`
* is there a benefit to do `UserDetails extends org.springframework.security.core.userdetails.UserDetails`, do we need all spring methods exposed? try to keep "our" API narrow
* the role of new field `userId` in `UserSetting` should be explained because there is also a `User` field
* agree `"system-process_" + CodeGenerator.generateUid();` but should we make this constant for an `SuperUser` instance?
* maybe not use `CurrentUserUtil.getCurrentUserDetails()` within a stream directly but resolve it before going into a stream operation?
* thinking about situations like `user != null ? user.getUsername() : "[None]",` - would it make sense to have a `GuestUser` class instead so `UserDetails` is never null but its methods might return null?
* saw a couple of `// TODO: MAS:` comments 
* `getSharingPredicates` resolves `CurrentUserGroupInfo`, is this "expensive"? we use this method quite a bit
* in some places the `CurrentUserUtil.getCurrentUserDetails()` is checked for null and then a `CurrentUserUtil.getCurrentUserDetails().something()` is done, it might be cheap but we should avoid that and store the instance in a field also for consistency as in theory the result of the call could change between multiple calls
* why do we need more implementations of `UserDetails` than `UserDetailsImpl`and `SuperUser`? e.g. `DhisOidcUser`, can't they just create instances of `UserDetailsImpl`? that is easier to understand and less risk of having unwanted state in the session
* `setActiveLinkedAccounts` uses `Instant.now()` two times instead of using same instance => could be different value
* `UserDetails.fromUser(...)` should be a todo for future PR to eliminate as much as possible by using `UserDetails` higher up (caller)
* `JdbcEventStore#getEventCount` uses `UserDetails.fromUser(user)` as a query parameter, that seems wrong?! those should be a persisted object like the `User`?!
* `EventImporterUserService#getCurrentUser()` seems like an addition that should be deprecated for be replaced with use of `UserDetails` on the caller side?
* `@CurrentUser UserDetailsImpl` in controller parameters should be `@CurrentUser UserDetails` => we know how to fetch the instance in the resolver
* `CurrentUserHandlerMethodArgumentResolver` and `CurrentUserInfoHandlerMethodArgumentResolver` seem to be duplicates now after we merged the old user and userinfo objects 
* `CurrentUserHandlerMethodArgumentResolver`: `if (type == UserDetailsImpl.class) {` => `if (type == UserDetails.class) {`, also the `throw new NotAuthenticatedException();` logic further below done for `User` should be the same for `UserDetails` ?!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserSettingStore.java
```diff
@@ -101,12 +102,12 @@ public void deleteUserSetting(UserSetting userSetting) {
...
-    String hql = "delete from UserSetting us where us.user = :user";
-
-    getSession().createQuery(hql).setParameter("user", user).executeUpdate();
-  }
+  //  @Override
+  //  public void removeUserSettings(String username) {
+  //    String hql = "delete from UserSetting us where us.user.username = :username";
+  //
+  //    getSession().createQuery(hql).setParameter(username, username).executeUpdate();
+  //  }
```
david-mackessy, line 110: if not needed can we remove? if needed can a TODO be added so it's easily identifiable?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -1053,11 +1056,16 @@ private ImportContext createDataValueSetImportContext(
       DataValueSet data,
       BatchHandler<DataValue> dataValueBatchHandler,
       BatchHandler<DataValueAudit> auditBatchHandler) {
-    final User currentUser = currentUserService.getCurrentUser();
+
+    String currentUsername = CurrentUserUtil.getCurrentUsername();
+    User currentUser = userService.getUserByUsername(currentUsername);
```
david-mackessy, line 1061: this pattern seems to come up a lot:
- get current username
- then get current user (using current username from above)

could they be their own method to reduce code?

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/DhisConvenienceTest.java
```diff
@@ -2863,17 +2880,21 @@ protected User preCreateInjectAdminUser() {
...
+
+    UserRole role = createUserRole("Superuser_Test_" + uid, "ALL");
+    role.setUid(uid);
 
     User user = new User();
+    user.setUid(uid);
     user.setFirstName("Admin");
     user.setSurname("User");
     user.setUsername(DEFAULT_USERNAME + "_test");
+    //    user.setUsername(DEFAULT_USERNAME + "_test_" + uid);
```
david-mackessy, line 2893: remove if not needed

david-mackessy (commented):
nice work Morten, It was a big undertaking :)
I would mention UserDetails somewhere in the PR title as that is probably the biggest part of this change.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobConfigurationService.java
netroms: This is updated now, class now checks for running profiles and changes propagation if it's a test running.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobConfigurationService.java
```diff
@@ -94,16 +95,20 @@ public class DefaultJobConfigurationService implements JobConfigurationService {
...
   public String create(JobConfiguration config) throws ConflictException {
     config.setAutoFields();
     jobConfigurationStore.save(config);
     return config.getUid();
   }
 
   @Override
-  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  //  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  @Transactional
```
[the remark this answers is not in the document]
netroms: This is updated now, class now checks for running profiles and changes propagation if it's a test running.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
netroms: Yes, this is something I want to refine further, I was trying to find the most common pattern as I was refactoring here. I want to do this in a follow-up PR. Also, getting rid of more of the user == null means super user.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/hibernate/HibernateDataElementStore.java
netroms: My intention to remove the CurrentUserGroupInfo all together, but I need to do that in a separate PR. The check for size inconsistency was just a convenience during testing. I will remove that.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/InternalHibernateGenericStoreImpl.java
netroms: Don't think so, this code is been unmodified in the codebase for a long time.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEnrollmentService.java
```diff
@@ -253,7 +257,12 @@ public void validate(EnrollmentQueryParams params) throws IllegalQueryException
       throw new IllegalQueryException("Params cannot be null");
     }
 
-    User user = params.getUser();
+    String username =
+        params.getCurrentUserDetails() != null
+            ? params.getCurrentUserDetails().getUsername()
+            : null;
+
+    User user = userService.getUserByUsername(username);
```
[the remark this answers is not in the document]
netroms: There are several null checks on the user in the validation checks in the method, which were there before the refactor, so I assume it is null safe.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
netroms: Yes, this is correct.

PR #15731, opened 2023-11-17
chore/Update information in pom.xml

jason-p-pickering:
Updated some out of date information in the pom.xml

### DIFF
--- a/dhis-2/pom.xml
+++ b/dhis-2/pom.xml
@@ -7,11 +7,12 @@
   <packaging>pom</packaging>
   <name>DHIS 2</name>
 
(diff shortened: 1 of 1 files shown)


dhis-2/pom.xml
```diff
@@ -7,11 +7,11 @@
   <packaging>pom</packaging>
   <name>DHIS 2</name>
```
amcgee, line 8:
```suggestion
  <name>DHIS2</name>
```

amcgee (changes requested):
Thanks Jason!  Added a few comments/suggestions.

Philip-Larsen-Donnelly (approved):

PR #15732, opened 2023-11-17
fix: Move TE request validations to appropiate layers [TECH-1656]

muilpp:
We had a few validations regarding tracked entities that were scattered around the service, outside the mapper. They were mainly `decideAccess()`, `validate()`, and `validateSearchScope()`.
Now the three of them are split and all the contained validations are in the mapper.

Ticket: https://dhis2.atlassian.net/browse/TECH-1656


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -191,9 +190,7 @@ private void validateUser(
       User user, Program program, ProgramStage programStage, OrganisationUnit requestedOrgUnit)
       throws ForbiddenException {
 
-    if (user == null
-        || user.isSuper()
-        || user.isAuthorized(F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS)) {
+    if (user == null || user.isSuper()) {
```
enricocolasante, line 193: This seems like a bug fixing.
Should it be here in this PR?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParamsMapper.java
```diff
@@ -153,6 +155,60 @@ public TrackedEntityOperationParams map(
     return builder.build();
   }
 
+  private void validateRequestParams(TrackedEntityRequestParams params) throws BadRequestException {
+
+    String violation = null;
```
enricocolasante, line 160: Here we are declaring a variable and assign every time a different error.
Can we just remove the variable and throw the exception in the if statements?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParamsMapper.java
```diff
@@ -153,6 +155,60 @@ public TrackedEntityOperationParams map(
...
+    if (params.getEventStatus() != null
+        && (params.getEventOccurredAfter() == null || params.getEventOccurredBefore() == null)) {
+      violation = "Event start and end date must be specified when event status is specified";
+    }
+
+    if (params.getUpdatedWithin() != null
+        && (params.getUpdatedAfter() != null || params.getUpdatedBefore() != null)) {
+      violation =
+          "Last updated from and/or to and last updated duration cannot be specified simultaneously";
+    }
```
enricocolasante, line 200: Can you review the naming here?
For example, we don't have `incident date` anymore and `last updated` is just called updated now.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -106,9 +130,28 @@ public TrackedEntityQueryParams map(TrackedEntityOperationParams operationParams
...
 
+    validateGlobalSearchParameters(params);
+
     return params;
   }
 
+  private List<TrackedEntityType> filterAndValidateTrackedEntityTypes(
+      User user, TrackedEntityType trackedEntityType, Program program) throws BadRequestException {
+    List<TrackedEntityType> trackedEntityTypes = new ArrayList<>();
+    if (trackedEntityType == null) {
```
enricocolasante, line 141: If program is not null, should we still define the `trackedEntityTypes`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -66,22 +75,36 @@ class TrackedEntityOperationParamsMapper {
 
   @Nonnull private final TrackedEntityAttributeService attributeService;
 
+  @Nonnull private final AclService aclService;
+
+  @Nonnull private final TrackedEntityStore trackedEntityStore;
```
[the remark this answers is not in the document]
enricocolasante: I agree with you.
Can we at least add a comment or a TODO to it?

enricocolasante (commented):
Most of the comments are in fact on the old code that was just moved somewhere else but I think that now is the moment to clean up that code too.

The check removed for `F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS` seems a big deal to me, I am not sure if it is supposed to be in this PR

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
enricocolasante: I think so because this probably needs to be backported, right?

muilpp: it's possible yes

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
muilpp: when the program is not null we still need to do it, but that's a good point, because we don't need to do it if a `trackedEntityType` is present in the request.

enricocolasante: If Program is not null, shouldn't be somoething like this?
`trackedEntityTypes = List.of(program.getTrackedEntityType)`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParamsMapper.java
enricocolasante: The parameters used in the payload didn't change so they are not in the release notes.
Here it is just a matter to check how things are called in the if statement and change the error message accordly.
I would even use a notation with back quotes to highlight the parameters that needs to be corrected

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
muilpp: I did it this way because it's how it's done before this refactor.
When no `trackedEntityType` is specified, it gets all the ones accessible by the user. Is there a reason why we should use the one of the program instead?

 https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityService.java#L472

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -106,9 +134,37 @@ public TrackedEntityQueryParams map(TrackedEntityOperationParams operationParams
...
+
+  private List<TrackedEntityType> filterAndValidateTrackedEntityTypes(User user, Program program)
+      throws BadRequestException {
+    List<TrackedEntityType> trackedEntityTypes =
+        trackedEntityTypeService.getAllTrackedEntityType().stream()
+            .filter(tet -> aclService.canDataRead(user, tet))
+            .toList();
+
+    if (program == null && trackedEntityTypes.isEmpty()) {
+      throw new BadRequestException("User has no access to any Tracked Entity Type");
```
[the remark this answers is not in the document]
muilpp: It's also possible to run a request if `trackedEntities` is present, also when `program` and `trackedEntityType` are null.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -106,9 +134,29 @@ public TrackedEntityQueryParams map(TrackedEntityOperationParams operationParams
...
+    if (program != null && program.getTrackedEntityType() != null) {
+      return List.of(program.getTrackedEntityType());
+    } else if (trackedEntityType == null) {
+      return filterAndValidateTrackedEntityTypes(user);
+    }
+
+    return emptyList();
+  }
+
+  private List<TrackedEntityType> filterAndValidateTrackedEntityTypes(User user) {
```
enricocolasante, line 154: I still think we should validate if user has no access to any Tracked Entity Types because if we skip this validation in the case the client specifies something in trackedEntities then we could return some results even if the user has no access to.

muilpp: When we run the query against the DB we join the tracked entity type table and we filter by the value we found here. In case there's no value, the result of the query will be an empty list, so in theory we should not leak any information. But I agree that it's better to stop the request here instead of running a query against a database when we already know we'll return an empty list.

PR #15749, opened 2023-11-21
chore: Remove unnecessary check in CurrentUserUtil. Add tests

david-mackessy:
- remove what seems to be a redundant check in `CurrentUserUtil#getCurrentUserDetails`
  - it always returns `null` if it's a `String` so why do we perform the additional check `if (!"anonymousUser".equals(principal)`
- add some tests for `CurrentUserUtil` checking `isSuper`


jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -94,10 +94,6 @@ public static CurrentUserDetails getCurrentUserDetails() {
     // Principal being a string implies anonymous authentication
     // This is the state before the user is authenticated.
     if (principal instanceof String) {
-      if (!"anonymousUser".equals(principal)) {
-        return null;
-      }
-
```
vietnguyen, line 100: This `anonymousUser` is a hardcoded username for [Anonymous Authentication](https://docs.spring.io/spring-security/reference/servlet/authentication/anonymous.html) provided by Spring Security. 
I don't think we have a unit test for that type of authentication.  So not sure if this is safe to be removed. 
@netroms Can you verify if we need this ?

netroms: Yes, this is a corner case we don't have proper coverage for. There at least were cases related to the old Struts code, that would call this method when there is no user currently logged in, and the principal would be "anonymous". Returning null here is also an issue, because a lot of the code unfortunately treats null as admin. I would not remove this until we are rid of all the old legacy Struts security related code.

netroms (changes requested):
I would not remove this until we get rid of all the Struts security related code.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
david-mackessy: there are 2 flows here:
1. if principal is not anonymousUser -> return null
2. if principal is anonymousUser -> return null
It's not doing anything because it's already wrapped in a String check which returns null

jbee: As David said code wise this has no function that would change the behaviour, maybe we should turn it into a comment instead to document the case needs handling?

david-mackessy: we can have a chat at the next team meeting about it

david-mackessy: agreed at team meeting to merge as this has no impact on the code flow at all.

david-mackessy:
> I would not remove this until we get rid of all the Struts security related code.

agreed at team meeting to merge as this has no impact on the code flow at all.

netroms (approved):

PR #15752, opened 2023-11-22
fix: getScheduledNotifications() should take in DataSetNotificationTrigger

vietnguyen:
### Issue
- `HibernateDataSetNotificationTemplateStore.getScheduledNotifications` takes in parameter `org.hisp.dhis.program.notification.NotificationTrigger` which is incorrect and cause error 
```
java.lang.IllegalArgumentException: Parameter value [SCHEDULED_DAYS_DUE_DATE] did not match expected type [org.hisp.dhis.dataset.notifications.DataSetNotificationTrigger (n/a)]
```
### Fix
- `HibernateDataSetNotificationTemplateStore.getScheduledNotifications` should take in `org.hisp.dhis.dataset.notifications.DataSetNotificationTrigger`
- Added `DataSetNotificationTemplateStoreTest`

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/notifications/DataSetNotificationTemplateService.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/notifications/DataSetNotificationTemplateService.java
@@ -29,7 +29,6 @@ package org.hisp.dhis.dataset.notifications;
 
 import java.util.List;
 import org.hisp.dhis.dataset.DataSet;
-import org.hisp.dhis.program.notification.NotificationTrigger;
 
 /** Created by zubair@dhis2.org on 20.07.17. */
 public interface DataSetNotificationTemplateService {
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/notifications/DataSetNotificationTemplate.java
```diff
@@ -93,6 +93,29 @@ public DataSetNotificationTemplate(
     this.sendStrategy = sendStrategy;
   }
 
+  public DataSetNotificationTemplate(
```
jbee, line 96: Could we use lombok here?

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/notifications/DataSetNotificationTemplate.java
vietnguyen: yes, forgot about lombok.

david-mackessy (approved):

PR #15753, opened 2023-11-22
refactor: system and DB info as immutable objects

jbee:
### Summary
The original goal was to add the DB time info to the system info so it can be considered for time-zone related issues.
The main changes were needed to make the added information be dynamic so the time is not frozen since startup.

The mutable info objects were hard to understand so the main changes become to make them into immutable objects instead.
The `DatabaseInfo` is no longer a singleton bean and must always be obtained from the `DatabaseInfoProvider` as it contains fields that are not stable. 

Also integration tests now use the actual `SystemService` implementation not a mocked one.

### Automatic Testing
Added 2 scenarios that fetch the system info, one with and one without super access.
Added a unit test to check all sensitive information are cleared.

### Manual Testing
* check `/api/system/info` for new property database `time`
* check pressing F5 (refresh) does result in a updated `time` value


david-mackessy:
nice 👍
was `MonitoringService` moved into the API module just to include it in a Javadoc comment?

david-mackessy (approved):

netroms (approved):

jbee:
> why was MonitoringService moved into the API module?

Yes, I moved the interface and data model part of it to the API module.
For most domains this has always been the case. Later the domains were completely encapsulated in their own maven module including their interfaces and data exchange types. This tends to cause cyclic dependencies so whenever I encounter some I now  started to move over the API part of it to the API module. This is also a bit in preparation to simplify the maven module landscape as discussed in https://github.com/dhis2/notes/discussions/342

PR #15754, opened 2023-11-22
feat: Add read ACL check when no namespace protection [DHS2-16134]

david-mackessy:
# Summary
This feature adds an ACL check to `DatastoreService#readProtectedIn` when there is no namespace protection. The following endpoints can be used to test the feature:
- `GET` `/api/dataStore/{namespace}/{key}`
- `GET` `/api/dataStore/{namespace}/{key}/metaData`

## Background
There was already an existing read `Sharing` check performed for namespace entries that had namespace protection.
The read check has been updated to incorporate scenarios where there is no namespace protection in place (namespace protection is only setup programmatically currently. All namespaces setup through the API will not have any namespace protection).

## DatastoreEntry Read criteria when no namespace protection
- Superuser can always read
- Any user can read by default (no Sharing changes made)
- Public access removed & User is not the owner
  - User cannot access
  - User can access when User added to Shared Users
  - User can access when User UserGroup added to Shared UserGroups

# Testing
## Automated
- Multiple e2e tests added to cover different scenarios

## Manual
To manually test this feature, the following endpoints can be used:
- `GET` `/api/dataStore/{namespace}/{key}`
- `GET` `/api/dataStore/{namespace}/{key}/metaData`

To add an entry to a new namespace:

`POST` `/api/dataStore/{myNamespace}/{myKey}` with sample body
```json
{
    "name": "test",
    "project": "dhis2"
}
```

To get the ID of a DataEntry use:
`GET` `/api/dataStore/{namespace}/{key}/metaData`

To remove public access of a DataEntry:

`POST` `/api/sharing?type=dataStore&id={dataStoreEntryId}` with body
```json
{
    "object": {
        "publicAccess": "--------",
        "externalAccess": false,
        "user": {},
        "userAccesses": [],
        "userGroupAccesses": []
    }
}
```

To share access of a DataEntry with another User & remove public access:

`POST` `/api/sharing?type=dataStore&id={dataStoreEntryId}` with body
```json
{
    "object": {
        "publicAccess": "--------",
        "externalAccess": false,
        "user": {},
        "userAccesses": [
            {
                "id": "{userId}",
                "access": "r-------"
            }
        ],
        "userGroupAccesses": []
    }
}
```

To share access with a UserGroup & remove public access use:

```json
{
    "object": {
        "publicAccess": "--------",
        "externalAccess": false,
        "user": {},
        "userAccesses": [],
        "userGroupAccesses": [
            {
                "id": "{userGroupId}",
                "access": "r-------"
            }
        ]
    }
}
```


vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datastore/DefaultDatastoreService.java
```diff
@@ -190,6 +211,23 @@ private <T> T readProtectedIn(String namespace, T whenHidden, Supplier<T> read)
...
 
+  private <T> boolean isDatastoreEntryWithNoProtection(
+      T res, DatastoreNamespaceProtection protection) {
+    return (res instanceof DatastoreEntry) && (protection == null);
+  }
+
+  private <T> boolean isDatastoreEntryAndProtectionSharingRespected(
+      T res, DatastoreNamespaceProtection protection) {
+    return (res instanceof DatastoreEntry)
+        && (protection != null && protection.isSharingRespected());
```
jbee, line 222: I wonder if we simply always do ACL now. When only protected namespaces had sharing as an option it made sense to check this but now if all entries are by default sharing protected it does not make much sense that a protected namespace basically can opt-out of the sharing level protection.

david-mackessy: happy to update if that's what we want to do from now on.

jbee: I think we should :)

david-mackessy: ok, so we'll get rid of the `sharingRespected` property as sharing will always apply from now on. I'll work on that in a separate issue then to keep things separate as it impacts writes as well as reads

PR #15777, opened 2023-11-27
feat: Remove sharingRespected property from namespace protection [DHI…

david-mackessy:
…S2-16215]

# Summary
Now that sharing is fully implemented in the datastore, the `DatastoreNamespaceProtection` property `sharingProtected` is no longer needed. Sharing can be set in the usual manner. This simplifies access to the datastore.

There are 2 levels of access used in the datastore:
- namespace protection
- standard sharing (read/write ACL)

Only programmatic namespaces used the `sharingProtected` property previously. 
All datastore entries are public (rw------) by default and need explicit updating to restrict sharing access.
Datastore entries are considered metadata.

# Testing
## Automated
Multiple e2e tests added for `update` and `delete` scenarios
There's already many `read` datastore entry e2e tests in place

## Manual
Datastore entries should have the following behaviour:
- superuser can create/read/update/delete
- object owner can create/read/update/delete
- basic user can create/read/update/delete by default
- when default public sharing removed:
  - basic user can create/read/update/delete only when given explicit access (through user or usergroup)

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/datastore/DatastoreNamespaceProtection.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/datastore/DatastoreNamespaceProtection.java
@@ -67,8 +67,6 @@ public class DatastoreNamespaceProtection {
 
   private final String namespace;
 
-  private final boolean sharingRespected;
-
   private final ProtectionType reads;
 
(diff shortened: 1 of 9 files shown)


jbee (approved):
Nice one!

vietnguyen (approved):

PR #15783, opened 2023-11-28
fix: re-enable Hibernate QueryCache

vietnguyen:
The class `HibernateConfigurationProvider` is not in used anymore because: 

1. All hibernate configurations property have been moved to `dhis.conf` and can be retrieved by `DhisConfigurationProvider`
2. Ehcache is still used for hibernate second level cache and query cache. However its cluster settings are not in used as we are using Redis to synchronize cache between instances. It's configured in `CacheInvalidationConfiguration`.
4. The hbm.xml files loader has also been moved to `HibernateConfig` class.

So this PR remove `HibernateConfigurationProvider` class and all related files. 

### Test
- Added `HibernateQueryCacheTest` together with new `HibernateCacheBaseTest` because all other tests have query cache disabled.


dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/HibernateEncryptionConfig.java
```diff
@@ -101,7 +101,9 @@ public MethodInvokingFactoryBean addProvider() {
   }
 
   private Object getConnectionProperty(String key, String defaultValue) {
-    String value = hibernateConfigurationProvider.getConfiguration().getProperty(key);
+    String value =
+        dhisConfigurationProvider.getPropertyOrDefault(
+            ConfigurationKey.ENCRYPTION_PASSWORD, "J7GhAs287hsSQlKd9g5");
```
david-mackessy, line 106: this key is marked as sensitive, should we be explicitly setting this in plain sight here or should we be doing something else?
I'm not sure how this key is used, if it's not set in the config file, does it cause issues if it's not set?

netroms: This is suboptimal and is well known.
We should not use a default password here but rather generate a unique on first boot, outside this process and inject it safely. The facilities to do this is lacking and requires a bigger refactoring and feature work to get in place. It is on the security backlog and roadmap.

netroms (approved):

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/HibernateEncryptionConfig.java
david-mackessy: great

david-mackessy (approved):

jbee (approved):

PR #15793, opened 2023-11-29
fix: Investigate simultaneous updates of data values is not updating at all [ DHIS2-16080 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16080

In the old tracker, we have the `/events` endpoint:
` @PutMapping(value = "/{uid}/{dataElementUid}"` 
which is supposed to update the event data values although it goes through the generic event update.
The front end sends multiple requests in a very quick sequence to this endpoint.

There are 2 points about this issue:
- With the `WorkContextLoader`, info about the existing `eventdatavalues` field is preloaded. Therefore, if we have a concurrent update we might preload data element values that are no longer valid and have been updated in the meanwhile
- Apparently, using the spring JdbcTemplate `batchUpdate`, the concurrency and waiting for the same row update is not in place because it has a default isolation level.

### Solution - Relying on `@Transactional`
One possible solution is to process one data element value at a time and save it to the `eventdatavalues` JSON.
This is done using the Postgres function `jsonb_set` or `jsonb_insert` which updates or inserts at specific positions inside the JSON.
`@Transactional `should handle the concurrency for us if we use hibernate queries. Therefore, multiple concurrent requests for the same row are queued and handled by the JPA at the database level.

### Alternative solution
An alternative approach to this issue is to use locking. There are 2 possible locking we can use:

- Optimistic locking: hibernate recommends we use the `@Version` field so we can catch if the row we are updating is the current version or not. This can have a big impact on the current state of the `event` table
- Pessimistic locking. This will eventually result in using the `SELECT .. FOR UPDATE` syntax which can slow down performances and create deadlocks. Note this should correspond to the isolation level of `SERIALIZABLE`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/event/persistence/DefaultEventPersistenceService.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/event/persistence/DefaultEventPersistenceService.java
@@ -27,17 +27,22 @@
  */
 package org.hisp.dhis.dxf2.deprecated.tracker.event.persistence;
 
+import static java.lang.String.format;
 import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
+import javax.persistence.EntityManager;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.dxf2.deprecated.tracker.event.EventCommentStore;
 import org.hisp.dhis.dxf2.deprecated.tracker.event.EventStore;
 import org.hisp.dhis.dxf2.deprecated.tracker.importer.context.WorkContext;
 import org.hisp.dhis.dxf2.deprecated.tracker.importer.mapper.ProgramStageInstanceMapper;
+import org.hisp.dhis.eventdatavalue.EventDataValue;
 import org.hisp.dhis.program.Event;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
@@ -51,6 +56,10 @@ public class DefaultEventPersistenceService implements EventPersistenceService {
   @Nonnull private final EventStore jdbcEventStore;
 
   @Nonnull private final EventCommentStore jdbcEventCommentStore;
+
+  @Nonnull private final EntityManager entityManager;
+
+  @Nonnull private final ObjectMapper mapper;
 
   @Override
   @Transactional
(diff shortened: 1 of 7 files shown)


muilpp:
> @muilpp thanks for reviewing. AFAIK, this is not an endpoint for massive updates, it shouldn't be an issue processing one by one. I haven't tried, but yes you probably could change the `batchupdate` method transaction level to catch a concurrent update. However, you still have the problem of reloading the current state of the `eventdatavalues` field. So, eventually, although this is not the best solution, it should be better than using locking.

Yeah, I agree it's still better than using locks. 
Just one last doubt, when using @Transactional in an environment with concurrent transactions and with our current default isolation level, are we safe to assume that the transactions will be queued? It was my understanding that we don't have such guarantee unless the isolation level is `Serializable`.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/event/persistence/DefaultEventPersistenceService.java
```diff
@@ -112,6 +121,35 @@ public void delete(
...
+            "UPDATE event SET "
+                + "eventdatavalues= (CASE"
+                + "        WHEN eventdatavalues->:de IS NOT NULL"
+                + "        THEN jsonb_set(eventdatavalues, '{%1$s}', (eventdatavalues->:de) || '%2$s')"
+                + "        WHEN eventdatavalues->:de IS NULL"
+                + "        THEN jsonb_insert(eventdatavalues, '{%1$s}', '%2$s')"
+                + "    END) ,"
+                + " lastupdated = current_timestamp "
+                + "WHERE  uid = :event",
+            uid, mapper.writeValueAsString(de));
```
jbee, line 144: Using a text block (tripple quotes) should make this more readable ;)

lucaCambi77:
> > @muilpp thanks for reviewing. AFAIK, this is not an endpoint for massive updates, it shouldn't be an issue processing one by one. I haven't tried, but yes you probably could change the `batchupdate` method transaction level to catch a concurrent update. However, you still have the problem of reloading the current state of the `eventdatavalues` field. So, eventually, although this is not the best solution, it should be better than using locking.
> 
> Yeah, I agree it's still better than using locks. Just one last doubt, when using @transactional in an environment with concurrent transactions and with our current default isolation level, are we safe to assume that the transactions will be queued? It was my understanding that we don't have such guarantee unless the isolation level is `Serializable`.

@muilpp that is a good point. I am not 100% sure. However, i figure that is probably not the case in this context using JdbcTemplate and batchupdate. While using the `entitymanager` it works as expected

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/event/persistence/DefaultEventPersistenceService.java
lucaCambi77: ok thanks @jbee i will

vietnguyen:
I guess there is a chance that the batch update uses a different database transaction from the work context loader which is a readonly transaction. The default read_committed in postgresql doesn't work in that scenario. 
It seems your solution solve that issue by put those operations into one transaction. You could turn on query logs in postgres and double check the result.

vietnguyen (approved):

zubaira (approved):

PR #15801, opened 2023-11-30
chore:Replace organisation units with no groups integrity check with SQL

jason-p-pickering:
Replaces the integrity check relating to organisation units with no groups with an SQL based equivalent.


jbee (approved):

david-mackessy (approved):
nice, and covered with tests, even better 👍

PR #15804, opened 2023-11-30
feat: improved data integrity execution [DHIS2-16223] (2.40)

jbee:
Manual backport of #15789 by picking entire files from `master` for anything directly data integrity related. 
Changes to job scheduling related files were manually changed the same way there were in master.


vietnguyen (approved):

jbee:
> It's a large backport :) Just wondering what is the reasoning behind backporting such a large backport if it's not essential or fixing a bug.

We always had the plan to backport the data integrity improvements, see DHIS2-14849. We just never got around to make a PR that worked. The reason is that the users of data integrity checks are mainly on 2.39 so they would not benefit much if we only have it in 2.40+. Also we want feedback on the extended DI checks. Last but not least DI is a fairly independent isle so it has little risk of breaking anything else and it is well tested so backporting is not such a big risk overall.

PR #15812, opened 2023-12-01
chore: Replace data element no groups Java check with SQL based check

jason-p-pickering:
- Replaces check for aggregate data elements with no groups with a Java based check
- Add unit test in legacy report


david-mackessy (approved):
nice find, the sql type check was always there but never used?

jbee (approved):

PR #15834, opened 2023-12-05
fix: Validate user is allowed to use mode ALL [TECH-1589] [2.38]

muilpp:


muilpp:
not needed, this validation we do in the mapper already

PR #15840, opened 2023-12-06
fix: only restrict set of checks when executing them [DHIS2-16256]

jbee:
### Summary
When executing data integrity checks the default (no set of checks was explicitly specified) is to only run checks that are not marked as "slow". This filter wrongly was also used when expanding checks to read the results from the cache. This meant that slow checks would not be included unless explicitly named.

### Automatic Testing
Modified existing tests with additional checks that verify that reading check descriptions and check summaries after execution does include the slow checks which means they are not filtered.

### Manual Testing
* run checks POST `/api/dataIntegrity/summary` (no `checks` parameter to run all defaults)
* run a specific slow check POST `/api/dataIntegrity/summary?checks=data_elements_aggregate_abandoned`
* wait some
* fetch results GET `/api/dataIntegrity/summary` should include the slow check `data_elements_aggregate_abandoned`


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegritySummaryControllerTest.java
```diff
@@ -56,6 +58,9 @@ void testLegacyChecksHaveSummary() {
       JsonDataIntegritySummary summary = getSummary(check);
       assertTrue(summary.getCount() >= 0, "summary threw an exception");
     }
+    JsonMap<JsonDataIntegritySummary> checksByName =
+        GET("/dataIntegrity/summary?timeout=1000").content().asMap(JsonDataIntegritySummary.class);
+    assertEquals(31, checksByName.size());
```
jbee, line 63: FYI: all of these are slow so any non-zerp number means slow tests are included

jason-p-pickering (approved):
Looks good. Thanks for fixing

PR #15864, opened 2023-12-08
feat: [WIP] add pagination support for icons [DHIS2-16046]

zubaira:
[DHIS2-16046](https://dhis2.atlassian.net/browse/DHIS2-16046)

[DHIS2-16046]: https://dhis2.atlassian.net/browse/DHIS2-16046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
```diff
@@ -87,15 +88,19 @@ public List<CustomIcon> getIconsByKeywords(String[] keywords) {
...
+  public List<CustomIcon> getAllIcons(IconCriteria iconCriteria) {
     final String sql =
         """
             select c.key as iconkey, c.description as icondescription, c.keywords as keywords, f.uid as fileresourceuid, u.uid as useruid
             from customicon c join fileresource f on f.fileresourceid = c.fileresourceid
             join userinfo u on u.userinfoid = c.createdby
             """;
 
-    return jdbcTemplate.query(sql, customIconRowMapper);
+    if (Boolean.TRUE.equals(iconCriteria.isSkipPaging())) {
```
muilpp, line 99: can this be only `if (iconCriteria.isSkipPaging()) {`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -72,8 +72,9 @@ public class DefaultIconService implements IconService {
 
   @Override
   @Transactional(readOnly = true)
-  public List<Icon> getIcons() {
-    return Stream.concat(defaultIcons.values().stream(), customIconStore.getAllIcons().stream())
+  public List<Icon> getIcons(IconCriteria iconCriteria) {
```
muilpp, line 75: should we add a couple of tests to validate the pagination works?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
zubaira: @muilpp Yes, it can be. But Sonarcloud does not like it and complains :)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
zubaira: on the way

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
muilpp: in this case, I think we should use `if (iconCriteria.isPagingRequest())`
https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/webapi/controller/event/webrequest/PagingAndSortingCriteriaAdapter.java#L70

enricocolasante (commented):
The response of the endpoint should be changed as well in order to provide the information of pagination like it is done for any metadata endpoint

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,17 +128,58 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
   @GetMapping
   public @ResponseBody List<IconResponse> getAllIcons(
       @RequestParam(required = false) String[] keywords) {
+
+    IconCriteria iconCriteria = IconCriteria.of(0, 0, true);
```
enricocolasante, line 132: I can understand the idea here but it is quite difficult to deduce the behavior from the code.
Yu could create a static constructor in `IconCriteria` named `SkipPagination` or something similar

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,17 +128,58 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
+      @RequestParam(required = false) String[] keywords) {
+
+    IconCriteria iconCriteria = IconCriteria.of(page, size, skipPaging);
+
+    List<CustomIcon> icons;
+
+    if (keywords == null) {
+      icons = iconService.getCustomIcons(iconCriteria);
+    } else {
+      icons = iconService.getCustomIcons(keywords);
```
enricocolasante, line 159: If there are keywords then there is no pagination?
I am not sure why this distinction is made here.
In my mind, keywords would be a field in `IconCriteria` and you would apply this distinction only in the store where you will construct the query with or without the filter on keywords.
This way, you will also have pagination in any case

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,17 +128,58 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
+    }
+
+    int count = iconService.getIconCount();
+
+    List<IconResponse> iconResponses = icons.stream().map(iconMapper::from).toList();
+
+    PagingWrapper<IconResponse> listPagingWrapper = new PagingWrapper<>();
```
enricocolasante, line 166: This is the pager used in Tracker and we shouldn't use it.
We should use something like here https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java#L199-L269

Anyway, this pagination issue will be tackled in tracker in https://dhis2.atlassian.net/browse/TECH-1670, maybe you can even leave as it is now and we can go back here when we fix Tracker so we will have a common solution for it

enricocolasante (changes requested):
If there are keywords pagination is ignored

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: this has been in use outside the tracker as well.
`ProgramNotificationInstanceController
`DeduplicationController. 
Secondly, the custom icon does not seem to fall in the metadata domain.

zubaira: I agree to have common solution. There has been many ways we can do paging in dhis2-core. We should have a uniform way for pagination implementation.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: when keywords are provided then response will not contains lot of Icons. It is kind of filter that is applied in the query to filter out Icons.

PR #15865, opened 2023-12-08
fix: Missing support for headers + NPE/clean-up [DHIS2-16065]

maikelarabori:
This PR adds support for "headers" param, fixes a NPE and also fixes an exception handling that was not working correctly.

I noticed that exactly the same items and columns were used across many places. I grouped them into an Enum.


gnespolino (approved):
good one 👍

vietnguyen (approved):

PR #15867, opened 2023-12-08
fix: Skip ownership validation when user is authorized [TECH-1589]

muilpp:
We don't want to validate ownership when the user is not super but it's authorized to use org unit mode `ALL`.
Ticket: https://dhis2.atlassian.net/browse/TECH-1589


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -101,7 +102,7 @@ public Enrollment getEnrollment(
       throws ForbiddenException {
     User user = currentUserService.getCurrentUser();
     List<String> errors = trackerAccessManager.canRead(user, enrollment, false);
-    if (!errors.isEmpty()) {
+    if (!errors.isEmpty() && !user.isAuthorized(F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS)) {
```
enricocolasante, line 105: I am a little bit confused.
If user has the authority but he is not using ALL mode, should we still skip the checks

muilpp: nice catch, no we should skip the validation in this case

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -92,16 +94,21 @@ public Enrollment getEnrollment(String uid, EnrollmentParams params, boolean inc
...
+      @Nonnull Enrollment enrollment,
+      EnrollmentParams params,
+      boolean includeDeleted,
+      OrganisationUnitSelectionMode orgUnitMode)
       throws ForbiddenException {
     User user = currentUserService.getCurrentUser();
     List<String> errors = trackerAccessManager.canRead(user, enrollment, false);
-    if (!errors.isEmpty()) {
+    boolean skipValidation =
+        user.isAuthorized(F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS) && orgUnitMode == ALL;
```
enricocolasante, line 110: Based on operation params validations, could we assume that if the mode is ALL, then it was already verified that the user has such authority?
Maybe it is not a big difference in the code but the concept is that after the operation validation, you can trust the orgUnitMode that is there

muilpp: In this case we can't, because it could be the user has the authority `ALL`, meaning it's a superuser.
But this just gave me a better idea on how to handle that, I'll refactor this a bit more.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -277,59 +286,54 @@ public void decideAccess(EnrollmentQueryParams params) {
...
       throws ForbiddenException {
     List<Enrollment> enrollmentList = new ArrayList<>();
     User user = currentUserService.getCurrentUser();
 
     for (Enrollment enrollment : enrollments) {
       if (enrollment != null
-          && trackerOwnershipAccessManager.hasAccess(
-              user, enrollment.getTrackedEntity(), enrollment.getProgram())) {
-        enrollmentList.add(getEnrollment(enrollment, params, includeDeleted));
+          && (trackerOwnershipAccessManager.hasAccess(
```
enricocolasante, line 333: Just a small performance fix, if we swap the order here, putting `orgUnitMode == ALL` in the first place, if it is true, then the ownershipCheck, that here is the heavier operation, will not happen

enricocolasante (approved):
Just a small comment

zubaira (approved):

PR #15873, opened 2023-12-10
Remove duplicate Java check for data elements in datasets with different period types

jason-p-pickering:
…erent period types


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/FlattenedDataIntegrityReport.java
```diff
@@ -53,7 +53,7 @@ public class FlattenedDataIntegrityReport implements WebMessageResponse {
   @JsonProperty private final List<String> dataElementsWithoutGroups;
 
   @JsonProperty
-  private final Map<String, List<String>> dataElementsAssignedToDataSetsWithDifferentPeriodTypes;
```
jbee, line 56: You cannot change this field. If you just have one value wrap it in a list when coping over to the map.

jbee (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataintegrity/FlattenedDataIntegrityReport.java
jason-p-pickering: Reverted this back. As requested, made changes in the SQL to return the UIDs of the datasets in the `refs` field.

david-mackessy (approved):

jbee (approved):

PR #15880, opened 2023-12-11
fix: make dataDimensionType not null for catOptionGroup and groupSet

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-14694

DataDimensionType should be not-null for CategoryOptionGroup and CategoryOptionGroupSet.


jbee (approved):

vietnguyen:
> wondering if this could cause issues to entities that currently have null for this value in the DB? (or is that even possible)
> If it is possible, imagine a `GET` is performed for an entity with a null `dataDimensionType` and then that entity is flushed and might throw an error.
> Or is there a default `dataDimensionType` in play here somewhere?

@david-mackessy I thought about that. In this case I think we can't automatically update existing records because we can't be sure which is the correct value. 
So i think it's better not to back-port this fix, then add a note to the breaking change section in upgrade notes for 2.41. User would need to manually correct their data before upgrade.
Beside, this feature is requested and only used by PEPFAR so I think the fix won't affect other implementations.

david-mackessy (approved):

PR #15881, opened 2023-12-11
feat: datastore partial and rolling updates [DHIS2-16224]

jbee:
### Summary
Adds two (and a half) features to the datastore update API
* partial update of a complex JSON value providing a `path` to the part of the value that should be changed (no need to resend the entire value)
* fixed collection update of arrays providing a `roll` size. This means "atomic" insert at the end, discarding the head if the size is or exceeds the roll size
* delete values using undefined body

### Fixed Collections
Fixed collections is a term used for a list of a fixed size that allows an insert while the size is constant. Usually this is done by shifting the elements through the fixed window so that the oldest is removed in order to insert a new element.

With the existing JSONB functions it was easiest to insert via append (`||` concat operator) and remove via index (`- 0` operator). The size of the fixed collection is thereby supplied via the `roll` parameter. The presence of the roll parameter also triggers the fixed collection behaviour. 

### Edge Cases
* If the target value is not an array, null, or undefined but another JSON type this results in a simple override of the value with the new provided value. 
* If the rolled array is larger than the roll size this will **not** discard more then one of the existing elements and cut the array to the roll length. Instead the current length is maintained.
* If the provided roll size is negative the current size is simply maintained

### Automatic Testing
Moves any existing update related test to a class with a postgres setup as H2 will not support the new update via SQL.

Adds a new integration test class with one scenario for each possible update scenario. 

### Manual Testing
There are many scenarios from the combination of the `path` and `roll` parameters for a `PUT` `/api/dataStore/<ns>/<key>` endpoint in combination with the current and new value of the entry, such as:
* `PUT` `/api/dataStore/test/test?roll=2` on value `[1,2]` and new value `4`
* `PUT` `/api/dataStore/test/test?roll=2&path=a` on value `{"a": [1,2] }` and new value `4`

All cases are covered by this test https://github.com/dhis2/dhis2-core/blob/32e30e025b55a334ac3637629c08e739137abc07/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatastoreUpdateControllerIntegrationTest.java

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datastore/hibernate/HibernateDatastoreStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datastore/hibernate/HibernateDatastoreStore.java
@@ -267,13 +267,18 @@ public class HibernateDatastoreStore extends HibernateIdentifiableObjectStore<Da
         """
           update keyjsonvalue
           set jbvalue = case jsonb_typeof(jsonb_extract_path(jbvalue, VARIADIC cast(:path as text[])))
-            when 'null' then jsonb_set(jbvalue, cast(:path as text[]), to_jsonb(ARRAY[cast(:value as jsonb)]))
             when 'array' then case
               when :size < 0 or jsonb_array_length(jsonb_extract_path(jbvalue, VARIADIC cast(:path as text[]))) >= :size
                 then jsonb_set(jbvalue, cast(:path as text[]), (jsonb_extract_path(jbvalue, VARIADIC cast(:path as text[])) - 0) || to_jsonb(ARRAY[cast(:value as jsonb)]), false)
               else jsonb_set(jbvalue, cast(:path as text[]), jsonb_extract_path(jbvalue, VARIADIC cast(:path as text[])) || to_jsonb(ARRAY[cast(:value as jsonb)]), false)
               end
-            else jsonb_set(jbvalue, cast(:path as text[]), cast(:value as jsonb))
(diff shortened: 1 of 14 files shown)


netroms (approved):

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datastore/hibernate/HibernateDatastoreStore.java
```diff
@@ -172,15 +174,138 @@ public DatastoreEntry getEntry(String namespace, String key) {
...
+    // language=SQL
+    String sql = "select count(*) from keyjsonvalue v where v.namespace = :ns";
+    Object count = getSession().createNativeQuery(sql).setParameter("ns", ns).uniqueResult();
+    if (count == null) return 0;
+    if (count instanceof Number n) return n.intValue();
+    throw new IllegalStateException("Count did not return a number but: " + count);
+  }
+
+  @Override
+  public boolean updateEntry(
```
david-mackessy, line 194: this method would read more easily/nicely if each flow was its own method, making it much easier to comprehend at the top level what this method does, without having to look through the implementation details for each flow.

jbee: I'll make a cleanup PR

jbee: https://github.com/dhis2/dhis2-core/pull/15899 :)

PR #15888, opened 2023-12-11
fix: System settings endpoint returns expected Cache-Control header v…

david-mackessy:
…alue [DHIS2-15196]

# Summary
Security issue raised so that we set the following values for the `Cache-Control` header on the `/api/systemSettings` endpoint:
- no-cache
- no-store
- must-revalidate

Code added to `ContextUtils` for easy reuse.

## Note
This was the easiest and simplest way to have the combination of these 3 header values contained in 1 header key.
The Spring `CacheControl` class is not nice to work with at all when trying to get these 3 specific values together (mix of static and non-static methods)

# Testing
## Automated
- e2e test added

## Manual
- make call to `GET /api/systemSettings`
- check header `Cache-Control`
- ensure its value is `no-cache, no-store, must-revalidate`

[DHIS2-15196]: https://dhis2.atlassian.net/browse/DHIS2-15196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/ContextUtils.java
```diff
@@ -189,6 +189,12 @@ private void configureResponse(
     }
   }
 
+  public static HttpHeaders getSecureNoCacheHeader() {
```
jbee, line 192: I think using the word secure here is easily misleading the reader to think this has anything to do with security and authentication when it is just a working way of forcing the client to not cache the response. I also think we should maybe follow the spring methods in naming for this and simply call it `noCacheNoStoreMustRevalidate`.

david-mackessy: cool, i'll update

vietnguyen (approved):

jbee (approved):

PR #15904, opened 2023-12-12
fix: json patch throw error when remove null property

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-16299
### Issue

**Problem in JsonPatchManager**: 
The `JsonPatchManager` relies on the presence of properties within a `JsonNode` to validate patch paths. However, our Jackson ObjectMapper does not serialize properties with `null` values into `JsonNode`. This leads to an issue where patch paths referencing `null` properties are incorrectly deemed invalid, despite these properties existing in the original object.

### Fix

- Validate the patch path in `JsonpPatchManager` before applying the patch operators.
- Remove the path validation in `RemoveOperation` and related tests.

### Test

**Validation Procedure**:

1. **Initial Setup**: 
   Begin by creating a new UserGroup instance where the `code` property is intentionally set to `null`.

2. **Patch Request**: 
   Execute a JSON patch request to update the `code` property. Use the following `curl` command:

```
curl -X PATCH -v http://localhost:9090/api/userGroups/vAvEltyXGbD
-H "Content-Type: application/json-patch+json"
-d '[{ "op": "replace", "path": "/code", "value": "bonthe" }]'
-u admin:district
```

3. **Outcome Verification**: 
Confirm that the `code` property is correctly updated to "bonthe".


jbee (approved):

david-mackessy (approved):
lovely

PR #15929, opened 2023-12-13
feat: Add type filter in Icon endpoint [DHIS2-16044]

zubaira:
[DHIS2-16044](https://dhis2.atlassian.net/browse/DHIS2-16044)

[DHIS2-16044]: https://dhis2.atlassian.net/browse/DHIS2-16044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconCriteria.java
enricocolasante, line 1: You are creating the same class with a different behavior in a different PR.
This is going to create a big conflict.
As you are the author of both PR it would be good to think of an order of PRs that needs to be merged, or maybe close the other PR for now.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -137,6 +140,25 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
     return icons.stream().map(iconMapper::from).toList();
   }
 
+  @GetMapping(value = "/type/{type}")
```
enricocolasante, line 143: I am just curious about the API design here, why type is not a RequestParam?
Default value of type param could be both so in case is not specified it will have the same behavior of `getAllIcons`, that could be removed then

enricocolasante (commented):
This PR is in conflict with the pagination PR, which one is the "right" one?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconCriteria.java
zubaira: We have not yet decided about the fate of pagination. So i am not sure when pagination stuff is going to be merged.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: right. Initially, i created `api/icons?type` and that was conflicting with `api/icons?iconkey`. So then i had to introduce another path `api/icons/type`. but yes `type` can be RequestParam.

enricocolasante: What do you mean they were conflicting?

In general, it would be good to have one flow for retrieving and have all the parameters in the criteria.
Here we already have 3 different calls to the service (getIcons(), getIcons(keywords) and getIcons(criteria)) that are doing the same thing but filtering on different fields.
We could have just one getIcons(criteria) with all the params and the flow would be much simpler

zubaira: Ambiguous mapping `IconController#getIconByType(String, String[])` and `IconController#getAllIcons(String[])`

zubaira: for multiple getIcon methods. I will check if they all can be replaced by one.

zubaira: having everything in one getIcon(criteria) would make it very messy. It looks abit redundant now, but i think
in the current shape they are abit simpler.
Custom types are coming from database where as defaults are bundled in the war file. In some cases we need all of them but in others, we need all but with keyword filtration. So there are many scenarios that a single method needs to cover and that would make one method unnecessarily complex.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -72,20 +72,23 @@ public class DefaultIconService implements IconService {
 
   @Override
   @Transactional(readOnly = true)
-  public List<Icon> getIcons() {
-    return Stream.concat(defaultIcons.values().stream(), customIconStore.getAllIcons().stream())
-        .toList();
-  }
+  public List<? extends Icon> getIcons(IconCriteria iconCriteria) {
+    if (IconType.DEFAULT.equals(iconCriteria.getType())) {
```
enricocolasante, line 76:
```suggestion
    if (IconType.DEFAULT == iconCriteria.getType()) {
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -72,20 +72,23 @@ public class DefaultIconService implements IconService {
...
+      return defaultIcons.values().stream()
+          .filter(icon -> Set.of(icon.getKeywords()).containsAll(iconCriteria.getKeywords()))
+          .toList();
+    }
+
+    if (IconType.CUSTOM.equals(iconCriteria.getType())) {
```
enricocolasante, line 82:
```suggestion
    if (IconType.CUSTOM == iconCriteria.getType()) {
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
```diff
@@ -77,7 +77,12 @@ public CustomIcon getIconByKey(String key) {
   }
 
   @Override
-  public List<CustomIcon> getIconsByKeywords(String[] keywords) {
+  public List<CustomIcon> getIconsByKeywords(List<String> keywords) {
+
+    if (keywords.isEmpty()) {
```
enricocolasante, line 82: I would rather completely remove the `getAllIcons()` method and add `where keywords @> string_to_array(?,',')` to the sql only if keywords is not empty.

This way if something changes in the query you do not have 2 places to change

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -87,12 +88,16 @@ protected void initTest() throws IOException {
   void shouldGetAllIconsWhenRequested() {
     Map<String, DefaultIcon> defaultIconMap = getAllDefaultIcons();
 
+    IconCriteria iconCriteria = new IconCriteria();
```
enricocolasante, line 91: You could also add one test similar to this one just using the default values from `iconCriteria`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -124,15 +124,9 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
-    List<Icon> icons;
-
-    if (keywords == null) {
-      icons = iconService.getIcons();
-    } else {
-      icons = iconService.getIcons(keywords);
-    }
+  public @ResponseBody List<IconResponse> getAllIcons(IconCriteria iconCriteria) {
+
+    List<? extends Icon> icons = iconService.getIcons(iconCriteria);
```
enricocolasante, line 129:
```suggestion
    List<Icon> icons = iconService.getIcons(iconCriteria);
```

or even better inline it in the return statement

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
zubaira: done

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
zubaira: done

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconCriteria.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.NoArgsConstructor;
+
+/** Represents query parameters sent to {@link IconController}. */
+
+/**
+ * @author Zubair Asghar
+ */
+@Data
+@NoArgsConstructor
+public class IconCriteria {
```
teleivo, line 42: Is this `Criteria` suffix a convention in platform? It always reminds me of hibernate. In tracker we use `RequestParams` in the controller layer and `QueryParams` in the store layer. These are easier at least for me to mentally attribute to one of the layers.

Or is it that this is used in all layers?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconType.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.hisp.dhis.icon;
+
+import java.util.Optional;
+
+/**
+ * @author Zubair Asghar
+ */
+public enum IconType {
+  ALL("all"),
```
teleivo, line 36: What is the meaning of this IconType class? As all is AFAIK not an icon type we actually have.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -124,17 +124,9 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
-  public @ResponseBody List<IconResponse> getAllIcons(
-      @RequestParam(required = false) String[] keywords) {
-    List<Icon> icons;
-
-    if (keywords == null) {
-      icons = iconService.getIcons();
-    } else {
-      icons = iconService.getIcons(keywords);
-    }
+  public @ResponseBody List<IconResponse> getAllIcons(IconCriteria iconCriteria) {
```
teleivo, line 127: Would `getIcons` make more sense as a method name? As it does not necessarily return all icons if keywords are given. Or does all mean something else here 😅

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -87,12 +88,30 @@ protected void initTest() throws IOException {
...
+    iconCriteria.setType(IconType.ALL);
+    iconCriteria.setKeywords(List.of());
```
teleivo, line 93: keywords is an empty list by default AFAIK

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -87,12 +88,30 @@ protected void initTest() throws IOException {
...
+
+    assertEquals(
+        defaultIconMap.size() + 1,
+        iconService.getIcons(iconCriteria).size(),
+        String.format(
+            "Expected to find %d icons, but found %d instead",
+            defaultIconMap.size() + 1, iconService.getIcons(iconCriteria).size()));
```
teleivo, line 100: Since this is an integration test it would be hard to debug on github if that assertion fails. I would print the list in addition, maybe like

```suggestion
        String.format(
            "Expected to find %d icons, but found %d instead: %s",
            defaultIconMap.size() + 1, iconService.getIcons(iconCriteria).size(), iconService.getIcons(iconCriteria)));
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -87,12 +88,30 @@ protected void initTest() throws IOException {
...
+  }
+
+  @Test
+  void shouldGetAllIcons() {
```
teleivo, line 104: What is the difference between this and the above test? Should this show that `shouldGetAllIconsByDefault`? if so I would rename the test. I also wonder if we need an integration test for this as this only tests IconCriteria.type defaults to ALL

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: in cases when there are no keywords and IconType.ALL then it will return all icons.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconCriteria.java
zubaira: Agreed, uniform convention makes it easier. But this is not tracker specific thing and we have criteria objects in many other modules.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconType.java
zubaira: make sense. Renaming it to `IconTypeFilter` or something like that

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
zubaira: fixed.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -40,17 +40,10 @@ public interface IconService {
...
-   * Gets icons tagged with all given keywords.
-   *
-   * @param keywords collection of keywords
-   * @return a collection of matching icons
-   */
-  List<Icon> getIcons(String[] keywords);
+  List<? extends Icon> getIcons(IconCriteria iconCriteria);
 
   /**
    * Gets the icon associated to a key, if it exists
```
lucaCambi77, line 49: I don't know all of it, but if it implements an interface, it should return the `Icon` in the first place to avoid IDE warnings and make the code cleaner.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
```diff
@@ -65,6 +67,8 @@ void shouldCreateCustomIconWhenFileResourceExist() throws IOException {
                     + createFileResource()
                     + "', 'keywords':"
                     + keywords
+                    + ", 'custom' :"
+                    + custom
                     + "}")
             .content(HttpStatus.CREATED)
```
lucaCambi77: You should be able to use a text block and create a more readable `key` with something like
```
    String key =
        """
{'key': '%s','description': '%s','fileResourceUid': '%s','keywords': %s}"""
            .formatted(iconKey, description, createFileResource(), keywords);
    JsonWebMessage message =
        POST("/icons/", key).content(HttpStatus.CREATED).as(JsonWebMessage.class);
```

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
```diff
@@ -126,6 +121,19 @@ void shouldDeleteIconWhenKeyExists() throws IOException {
...
+    String fileResourceId = createFileResource();
+    createIcon(fileResourceId);
+
+    JsonArray iconArray = GET("/icons/type/custom").content(HttpStatus.OK);
+
+    assertEquals(iconKey, iconArray.getObject(0).getString("key").string());
+    assertEquals(description, iconArray.getObject(0).getString("description").string());
+    assertEquals(fileResourceId, iconArray.getObject(0).getString("fileResourceUid").string());
+    assertEquals(keywords, iconArray.getObject(0).getArray("keywords").toString());
+  }
```
lucaCambi77: It would be nice to wrap this kind of assertion in `assertAll()`. Also, with an error message in case it fails

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
zubaira: Yes. I did try that and surprising for me that IDE started complaining. So I then had to use this

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
zubaira: fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
lucaCambi77: I think the best way is to change the signature in the interface:

`
List<Icon> getIcons(IconCriteria iconCriteria);
`

then in the `getIcons()` method in `DefaultIconService` you can return:

`
return new ArrayList<>(customIconStore.getIconsByKeywords(iconCriteria.getKeywords()));
`

lucaCambi77: You should then be able to use `switch` :

```
  @Override
  public List<Icon> getIcons(IconCriteria iconCriteria) {
    switch (iconCriteria.getType()) {
      case DEFAULT -> {
        return defaultIcons.values().stream()
            .filter(icon -> Set.of(icon.getKeywords()).containsAll(iconCriteria.getKeywords()))
            .toList();
      }
      case CUSTOM -> {
        return new ArrayList<>(customIconStore.getIconsByKeywords(iconCriteria.getKeywords()));
      }
      default -> {
        return Stream.concat(
                defaultIcons.values().stream()
                    .filter(
                        icon -> Set.of(icon.getKeywords()).containsAll(iconCriteria.getKeywords()))
                    .toList()
                    .stream(),
                customIconStore.getIconsByKeywords(iconCriteria.getKeywords()).stream())
            .toList();
      }
    }
  }
```

zubaira: That's why i have mentioned it [here](https://github.com/dhis2/dhis2-core/pull/15929#discussion_r1450861309.).
So either we have to fix the code or the sonarcloud configurations.

Screenshot 2024-01-15 at 09 17 22

lucaCambi77: ok but that is about mutability and should be fixable, I'll have a look

lucaCambi77:
```
  private final Map<String, Icon> defaultIcons =
      Arrays.stream(DefaultIcon.Icons.values())
          .map(DefaultIcon.Icons::getVariants)
          .flatMap(Collection::stream)
          .collect(Collectors.toMap(DefaultIcon::getKey, Function.identity()));

  @Override
  public List<Icon> getIcons(IconCriteria iconCriteria) {
    switch (iconCriteria.getType()) {
      case DEFAULT -> {
        return defaultIcons.values().stream()
            .filter(icon -> Set.ofli(icon.getKeywords()).containsAll(iconCriteria.getKeywords()))
            .toList();
      }
      case CUSTOM -> {
        return new ArrayList<>(customIconStore.getIconsByKeywords(iconCriteria.getKeywords()));
      }
      default -> {
        return Stream.concat(
                defaultIcons.values().stream()
                    .filter(
                        icon -> Set.of(icon.getKeywords()).containsAll(iconCriteria.getKeywords()))
                    .toList()
                    .stream(),
                customIconStore.getIconsByKeywords(iconCriteria.getKeywords()).stream())
            .toList();
      }
    }
  }
```
This should work, but you must roll back the `stream` change and make the `getIconsByKeywords` return a List<CustomIcon>. The toList() enforced in the latest version of Java returns the implementation rather than the abstraction, so it is a bit unfortunate.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
teleivo: Right, so the method is a more general one hence `getAllIcons` is misleading as that is just one of the use cases it supports.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconType.java
teleivo: `IconTypeFilter` makes sense to me 👍🏻

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParamsMapper.java
```diff
@@ -0,0 +1,75 @@
...
+ * @author Zubair Asghar
+ */
+@Component
+@RequiredArgsConstructor
+public class IconRequestParamsMapper {
+
+  public IconOperationParams map(IconRequestParams iconRequestParams) throws BadRequestException {
+
+    validate(iconRequestParams);
+    IconOperationParams operationParams = new IconOperationParams();
```
teleivo, line 47: We are trying to get away from these old style of classes on which will call setters after instantiating them. If a class has many optional fields we use a builder. The goal is that an instance of an object is valid once created 😄 Either we achieve that using a good old constructor/factory or a builder.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
```diff
@@ -107,30 +104,44 @@ void shouldUpdateIconWhenKeyExists() throws IOException {
...
-    createIcon(createFileResource());
+    createIcon(createFileResource(), keywords);
 
     JsonObject response = DELETE(String.format("/icons/%s", iconKey)).content();
 
     assertEquals(String.format("Icon %s deleted", iconKey), response.getString("message").string());
   }
 
-  private String createIcon(String fileResourceId) {
+  private String createIcon(String fileResourceId, String keywords) {
```
teleivo, line 136: Do I understand correctly that a user can only create a custom icon? If so I would name this `createCustomIcon` to make this obvious. This makes it easier to understand some of the test cases 😅

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
```diff
@@ -107,30 +104,44 @@ void shouldUpdateIconWhenKeyExists() throws IOException {
...
+
+    verifyAll(iconArray, keywords, fileResourceId);
+  }
+
+  @Test
+  void shouldGetIconsWithSpecifiedKey() throws IOException {
+    String fileResourceId = createFileResource();
+    createIcon(fileResourceId, keyword);
+
+    JsonArray iconArray = GET("/icons?keywords=m1").content(HttpStatus.OK);
```
teleivo, line 122: You can have a look at EnrollmentsExportControllerTest on how we use jsontree 😄 

I think you should be able to do

```suggestion
    JsonList<JsonIcon> icons = GET("/icons?keywords=m1")
       .content(HttpStatus.OK)
       .asList(JsonIcon.class);
```

or `.getList("someKey, JsonIcon.class);` if the icons returned in a map.


All our custom assertions like this are usually prefixed with `assert`. With the above change you can then do

`assertCustomIcon(icons.get(0), ...)`

this makes your assertion reusable as it does not include that you expect the icon to be in position 0 in a list/array.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconOperationParams.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package org.hisp.dhis.icon;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** Represents query parameters sent to {@link org.hisp.dhis.icon.IconService}. */
```
teleivo, line 36: please merge the 2 separate javadocs

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -88,12 +88,34 @@ protected void initTest() throws IOException {
...
     Map<String, DefaultIcon> defaultIconMap = getAllDefaultIcons();
 
+    IconOperationParams operationParams = new IconOperationParams();
+    operationParams.setIconTypeFilter(IconTypeFilter.ALL);
+    operationParams.setKeywords(List.of());
+
     assertEquals(
         defaultIconMap.size() + 1,
-        iconService.getIcons().size(),
+        iconService.getIcons(operationParams).size(),
```
teleivo, line 97: Tests to me are easier to read if they follow the same structure. I would like to see with a glimpse whats setup, whats being tested/called, whats the expected outcome. Having three groups for each achieves that for example.

I would at least extract the `iconService.getIcons(operationParams)` which is whats tested AFAIK

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
```diff
@@ -137,4 +132,35 @@ public void update(CustomIcon customIcon) {
         customIcon.getLastUpdated(),
         customIcon.getKey());
   }
+
+  private String buildIconQuery(IconOperationParams iconOperationParams, String sql) {
+    SqlHelper hlp = new SqlHelper(true);
+
+    if (iconOperationParams.hasLastUpdated()) {
```
teleivo, line 139: @stian-sandvold or @enricocolasante did we want to move away from this style of generating SQL? Rather use MapSqlParameterSource and NamedParameterJdbcTemplate to have some SQL injection prevention.

zubaira: SQL injection risk is there if SQL contains input from the user. That is why validation is added here to make sure
user-entered input are valid.
I can change it to include parameterized input.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParamsMapper.java
zubaira: I normally use Builder where there is a need to create complex object but lot of fields can be optional. 
IconOperationParams does not entail that complexity.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/jdbc/JdbcCustomIconStore.java
teleivo: Keywords are also user input. I am not sure there is no way to exploit this. My point is better safe than sorry by replacing an error prone practice with a safer one.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParams.java
```diff
@@ -0,0 +1,63 @@
...
+@Data
+@NoArgsConstructor
+public class IconRequestParams {
+
+  static final String DEFAULT_FIELDS_PARAM =
+      "key,keywords,description,fileResourceUid,createdByUserUid,href";
+
+  private List<String> keywords = new ArrayList<>();
+  private IconTypeFilter type = IconTypeFilter.ALL;
+  private Date created;
```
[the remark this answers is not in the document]
stian-sandvold: Platform standard: filter=...
If we for some reason cannot piggyback on the platform filter, at least use clearer names for the params, like createdAfter/lastUpdatedAfter.

Can icons be updated? Yes, keywords and description, but changing the icon or the key/name should not be :)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParams.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @author Zubair Asghar
+ */
+@Data
+@NoArgsConstructor
+public class IconRequestParams {
+
+  static final String DEFAULT_FIELDS_PARAM =
+      "key,keywords,description,fileResourceUid,createdByUserUid,href";
```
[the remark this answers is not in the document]
stian-sandvold: Depends whether default should be minimal or not;

The bare minimum to show would be key and fileResourceUid. This would be used by all consumers, since it's the minimum required to link an icon reference (key) to a file (fileResourceUid).

However, keywords and descriptions are useful by adding more context, so if you're browsing the api, this makes sense to include; But for our internal apps, it's primarily relevant for maintenance app (Unless we want ALT text for images). And if we starting thinking we add anything relevant for maintenance, I would say this is a good default.

This I feel would be more useful to get feedback from the frontend teams instead (@vgarciabnz @Birkbjo @JoakimSM @janhenrikoverland ) But if there is no preference, I'd be fine with what is in this list :)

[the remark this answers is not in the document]
vgarciabnz: It looks fine to me. Anyway, the Android app always explicitly includes the list of fields in the request to avoid downloading unused properties.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -113,12 +135,16 @@ void shouldGetAllKeywordsWhenRequested() {
             .flatMap(Arrays::stream)
             .collect(Collectors.toSet());
 
+    IconOperationParams operationParams = new IconOperationParams();
+    operationParams.setIconTypeFilter(IconTypeFilter.ALL);
+    operationParams.setKeywords(keywordList.stream().toList());
```
teleivo, line 140: keywords is an empty list by default AFAIK. I don't think this is necessary

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -88,12 +88,34 @@ protected void initTest() throws IOException {
   void shouldGetAllIconsWhenRequested() {
     Map<String, DefaultIcon> defaultIconMap = getAllDefaultIcons();
 
+    IconOperationParams operationParams = new IconOperationParams();
+    operationParams.setIconTypeFilter(IconTypeFilter.ALL);
+    operationParams.setKeywords(List.of());
```
teleivo, line 93: keywords is an empty list by default AFAIK. I don't think this is necessary

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -124,17 +130,15 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
-  public @ResponseBody List<IconResponse> getAllIcons(
-      @RequestParam(required = false) String[] keywords) {
-    List<Icon> icons;
-
-    if (keywords == null) {
-      icons = iconService.getIcons();
-    } else {
-      icons = iconService.getIcons(keywords);
-    }
+  public @ResponseBody List<ObjectNode> getAllIcons(IconRequestParams iconRequestParams)
```
teleivo, line 133: github seems to have lost our conversation on this one. https://github.com/dhis2/dhis2-core/pull/15929#discussion_r1450022903 still applies

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParams.java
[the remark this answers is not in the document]
teleivo: > Depends whether default should be minimal or not;

I think the default should be a sensible minimum 😁 like we don't return nested entities in tracker export by default to avoid fetching everything just because someone is browsing the API 😁 

Any app should use the approach @vgarciabnz in Android takes. Always specify the fields you need.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
zubaira: fixed

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
zubaira: fixed

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/icon/IconControllerTest.java
zubaira: fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconOperationParams.java
zubaira: fixed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParamsMapper.java
```diff
@@ -0,0 +1,99 @@
...
+
+    return operationParams;
+  }
+
+  private void validateRequestParams(IconRequestParams iconRequestParams)
+      throws BadRequestException {
+
+    if (iconRequestParams.hasCreatedStartDate()
+        && !DateUtils.dateIsValid(
+            DateUtils.getMediumDateString(iconRequestParams.getCreatedStartDate()))) {
```
teleivo, line 66: We should support timestamps right @stian-sandvold? Looks like the `getMediumDateString` uses format `yyyy-MM-dd`

zubaira: yes @teleivo.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconOperationParams.java
```diff
@@ -0,0 +1,71 @@
...
+import lombok.NoArgsConstructor;
+
+/**
+ * Represents query parameters sent to {@link org.hisp.dhis.icon.IconService}.
+ *
+ * @author Zubair Asghar
+ */
+@Data
+@NoArgsConstructor
+public class IconOperationParams {
```
vgarciabnz, line 43: I have just realized it is not possible to filter a list of "key", is it? 
In Android, the workflow would be:
1. Download user-related metadata (programs assigned, data elements, etc)
2. Collect the list of icons assigned to that metadata and filter those whose icon has not been downloaded.
3. Request the "/api/icons" to download the information about those icons
4. Download the file resources

In the step 3, we could do a single request if it were possible to filter by key (something similar to `filter=id:in:id1,id2,id3,..` in metadata). I assume we need the JsonIcon object in order to be able to download the actual file resource. Or is it possible to download the file resource without downloading the JsonIcon object?

zubaira: To download the file you need to hit this endpoint` /{key}/icon`. `{key}` is the fileResourceUid that you get after downloading CustomIcon.

vgarciabnz: In that case, I would suggest to add a filter such as `key=icon1,icon2,icon3,...` in order to download a list of CustomIcons in a single query instead of doing a request per icon. It doesn't necessarily have to be part of this PR

zubaira: yes this is possible using  `api/icons?keys=icon1,icon2,icon3`

vgarciabnz (approved):

PR #15942, opened 2023-12-14
fix: data value import with completion - use today for any future date [DHIS2-16290]

jbee:
### Summary
Jason P. was observing users posting future dates for the `completeDate` property of the data set import. 
Presumably users would do so to "game the system" in a way.
This fix prevents this by replacing any future date with the current date.
In addition it will also allow to use `true` (completed today) and `false` (not completed) as values.

### Automatic Testing
New test was added to check the date parsing and defaults work as expected.

### Manual Testing
* perform a data value set import with a future date
* check the completion record is updated not to the future date but to the date of the import
* do the same with a past date
* check the past date is accepted as is


david-mackessy (approved):
Nice. I think [the docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/data.html?h=complete#webapi_sending_data_values) might need an update so it's clear that we're using the execution time (for supplied future dates), as it's a behaviour change that some people might not be expecting

vietnguyen (approved):

PR #15955, opened 2023-12-15
fix: sort authorities list case insensitive [DHIS2-10843]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-10843

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AuthoritiesControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AuthoritiesControllerTest.java
@@ -27,9 +27,11 @@
  */
 package org.hisp.dhis.webapi.controller;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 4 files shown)


netroms (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AuthoritiesControllerTest.java
```diff
@@ -46,9 +45,22 @@ class AuthoritiesControllerTest extends DhisControllerConvenienceTest {
...
+                            .getString("name")
+                            .string()
+                            .toLowerCase()
+                            .compareTo(
+                                systemAuthorities
+                                    .getObject(i)
+                                    .getString("name")
+                                    .string()
+                                    .toLowerCase())
+                        < 0));
```
jbee, line 64: If I get this correct you want to make sure they are all different when lower-cased and maybe also sorted?
I think the better way to do this is to use `stringValues()` on the JsonArray, giving you a `List<String>`. 
You then can map `toLowerCase`. Then sort a copy and compare that for equals, also copy it in a set and compare that for size to make sure there are no duplicates.

jbee (approved):

PR #15956, opened 2023-12-15
fix: Ensure task summary available on empty file import [DHIS2-7881]

david-mackessy:
# Summary 
When importing an empty json file asynchronously in the data value set API, the job completes but there is no task summary produced which causes the UI to crash when retrieving an empty task summary body.

## Fix
Add import summaries in catch blocks

## Notes
All formats were tested (not just the reported json issue) as good housekeeping.
The fix in `DefaultDataValueSetService` fixes the same issue for formats:
- json
- xml
- pdf

The ADX format uses a different service `DefaultAdxDataService` which also had the same issue, so the try block was moved up slightly to include the thrown exception origin.

All formats now display an import summary when an empty file is imported async

# Testing
## Automated
2 e2e tests added covering new behaviour

## Manual
To manual test follow steps in [Jira](https://dhis2.atlassian.net/browse/DHIS2-7881)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -52,7 +52,7 @@ jobs:
           cache: maven
 
       - name: Codeql-init
-        uses: github/codeql-action/init@v2
+        uses: github/codeql-action/init@v3
         with:
           languages: java
           config-file: ./.github/codeql/codeql-config.yml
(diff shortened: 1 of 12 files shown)


jbee (approved):
We should move the execution to use the job scheduler not a `ExecutorService` but that is another ticket to be created :D

vietnguyen (approved):

PR #15979, opened 2023-12-18
fix: Return accessible enrollments instead of exception [TECH-1589]

muilpp:
In `/tracker/enrollments` we return an exception if either of the enrollment's program or tracked entity type are not accessible. We just want to raise an exception if the parameters are explicitly in the request, if not, in requests like `/tracker/enrollments?orgUnitMode=ALL`, we want to return all user accessible results.  

Ticket: https://dhis2.atlassian.net/browse/TECH-1589


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapper.java
```diff
@@ -178,6 +179,13 @@ private TrackedEntity validateTrackedEntity(String uid) throws BadRequestExcepti
       throw new BadRequestException("Tracked entity is specified but does not exist: " + uid);
     }
 
+    if (trackedEntity.getTrackedEntityType() != null
+        && !aclService.canDataRead(user, trackedEntity.getTrackedEntityType())) {
+      throw new BadRequestException(
```
muilpp, line 184: is there any reason why we should not do this?

enricocolasante (commented):
Return 403 code for sharing validation

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapperTest.java
```diff
@@ -186,7 +185,7 @@ void shouldMapOrgUnitsWhenOrgUnitUidsAreSpecified()
   }
 
   @Test
-  void shouldThrowExceptionWhenOrgUnitNotFound() {
+  void shouldFailWhenOrgUnitNotFound() {
     EnrollmentOperationParams operationParams =
         EnrollmentOperationParams.builder()
```
lucaCambi77, line 190: I think ... shouldFail ... is a bit generic here, it could be more like .. forbiddenAccesWhen ... or ... shouldThrowForbiddenException ...

muilpp: I don't really have a strong opinion on this, I just did it because in other tests we already had "shouldFail", so I wanted to harmonize this. But I'm fine with having a more explicit test name. I'll change it back.

enricocolasante (approved):

lucaCambi77 (approved):

PR #15988, opened 2023-12-19
fix: Fix issue related to updating modifiable properties of a category combo

jason-p-pickering:
- Users should be able to modify the name, short name, and totals properties of a category combo. However, they should not be able to modify the categories of a category combo if data exists in the system, as this would make the data "unreachable". 
- This PR corrects how the old and proposed objects are compared by using only the set of categories UIDs as the basis of comparison.
- Added an API test to test the modification of the category combo when it has data, and when it does not have data.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
@@ -32,7 +32,6 @@ import static org.hisp.dhis.web.WebClient.Body;
 import static org.hisp.dhis.web.WebClientUtils.assertStatus;
 
 import org.hisp.dhis.common.ValueType;
-import org.hisp.dhis.datavalue.DataValueService;
 import org.hisp.dhis.jsontree.JsonList;
 import org.hisp.dhis.jsontree.JsonObject;
 import org.hisp.dhis.web.HttpStatus;
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.webapi.controller;
 
-import static java.lang.String.format;
 import static org.hisp.dhis.web.WebClient.Body;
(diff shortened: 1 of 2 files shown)


jbee (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  String dataElementId;
+
+  String orgUnitId;
+  String categoryColor;
+  String categoryTaste;
+
+  @Test
+  void testModificationNoData() {
+    setupTest();
```
jbee, line 55: If all your test methods share a common setup you should annotate the `setupTest` method with `@BeforeEach` instead of calling it implicitly here.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    String body =
+        format(
+            "{"
+                + "'dataElement':'%s',"
+                + "'categoryOptionCombo':'%s',"
+                + "'period':'20220102',"
+                + "'orgUnit':'%s',"
+                + "'value':'24',"
+                + "'comment':'OK'}",
```
jbee, line 90: I recommend using a text-block instead, try this:
```java
// language=JSON
String body = """
...
""".format(...)
``

File: dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/CategoryComboModificationControllerTest.java [hunk 3]
```diff
@@ -0,0 +1,216 @@
...
+            POST("/categoryOptions", "{ 'name': 'Sour', 'shortName': 'Sour' }"));
+
+    String categoryOptionRed =
+        assertStatus(
+            HttpStatus.CREATED, POST("/categoryOptions", "{ 'name': 'Red', 'shortName': 'Red' }"));
+
+    categoryColor =
+        assertStatus(
+            HttpStatus.CREATED,
+            POST(
```
jbee, line 170: I think you should add a helper method for creating categories with essentially this part where you only pass in the body. This will make this much more readable as you don't get the breakdown like this. Also you can use text-blocks again for the body ;)

david-mackessy (approved):

PR #15994, opened 2023-12-19
task: Support the zip/gzip format in tracker exporter endpoints [ DHIS2-16165 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16165

Apply the same logic as in the old tracker events endpoint we can find in [AbstractRootNodeMessageConverter](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/AbstractRootNodeMessageConverter.java).

Two new GET mapping are created in `tracker/events` applying to:
- application/json+gzip (.../**events.json.gz**)
- application/json+zip (.../**events.json.zip**)

Also, we harmonize the CSV compression for `EventsExportController` and `TrackedEntititesExportController` with the same logic.
Using the same logic as the old tracker, the Rest invocation will accept the `attachment` parameter, defaulting to `event.json.zip` or `event.json.gzip` if missing.

This also introduces the `EventExportTestConfiguration`, which can be used to mock configuration beans. This way, we don't need to invoke the service, also because the event query uses Postgres functions and doesn't work with the h2 database. Overall, the MVC tests should test only the web layer and not the services. We can still invoke the real method when applicable or fake a response if we want to.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventExportTestConfiguration.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventExportTestConfiguration.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2004-2023, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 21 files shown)


enricocolasante (commented):
It looks good to me!
But I think we also need to support the same compression for csv format.
You can see in the old tracker that this is possible and in the Import/Export app it is supported to export compressed events in csv format

lucaCambi77:
> It looks good to me! But I think we also need to support the same compression for csv format. You can see in the old tracker that this is possible and in the Import/Export app it is supported to export compressed events in csv format

Ok, I've missed that. Actually, there is already the gzip support for csv. We can start from that and complete the job then

vietnguyen (approved):

enricocolasante (approved):

PR #16049, opened 2024-01-02
feat: Download is missing support for NAME, CODE, UID and ID

d-bernat:
Introduce support for the outputIdScheme parameter. Depending on the selected schema, all dimensions/headers should adhere to the specified scheme. The default scheme is 'UID.' The OutputIdScheme serves as an enumerator (with options such as ID, UID, UUID, CODE, NAME). Each identifiable object can be recognized by any of the listed values.

The request parameter outputIdScheme dictates the type of ID in the response. If outputIdScheme is not included in the incoming request, UID is applied by default. All IDs, except UID, are enhancements of the SQL result set (analytics table) performed using Hibernate ORM. Consequently, some columns (_name) from the analytics tables have become obsolete and were removed from both analytics exports and SQL select queries.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/outlier/service/AnalyticsOutlierService.java
```diff
@@ -283,4 +303,18 @@ private void setRows(Grid grid, List<Outlier> outliers, OutlierRequest request)
           grid.addValue(v.getUpperBound());
         });
   }
+
+  private String getIdProperty(IdentifiableObject object, String uid, IdScheme idScheme) {
```
maikelarabori, line 307: Javadoc is missing.

maikelarabori (commented):
Hi @d-bernat, looks good. Just one minor comment related to the code. Please review it.

Also, please, add a description to the PR. Usually we describe at least what has changed and the motivation for it.

Cheers,
Maikel

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/outlier/service/AnalyticsOutlierService.java
d-bernat: done

maikelarabori (approved):

enricocolasante (approved):

PR #16059, opened 2024-01-03
fix: Expensive subquery part of events API response [DHIS2-16398]

larshelge:
The SQL query used to retrieve as part of the `/api/events` endpoint response is unnecessarily slow. The expensive part is a subquery which aggregates across _all_ category option combos. On databases with many category option combos, this is very expensive. This applies to both `JdbcEventStore` classes.

This subquery is optimized with a `lateral` join and a `where` clause to limit to the relevant attribute/category option combos.

Existing tests should cover this change.


enricocolasante (approved):
Nice catch!
Just to comment that changing both classes will improve performance in deprecated `api/events` endpoint and also in `api/tracker/events`

teleivo (approved):

PR #16073, opened 2024-01-04
feat: marker annotations to indicate non standard tx management [DHIS2-16402]

jbee:
Introduces two marker annotations that are used instead of spring's `@Transactional` to document that although the annotated methods being service level methods they should not be annotated with `@Transactional` because ...
* `NonTransactional`: ... the method is not doing database access (at all)
* `IndirectTransactional`: ... the method uses other means to access the database, e.g. TX templates or JDBC level


teleivo (approved):

lucaCambi77 (commented):
Should this be added to wow and enforced as a practice? Also, should this have an Open API tag?

enricocolasante (approved):

jbee:
> Should this be added to wow and enforced as a practice?

Yes, I'll add something


>  Also, should this have an Open API tag?

In what way? Maybe you can explain the idea you had?

PR #16077, opened 2024-01-04
feat: adds totalPages and paging request parameter alias [DHIS2-16401]

jbee:
### Summary
To align our different APIs more to use same names for common request parameters a `totalPages` alias for `total` was added. Similarly the new request parameter `paging` is the opposite of `headless`. 

When both request parameters for the same property are supplied but they do not agree a BAD REQUEST is thrown.
This follows how tracker already handles similar situations.

### Automatic Testing
New test scenarios were added to show the effectiveness of the added parameters as well as the scenario where the aliases disagree.


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/GistPagerControllerTest.java
```diff
@@ -42,9 +43,12 @@
...
   void testPager_Total_ResultBased() {
-    JsonObject gist =
-        GET("/users/{uid}/userGroups/gist?fields=name,users&total=true", getSuperuserUid())
-            .content();
+    String baseUrl = "/users/{uid}/userGroups/gist?fields=name,users";
+    JsonObject gist = GET(baseUrl + "&total=true", getSuperuserUid()).content();
+    assertHasPager(gist, 1, 50, 1);
+    gist = GET(baseUrl + "&totalPages=true", getSuperuserUid()).content();
+    assertHasPager(gist, 1, 50, 1);
+    gist = GET(baseUrl + "&totalPages=true&total=false", getSuperuserUid()).content();
```
teleivo, line 51: In tracker we opted to reject contradictory parameters as its not clear what the user wants 🤔

jbee: :+1: will follow that

teleivo (approved):

vietnguyen (approved):

PR #16086, opened 2024-01-05
feat: Add sorting into Visualization [DHIS2-16369]

maikelarabori:
Adding a `Sorting` object into `Visualization`. This will allow users to save the sorting direction they want for a particular `Visualization`, in the Data Visualizer app.
The sorting direction has to be consistent with the dimensions (columns) defined in the `Visualization` object.
Only dimension columns can be sorted.
Introduces a new DB column in the `visualization` table.


gnespolino (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/visualization/Visualization.java
```diff
@@ -807,6 +826,22 @@ private List<DimensionalItemObject> getDimensionalItemObjects(String dimension)
...
+        s -> {
+          if (isBlank(s.getDimension()) || s.getDirection() == null) {
+            throw new IllegalArgumentException("Sorting is not valid");
+          } else if (columns.stream()
+              .noneMatch(c -> containsAny(s.getDimension(), c.split("\\.")))) {
+            throw new IllegalStateException(s.getDimension());
+          }
+        });
+  }
+
```
vietnguyen, line 844: Would be great to put this into a ValidationUtil class or Service class.

vietnguyen (approved):

PR #16096, opened 2024-01-08
tech: renames audit to changeLog [ TECH-1659 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/TECH-1659

This PR is a first step to move some parts of the audit to changeLog. 
We want to split parts of the audit that relate to a change log and make the difference from what is intended to be an actual [audit](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/audit/Audit.java).
In this PR, we move:

- `TrackedEntityDataValueAudit` to `TrackedEntityDataValueChangeLog`
- `TrackedEntityAttributeValueAudit` to `TrackedEntityAttributeValueChangeLog`
- `TrackedEntityAudit` to `TrackedEntityChangeLog`

We also rename related services/stores and variable names.

### Later steps:

see https://dhis2.atlassian.net/browse/DHIS2-16126

The endpoint should move from `/audits` to `/changeLog`. Therefore, `/audits` deprecation will be in place.
Move the `changelog` tracker domain-related endpoints to dedicated tracker endpoints.
Rename the tables with the `audit` suffix to the `changelog` suffix.

## Note

Notice that `AuditType` moves to `ChangeLogType`.  When the database table and columns are renamed, the old endpoint must map `changelogtype` to `audittype` (if we keep this data structure).


dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/ChangeLogType.java
```diff
@@ -25,14 +25,14 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package org.hisp.dhis.common;
+package org.hisp.dhis.changelog;
 
 /**
  * This class is deprecated in favor of new async auditing solution, do not use.
```
teleivo, line 31: Should this doc change?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/ChangeLogType.java
```diff
@@ -25,14 +25,14 @@
...
  *
  * @author Halvdan Hoem Grelland
  */
-public enum AuditType {
+public enum ChangeLogType {
   CREATE,
   UPDATE,
   DELETE,
```
teleivo, line 38: Is this class truly separate from the other type of audits we are trying to separate from? I see `READ`, `SEARCH` which sound more like the traces of every action i.e. audit and not like changelog 🤔 

If this class is used for both purposes does it make sense to split it? Like the approach we used so far with old/new tracker. Duplicate first, then rename.

lucaCambi77: That is a good point. The "real" [audit](https://github.com/dhis2/wow-backend/blob/master/guides/auditing.md) is already in place, but it uses another [AuditType](https://github.com/dhis2/dhis2-core/blob/91df59ab9a47d3975ef63108b22882706e4c7219/dhis-2/dhis-api/src/main/java/org/hisp/dhis/audit/AuditType.java#L33), and it is a bit more complex flow that uses JMS.
 @mortenoh   knows more about auditing in any case, so I guess he can answer more specific questions about the semantics of particular choices.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/ChangeLogType.java
lucaCambi77: yes it needs an update thanks

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/ChangeLogType.java
teleivo: Ok, got it thank you! We change `org.hisp.dhis.common.AuditType` => `ChangeLog` while
`org.hisp.dhis.audit.AuditType` remains as the "real" audit type 😅 .

teleivo (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/ChangeLogType.java
lucaCambi77: Yes, @teleivo, I agree that would need some refinement. Likewise, I am unsure that everything we are moving in the current `/audits` is an actual change log. So, we'll need to make some clean when making the new endpoints.

enricocolasante (approved):

PR #16099, opened 2024-01-08
feat: Merge indicators [DHIS2-14456]

david-mackessy:
# Summary
New feature allowing merging of `Indicator`s.
New endpoint added `POST` `/api/indicators/merge`

The crux of this feature is making sure we retrieve all metadata types with `Indicator` refs and updating them. This can include such actions as:
- removing source indicator -> adding target indicator
- setting indicator as target
- replacing any occurrences of source refs within strings with target refs (custom forms/numerator/denominator)

Details about the impacted metadata is in [Jira](https://dhis2.atlassian.net/browse/DHIS2-14456) and docs PR below.

## Refactor
Small refactor regarding the initial validation as they are similar for each merge type (validating `UID`s and valid `IdentifiableObject`s. The same initial validation can be reused across different merge types.

# Testing
## Automated
- e2e test checking a valid merge and all impacted `Indicator` associations
- int test added checking multiple scenarios
- web tests added checking auth and other scenarios

## Manual
To manual test
`POST` to `/api/indicators/merge`
with merge body like:
```json
{
    "sources": ["lOiynlltFdy", "sMTMkudvLCD"],
    "target": "btgZ1oeF9pJ",
    "deleteSources": true
}
```

# Docs
[Docs PR](https://github.com/dhis2/dhis2-docs/pull/1295)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
```diff
@@ -480,14 +481,18 @@ public List<DataElement> getDataElements() {
...
+  /**
+   * Returns all indicators in the data dimensions. The returned set is immutable. Return as
+   * distinct set instead of list as there can be many data dimension items referencing the same
+   * indicator
+   */
   @JsonIgnore
-  public List<Indicator> getIndicators() {
-    return ImmutableList.copyOf(
+  public Set<Indicator> getIndicators() {
+    return ImmutableSet.copyOf(
```
jbee, line 491: We can use `Set.copyOf` or collect to an unmodifiable set in the first place

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/SectionService.java
```diff
@@ -84,4 +85,15 @@ public interface SectionService {
    * @return a Collection of Sections.
    */
   List<Section> getAllSections();
+
+  /**
+   * Retrieves all {@link Section}s referencing the provided {@link Indicator}s.
+   *
+   * @return a Collection of {@link Section}s.
+   */
+  List<Section> getSectionsByIndicators(List<Indicator> indicators);
```
jbee, line 94: Could we reference the indicators we look for by `UID`, so `List<UID>`?
Alternatively, I think we should not always go through a service to use the store. It is ok for the merge service to use other stores. Just if something is a service layer API it should avoid accepting mapped entities as parameters.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/indicator/IndicatorService.java
```diff
@@ -57,6 +57,24 @@ public interface IndicatorService {
...
 
+  List<Indicator> getIndicatorsByUid(List<String> uids);
+
+  /**
+   * Get all {@link Indicator}s that have the uid contained in its numerator
+   *
+   * @param uid to search
+   * @return matching {@link Indicator}s
+   */
+  List<Indicator> getIndicatorsWithNumeratorContaining(String uid);
```
jbee, line 68: FYI: We have a `UID` type that we try to slowly introduce. The service API methods would be a good place to start using it even if we do not use it in our model objects yet.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/merge/MergeValidator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Set;
+import org.hisp.dhis.common.IdentifiableObject;
+import org.hisp.dhis.common.UID;
+import org.hisp.dhis.feedback.MergeReport;
+
+/**
+ * @author david mackessy
+ */
+public interface MergeValidator {
```
jbee, line 38: This looks nice!
Could do with a bit of javadoc on the methods explaining what check they do. It is not entirely clear from the name for verifySource/Target.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/DefaultMergeValidator.java
```diff
@@ -0,0 +1,210 @@
...
+              () -> {
+                mergeReport.addErrorMessage(
+                    new ErrorMessage(
+                        doesNotExistErrorCode(clazz.getSimpleName()), mergeObjectType, uid));
+                return Optional.empty();
+              });
+    } else {
+      mergeReport.addErrorMessage(
+          new ErrorMessage(noTargetErrorCode(clazz.getSimpleName()), mergeObjectType, uid));
+      return Optional.empty();
```
jbee, line 146: That is a bit over my use of `Optional` limit :D

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/DefaultMergeValidator.java
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * @author david mackessy
+ */
+@Component
+@RequiredArgsConstructor
+public class DefaultMergeValidator implements MergeValidator {
+
+  private final IdentifiableObjectManager manager;
+  private static final String INDICATOR_TYPE = "IndicatorType";
+  private static final String INDICATOR = "Indicator";
```
jbee, line 56: Since you use the `X.class.getSimpleName()` as argument to the switches where these are used I think you should also initialise the constants the same way.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/indicator/IndicatorMergeProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author david mackessy
+ */
+@Component
+@RequiredArgsConstructor
+public class IndicatorMergeProcessor implements MergeProcessor {
+  private final MergeService indicatorMergeService;
+
+  @Override
+  public MergeService getMergeService() {
+    return indicatorMergeService;
```
jbee, line 47: I feel like there now is too much surface API concerned with merge. We have service, handler, processor. 
What is a bit inconsistent with the terms as we use it normally is that there is one `MergeService` per merge type which is resolved. This creates the need to keep track of the correct implementation for the particular type. Wouldn't it be simpler to make that part of the implementation become private by adding the `MergeType` to the `MergeParams`  so that the same `MergeService` (to the outside) handles any merge request. Inside we could still have the different classes/components be responsible for one type each, just that this code organisation is not exposed to the caller making it harder to understand how to call this correctly? Makes sense?
And that with that change I hope that maybe one of the handler or processor abstractions also dissolves. But we talked about this and this was something like the glue between controller and service call chains. So if we still need it that is fine.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/indicator/handler/MetadataIndicatorMergeHandler.java
```diff
@@ -0,0 +1,205 @@
...
+import org.springframework.stereotype.Service;
+
+/**
+ * Merge handler for metadata entities.
+ *
+ * @author david mackessy
+ */
+@Service
+@RequiredArgsConstructor
+public class MetadataIndicatorMergeHandler {
```
jbee, line 58: The `Metadata` prefix seems inconsistent with the rest where we just said `Indicator*` or I am not getting why this is different.

jbee (commented):
Looks good :+1: 
A few minor comments.

One bigger thought on the `MergeService` API. Maybe we discuss to see if that is a good idea or not.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
david-mackessy: i'll update this

dhis-2/dhis-api/src/main/java/org/hisp/dhis/indicator/IndicatorService.java
david-mackessy: i'll update this

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/SectionService.java
david-mackessy: The underlying store call uses the `Indicator`s to retrieve items from the mapping table `sectionindicators`.
If we were to use `UID`s here then we would have to get the full `Indicator` objects again before making the store call. (basically going from `Indicator` -> `UID` -> `Indicator` which doesn't seem worth it).
Can chat about it if it's not clear.

david-mackessy: I can update to use the store directly here though.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/merge/MergeValidator.java
david-mackessy: all the javadocs are in the implementation :) Not sure if we have a convention for this? Should it be in the interface or the implementation? I've seen a mix.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/DefaultMergeValidator.java
david-mackessy: that's a very low limit :) 
all it does is get an optional and do something if it's present or return empty if not present.
i'll extract a method so it reads a bit nicer.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/DefaultMergeValidator.java
david-mackessy: that sounds like a good idea. I tried and believe it or not the switch doesn't allow them to be used (shows compile error `constant expression is required`)

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/indicator/handler/MetadataIndicatorMergeHandler.java
david-mackessy: happy to remove it.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/indicator/IndicatorMergeProcessor.java
david-mackessy: we chatted about this and are happy with the current implementation, after walking through a few things.

jbee (approved):

stian-sandvold (approved):

vietnguyen (approved):
Nice work David. 
Would be great to have few more tests for the methods with sql queries in store layer.

stian-sandvold:
> Nice work David. Would be great to have few more tests for the methods with sql queries in store layer.

So sorry Viet! I just merged exactly when you gave your feedback :) Feel free to re-open this @david-mackessy or create a separate PR if you want to add more tests

david-mackessy:
yep I'll add a few store tests in a separate PR 👍

PR #16111, opened 2024-01-09
fix: validation of /tracker/trackedEntities?order TECH-1648 [2.40]

teleivo:
Similar to https://github.com/dhis2/dhis2-core/pull/16097 but for trackedEntities.

In https://dhis2.atlassian.net/browse/TECH-1611 we harmonized ordering trackedEntities in 2.41. This now allows ordering only by field names of new tracker. Previously, users could also order by 'created' since TrackedEntityInstanceQueryParams.OrderColumn is responsible for mapping field names and is used by new and old tracker.

This change rejects any order request for non supported fields with a 400 like in 2.41.

* supported field names comes from master at
https://github.com/dhis2/dhis2-core/blob/d9af17c0fdac03b318208f330a76ab95d4bba1ac/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityMapper.java#L56-L68
* mapping to internal field names is already done by `TrackedEntityInstanceQueryParams.OrderColumn` which contains an entry for every supported field

### DIFF
--- a/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
+++ b/dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
@@ -43,12 +43,10 @@ import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 
-import java.util.Collections;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/TrackerTrackedEntityCriteriaMapperTest.java
```diff
@@ -215,7 +216,7 @@ void testMapping() throws BadRequestException, ForbiddenException {
     criteria.setSkipPaging(false);
     criteria.setIncludeDeleted(true);
     criteria.setIncludeAllAttributes(true);
-    criteria.setOrder(Collections.singletonList(OrderCriteria.of("created", SortDirection.ASC)));
+    criteria.setOrder(Collections.singletonList(OrderCriteria.of("createdAt", SortDirection.ASC)));
```
lucaCambi77, line 220: Do we have MVC tests where we are testing the 400 response?

teleivo: Covered by test `testMappingOrderParamsGivenInvalidField` below IMHO

lucaCambi77 (approved):

enricocolasante (approved):

PR #16112, opened 2024-01-09
fix: job scheduler delayed execution window handling and info [DHIS2-15027]

jbee:
### Summary
Main fix is to not use "now" as the last execution time when a job has not been executed before when computing the next execution time because that could lead to missing an execution and thereby deferring it indefinitely. The "last" execution must always be in the past for this computation. So now is shifted back by the maximum delay. This should be safe for any setting.

This comes with a semantic change as well (not since 2.40 but since the feature was introduced in 2.41). When creating a job it will always execute as long as now is in the window between the intended execution time and the maximum delayed execution time. This means a newly created job running every day at 7pm will run if I create it 8pm with a the setting for maximum CRON delay being longer than 1h. 

**Further improvements:**

* when a job actually runs (start) this is logged on `debug`
* renames the `HeartbeatJob` to `Housekeeping` (was missed when the name changed)
* updates the `lastUpdated` field with each transition of the `jobStatus` to `now()`
* uses the 4h default as constant to calculate the `getNextExecutionTime()` and `getMaxDelayedExecutionTime()`
* fixes a bug in the computation of the loop alignment (cosmetic, just to have 20,40,60/00 instead of some random second of the minute when the scheduling loop is active)
* `SchedulerEntry` now has 
  * `maxDelayedExecutionTime` : the correct end of the execution window instant (setting dependent)
  * `secondsToNextExecutionTime`: seconds to next run, for convenience as this is server time based not user local time based
  * `secondsToMaxDelayedExecutionTime`: seconds to next run end of execution window (same as above)
* `JobConfiguration` now also has `getMaxDelayedExecutionTime()` but similar to `getNextExecutionTime()` both of these are only correct 100% of the time if the setting for max delay has not been changed from its default of 4h

### Automatic Testing
Existing tests were adjusted and extended to reflect the change in behaviour for newly created jobs with CRON

### Manual Testing
1.
* create a job with CRON that soon should run
* wait 
* check it did run

2.
* create a job with CRON that runs daily about 1h ago to current time
* wait 1min
* check it did run (because it was in 4h window from 1h ago)

3.
* create a job with CRON that runs that runs daily about 5h from current time
* wahit a bit (1min)
* check it did not run, maybe also check the indicated next execution time (for fully correct values use `/api/scheduler/`)4.

4.
* query `/api/scheduler` 
* check the new fields for jobs created above  `maxDelayedExecutionTime`, `secondsToNextExecutionTime` and `secondsToMaxDelayedExecutionTime` and see if their values add up to your expectations

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobScheduler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobScheduler.java
@@ -99,7 +99,7 @@ public class JobScheduler implements Runnable, JobRunner {
 
   public void start() {
     long loopTimeMs = LOOP_SECONDS * 1000L;
-    long alignment = currentTimeMillis() % loopTimeMs;
+    long alignment = loopTimeMs - (currentTimeMillis() % loopTimeMs);
     Executors.newSingleThreadScheduledExecutor()
         .scheduleAtFixedRate(this, alignment, loopTimeMs, TimeUnit.MILLISECONDS);
     scheduling.set(true);
@@ -221,6 +221,7 @@ public class JobScheduler implements Runnable, JobRunner {
               jobId, start.atZone(ZoneId.systemDefault())));
       return;
     }
+    log.debug("Running job %s");
     JobProgress progress = null;
     try {
       AtomicLong lastAlive = new AtomicLong(currentTimeMillis());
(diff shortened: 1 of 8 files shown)


vietnguyen (approved):

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobScheduler.java
```diff
@@ -221,6 +221,7 @@ private void runDueJob(JobConfiguration config, Instant start) {
               jobId, start.atZone(ZoneId.systemDefault())));
       return;
     }
+    log.debug("Running job %s");
```
teleivo, line 224: @jbee this shows up as `Running job %s` in the logs I assume this should be `log.debug("Running job {}", jobId);`

PR #16186, opened 2024-01-16
fix: Password expired form not working

netroms:
## Summary 

Fixes a bug in the `AccountController` causing the expired password reset submit form to fail with an “unauthorised”/401 error message.

### Manual tests
1. Enable password expiry in the Settings app
2. Manually modify the testing user in the database at the `userinfo` table, and set `passwordlastupdated` column to a date before the expiry time. 
3. Log in as normal, and observe you are taken to the password reset page.
4. Reset password (this step failed before this fix)

### Automatic tests
Due to the nature of this feature, there is no easy way to make an automatic test.
A full test for this feature needs to run against a live server. 
The new UI testframework in the e2e tests can be used, but will take a considerable time to make it.
I made a Jira feature issue to make such a test. [DHIS2-16492](https://dhis2.atlassian.net/browse/DHIS2-16492)
This feature will also soon be replaced with the new login app.

### Jira
[DHIS2-16490](https://dhis2.atlassian.net/browse/DHIS2-16490)

[DHIS2-16490]: https://dhis2.atlassian.net/browse/DHIS2-16490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-16492]: https://dhis2.atlassian.net/browse/DHIS2-16492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
@@ -141,6 +141,7 @@ public enum ErrorCode {
   E2042("Min value must be specified"),
   E2043("Max value must be specified"),
   E2044("Max value must be greater than min value"),
+  E2045("Case insensitive operators can only be used with constant values"),
 
   /* Outlier detection */
   E2200("At least one data element must be specified"),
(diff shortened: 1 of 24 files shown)


amcgee:
@netroms thanks for the quick fix!  Let's make sure we have a Jira issue and link it to this PR before merging

david-mackessy (approved):
It would be nice if the summary included the cause and the fix for easy reading.
It looks like we're using the `SystemUser` to update the password now.
Just curious, what would happen if a non-registered `User` calls this endpoint now?

jbee (approved):

PR #15715, opened 2023-11-16
chore: use alternative isDescendant check where possible

david-mackessy:
# Summary
While working on #15639 it was discovered that a lot of other areas of the code were also making unnecessary DB calls (get `OrgUnit`) when checking if an `OrgUnit` was a descendant of another `OrgUnit`.
This PR removes the `OrganisationUnitService` methods (that contain DB calls):
- `boolean isDescendant(OrganisationUnit organisationUnit, Set<OrganisationUnit> ancestors)`
- `boolean isDescendant(OrganisationUnit organisationUnit, OrganisationUnit ancestor)`

and replaces them with the `OrganisationUnit` methods directly (that perform checks on in-memory Objects):
- `boolean isDescendant(Collection<OrganisationUnit> ancestors)`
- `boolean isDescendant(OrganisationUnit ancestor)`

# Testing
- These changes are covered by a multitude of existing tests (both directly & indirectly)
- Also added a few more assertions in existing test

# Notes
Added a `parent` mapping in the `TrackedEntityMapper` class. This helped resolve a Tracker test with some setup issues. Tagged a Tracker Dev in the PR to check if it looks ok.

PR #16194, opened 2024-01-17
fix: prevent a user change access to a higher organisation unit

netroms:
Backport of: #15149

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
@@ -69,6 +69,8 @@ import org.hisp.dhis.dataset.DataSet;
 import org.hisp.dhis.feedback.ErrorCode;
 import org.hisp.dhis.feedback.ErrorReport;
 import org.hisp.dhis.hibernate.exception.UpdateAccessDeniedException;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.hisp.dhis.organisationunit.OrganisationUnitService;
 import org.hisp.dhis.period.PeriodType;
 import org.hisp.dhis.security.PasswordManager;
 import org.hisp.dhis.security.SecurityService;
@@ -109,6 +111,8 @@ public class DefaultUserService implements UserService {
 
   private final AclService aclService;
 
+  private final OrganisationUnitService organisationUnitService;
+
   public DefaultUserService(
       UserStore userStore,
       UserGroupService userGroupService,
@@ -118,7 +122,8 @@ public class DefaultUserService implements UserService {
       CacheProvider cacheProvider,
       @Lazy PasswordManager passwordManager,
       @Lazy SecurityService securityService,
-      AclService aclService) {
+      AclService aclService,
+      @Lazy OrganisationUnitService organisationUnitService) {
     checkNotNull(userStore);
     checkNotNull(userGroupService);
     checkNotNull(userRoleStore);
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -610,26 +617,45 @@ public boolean isAccountExpired(User user) {
...
+    checkIsInOrgUnitHierarchy(user.getTeiSearchOrganisationUnits(), currentUser, errors);
 
     return errors;
   }
 
-  private void validateUserGroups(User user, User currentUser, List<ErrorReport> errors) {
+  private void checkIsInOrgUnitHierarchy(
+      Set<OrganisationUnit> organisationUnits, User currentUser, List<ErrorReport> errors) {
+    for (OrganisationUnit orgUnit : organisationUnits) {
+      boolean inUserHierarchy = organisationUnitService.isInUserHierarchy(currentUser, orgUnit);
```
jbee, line 646: @david-mackessy didn't we do this based on the user now?

david-mackessy: yes, this seems like it's falling into the same trap again, going to the DB for each `OrgUnit` when we already have everything needed to perform the check (it seems here). This was the [related PR](https://github.com/dhis2/dhis2-core/pull/15715).
It looks like we could just call `unit.isDescendant(user.getOrganisationUnits())` here to avoid any further DB calls.
Also, we seem to iterate through every `OrgUnit` here regardless whether we find a match or not. Do we need to know each answer or can we exit early if we find a match?

david-mackessy:
what PR is this a backport of? it's not easily found as a reviewer.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: I think we need to check all in the list.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -610,26 +617,59 @@ public boolean isAccountExpired(User user) {
...
+  //      if (!orgUnit.isDescendant(currentUser.getOrganisationUnits())) {
+  //        errors.add(
+  //            new ErrorReport(
+  //                OrganisationUnit.class,
+  //                ErrorCode.E7617,
+  //                orgUnit.getUid(),
+  //                currentUser.getUsername()));
+  //      }
+  //    }
+  //  }
```
jbee, line 670: ?

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: Seems that E2E tests fails when I use the orgUnit.isDescendant() method.

jbee: I was commenting because of the commented out code, I guess it can be removed

netroms: It is just there while I experiment with both methods

PR #16197, opened 2024-01-17
fix: Outliers endpoint "orderBy" not respecting "headers" name

d-bernat:
The 'Outliers API' offers the 'OrderBy' query parameter for sorting database data in the SELECT SQL statement. However, the parameter did not adhere to the convention of having the same name as the header names.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -157,7 +157,7 @@ public enum ErrorCode {
   E2210("Data end date not allowed"),
   E2211("Algorithm min-max values not allowed"),
   E2212("Specifying both a start date/end date and a relative period is not allowed"),
-
+  E2213("Order by value is not compatible with algorithm"),
```
gnespolino, line 160: could we add `{0}` to return which algorithm ?

gnespolino (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
maikelarabori:
```suggestion
  E2213("Value of param orderBy is not compatible with algorithm `{0}`"),
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -444,6 +444,7 @@ public enum ErrorCode {
 
   E7180(
       "The analytics outliers data does not exist. Please ensure analytics job was run and did not skip the outliers"),
+  E7181("Order by param `{0}` does not exist"),
```
maikelarabori, line 447:
```suggestion
  E7181("Column `{0}` specified, in orderBy, does not exist"),
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/outlier/data/OutlierQueryParser.java
```diff
@@ -164,4 +168,54 @@ private List<Period> getPeriods(RelativePeriodEnum relativePeriod, Date relative
...
         .toList();
   }
+
+  /**
+   * The method retrieves the enumerator item related to the column for order by statement.
+   *
+   * @param orderBy is the order by field. It has to be the same as header item name.
+   * @return the {@link Order}.
+   */
+  private Order getOrderBy(String orderBy) {
```
maikelarabori, line 178: We could, probably, replace all these "IFs" by creating a method inside the enum Order that returns the respective Order based on the `orderBy` (which should be renamed to `headerName`).

d-bernat: done

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
d-bernat: not all columns are eligible for order by and still can exist. What about Column `{0}` specified, in orderBy, is not eligible for orderBy or does not exist

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
d-bernat: done

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
maikelarabori: Makes sense @d-bernat 👍

maikelarabori (approved):

PR #16198, opened 2024-01-17
fix: utf-8 encoding in email sender (2.40)

vietnguyen:
backport missing codes from https://github.com/dhis2/dhis2-core/pull/14166


Philip-Larsen-Donnelly (approved):
approved as backport

PR #16210, opened 2024-01-18
refactor: Replace usage of statement builder

larshelge:
Postgres is the only supported database platform. H2 is used for unit tests, and PostgreSQL through testcontainers is used for integration tests. This means there is reduced need for the `StatementBuilder` interface, which was originally designed to support both Postgres and MySQL as database for deployments. This PR removes unnecessary usage of `StatementBuilder`.

The SonarCloud warnings point to code not touched by this PR.


jbee (approved):
Thanks for the cleanup!

I don't see any classes actually removed in this PR. What makes the statement builder still required? Is it used elsewhere?

larshelge:
Good question @jbee . The statement builder was initially meant only for SQL syntax that was different across Postgres, MySQL and H2. I can see it has now been entangled into the program indicator engine, with methods to construct program indicator logic, but with only one implementation (for Postgres). I can work with the relevant people to those methods moved out and into a more suitable place, but I need some help there.

PR #16211, opened 2024-01-18
DO NOT MERGE: can we make total and pageCount optional

teleivo:
this would make the common Pager reusable by endpoints which offer disabling totals


teleivo:
closing this one. will create an issue to implement this properly

PR #16212, opened 2024-01-18
chore: Add API ping request ignoring config

netroms:
# Summary

Adds back, configuration to ignore authentication on calls to `/api/ping` endpoint. This was lost during the migration to non deprecated Spring security config.

## Manual test
1. Call endpoint `/api/ping` (without being logged in first)
2. Observe call is executed and returns 200


david-mackessy (approved):

david-mackessy:
having a test for it would mean we would catch if it's ever mistakenly changed in future

vietnguyen (approved):

PR #16225, opened 2024-01-18
fix: remove mergeMode from metadata import params

vietnguyen:
- In Metadata Import Service, we have MergeMode which can be MERGE or REPLACE.
- Currently, the MERGE mode is not working properly. It doesn’t support merging  collections with item has type of embedded object or JsonB object. In this case, the PATCH endpoint is more suitable.
- Therefore, we should remove the MergeMode parameter from the metadata import service and the merge operation should always be REPLACE


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataImportService.java
```diff
@@ -184,6 +184,11 @@ public MetadataImportParams getParamsFromMap(Map<String, List<String>> parameter
     params.setFirstRowIsHeader(getBooleanWithDefault(parameters, "firstRowIsHeader", true));
     params.setAsync(getBooleanWithDefault(parameters, "async", false));
 
+    if (params.getMergeMode() == MergeMode.MERGE) {
+      throw new MetadataImportException(
+          "Merge mode MERGE is not supported anymore, please try again with REPLACE mode.");
```
larshelge, line 189: Maybe something like:

> Merge mode MERGE is no longer supported, only merge mode REPLACE is available

larshelge:
It is good to give feedback through the API. I guess we can remove the entire parameter in 2.42 if only a single mode is supported anyway.

vietnguyen:
> It is good to give feedback through the API. I guess we can remove the entire parameter in 2.42 if only a single mode is supported anyway.

Yes, agree. We will remove the it from 2.42.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataImportService.java
vietnguyen: fixed

netroms (approved):

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataImportService.java
```diff
@@ -184,6 +184,11 @@ public MetadataImportParams getParamsFromMap(Map<String, List<String>> parameter
     params.setFirstRowIsHeader(getBooleanWithDefault(parameters, "firstRowIsHeader", true));
     params.setAsync(getBooleanWithDefault(parameters, "async", false));
 
+    if (params.getMergeMode() == MergeMode.MERGE) {
+      throw new MetadataImportException(
+          "Merge mode MERGE is no longer supported, only merge mode REPLACE is available.");
```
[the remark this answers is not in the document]
vietnguyen: This is at the request parameters validation process in the controller, in which we usually throw direct exception. Then those exception will be caught by the `CrudControllerAdvice`. I'm not sure if we agree to always use `ErrorCode` at this stage. We do throw ErrorCode when in metadata import process. 
I think the whole error catching of `api/metadata` endpoint need some refactoring including exceptions thrown by database layer always result in http error 500.
So i think this could be another refactoring PR, will discuss with team.

jbee (commented):
As we discussed please also remove the `MergeMode.MERGE` constant.

vietnguyen:
> As we discussed please also remove the `MergeMode.MERGE` constant.

@jbee There are few internal uses of the `Merge` mode which I'm not sure if we have enough test coverage for them.

1. [UserController.java#L481](https://github.com/dhis2/dhis2-core/blob/3d3a3b31b4ed9846ef18ec0cec271913a315b9d2/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java#L481)
2. [DefaultGmlImportService.java#L297](https://github.com/dhis2/dhis2-core/blob/3d3a3b31b4ed9846ef18ec0cec271913a315b9d2/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/gml/DefaultGmlImportService.java#L297)

PR #16251, opened 2024-01-23
chore: update to 0.11.1 API

jbee:


dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/webapi/json/domain/JsonObjectReport.java
```diff
@@ -40,8 +40,8 @@ default Class<?> getKlass() {
     return getString("klass").parsedClass();
   }
 
-  default int getIndex() {
-    return getNumber("index").intValue();
+  default Integer getIndex() {
```
jbee, line 43: FYI: This inconsistency was found by the new validation. `int` meant that this value should be present but it never was a field that always is available so the proper way to encode is to use `Integer` or provide a default and mark it as optional.

PR #16255, opened 2024-01-23
feat: Database table and SQL builder abstraction

larshelge:
Introduces a database abstraction layer with a data model and a SQL builder for PostgreSQL. Applies the abstraction to the resource table service.

The database data model includes the following classes:

* `Table`
* `Column` with `Nullable` constraint
* `DataType`
* `Index` with `Unique` constraint

The `SqlBuilder` interface offers methods for creating tables, analyze, vacuum, rename, exists and drop. Currently a `PostgreSqlBuilder` implementation exists.

The `SqlBuilder` interface is applied to the resource table service as a part of an overall refactor. Each resource table class now produces a `Table` instance which defines the columns, primary key and indexes of the resource table.

The benefits of this approach are:

* The SQL operations and query building are separated from the business logic of creating the resource tables, which means that the SQL builder can be tested in isolation and re-used for several types of dynamic tables.
* Client code can take an object-oriented approach to defining tables and columns and don't have to mix table logic with SQL building logic.
* Allow for introducing new implementations of the `SqlBuilder` interface for other database systems without making changes to the service code.

The new table data model is part of the effort for a introducing a dedicated analytics database for DHIS 2. The next steps are:

* Apply the database table model and SQL builder to the analytics table generator.
* Introduce a new implementation of the `SqlBuilder` interface for the DBMS to be selected as dedicated analytics database.

Unit tests and integration tests are included.

This PR may be best reviewed by checking out the branch and inspecting the `org.hisp.dhis.db` and `org.hisp.dhis.resourcetable` packages in the `dhis-service-analytics` project.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/CategoryResourceTable.java
```diff
@@ -35,67 +36,82 @@
...
-        "create "
-            + tableType
-            + " table "
-            + getTempTableName()
-            + " ("
-            + "categoryoptioncomboid bigint not null, "
-            + "categoryoptioncomboname varchar(255), ";
+  private List<Column> getColumns() {
+    List<Column> columns =
+        Lists.newArrayList(
```
jbee, line 74: `List.of` please :)

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/CategoryResourceTable.java
```diff
@@ -35,67 +36,82 @@
...
+            new Column("categoryoptioncomboid", DataType.BIGINT, Nullable.NOT_NULL),
+            new Column("categoryoptioncomboname", DataType.VARCHAR_255));
 
     UniqueNameContext nameContext = new UniqueNameContext();
```
jbee, line 78: I wonder if something like the unique name task can and should become part of the builder?!

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/OrganisationUnitGroupSetResourceTable.java
```diff
@@ -28,64 +28,100 @@
...
   }
 
   @Override
   public Optional<String> getPopulateTempTableStatement() {
     String sql =
         "insert into "
-            + getTempTableName()
+            + toStaging(TABLE_NAME)
             + " "
             + "select ou.organisationunitid as organisationunitid, ou.name as organisationunitname, null as startdate, ";
```
jbee, line 122: Is there a reason that this SQL is not part of the responsibilities of the builder?

jbee (approved):
The new foundation improves the readability quite a bit.
I am wondering if we should use `record` classes or if you plan to backport this. 

I would hope that we can make building the actual SQL in the builder more based on templates than using string concatenation as we then can benefit from SQL validation if we use language injection in the IDE. But that is always something we can look at later.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/CategoryResourceTable.java
larshelge: This part is very specific to the resource table, and we depend on the existing table names in analytics, so fits better here.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/OrganisationUnitGroupSetResourceTable.java
larshelge: Good question. I wanted to make SqlBuilder general and multi-purpose. This logic is very specific and different for each resource table type.

I did think of having ResourceTable  extend Table, but these methods require variable input arguments so it got tricky. We can revisit this later for sure.

larshelge:
I agree using templates/prepared statements/stored procedures is good in general and something we should do more of. Creating tables and indexes means dynamic SQL which is not a great fit though.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/db/model/Column.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Represents a database table column.
+ *
+ * @author Lars Helge Overland
+ */
+@Getter
+@RequiredArgsConstructor
+public class Column {
+  /** Column name. Required. */
+  private final String name;
```
david-mackessy, line 43: If this property is required then having some validation in place to match our expectation would help reduce potential bugs and unexpected state. E.g. it's currently possible to do something like this `new Column(null, null, null)`
This goes for any new proposed DB model class with a 'required' comment above it.

david-mackessy:
Without knowing too much about the dynamic analytics tables, it looks like a nice improvement on what is currently there (particularly the sql string concat). I would be in favour of using actual sql rather than through a custom builder though. But if this helps this use case regarding code reuse & testability great 👍

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/db/model/Column.java
larshelge: Yes agree. I had Lombok @Nonull but noticed it is banned in our code base now.

jbee: Use can use `javax` annotations

larshelge: Okay. Does that take effect during runtime or is it only to assist static analysis tools?

PR #16275, opened 2024-01-24
fix: prevent 500 when job input file is no longer available [DHIS2-16521]

jbee:
Takes care of a possible exception that would end in a HTTP 500 in case the file for a file resource is not in the storage (even though it should be).


david-mackessy (approved):
I think that may have been a reported potential bug in sonar as well. 2 birds with 1 stone 👍

larshelge (approved):

PR #16280, opened 2024-01-24
fix: don't throw DeleteNotAllowException during metadata import

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-15772

### Issue
- When importing with importStrategy=DELETE if `DeleteNotAllowException` is thrown then it will abort the current import process. 

### Fix
- Remove the  method call `store.delete()`.
- Instead, explicitly call `DeleteManager.onDeletionWithoutRollBack` to avoid transaction rollback. Then call `session.delete(entity)` and catch `DeleteNotAllowException`
- Need to decrease the `stats.delete` because it has been +1 in validation process.

### Test
- Added test in ControllerTest
- Can manually test using the Import/Export app.


david-mackessy (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MetadataImportExportControllerTest.java
```diff
@@ -441,4 +441,34 @@ void exportUserWithSkipSharing() {
...
+  @DisplayName(
+      "Should return error in import report if deleting object is referenced by other object")
+  void testDeleteWithException() {
+    POST(
+            "/metadata",
+            "{\"optionSets\":"
+                + "    [{\"name\": \"Device category\",\"id\": \"RHqFlB1Wm4d\",\"version\": 2,\"valueType\": \"TEXT\"}]"
+                + ",\"dataElements\":"
+                + "[{\"name\":\"test DataElement with OptionSet\", \"shortName\":\"test DataElement\", \"aggregationType\":\"SUM\",\"domainType\":\"AGGREGATE\",\"categoryCombo\":{\"id\":\"bjDvmb4bfuf\"},\"valueType\":\"NUMBER\",\"optionSet\":{\"id\":\"RHqFlB1Wm4d\"}"
+                + "}]}\n")
```
jbee, line 455: better use a text block 
```java
"""
  {...
  }"""
```

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/DefaultObjectBundleService.java
```diff
@@ -376,14 +380,16 @@ private <T extends IdentifiableObject> TypeReport handleDeletes(
...
+            deletionManager.onDeletionWithoutRollBack(new ObjectDeletionRequestedEvent(object));
+            session.delete(object);
+            bundle.getPreheat().remove(bundle.getPreheatIdentifier(), object);
           } catch (DeleteNotAllowedException ex) {
-            lastEx.set(ex);
-            throw ex;
+            objectReport.addErrorReport(
+                new ErrorReport(klass, new ErrorMessage(ex.getMessage(), ErrorCode.E4030, null)));
+            typeReport.getStats().incIgnored();
+            typeReport.getStats().decDeleted();
```
jbee, line 390: Maybe it is time to move the try catch block into a helper method for delete

jbee (approved):

PR #16286, opened 2024-01-24
feat: persistent errors for data value import [DHIS2-16531]

jbee:
### Summary
Adds persistent errors to the data value import.

As the GeoJSON import is based on the same conflict model I also update the GeoJSON import to support this feature.

The PR also changes the `code` param from `ErrorCode` to `String` because internally multiple enums are used (tacker+metadata) so it is not correct to expect just `ErrorCode` values.

### Manual Testing
* open import/export app
* export some data (select so you get a small set)
* edit the file, change a UID of a referenced object, e.g. a data element to a non existing UID
* import the now broken data file
* check the issue is listed in the summary shown in the import/export app
* browse to `/api/jobConfigurations/errors` 
* look for the job created for the import and check its errors contain the same error information


david-mackessy (approved):
if existing tests already cover any changes made then great, otherwise a test confirming any new behaviour would also be great.

netroms (approved):

PR #16294, opened 2024-01-25
chore: sonar warnings on equals methods not being sufficient

jbee:
The issue(s) indicated by sonar was the equals method. 

I went ahead and used lombok generated methods for `equals`, `hashCode`, `toString` and getter/setters where possible.

In some cases this slightly changed the equals/hashCode as they were wrongly inconsistent using `id` in one but not the other.
I then removed use of `id` in `hashCode` so it is consistent with `equals`.


david-mackessy (approved):
that's some nice cleaning :)

larshelge (approved):

PR #16295, opened 2024-01-25
feat: Distinguish non-existent repetitions from empty values in TEI API [DHIS2-16232]

d-bernat:
The data values retrieved from the database do not carry information about their origin. The database delivers <<NULL>> for missing values, which is a reliable indication of the user's intention not to set the data value for primitive data types. However, for complex data like JSON structures, another possibility exists where the data does not exist at all in the database, and a value cannot be set. To distinguish the origin, a row context structure is used, such as the following example:
`"rowContext": {
        "1": {
            "17": {
                "valueStatus": "ND"
            }
        }
    }`.

- "1" -> row index (second row)
- "17" -> column index (18th column)
- ND* -> not defined (the data does not exist )
- 

* There are three shortcuts used for origin identification:

- S -> set
- NS -> not set
- ND -> not defined.


d-bernat:
> Looks good @d-bernat. Could you add some e2e tests?
> 
> Thx!

E2E done

maikelarabori (approved):

gnespolino (approved):

PR #16299, opened 2024-01-25
fix: Exporting users with skipSharing should have expected fields [DHIS2-16512]

david-mackessy:
# Summary
## Issue
Endpoint: `GET` `/api/metadata.json?skipSharing=true&download=true&users=true`
When exporting users and choosing the `skipSharing=true` option, the returned payload is missing the following fields:
- `username`
- `userRoles`

## Cause
This was working up until this PR (#13029), which tried to rectify issues with Tracker field filtering. This inadvertently caused this bug.
That PR(#13029) performs `String::startsWith` checks which helped resolve the Tracker field filtering but it now catches more than we want.
e.g. if the field starts with `user` it will be excluded from the final payload (`username` & `userRoles` were being removed by mistake)

## Fix
Instead of performing `String::startsWith` we now just check the root field vs the field to exclude. If they are equal then we exclude.

# Testing
## Automated
- web test added
- int test added
- unit tests added for new helper method
- there were a lot of tests in the original PR(#13029) which give confidence that the original Tracker requirements still work 🙂

## Manual
`GET` `/api/metadata.json?skipSharing=true&download=true&users=true`
- check that `username` & `userRoles` fields are present

# Info
These are the [fields](https://github.com/dhis2/dhis2-core/blob/4516b49e4ca5c0473dca699d6e1222b4879b355f/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/fieldfilter/FieldFilterService.java#L38) used in the `skipSharing` option:
- user
- publicAccess
- userGroupAccesses
- userAccesses
- externalAccess
- sharing


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
```diff
@@ -295,6 +291,35 @@ private void applyExclusions(List<FieldPath> exclusions, Map<String, FieldPath>
...
+   *       <li>root.name.last
+   *     </ul>
+   *
+   * @param field is the actual field which should be checked. This will only be 1 word (no dot
+   *     notation)
+   * @return true if the field is equal to the fullFieldPath root <br>
+   */
+  public static boolean fieldEqualsRootField(String fullFieldPath, String field) {
+    if (ObjectUtils.anyNull(fullFieldPath, field)) return false;
+    String root = fullFieldPath.split("\\.")[0];
```
jbee, line 319: That seems a unnecessary complicated way to check this. `fullFieldPath.substring(0, fullFieldPath.indexOf('.'))` should give `root` I think.

jbee (approved):
Nice breakdown, fix and testing :100:

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldPathHelper.java
david-mackessy: the beauty of `fullFieldPath.split("\\.")[0]` is that it handles the lack of any `.` in the string.
whereas `fullFieldPath.indexOf('.')` could return -1 & then we're into using more logic to get the result we want.

vietnguyen (approved):

PR #16342, opened 2024-01-29
fix: nested indicator w/periodOffset [DHIS2-15434]

jimgrace:
See [DHIS2-15434](https://dhis2.atlassian.net/browse/DHIS2-15434).

In an indicator expression you can reference another indicator value using the `N{...}` syntax. When this is modified by a `.periodOffset()` function, it is desired that the indicator run in the offset period. Instead, it was running in the un-offset period.

The fix is simple. The `Indicator` object generated from the  `N{...}` syntax already references a `QueryModifiers` object containing the period offset (as well as other modifiers in effect such as aggregation type). We just needed to copy the query modifiers from the indicator into the context that parses the indicator expressions. This way any items parsed within the indicator expression will have the same query modifiers that the indicator had (unless or until they are overridden by other modifiers inside the nested indicator).

This means that any other query modifiers will also be applied inside the nested indicator, which should be a good thing. The other modifiers are:

- aggregation type
- year-to-date
- minimum date
- maximum date
- value type (only used within subexpressions so doesn't apply here)

If the user doesn't want these modifiers to be used inside the nested indicator, they can easily construct the expression such that the modifiers don't apply to the nested indicator reference.

[DHIS2-15434]: https://dhis2.atlassian.net/browse/DHIS2-15434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


maikelarabori (approved):

gnespolino (approved):
LGTM

d-bernat (approved):

PR #16354, opened 2024-01-29
feat: support for ! operators [DHIS2-16459]

gnespolino:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/params/dimension/DimensionParamItem.java
```diff
@@ -62,17 +73,34 @@ public static List<DimensionParamItem> ofStrings(List<String> items) {
...
   }
 
+  /**
+   * Returns the operator for the given string.
+   *
+   * @param operator the operator string
+   * @return the operator
+   */
   private static AnalyticsQueryOperator getOperator(String operator) {
+    // If the operator starts with !, will call this method again removing the ! and then negate the
```
maikelarabori, line 89: I didn't understand this comment and the reason to check for "!!"...

gnespolino: we want to mark operators starting with `!!` as invalid

d-bernat (approved):

maikelarabori (approved):

PR #16378, opened 2024-01-31
fix: Add tableName property to Index [DHIS2-16705]

larshelge:
Adds `tableName` property to `Index`. Goal is to bring `Index` and `AnalyticsIndex` closer together before eventually being merged.


david-mackessy (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/DataElementResourceTable.java
```diff
@@ -89,12 +89,25 @@ private List<Index> getIndexes() {
     return List.of(
         new Index(
             appendRandom("in_dataelementstructure_dataelementuid"),
+            toStaging(TABLE_NAME),
```
jbee, line 92: I wonder if a `Table` object now should be responsible to create tables so that the table name does not have to be passed. Also an `Index` is an object that depends upon a `Table`. Something like
```java
Table table = ...
List.of(table.newIndex(
  appendRandom("in_dataelementstructure_dataelementuid"),
  Unique.UNIQUE,
  List.of("dataelementuid")));
```
And then this internally passes on the table name to the `new Index` call just to make it read a bit nicer?
Or even have the `Index` reference the `Table` object not just the name.

jbee (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/DataElementResourceTable.java
larshelge: Yes agree, I'm aware of the slightly duplication/redundancy. Now in most applications you will first create the table, then populate it, and then build the indexes for performance reasons, meaning indexes are not created when the table is created.

The entire analytics table code is written in a way where indexes are somewhat independent of tables. So I did it like this to make it possible to apply to the analytics code base.

PR #16380, opened 2024-01-31
refactor: Use Index instead of AnalyticsIndex [DHIS2-16705]

larshelge:
Replaces `AnalyticsIndex` and analytics `IndexFunction` with `Index` and `IndexFunction` from db model package in analytics table service.

Removes the list of `Index` from `Table` in the table data model, to remove the duplication now that `Index` has a `tableName` property.

Updates resource table service accordingly.


gnespolino (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/CategoryOptionComboNameResourceTable.java
```diff
@@ -78,6 +79,11 @@ private List<String> getPrimaryKey() {
     return List.of("categoryoptioncomboid");
   }
 
+  @Override
+  public List<Index> getIndexes() {
+    return List.of();
+  }
```
david-mackessy, line 85: I'm not sure what the intention here is with `ResourceTable`s & `Index`es, seeing a lot of empty lists being returned in this PR in this override method. Looks like a code smell to me.
If all of these tables will eventually have indexes then i think this is fine, if they might not ever then a slightly different model may be more suited, like Table extending IndexTable or something like that. So straight away we know if a table has an index from its class signature as opposed to having to check a getter method.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/resourcetable/table/CategoryOptionComboNameResourceTable.java
larshelge: Yes agree, we could make a default method that returns an empty list. The situation is that about a 1/3 of the resource table have indexes, and the interface lets tables provide their indexes.

PR #16381, opened 2024-01-31
fix: Unable to store image in different sizes

lucaCambi77:
When sending an event to the `FileResourceEventListener`, if we update a file resource, we cannot get the `UserDetails` object.
With this issue, we save the user in the `ImageFileSavedEvent` and retrieve it  from the security context before updating the file resource object.

### Note

This is a quick solution for now. As discussed with @jbee, we should move the the Job scheduler and an issue is created for that :
https://dhis2.atlassian.net/browse/TECH-1687

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/fileresource/FileResourceEventListener.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/fileresource/FileResourceEventListener.java
@@ -30,12 +30,15 @@ package org.hisp.dhis.fileresource;
 import java.io.File;
 import java.util.Map;
 import java.util.stream.Stream;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
+import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.fileresource.events.BinaryFileSavedEvent;
 import org.hisp.dhis.fileresource.events.FileDeletedEvent;
 import org.hisp.dhis.fileresource.events.FileSavedEvent;
 import org.hisp.dhis.fileresource.events.ImageFileSavedEvent;
+import org.hisp.dhis.user.AuthenticationService;
 import org.joda.time.DateTime;
 import org.joda.time.Period;
 import org.joda.time.format.PeriodFormat;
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/fileresource/FileResourceEventListener.java
```diff
@@ -78,19 +78,21 @@ public void save(FileSavedEvent fileSavedEvent) {
...
     FileResource fileResource =
-        fileResourceService.getFileResource(imageFileSavedEvent.getFileResource());
+        fileResourceService.getFileResource(imageFileSavedEvent.fileResource());
 
     String storageId = fileResourceContentStore.saveFileResourceContent(fileResource, imageFiles);
 
     if (storageId != null) {
       fileResource.setHasMultipleStorageFiles(true);
 
+      authenticationService.obtainAuthentication(imageFileSavedEvent.userUid());
```
jbee, line 94: To not leak the user context in the worker thread this should use a try-finally where in the finally you call the clear on the authenticationService.

jbee (commented):
Looks good, just needs clearing of the user context after the action is performed.

muilpp (approved):

jbee (approved):

zubaira:
> Looks good, just needs clearing of the user context after the action is performed.

@lucaCambi77 @jbee 
this bug is the result of currentUser being null. There could be many other places like this. Is this fix applicable everywhere or CurrentUserUtil has to be updated?

lucaCambi77:
> > Looks good, just needs clearing of the user context after the action is performed.
> 
> @lucaCambi77 @jbee 
> this bug is the result of currentUser being null. There could be many other places like this. Is this fix applicable everywhere or CurrentUserUtil has to be updated?

@zubaira this should happen in an async context. I had a quick look during this issue and the only place I've seen is this. If it happens in other places we shouldn't fix everywhere but rather understand why, with a security context set, we have this error.

zubaira:
> > > Looks good, just needs clearing of the user context after the action is performed.
> > 
> > 
> > @lucaCambi77 @jbee
> > this bug is the result of currentUser being null. There could be many other places like this. Is this fix applicable everywhere or CurrentUserUtil has to be updated?
> 
> @zubaira this should happen in an async context. I had a quick look during this issue and the only place I've seen is this. If it happens in other places we shouldn't fix everywhere but rather understand why, with a security context set, we have this error.

@lucaCambi77 yes we do have this problem at other places as well and those are not async operations. With that i suspect there could be more places where we can potentially face this issue.

PR #16386, opened 2024-02-01
fix: Pre-fill user-invite fields and align form fields with user-app …

netroms:
…form


david-mackessy (approved):
If this is a fix for a bug, then it should be accompanied with a test showing that the bug no longer exists.
If it's hard to recreate the condition in a test, just call this out so it's easy for a reviewer to understand.

larshelge (approved):

PR #16387, opened 2024-02-01
feat: Add table inheritance to model and SQL builder [DHIS2-16705]

larshelge:
Adds a parent table to the table model, and support for table inheritance to SQL builder. Adds basic property validation for table.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/db/model/Table.java
```diff
@@ -61,20 +63,27 @@ public class Table {
   /** Whether table is logged or unlogged. PostgreSQL-only feature. */
   private final Logged logged;
 
+  /**
+   * The parent table. This table will inherit from the parent table, if specified. Optional, may be
+   * null.
+   */
+  private final Table parent;
```
jbee, line 70: You can annotate with `@CheckForNull` to express that it might be null in a way that helps tools to analyse this correctly and that is quicker to grasp when reading the code

PR #16388, opened 2024-02-01
chore: Prevent XXE XML attack, sonar issue

david-mackessy:
# Summary
## Issue
Sonar identified 2 possible areas where [XXE XML attacks](https://sonarcloud.io/organizations/dhis2/rules?open=java%3AS2755&rule_key=java%3AS2755) could be executed.

## Fix
Ensure that we don't allow processing of external DTD or stylesheet by disabling those features.

# Testing
## Automated
Int test added confirming malicious xml is not processed

# Info
There are only 2 places where this was identified in the PRD code. We are also moving away from supporting XML. If there were any more instances of this or if we intended on supporting XML then we could just create a wrapper around the code in question (and implement these safe settings) to reduce the possibility of it popping up somewhere else in code in future.


netroms (approved):

jbee (approved):
Nice one!

PR #16392, opened 2024-02-01
fix: request param of type dates can be repeated [DHIS2-16708]

gnespolino:
Switched several date param in TEI/CPL analytics to use collection instead of string, to allow multiple params


maikelarabori (approved):
It would be nice to add an e2e test, just in case.

d-bernat (approved):

PR #16393, opened 2024-02-01
fix: SMS gateway do not require password and authToken to be send in an update [DHIS2-16695]

jbee:
### Summary 
Main goal of the PR was to allow updates of a SMS gateway configuration without the need to send in the `password` or `authToken`. If they are blank we keep the existing value, if they have a new value the new value is used. 

In connection with this the attempt was to also make these fields hidden in the API but that requires a wider general change I didn't want to include in this PR. In this context the `/config/sms` endpoint was removed as it is likely unused and outdated and exposes the entire SMS config as one object which makes it even harder to properly protect. 

The refactoring to use lombok `@Getter` and `@Setter` had also revealed bugs in the schema building where getter/setters are not found because of a naming convention mismatch which then causes the `Property` to be broken because of another bug which then causes a stack overflow exception when field filtering the list of gateways. So these issues were addressed in this PR as well. 

### Automatic Testing
Added a tests scenario to ensures the issue with the naming convention is fixed and getter/setters are found even in cases where the convention isn't followed as done by lombok generated methods.

Added test scenarios and asserts to the gateway API tests to make sure an update can be done without sending the fields in question but the fields are still kept internally. 

### Manual Testing 
* create a gateway (of type bulk and also clickatell)
* edit the gateway and leave the passoword/authToken field blank (needs UI update or use of API)
* confirm the fields still have a value after the update (currently the api `/api/systemSettings/keySmsSetting` can be used) otherwise check directly in the DB


dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/introspection/JacksonPropertyIntrospector.java
```diff
@@ -309,22 +311,12 @@ private static List<Property> collectProperties(Class<?> klass) {
...
-                    && m.getParameterTypes().length == 0);
-
-    Multimap<String, Method> multimap = ReflectionUtils.getMethodsMultimap(klass);
+    List<Field> fields = findFields(klass, f -> f.isAnnotationPresent(JsonProperty.class));
 
     Map<String, Property> propertyMap = new HashMap<>();
 
     for (var field : fields) {
-      Property property = new Property(klass, null, null);
+      Property property = new Property(field.getType(), null, null);
```
jbee, line 319: FYI: `klass` would be the type of the object having the `field` so it is the entirely wrong type for the property but it usually would be updated latter when a getter is found to the getter return type which is why it usually works unless you don't find the getter.

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/introspection/JacksonPropertyIntrospector.java
```diff
@@ -357,20 +353,14 @@ private static List<Property> collectProperties(Class<?> klass) {
         continue;
       }
 
-      Property property = new Property(klass, method, null);
+      Class<?> type = method.getReturnType();
+      Property property = new Property(type, method, null);
```
jbee, line 357: FYI: Same here, `klass` was the wrong type but that was overridden later with correct type.

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/introspection/JacksonPropertyIntrospector.java
```diff
@@ -309,22 +311,12 @@ private static List<Property> collectProperties(Class<?> klass) {
...
-        ReflectionUtils.findFields(klass, f -> f.isAnnotationPresent(JsonProperty.class));
-
-    List<Method> methods =
-        ReflectionUtils.findMethods(
-            klass,
-            m ->
-                AnnotationUtils.findAnnotation(m, JsonProperty.class) != null
-                    && m.getParameterTypes().length == 0);
-
-    Multimap<String, Method> multimap = ReflectionUtils.getMethodsMultimap(klass);
```
jbee, line 322: FYI: This map was used to see if there is a setter for the expected name but it is unnecessary complicated and also would just take the first method with a matching name not making sure it is actually a setter with matching type parameter. This now uses `findSetterMethod(fieldName, klass, type)` instead where `type` is the field or getter return type.

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/ReflectionUtils.java
```diff
@@ -229,7 +230,7 @@ public static Method findGetterMethod(String fieldName, Class<?> clazz) {
         }
       }
     }
-
+    if (property.matches("(?:is|has|get)[A-Z].*")) return _findMethod(clazz, property);
```
jbee, line 233: FYI: So the issue here is with fields and getters that do not name align, for example a field names `isDefault` and the getter being named `isDefault` (not `getIsDefault`). So after having failed to find the getter by adding a prefix we then added this fallback strategy to see if the property name itself already got a usual getter prefix and in that case we try the exact property name as getter name.

larshelge (approved):

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/introspection/JacksonPropertyIntrospector.java
larshelge: Would be to replace the "klass" naming with "type" consistently. Spring uses "type" which I guess is more common and descriptive.

david-mackessy (approved):
nice one

dhis-2/dhis-services/dhis-service-schema/src/main/java/org/hisp/dhis/schema/introspection/JacksonPropertyIntrospector.java
jbee: Renamed now

PR #16402, opened 2024-02-02
fix: make sure background threads check for not logged in current user

netroms:
# Summary
After the user refactoring in #15727, some places where a separate/background thread is accessing the current user, it will fail because they now have to check for no user before calling `CurrentUserUtil.getCurrentUserDetails/getCurrentUsername`.
This PR fixes one such issue in the `DefaultMessageService`.

## Automatic test
MessageServiceTest#testSendMessageWithNoCurrentUser()


jbee (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/message/MessageServiceTest.java
```diff
@@ -192,4 +192,18 @@ void testGetMessageConversations() {
...
+    clearSecurityContext();
+    MessageConversationParams params =
+        new MessageConversationParams.Builder()
+            .withRecipients(Set.of(userA))
+            .withSubject("subject")
+            .withText("text")
+            .withMessageType(MessageType.SYSTEM)
+            .build();
+
+    messageService.sendMessage(params);
```
david-mackessy, line 207: should there be an assertion somewhere in this test?

david-mackessy (approved):

PR #16403, opened 2024-02-02
chore: Validate user-supplied url before using, sonar issue

david-mackessy:
# Summary
## Issue
We have an endpoint `POST` `/api/synchronization/metadataPull` which accepts a user-supplied string payload.
We make a http call with the user-supplied parameter without any validation.
The endpoint can only be called by a user with the `ALL` role but we should still perform some check on the supplied param.
This is a Sonar identified [SSRF vulnerability](https://sonarcloud.io/organizations/dhis2/rules?open=javasecurity%3AS5144&rule_key=javasecurity%3AS5144).

## Proposed Fix
Proposed fix is to only accept a supplied URL if its base URL is a member of the `system.remote_servers_allowed` list obtained from config.
New config key `system.remote_servers_allowed` which retrieves the allowed list from the config file.

# Testing
## Automated
unit ->  config
web -> controller

## Manual
Manually tested by:
1. adding this to local config:
`system.remote_servers_allowed=https://raw.githubusercontent.com/dhis2/`
2. `POST` `/api/synchronization/metadataPull` with text payload `https://raw.githubusercontent.com/dhis2/dhis2-metadata-repo/master/repo/221/sims-data-elements/metadata.json`
3. Result is successful import of 404 created data elements

## Docs
[docs update](https://github.com/dhis2/dhis2-docs/pull/1313)

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/analytics/AnalyticsConstants.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/analytics/AnalyticsConstants.java
@@ -35,4 +35,12 @@ public class AnalyticsConstants {
   public static final String ANALYTICS_TBL_ALIAS = "ax";
 
   public static final String ANALYTICS_TBL_TEMP_SUFFIX = "_temp";
+
+  public static final String OWNERSHIP_TBL_ALIAS = "own";
+
+  public static final String DATE_PERIOD_STRUCT_ALIAS = "ps";
+
+  public static final String ORG_UNIT_STRUCT_ALIAS = "ous";
+
+  public static final String ORG_UNIT_GROUPSET_STRUCT_ALIAS = "ougs";
 }
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/OrgUnitField.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/OrgUnitField.java
@@ -30,14 +30,14 @@ package org.hisp.dhis.analytics;
 import static com.google.common.base.Enums.getIfPresent;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static org.apache.commons.lang3.ObjectUtils.isEmpty;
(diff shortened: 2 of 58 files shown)


dhis-2/dhis-support/dhis-support-external/src/test/resources/dhis.conf
```diff
@@ -1,5 +1,6 @@
 monitoring.api.enabled = off
 enable.query.logging = true
 method.query.logging.enabled = false
+system.remote_servers_allowed=https://validtesturl.com/,https://validtesturl2.com/
```
larshelge, line 4: Use spaces around equals.

larshelge (approved):

dhis-2/dhis-support/dhis-support-external/src/test/java/org/hisp/dhis/external/conf/DhisConfigurationProviderTest.java
```diff
@@ -51,15 +70,50 @@ void isOn() {
...
+  @ParameterizedTest
+  @NullSource
+  @ValueSource(strings = {"", " ", "https://invalidurl.com/fail"})
+  @DisplayName("invalid URLs should return false")
+  void invalidUrlTest(String url) {
+    // given there are 2 remote servers in the test config allowed list
+    // when we check if an invalid url is in the allowed list
+    boolean urlIsAllowed = configProvider.remoteServerIsInAllowedList(url);
+
+    // then it should be false
```
jbee, line 103: Comments are good if they add any information but _it should be false_ is exactly what `assertFalse(urlIsAllowed);` already tells us. Nothing can be learned here so then it is just additional clutter to skip over. I suggest if you want to enhance this think about a message to add to the `assertTrue`/`assertFalse` because these in particular don't read very well without a message when checking booleans that do not simply reflect a boolean field somewhere.

jbee (approved):

dhis-2/dhis-support/dhis-support-external/src/test/java/org/hisp/dhis/external/conf/DhisConfigurationProviderTest.java
david-mackessy: yep fair enough, it's just a habit when creating the skeleton of the test :)

PR #16382, opened 2024-01-31
fix: Relationships are not filtered by deleted in tracker/relationships endpoint

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16706

The endpoint `/tracker/relationships` does not filter by the `deleted` property. 
With this PR, we align the relationships endpoint with the other tracker endpoints and add the `includeDeleted` property in the request parameter, which defaults to false if not specified.

### Note:

This ended up with a refactoring of the Store by assigning responsibilities to extract the relationships list or the count because it was not clear what was going on

PR #16421, opened 2024-02-05
fix: Relationships are not filtered by deleted in tracker/relationships endpoint [2.39] 

lucaCambi77:
backport of https://github.com/dhis2/dhis2-core/pull/16382

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/relationship/RelationshipDeletionHandlerTest.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/relationship/RelationshipDeletionHandlerTest.java
@@ -87,7 +87,8 @@ class RelationshipDeletionHandlerTest {
 
   @Test
   void deleteTrackedEntityInstance() {
-    when(relationshipService.getRelationshipsByTrackedEntityInstance(any(), anyBoolean()))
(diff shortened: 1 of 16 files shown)


dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/relationship/RelationshipDeletionHandlerTest.java
```diff
@@ -87,15 +87,16 @@ void allowDeleteRelationshipTypeWithoutData() {
...
+            any(), anyBoolean(), anyBoolean()))
         .thenReturn(singletonList(new Relationship()));
 
     ObjectDeletionRequestedEvent event =
         new ObjectDeletionRequestedEvent(new TrackedEntityInstance());
     deletionManager.onDeletion(event);
 
     verify(relationshipService, atLeastOnce())
-        .getRelationshipsByTrackedEntityInstance(any(), anyBoolean());
+        .getRelationshipsByTrackedEntityInstance(any(), anyBoolean(), anyBoolean());
```
teleivo, line 99: `anyBoolean` can be replaced with false (or whatever representation of false is ok for mockito here), as the value will always be false in the deletion handler right?

teleivo (approved):

PR #16423, opened 2024-02-05
Dhis2 16728: Abandoned data elements

jason-p-pickering:
Adds missing unit test for abandoned data elements. Previously, the unit test for abandoned data elements was missing. This adds a new test to create a data element in the past which has no data attached to it, which should trigger the integrity check.


jbee (approved):
Would it be wise to also have a test for a data element that should not have the issue based on the date/age comparison you changed?

david-mackessy (approved):

PR #16446, opened 2024-02-06
feat: add tracker endpoint to download data element files DHIS2-16681

teleivo:


jbee (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -69,8 +78,69 @@ class DefaultEventService implements EventService {
...
 
+  private final FileResourceService fileResourceService;
+
   private final EventOperationParamsMapper paramsMapper;
 
+  @Override
+  public FileResourceStream getFileResource(UID eventUid, UID dataElementUid)
+      throws NotFoundException, ConflictException {
+    Event event = eventService.getEvent(eventUid.getValue());
+    if (event == null) {
```
muilpp, line 89: I think an event can never be null at this point, can it?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -69,8 +78,69 @@ class DefaultEventService implements EventService {
...
+      throw new NotFoundException(DataElement.class, dataElementUid.getValue());
+    }
+
+    if (!dataElement.getValueType().isFile()) {
+      throw new NotFoundException(
+          "Data element " + dataElementUid.getValue() + " is not a file (or image).");
+    }
+
+    User currentUser = userService.getUserByUsername(CurrentUserUtil.getCurrentUsername());
+    List<String> errors = trackerAccessManager.canRead(currentUser, event, dataElement, false);
```
muilpp, line 104: his is just for me to understand how this works.

1. If a user has access to the program, program stage and event, the request can still be invalid if he lacks access to any of the data elements in that event?
2. Should we validate that the supplied data element uid is part of the event?
3. EventService.getEvent is already validating the sharing access to the event, seems that we are running the same validation again just to make sure the user has access to the data element.

muilpp (approved):

PR #16449, opened 2024-02-07
fix: Improve performance of min max validation query [DHIS2-16750]

larshelge:
This PR is about the min max value validation service used by the data entry app. The SQL query produced by the validation can be improved in terms of performance.

The query filters by the path column of organisation unit, which makes the query slow. The validation is always done for a single org unit, also not including the org unit sub-hierarchy. This means that the query can filter directly on the org unit ID, which is much faster.


jbee (approved):

david-mackessy (approved):
nice one!

vietnguyen (approved):

PR #16463, opened 2024-02-08
fix: FollowUp is not persisted in programStageWorkingListFilters

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16738

With program stage working list we can define a `programStageQueryCriteria`, for example;
```
{
  "name": "test",
  "program": {
    "id": "IpHINAT79UW"
  },
  "programStage": {
    "id": "A03MvHHogjR"
  },
  "programStageQueryCriteria": {
    "displayColumnOrder": [
      "w75KJ2mc4zz",
      "zDhUuAYrxNC",
      "APtutTb0nOY"
    ],
    "order": "createdAt:desc",
    "enrollmentStatus": "ACTIVE",
    "followUp": "true",
    "eventOccurredAt": {
      "type": "RELATIVE",
      "period": "TODAY"
    },
    "attributeValueFilters": [],
    "dataFilters": []
  }
}
```
In the capture app, we can then convert the `programStageQueryCriteria` filters to an API call to filter, for example, a tracked entity list.
The `followUp` filter is currently missing, and we have fixed it with this PR. This is the API parameter for the [request](https://github.com/dhis2/dhis2-core/blob/cf8a3809d42893f9bcdca782c92acbb818ed254a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParams.java#L136), although does not reflect the class [property](https://github.com/dhis2/dhis2-core/blob/cf8a3809d42893f9bcdca782c92acbb818ed254a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Enrollment.java#L90)

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
@@ -102,5 +102,5 @@ public class ProgramStageQueryCriteria implements Serializable {
   private List<AttributeValueFilter> attributeValueFilters = Collections.emptyList();
 
   /** Property to filter events based on {@link org.hisp.dhis.program.Enrollment#followup property */
-  @JsonProperty private boolean followup;
+  @JsonProperty private boolean followUp;
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
@@ -100,4 +100,7 @@ public class ProgramStageQueryCriteria implements Serializable {
   /** Property to filter tracked entity instances based on tracked entity attribute values */
   @JsonProperty @Builder.Default
   private List<AttributeValueFilter> attributeValueFilters = Collections.emptyList();
+
+  /** Property to filter events based on {@link org.hisp.dhis.program.Enrollment#followup property */
+  @JsonProperty private boolean followup;
 }
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/ProgramStageWorkingListControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/ProgramStageWorkingListControllerTest.java
@@ -263,33 +263,32 @@ class ProgramStageWorkingListControllerTest extends DhisControllerConvenienceTes
 
   @Test
   void shouldCreateWorkingListWithProgramStageQueryCriteria() {
-
     String workingListJson =
         assertStatus(
             HttpStatus.CREATED,
             POST(
                 "/programStageWorkingLists",
                 """
-         {
-         'program': {'id': '%s'},
-         'programStage': {'id': '%s'},
-         'name':'workingListName',
-         'programStageQueryCriteria': {
-           "displayColumnOrder": [
-             "w75KJ2mc4zz",
-             "zDhUuAYrxNC",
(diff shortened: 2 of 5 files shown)


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/ProgramStageWorkingListControllerTest.java
```diff
@@ -258,17 +261,77 @@ void shouldDeleteWorkingListWhenDeleteRequested() {
...
+  void shouldCreateWorkingListWithProgramStageQueryCriteria() {
+
+    String workingListJson =
+        assertStatus(
+            HttpStatus.CREATED,
+            POST(
+                "/programStageWorkingLists",
+                """
+         {
+         'program': {'id': '%s'},
```
teleivo, line 274: Can we indent the JSON inside the `{}`?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/ProgramStageWorkingListControllerTest.java
```diff
@@ -258,17 +261,77 @@ void shouldDeleteWorkingListWhenDeleteRequested() {
     assertEquals(HttpStatus.OK, response.status());
   }
 
+  @Test
+  void shouldCreateWorkingListWithProgramStageQueryCriteria() {
+
```
teleivo, line 266: please remove leading newlines most of our tests don't have them

teleivo (changes requested):
We should not expose `followup` in the API. We just worked on getting rid of that typo in the exporter 😓

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
```diff
@@ -102,5 +102,5 @@ public class ProgramStageQueryCriteria implements Serializable {
   private List<AttributeValueFilter> attributeValueFilters = Collections.emptyList();
 
   /** Property to filter events based on {@link org.hisp.dhis.program.Enrollment#followup property */
-  @JsonProperty private boolean followup;
+  @JsonProperty private boolean followUp;
```
teleivo, line 105: AFAIK before in the PR description you said that the property has to be called `followup` because of the class used for the DB model. Does this now need some kind of mapping somewhere?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/ProgramStageWorkingListControllerTest.java
```diff
@@ -263,33 +263,32 @@ void shouldDeleteWorkingListWhenDeleteRequested() {
...
-           'enrollmentStatus': 'ACTIVE',
-           'followup': 'true',
-           'eventOccurredAt': {
-             'type': 'RELATIVE',
-             'period': 'TODAY'
-           }
-         }
-        }
-        """
+               {
```
teleivo, line 272: This JSON still looks odd. I would expect it to look like

```json
{
  "program": {}
}
```

not

```json
{
"program": {}
}
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstageworkinglist/ProgramStageQueryCriteria.java
lucaCambi77: the front end converts the parameter to the API request, but I thought using `followup`  was consistent with the class name. So if we use `followUp` it is consistent with API although do not reflect the class property

teleivo (approved):

muilpp (approved):

PR #16466, opened 2024-02-08
fix: Do not export storedby with aggregate data exchanges

jason-p-pickering:
This PR changes how data with external aggregate data exchanges is performed. The `storedby` attribute is sent as unspecified, which should have the effect of ensuring that the data on the target server side is properly attributed to the user who performs the import.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.scheduling;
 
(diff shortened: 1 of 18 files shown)


larshelge:
On stored by, I think it is fine to just not set it in the data value set analytics API.

Setting it to [aggregated] was just a random choice. We also set this value in the comment if you need to distinguish aggregated from captured data values.

If you have time, you can change the PR to simply remove this line:

https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsUtils.java#L482

larshelge (approved):

gnespolino (approved):

PR #16467, opened 2024-02-08
refactor: Rename encode to singleQuote [DHIS2-16705]

larshelge:
* Renames method `encode` with quoting (`true`) to `singleQuote` as encode is a misleading name.
* Renames method `encode` without quoting (`false`) to `escape` which is more descriptive.
* Adds backslash to the escaped characters.
* Replace and remove duplicated methods in `AnalyticsSqlUtils` which exist in `SqlUtils`.
* Remove method `removeQuote`.


maikelarabori (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
```diff
@@ -201,12 +201,24 @@ public interface SqlBuilder {
    */
   String quote(String relation);
 
+  /**
+   * @param relation the relation to quote, e.g. a column name.
+   * @return an aliased and double quoted relation.
+   */
+  String quote(String alias, String relation);
```
jbee, line 208: are these truly database dependent? If so `quote` or `singleQuote` is not a better name as another database might do something else to encode a literal. If we assume all these are always using a single quote as literals these seem misplaced as a abstract function and should be static functions.

jbee (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
larshelge: Yes agreed @jbee , was actually thinking the same. You are right. We should find a more agnostic name. The best I can think of is "relation" vs "value". In Postgres and many databases, you would double quote a relation, meaning table name, index name, column name, while you would single quote values, such as items in where filters, e.g. `where dx in ('a', 'b')`. So perhaps `valueQuote`. Open for ideas.

PR #16476, opened 2024-02-09
feat: create login endpoint for new login app

netroms:


david-mackessy (approved):
Are there docs for the new endpoint? (including the difference between `/auth` & `/authentication` )
I'm not sure about using the embedded jetty module this way for e2e tests for a few reasons:
- Ideally e2e tests should know nothing about the server apart from the public API
- using types from code (`LoginResponse`) as the response to an e2e test could have some unknown side effects that we may miss, like dropping some property during deserialization etc.
- the startup used for the test is not using the PRD code to startup

It feels like this sits between int & e2e test.

vietnguyen (approved):

PR #16481, opened 2024-02-12
fix: Add variable support for job progress log messages

larshelge:
Adds Slf4j-style variable substitution support for `JobProgress` for more elegant logging and to avoid string concatenation.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobProgress.java
@@ -236,8 +236,8 @@ public interface JobProgress {
     startingStage(description, 0, onFailure);
   }
 
-  default void startingStage(String format, Object... args) {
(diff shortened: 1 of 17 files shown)


david-mackessy (approved):

mortenoh (approved):

jbee (approved):
A side note: while this works for the simplest form of `startingX` and `completedX` calls there are already more complex forms which now are different in that they do not allow for this substitution. These will often use other means of formatting or assembling messages which will leave a feeling of inconsistency. We should look into this in another step. I think at least in some places we can make this more consistent again but there will also be places where it is hard as the messages are generated from the data.

larshelge:
Yes agree. It is somewhat inconsistent. Varargs should be the last argument. I did not want to go into a very large refactor in this PR.

To be fair, Slf4J also uses a mix of regular string and format/argument style messages:

https://www.slf4j.org/api/org/slf4j/Logger.html#info(org.slf4j.Marker,java.lang.String)

PR #16482, opened 2024-02-12
fix: Do not export storedby with aggregate data exchanges [DHIS2-16761]

jason-p-pickering:


david-mackessy (approved):
if this is a backport, can that be put in the description. That's the convention we usually use, which makes it easy for the reviewer to see the original PR.
e.g. backport of #{pr-number}

PR #16484, opened 2024-02-12
feat: add support for unpooled DB pool type

claudemamo:
Refs: DHIS2-16396


dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -176,13 +178,16 @@ public ConfigKeyMapper getMapper() {
...
   }
 
+  private DatabasePoolUtils() {}
+
   public static DataSource createDbPool(PoolConfig config)
       throws PropertyVetoException, SQLException {
     Objects.requireNonNull(config);
 
     DbPoolType dbType = DbPoolType.valueOf(config.dbPoolType.toUpperCase());
+    log.info(String.format("Database pool type value is [%s]", dbType));
```
david-mackessy, line 188: no need for the string format, logging can handle the param e.g.
`log.info("Database pool type value is {}", dbType);`

dhis-2/dhis-support/dhis-support-hibernate/src/test/java/org/hisp/dhis/datasource/DatabasePoolUtilsTest.java
```diff
@@ -0,0 +1,169 @@
...
+            .dbPoolType(DatabasePoolUtils.DbPoolType.UNPOOLED.name())
+            .jdbcUrl("jdbc:fake:db")
+            .username("")
+            .password("")
+            .dhisConfig(mockDhisConfigurationProvider);
+
+    DataSource dataSource = DatabasePoolUtils.createDbPool(poolConfigBuilder.build());
+    assertInstanceOf(DriverManagerDataSource.class, dataSource);
+
+    assertEquals("Database pool type value is [UNPOOLED]", getLogEntry());
```
david-mackessy, line 92: Instead of asserting on log entries to confirm the chosen pool config, could we assert on the actual DataSource properties returned from those methods?
Each datasource seems to have unique requirements which could be confirmed with assertions. This would leave the test in a much more reliable state, being able to catch actual changes in code (which a log statement cannot)
That would also remove the need for all the logging setup & dependency in this PR.

stian-sandvold (approved):

david-mackessy:
is there be a docs update to go with this change?

dhis-2/dhis-support/dhis-support-hibernate/src/test/java/org/hisp/dhis/datasource/DatabasePoolUtilsTest.java
claudemamo: I actually want to test that logging is happening give that this was specifically requested by @bobjolliffe

claudemamo:
> is there a docs update to go with this change?

I've opened a PR here: https://github.com/dhis2/dhis2-docs/pull/1319

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -252,6 +259,34 @@ private static DataSource createHikariDbPool(PoolConfig config) throws SQLExcept
...
+        dhisConfig.getProperty(mapper.getConfigKey(CONNECTION_DRIVER_CLASS));
+    final String jdbcUrl =
+        firstNonNull(
+            config.getJdbcUrl(), dhisConfig.getProperty(mapper.getConfigKey(CONNECTION_URL)));
+    final String username =
+        firstNonNull(
+            config.getUsername(), dhisConfig.getProperty(mapper.getConfigKey(CONNECTION_USERNAME)));
+    final String password =
+        firstNonNull(
+            config.getPassword(), dhisConfig.getProperty(mapper.getConfigKey(CONNECTION_PASSWORD)));
```
david-mackessy, line 277: There are many duplicated lines in this class, could we possibly reuse some code here?
Using sonarlint locally identifies these.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -176,14 +178,19 @@ public ConfigKeyMapper getMapper() {
...
   public static DataSource createDbPool(PoolConfig config)
       throws PropertyVetoException, SQLException {
     Objects.requireNonNull(config);
 
     DbPoolType dbType = DbPoolType.valueOf(config.dbPoolType.toUpperCase());
+    log.info("Database pool type value is [{}]", dbType);
 
     if (dbType == DbPoolType.C3P0) {
       return createC3p0DbPool(config);
+    } else if (dbType == DbPoolType.UNPOOLED) {
```
david-mackessy, line 192: would using a switch on the enum value be better here (instead of ifs)? It allows the compiler to help us make sure that all DbTypes are handled (ensuring if any further are added that we handle them as well)

dhis-2/dhis-support/dhis-support-hibernate/src/test/java/org/hisp/dhis/datasource/DatabasePoolUtilsTest.java
david-mackessy: sorry @claudemamo I added a few more comments after pulling the code locally and checking it again (as opposed to browsing the changes in GitHub)

david-mackessy: that log assertion proves the code path which is fine (it still leaves the datasource properties untested)

claudemamo: The point of that assertion is to ensure logging occurs. I'm confirming that the right path is being exercised by asserting the data source type. For example, the `unpooled` setting should return a `DriverManagerDataSource`. Will not be testing the individual properties since I feel I obtained sufficient test coverage and it's definitely more than what we had before (which was none!). Feel free to follow-up on this PR if you don't think that's the case.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
claudemamo: Reverted some of the code.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -176,25 +178,19 @@ public ConfigKeyMapper getMapper() {
...
-
-    String msg =
-        String.format(
-            "Database pool type value is invalid, can not create a database pool! Value='%s'",
-            config.dbPoolType);
-    log.error(msg);
-
-    throw new IllegalArgumentException(msg);
+    return switch (dbType) {
+      case C3P0, UNPOOLED -> createC3p0DataSource(config);
```
david-mackessy, line 191: this invalidates a previously resolved comment.
if `UNPOOLED` is the chosen config we should really call the unpooled method (not have it buried in the C3P0 method). Currently, this is very misleading as we are not creating a C3PO pool.

david-mackessy: ah ok, i thought they were 3 separate, unrelated setups 👍
it needs to be really clear that they are both C3P0 so (the names don't convey this)
would a user know this from the config setting `unpooled`? (unless the docs explain this well)

jason-p-pickering: I thought we were getting rid of the now abandoned C3P0 library? Maybe its better to anticipate this now?

claudemamo: I've now replaced it with another un-pooled data source class from the Spring Framework.

jason-p-pickering (approved):
Looks good to me. I think this is much clearer in terms of looking at the code, and might help when we decide to get rid of the C3P0 pool.

PR #16488, opened 2024-02-12
fix(app): use correct authority for overidden bundled-apps

Birkbjo:
https://dhis2.atlassian.net/browse/DHIS2-16794

Testing a fix for above issue. The gist is that users without `ALL` or `M_dhis-web-app-management` won't be able to see overridden bundled-apps, causing the app to fall back to the bundled one.
This results in eg. super users seeing an updated app, while users without those authorities will see the original bundled version. 

### More details

Bundled apps have authorities like `M_dhis-web-data-visualizer`, while uploaded apps gets authorities like `M_datavisualizer` (due to the code in `getSeeAppAuthority`). So "overidden" bundled apps "falls between"-cases. Authorities generated by `getSeeAppAuthority` for "overridden" bundled apps are not part of systems authorities because "bundled" apps are filtered out in [`AppsSystemAuthoritiesProvider`](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/security/authority/AppsSystemAuthoritiesProvider.java#L53). Which is sane behaviour - we don't want duplicates of `M_dhis-web-data-visualizer` and `M_datavisualizer` authorities. If I understand things correctly, it seems like authorities for "bundled" apps are generated by `ModuleSystemAuthoritiesProvider`. 

The problems arise when we check if the user has access to the app - but now using the wrong authority (`M_datavisualizer`) that doesn't exist in the system.  

Note that bundled apps that are not overridden are using a [different method of checking access](https://github.com/dhis2/dhis2-core/blob/de98d605212216fe99a4d1da2ea3f132e3a4c275/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/security/config/DhisWebCommonsWebSecurityConfig.java#L206) And are thus not affected by this bug. 

The chain:

1. [`doFilterInternal`](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/AppOverrideFilter.java#L87) is called to check whether a bundled app has been overriden, if it is it would redirect to that source instead of the bundled source.
2. The app is bundled, thus `appManager.getApp` is called. 
3. `AppManager.getApp` tries to find the app, ultimately filtering using `isAccessable`
4. `isAccessible` check if user has `ALL, M_dhis-web-app-management`, or the authority from `app.getSeeAppAuthority`.
5. `getSeeAppAuthority` returns the wrong authority, causing `isAccessible` to return `false`. `appManager.getApp()` thus returns null.
6. Back in `doFilterInternal` the `app` is `null`, and thus the backend serves the original bundled app. 

I'm not sure this is the best way to fix it. I think it would be cleaner for these authorities to be in the `authorities`-HashSet in the app. But it doesn't seem like that is really used anymore? All these `bundled`-overrides seems a bit hacky and "bolted" on to me, but I guess this solution follows that 🤷


dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
```diff
@@ -439,6 +439,10 @@ public String getUrlFriendlyName() {
   }
 
   public String getSeeAppAuthority() {
+    if (isBundled()) {
+      return SEE_APP_AUTHORITY_PREFIX + AppManager.BUNDLED_APP_PREFIX + getShortName();
```
Birkbjo, line 443: Might want to call `getUrlFriendlyName` instead of short-name directly.

HendrikThePendric (approved):
I can confirm that this indeed solves the bug I found. A backend engineer should assess the code changes.

amcgee (approved):
Great work @Birkbjo and @HendrikThePendric !

This change looks correct to me, @stian-sandvold maybe you can take a quick look as well?  I think adding a unit test would be great.  It would be great to clean up a lot of this app serving code, but this is a great solution to the immediate problem with the code we have now.

We'll want to backport this to v38-40 at least as well, tagging @Philip-Larsen-Donnelly here as FYI

vietnguyen (approved):

PR #16489, opened 2024-02-12
feat: Remove the character "/" from the Org. Unit structure in response rows[DHIS2-16780]

d-bernat:
- The organisation unit name hierarchy format was changed. Instead of 
   /Sierra Leone/Bo/Kakua/Yemoh Town CHC is
   Sierra Leone / Bo / Kakua / Yemoh Town CHC required.
- The header names of Z-score and Modified Z-score were modified.


jason-p-pickering:
Hmm, I strongly suspect this is going to break a lot of external scripts which may depend on that leading slash. What is the reasoning here?

maikelarabori (approved):

gnespolino (approved):

PR #16490, opened 2024-02-12
refactor: Reuse method for analytics tables [DHIS2-16705]

larshelge:
* Reuses method for `JdbcEventAnalyticsTableManager`.
* Replaces `Collectors.toList()` with `toList()`.
* Adds `SqlBuilderProvider` for future use.
* Simplifies code in `AnalyticsDataSourceConfig`.
* Moves `PoolConfig` class to a public class file.


dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/model/PoolConfig.java
```diff
@@ -0,0 +1,69 @@
...
+import lombok.Data;
+import org.hisp.dhis.datasource.DatabasePoolUtils.ConfigKeyMapper;
+import org.hisp.dhis.external.conf.DhisConfigurationProvider;
+
+/**
+ * Encapsulation of a database connection pool configuration.
+ *
+ * @author Morten Svanæs <msvanaes@dhis2.org>
+ */
+@Data
```
jbee, line 41: A config usually is an immutable object so I'd go for `@Value` + `@Builder` instead

jbee (approved):

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/model/PoolConfig.java
larshelge: Nice. Fixed.

gnespolino (approved):

maikelarabori (approved):

PR #16494, opened 2024-02-13
DHIS2-8065:Require a specific authority to attribute data values

jason-p-pickering:
As detailed in DHIS2-8065, the current behavior of the system allows for `storedBy`, `created` and `lastUpdated` fields to be overridden by values contained in the data value set payload. This creates significant issues for accountability, auditing and debugging in systems who receive data via the `dataValueSets` endpoint. 

 This change in the API behavior will respect the current behavior for superusers who need to import data on behalf of other users or for other purposes. However, for non-superusers, the values specified for these fields in the payload will be ignored. The `storedBy` will be set to the current user name of the user importing the data. The `created` value will not be able to be be overridden. If the value is an existing value, the `created` date will be maintained, otherwise it will default to the current date. 
For `lastUpdated` this will always default to the current time regardless of what is specified in the payload for non-superusers.


jason-p-pickering:
Thinking more about this, it would seem we should introduce a separate authority "Allow data value attribution". This would allow users which have this authority to attribute either via the `dataValue` or `dataValueSets` end points the `storedBy`, `created` and `lastUpdated` fields of data values. Considering that this behavior has been in the code base for a very long time, its not really known how people are using it. However, by granting this authority (without granting super) this would allow what is essentially the current behavior. 

It is worth pointing out however that overriding the data value attribution is not possible via the `dataValue` endpoint, so the question might be whether we maintain this behavior or also allow overriding of these attributes via that point as well via the proposed authority.

PR #16495, opened 2024-02-13
fix: JasperReport failed to export pdf with arabic characters

vietnguyen:
- JasperReport replaced iText with OpenPDF from 6.20.1 as mentioned [here](https://github.com/TIBCOSoftware/jasperreports/releases/tag/6.20.1). This has caused an issue for PDF export with Arabic characters as reported in [DHIS2-16496](https://dhis2.atlassian.net/browse/DHIS2-16496)
- This PR revert it to 6.20.0 and add item to ignore list for dependabot. Created a [ticket](https://dhis2.atlassian.net/browse/DHIS2-16796) to do the upgrade later.
- Currently can't find a quick way to write unit test to check if the character in the pdf is translated, so include the test screenshot on my local 

Screen Shot 2024-02-13 at 6 44 19 PM

[DHIS2-16496]: https://dhis2.atlassian.net/browse/DHIS2-16496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbee (approved):

david-mackessy (approved):

andreasrosdal:
If this is caused by a bug in the OpenPDF library, please help by reporting this as a bug here so we can help solve it.

https://github.com/LibrePDF/OpenPDF

PR #16507, opened 2024-02-14
feat: Support new custom labels [DHIS2-16598]

maikelarabori:
There are new custom labels added recently, in the Maintenance app, that should be displayed in analytics headers if any of them are defined by users.
This PR will add support for these labels in the respective analytics endpoints.


d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/event/data/LabelMapperTest.java
```diff
@@ -47,10 +47,65 @@ class LabelMapperTest {
...
+  private static final String ORG_UNIT = "Organisation unit";
+
+  private static final String PROGRAM_STAGE = "Program stage";
+
+  private static final String ENROLLMENT = "Enrollment";
+
+  private static final String EVENT = "Event";
+
+  @Test
+  void testGetHeaderNameFor_EVENT() {
```
larshelge, line 61: Use standard camel-case for method names (`_event`).

maikelarabori: The idea here is to give a better context and make it clear that these tests are directly related to the CONSTANTS. In this case `EVENT`.

I can make it lowercase, but it will reduce the readability and at first sight, it will give the impression that this is testing an Event object, instead...

gnespolino (approved):

PR #16487, opened 2024-02-12
feat: add tracker endpoint to download attribute files DHIS2-16701

teleivo:
As for events https://github.com/dhis2/dhis2-core/pull/16446 add `GET "/tracker/trackedEntities/{trackedEntity}/attributes/{attribute}/file"`.

* extract serving of files from event controller into `FileResourceRequestHandler` for reuse in tracked entity controller
* implement finding file resource and its stream analogous to the implementation for events https://github.com/dhis2/dhis2-core/pull/16446

PR #16508, opened 2024-02-14
feat: feat: add tracker endpoint to download attribute image DHIS2-16763

teleivo:
As for events https://github.com/dhis2/dhis2-core/pull/16473 add `GET "/tracker/trackedEntities/{trackedEntity}/attributes/{attribute}/image"`

* endpoint was already excluded from Springs shallow etag filter in https://github.com/dhis2/dhis2-core/pull/16487
* contain the ugliness of us having to re-compute small, medium and large image md5 hashes in the `FileResourceStream` class. This also allows us to hide the fact that we need to resort to mutation 😅

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FileResourceStream.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FileResourceStream.java
@@ -27,12 +27,22 @@
  */
 package org.hisp.dhis.tracker.export;
 
+import com.google.common.hash.Hashing;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.util.NoSuchElementException;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
-import lombok.Setter;
+import org.hisp.dhis.feedback.BadRequestException;
+import org.hisp.dhis.feedback.ConflictException;
 import org.hisp.dhis.fileresource.FileResource;
+import org.hisp.dhis.fileresource.FileResourceService;
+import org.hisp.dhis.fileresource.ImageFileDimension;
+import org.hisp.dhis.util.ObjectUtils;
 
 /**
  * FileResourceStream holds a file resource and a supplier to open an input stream to the file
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FileResourceStream.java
```diff
@@ -43,5 +53,96 @@
...
+        () -> {
+          try {
+            return fileResourceService.openContentStream(fileResource);
+          } catch (NoSuchElementException e) {
+            // Note: we are assuming that the file resource is not available yet. The same approach
+            // is taken in other file endpoints or code relying on the storageStatus = PENDING.
+            // All we know for sure is the file resource is in the DB but not in the store.
+            throw new ConflictException(
+                "The content is being processed and is not available yet. Try again later.");
+          } catch (IOException e) {
```
lucaCambi77, line 72: Would it be worth moving those strings as constants? I see we have in quite a few places.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FileResourceStream.java
```diff
@@ -43,5 +53,96 @@
...
+        content = fileResourceService.copyImageContent(fileResource, imageDimension);
+        fileResourceStream.inputStreamSupplier = () -> new ByteArrayInputStream(content);
+      } catch (NoSuchElementException e) {
+        // Note: we are assuming that the file resource is not available yet. The same approach
+        // is taken in other file endpoints or code relying on the storageStatus = PENDING.
+        // All we know for sure is the file resource is in the DB but not in the store.
+        throw new ConflictException(
+            "The content is being processed and is not available yet. Try again later.");
+      } catch (IOException e) {
+        throw new ConflictException(
```
lucaCambi77, line 119: Is it also worth having unit tests for all these different ifs and exceptions that are thrown?

teleivo: These are tested via the controller tests 👌🏻

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FileResourceStream.java
teleivo: I don't want to adapt code other than tracker right now. For tracker they are contained in this class.

We will work on improving the serving of files in general across products after the next release 😄

muilpp (approved):

lucaCambi77 (approved):

PR #16513, opened 2024-02-14
feat: Add data values change log endpoint [DHIS2-16740]

muilpp:
First PR of the change log epic.

In this one I'm creating a new endpoint to get event changelog data, including:

- ACL validations: Checking sharing access to the event and the access to the event org unit.
- Building the response in the format agreed in the [design document](https://docs.google.com/document/d/1oij16GPVUUDxzjX2-hSssXenntiadZni7jKCEAcWdwM/edit?usp=sharing).
- New service and store to fetch the event changelog data. We need a new method because of the new format expected by frontend.
- Deprecate the old service now that we have a new one.

Sorting and paging are out of scope, they will be implemented in a future ticket.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/EventDataValueChangeLog.java
```diff
@@ -0,0 +1,45 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.changelog;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Date;
+import org.hisp.dhis.program.UserInfoSnapshot;
+
+public record EventDataValueChangeLog(
```
teleivo, line 34: Does this not represent an events changelog? In which case we would rather call it `EventChangeLog`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/EventDataValueChangeLog.java
```diff
@@ -0,0 +1,45 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.changelog;
```
teleivo, line 28: I think this should live in `org.hisp.dhis.tracker.export.changelog` as its specific to new tracker. We are trying to have all tracker related classes in as few modules/main packages as possible.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/ChangeLogStore.java
```diff
@@ -0,0 +1,41 @@
...
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.export.changelog;
+
+import java.util.List;
+import org.hisp.dhis.changelog.EventDataValueChangeLog;
+
+public interface ChangeLogStore {
```
teleivo, line 33: I don't think we need a store interface.

We only have one implementation.
We don't test the service separate from the store, so no need for a service for mocks.
The store implementation should be package private and only used in the ChangeLogService.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/JdbcChangeLogStore.java
```diff
@@ -0,0 +1,97 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.hisp.dhis.changelog.EventDataValueChangeLog;
+import org.hisp.dhis.changelog.EventDataValueChangeLog.Change;
+import org.hisp.dhis.program.UserInfoSnapshot;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.stereotype.Repository;
+
+@Slf4j
```
teleivo, line 40: No need to add slf4j if we don't log

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/JdbcChangeLogStore.java
```diff
@@ -0,0 +1,97 @@
...
+      };
+
+  @Override
+  public List<EventDataValueChangeLog> getEventChangeLog(String eventUid) {
+    final String sql =
+        """
+            select
+              case
+                when cl.audittype = 'CREATE' and cl.currentchangelogvalue is null then cl.previouschangelogvalue
+                when cl.audittype = 'CREATE' and cl.currentchangelogvalue is not null then cl.previouschangelogvalue
```
teleivo, line 72: There are two cases for CREATE here with an `and` with the two possible values of `cl.previouschangelogvalue` leading to the same result. Can we simplify this to 

```suggestion
                when cl.audittype = 'CREATE' then cl.previouschangelogvalue
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/JdbcChangeLogStore.java
```diff
@@ -0,0 +1,97 @@
...
+                when cl.audittype = 'UPDATE' and cl.currentchangelogvalue is null then cl.currentValue
+                when cl.audittype = 'UPDATE' and cl.currentchangelogvalue is not null then cl.currentchangelogvalue
+              end as currentValue,
+              case
+                when cl.audittype = 'DELETE' then cl.previouschangelogvalue
+                when cl.audittype = 'UPDATE' and cl.currentchangelogvalue is null then cl.previouschangelogvalue
+                when cl.audittype = 'UPDATE' and cl.currentchangelogvalue is not null then cl.previouschangelogvalue
+              end as previousValue, cl.created as updatedAt, cl.modifiedby as userName, cl.dataElementUid as dataElementUid, cl.firstname, cl.surname, cl.username, cl.useruid
+            from
+              (select t.created, d.uid as dataElementUid, t.modifiedby, t.audittype, u.firstname, u.surname, u.username, u.uid as useruid,
```
teleivo, line 82: please stick to lower case in all sql to stay consistent 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,174 @@
...
+        .flatMap(
+            e ->
+                e.getDataValues().stream()
+                    .filter(
+                        dv -> dv.getDataElement().getIdentifier().equalsIgnoreCase(dataElementUid))
+                    .findFirst())
+        .ifPresent(dv -> dv.setValue("new value"));
+    assertNoErrors(trackerImportService.importTracker(importParams, trackerObjects));
+
+    List<EventDataValueChangeLog> changeLogList =
```
teleivo, line 154:
```suggestion
    List<EventDataValueChangeLog> changeLogs =
```

the type is already in the type

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerByIdTest.java
```diff
@@ -687,6 +695,69 @@ private EventDataValue dataValue(DataElement de, String value) {
...
+
+    JsonWebMessage msg =
+        POST("/tracker?async=false&importStrategy=UPDATE", json)
+            .content(HttpStatus.OK)
+            .as(JsonWebMessage.class);
+
+    JsonEvent changeLogJson =
+        GET("/tracker/events/{id}/changelog", event.getUid())
+            .content(HttpStatus.OK)
+            .as(JsonEvent.class);
```
teleivo, line 756: That is not intentional right? We are not returning an event here 🤔

teleivo (commented):
One thing that came to mind was how we write changelog. We should make sure we are happy with the casing at least in what we publicly expose.

In our classes we use `ChangeLog` while in the API path we use 
`changelog` even though we use camel case in our paths. Not sure what we currently have in our JSON body.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerByIdTest.java
muilpp: yeah, don't pay too much attention to this test class... yet

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/ChangeLogStore.java
muilpp: I'm not sure I follow.
If we don't use an interface, what do you mean by the store implementation?

Are you suggesting we should create an instance of `JdbcChangeLogStore` manually, not managed by Spring?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/changelog/EventDataValueChangeLog.java
muilpp: I agree that it makes more sense to have everything related to tracker in the service tracker module. The issue is that when it comes to at least the exporters, we currently separate each class by event, enrollment, tracked entity, relationship, and now change log. However, we simply add classes to each package without considering the layer they belong to.

Is this just the starting point for cleaning things up? Are we planning to implement something else down the line, such as onion architecture?

muilpp:
> One thing that came to mind was how we write changelog. We should make sure we are happy with the casing at least in what we publicly expose.
> 
> In our classes we use `ChangeLog` while in the API path we use `changelog` even though we use camel case in our paths. Not sure what we currently have in our JSON body.

I'm using ChangeLog in our classes because of legacy reasons, but seems that the accepted term is change log, that's the reason why I'm using it in the API path

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentitydatavalue/TrackedEntityDataValueChangeLogService.java
```diff
@@ -39,6 +39,10 @@ public interface TrackedEntityDataValueChangeLogService {
   void addTrackedEntityDataValueChangeLog(
       TrackedEntityDataValueChangeLog trackedEntityDataValueChangeLog);
 
+  /**
+   * @deprecated use ChangeLogService.getEventChangeLog(UID) instead
```
teleivo, line 43: Can we use a `{@link to class}` in here? This way its easy to navigate and will easily be updated on refactorings.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/changelog/ChangeLogService.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hisp.dhis.feedback.NotFoundException;
+
+public interface ChangeLogService {
+
+  /**
+   * Retrieves the change log data for a particular event
+   *
+   * @return list with the change logs of the supplied event, if any
+   */
+  List<EventChangeLog> getEventChangeLog(UID eventUid) throws NotFoundException;
```
teleivo, line 41:
```suggestion
  List<EventChangeLog> getEventChangeLog(UID event) throws NotFoundException;
```

Lets get rid of type info in our var names where we can 😄

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+  @Test
+  void shouldFailWhenEventDoesNotExist() {
+    assertThrows(
+        NotFoundException.class,
+        () -> changeLogService.getEventChangeLog(UID.of(CodeGenerator.generateUid())));
+  }
+
+  @Test
+  void shouldFailWhenEventOrgUnitIsNotAccessible() {
+    injectSecurityContextUser(manager.get(User.class, "o1HMTIzBGo7"));
```
teleivo, line 88: A helper like

```suggestion
    testAsUser("o1HMTIzBGo7");
```

might be nice

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    assertThrows(
+        NotFoundException.class, () -> changeLogService.getEventChangeLog(UID.of("G9PbzJY8bJG")));
+  }
+
+  @Test
+  void shouldReturnChangeLogsWhenDataValueIsCreated() throws NotFoundException {
+    injectSecurityContextUser(manager.get(User.class, "M5zQapPyTZI"));
+
+    Event event = manager.get(Event.class, "QRYjLTiJTrA");
```
teleivo, line 122:
```java
    Event event = manager.get(Event.class, "QRYjLTiJTrA");
    assertNotNull(event);
```

could be turned into `getEvent("QRYjLTiJTrA")` or `assertEvent`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+    assertNotNull(event);
+    String dataElementUid = event.getEventDataValues().iterator().next().getDataElement();
+    DataElement dataElement = manager.get(DataElement.class, dataElementUid);
+    assertNotNull(dataElement);
```
teleivo, line 126: I think we should also hide the annoyance of working with the set here 😅 
like in a `getFirstDataValue(event)`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    TrackerObjects trackerObjects = fromJson("tracker/event_and_enrollment.json");
+    updateDataValue(trackerObjects, event.getUid(), dataElementUid, "");
```
teleivo, line 149: move 

    TrackerObjects trackerObjects = fromJson("tracker/event_and_enrollment.json");
    assertNoErrors(trackerImportService.importTracker(importParams, trackerObjects));

into `updateDataValue`

It will simplify its signature and make the test even nicer to read

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+    assertNoErrors(trackerImportService.importTracker(importParams, trackerObjects));
+
+    List<EventChangeLog> changeLogs = changeLogService.getEventChangeLog(UID.of("QRYjLTiJTrA"));
+
+    assertAll(
+        () -> {
+          validateNonValueElements(2, changeLogs, importUser, dataElementUid);
+          assertNull(changeLogs.get(0).change().dataValue().currentValue());
+          assertEquals("15", changeLogs.get(0).change().dataValue().previousValue());
```
teleivo, line 187: I might not fully get this case yet 😅 
The changelog 1 is for the initial create right?
The changelog 2 is for a delete and shows that no matter how often you delete a deleted value it will only lead to one change, right?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+    updateDataValue(trackerObjects, event.getUid(), dataElementUid, "");
+    assertNoErrors(trackerImportService.importTracker(importParams, trackerObjects));
+
+    List<EventChangeLog> changeLogs = changeLogService.getEventChangeLog(UID.of("QRYjLTiJTrA"));
+
+    assertAll(
+        () -> {
+          validateNonValueElements(3, changeLogs, importUser, dataElementUid);
+          assertNull(changeLogs.get(0).change().dataValue().currentValue());
+          assertEquals("20", changeLogs.get(0).change().dataValue().previousValue());
```
teleivo, line 274: I think an assertion like

`assertChange(change, prev, cur)`

would make this even better to read

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+                    .findFirst())
+        .ifPresent(dv -> dv.setValue(newValue));
+  }
+
+  private void validateNonValueElements(
+      int expectedListSize,
+      List<EventChangeLog> changeLogs,
+      User importUser,
+      String dataElementUid) {
+    assertEquals(expectedListSize, changeLogs.size());
```
teleivo, line 301: Since this is an integration tests we should be more careful with our assertions. Flaky tests in CI are annoying to fix with assertions that only state expected 3 got 2 😅 

Also you are using an `assertAll` where assertions that access an index lets say at 2 would cause an NPE in case the list only has 1 element.

You might need to rearrange your assertions.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/changelog/ChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    for (EventChangeLog eventChangeLog : changeLogs) {
+      assertEquals(importUser.getUsername(), eventChangeLog.updatedBy().getUsername());
```
teleivo, line 304: assertions like this should also be in an assertAll

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerPostgresTest.java
```diff
@@ -0,0 +1,253 @@
...
+            .as(JsonWebMessage.class);
+    assertEquals(HttpStatus.OK.toString(), importResponse.getStatus());
+
+    JsonWebMessage changeLogResponse =
+        GET("/tracker/events/{id}/changelog", event.getUid())
+            .content(HttpStatus.OK)
+            .as(JsonWebMessage.class);
+    assertEquals(1, changeLogResponse.asList(JsonList.class).size());
+
+    JsonObject changeLogObject = changeLogResponse.asList(JsonList.class).get(0).asObject();
```
teleivo, line 176: We usually create something like https://github.com/dhis2/dhis2-core/blob/b69cfc3a2245cba91ca4641711ffbbd782ed8b89/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/JsonEvent.java#L34

to make writing assertions easier.

teleivo (commented):
Very nice! We are already in the clean up/tweaking stage 🤩

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentitydatavalue/TrackedEntityDataValueChangeLogService.java
muilpp: That was my idea, but I couldn't find a way to link to a class in a different module, is that even possible?

teleivo: ah sorry, should be but not here as the dhis-api does not depend on service-tracker but the other way round

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventChangeLogStore.java
```diff
@@ -0,0 +1,92 @@
...
+              end as previousValue, cl.created as updatedAt, cl.modifiedby as userName, cl.dataElementUid as dataElementUid, cl.firstname, cl.surname, cl.username, cl.useruid
+            from
+              (select t.created, d.uid as dataElementUid, t.modifiedby, t.audittype, u.firstname, u.surname, u.username, u.uid as useruid,
+                  LAG (t.value) OVER (PARTITION BY t.eventid, t.dataelementid ORDER BY t.created DESC) AS currentchangelogvalue,
+                  t.value as previouschangelogvalue,
+                  e.eventdatavalues -> d.uid  ->> 'value' as currentValue
+              from trackedentitydatavalueaudit t
+              join event e using (eventid)
+              join dataelement d using (dataelementid)
+              join userinfo u on u.username = t.modifiedby
```
lucaCambi77, line 84: @muilpp I think the `trackedentitydatavalueaudit` table needs to change now and get a proper name like `trackedentitydatavaluechangelog`. This was agreed upon here https://dhis2.atlassian.net/browse/TECH-1659
so I think we should start doing these migrations

muilpp: The thing is this table needs a lot more changes than that. 
We should change the model and have previous value and current value in the same entry.
Then, we only store the current event data value in case it's created, if it's updated or deleted we store the previous only, so that should change too, but it also requires a migration.
Since we are tight on time, I think it's better to do these refactors/migrations after the release, so we can have a good plan on how to tackle all of this, Do you agree @lucaCambi77 @teleivo?

teleivo: I agree. We need more time to investigate all the consequences and what the ideal would be. I would therefore postpone any migrations of changelog related tables.

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    assertThrows(
+        NotFoundException.class,
+        () -> eventChangeLogService.getEventChangeLog(UID.of("H0PbzJY8bJG")));
+  }
+
+  @Test
+  void shouldFailWhenProgramWithoutRegistrationAndNoAccessToEventOrgUnit() {
+    injectSecurityContextUser(manager.get(User.class, "o1HMTIzBGo7"));
+    testAsUser("o1HMTIzBGo7");
```
lucaCambi77, line 115: code duplication

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerPostgresTest.java
```diff
@@ -0,0 +1,252 @@
...
+import org.hisp.dhis.web.HttpStatus;
+import org.hisp.dhis.webapi.DhisControllerIntegrationTest;
+import org.hisp.dhis.webapi.json.domain.JsonWebMessage;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+class EventsExportControllerPostgresTest extends DhisControllerIntegrationTest {
+  private static final String DATA_ELEMENT_VALUE = "value";
+  private static final String NEW_DATA_ELEMENT_VALUE = "new value";
```
lucaCambi77, line 66: I am not sure about this kind of testing, we already have tests in in the integration part so what we should do in MVC testing is maybe checking that we are invoking the service with the parameters we expect. An e2e test makes more sense. But that is my opinion.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventChangeLogStore.java
lucaCambi77: ok, sorry, I haven't followed that part. We should also get rid of the `ChangeLogType.READ` type and move that to the real audit i think or remove it completely. Currently we add to change log in `addTrackedEntityChangeLog` of `DefaultTrackedEntityService` for read operations

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerPostgresTest.java
muilpp: That's a good point, but in other tracker mock mvc tests we do this kind of testing too, plus the performance should be better here than in an e2e test, shouldn't it?
Maybe we can discuss in slack with the rest of the team?

teleivo: We had this discussion before. To me tests like this are less brittle, easier to write and faster to run while still giving a high degree of confidence that the feature is working.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventChangeLogStore.java
teleivo: we will, we just don't have the time right now. soft freeze in super close

lucaCambi77 (approved):

PR #16524, opened 2024-02-15
feat: UNLOGGED by default + bug introduced recently [DHIS2-16809]

maikelarabori:
This PR enables `UNLOGGED` tables by default in the tables generated during the Analytics Export process.

During my tests, related to this change, I discovered that some recent changes had broken the logic to generate the `UNLOGGED` tables. The code was not considering partitions anymore, which increased the export time by around 30% (on my laptop).

I fixed this issue, and now the export respects the flag accordingly, including the partition tables.
Some tests and assertions were added to prevent this from happening again in the future.


d-bernat (approved):

larshelge:
Thanks for fixing the logging setting on table partitions @maikelarabori .

PR #16546, opened 2024-02-19
feat: Outliers, Support skipRound [DHIS2-16837]

d-bernat:
The outliers API has to support skipRound query parameter. The default scale for rounding of outliers floating point type values (z-score, modified z-score, standard deviation, etc...)  is 1. When skipRound is true, scale is 10.


maikelarabori (approved):
Looks good 👍 
Would be nice to have at least one test with skip rounding set to true (unless I did not see it)

d-bernat:
> Looks good 👍 Would be nice to have at least one test with skip rounding set to true (unless I did not see it)

Done

gnespolino (approved):

PR #16558, opened 2024-02-20
feat: data sharing for AggregateDataExchange

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-16814
### Summary
- Enable data sharing for AggregateDataExchange.
- In endpoint `aggregateDataExchanges/{uid}/exchange` validate if user has data write permission for specified AggregateDataExchange object.
- Add flyway script to enable data sharing for existing records in database. This is done by copying Metadata Write to Data Write value.
- Add tests to verify all utils methods used in the migration script.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeJob.java
+++ b/dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeJob.java
@@ -40,6 +40,8 @@ import org.hisp.dhis.scheduling.JobType;
 import org.hisp.dhis.scheduling.parameters.AggregateDataExchangeJobParameters;
 import org.hisp.dhis.system.notification.NotificationLevel;
 import org.hisp.dhis.system.notification.Notifier;
+import org.hisp.dhis.user.UserDetails;
+import org.hisp.dhis.user.UserService;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeJob.java
```diff
@@ -65,6 +69,11 @@ public void execute(JobConfiguration config, JobProgress progress) {
     AggregateDataExchangeJobParameters params =
         (AggregateDataExchangeJobParameters) config.getJobParameters();
 
+    UserDetails currentUser =
+        config.getExecutedBy() != null
+            ? UserDetails.fromUser(userService.getUser(config.getExecutedBy()))
+            : CurrentUserUtil.getCurrentUserDetails();
```
jbee, line 75: Should be just `CurrentUserUtil.getCurrentUserDetails()` - if the `executedBy` user is set at job construction, that user will be the current user when running the `execute` method of the job. 

Note that if no user is set at construction this is the `SuperUser`!

jbee (changes requested):

dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeJob.java
vietnguyen: Done

larshelge (approved):

netroms (approved):

jbee (approved):

PR #16559, opened 2024-02-20
feat: Require a specific authority to attribute data values (DHIS2-8065)

jason-p-pickering:


stian-sandvold (changes requested):
Thanks for the PR Jason!

I don't think this would be a good solution for the problem though. I think we should not introduce superuser/authorities to the mix, and we shouldn't let created/lastUpdated be changed at all. I added a new comment to the Jira issue - If you want you can ping me and we can have a call about what solution we should go for.

PR #16562, opened 2024-02-20
feat(DHIS2-16132): add ability to customise the display of a section form

kabaros:
This PR adds configuration option to the Section metadata in order to be able to display a form transposed, and grouped by data elements in the data entry app. Implements [DHIS2-16132 ](https://dhis2.atlassian.net/browse/DHIS2-16132).

## Implementation details

- The `displayOptions` was added as a JSON type to allow us flexibility to add more options without updating the backend. See discussion here.
  - The JSON object only supports these properties: `pivotMode` and `pivotedCategory` (this is a FE concern though, irrelevant to this PR) 
- [This line](https://github.com/dhis2/dhis2-core/pull/16562/files#diff-3f4dd79ed04bd98558f59e7b97ec192556f92cbe19bbae558a60e6516696cbeaR88) was added to  `DefaultDataSetMetadataExportService` as otherwise the `dataEntry/metadata` eTag is not updated when a section is updated and the new configuration is not shown to the user (unless the server is restarted?)
- The change includes a migration to create the new columns (applied in v41)

image

## Background

We are aiming to add more form configuration options as part of an initiative to provide configurations natively to data entry forms to reduce the necessity for custom forms. Users are currently building custom forms as a workaround for shortcomings of the configuration options (ability to transpose, or customise a cell design) or implementation (such to avoid issues with RTL issues).

This is [an RFC](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/121995265/Form+configuration+RFC) that describes the approach and the priorities for form configuration options. This is based on a [thorough investigation](https://docs.google.com/presentation/d/1pSeFoF91HDYNeL88GghKLvj3wpYf0qUCdhkKXklB2MY/edit#slide=id.g2462c2dd6c9_0_53) by the functional design team for custom form use cases in real-life implementations. Based on that investigation, the ability to transpose forms and adding free-text descriptions were one of the main reasons people choose to go the custom forms route so we're tackling these first.

### UI

[form-config-4.webm](https://github.com/dhis2/dhis2-core/assets/1014725/2774b230-52a3-4f20-bfde-63ec544c35e4)

## Related PRs:

- [Maintenance app](https://github.com/dhis2/maintenance-app/pull/2802)
- [Data entry app (beta)](https://github.com/dhis2/aggregate-data-entry-app/pull/367)
- [Backend](https://github.com/dhis2/dhis2-core/pull/16562)

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/Section.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/Section.java
@@ -33,10 +33,8 @@ import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import org.hisp.dhis.category.CategoryCombo;
 import org.hisp.dhis.common.BaseIdentifiableObject;
@@ -67,11 +65,7 @@ public class Section extends BaseIdentifiableObject implements MetadataObject {
 
   private boolean disableDataElementAutoGroup;
 
-  private String textBeforeSection;
-
-  private String textAfterSection;
-
-  private Map<String, String> displayOptions = new HashMap<>();
+  private String displayOptions;
 
   // -------------------------------------------------------------------------
   // Constructors
@@ -224,31 +218,11 @@ public class Section extends BaseIdentifiableObject implements MetadataObject {
 
   @JsonProperty
   @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
-  public String getTextBeforeSection() {
-    return textBeforeSection;
-  }
-
-  public void setTextBeforeSection(String textBeforeSection) {
-    this.textBeforeSection = textBeforeSection;
-  }
-
-  @JsonProperty
-  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
-  public String getTextAfterSection() {
-    return textAfterSection;
-  }
-
-  public void setTextAfterSection(String textAfterSection) {
-    this.textAfterSection = textAfterSection;
-  }
-
-  @JsonProperty
-  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
-  public Map<String, String> getDisplayOptions() {
+  public String getDisplayOptions() {
     return displayOptions;
   }
 
-  public void setDisplayOptions(Map<String, String> displayOptions) {
+  public void setDisplayOptions(String displayOptions) {
     this.displayOptions = displayOptions;
   }
 
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/Section.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/Section.java
@@ -33,8 +33,10 @@ import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import org.hisp.dhis.category.CategoryCombo;
 import org.hisp.dhis.common.BaseIdentifiableObject;
@@ -64,6 +66,12 @@ public class Section extends BaseIdentifiableObject implements MetadataObject {
   private boolean showColumnTotals;
 
   private boolean disableDataElementAutoGroup;
+
+  private String textBeforeSection;
+
+  private String textAfterSection;
+
+  private Map<String, String> displayOptions = new HashMap<>();
 
   // -------------------------------------------------------------------------
   // Constructors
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_45__add_section_config_fields.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE section ADD COLUMN IF NOT EXISTS transposesection BOOLEAN DEFAULT FALSE;
+ALTER table section add column if not exists transposedgroups jsonb default '[]'::jsonb;
```
kabaros, line 2: @david-mackessy - I am working on this functionality to add extra configuration options to the Section metadata (see PR for more details). For this field `transposedGroups` , we want to be able to configure 1+ `categories` to be transposed/pivoted when rendering the section form on the Frontend.

I modeled it as a soft link with a `json` field. This felt like the simplest thing to do, and could provide some flexibility with this field (i.e. we might want to transpose data elements too in the future). I even thought the four fields maybe can be one json field, to avoid updating the backend as we experiment with more configuration options if possible. This means that we could have dangling categories when they're deleted but in this context (configuring forms), the FE is very permissive and would fall back to showing the categories that exist similar to what would happen if this was cascade deleted in a relationship.

In the codebase, I see this pattern used in few places (but mostly analytics) but also sometimes we're normalising with a proper relationship (i.e. greyFields). So I am wondering what's the right way to go ahead with this?

(ps: this PR is missing other local files, but just want to discuss this specific point first)

cc' @Birkbjo @stian-sandvold

david-mackessy: Hi @kabaros, using jsonb will provide you with a lot of flexibility. You can update & query the jsonb model as you please without needing further DB changes. Without knowing the requirements or how large the jsonb may become, just try to cover different scenarios so you know you're happy with performance if the model becomes very large.

It sounds like you're looking for some kind of dynamic table effect, speaking with Analytics (Maikel, Giusseppe, Dusan) may provide useful also.

We use jsonb in 2 different ways (AFAIK):
- As a user type (see the `JsonBinaryType` in our codebase and which classes use it and how it's used,`Sharing` and `Translation` are examples)
- As a plain string value (`DataStore` jbPlainValue is an example)

You probably have to choose which one `transposedGroups` should be. I don't have enough experience with these to suggest which would be better or easier to use.

I'll tag @jbee as well to see if he has any suggestions or concerns.

kabaros: Thanks @david-mackessy for the pointers, I added the field as jsonb, and mapped it with `jbObject` (similar to its use in systemSettings and MapView) - this provides the flexibility for the FE to define any JSON payload without having to update the backend.

kabaros: it'd be great if you can help me with solving the sonar issue - not sure why it's happening as the same pattern exists in [MapView](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/mapping/MapView.java#L197) and a couple other places.

kabaros: I fixed the sonar issue, by adding `transient` to the `displayOptions` per the suggestion from Sonar itself. Initially I thought this would mean that the property is not serialised in the JSON but that doesn't seem to be the case and it works as expected - let me know if that potentially has other implications.

david-mackessy (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_45__add_section_config_fields.sql
david-mackessy: I'm not sure why a map is being used. What will the map keys be? the json root fields? If so, it feels a bit messy doing it like this.
I think you could get what you're looking for with string. The `DatastoreEntry` field `jbPlainValue` is another example of using string to store in a jsonb column.
If you're happy with what's there & it satisfies your requirements then 👍

kabaros: the json I am sending looks like

```json
{
"pivotMode": "x",
"pivotedCategory": "XXX"
}
```

is that not suitable to be modeled as a map of strings to strings? We 'd be adding more properties (these are the keys), but they will always be strings keys to strings values. I will look at the `jbbPlainValue` as well though.

david-mackessy: ok i see. As you mentioned earlier, it looks like you're settling on a middle ground rather than having the full flexibility. But if you know this will only ever be a straight string -> string mapping then fair enough.

jbee:
I feel we should discuss adding multiple fields for styling information. Once you have 1 JSONB column for this it should be all that is needed. I assume nobody would want to filter on these information so having extra columns seems complicating the backend model unnecessary. Any visual/style concerns should be bundled in a single "style" JSONB column. This not just saves us columns now but also in the future when more or different styling information is wanted by the app it can just change what JSON is stored and the backend does not need to change at all. In any case I feel having a chat for this kind of change would be the best way to proceed.

jbee (commented):
I think using individual fields is not a good way to model style concerns when JSON(B) is available.

stian-sandvold:
> I feel we should discuss adding multiple fields for styling information. Once you have 1 JSONB column for this it should be all that is needed. I assume nobody would want to filter on these information so having extra columns seems complicating the backend model unnecessary. Any visual/style concerns should be bundled in a single "style" JSONB column. This not just saves us columns now but also in the future when more or different styling information is wanted by the app it can just change what JSON is stored and the backend does not need to change at all. In any case I feel having a chat for this kind of change would be the best way to proceed.

Good point Jan. I do think in this case having the text as separate fields makes sense, as it's not really "styling" per se, rather it's information to be displayed in the forms that might need or require translations in the future. I also think we should discuss (in the future) how we might see styling/display-related properties actually being related to metadata should be done. Currently, I think it's a bit messy having all kinds of different properties in the same object. Hiding it all in a style jsonb object is one solution, but I think we should dig deeper into that later!

stian-sandvold (approved):

kabaros:
@jbee @david-mackessy @stian-sandvold - After discussing with JB, I updated the PR to:

- get rid of the extra text columns, and only have the one JSONB column.
- change the mapped type from a Hashmap to a string - this is consistent with other use cases, but is inconvenient from a FE perspective as we always have to parse the string returned into a JSON object when receiving, and stringify it before sending. So maybe that's something we can keep in mind for the future.

Let me know if it looks good for you.

jbee (approved):

david-mackessy (approved):

PR #16563, opened 2024-02-20
feat: Add pagination on new change log endpoint [DHIS2-16856]

muilpp:
* inline fields from `common.Pager` into `org.hisp.dhis.tracker.export.Page` as there is no benefit for us in using `common.Pager` internally. The `common.Pager` does not support not returning totals which we have to.
* add `org.hisp.dhis.tracker.export.Page.with` instance method to copy a Page using a different list of items. This makes it easier in controllers to return the `fields` filtered list based on the same `Page` returned from a service
* use `LongSupplier` for counts as total is represented using a `long`
* return `prevPage` and `nextPage` in change log endpoint. The links are absolute as this is what we do for metadata.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
@@ -28,40 +28,42 @@
 package org.hisp.dhis.tracker.export;
 
 import java.util.List;
+import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
 @Getter
 @ToString
 @EqualsAndHashCode
 public class Page<T> {
   private final List<T> items;
-  private final Integer page;
-  private final Integer pageSize;
+  private final int page;
+  private final int pageSize;
   private final Long total;
-  private final boolean prev;
-  private final boolean next;
+  private final Boolean prev;
+  private final Boolean next;
 
-  private Page(List<T> items, int page, int pageSize, long total, boolean prev, boolean next) {
-    this.items = items;
-    this.page = page;
-    this.pageSize = pageSize;
-    this.total = total;
-    this.prev = prev;
-    this.next = next;
+  /**
+   * Create a new page based on an existing one but with given {@code items}. Page related counts
+   * will not be changed so make sure the given {@code items} match the previous page size.
+   */
+  public <U> Page<U> withItems(List<U> items) {
+    return new Page<>(items, this.page, this.pageSize, this.total, this.prev, this.next);
   }
 
   public static <T> Page<T> withTotals(List<T> items, int page, int pageSize, long total) {
-    return new Page<>(items, page, pageSize, total, false, false);
+    return new Page<>(items, page, pageSize, total, null, null);
   }
 
   public static <T> Page<T> withoutTotals(List<T> items, int page, int pageSize) {
-    return new Page<>(items, page, pageSize, 0, false, false);
+    return new Page<>(items, page, pageSize, null, null, null);
   }
 
   public static <T> Page<T> withPrevAndNext(
       List<T> items, int page, int pageSize, boolean prev, boolean next) {
-    return new Page<>(items, page, pageSize, 0, prev, next);
+    return new Page<>(items, page, pageSize, null, prev, next);
   }
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
@@ -30,25 +30,38 @@ package org.hisp.dhis.tracker.export;
 import java.util.List;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
 import lombok.ToString;
-import org.hisp.dhis.common.Pager;
 
-@RequiredArgsConstructor(staticName = "of")
 @Getter
 @ToString
 @EqualsAndHashCode
 public class Page<T> {
   private final List<T> items;
-  private final Pager pager;
+  private final Integer page;
+  private final Integer pageSize;
+  private final Long total;
+  private final boolean prev;
+  private final boolean next;
 
-  /**
-   * Indicates if the {@link #pager} contains the total number of items. The current {@link #pager}
-   * does not reveal that information as it was meant for metadata that does not allow disabling the
-   * return of totals.
-   */
-  private final boolean pageTotal;
+  private Page(List<T> items, int page, int pageSize, long total, boolean prev, boolean next) {
+    this.items = items;
+    this.page = page;
+    this.pageSize = pageSize;
+    this.total = total;
+    this.prev = prev;
+    this.next = next;
+  }
 
-  private final boolean hasPrevious;
-  private final boolean hasNext;
+  public static <T> Page<T> withTotals(List<T> items, int page, int pageSize, long total) {
+    return new Page<>(items, page, pageSize, total, false, false);
+  }
+
+  public static <T> Page<T> withoutTotals(List<T> items, int page, int pageSize) {
+    return new Page<>(items, page, pageSize, 0, false, false);
+  }
+
+  public static <T> Page<T> withPrevAndNext(
+      List<T> items, int page, int pageSize, boolean prev, boolean next) {
+    return new Page<>(items, page, pageSize, 0, prev, next);
+  }
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
@@ -48,4 +48,7 @@ public class Page<T> {
    * return of totals.
    */
   private final boolean pageTotal;
+
+  private final boolean hasPrevious;
+  private final boolean hasNext;
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Date;
 import org.hisp.dhis.program.UserInfoSnapshot;
@@ -40,6 +41,6 @@ public record EventChangeLog(
 
   public record DataValueChange(
       @JsonProperty String dataElement,
-      @JsonProperty String previousValue,
-      @JsonProperty String currentValue) {}
+      @JsonInclude @JsonProperty String previousValue,
+      @JsonInclude @JsonProperty String currentValue) {}
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Date;
 import org.hisp.dhis.program.UserInfoSnapshot;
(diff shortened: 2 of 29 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/Page.java
```diff
@@ -48,4 +48,7 @@ public class Page<T> {
    * return of totals.
    */
   private final boolean pageTotal;
+
+  private final boolean hasPrevious;
```
muilpp, line 52: Need to add javadoc here

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
```diff
@@ -40,6 +41,6 @@ public record Change(@JsonProperty DataValueChange dataValue) {}
 
   public record DataValueChange(
       @JsonProperty String dataElement,
-      @JsonProperty String previousValue,
-      @JsonProperty String currentValue) {}
+      @JsonInclude @JsonProperty String previousValue,
```
teleivo, line 44: Why is `@JsonInclude` needed in addition to `@JsonProperty`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -27,16 +27,17 @@
...
+import org.hisp.dhis.tracker.export.Page;
+import org.hisp.dhis.tracker.export.PageParams;
 
 public interface EventChangeLogService {
 
   /**
    * Retrieves the change log data for a particular event
    *
-   * @return list with the change logs of the supplied event, if any
+   * @return paginated list of change logs of the supplied event, if any
```
teleivo, line 40:
```suggestion
   * @return event change logs page
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
```diff
@@ -139,5 +152,7 @@ public static class Pager {
     @JsonProperty private Integer pageSize;
     @JsonProperty private Long total;
     @JsonProperty private Integer pageCount;
+    @JsonProperty private String prev;
```
teleivo, line 155: The names have to match the https://github.com/dhis2/dhis2-core/blob/40e950822f2c1346518958022fc22fe61e07f959/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Pager.java#L47-L49

`https://play.im.dhis2.org/dev/api/organisationUnits?page=2&pageSize=2&fields=displayName`

```json
{
  "pager": {
    "page": 2,
    "total": 1332,
    "pageSize": 2,
    "nextPage": "https://play.im.dhis2.org/dev/api/organisationUnits?page=3&pageSize=2&fields=displayName",
    "prevPage": "https://play.im.dhis2.org/dev/api/organisationUnits?pageSize=2&fields=displayName",
    "pageCount": 666
  },
  "organisationUnits": [
    {
      "displayName": "Agape CHP"
    },
    {
      "displayName": "Ahamadyya Mission Cl"
    }
  ]
}
```

We could add a unit test to compare JSON produced by common.Pager and our view.Page as a safety-net until we can reuse the common.Pager.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/ChangeLogRequestParams.java
```diff
@@ -0,0 +1,52 @@
...
+@OpenApi.Shared(name = "ChangeLogRequestParams")
+@OpenApi.Property
+@Data
+@NoArgsConstructor
+public class ChangeLogRequestParams {
+
+  static final String DEFAULT_FIELDS_PARAM = "createdBy, change, createdAt";
+
+  @OpenApi.Property(defaultValue = "1")
+  private Integer page;
```
teleivo, line 46: The way you wrote it here is for endpoints that allow turning off pagination. We do not allow that for changelogs.

So it should rather be

```suggestion
  private int page = 1;
```

with no need for the extra `@OpenApi.Property` on the field the default will be picked up.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/ChangeLogRequestParams.java
```diff
@@ -0,0 +1,52 @@
...
+import org.hisp.dhis.fieldfiltering.FieldFilterParser;
+import org.hisp.dhis.fieldfiltering.FieldPath;
+
+@OpenApi.Shared(name = "ChangeLogRequestParams")
+@OpenApi.Property
+@Data
+@NoArgsConstructor
+public class ChangeLogRequestParams {
+
+  static final String DEFAULT_FIELDS_PARAM = "createdBy, change, createdAt";
```
teleivo, line 43:
```suggestion
  static final String DEFAULT_FIELDS_PARAM = "change,createdAt,createdBy";
```

this value reflects what users would usually send in their request which will not have spaces

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
```diff
@@ -296,9 +296,40 @@ ResponseEntity<InputStreamResource> getEventDataValueImage(
   }
 
   @GetMapping("/{uid}/changeLog")
-  List<EventChangeLog> getEventByUid(@OpenApi.Param({UID.class, Event.class}) @PathVariable UID uid)
+  Page<ObjectNode> getEventByUid(
```
teleivo, line 299:
```suggestion
  Page<ObjectNode> getEventChangeLogByUid(
```

OpenAPI cannot handle overloaded methods and this method is not returning events.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
```diff
@@ -296,9 +296,40 @@ ResponseEntity<InputStreamResource> getEventDataValueImage(
...
+    org.hisp.dhis.tracker.export.Page<EventChangeLog> changeLogs =
+        eventChangeLogService.getEventChangeLog(uid, pageParams);
+
+    if (changeLogs.isHasNext()) {
+      changeLogs
+          .getPager()
+          .setNextPage(
+              request.getRequestURI()
+                  + String.format(
+                      "?page=%d&pageSize=%d",
```
teleivo, line 317: This is insufficient if we provide non-pagination parameters like `fields`.

If you try 

`https://play.im.dhis2.org/dev/api/organisationUnits?page=2&pageSize=2&fields=displayName`

you see that the query parameters are preserved

```json
{
  "pager": {
    "page": 2,
    "total": 1332,
    "pageSize": 2,
    "nextPage": "https://play.im.dhis2.org/dev/api/organisationUnits?page=3&pageSize=2&fields=displayName",
    "prevPage": "https://play.im.dhis2.org/dev/api/organisationUnits?pageSize=2&fields=displayName",
    "pageCount": 666
  },
  "organisationUnits": [
    {
      "displayName": "Agape CHP"
    },
    {
      "displayName": "Ahamadyya Mission Cl"
    }
  ]
}
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
muilpp: to see json elements with null values, if not they are skipped

teleivo: This is different to what we usually do 🤔 Did we decide on whether we would add a `type` field for `create`, `update`, `delete`?

muilpp: it's not required by design, front end didn't answer me yet on whether they'd like to see it while testing.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramNotificationInstanceController.java
```diff
@@ -126,7 +126,10 @@ public ProgramNotificationInstanceController(
       return Page.withPager(
           ProgramNotificationInstanceSchemaDescriptor.PLURAL,
           instances,
-          org.hisp.dhis.tracker.export.Page.of(instances, pager, true, false, false));
+          org.hisp.dhis.tracker.export.Page.withTotals(
+              instances, pager.getPage(), pager.getPageSize(), (int) pager.getTotal()),
+          null,
+          null);
```
teleivo, line 132: why do we need to pass in `null`? I thought our API in `export.Page` does not require that.

* without totals: no totals and no prev/next
* with totals: total/pageCount but no prev/next
* with prev and next: no totals

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/ChangeLogRequestParams.java
```diff
@@ -27,26 +27,28 @@
...
-  private Integer page;
+  private int page = 1;
 
-  @OpenApi.Property(defaultValue = "50")
-  private Integer pageSize;
+  private int pageSize = 50;
 
   private List<FieldPath> fields = FieldFilterParser.parse(DEFAULT_FIELDS_PARAM);
+
+  private List<OrderCriteria> order = new ArrayList<>();
```
teleivo, line 53: why are you adding order? Let's get pagination done first without order if possible

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java
```diff
@@ -308,28 +308,25 @@ Page<ObjectNode> getEventChangeLogByUid(
...
+          request.getRequestURI()
+              + String.format(
+                  "?page=%d&pageSize=%d", requestParams.getPage() - 1, requestParams.getPageSize());
     }
 
     List<ObjectNode> objectNodes =
         fieldFilterService.toObjectNodes(changeLogs.getItems(), requestParams.getFields());
 
-    return Page.withPager("changeLogs", objectNodes, changeLogs);
+    return Page.withPager("changeLogs", objectNodes, changeLogs, previousPage, nextPage);
```
teleivo, line 330: I think we can create a simpler API that takes care of doing the prev/nextPage transformation transparently. This way every controller will get that for free 🥳 .

Lets use the `export.Page.with` instance method described before in a service comment.

The reason we have to pass in `objectNodes` is that we filtered the `changeLogs.items`. So we cannot use the `changeLogs.items` in `Page.withPager`.

It could look like this in the end 😋 

```java
return Page.withPager("changeLogs", changeLogs.with(objectNodes), request);
```

We pass
* the JSON key under which the `changeLogs.items` should be returned
* the `export.Page` (here we reuse the pagination information but only replace the items as they were filtered; without resorting to mutations)
* and the request to fill in the `prev/nextPage` if `changeLogs.isPrev/isNext` is true

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
```diff
@@ -87,60 +87,56 @@ public class Page<T> {
...
+        next);
   }
 
   /**
    * Returns a page which will only serialize the items into {@link #items} under given {@code key}.
    * All other fields will be omitted from the JSON.
    */
   public static <T> Page<T> withoutPager(String key, List<T> items) {
-    return new Page<>(key, items, null, false);
+    return new Page<>(key, items, 0, 0, null, 0, null, null);
```
teleivo, line 139: For fields to be omitted from the JSOn they have to be null. See javadoc for the purpose of this factory.

teleivo (commented):
We should validate the pagination params in the controller. We do that in the exporters using https://github.com/dhis2/dhis2-core/blob/f689bf2f717610a8a793f84167343dac84861234/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java#L406 We might only need to validate their range as this endpoint does not support turning pagination off/counts on/off.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
teleivo: We decided not return the keys if the values are null as its consistent with what we do.

teleivo (approved):

lucaCambi77:
One way to get rid of the null pointer for spring `queryForObject` could be to use a call back handler:

```
...

    RowCountHandler countHandler = new RowCountHandler();

    jdbcTemplate.query(sql, mapSqlParameterSource, countHandler);

    return countHandler.getCount();
  }

  class RowCountHandler implements RowCallbackHandler {
    int count;

    @Override
    public void processRow(ResultSet rs) throws SQLException {
      count = rs.getInt("count(*)");
    }

    public int getCount() {
      return count;
    }
  }
```

lucaCambi77 (approved):

PR #16567, opened 2024-02-21
feat: Implement configurable labels for certain DHIS2 terminology [Tracker] [More labels]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16326

Some labels have already been added, see https://github.com/dhis2/dhis2-core/pull/16288.

However, there is a new requirement for using labels in the program context during Android testing when we don't need to configure or refer to a specific program stage. Eventually, Android will use the label for the program stage only if it exists; otherwise, it will use the one in the program context.

### Note
Again, this is a temporary solution and is considered a technical debt. This feature will be removed in the future


teleivo (approved):

david-mackessy (approved):
I'm not familiar with the concept of labels on metadata or its usage. From a code standpoint it looks fine.

PR #16570, opened 2024-02-21
chore: custom icon refactor [DHIS2-16872]

zubaira:
[DHIS2-16872](https://dhis2.atlassian.net/browse/DHIS2-16872)

[DHIS2-16872]: https://dhis2.atlassian.net/browse/DHIS2-16872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- Move `DefaultIcon` to the database with `custom` set to `false`
- Add `custom` flag
- Make `keywords` `jsonb` object

IconController
- Added endpoint for create/update/delete/get

CustomIcon
- Removed `String[] keywords` with `Set<String> keywords`
- Added `createdBy/lastUpdatedBy` and `created/lasupdated`

Services/Store
- Removed `JdbcCustomIconStore` as now we have `HibernateCustomIconStore`

Migration
- Added flyway migration script to put all bundled/default icons into the database


dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconOperationParams.java
```diff
@@ -35,21 +35,21 @@
...
 /**
- * Represents query parameters sent to {@link org.hisp.dhis.icon.IconService}.
+ * Represents query parameters sent to {@link org.hisp.dhis.icon.CustomIcon}.
  *
  * @author Zubair Asghar
  */
 @Data
 @NoArgsConstructor
-public class IconOperationParams {
+public class CustomIconOperationParams {
```
jbee, line 44: _Operation_ is not a pattern we use and if we would use if for manipulation, not querying.
How about `IconQueryParams` ?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconOperationParams.java
```diff
@@ -76,4 +76,8 @@ public boolean hasKeywords() {
   public boolean hasKeys() {
     return !keys.isEmpty();
   }
+
+  public boolean hasCustom() {
```
jbee, line 80: This is an easily misunderstood name. Either find a more communicating name or add a javadoc explaining what this means.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconService.java
```diff
@@ -63,14 +62,23 @@ public interface IconService {
...
   CustomIcon getCustomIcon(String key) throws NotFoundException;
 
+  /**
+   * Gets the custom icon associated to a uid, if it exists
+   *
+   * @param uid key of the icon to find
+   * @return custom icon associated to the uid, if found
+   * @throws NotFoundException if no custom icon exists with the provided uid
+   */
+  CustomIcon getCustomIconByUid(String uid) throws NotFoundException;
```
jbee, line 72: I thought we were clear on icons not having a UID as it collides with the `key`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
```diff
@@ -55,170 +48,172 @@
...
-            customIconStore.getIcons(iconOperationParams))
-        .collect(Collectors.toList());
+  public Set<CustomIcon> getCustomIcons(CustomIconOperationParams iconOperationParams) {
+    return customIconStore.getCustomIcons(iconOperationParams);
   }
 
   @Override
   @Transactional(readOnly = true)
-  public long count(IconOperationParams iconOperationParams) {
+  public long count(CustomIconOperationParams iconOperationParams) {
```
jbee, line 70: Please avoid names like `iconOperationParams` that only repeat the type name. Long names like that add no new information to the code. Either use a short name like `params` or a name that adds new information, like `filters`, `criteria`, `restrictions` or alike.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
```diff
@@ -55,170 +48,172 @@
...
 
     return customIcon;
   }
 
   @Override
-  public Resource getDefaultIconResource(String key) throws NotFoundException {
-    if (defaultIcons.containsKey(key)) {
-      return new ClassPathResource(
-          String.format("%s/%s.%s", ICON_PATH, key, DefaultIcon.Icons.SUFFIX));
+  @Transactional
```
jbee, line 97: Sounds to me like a `readOnly` TX

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
```diff
@@ -55,170 +48,172 @@
  * @author Kristian Wærstad
  */
 @RequiredArgsConstructor
-@Service("org.hisp.dhis.icon.IconService")
-public class DefaultIconService implements IconService {
+@Service("org.hisp.dhis.icon.CustomIconService")
+public class DefaultCustomIconService implements CustomIconService {
+
+  private static final String CUSTOM_ICON_KEY_PATTERN = "^[a-zA-Z0-9*-_+]+$";
```
jbee, line 54: If we have a pattern anyhow should we define a max length? Instead of `+` do a `{1,50}` or something? At the very least the `key` column will (or should) have a max length we can assume here.

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_47__Add_default_icons_into_database.java
```diff
@@ -0,0 +1,112 @@
...
+      """
+      INSERT INTO public.customicon(
+      id, uid, code, iconkey, fileresourceid, description, keywords, created, lastupdated, createdby, lastupdatedby, custom)
+      VALUES (nextval('hibernate_sequence'), generate_uid(), ?, ?, null, ?, ?, now(), now(), null, null, false );
+      """;
+
+  @Override
+  public void migrate(Context context) throws Exception {
+
+    Map<String, CustomIcon> customIconMap =
```
jbee, line 69: This is not a valid way of doing this as a number of things can change in this base data. The `Icon.values()` may change, the variants may change. A script needs to be constant. Better you use the code to print out the statements to console and then copy them to a SQL flyway script so they never change.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_custom_icon.sql
```diff
@@ -1,22 +1,28 @@
...
-    created        TIMESTAMP    NULL DEFAULT CURRENT_TIMESTAMP,
-    createdby      bigint       NOT NULL,
-    CONSTRAINT customicon_pkey PRIMARY KEY (customiconid),
-    CONSTRAINT customicon_ukey UNIQUE ("key"),
+    id                  bigint         NOT NULL,
+    uid                 character      varying(11) NOT NULL,
+    code                character      varying(50),
+    iconkey             varchar(100)   NOT NULL,
+    fileresourceid      bigint         NULL,
+    description         text           NOT NULL,
```
jbee, line 12: Can/should we force that all icons have a description?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_custom_icon.sql
```diff
@@ -1,22 +1,28 @@
...
-    keywords       text[]       NULL,
-    created        TIMESTAMP    NULL DEFAULT CURRENT_TIMESTAMP,
-    createdby      bigint       NOT NULL,
-    CONSTRAINT customicon_pkey PRIMARY KEY (customiconid),
-    CONSTRAINT customicon_ukey UNIQUE ("key"),
+    id                  bigint         NOT NULL,
+    uid                 character      varying(11) NOT NULL,
+    code                character      varying(50),
+    iconkey             varchar(100)   NOT NULL,
+    fileresourceid      bigint         NULL,
```
jbee, line 11: Shouldn't all icons have a file resource?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_custom_icon.sql
```diff
@@ -1,22 +1,28 @@
...
-    customiconid   int8 GENERATED ALWAYS AS IDENTITY,
-    "key"          varchar(100) NOT NULL,
-    fileresourceid int8         NOT NULL,
-    description    text         NULL,
-    keywords       text[]       NULL,
-    created        TIMESTAMP    NULL DEFAULT CURRENT_TIMESTAMP,
-    createdby      bigint       NOT NULL,
-    CONSTRAINT customicon_pkey PRIMARY KEY (customiconid),
-    CONSTRAINT customicon_ukey UNIQUE ("key"),
+    id                  bigint         NOT NULL,
```
jbee, line 7: This should be `<tablename>id`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/CustomIconTest.java
```diff
@@ -0,0 +1,282 @@
...
+
+  private final Set<String> keywords = new HashSet<>();
+  private final String Key = "iconKey";
+  private final String uid = CodeGenerator.generateUid();
+
+  private CustomIcon customIcon;
+  private FileResource fileResource;
+  private User currentUser;
+
+  @SneakyThrows
```
jbee, line 76: Don't use `@SneakyThrows`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconMapper.java
```diff
@@ -35,71 +35,78 @@
...
+    return customIcon;
   }
 
-  public CustomIcon to(IconDto iconDto) throws BadRequestException {
-    Optional<FileResource> fileResource =
-        fileResourceService.getFileResource(iconDto.getFileResourceUid(), CUSTOM_ICON);
-    if (fileResource.isEmpty()) {
-      throw new BadRequestException(
-          String.format("File resource with uid %s does not exist", iconDto.getFileResourceUid()));
+  public void merge(CustomIcon persisted, CustomIconRequest customIconRequest)
```
jbee, line 75: I don't think this is a good way to implement updates. Better pass down the request object to the store layer and build SQL from it. That will avoid needing the file resource (or any other non updated) property. Service layer should be responsible for validating the update based on the request.

jbee (commented):
The naming should be consistent and as icons are both default (provided) and custom (user uploaded) I think the name/prefix should be `Icon` and not `CustomIcon`. Also in DB I would use `icon` table and in controller `/icons` path. 

I thought we talked about not making it a `BaseIdentifiableObject` and use SQL to implement the query? I feel like the requirements have changed again without me being in the loop.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_custom_icon.sql
```diff
@@ -1,22 +1,28 @@
...
+    code                character      varying(50),
+    iconkey             varchar(100)   NOT NULL,
+    fileresourceid      bigint         NULL,
+    description         text           NOT NULL,
+    keywords            jsonb          NULL,
+    created             TIMESTAMP      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    lastupdated         TIMESTAMP      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    createdby           bigint         NULL,
+    lastupdatedby       bigint         NULL,
+    custom              Boolean        NOT NULL,
```
teleivo, line 18:
```suggestion
    custom              boolean        NOT NULL,
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/CustomIconTest.java
zubaira: fixed

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/CustomIconTest.java
```diff
@@ -0,0 +1,282 @@
...
+package org.hisp.dhis.icon;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import com.google.common.collect.Sets;
```
teleivo, line 37: please use java.util.Set

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/CustomIconRequestParams.java
```diff
@@ -42,19 +42,18 @@
...
   private List<String> keys = new ArrayList<>();
   private List<String> keywords = new ArrayList<>();
-  private IconTypeFilter type = IconTypeFilter.ALL;
   private Date createdStartDate;
   private Date createdEndDate;
   private Date lastUpdatedStartDate;
   private Date lastUpdatedEndDate;
-  private boolean paging = true;
+  private IconTypeFilter type = IconTypeFilter.ALL;
+  private boolean paging = false;
```
teleivo, line 56: did we settle on paging being off by default?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_custom_icon.sql
zubaira: Yes all need file but customicon needs a reference to that external file via fileResource. Whereas bundled/default icon does not need that.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -37,177 +37,225 @@
...
+      throws IOException, WebMessageException {
 
-    return new PaginatedIconResponse(pager, objectNodes);
+    if (!iconService.customIconExists(key)) {
+      throw new WebMessageException(
+          WebMessageUtils.notFound(String.format("CustomIcon with key %s not found", key)));
+    }
+
+    String location = response.encodeRedirectURL("/icons/" + key + "/icon");
+    response.sendRedirect(ContextUtils.getRootPath(request) + location);
```
teleivo, line 184: Why do we redirect?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconMapper.java
```diff
@@ -35,71 +35,78 @@
...
+      }
     }
+
+    CustomIcon customIcon = new CustomIcon();
+    customIcon.setKey(customIconRequest.getKey());
+    customIcon.setDescription(customIconRequest.getDescription());
+    customIcon.setKeywords(customIconRequest.getKeywords());
+    customIcon.setCustom(customIconRequest.getCustom());
+    customIcon.setCode(customIconRequest.getCode());
+    customIcon.setFileResource(fileResource.orElse(null));
```
teleivo, line 70: Should this ever be null?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconService.java
zubaira: i think we should keep it if we plan to make custom icon a metadata object and for sake of consistency as well.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/CustomIconRequest.java
```diff
@@ -28,20 +28,27 @@
...
+
   @JsonProperty private String description;
 
-  @JsonProperty private String[] keywords;
+  @JsonProperty private Set<String> keywords;
+
+  @JsonProperty private String fileResourceId;
 
-  @JsonProperty private String fileResourceUid;
+  @Builder.Default @JsonProperty private Boolean custom = true;
```
teleivo, line 53: Is this only used to create/update custom icons? if so we would not need this field

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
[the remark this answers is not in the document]
teleivo: @zubaira did you fix this alert?

zubaira:
> The naming should be consistent and as icons are both default (provided) and custom (user uploaded) I think the name/prefix should be `Icon` and not `CustomIcon`. Also in DB I would use `icon` table and in controller `/icons` path.
> 
> I thought we talked about not making it a `BaseIdentifiableObject` and use SQL to implement the query? I feel like the requirements have changed again without me being in the loop.

i think it was decided NOT to extend `AbstractReadOnlyController` and implement the filtering part.
Secondly i really don't see any harm in making it `BaseIdentifiableObject` if we have a plan to make it metadata object eventually.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconOperationParams.java
zubaira: fixed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
zubaira: fixed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultCustomIconService.java
```diff
@@ -55,170 +48,172 @@
...
-    CustomIcon icon = validateCustomIconExists(key);
-    getFileResource(icon.getFileResourceUid()).setAssigned(false);
-    customIconStore.delete(key);
+  private void validateCustomIconKey(String key) throws BadRequestException {
+    Matcher matcher = pattern.matcher(key.trim());
+
+    if (!matcher.matches()) {
+      throw new BadRequestException(
+          String.format(
+              "CustomIcon key %s is not valid. Alphanumeric and special characters are allowed",
```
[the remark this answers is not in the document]
zubaira: we already have keys with `-` and `_`.  May i can change it with some example key name

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: https://dhis2.atlassian.net/browse/DHIS2-16041

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconMapper.java
zubaira: yes. if there is a change in default icons. no?

teleivo: can default icons be changed by a user via our API?

zubaira: description or addition/removal from keywords. These are cases where default icons can be updated.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconService.java
jbee: Hm... my last understanding was that icons are a bad fit for classic metadata and making them look like metadata is more pain than gain.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/CustomIconRequest.java
jbee: This is so we can distinguish user provided (custom) icons from bundled icons later assuming they all are in the same table(s)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/CustomIconService.java
zubaira: fixed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/CustomIconRequest.java
teleivo: Sure but I this class is used in the controller for POST addIcon and PUT updateIcon. I assume a default icon cannot be created or updated, right? If that is the case it does not make sense to add a flag. It would then also be clearer if this is called CustomIconRequest.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -57,141 +53,139 @@
...
 
   @Override
   @Transactional
-  public void addCustomIcon(CustomIcon customIcon) throws BadRequestException, NotFoundException {
-    validateIconDoesNotExists(customIcon.getKey());
-    FileResource fileResource = getFileResource(customIcon.getFileResourceUid());
-    UserDetails currentUserDetails = CurrentUserUtil.getCurrentUserDetails();
-    customIconStore.save(customIcon, fileResource, currentUserDetails);
-  }
+  public void addIcon(Icon icon) throws BadRequestException, NotFoundException, SQLException {
```
teleivo, line 111: Please add a `javax.annotation.Nonnull` to the `icon` parameter on the interface and implementation.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -57,141 +53,139 @@
...
+
+    if (Boolean.TRUE.equals(icon.getCustom())) {
+      FileResource fileResource = getFileResource(icon.getFileResource().getUid());
+      fileResource.setAssigned(true);
+      fileResourceService.updateFileResource(fileResource);
     }
 
-    if (description != null) {
-      icon.setDescription(description);
+    if (icon.getCreatedBy() == null) {
```
teleivo, line 126: Could a user set this? Just wondering if we should always set it no matter if its null or not like with `created`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -57,141 +53,139 @@
...
+    iconStore.update(icon);
   }
 
   @Override
   @Transactional
-  public void deleteCustomIcon(String key) throws BadRequestException, NotFoundException {
-    CustomIcon icon = validateCustomIconExists(key);
-    getFileResource(icon.getFileResourceUid()).setAssigned(false);
-    customIconStore.delete(key);
+  public void deleteIcon(Icon icon) throws BadRequestException, NotFoundException {
```
teleivo, line 158: Would `deleteIcon(String key)` simplify things? I imagine that is all we need to identify the icon.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -57,141 +53,139 @@
...
+      FileResource fileResource = getFileResource(persistedIcon.getFileResource().getUid());
+      fileResource.setAssigned(false);
+      fileResourceService.updateFileResource(fileResource);
+    }
+
+    iconStore.delete(persistedIcon);
+  }
+
+  private void validateIconKey(String key) throws BadRequestException {
+    Matcher matcher = pattern.matcher(key.trim());
```
teleivo, line 172: Do/should we also trim the key before persisting? Or maybe simpler reject a key with whitespace.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -206,27 +200,35 @@ private void validateIconKeyNotNullOrEmpty(String key) throws BadRequestExceptio
...
-  private CustomIcon validateCustomIconExists(String key)
-      throws NotFoundException, BadRequestException {
+  private Icon validateIconExists(String key) throws NotFoundException, BadRequestException {
     validateIconKeyNotNullOrEmpty(key);
-    return getCustomIcon(key);
+    return getIcon(key);
+  }
+
+  private Icon validateIconExists(Icon icon) throws NotFoundException, BadRequestException {
+    if (icon == null) {
```
teleivo, line 228: There are a bunch of these methods that look like we are repeating ourselves. Can this not be `!validateIconExists(icon)`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,279 @@
...
+            });
+
+    return icons.isEmpty() ? null : icons.get(0);
+  }
+
+  @Override
+  public void save(Icon icon) throws SQLException {
+
+    String sql =
+        "INSERT INTO icon (iconkey,description,keywords,fileresourceid,createdby,created,lastupdated,custom) VALUES (:key,:description,cast(:keywords as jsonb),:fileresourceid,:createdby,now(),now(),:custom) ";
```
teleivo, line 117: Why are we using `now()` here? are we not using `setAutoFields()` which already sets these fields? Same in the update

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,279 @@
...
+      sql =
+          getPaginatedQuery(
+              params.getPager().getPage(), params.getPager().getPageSize(), sql, parameterSource);
+    }
+
+    return namedParameterJdbcTemplate
+        .query(
+            sql,
+            parameterSource,
+            (rs, rowNum) -> {
```
teleivo, line 181: Can we reuse this mapper? I think this one is duplicated

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,279 @@
...
+      parameterSource.addValue("keys", params.getKeys());
+    }
+
+    return sql;
+  }
+
+  private String getPaginatedQuery(
+      int page, int pageSize, String sql, MapSqlParameterSource mapSqlParameterSource) {
+
+    sql = sql + " LIMIT :limit OFFSET :offset ";
```
teleivo, line 252: Do we have an `order by` clause in the sql? if not we should add one for it to be deterministic.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,279 @@
...
+    mapSqlParameterSource.addValue("offset", offset);
+
+    return sql;
+  }
+
+  private Set<String> convertKeywordsJsonIntoSet(String jsonString) {
+    try {
+      return keywordsMapper.readValue(jsonString, new TypeReference<Set<String>>() {});
+    } catch (IOException e) {
+      log.error("Parsing keywords json failed, string value: '{}'", jsonString);
```
teleivo, line 266: You can log the exception https://www.baeldung.com/slf4j-log-exceptions

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_47__Add_default_icons_into_database.java
```diff
@@ -0,0 +1,110 @@
...
+import org.hisp.dhis.icon.DefaultIcon;
+import org.hisp.dhis.icon.Icon;
+import org.postgresql.util.PGobject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Zubair Asghar
+ */
+public class V2_41_47__Add_default_icons_into_database extends BaseJavaMigration {
```
teleivo, line 51: Not sure where @jbee comment went on using this code to generate a `.sql` migration file. We need the migration to be an immutable file. If one day we need to make a change to `DefaultIcon` we will need to write another SQL migration and update existing data or insert new data.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_icon.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE IF NOT EXISTS icon
+(
+    iconkey             varchar(100)   NOT NULL,
+    fileresourceid      bigint         NULL,
+    description         text           NULL,
+    keywords            jsonb          NULL,
+    created             TIMESTAMP      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    lastupdated         TIMESTAMP      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    createdby           bigint         NULL,
+    custom              Boolean        NOT NULL,
```
teleivo, line 14:
```suggestion
    custom              boolean        NOT NULL,
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -57,210 +52,169 @@
...
-    iconService.addCustomIcon(
-        new CustomIcon(
-            "iconKey", "description", keywords, fileResource.getUid(), getCurrentUser().getUid()));
+    keywords.addAll(Set.of("k1", "k2", "m1"));
+    fileResource = createAndPersistFileResource('A');
+    icon = new Icon(Key, "description", keywords, true, fileResource);
+
+    try {
+      iconService.addIcon(icon);
+    } catch (NotFoundException | BadRequestException | SQLException e) {
```
teleivo, line 89: Why are you catching this? the test should fail if the setup fails.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,88 +148,115 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
+      throws NotFoundException, WebMessageException {
 
-    Pager pager = null;
+    Icon icon = iconService.getIcon(key);
 
-    if (iconRequestParams.isPaging()) {
-      long total = iconService.count(iconOperationParams);
-      pager = new Pager(iconRequestParams.getPage(), total, iconRequestParams.getPageSize());
-      iconOperationParams.setPager(pager);
+    if (icon == null) {
```
teleivo, line 164: `getIcon` should throw a `NotFoundException` if the icon is not found. This is directly translated by the `CrudControllerAdvice`. So you don't need to do anything in this case.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,88 +148,115 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
 
-    WebMessage webMessage = new WebMessage(Status.OK, HttpStatus.OK);
-    webMessage.setMessage(String.format("Icon %s deleted", iconKey));
-    return webMessage;
+    if (icon == null) {
+      throw new WebMessageException(
+          WebMessageUtils.notFound(String.format("Icon with key %s not found", key)));
+    }
+
+    iconService.deleteIcon(icon);
```
teleivo, line 239: I think this should be simplified to `deleteIcon(key)`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
```diff
@@ -28,573 +28,514 @@
...
-  private final String[] keywords;
+  public static final String SUFFIX = "svg";
+
+  @Getter private final String key;
 
-  private Date created;
+  @Getter private final String description;
 
-  private Date lastUpdated;
+  private final String[] keywords;
```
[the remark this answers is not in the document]
zubaira: this is what we had in bundled icon. These will be in the database so does not matter if older one has `List` or `Array`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
zubaira: because it's convenient. Does it pose any problem?

jbee (changes requested):
Looks pretty good!
Some suggestions and some smaller points to address left.
The main thing I am confused about is why custom and default icons are treated any different for e.g. download. As I understand it, after the bundled icons are inserted by the script they can and should be treated the same except for maybe disallowing update and delete.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
```diff
@@ -164,6 +164,11 @@ protected void postProcessResponseEntity(
    */
   protected void forceFiltering(final WebOptions webOptions, final List<String> filters) {}
 
+  /** To override default fields if they are not provided. */
+  protected void getFields(List<String> fields) {
+    fields.addAll(Preset.defaultPreset().getFields());
+  }
+
```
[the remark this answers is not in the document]
zubaira: i dont know why you still see this but this was reverted

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,88 +148,115 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
+  @Deprecated
+  public void getIconData(
+      HttpServletResponse response, HttpServletRequest request, @PathVariable String key)
+      throws IOException, WebMessageException {
 
-    return new PaginatedIconResponse(pager, objectNodes);
-  }
+    if (!iconService.iconExists(key)) {
+      throw new WebMessageException(
+          WebMessageUtils.notFound(String.format("Icon with key %s not found", key)));
```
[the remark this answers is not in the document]
zubaira: fixed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -126,88 +148,115 @@ public void getIconData(@PathVariable String key, HttpServletResponse response)
...
+          BadRequestException,
+          SQLException {
 
-    WebMessage webMessage = new WebMessage(Status.OK, HttpStatus.OK);
-    webMessage.setMessage(String.format("Icon %s updated", iconDto.getKey()));
-    return webMessage;
+    Icon persisted = iconService.getIcon(key);
+
+    if (persisted == null) {
+      throw new WebMessageException(
```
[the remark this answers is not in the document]
zubaira: fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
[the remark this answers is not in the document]
teleivo: Well it does matter if you actually want a Set when working with this type. See your getKeywords which transforms this. This is unnecessary if all you want is a Set.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/Icon.java
```diff
@@ -27,16 +27,90 @@
...
+    }
+
+    Icon other = (Icon) obj;
+    return Objects.equals(key, other.key);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(key);
+  }
```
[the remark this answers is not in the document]
zubaira: fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -29,6 +29,7 @@
 
 import java.sql.SQLException;
 import java.util.Set;
+import lombok.NonNull;
```
teleivo: that is the wrong one see https://github.com/dhis2/wow-backend/blob/59d3a443c755080a0da845251ca5db2d07dcd4a0/guides/preferred_patterns.md?plain=1#L27

zubaira: i am using `javax.annotation`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
zubaira: fixed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -98,7 +98,10 @@ public Icon getIconByKey(String key) {
   @Override
   public void save(Icon icon) throws SQLException {
     String sql =
-        "INSERT INTO icon (iconkey,description,keywords,fileresourceid,createdby,created,lastupdated,custom) VALUES (:key,:description,cast(:keywords as jsonb),:fileresourceid,:createdby,now(),now(),true) ";
+        """
+            INSERT INTO icon (iconkey,description,keywords,fileresourceid,createdby,created,lastupdated,custom)
+            VALUES (:key,:description,cast(:keywords as jsonb),:fileresourceid,:createdby,now(),now(),true)
```
teleivo, line 103: I commented on this somewhere, why are you using `now()`? We are setting created/lastUpdated by `setAutoFields` in the service.

zubaira: I think yes and I did reply that. 
using `now()` is convenient and does not pose any problem

teleivo: It is odd to do that no? Why are we setting it in Java while we are not even using it since we override it in the SQL. @jbee are we doing this elsewhere? I imagine this can lead to some headaches as its surprising.

zubaira: What sort of headache?

jbee: We are using this in a few places when we are not using hibernate managed insert and updates. There should not be an issue doing this. You can get odd data when DB "now" is set to another timezone than the DHIS2 server but if that is the case we have a lot of issues :confused:

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconService.java
```diff
@@ -37,31 +37,31 @@
...
+   * Get list of Icons based on filters provided in {@link IconOperationParams}
    *
-   * @param key key of the icon to find
-   * @return icon associated to the key, if found
-   * @throws NotFoundException if no icon exists in the database with the provided key
+   * @param params filters to build query
+   * @return list of Icons
    */
-  Icon getIcon(String key) throws NotFoundException;
+  Set<Icon> getIcons(IconOperationParams params);
```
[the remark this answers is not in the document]
jbee: Good point, I missed this but I agree, `List` should be the more fitting and future proof option even if order is mostly not relevant for icons.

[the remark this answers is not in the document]
zubaira: Well that was the initial thinking behind using Set.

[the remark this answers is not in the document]
zubaira: Will change it though.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -102,6 +104,26 @@ void shouldGetIconDataWhenKeyBelongsToIcon() throws NotFoundException {
...
+    icon2.setKey("facility-agent");
+
+    iconService.addIcon(icon1);
+    iconService.addIcon(icon2);
+
+    IconOperationParams operationParams = new IconOperationParams();
+    operationParams.setSearch("agent");
+    List<Icon> icons = iconService.getIcons(operationParams);
+
+    assertThat(icons.size(), is(2));
```
teleivo, line 123: if order matters please use `assertEquals(List.of(), actual)` else use our helper `assertContainsOnly` from org.hisp.dhis.utils.Assertions. See https://github.com/dhis2/wow-backend/blob/59d3a443c755080a0da845251ca5db2d07dcd4a0/guides/testing/testing.md?plain=1#L109

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,271 @@
...
+    if (params.hasKeys()) {
+      sql += hlp.whereAnd() + " c.iconkey IN (:keys )";
+
+      parameterSource.addValue("keys", params.getKeys());
+    }
+
+    if (params.hasSearch()) {
+      String searchValue = params.getSearch();
+
+      sql += hlp.whereAnd() + " c.iconkey ilike :search";
```
teleivo, line 219: Where did this requirement come from 🤔 ?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
teleivo: Please remove `assertThat(icons.size(), is(2));` as this will already be caught by `assertContainsOnly`. The way it is right now the test would fail with `expected 2 got 3` or so. We want to avoid such assertion errors in CI as it does not tell you enough to find the issue. That is why we use `assertContainsOnly` or `assertEquals(List.of(), actual)`. These two already assert on the size but print what's in the collection in case the assertion fails.

teleivo (commented):
Please address the outstanding comments. Especially the migration+fileresourceid not having a not null constraint.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
zubaira: may be @Birkbjo  can explain more. The client side needs `ilike` search for both keys and keywords. But for now we are only doing it for keys.

zubaira:
> Please address the outstanding comments. Especially the migration+fileresourceid not having a not null constraint.

because we don't have any files to upload for existing default icons.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconOperationParams.java
```diff
@@ -76,4 +76,8 @@ public boolean hasKeywords() {
   public boolean hasKeys() {
     return !keys.isEmpty();
   }
+
+  public boolean hasCustom() {
```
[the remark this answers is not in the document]
zubaira: renamed to includeCustomIcon

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
zubaira: yes we do before persisting

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.41/V2_41_2__Create_table_icon.sql
```diff
@@ -0,0 +1,27 @@
+
+
+-- icon table creation
+
+CREATE TABLE IF NOT EXISTS icon
+(
+    iconkey             varchar(100)   NOT NULL,
+    fileresourceid      bigint         NULL,
```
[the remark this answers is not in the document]
zubaira: yes they are different

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
zubaira: fixed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
zubaira: fixed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
Birkbjo: So this comes from the fact that icons are now paginated.
Previously searching icons are done on the client. Because key and keywords filter only support full text matching, which is kind of useless for searching.

So for the client to be able to use this we kind of need a search. And this should've searched keywords as well, but that's a jsonb column and the function that trivially does that requires postgres13.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
teleivo: Why do we trim here? Can we not take the users key at face value when they want to get/delete the icon?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/icon/IconTest.java
```diff
@@ -60,150 +57,157 @@
...
+    operationParams.setSearch("agent");
     List<Icon> icons = iconService.getIcons(operationParams);
 
-    assertEquals(
-        defaultIconMap.size() + 1,
-        iconService.getIcons(operationParams).size(),
-        String.format(
-            "Expected to find %d icons, but found %d instead: %s",
-            defaultIconMap.size() + 1, icons.size(), icons));
+    Assertions.assertContainsOnly(List.of(icon1, icon2), icons);
```
teleivo, line 120: Please use a static import as we do for all assertions

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractFullReadOnlyController.java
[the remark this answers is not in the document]
teleivo: I also see this on Github. I checked out the branch locally and these changes are in so you still need to revert them.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
teleivo: This is also not fixed. I am still seeing this on the branch when checking out the PR.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconRequestParams.java
```diff
@@ -44,16 +45,17 @@
...
 public class IconRequestParams {
 
-  static final String DEFAULT_FIELDS_PARAM =
-      "key,keywords,description,fileResourceUid,createdByUserUid,href";
+  static final String DEFAULT_FIELDS_PARAM = "key,keywords,description,fileResource,createdBy,href";
 
   private List<String> keys = new ArrayList<>();
   private List<String> keywords = new ArrayList<>();
-  private IconTypeFilter type = IconTypeFilter.ALL;
+  private List<OrderCriteria> order = new ArrayList<>();
```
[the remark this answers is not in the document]
zubaira: Why not?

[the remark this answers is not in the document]
zubaira: Giving flexibility to order based on different parameters

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -94,102 +96,135 @@ public class IconController {
...
+    linkService.generatePagerLinks(pager, Icon.class);
+
+    return new PaginatedIconResponse(pager, objectNodes);
   }
 
   @GetMapping(value = "/{key}/icon")
   public void getIconData(@PathVariable String key, HttpServletResponse response)
       throws NotFoundException, WebMessageException, IOException {
-
     Icon icon = iconService.getIcon(key);
```
Birkbjo, line 146: This route is hit every time an icon is requested as a static asset. Is this something we should consider performance wise? Could it be considered bad-performance wise to instantiate and grab all the data from the DB, when we really only need to check for the existence of `key` and get the `fileResource`. Or is this not a problem?

@teleivo

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -0,0 +1,304 @@
...
+    if (params.hasKeys()) {
+      sql += hlp.whereAnd() + " c.iconkey IN (:keys )";
+
+      parameterSource.addValue("keys", params.getKeys());
+    }
+
+    if (params.hasSearch()) {
+      String searchValue = params.getSearch();
+
+      sql += hlp.whereAnd() + "(c.iconkey ilike :search or c.keywords #>> '{}' ilike :search)";
```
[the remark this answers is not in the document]
zubaira: 👍

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
```diff
@@ -222,7 +222,7 @@ public WebMessage updateIcon(@PathVariable String key, HttpServletRequest reques
   public WebMessage deleteIcon(@PathVariable String key)
       throws NotFoundException, BadRequestException {
 
-    iconService.deleteIcon(key);
+    iconService.deleteIcon(key.trim());
```
teleivo, line 225: Why are you doing that? Can we not expect users to give us the exact key they want to delete/search for?

I would only validate on insert that they key is valid i.e. has on whitespace.

zubaira: handling unintended whitespace in user input

teleivo: Whitespace has to be percent encoded. If a user goes to that length it would result in a 404 as the key does not exist.

If you try adding spaces in a URL with a tool like curl will yell at you

```
curl 'http://admin:district@localhost:8080/api/tracker/events/ FV4JCI73wO2'
curl: (3) URL using bad/illegal format or missing URL
```

if you percent encode it you will get a 400 as in this particular case its expected to be a UID

```
curl 'http://admin:district@localhost:8080/api/tracker/events/%20FV4JCI73wO2'
{"httpStatus":"Bad Request","httpStatusCode":400,"status":"ERROR","message":"Value ' FV4JCI73wO2' is not valid for path parameter uid. UID must be an alphanumeric string of 11 characters starting with a letter."}
```

If you select the above url and right-click go in Chrome it will percent encode the space and you also get a 400.

We should not add code that does not add any benefit.

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -69,16 +77,51 @@ public class DefaultIconService implements IconService {
...
+
+            i.setFileResource(fileResource);
+
+            addIconInternal(i);
+          } catch (NotFoundException
+              | ConflictException
+              | IOException
+              | SQLException
+              | BadRequestException e) {
+            e.printStackTrace();
```
teleivo: Please log.error instead of printing the stack trace.

I am not sure when printing the stack trace like this would be appropriate. You want users to be able to configure what and how things get logged. This way they cannot.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -244,4 +277,37 @@ private Icon validateIconExists(String key) throws NotFoundException, BadRequest
     validateIconKeyNotNullOrEmpty(key);
     return getIcon(key);
   }
+
+  private List<Icon> createIconWithVariants(DefaultIcon defaultIcon) {
+
```
teleivo, line 254: remove these newlines please 😢

teleivo (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
zubaira: i think this code is outdated.

jbee:
@zubaira I converted the PR to draft as I know you are working on it but it had 2 approvals and might be merged accidentally otherwise

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/icon/IconController.java
teleivo: Icons have a max age of 1 year. The icons are served from the browsers cache if I open capture on dev. So unless they are not updated (or removed from the cache) icons should not hit the backend more than once / year / user.

PR #16574, opened 2024-02-21
fix: get TEA via program attributes else TET attributes DHIS2-16859

teleivo:
* `/api/tracker/trackedEntities/vOxUH373fy5/attributes/nDlikr3TUS6/image?program={programUid}`: validate ACL against TE and program + find attribute in program attributes
* `/api/tracker/trackedEntities/vOxUH373fy5/attributes/nDlikr3TUS6/image`: validate ACL against TE + find attributes in tracked entity type attributes of the TE

* simplify creating attributes and values (tracked entity attribute value -> tracked entity type attribute|program attribute -> tracked entity attribute)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityAttributeService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityAttributeService.java
@@ -303,10 +303,17 @@ public class DefaultTrackedEntityAttributeService implements TrackedEntityAttrib
 
   @Override
   @Transactional(readOnly = true)
-  public Set<TrackedEntityAttribute> getAllUserReadableTrackedEntityAttributes(Program program) {
-    List<TrackedEntityType> trackedEntityTypes = trackedEntityTypeService.getAllTrackedEntityType();
+  public Set<TrackedEntityAttribute> getProgramAttributes(Program program) {
     return getAllUserReadableTrackedEntityAttributes(
-        CurrentUserUtil.getCurrentUserDetails(), List.of(program), trackedEntityTypes);
+        CurrentUserUtil.getCurrentUserDetails(), List.of(program), List.of());
+  }
+
+  @Override
+  @Transactional(readOnly = true)
+  public Set<TrackedEntityAttribute> getTrackedEntityTypeAttributes(
+      TrackedEntityType trackedEntityType) {
+    return getAllUserReadableTrackedEntityAttributes(
+        CurrentUserUtil.getCurrentUserDetails(), List.of(), List.of(trackedEntityType));
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


muilpp:
nice to have a clear separation between program attributes and tracked entity attributes!
is there a reason not to follow the test naming convention `shouldXXwhenYY`?

muilpp (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityAttributeService.java
```diff
@@ -303,10 +303,17 @@ public Set<TrackedEntityAttribute> getAllUserReadableTrackedEntityAttributes(
...
+  }
+
+  @Override
+  @Transactional(readOnly = true)
+  public Set<TrackedEntityAttribute> getTrackedEntityTypeAttributes(
+      TrackedEntityType trackedEntityType) {
     return getAllUserReadableTrackedEntityAttributes(
-        CurrentUserUtil.getCurrentUserDetails(), List.of(program), trackedEntityTypes);
+        CurrentUserUtil.getCurrentUserDetails(), List.of(), List.of(trackedEntityType));
   }
```
lucaCambi77, line 317: If this runs with no security context the user can be null, but maybe it is not the case here

teleivo: At least for now its only run in sync as its only used in the GET of file/images.

I am going to bring this up in the backend meeting though. I think `CurrentUserUtil.getCurrentUserDetails()` should throw a checked exception as most of the time you expect the user to be there and anything else is an exception. We could then add a method that is used in the rare cases where you want to deal with the user not being set.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -173,22 +156,56 @@ private FileResource getFileResourceMetadata(
...
-  private TrackedEntityAttribute validateAttributeDataReadAccess(UID attributeUid, Program program)
+  /**
+   * Tracked entity attributes are fetched from the program if supplied, otherwise from the tracked
+   * entity. Access is determined through the program sharing and ownership if present. If no
+   * program is supplied, we fall back to tracked entity type ownership and the registering org unit
+   * of the tracked entity.
+   */
+  private TrackedEntityAttribute getAttribute(
+      UID attributeUid, TrackedEntity trackedEntity, @CheckForNull UID programUid)
       throws NotFoundException {
```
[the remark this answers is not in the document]
teleivo: That is exactly the purpose of this annotation. https://github.com/dhis2/wow-backend/blob/bf4b42b79b8f122db62e6fc0ebcdedde0bedf719/guides/preferred_patterns.md?plain=1#L20-L28

I can remove it from the private method if you like. We should definitely add these on our public methods.

lucaCambi77 (approved):

PR #16575, opened 2024-02-21
fix: run in memory jobs on any node [DHIS2-16497]

jbee:
The fix is twofold:

* always execute a "in-memory" job independent of leader status (a job submitted to a node for immediate execution that is not persisted)
* only synchronize the "runningRemotely" for job types that should only be run by the leader


jbee:
I used the wrong base branch in this - closing in favour of https://github.com/dhis2/dhis2-core/pull/16577

PR #16576, opened 2024-02-21
feat: Make org. unit param ("ou") not mandatory[DHIS2-16841]

d-bernat:
The organization unit request parameter in the Outliers API is now optional. The validator and SQL statement producer have been enhanced to accommodate this change.
The requested organization units are the subject of security constrain. 
Here's a breakdown of how it works:

- **User Permissions**: Each user has a set of permissions that dictate which organizational units they can access.
- **Data View Organization Units**: This likely refers to a predefined set of organizational units that are associated with the user's data view. These could be units they manage, units relevant to their role, or units they're explicitly granted access to.
- **Requested Units**: These are the organizational units that the user is trying to access.
- **Accessible Units**: The accessible units are determined by taking the intersection of the user's data view organization units and the requested units. This means that the user can only access units that are both in their data view and included in the request.
- **Absence of Requested Units**: If the user doesn't specify any organizational units in their request, they automatically have access to all units in their data view.
- **Super User**: The super user, likely an administrator or a user with elevated privileges, has access to all organizational units regardless of the constraints mentioned above. Their access is not restricted by data view organization units or requested units.

This system ensures that users can only access the organizational units that they're authorized to see based on their permissions and data view, while allowing the super user to have unrestricted access for administrative purposes.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
@@ -96,6 +96,10 @@ public class Program extends BaseNameableObject implements VersionedObject, Meta
 
   private String trackedEntityAttributeLabel;
 
+  private String programStageLabel;
+
+  private String eventLabel;
+
   private Set<OrganisationUnit> organisationUnits = new HashSet<>();
 
   private Set<ProgramStage> programStages = new HashSet<>();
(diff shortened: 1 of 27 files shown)


maikelarabori (commented):
I left a small comment @d-bernat.

Also, which behavior of the Data Visualizer have you followed? Did you mimic the same behavior as we agreed?

```
We have to mimic the same behavior, please. Unless there is some surprise.
```

You may have found some blocker, so we cannot mimic it...?

maikelarabori (approved):
Looks good 👍 Thx!

gnespolino (approved):

PR #16577, opened 2024-02-21
fix: run in memory jobs on any node [DHIS2-16497] (2.40)

jbee:
The fix is twofold:

* always execute a "in-memory" job independent of leader status (a job submitted to a node for immediate execution that is not persisted)
* only synchronize the "runningRemotely" for job types that should only be run by the leader


jimgrace (approved):
Tested in the test scenario that was used to isolate the problem for PEPFAR, and it works great. Analytics updates can be run in either the leader or non-leader nodes of a cluster.

PR #16594, opened 2024-02-23
fix: Add web test for Indonesian locales [DHIS2-13358]

david-mackessy:
- adding a web test checking the Indonesian locale codes returned
- this will help with fixes required in older versions (dev / v41 has the correct behaviour which this test confirms)

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_46__add_data_sharing_for_aggregate_data_exchange.java
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_46__add_data_sharing_for_aggregate_data_exchange.java
@@ -40,8 +40,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * In table aggregatedataexchange, copy Metadata Write value to Data Write value in sharing.users and
- * sharing.userGroups in sharing column. Public access must not be changed.
+ * In table aggregatedataexchange, copy Metadata Write value to Data Write value in sharing.users
+ * and sharing.userGroups in sharing column. Public access must not be changed.
  *
  * @author Viet Nguyen
  */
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_46__add_data_sharing_for_aggregate_data_exchange.java
```diff
@@ -40,8 +40,8 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * In table aggregatedataexchange, copy Metadata Write value to Data Write value in sharing.users and
- * sharing.userGroups in sharing column. Public access must not be changed.
+ * In table aggregatedataexchange, copy Metadata Write value to Data Write value in sharing.users
+ * and sharing.userGroups in sharing column. Public access must not be changed.
```
vietnguyen, line 44: Update flyway script will cause checksum error. Unless its really necessary, I think we should remove this change.

vietnguyen (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v41/V2_41_46__add_data_sharing_for_aggregate_data_exchange.java
david-mackessy: yeah, good catch. I can revert, but the formatting step will fail. Let's see if it allows merging with an invalid format.
I wonder how the unaccepted format got in?

david-mackessy: ah, [the original PR](https://github.com/dhis2/dhis2-core/pull/16558) did have a format error but it was merged anyway

vietnguyen: ah, I thought the formatter error would block the Merge button ? I will create another PR to fix it.

david-mackessy: maybe we should make the formatting a mandatory PR step, needing to pass. We can chat about that next week as a team and as a wider group I suppose.

PR #16602, opened 2024-02-26
Analytics outliers API: period not fully supported. Missing fixed types.[DHIS2-16894]

d-bernat:
TBD


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/outlier/data/OutlierQueryParser.java
```diff
@@ -233,12 +234,13 @@ private List<OrganisationUnit> applySecurityConstrain(
    * @param relativePeriod the {@link RelativePeriodEnum}.
```
maikelarabori, line 234: Please, update the Java doc, as we do not use RelativePeriodEnum anymore :
```
@param relativePeriod the period.
```
Once done, the enum `RelativePeriodEnum` can be removed from the `import` section.

maikelarabori (approved):
Looks good, just a minor comment. Thx

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/outlier/data/OutlierQueryParser.java
d-bernat: Done

gnespolino (approved):

PR #16625, opened 2024-02-27
feat: TE analytics API - Meta data should return only dimension and filter objects[DHIS2-16533]

d-bernat:
The TEI API provides metadata in response, including descriptions of relevant data entities. However, there may be responses where the quantity of delivered metadata items exceeds the dimensions and filters specified in the user request. To address this, proper filtering of the metadata content has been implemented.


gnespolino (approved):

maikelarabori (approved):
Thanks Dusan, questions clarified.

PR #16628, opened 2024-02-27
feat: add TEA changelogs endpoint [DHIS2-16741]

muilpp:
First PR to create an endpoint to serve tracked entities change logs.

It includes:
- ACL validations
- Response in the format agreed in the [design document](https://docs.google.com/document/d/1oij16GPVUUDxzjX2-hSssXenntiadZni7jKCEAcWdwM/edit?usp=sharing).
- New store to fetch all the necessary data from the database.
- Deprecate the old service
- Pagination

Sorting is out of the scope, will be done in an upcoming PR.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -0,0 +1,137 @@
...
+      // the advantage of doing this? Are we giving away any detail by saying the user has no access
+      // to an org unit?
+      throw new NotFoundException(TrackedEntity.class, trackedEntityUid.getValue());
+    }
+  }
+
+  private Set<String> validateTrackedEntityAttributes(TrackedEntity trackedEntity)
+      throws NotFoundException {
+    Set<TrackedEntityAttribute> attributes =
+        trackedEntityAttributeService.getAllUserReadableTrackedEntityAttributes(
```
teleivo, line 127: I think you can use https://github.com/dhis2/dhis2-core/blob/3f52715324afbee980bc6843433a94984c5e022a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackedEntityAttributeService.java#L313

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,112 @@
...
+      (rs, rowNum) -> {
+        UserInfoSnapshot createdBy = new UserInfoSnapshot();
+        createdBy.setUsername(rs.getString("userName"));
+        createdBy.setFirstName(rs.getString("firstname"));
+        createdBy.setSurname(rs.getString("surname"));
+        createdBy.setUid(rs.getString("useruid"));
+
+        return new TrackedEntityChangeLog(
+            createdBy,
+            rs.getTimestamp("updatedAt"),
```
teleivo, line 57: why is it not a `created` column 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,112 @@
...
+@Repository("org.hisp.dhis.tracker.export.trackedentity.JdbcTrackedEntityChangeLogStore")
+@RequiredArgsConstructor
+public class JdbcTrackedEntityChangeLogStore {
+
+  private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+
+  private static final RowMapper<TrackedEntityChangeLog> customTrackedEntityChangeLogRowMapper =
+      (rs, rowNum) -> {
+        UserInfoSnapshot createdBy = new UserInfoSnapshot();
+        createdBy.setUsername(rs.getString("userName"));
```
teleivo, line 50: DB columns/tables are all lowercase according to our [wow guide](https://github.com/dhis2/wow-backend/blob/b8f4a30fc72a0cff0ed1d6eaaa667747252382bd/guides/new_database_table_flyway_migration.md?plain=1#L45)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,112 @@
...
+        """
+          select cl.audittype,
+              case
+          when cl.audittype = 'CREATE' then cl.currentchangelogvalue
+          when cl.audittype = 'UPDATE' then cl.currentchangelogvalue
+          end as currentValue,
+              case
+          when cl.audittype = 'DELETE' then cl.currentchangelogvalue
+          when cl.audittype = 'UPDATE' then cl.previouschangelogvalue
+          end as previousValue, cl.created as updatedAt, cl.modifiedby as userName, cl.trackedEntityAttributeUid, cl.firstname, cl.surname, cl.username, cl.useruid, cl.audittype as type
```
teleivo, line 78: I would only add aliases in one place in the query. That makes it easier to follow. I moved all aliases into the subquery https://github.com/dhis2/dhis2-core/pull/16622/files

You also have a duplicate with different casing that is confusing `cl.username` and `cl.modifiedby as userName`. Store in https://github.com/dhis2/dhis2-core/pull/16622/files addresses these.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityChangeLogService.java
```diff
@@ -0,0 +1,43 @@
...
+package org.hisp.dhis.tracker.export.trackedentity;
+
+import java.util.List;
+import org.hisp.dhis.common.UID;
+import org.hisp.dhis.feedback.NotFoundException;
+
+public interface TrackedEntityChangeLogService {
+
+  /**
+   * Retrieves the change log data for a particular tracked entity
```
teleivo, line 37:
```suggestion
   * Retrieves the change log data for a particular tracked entity.
```

We write full sentences in javadocs.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/JsonTrackedEntityChangeLog.java
```diff
@@ -0,0 +1,65 @@
...
+    return getString("type").string();
+  }
+
+  default JsonChange getChange() {
+    return get("change").as(JsonChange.class);
+  }
+
+  interface JsonChange extends JsonObject {
+    default JsonTrackedEntityAttribute getTrackedEntityAttributeChange() {
+      return get("trackedEntityAttribute").as(JsonTrackedEntityAttribute.class);
```
teleivo, line 48: I would stick to our tracker naming in the API. https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Attribute.java

we say `attribute` https://github.com/dhis2/dhis2-core/blob/3f52715324afbee980bc6843433a94984c5e022a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/TrackedEntity.java#L84

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,112 @@
...
+            rs.getString("type"),
+            new Change(
+                new TrackedEntityChangeLog.TrackedEntityAttributeChange(
+                    rs.getString("trackedEntityAttributeUid"),
+                    rs.getString("previousValue"),
+                    rs.getString("currentValue"))));
+      };
+
+  public List<TrackedEntityChangeLog> getTrackedEntityChangeLog(
+      String trackedEntity, Set<String> attributes) {
```
teleivo, line 67:
```suggestion
      UID trackedEntity, Set<String> attributes) {
```

We should use our `UID` type where we can. We lose information if we move from UID in the service to a String here.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerPostgresTest.java
```diff
@@ -0,0 +1,206 @@
...
+    return a;
+  }
+
+  private String createJson(
+      TrackedEntity trackedEntity, TrackedEntityAttribute trackedEntityAttribute) {
+    return """
+        {
+          "trackedEntities": [
+          {
+            "created": "2015-08-06T21:20:42.878",
```
teleivo, line 175: please remove fields that are not needed or will even be deliberately ignore like `created`. `created` set by the server. (its also named `createdAt` in our API)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
teleivo: We lose type information if we move from UID to a String. Why would we do that?

The reason we are using it in the web is because it all started with OpenAPI and then conversion of controller parameters to UID.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/JsonTrackedEntityChangeLog.java
muilpp: don't pay too much attention to this test, it's still a work on progress, the current response looks like this:
Screenshot from 2024-02-28 11-37-49, and it's already called attribute. I think that's the one you mean, right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private Set<String> validateTrackedEntityAttributes(TrackedEntity trackedEntity)
+      throws NotFoundException {
+    Set<TrackedEntityAttribute> attributes =
+        trackedEntityAttributeService.getTrackedEntityTypeAttributes(
+            trackedEntity.getTrackedEntityType());
+
+    if (attributes.isEmpty()) {
+      throw new NotFoundException(TrackedEntity.class, trackedEntity.getUid());
```
muilpp, line 132: @teleivo does it make sense to throw a `NotFoundException` in this case?
It's gonna make it hard for the end user to figure out the reason why there are no change logs shown, and I think we don't give any hint away by saying that this particular tracked entity has no attributes associated to the tracked entity type. What do you think?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerPostgresTest.java
```diff
@@ -0,0 +1,349 @@
...
+    JsonUser createdBy = changeLog.getCreatedBy();
+    JsonAttributeValue attributeChange = changeLog.getChange().getAttributeValue();
+    UserDetails currentUser = CurrentUserUtil.getCurrentUserDetails();
+
+    assertAll(
+        () -> assertEquals(currentUser.getUid(), createdBy.getUid()),
+        () -> assertEquals(currentUser.getUsername(), createdBy.getUsername()),
+        () -> assertEquals(currentUser.getFirstName(), createdBy.getFirstName()),
+        () -> assertEquals(currentUser.getSurname(), createdBy.getSurname()),
+        () -> assertEquals("CREATE", changeLog.getType()),
```
teleivo, line 149: Please use the same style as in EventsExportControllerPostgresTest
assertUpdate, assertDelete, ...

You can extract them in order to reuse them.

dhis-2/dhis-test-web-api/src/test/resources/tracker/simple_metadata.json
```diff
@@ -0,0 +1,228 @@
...
+      },
+      "authorities": [
+        "F_TRACKED_ENTITY_INSTANCE_SEARCH_IN_ALL_ORGUNITS",
+        "ALL"
+      ]
+    }
+  ],
+  "trackedEntityAttributes": [
+    {
+      "lastUpdated": "2020-05-31T11:41:22.414",
```
teleivo, line 202: lastUpdated, created have an effect as they are set by us. It confused me a while back when debugging failed tests so we deleted them.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,112 @@
...
+import org.springframework.jdbc.core.namedparam.SqlParameterSource;
+import org.springframework.stereotype.Repository;
+
+@Repository("org.hisp.dhis.tracker.export.trackedentity.JdbcTrackedEntityChangeLogStore")
+@RequiredArgsConstructor
+public class JdbcTrackedEntityChangeLogStore {
+
+  private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+
+  private static final RowMapper<TrackedEntityChangeLog> customTrackedEntityChangeLogRowMapper =
```
lucaCambi77, line 50: Is there a reason why we keep using Spring jdbc? I think if there isn't a specific reason we should use the JPA standards with named queries which gives the same support plus it is a standard previous to using Spring in general i think

lucaCambi77 (commented):
Is there a plan for this release to move the ChangeLogType.SEARCH or READ to the real audit? If not I think we should specify in the release note that we currently have an hibryd situation and in the audit tables there is a mix of audit and change log

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -0,0 +1,133 @@
...
+      throw new NotFoundException(Program.class, programUid);
+    }
+
+    return program;
+  }
+
+  private void validateOwnership(User currentUser, TrackedEntity trackedEntity, Program program)
+      throws NotFoundException {
+    if (!trackerAccessManager
+        .canRead(currentUser, trackedEntity, program, currentUser.isSuper())
```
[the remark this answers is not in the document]
muilpp: I don't think it's wrong because the super user will likely have access to all org units in the system anyway. But if we don't skip we are doing an unnecessary validation. Also, we agreed with markus and enrico that when it comes to a super user, we would only validate its sharing settings.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerPostgresTest.java
muilpp: I'm refactoring this in the next PR, where I'm dealing with sorting

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
muilpp: This was a personal choice of mine; I find it easier to work with JDBC rather than the entity manager or native queries. From my experience, I had some trouble, particularly when dealing with Collections. 
Before I started implementing this, I took a look at the tracker code base and noticed that we are using JDBC, HQL, and criteria queries in our endpoints. I don't think we have agreed on using one or the other; perhaps we could discuss this further and find common ground.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -0,0 +1,138 @@
...
+                  where audit.audittype in ('CREATE', 'UPDATE', 'DELETE')
+                  and t.uid = :trackedEntity
+       """;
+
+    List<TrackedEntityChangeLog> changeLogs;
+    if (attributes.isEmpty()) {
+      sql +=
+          """
+              order by audit.created desc) cl
+              limit :limit offset :offset
```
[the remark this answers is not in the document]
muilpp: The problem I see with keeping everything in the subquery is that we'd have to format the string with the clause as a variable, and that won't be nice nor easy to read.
I agree this is not an ideal solution, but there's only one condition, and with the cleanup you suggested, there's little room to not provide an order, at least for now.

muilpp:
> Is there a plan for this release to move the ChangeLogType.SEARCH or READ to the real audit? If not I think we should specify in the release note that we currently have an hibryd situation and in the audit tables there is a mix of audit and change log

If I'm not mistaken, this will be done after the soft freeze, right @teleivo?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -0,0 +1,133 @@
...
+  public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
+      UID trackedEntityUid, UID programUid, PageParams pageParams) throws NotFoundException {
+    TrackedEntity trackedEntity =
+        trackedEntityService.getTrackedEntity(trackedEntityUid.getValue());
+    if (trackedEntity == null) {
+      throw new NotFoundException(TrackedEntity.class, trackedEntityUid.getValue());
+    }
+
+    User currentUser = userService.getUserByUsername(CurrentUserUtil.getCurrentUsername());
+    Set<String> trackedEntityAttributes = Collections.emptySet();
```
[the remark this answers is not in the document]
muilpp: I think here's where image/file and change logs differ. For images and files, we only need to get one attribute, the one containing the data of that particular resource. 
On the other side, for change logs, we need to get all attributes the user has access to, and, even if a program is provided, there will also be tracked entity type attributes the user needs to see, that's why in this case we get both.

teleivo:
> Is there a plan for this release to move the ChangeLogType.SEARCH or READ to the real audit? If not I think we should specify in the release note that we currently have an hibryd situation and in the audit tables there is a mix of audit and change log

We don't have time for it so no.

The new changelog endpoints do not mention `audit` anywhere. Any vars or references to audit code is internal and will be fixed later.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
teleivo: > Is there a reason why we keep using Spring jdbc?
It spread from the EventStore which existed before @muilpp and I joined. I as well prefer full control over the queries so writing SQL is what I would go for. Feel free to present the differences and the advantages of what you are suggesting.

We do not have time to discuss and adapt this right now though 😬

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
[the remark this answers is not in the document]
teleivo: Can we not do

```sql
order by audit.created desc
limit :limit offset :offset) cl
```

?

teleivo (approved):

lucaCambi77 (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
muilpp: I didn't mean to adapt it before the soft freeze, but after the release it would be good to discuss this because it keeps coming back. To at least we could be in sync in our team.

PR #16630, opened 2024-02-27
refactor: Use analytics_rs_ as prefix for resource tables [DHIS2-16705]

larshelge:
Apache Doris does not allow for table names that start with `_`. Tables must start with a letter. That is a problem since our resource tables start with `_`. Changes the resource table prefix to become `analytics_rs_`. 

This means resource tables will continue to be automatically excluded from backups.


david-mackessy (approved):
where does Apache Doris come into play here? Is that one of the candidates for an analytical DB?

maikelarabori (approved):

PR #16674, opened 2024-03-01
feat: Analytics outliers: round to two decimal digits, instead of only one[DHIS2-16954]

d-bernat:
All decimal numbers are now rounded to two decimal places. The default rounding scale remains unchanged and is set to 10 decimals.

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
@@ -179,6 +179,7 @@ public class OutliersDetection2AutoTest extends AnalyticsApiTest {
     // Assert rows.
     validateRow(
         response,
+        0,
         List.of(
             "l6byfWFUGaP",
             "Yellow Fever doses given",
@@ -200,6 +201,7 @@ public class OutliersDetection2AutoTest extends AnalyticsApiTest {
             "391.03"));
     validateRow(
         response,
+        1,
         List.of(
             "s46m5MS0hxu",
             "BCG doses given",
@@ -221,6 +223,7 @@ public class OutliersDetection2AutoTest extends AnalyticsApiTest {
             "213.99"));
     validateRow(
         response,
+        2,
         List.of(
             "dU0GquGkGQr",
             "Q_Early breastfeeding (within 1 hr after delivery) at BCG",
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
```diff
@@ -293,11 +290,11 @@ public void queryOutliertest7() throws JSONException {
         false);
 
     // Assert rows.
-    validateRow(response, 0, List.of("O05mAByOgAv", "OPV2 doses given", "98.5"));
-    validateRow(response, 1, List.of("vI2csg55S9C", "OPV3 doses given", "70.3"));
-    validateRow(response, 2, List.of("n6aMJNLdvep", "Penta3 doses given", "70.1"));
-    validateRow(response, 3, List.of("UOlfIjgN8X6", "Fully Immunized child", "59.8"));
-    validateRow(response, 4, List.of("I78gJm4KBo7", "Penta2 doses given", "59.0"));
+    validateRow(response, List.of("O05mAByOgAv", "OPV2 doses given", "98.48"));
```
maikelarabori, line 293: Is there a reason why you removed the indexes? If we have the `orderBy=modifiedzscore`, we should assert that the order is correct. I would keep the indexes. If the `orderBy` is not working it should be fixed...

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
```diff
@@ -340,11 +337,11 @@ public void queryOutliertest8() throws JSONException {
     validateHeader(response, 2, "zscore", "Z-score", "NUMBER", "java.lang.Double", false, false);
 
     // Assert rows.
-    validateRow(response, 0, List.of("l6byfWFUGaP", "Yellow Fever doses given", "3.7"));
-    validateRow(response, 1, List.of("s46m5MS0hxu", "BCG doses given", "3.1"));
+    validateRow(response, List.of("l6byfWFUGaP", "Yellow Fever doses given", "3.66"));
```
maikelarabori, line 340: Is there a reason why you removed the indexes? If we have the `orderBy=zscore`, we should assert that the order is correct. I would keep the indexes. If the `orderBy` is not working it should be fixed...

maikelarabori (commented):
Two minor comments, @d-bernat. Otherwise looks good, thx

gnespolino (approved):

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
d-bernat: done

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
d-bernat: done

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/outlier/OutliersDetection2AutoTest.java
```diff
@@ -340,11 +340,11 @@ public void queryOutliertest8() throws JSONException {
     validateHeader(response, 2, "zscore", "Z-score", "NUMBER", "java.lang.Double", false, false);
 
     // Assert rows.
-    validateRow(response, 0, List.of("l6byfWFUGaP", "Yellow Fever doses given", "3.7"));
+    validateRow(response, 0, List.of("l6byfWFUGaP", "Yellow Fever doses given", "3.66"));
     validateRow(response, 1, List.of("s46m5MS0hxu", "BCG doses given", "3.1"));
     validateRow(
```
maikelarabori, line 345: I think this line is still missing the index...

d-bernat: done

maikelarabori (approved):

PR #16676, opened 2024-03-01
feat: Add sorting feature to TEA change logs endpoint [DHIS2-16964]

muilpp:
This PR introduces:
- Sorting for TEA change logs. For now only sorting by creation data is supported, both ascending and descending, agreed during [the design phase](https://docs.google.com/document/d/1xPXZ9sBizX8Yv9J2CO2G5hFRPrVzoNMO71uRiNVug5I/edit#heading=h.8p6gdd5nuezo). 
- Filter by program uid when specified. In this case, besides showing the TETA available to the user, we also need to show program attributes linked to the supplied program.


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityChangeLogServiceTest.java
```diff
@@ -218,6 +219,8 @@ void shouldReturnChangeLogsWhenTrackedEntityAttributeValueIsCreated() throws Not
     programOwnerService.createOrUpdateTrackedEntityProgramOwner(
         trackedEntity.getUid(), program.getUid(), orgUnitA.getUid());
 
+    createAndPersistProgramAttribute(program);
+
```
lucaCambi77, line 231: this way of using helpers is confusing to me. you say `createAndPersistProgramAttribute` by passing the program. But then in the method you have the `trackedEntityAttribute` of the class scope. In my opinion, say you want to create program attributes, i would do a method like `createProgramAttributes` or `programAttributes` by passing the program and the attributes. At a first glance, i am not interested in what is going on inside the method, but i want to see a signature explaining what is going on

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -106,28 +116,42 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
...
+    }
 
     List<TrackedEntityChangeLog> changeLogs;
     if (attributes.isEmpty()) {
       sql +=
           """
-              order by %s) cl
-              limit :limit offset :offset
-          """.formatted(sortExpressions(order));
+              order by %s
```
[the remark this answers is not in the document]
teleivo: It will in the future. I don't think it needs to handle something that is not supported yet. We would also need a test for multiple expressions if we implement it but we don't yet support that.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -77,38 +106,52 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
...
-            .addValue("offset", (pageParams.getPage() - 1) * pageParams.getPageSize());
+
+    if (program != null) {
+      sql +=
+          """
+              join program_attributes pa using (trackedentityattributeid)
+              join program p using (programid)
+              where audit.audittype in ('CREATE', 'UPDATE', 'DELETE')
+              and t.uid = :trackedEntity
+              and p.uid = :programUid
```
teleivo, line 128:
```suggestion
              and p.uid = :program
```

Since you call the `t.uid` trackedEntity and not trackedEntityUid

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/JdbcTrackedEntityChangeLogStore.java
```diff
@@ -77,38 +106,52 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
...
-              limit :limit offset :offset
-          """;
+              order by %s
+              limit :limit offset :offset) cl
+          """
+              .formatted(sortExpressions(order));
     } else {
       sql +=
           """
               and tea.uid in (:attributes)
```
teleivo, line 150: You could even invert the if condition and add the `and tea.uid in (:attributes)` to `sql` then at the end add the limit/offset. This would reduce some nesting.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityChangeLogServiceTest.java
```diff
@@ -209,16 +227,20 @@ void shouldReturnChangeLogsWhenTrackedEntityAttributeValueIsCreated() throws Not
...
-            UID.of(trackedEntity.getUid()), UID.of(program.getUid()), defaultPageParams);
+            UID.of(trackedEntity.getUid()),
+            UID.of(program.getUid()),
+            defaultOperationParams,
+            defaultPageParams);
 
     assertNumberOfChanges(1, changeLogs.getItems());
     assertAll(
         () ->
             assertChange(
```
teleivo, line 242: can you please use/add the assertCreate/assertUpdate/Delete assertions we have for event changelogs. it makes the tests easier to read.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityChangeLogServiceTest.java
```diff
@@ -338,14 +373,39 @@ void shouldReturnChangeLogsFromAccessibleTEAWhenProgramNotSpecified() throws Not
...
+  void shouldReturnChangeLogsFromSpecifiedProgramOnlyWhenMultipleLogsExist()
+      throws NotFoundException {
+    updateAttributeValue(trackedEntityAttributeValue, ChangeLogType.CREATE, "new value");
+    updateAttributeValue(outOfScopeTrackedEntityAttributeValue, ChangeLogType.CREATE, "new value");
+    Program program = createAndAddProgram('C');
+
+    programOwnerService.createOrUpdateTrackedEntityProgramOwner(
+        trackedEntity.getUid(), program.getUid(), orgUnitA.getUid());
+
+    createAndPersistProgramAttribute(program);
```
teleivo, line 395: I don't get how the program influences the result. What attribute is a program attribute in here? Can we adapt the helper to make that clear and remove other things from the setup here that are just details?

muilpp: I'm creating values for two different TEA. 
The program influences the result because only one of these two TEA belong to this program, hence the name starting with "outOfScope" for the second TEA.
In this case, only the first TEA is a program attribute, because I'm creating it at the line you pinned.

The thing with moving something to the setup is it will affect other tests, and we don't want to create a program attribute or set the ownership of a program in all the tests.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -93,7 +93,11 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
     }
 
     return jdbcTrackedEntityChangeLogStore.getTrackedEntityChangeLog(
-        trackedEntityUid, trackedEntityAttributes, operationParams.getOrder(), pageParams);
+        trackedEntityUid,
+        trackedEntityAttributes,
+        programUid,
+        operationParams.getOrder(),
+        pageParams);
   }
```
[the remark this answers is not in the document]
muilpp: We could do it, but then we would have the operations mapper in the store signature, which feels wrong, since the tracker stores always use `QueryParams` instead. 
I didn't do that because it feels a bit of an overkill to have a mapper and `QueryParams` for these two parameters.

lucaCambi77 (approved):

PR #16679, opened 2024-03-02
chore: override the process of default field selection

zubaira:
`AbstractReadOnlyController` can now allow the user to override default-behaviour of fields selection.


teleivo (commented):
Is there a test for defaults?

jbee:
I am not sure we want flexibility this way because then the semantics of that the set of default fields are would be dependent on the object type. It might be better to have a different solution that only applies to your endpoint. But I will discuss this tomorrow with the platform team.

zubaira:
> I am not sure we want flexibility this way because then the semantics of that the set of default fields are would be dependent on the object type. It might be better to have a different solution that only applies to your endpoint. But I will discuss this tomorrow with the platform team.

This is not a solution for any specific endpoint or any specific object. If a user wants a different **default-behavior** then he should be able to do that.

jbee:
> This is not a solution for any specific endpoint or any specific object. If a user wants a different default-behavior then he should be able to do that.

This sounds like a new feature requirement - if so this should go the normal way of proposing this feature

zubaira:
closing it for now. Will be entertained later on.
https://github.com/dhis2/dhis2-core/pull/16679#issuecomment-1976470818

PR #16691, opened 2024-03-04
feat: Add new self registration endpoint [DHIS2-14615]

david-mackessy:
# Summary
New endpoint added for self registration: `POST` `/api/auth/register` (to be used by the new Login App)
- This is basically a copy of the old struts login endpoint `POST` `/api/account` which is now marked as deprecated
- extracted some methods (into `UserAccountService`) from the copied controller code to clean things up a bit

# Tests
## Automated
- web tests covering multiple registration scenarios

## Manual
Enable self reg without recaptcha:
`POST` `/api/systemSettings/keySelfRegistrationNoRecaptcha?value=true`

self register:
`POST` `/api/auth/register` with body
```
{
    "username": "testSelfReg",
    "firstName": "test",
    "surname": "selfReg",
    "password": "P@ssword123",
    "email": "test@test.com",
    "phoneNumber": "12345oooo",
    "g-recaptcha-response": "recap respo"
}
```

check me endpoint with credentials used above to verify it worked:
`GET` `/api/me`

# Info
Can't test recaptcha locally.

The old `AccountController` could use the new service methods if we wanted to do this in the future. Seeing as it's deprecated it's probably not worth the effort.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
```diff
@@ -0,0 +1,209 @@
...
+  private final ConfigurationService configService;
+  private final TwoFactorAuthenticationProvider twoFactorAuthProvider;
+  private final SystemSettingManager systemSettingManager;
+  private final PasswordValidationService passwordValidationService;
+
+  private static final int MAX_LENGTH = 80;
+  private static final int MAX_PHONE_NO_LENGTH = 30;
+
+  @Override
+  public void createAndAddSelfRegisteredUser(
```
vietnguyen, line 70: Do we need @Transactional here ?

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/SelfRegistrationForm.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class SelfRegistrationForm {
```
jbee, line 40: Can we keep common name patterns? Something with `Params` suffix.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
jbee: If not please use one of the alternative annotations to indicate what case it is, `@NonTransactional` and alike

netroms (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/UserAccountController.java
```diff
@@ -165,4 +171,15 @@ public void resetPassword(@RequestBody ResetPasswordRequest resetRequest)
 
     log.info("Account restored for user: {}", user.getUsername());
   }
+
+  @PostMapping("/register")
+  public WebMessageResponse register(
+      @RequestBody SelfRegistrationForm selfRegForm, HttpServletRequest request)
+      throws BadRequestException, IOException {
+    log.info("Self registration received");
+    userAccountService.validateUserFormInfo(selfRegForm, request);
```
jbee, line 180: Could we make validation part of the action? What would be the reason to make is into a separate call?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
david-mackessy: yep, i'll add it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/SelfRegistrationForm.java
david-mackessy: i'll update, i'll remember this eventually 😄

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/UserAccountController.java
david-mackessy: hmm, I was trying to break down the copied code into more readable chunks and this seemed to make sense. The validate method doesn't require a transaction so it can be kept separate.
Also, they were left separate so the `AccountController` could use either method if we want to reduce code duplication in that class.

PR #16697, opened 2024-03-04
fix: TEI: Bug in the repositioning of rowContext column indexes[DHIS2-16982].

d-bernat:
There is a bug in the process of repositioning the row context columns. The columns need to be reindexed according to the incoming headers order.


gnespolino (approved):

maikelarabori (approved):

maikelarabori (commented):
@d-bernat, it would be nice to see the `created` in the rows response, to make sure this is sorted and there are no repetitions.

PR #16702, opened 2024-03-05
refactoring: use ContextService to set prev/nextPage in tracker DHIS2-16950

teleivo:
to handle cases where DHIS2 is deployed behind a proxy.

* extract functions which parse the servlet/context and API paths from an HttpServletRequest from [DefaultContextService](https://github.com/dhis2/dhis2-core/blob/fac46e89f156f4c0206d42d68b0a333369256ed8/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/service/DefaultContextService.java#L58C17-L58C31). The request is often already present in a controller so there is no need to get it from a local tread var as done by this service
* Use the `getApiPath` in tracker as this is the method used by the `LinkService` that generates the links for metadata
* Removed duplicated function from `ContextUtils`

This is a mere refactoring. Tested via existing controller tests like [EventsExportControllerPostgresTest](https://github.com/dhis2/dhis2-core/blob/3d3a3b31b4ed9846ef18ec0cec271913a315b9d2/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerPostgresTest.java#L211-L221).

## Future

We still have outstanding Tracker issues in https://dhis2.atlassian.net/browse/DHIS2-16573 which we will tackle after the release.

We should be able to reuse the JSON handler in all exporter endpoints serving JSON. I think we can create dedicated handlers like https://github.com/dhis2/dhis2-core/blob/3d3a3b31b4ed9846ef18ec0cec271913a315b9d2/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/FileResourceRequestHandler.java#L48 and the JSON one. Each bundles the dependencies it needs to provide a response and controllers delegate to them. Right now we need to repeat ourselves when it comes to returning JSON and CSV which is not necessary.


jbee (commented):
I think we should do a quick walk through call for this so I can follow this more on the functional level.

muilpp (approved):

PR #16703, opened 2024-03-05
feat: serve global app shell at /apps/ [DHIS2-17021]

amcgee:
If an app with the name `global-app-shell` is installed this filter will take effect.

1. Serve the global app shell app at `/apps/` (including top-level subpaths such as `/apps/dashboard`)
2. Redirect from existing app paths (i.e. `/dhis-web-dashboard/index.html`) to the global app shell path for the same app (`/apps/dashboard`)
3. For convenience, redirect from `/apps/dashboard/` to `/apps/dashboard`
4. Inject DHIS2 base url into index.html files for installed apps, replacing `__DHIS_BASE_URL__` (ideally included in index.html as `<meta name="dhis-base-url" content="__DHIS2_BASE_URL__" />`). This is particularly helpful for the global app shell itself, but can also be leveraged by any other app
5. Skip the redirect (number 2 above) if the querystring contains `?redirect=false`

A few things still to iron out...
- [ ] This is opt-in, by default nothing changes and no global app shell is bundled.  Should we consider making this the default?  It would significantly increase the consistency benefits
- [ ] Bundled apps can't take advantage of the `__DHIS2_BASE_URL__` runtime replacement (this might be OK for now...)
- [ ] We should perhaps redirect from `/apps/dashboard` back to `/dhis-web-dashboard` if the app shell has been uninstalled
- [ ] Appending a filename from a particular application (i.e. `webapp.manifest` or `plugin.html`) manually to the global app shell public URL will no longer work.  For example `/apps/dashbaord/webapp.manifest` or `/apps/dashboard/package.json` will return a 404.  This could be confusing and obnoxious, but mostly for developers and expert users... perhaps these paths should redirect back to their source counterparts...?


stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/GlobalAppShellFilter.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private boolean redirectLegacyAppPaths(
+      HttpServletRequest request, HttpServletResponse response, String path) throws IOException {
+    String queryString = request.getQueryString();
+    Matcher m = APP_PATH_PATTERN.matcher(path);
+
+    boolean matchesPattern = m.find();
+    boolean isIndexPath = path.endsWith("/") || path.endsWith("/index.html");
+    boolean hasRedirectFalse = queryString == null || !queryString.contains("redirect=false");
```
stian-sandvold, line 94: Is the naming a bit confusing here?

"hasRedirectFalse" I read as "Does the request have "redirect=false" present. However, in the assignment we check if "redirect=false" is NOT there.

So if we have "redirect=false", hasRedirectFalse == false.
And if we have null or "redirect=true" or no mention of redirect, hasRedirectFalse == true

If I understand your description of the PR, this should be opt-in, so we dont redirect by default, but here it seems we do redirect by default, unless we say redirect=false.

So I think if we do redirect for null, then we should rename the variable?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/GlobalAppShellFilter.java
```diff
@@ -0,0 +1,136 @@
...
+    if (matchesPattern && isIndexPath && hasRedirectFalse) {
+      String appName = m.group(1);
+      response.sendRedirect(request.getContextPath() + GLOBAL_APP_SHELL_PATH_PREFIX + appName);
```
stian-sandvold, line 97: Also, if redirect=false, should we redirect?

stian-sandvold: Maybe its just the name of the variable that confuses me

amcgee: Variable name was incorrect, fixed now

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/GlobalAppShellFilter.java
amcgee: Fixed now, variable name was incorrect

stian-sandvold (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
```diff
@@ -273,11 +276,28 @@ public void renderApp(
         response.setContentType(mimeType);
       }
 
-      response.setContentLengthLong(resource.contentLength());
-      response.setHeader(
-          "Last-Modified", DateUtils.toHttpDateString(new Date(resource.lastModified())));
-
-      StreamUtils.copyThenCloseInputStream(resource.getInputStream(), response.getOutputStream());
+      if (filename.endsWith(".html")) {
```
jbee, line 279: Can this condition be made more strict? 
I would assume we know more about which HTML documents need the below line-by-line replacement or is it really any of them?

jbee (commented):
It is not clear to me what sort of transformation or replacement this does but it feels like to general applying to way more requests that it has to be applied to. 
Overall it feels we "bold on" a solution for global shell instead of designing it. Not sure how fair that assessment is from a outside view perspective. I just see the way it interferes with requests and all the ways you mention that one does need to consider to install or remove it seem very complicated to me. I guess the server side of global shell would fall into the responsibility of the platform team but AFAIK nobody of the backend platform team has been involved in designing a working principle. That is maybe not the best way to introduce features like this.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
jbee: As discussed in today's meeting: we can limit this to a fixed list of files.

jbee (approved):

PR #16712, opened 2024-03-06
feat: Add invite user endpoint [DHIS2-14617]

david-mackessy:
# Summary
New endpoint added for an invite registration: `POST` `/api/auth/invite` (to be used by the new Login App)

This is basically a copy of the old struts login endpoint `POST` `/api/account` which is now marked as deprecated.
Extracted some methods (into `UserAccountService`) from the copied controller code to clean things up a bit.

Also rename `/register` endpoint to `/registration`

# Tests
## Automated
web tests covering multiple invite scenarios

## Manual
As a prerequisite to using the new endpoint, an Admin should setup the proposed user to invite first:
`POST` `/api/users`
```json
{
    "uid": "Uid00000001",
    "name": "Alan Partridge",
    "firstName": "Alan",
    "surname": "Partridge",
    "password": "Test123!",
    "username": "alanP",
    "userRoles": [
        {
            "id": "{userRoleId}"
        }
    ]
}
```
Then insert these values directly into the DB as they cannot be set through the API
`update userinfo set restoretoken = '$2a$10$fScYIKiJx6sBWBm/U0QgR.fPlLJeMXOu0CmuharO7v5XVOSZRZ.p.' where uid = '{uidOfCreatedUser}';`
`update userinfo set restoreexpiry  = '2040-06-03T00:00:00.000' where uid = '{uidOfCreatedUser}';`
`update userinfo set idtoken  = 'idToken' where uid = '{uidOfCreatedUser}';`

Turn off recaptcha (recaptcha should be checked in the Login App UI testing:
`POST` `/api/systemSettings/keySelfRegistrationNoRecaptcha?value=true`

complete invite register:
`POST` `/api/auth/invite` with body:

> ** Note ** 
> The token field is Base64 encoded and needs to start with `ID` for this invited user test
> This criteria is part of the recaptcha validation in `RestoreOptions#getRestoreOptions`
> Unencoded it is `idToken:IDrestoreToken`

```json
{
    "username": "alanP",
    "firstName": "Alan",
    "surname": "Partridge",
    "password": "Test123!",
    "email": "test@test.com",
    "phoneNumber": "12345oooo",
    "g-recaptcha-response": "recap disabled",
    "token": "aWRUb2tlbjpJRHJlc3RvcmVUb2tlbg=="
}
```

check `me` endpoint with credentials of invited user to verify it worked:
`GET` `/api/me`

# Info
Can't test recaptcha locally.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/SelfRegistrationParams.java
```diff
@@ -27,24 +27,13 @@
...
-  @JsonProperty String firstName;
-  @JsonProperty String surname;
-  @JsonProperty String password;
-  @JsonProperty String email;
-  @JsonProperty String phoneNumber;
-
-  @JsonProperty("g-recaptcha-response")
-  String recaptchaResponse;
-}
+public class SelfRegistrationParams extends UserRegistrationParams {}
```
jbee, line 39: Can't we reuse the `UserRegistrationParams` if there is nothing more in here?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/UserRegistrationParams.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@Setter
+@ToString
+@NoArgsConstructor
+public abstract class UserRegistrationParams {
```
jbee, line 40: `abstract` and "data" types is sending confusing signals.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
```diff
@@ -67,92 +67,115 @@ public class DefaultUserAccountService implements UserAccountService {
   private static final int MAX_LENGTH = 80;
   private static final int MAX_PHONE_NO_LENGTH = 30;
 
+  @Override
+  public void validateSelfRegUser(UserRegistrationParams userRegParams, String remoteAddress)
```
jbee, line 71: I am also wondering what kind of registration other than a "self" registration there is? We should find a simple language for this, like "registration" is a user driven registration, "invite" a admin driven registration.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
```diff
@@ -67,92 +67,115 @@ public class DefaultUserAccountService implements UserAccountService {
...
   }
 
   @Override
-  public void validateUserFormInfo(SelfRegistrationParams selfRegForm, HttpServletRequest request)
-      throws BadRequestException, IOException {
-    log.info("Validating user info");
-    validateCaptcha(selfRegForm.getRecaptchaResponse(), request);
-    validateUserForm(selfRegForm);
+  @Transactional
+  public void updateInvitedRegisteredUser(
```
jbee, line 116: Is this when the invited user confirms and the account is created? What if we actually call it that? `confirmInvite`?

jbee (commented):
I think we should refine the naming. Drop the "complete" in combination with "registration" because this can be confused with the "complete data set registration". Then call user driven account creation a "registration" and admin driven account creation "invite" and "confirmInvite"? Something like that.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/SelfRegistrationParams.java
david-mackessy: It was implemented this way so that validating a password can accept the abstract class (seeing as they share most of their properties). It could be changed to have 2 validate password methods, One taking the `SelfRegistrationParams` and the other taking `CompleteRegistrationParams`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/UserRegistrationParams.java
david-mackessy: see above comment, we could keep the 2 param classes completely separate (without this class) and have some duplication in some areas. It seemed a shame not to try and reuse these fields.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
david-mackessy: The User is already created by the admin is my understanding, so this is actually updating the user. I don't mind what it's called.

david-mackessy:
> I think we should refine the naming. Drop the "complete" in combination with "registration" because this can be confused with the "complete data set registration". Then call user driven account creation a "registration" and admin driven account creation "invite" and "confirmInvite"? Something like that.

@jbee Those 2 endpoints `/register` and `/completeRegistration` are used by the Login App now. This was the info in the feature which one would presume was chosen for a good reason. These are the kinds of issues that arise when time, care and attention are not given to features. This whole area needs a facelift :)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserAccountService.java
david-mackessy: updated

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/UserInviteRegistrationParams.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class UserInviteRegistrationParams extends RegistrationParams {
```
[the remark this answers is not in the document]
david-mackessy: updated

jbee (approved):

vietnguyen (approved):

PR #16717, opened 2024-03-07
fix: api/pdfForm return error 500 if db_locale is null

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-16937
### Issue
- Below method call can return null
```CurrentUserUtil.getUserSetting(UserSettingKey.DB_LOCALE)```

- Currently the Database Locale setting doesn’t have a default value in both SystemSetting and UserSetting. I guess this is intended because the `UI_Locale` does have en as the default value. 
- In `BaseIdentifiableObject`, when getting translation value of an object property, if the locale is null then we just return the exact value from database. This value may be in a language which is different from UI default locale ( English ). 
- For example, user may create a DataElement with name 數據元素 in an DHIS2 instance with UI_Locale=en and DB_Locale=null. I think there are use cases for this way of work.
### Fix
- This PR make the PdfFormController to use `UI_Locale`. The `DB_Locale` is already used in method `getDisplay*()` in `BaseIdentifiableObject`. So the result PDF file will have the labels use `UI_Locale` and all object names getting from database use `DB_Locale`. 

### Test
- Added a test to make sure no 500 error return if `DB_Locale` is null.

### Todo
- Go through all callers of `CurrentUserUtil.getUserSetting(UserSettingKey.DB_LOCALE)` and make sure they already handle the case when `null` returned.

### CODE REVIEW

File: dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserSettingService.java [hunk 1]
```diff
@@ -177,14 +178,14 @@ public Serializable getUserSetting(UserSettingKey key, String username) {
...
-                        && userSetting.getValue() != null)
-            .collect(Collectors.toMap(UserSetting::getName, UserSetting::getValue));
+    Map<String, Serializable> result = new HashMap<>();
+    Sets.newHashSet(getUserSettings(user)).stream()
+        .filter(
+            userSetting ->
+                userSetting != null
+                    && userSetting.getName() != null
+                    && userSetting.getValue() != null)
+        .forEach(userSetting -> result.put(userSetting.getName(), userSetting.getValue()));
```
vietnguyen, line 188: This is to make sure `getUserSettingsWithFallbackByUserAsMap()` never return `null`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetailsImpl.java
```diff
@@ -59,7 +60,7 @@ public class UserDetailsImpl implements UserDetails {
   private Collection<GrantedAuthority> authorities;
   private Set<String> allAuthorities;
   private Set<String> allRestrictions;
-  private Map<String, Serializable> userSettings;
+  private Map<String, Serializable> userSettings = new HashMap<>();
```
vietnguyen, line 63: This is to make sure `UserDetails.getUserSettings()` never return `null`

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserSettingService.java
```diff
@@ -177,14 +178,14 @@ public Serializable getUserSetting(UserSettingKey key, String username) {
...
-    Map<String, Serializable> result =
-        Sets.newHashSet(getUserSettings(user)).stream()
-            .filter(
-                userSetting ->
-                    userSetting != null
-                        && userSetting.getName() != null
-                        && userSetting.getValue() != null)
-            .collect(Collectors.toMap(UserSetting::getName, UserSetting::getValue));
+    Map<String, Serializable> result = new HashMap<>();
+    Sets.newHashSet(getUserSettings(user)).stream()
```
jbee, line 182: Should use `Set.of` instead of `Sets.newHashSet` but if the next step is to use `stream()` there is no need to "copy" the result of `getUserSettings(user)`, a `stream()` processing is not affecting the iterated collection.

vietnguyen: nice catch, this is an overlook when I fix the null issue of the code.

jbee (approved):

PR #16725, opened 2024-03-07
fix: add search parameter to icons

Birkbjo:
Adds `search`-param to `/icons` API.

This allows searching for a part of the `iconkey`, or partial matches in any keyword.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
@@ -29,6 +29,11 @@ package org.hisp.dhis.icon;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+
+import static org.apache.commons.lang3.StringUtils.wrap;
+import static org.hisp.dhis.commons.util.TextUtils.doubleQuote;
+import static org.hisp.dhis.dataitem.query.shared.StatementUtil.addIlikeReplacingCharacters;
+
 import java.io.IOException;
 import java.sql.SQLException;
 import java.sql.Types;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -210,6 +215,15 @@ private String buildIconQuery(
       parameterSource.addValue("keys", params.getKeys());
     }
 
+    if (params.hasSearch()) {
+      String searchValue = params.getSearch();
+
+      sql += hlp.whereAnd() + " c.iconkey ilike :search";
+      sql += " OR jsonb_path_exists(keywords, '$[*] ? (@ like_regex " + doubleQuote(searchValue) + ")')";
```
Birkbjo, line 222: I had ALOT of problems getting the correct escape params to work here. Whatever I tried I was not able to use the parameter source to put in the searchValue in the `jsonb` query.

I did intent to add start-of-string anchor to this, but I see I forgot about that. Any thoughts about that?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -210,6 +215,15 @@ private String buildIconQuery(
...
+
+      parameterSource.addValue("search", wrap(addIlikeReplacingCharacters(searchValue), '%'));
```
Birkbjo, line 224: not sure if I should use `addIlikeReplacingCharacters`, but it seemed the safest choice. However this is in the `dataitem`-package, I'm not sure how strict you are on including code from other modules like that?

jbee: You might have to add the module dependency if it wasn't already there

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
Birkbjo: Yeah, vscode took care of that

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
Birkbjo: Alright, I managed to add the parameter using `parameterSource`. But I had to add doubleQuotes in the SQL itself for it to work.

PR #16732, opened 2024-03-08
fix: Return 404 instead of 403 when no access [DHIS2-16785]

muilpp:


muilpp:
we'll postpone this until it's clear when to return 404 vs 403

PR #16734, opened 2024-03-08
fix: Do not swallow stack trace from JDK exceptions

larshelge:
This PR is about exception handling. Currently in `CrudControllerAdvice`, general JDK exceptions such as `java.lang.IllegalStateException` and `java.lang.IllegalArgumentException` are caught and returned with a generic "error occurred" message. 

This is not ideal since it will swallow exceptions coming from the JDK or external frameworks, potentially from programming errors, and there is no way to distinguish the two.

For DHIS 2, it is typically better to rely on DHIS 2-specific exceptions to indicate validation failing, etc.

This fix adds stack trace printing for `java.lang.IllegalStateException` and `java.lang.IllegalArgumentException`.

The alternative is to refactor and replace use of these exceptions in DHIS 2, but that will be a huge undertaking.

Replaces use of `IllegalArgumentException` to `IllegalQueryException` in tracked entity service query, which is more appropriate since they signal application level validation error.

Renames variables `exception` to `ex` for consistency.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
@@ -361,7 +361,7 @@ public class CrudControllerAdvice {
 
   @ExceptionHandler(Dhis2ClientException.class)
   @ResponseBody
-  public WebMessage dhis2ClientException(Dhis2ClientException ex) {
+  public WebMessage dhis2ClientExceptionHandler(Dhis2ClientException ex) {
     return conflict(ex.getMessage(), ex.getErrorCode());
   }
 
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -489,14 +489,34 @@ public void servletExceptionHandler(ServletException ex) throws ServletException
...
+   * Handles {@link IllegalArgumentException} and prints the stack trace to standard error. {@link
+   * IllegalArgumentException} is used in DHIS 2 application code but also by various frameworks to
+   * indicate programming errors, so stack trace must be printed and not swallowed.
+   */
+  @ExceptionHandler(IllegalArgumentException.class)
   @ResponseBody
-  public WebMessage handleBadRequest(Exception exception) {
-    return badRequest(exception.getMessage());
+  public WebMessage illegalArgumentExceptionHandler(IllegalArgumentException ex) {
+    ex.printStackTrace();
```
netroms, line 506: Is it better to use the logging framework for this? If the log output is not the same, some log consumers might have trouble parsing it.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -489,14 +489,34 @@ public void servletExceptionHandler(ServletException ex) throws ServletException
...
+
+  /**
+   * Handles {@link IllegalStateException} and prints the stack trace to standard error. {@link
+   * IllegalArgumentException} is used in DHIS 2 application code but also by various frameworks to
+   * indicate programming errors, so stack trace must be printed and not swallowed.
+   */
+  @ExceptionHandler(IllegalStateException.class)
+  @ResponseBody
+  public WebMessage illegalArgumentExceptionHandler(IllegalStateException ex) {
+    ex.printStackTrace();
```
netroms, line 518: Same as above?

netroms (changes requested):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
larshelge: Yes agreed. I did it like this for consistency, as we use `ex.printStackTrace();` for the catch-all error handler `defaultExceptionHandler` further down. Happy to change it. Though we should probably do the same in all these three methods. Thoughts @netroms @jbee ?

david-mackessy (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
netroms: Yeah, I think we should never use e.printStackTrace(), since it will always print directly to stdout, unformatted.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
netroms: Yes

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
larshelge: That's okay. Updated it now. All three places now use `log.error`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
larshelge: That's okay. Updated it now. All three places now use `log.error()`.

mortenoh (approved):

PR #16737, opened 2024-03-10
fix: Make LDAP authentication work again [DHIS2-17036] [39]

larshelge:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -77,18 +85,44 @@ public static CurrentUserDetails getCurrentUserDetails() {
...
-    // This is the state before the user is authenticated.
+    // String principal implies anonymous authentication and the state before user is authenticated
+
     if (principal instanceof String) {
-      if (!"anonymousUser".equals(principal)) {
+      if (!ANONYMOUS_USER.equals(principal)) {
         return null;
       }
 
       return null;
```
larshelge, line 95: This logic a bit redundant.

PR #16747, opened 2024-03-11
fix: SMS gateway configuration - do not require confidential properties on update (2.40)

jbee:
**based on** cherry-pick backport of #16486 and #16393 

Changes after cherry-pick
* ignoring "non functional" changes to files that do not exist in 2.40
* undo Java 17 usages
* remove usages of non-available informational annotations
* add in accidentally deleted check in the service


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/DefaultGatewayAdministrationService.java
```diff
@@ -86,20 +85,15 @@ public void setDefaultGateway(SmsGatewayConfig config) {
...
-
-    if (persisted != null) {
-      return true;
-    }
 
     config.setUid(CodeGenerator.generateCode(10));
 
     SmsConfiguration smsConfiguration = getSmsConfiguration();
+    if (smsConfiguration.getGateways().stream().anyMatch(c -> c.getClass() == config.getClass()))
+      return false;
```
jbee, line 96: FYI: I found that I had removed this check accidentally so I need to re-add it in master/2.41
(the original check was coded differently but this is the essence of the logic)

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/DefaultGatewayAdministrationService.java
```diff
@@ -121,58 +115,62 @@ public boolean addGateway(SmsGatewayConfig config) {
...
-      List<GenericGatewayParameter> updatedList =
-          updatedGenericConfig.getParameters().stream()
-              .filter(GenericGatewayParameter::isConfidential)
-              .collect(Collectors.toList());
-
-      for (GenericGatewayParameter p : updatedList) {
-        if (!isPresent(persistedList, p)) {
-          p.setValue(pbeStringEncryptor.encrypt(p.getValue()));
+    if (persisted instanceof GenericHttpGatewayConfig
+        && updated instanceof GenericHttpGatewayConfig) {
```
david-mackessy, line 146: just a small suggestion, this instanceof check always gets executed even if the instanceof check on line 138 is true. Maybe some reorganisation of the code could allow not checking when we already know the answer.

PR #16755, opened 2024-03-12
chore: replace string concat with java text blocks [DHIS2-16705]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-16965
- Replace String concatenation in analytic queries with Java Text blocks.
- Utilize the function `TextUtils.replace(String,Map)` which allows substitute a variable template `${text}` in the provided String with value from the provided Map.
- Also turned on `undefinedVariableException` which will throw `IllegalArgumentException` if a variable is not defined.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
```diff
@@ -764,36 +761,36 @@ private List<Integer> getDataYears(
...
-            + firstDataYear
-            + " and temp.supportedyear <= "
-            + latestDataYear;
+        replace(
+            """
+            select temp.supportedyear from \
+            (select distinct extract(year from ${eventDateExpression}) as supportedyear \
+            from event psi \
+            inner join enrollment pi on psi.enrollmentid = pi.enrollmentid \
+            where psi.lastupdated <= '${startTime}' and pi.programid = ${programId} \
```
larshelge, line 781: Move the second `and` to a new line.

larshelge (approved):

maikelarabori (approved):

PR #16772, opened 2024-03-13
fix: SMS gateway configuration - do not require confidential properties on update (2.38) 

jbee:
backport based on cherry-pick of #16759

There were couple of conflicting files because in 2.38 the view restriction was still intact. 
I picked the files from 2.39 to resolve this so 2.38 and 2.39 are equal in this regard.


jbee:
Closing this as 2.38 is out of support - the change would not be released and is not particularly important.

PR #16783, opened 2024-03-14
fix: Validate acl access depending on program param [DHIS2-17070]

muilpp:
When requesting a single TE we can't validate the registering org unit, because that enrollment might have been transferred.

There are two possible scenarios here:
- The request specifies a program: In this case, we validate the user has ownership access to the program/org unit pair.
- The request does not specify a program: Then, we need to make sure the user has access (data + metadata) to at least one program. On top of that, we also need to validate ownership or the registering org unit depending on the program. We need to do this because an enrollment might have been transferred and deleted later on, in which case the TE won't have any enrollments but the ownership will still point to the org unit it was transferred to. In this case, we'd only return the TE with its TETA.
 
Ticket: https://dhis2.atlassian.net/browse/DHIS2-17070


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/deprecated/tracker/TrackedEntityInstanceSupportServiceTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class TrackedEntityInstanceSupportServiceTest {
+
+  @Mock private TrackedEntityInstanceService trackedEntityInstanceService;
+
```
lucaCambi77, line 58: I would not add files to the old tracker because of the deprecation, so is there a way to add them to existing tests?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -222,19 +220,9 @@ public TrackedEntity getTrackedEntity(
...
     TrackedEntity trackedEntity;
     if (program != null) {
       trackedEntity = getTrackedEntity(uid, program, params, includeDeleted);
-
-      if (!trackerAccessManager.canRead(currentUser, trackedEntity, program, false).isEmpty()) {
-        if (program.getAccessLevel() == AccessLevel.CLOSED) {
-          throw new ForbiddenException(TrackerOwnershipManager.PROGRAM_ACCESS_CLOSED);
-        }
-        throw new ForbiddenException(TrackerOwnershipManager.OWNERSHIP_ACCESS_DENIED);
-      }
```
lucaCambi77: I'm sorry you're removing this, but can you explain where we are going to check program access? Is this in the program service? so we don't do the acl check anymore?

lucaCambi77:
>  we validate ownership, and if no program is present, we validate the enrollment org unit.
> 

we validate enrollment or te orgunit?

muilpp:
> > we validate ownership, and if no program is present, we validate the enrollment org unit.
> 
> enrollment orgunit is te orgunit?

yes, I changed the description, it was not clear enough

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
muilpp: we do it in the same service, but inside the `getTrackedEntity` method now: https://github.com/dhis2/dhis2-core/pull/16783/files#diff-069b27403f4f267ecb5e14378a2e9c32d339e92acb494e688b821df93de33acbR274
I think it's a bit cleaner this way

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/deprecated/tracker/TrackedEntityInstanceSupportServiceTest.java
muilpp: I didn't find any where I could use `TrackedEntityInstanceSupportService` and where it would make conceptual sense to at the same time. 
This test class has no dependency to the new API, so we should be able to deleted when the time comes without major problems.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
lucaCambi77: ah ok, but that is a private method then, and not implementing the interface

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -269,21 +248,71 @@ public TrackedEntity getTrackedEntity(
...
+      if (program.getAccessLevel() == AccessLevel.CLOSED) {
+        throw new ForbiddenException(TrackerOwnershipManager.PROGRAM_ACCESS_CLOSED);
+      }
+      throw new ForbiddenException(TrackerOwnershipManager.OWNERSHIP_ACCESS_DENIED);
+    }
+
+    return mapTrackedEntity(
+        daoTrackedEntity,
+        params,
+        userService.getUserByUsername(CurrentUserUtil.getCurrentUsername()),
```
teleivo, line 307: We have the `currentUser` already

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -269,21 +248,71 @@ public TrackedEntity getTrackedEntity(
...
+  private TrackedEntity getTrackedEntity(
+      String uid, Program program, TrackedEntityParams params, boolean includeDeleted)
+      throws NotFoundException, ForbiddenException {
+    TrackedEntity daoTrackedEntity = trackedEntityStore.getByUid(uid);
+    addTrackedEntityAudit(daoTrackedEntity, CurrentUserUtil.getCurrentUsername());
+    if (daoTrackedEntity == null) {
+      throw new NotFoundException(TrackedEntity.class, uid);
+    }
+
+    User currentUser = userService.getUserByUsername(CurrentUserUtil.getCurrentUsername());
```
teleivo, line 296: I think it would be clearer what the difference is between having a program and not having one if we split the code into `get, validate, map`. `get` and `map` are AFAIK the same in both cases.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -267,21 +247,69 @@ public TrackedEntity getTrackedEntity(
...
+   *
+   * @return the TE object if found and accessible by the current user
+   * @throws NotFoundException if uid does not exist
+   * @throws ForbiddenException if TE registration org unit not in user's scope or not enough
+   *     sharing access
+   */
+  private TrackedEntity getTrackedEntity(
+      String uid, TrackedEntityParams params, boolean includeDeleted)
+      throws NotFoundException, ForbiddenException {
+    TrackedEntity daoTrackedEntity = getTrackedEntity(uid);
```
teleivo, line 264:
```suggestion
    TrackedEntity trackedEntity = getTrackedEntity(uid);
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -267,21 +247,69 @@ public TrackedEntity getTrackedEntity(
...
+   * Gets a tracked entity based on the program and org unit ownership
+   *
+   * @return the TE object if found and accessible by the current user
+   * @throws NotFoundException if uid does not exist
+   * @throws ForbiddenException if TE owner is not in user's scope or not enough sharing access
+   */
+  private TrackedEntity getTrackedEntity(
+      String uid, Program program, TrackedEntityParams params, boolean includeDeleted)
+      throws NotFoundException, ForbiddenException {
+    TrackedEntity daoTrackedEntity = getTrackedEntity(uid);
```
teleivo, line 285:
```suggestion
    TrackedEntity trackedEntity = getTrackedEntity(uid);
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -267,21 +247,69 @@ public TrackedEntity getTrackedEntity(
...
+        throw new ForbiddenException(TrackerOwnershipManager.PROGRAM_ACCESS_CLOSED);
+      }
+      throw new ForbiddenException(TrackerOwnershipManager.OWNERSHIP_ACCESS_DENIED);
+    }
+
+    return mapTrackedEntity(daoTrackedEntity, params, userDetails, includeDeleted);
+  }
+
+  private TrackedEntity getTrackedEntity(String uid) throws NotFoundException {
+    TrackedEntity daoTrackedEntity = trackedEntityStore.getByUid(uid);
```
teleivo, line 299:
```suggestion
    TrackedEntity trackedEntity = trackedEntityStore.getByUid(uid);
```

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -253,35 +230,123 @@ public TrackedEntity getTrackedEntity(
...
-                .filter(att -> tetAttributes.contains(att.getAttribute().getUid()))
-                .collect(Collectors.toCollection(LinkedHashSet::new));
-        trackedEntity.setTrackedEntityAttributeValues(tetAttributeValues);
+      UserDetails userDetails = CurrentUserUtil.getCurrentUserDetails();
+      trackedEntity =
+          mapTrackedEntity(getTrackedEntityNoDataAcl(uid), params, userDetails, includeDeleted);
+      if (programService.getAllPrograms().stream()
+          .noneMatch(
+              p -> trackerAccessManager.canRead(userDetails, trackedEntity, p, false).isEmpty())) {
+        throw new ForbiddenException(TrackedEntity.class, uid);
```
muilpp, line 239: I'd return `ForbiddenException` instead of `NotFoundException` to avoid breaking the existing API.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -253,35 +233,117 @@ public TrackedEntity getTrackedEntity(
...
+   * @throws ForbiddenException if TE is not accessible
+   */
+  private TrackedEntity getTrackedEntity(String uid) throws NotFoundException, ForbiddenException {
+    TrackedEntity trackedEntity = trackedEntityStore.getByUid(uid);
+    addTrackedEntityAudit(trackedEntity, CurrentUserUtil.getCurrentUsername());
+    if (trackedEntity == null) {
+      throw new NotFoundException(TrackedEntity.class, uid);
+    }
+
+    if (programService.getAllPrograms().stream()
```
muilpp, line 314: @enricocolasante 
`getProgramsByTrackedEntityType` does not validate metadata access, so I think it's better to leave it as it is now and filter by the TET later on. Do you agree?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -107,6 +111,18 @@ public List<String> canWrite(UserDetails user, TrackedEntity trackedEntity) {
   @Override
   public List<String> canRead(
       UserDetails user, TrackedEntity trackedEntity, Program program, boolean skipOwnershipCheck) {
+    List<String> errors = canReadProgramAndTrackedEntityType(user, trackedEntity, program);
+
+    if (!skipOwnershipCheck && !ownershipAccessManager.hasAccess(user, trackedEntity, program)) {
```
lucaCambi77, line 116: I'd say that considering this span for multiple programs with the same TET of the tracked entity, we can make the signature generic to accept a list of programs and iterate inside this method to get the list of errors. Or create another method with this behavior.

lucaCambi77: or at least the _OWNERSHIP_ACCESS_DENIED_ error should indicate what is the program

lucaCambi77, line 116: @muilpp you are saying you are validating the program metadata access. Where is that happening? I see here we are validating the data access of the program

muilpp: the thing with the error message is that in the end we just check if the returned list is empty or not, but we don't use the error messages within. That's also true for this case, where I just want to see if the user has access to at least one program, so the error message itself is not important.
I'd be in favor of start using the error messages, but that'd require a bit of a design, since we might break the API (specially when it comes to breaking the glass).

lucaCambi77 (commented):
@muilpp, the PR is getting really difficult to follow; I'll try to do a review.
Mainly, I have questions about this part:
```

  if (programService.getAllPrograms().stream()
        .anyMatch(
            p ->
                p.getTrackedEntityType() == trackedEntity.getTrackedEntityType()
                    && trackerAccessManager
                        .canRead(getCurrentUserDetails(), trackedEntity, p, false)
                        .isEmpty())) 
```

I don't think we should use `==` to compare objects; if I am not wrong, that checks the object reference in Java. I think we should probably compare the uids.

Another point concerns the consistency of this check. For example, is it a check only for this part, or do you want to make it for other parts where a user wants to access a TE, such as when fetching a list of TEs or in the change log endpoint, etc.?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -253,35 +234,136 @@ public TrackedEntity getTrackedEntity(
...
+                    && trackerAccessManager
+                        .canRead(getCurrentUserDetails(), trackedEntity, p, false)
+                        .isEmpty())) {
+      return trackedEntity;
+    }
+
+    throw new ForbiddenException(TrackedEntity.class, uid);
+  }
+
+  /**
```
enricocolasante, line 330: It's a purely style comment. I would put this java doc in the call of the method so it is even more clear that the workaround is the call to the method itself

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -253,35 +234,136 @@ public TrackedEntity getTrackedEntity(
...
+    if (trackedEntity == null) {
+      throw new NotFoundException(TrackedEntity.class, uid);
+    }
+
+    initializeTrackedEntityOrgUnitParents(trackedEntity);
+
+    if (programService.getAllPrograms().stream()
+        .anyMatch(
+            p ->
+                p.getTrackedEntityType() == trackedEntity.getTrackedEntityType()
```
enricocolasante, line 320:
```suggestion
                Objects.equals(p.getTrackedEntityType(), trackedEntity.getTrackedEntityType());
```

muilpp:
> @muilpp, the PR is getting really difficult to follow; I'll try to do a review. Mainly, I have questions about this part:
> 
> ```
> 
>   if (programService.getAllPrograms().stream()
>         .anyMatch(
>             p ->
>                 p.getTrackedEntityType() == trackedEntity.getTrackedEntityType()
>                     && trackerAccessManager
>                         .canRead(getCurrentUserDetails(), trackedEntity, p, false)
>                         .isEmpty())) 
> ```
> 
> I don't think we should use `==` to compare objects; if I am not wrong, that checks the object reference in Java. I think we should probably compare the uids. Also, what about filtering for programs with registration?
> 
> Another point concerns the consistency of this check. For example, is it a check only for this part, or do you want to make it for other parts where a user wants to access a TE, such as when fetching a list of TEs or in the change log endpoint, etc.?

You are right, I should have used the equals operator, I was wrong to assume I was comparing enums, nice catch!
About filtering programs with registration, we are implicitly doing it while filtering by tracked entity type, since program without registration don't have any.

Regarding the consistency, it's something I'd like to do after I'm done with this PR. As you mentioned, it's getting hard to follow, adding more changes and more tests would make it even harder.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1578,6 +1579,184 @@ void shouldReturnAllEntitiesWhenSuperuserAndModeAll()
...
+    injectSecurityContextUser(user);
+    ForbiddenException exception =
+        assertThrows(
+            ForbiddenException.class,
+            () ->
+                trackedEntityService.getTrackedEntity(
+                    trackedEntityA.getUid(), null, TrackedEntityParams.TRUE, false));
+
+    assertContains(
+        String.format("User has no access to TrackedEntity:%s", trackedEntityA.getUid()),
```
teleivo, line 1756: Maybe a space in between `:` and the UID would make the error more readable. Also seems like you are doing a full comparison in which case assertEquals is what you want. assertContains is nice if you can assert on only a bit of the error message so you can change the message without having to change the test as well.

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1578,6 +1579,184 @@ void shouldReturnAllEntitiesWhenSuperuserAndModeAll()
     assertContainsOnly(Set.of(trackedEntityA.getUid()), uids(trackedEntities));
   }
 
+  @Test
+  void shouldFailWhenRequestingSingleTEAndProvidedProgramDoesNotExist() {
+    String programUid = "made up program";
```
teleivo, line 1584: I would use valid UIDs so we can also use the UID type where possible.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -107,6 +111,18 @@ public List<String> canWrite(UserDetails user, TrackedEntity trackedEntity) {
...
+      errors.add(OWNERSHIP_ACCESS_DENIED);
+    }
+
+    return errors;
+  }
+
+  @Override
+  public List<String> canReadProgramAndTrackedEntityType(
```
teleivo, line 124: When would I use this method over `public List<String> canRead(UserDetails user, TrackedEntity trackedEntity, Program program, boolean skipOwnershipCheck)`?

The signatures are so close 😅 

Also if all we do is `!errors.isEmpty()` we should return a bool.

lucaCambi77:
> > @muilpp, the PR is getting really difficult to follow; I'll try to do a review. Mainly, I have questions about this part:
> > ```
> > 
> >   if (programService.getAllPrograms().stream()
> >         .anyMatch(
> >             p ->
> >                 p.getTrackedEntityType() == trackedEntity.getTrackedEntityType()
> >                     && trackerAccessManager
> >                         .canRead(getCurrentUserDetails(), trackedEntity, p, false)
> >                         .isEmpty())) 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't think we should use `==` to compare objects; if I am not wrong, that checks the object reference in Java. I think we should probably compare the uids. Also, what about filtering for programs with registration?
> > Another point concerns the consistency of this check. For example, is it a check only for this part, or do you want to make it for other parts where a user wants to access a TE, such as when fetching a list of TEs or in the change log endpoint, etc.?
> 
> You are right, I should have used the equals operator, I was wrong to assume I was comparing enums, nice catch! About filtering programs with registration, we are implicitly doing it while filtering by tracked entity type, since program without registration don't have any.
> 
> Regarding the consistency, it's something I'd like to do after I'm done with this PR. As you mentioned, it's getting hard to follow, adding more changes and more tests would make it even harder.

ok i am asking because i am working on the relationships, so let's talk about it if you'll also work on that

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
muilpp: Changed to `assertEquals`.
The space before the UID I didn't want to add because this API response is used in other places and I'm not sure whether there are clients relying on this error message in its current format.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
muilpp: I added a javadoc to both methods, hopefully is a bit clearer.
The difference is `canRead` validates the TE/program ownership, while `canReadProgramAndTrackedEntityType` does not.
Anyway, this is needed now because of the way the let the clients know it's possible to break the glass. This is something we need to fix (there's at ticket for it already). I think we should be able to remove this method when we work on it.

PR #16786, opened 2024-03-14
chore: use UserDetails (in controllers) - 1

jbee:
Uses `UserDetails` instead of `User` where possible (starting from controller layer)


jbee:
Since it has failing tests I think I redo this later when other user improvements have been made

PR #13756, opened 2023-04-28
fix: Filter out from sync teis updated before last sync [DHIS2-15223]

enricocolasante:
When retrieving teis in sync mode, teis that were not updated after last sync were included in the process.
Changed the query to not include teis that were updated before last sync and also consider `skipChangedBefore` parameter to ignore teis that were updated before a specific date

PR #16787, opened 2024-03-14
fix: Filter out from sync teis updated before last sync [DHIS2-15223] [2.37]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-15223

Backport of https://github.com/dhis2/dhis2-core/pull/13756

As described in the issue, we normally do not fix unsupported versions, but in this case, as agreed with @Bekkalizer, we make an exception as requested.


dhis-2/.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+{
```
muilpp, line 1: do we need to add this config file?

lucaCambi77: ah ok i was trying out Visual Studio, so I guess we can also add the ignore on master

PR #16793, opened 2024-03-14
fix: check event in enrollment with no tracked entity

lucaCambi77:
Check whether an enrollment from an event is missing the tracked entity.
This should not happen and currently seems related to test data in play or SL db. However, if an instance has this problem, we report an error by indicating that the system can have corrupted data.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataRelationsValidator.java
```diff
@@ -322,4 +332,28 @@ private Program getEnrollmentProgramFromEvent(TrackerBundle bundle, Event event)
...
+
+    if (enrollment == null) {
+      return !Objects.isNull(
+          bundle
+              .findEnrollmentByUid(event.getEnrollment())
+              .map(org.hisp.dhis.tracker.imports.domain.Enrollment::getTrackedEntity)
+              .orElse(null));
+    }
+
+    return !Objects.isNull(enrollment.getTrackedEntity());
```
muilpp, line 357: just out of curiosity, why do you prefer `Objects.isNull` rather than` == null`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataRelationsValidator.java
```diff
@@ -322,4 +332,28 @@ private Program getEnrollmentProgramFromEvent(TrackerBundle bundle, Event event)
     }
     return null;
   }
+
+  /**
+   * Given a program with registration, check whether an event in the preheat or payload refers to
+   * enrollment with an existing tracked entity. This should not happen because an enrollment
+   * without a tracked entity is not allowed, see {@link
```
muilpp, line 339: I find this sentence a bit confusing.
I think it would be more clear to say something like this:
_It should always find a tracked entity, because an enrollment without a tracked entity is not allowed_

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityOwnershipValidator.java
```diff
@@ -80,7 +79,10 @@ class SecurityOwnershipValidator implements Validator<org.hisp.dhis.tracker.impo
   @Nonnull private final OrganisationUnitService organisationUnitService;
 
   private static final String ORG_UNIT_NO_USER_ASSIGNED =
-      " has no organisation unit assigned, so we skip user validation";
+      "Event %s has no organisation unit assigned, so we skip user validation";
```
muilpp, line 82: when would an event have no org unit assigned? in case it's an event program?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataRelationsValidatorTest.java
```diff
@@ -750,6 +757,113 @@ void eventValidationFailsWhenEventAOCAndEventCOsAreSetAndInProgramCCButNotAllCOs
...
   }
 
+  @Test
+  void eventOfProgramWithRegistrationInEnrollmentWithoutTrackedEntity() {
+    OrganisationUnit orgUnit = organisationUnit(ORG_UNIT_ID);
+    when(preheat.getOrganisationUnit(MetadataIdentifier.ofUid(ORG_UNIT_ID))).thenReturn(orgUnit);
+
+    Program program = setupProgram(orgUnit);
+
+    CategoryCombo defaultCC = defaultCategoryCombo();
```
muilpp, line 767: the setups of these tests seems to be quite similar when it comes to category combos, could we have a helper method initialize it?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityOwnershipValidator.java
lucaCambi77: yes correct that is just a refactor to align with the new constant I've added

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataRelationsValidator.java
lucaCambi77: not a particular reason, it looks more compact to me. Also, in the previous if using the `==null` made the code difficult to read

muilpp (approved):

teleivo (approved):

PR #16794, opened 2024-03-15
fix: broken ldap login

netroms:
## Summary
Fixes broken LDAP login and makes the new AuthenticationController compatible with LDAP authentication.


mortenoh (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/ldap/authentication/DhisBindAuthenticator.java
```diff
@@ -58,19 +58,20 @@ public DhisBindAuthenticator(BaseLdapPathContextSource contextSource, UserStore
...
+    if (user.hasLdapId() && user.isExternalAuth()) {
       log.debug(
           "LDAP authentication attempt with username: '{}' and LDAP ID: '{}'",
           user.getUsername(),
           user.getLdapId());
       authentication =
           new UsernamePasswordAuthenticationToken(
               user.getLdapId(), authentication.getCredentials());
+    } else {
+      throw new BadCredentialsException("Incorrect user credentials");
```
vietnguyen, line 74: Do we need to do log.error for `user.hasLdapId()` and  `user.isExternalAuth()` before throwing the exception ?

larshelge: Yes that is a good idea @vietnguyen .

larshelge: I think we can simply include the username in the exception message. That will make it clear which user it relates to, and the lack of LDAP ID / external auth will be implicit.

PR #16798, opened 2024-03-15
perf: default icon insert with single lookup when all exist [DHIS2-16872]

jbee:
Splits the creation of default icons in two service methods
1. find the missing icons
2. insert each missing item with tracking

This way we have detailed tracking and error handling for the insert but only a single query to find that no work needs to be done when all icons were inserted. 

Also the creation of the file resource is done in an independent transaction before the icon is created using the same `addIcon` method that we use for custom icons.

Last but not least the `Icon` `key`,`name` and `storageKey` for the created `FileResource` are all aligned to use the `DefaultIcon#key` and domain to create the unique values. This way it is way easier when looking at the file resource table to tell what resource belongs to which icon. 

Also should an icon insert fail for some reason it is not re-attempted until the next restart of the server to avoid endless tries.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/JdbcIconStore.java
```diff
@@ -78,6 +80,21 @@ public class JdbcIconStore implements IconStore {
...
   private final FileResourceStore fileResourceStore;
 
+  @Override
+  public List<String> getKeysWithoutIcon(Set<String> keys) {
+    String sql =
+        """
+    SELECT ARRAY(
+                   SELECT UNNEST(ARRAY [:keys])
+                   except
+                   SELECT icon.iconkey from icon)""";
```
jbee, line 90: I am no SQL guru - this is the best I could find to return only those keys that do not have a row in icon

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -67,33 +71,48 @@ public class DefaultIconService implements IconService {
...
+  }
+
+  @Override
+  @Transactional
+  public String uploadDefaultIcon(Icon icon) throws ConflictException {
+    String fileResourceId = CodeGenerator.generateUid();
+    Resource resource = getDefaultIconResource(icon.getKey());
+    try {
+      FileResource fileResource =
+          FileResource.ofKey(CUSTOM_ICON, icon.getKey(), MediaType.IMAGE_PNG);
```
teleivo, line 104: Why is `ofKey` called with `CUSTOM_ICON` in the method `uploadDefaultIcon`? And why `png`? At least right now default icons like `https://play.im.dhis2.org/dev/api/42/icons/clinical_fe_positive/icon` are served as svg 🤔

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/IconStore.java
```diff
@@ -71,12 +70,9 @@ public interface IconStore {
   List<Icon> getIcons(IconQueryParams params);
 
   /**
-   * Check if the store contains entries for all provided keys
-   *
-   * @param keys a set of keys
-   * @return true, if all exist
+   * @return all existing keys
```
teleivo, line 73:
```suggestion
   * @return all existing keys (including variants)
```

I am constantly thinking about if what I am looking at is the normalized key or the key of a variant. Not sure if someone else has the same difficulty 😅

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -130,15 +149,10 @@ public boolean iconExists(String key) {
   public void addIcon(@Nonnull Icon icon)
       throws BadRequestException, NotFoundException, SQLException {
 
-    if (!icon.isCustom()) {
+    if (icon.getOrigin() == null && !icon.isCustom()) {
```
teleivo, line 152: It's not obvious what this `origin` field has to do with an icon being custom or not. This is only set for default icons right? Can this then be moved into the `isCustom`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -67,33 +71,48 @@ public class DefaultIconService implements IconService {
...
-                  fileResourceId, MediaType.IMAGE_PNG, resource.getInputStream(), CUSTOM_ICON);
-
-          i.setFileResource(fileResource);
-
-          addIconInternal(i);
-        } catch (Exception ex) {
-          throw new RuntimeException(ex);
+  @Transactional(readOnly = true)
+  public List<Icon> findNonExistingDefaultIcons() {
+    Set<String> existing = Set.copyOf(iconStore.getAllKeys());
```
teleivo, line 82: We could directly return a `Set` from `getAllKeys` as this is what we need.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
```diff
@@ -27,515 +27,454 @@
...
+    this.keywords = Set.of(keywords);
+    this.variants =
+        Set.of("positive", "negative", "outline").stream()
+            .map(variant -> "%s_%s".formatted(key, variant))
+            .toList();
   }
 
-  public Set<String> getKeywords() {
-    return Arrays.stream(this.keywords).collect(Collectors.toSet());
+  public List<Icon> createIcons() {
```
teleivo, line 475:
```suggestion
  public List<Icon> getIconsWithVariants() {
```

or `toIconsWithVariants` 🤷🏻 

Maybe this is only me but `create` reminds me of persisting. Please resolve this if we never use `create` in that regard in our code.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
```diff
@@ -67,33 +71,48 @@ public class DefaultIconService implements IconService {
...
+          FileResource.ofKey(CUSTOM_ICON, icon.getKey(), MediaType.IMAGE_PNG);
+      fileResource.setUid(fileResourceId);
+      fileResource.setAssigned(true);
+      try (InputStream image = resource.getInputStream()) {
+        fileResourceService.syncSaveFileResource(fileResource, image);
+      }
+      icon.setFileResource(fileResource);
+      return fileResourceId;
+    } catch (IOException ex) {
+      ignoredAfterFailure.add(icon.getOrigin());
```
teleivo, line 113: Why are we ignoring these and not retrying later?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/HousekeepingJob.java
```diff
@@ -94,10 +94,22 @@ public void execute(JobConfiguration config, JobProgress progress) {
...
+    List<Icon> missing = progress.runStage(List.of(), iconService::findNonExistingDefaultIcons);
+    progress.startingStage("Insert default icons", missing.size(), SKIP_ITEM);
+    progress.runStage(missing, Icon::getKey, this::createDefaultIcon);
 
     progress.completedProcess(null);
   }
+
+  private void createDefaultIcon(Icon icon) {
+    try {
+      // note that these are split in two to have independent TX boundaries
```
teleivo, line 107: The why is not clear to me from this comment.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
teleivo: How would a user notice this error and potentially fix it?

netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
jbee: I don't know :) I just kept the arguments as there were earlier just passing them to a new method.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
jbee: I was back on forth on this one. I think it makes sense to expect a sorted response from a all values lookup. I just need fast lookup so I made it into a set later.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/icon/DefaultIcon.java
```diff
@@ -27,515 +27,454 @@
...
 
-  private final String[] keywords;
-
-  DefaultIcon(String key, String description, String[] keywords) {
+  DefaultIcon(String key, String description, String... keywords) {
     this.key = key;
     this.description = description;
-    this.keywords = keywords;
+    this.keywords = Set.of(keywords);
+    this.variants =
```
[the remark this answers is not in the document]
jbee: Good point, I shall rename :)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
jbee: Because it is likely this keeps failing for the same reason causing an infinite sequence of tries.
This is a very simple way to avoid that.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/icon/DefaultIconService.java
teleivo: Interesting 😅

PR #16804, opened 2024-03-18
feat: custom icons [DHIS2-16872]

jbee:
chery-pick backport from #16570 and #16798 and #16805


zubaira (approved):

jbee:
NOTE: this is waiting for some more name changes Ivo is working on. These need to be consistent for the versions.

PR #16806, opened 2024-03-18
fix: broken login when user has org units associated

netroms:
### Summary
Fixes issue during user details creation, when hydrating Hibernate managed OrgUnits to populate the UserDetails object.
Since there are many calls to `CurrentUserUtil.getCurrentUserDetails()` in the getters themselves in several places, this will cause an exception since we are not logged in yet. Since we only need the OrgUnit IDs during creation of the UserDetails, separate data store/SQL is made to fetch only the UIDs.

Also makes some startup data populators use 'current user safe' service access methods, i.e. running as SystemUser.

## Automatic test
OrganisationUnitStoreTest#testOrganisationUnitsUidByUser()


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -171,8 +188,14 @@ default boolean isInUserDataHierarchy(String orgUnitPath) {
...
-    if (orgUnitPath == null) return false;
-    for (String uid : orgUnitPath.split("/")) if (orgUnitIds.contains(uid)) return true;
+    if (orgUnitPath == null) {
+      return false;
+    }
+    for (String uid : orgUnitPath.split("/")) {
+      if (orgUnitIds.contains(uid)) {
+        return true;
+      }
+    }
```
jbee, line 198: Do you have your IDE configured to insert brackets for all block statements?

jbee (commented):
Do I understand correctly that the issue is that a `User` used to create a `UserDetails` might be detached and therefore not correctly populated when it comes to these lists so therefore you changed this to query it directly?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
netroms: Sorry about that, it might be my IDE yeah. I will revert it.

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/organisationunit/hibernate/HibernateOrganisationUnitStore.java
```diff
@@ -86,6 +86,35 @@ public HibernateOrganisationUnitStore(
...
+
+  private static String getOrgUnitTablesUids(String username, String orgUnitTableName) {
+    return """
+        SELECT ou.uid
+        FROM organisationunit ou
+                 JOIN %s um ON ou.organisationunitid = um.organisationunitid
+                 JOIN userinfo ui ON um.userinfoid = ui.userinfoid
+        WHERE ui.username = '%s';
+        """
+        .formatted(orgUnitTableName, username);
```
vietnguyen, line 115: maybe we could have one single query for those properties.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -909,8 +910,21 @@ public UserDetails createUserDetails(User user) {
 
     Map<String, Serializable> userSettings = userSettingService.getUserSettingsAsMap(user);
 
+    List<String> organisationUnitsUidsByUser =
+        organisationUnitService.getOrganisationUnitsUidsByUser(user.getUsername());
+    List<String> searchOrganisationUnitsUidsByUser =
```
vietnguyen, line 915: the user.organisationUnit settings should not be updated regularly, maybe we could use a cache here.

PR #16855, opened 2024-03-20
fix: Do not set created and lastupdated on analytics datavalueset export

larshelge:


jason-p-pickering:
This is a fix of https://github.com/dhis2/dhis2-core/pull/16483 recommended by @larshelge . Instead of setting the `storedBy` fields to "" (blank) set the explicitly to `null`. Also leave out the created and lastUpdated. The effect of this should be to use the `currentUser` of user who imports the data on the target server side. The `created` and `lastUpdated` fields should then also default to the server time instead of what is sent by the client.

david-mackessy (approved):
should the analytics team be also added as a reviewer seeing as it's an analytics PR?

jason-p-pickering:
Tagging @dhis2/analytics-backend for review.

jbee (approved):

PR #16879, opened 2024-03-21
chore: Bump flyway version and upgrade Postgres version to 13

enricocolasante:
- Ignoring flyway upgrades for version 2.41
- Import of `flyway-database-postgresql` dependency is needed instead of `flyway-core` but as the former is not directly used in our code and the latter is, we need to update the list of `usedUndeclared` and `unusedDeclared` dependencies in the `pom.xml` file
- Upgrade Postgres version to 13 (minimum version suggested) used by TestContainers for integration tests and in Docker image used by e2e tests


david-mackessy (approved):

muilpp (approved):

gnespolino (approved):
lgtm

stian-sandvold (approved):

PR #16880, opened 2024-03-21
fix: Validate user has access to relationship type [DHIS2-17082]

muilpp:
When the user tries to create, update or delete a relationship, we are not validating its access to the relationship type.
This PR adds a `SecurityOwnershipValidator`, as done in TE, enrollments and events validators, to check the aforementioned access.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-17082


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/relationship/RelationshipValidator.java
```diff
@@ -53,7 +53,8 @@ public RelationshipValidator() {
                 new MetaValidator(),
                 new LinkValidator(),
                 new ConstraintValidator(),
-                new DuplicationValidator()));
+                new DuplicationValidator(),
+                securityOwnershipValidator));
```
enricocolasante, line 57: I would put this new validator right after the `MetaValidator`.
As soon as it is validated that the relationship type exists we should check if the user has access to it.
Running this validator as the last one could potentially disclose some information that the user shouldn't see, for example, the user shouldn't be informed that there is a duplicated relationship if it has no access to the relationship type.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/mappers/RelationshipMapper.java
```diff
@@ -64,5 +64,6 @@ public interface RelationshipMapper extends PreheatMapper<Relationship> {
   @Mapping(target = "name")
   @Mapping(target = "attributeValues")
   @Mapping(target = "bidirectional")
+  @Mapping(target = "sharing")
```
teleivo, line 67: Why are we not doing this elsewhere @enricocolasante ? And why do the other security validators work 🤷🏻 We discussed this a while back but I don't remember what the outcome was.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/relationship/SecurityOwnershipValidatorTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Test
+  void shouldFailWhenUserHasNoAccessToRelationshipType() {
+    when(aclService.canDataWrite(bundle.getUser(), relationshipType)).thenReturn(false);
+
+    validator.validate(reporter, bundle, relationship);
+
+    assertNotEmpty(reporter.getErrors());
+    assertContains(
+        "has no data write access to relationship type", reporter.getErrors().get(0).getMessage());
```
teleivo, line 90: Please use the nice assertions we have like

```java
assertHasError(reporter, enrollment, ValidationCode.E1014);
```

in org.hisp.dhis.tracker.imports.validation.validator.AssertValidations. Check other Validator tests. We should also have some now for no errors.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/RelationshipImportTest.java
```diff
@@ -80,6 +82,17 @@ void successImportingRelationships() throws IOException {
     assertThat(importReport.getStats().getCreated(), is(2));
   }
 
+  @Test
+  void shouldFailWhenUserNotAuthorizedToCreateRelationship() throws IOException {
+    userA = userService.getUser("o1HMTIzBGo7");
+    TrackerImportParams params = TrackerImportParams.builder().userId(userA.getUid()).build();
+    ImportReport importReport =
+        trackerImportService.importTracker(params, fromJson("tracker/relationships.json"));
+    assertThat(importReport.getStatus(), is(Status.ERROR));
```
teleivo, line 91: that is already asserted by assertHasError

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/mappers/RelationshipMapper.java
enricocolasante: I am not sure I understand what you mean by elsewhere.
We are mapping `sharing` for other metadata, like `TrackedEntityType` or `Program`

muilpp: I think @teleivo meant why are we not using the `RelationshipMapperType` like we do in the `TrackerMapper` for the TET.
I refactored this in the last commit: https://github.com/dhis2/dhis2-core/pull/16880/commits/d921f3d4b014bde82d42d636525046b8fdfdd77c

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/relationship/SecurityOwnershipValidatorTest.java
```diff
@@ -70,25 +73,25 @@ public void setUp() {
...
     reporter = new Reporter(idSchemes);
     relationship = new Relationship();
+    relationshipType = new RelationshipType();
+
+    when(bundle.getPreheat().getRelationshipType(any())).thenReturn(relationshipType);
+    when(bundle.getStrategy(relationship)).thenReturn(CREATE);
   }
 
   @Test
   void shouldFailWhenUserHasNoAccessToRelationshipType() {
```
lucaCambi77: Are we covering the different test cases? I understand this validation happens when the strategy is update/delete/create. Also, I see the integration and JSON cover only the negative case, but I don't have a full vision of what is already present

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/relationship/SecurityOwnershipValidator.java
```diff
@@ -0,0 +1,69 @@
...
+import org.hisp.dhis.tracker.imports.domain.Relationship;
+import org.hisp.dhis.tracker.imports.validation.Reporter;
+import org.hisp.dhis.tracker.imports.validation.ValidationCode;
+import org.hisp.dhis.tracker.imports.validation.Validator;
+import org.springframework.stereotype.Component;
+
+@Component(
+    "org.hisp.dhis.tracker.imports.validation.validator.relationship.SecurityOwnershipValidator")
+@RequiredArgsConstructor
+@Slf4j
```
teleivo, line 46: No need to add the logging dependency if you don't log.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/RelationshipImportTest.java
teleivo: `assertThat(importReport.getStatus(), is(Status.ERROR));` is already asserted by `assertHasError`. Can you please remove that line.

And add newline between setup, method under test and assertions. So it's easy to spot them.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/mappers/RelationshipMapper.java
teleivo: Perfect! Now it's clear that we map the sharing for metadata. Before I thought we map it now for data which we do not do elsewhere.

teleivo (approved):

PR #16882, opened 2024-03-21
fix: Use appropriate JDBC template with read replicas [DHIS2-17062]

david-mackessy:
# Issue
When DHIS2 is set up using a Postgres read replica DB, any SQL view (which has an underlying write) will be retrieved using the `readOnlyJdbcTemplate`, which causes an error `cannot execute ... in a read-only transaction`.

# Cause
The method `HibernateSqlViewStore#populateSqlViewGrid` always only ever used the `readOnlyJdbcTemplate`. This never causes any issues when DHIS2 is set up to only use a primary Postgres DB (no read replica DBs) as the `readOnlyJdbcTemplate` falls back to using the normal `jdbcTemplate` when no read replica DBs are found at startup.
If a read replica DB is detected at startup, then the `readOnlyJdbcTemplate` would be used and would target the read replica DB, wherever it is used in the code.

This is only an issue when sql views perform writes underneath (e.g. a select query that calls a DB function, which in turn creates a temp table).

This issue would only ever be seen when the following config is in place:
- using read only replicas
- using sql views that perform writes (`system.sql_view_write_enabled=on`)

# Fix
In order to use the correct jdbc template, when we know the SQL view has an underlying write, then we pass the enum `TransactionType.WRITE` to the store so the normal `jdbcTemplate` can perform the operation.
When we know the SQL view is read only, then we pass the enum `TransactionType.READ` so the `readOnlyJdbcTemplate` is used.

# Testing
Because of the complicated set up involving a read replica DB, this is extremely difficult to test automatically. It suits a manual test, preferably by the reporter of the issue.

## Automated
Simple mock test added to ensure the correct template is used based on the enum value passed in.

## Manual
Set up DHIS2 to use a read replica DB - [see guide](https://github.com/dhis2/wow-backend/blob/master/guides/postgres/postgres_read_replica.md)
Then follow the steps in the Jira to check if issue has been resolved.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/TransactionType.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/TransactionType.java
@@ -1,0 +1,6 @@
+package org.hisp.dhis.common;
+
+public enum TransactionType {
+  READ,
+  WRITE
+}
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/TransactionType.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/TransactionType.java
@@ -1,3 +1,30 @@
+/*
+ * Copyright (c) 2004-2024, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/TransactionType.java
```diff
@@ -0,0 +1,33 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.common;
+
+public enum TransactionType {
```
larshelge, line 30: Looks nice. I think the term "transaction mode" / `TransactionMode` is slightly more established and descriptive than "type".

dhis-2/dhis-api/src/main/java/org/hisp/dhis/sqlview/SqlViewStore.java
```diff
@@ -40,7 +41,17 @@ public interface SqlViewStore extends IdentifiableObjectStore<SqlView> {
 
   void dropViewTable(SqlView sqlView);
 
-  void populateSqlViewGrid(Grid grid, String sql);
+  /**
+   * This method will use the appropriate jdbcTemplate depending how DHIS2 has been setup.<br>
+   * <br>
```
larshelge, line 46: Use `<p>` instead of br for Javadoc paragraphs.

vietnguyen (approved):

netroms (approved):

PR #16915, opened 2024-03-22
fix: code format in analytic queries [DHIS2-16705]

vietnguyen:
- Java text block ignore trailing space so we need to put `\s` to preserve the space.


maikelarabori (approved):
I'm approving it, but I'm not sure this is a good idea to start adding such characters on SQL statements... The SQL can be formatted by a text editor.

larshelge (approved):

larshelge:
Yep agreed @maikelarabori . The `\s` represents a whitespace character in Java 15 blocks, not newline. This is more to comply with how to original code was written. We should refactor to a more elegant solution.

PR #16934, opened 2024-03-25
fix: icon exists checks file resource and content [DHIS2-16872]

jbee:
### Summary
Main change is to handle state inconsistency between `Icon`, `FileResource` and file content as 404 or 400 exception (depends on case) and to auto-repair default icons which link to a non-existent file resource.

Further improvements:
* move validation into service layer
* remove unnecessary declaration of `throws SQLException`
* remove unnecessary mapping/copying of input params
* add `@Nonnull` and `@CheckForNull` annotations in APIs
* add `required = true` to `@JsonProperty` where suitable
* rename params classes to follow name patterns used elsewhere
* use params objects for add/update server layer input instead of entity object (avoids having to check stable properties are indeed unchanged, avoids stale state objects)
* add limits to user provided input for `page` and `pageSize`
* move mock tests to be integration tests (too much hassle to fix the test setup each time something changes slightly)
* fix general test setup to use synchronous file store for `FileResource` content in tests (no scheduling) so that the content is actually available during the test as this is now checked when an icon is fetched (`getIcon(key)`)
* removed duplication in code 
* only set image as assigned when the icon is saved successful

### Automatic Testing
The mock tests were moved to integration tests to not have to deal with fixing test setup each time something changes slightly. 

### Manual Testing
* open maintenance app, select _Icons_ card
* add a custom icon 
* update the icon
* delete the icon
* browse to `/dhis-web-maintenance/index.html#/edit/programSection/program/IpHINAT79UW` as an example for an object that uses icons, click the change icon button and play with the selection dialog


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
```diff
@@ -207,10 +212,13 @@ public FileResource saveFileResource(String uid, MultipartFile file, FileResourc
...
 
     if (uid != null && fileResourceService.fileResourceExists(uid)) {
-      throw new WebMessageException(
-          conflict(ErrorCode.E1119, FileResource.class.getSimpleName(), uid));
+      throw new ConflictException(ErrorCode.E1119, FileResource.class.getSimpleName(), uid);
+    }
+    if (!jobRunner.isScheduling()) {
+      fileResourceService.syncSaveFileResource(fileResource, Files.readAllBytes(tmpFile.toPath()));
+    } else {
+      fileResourceService.asyncSaveFileResource(fileResource, tmpFile);
```
jbee, line 220: FYI: This is a general fix to make sure file content is available in integration tests, the asynchronous save was not happening in test setup but at the same time I did not want to change this for prod as it might affect many callers.

netroms (approved):

vietnguyen (approved):

PR #16955, opened 2024-03-27
feat: Analytics enrollment endpoint: remove the use of default period for empty enrollment/incident date[DHIS2-17103]

d-bernat:
TBD

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsPeriodCriteriaUtils.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsPeriodCriteriaUtils.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2004-2024, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsPeriodCriteriaUtils.java
```diff
@@ -0,0 +1,80 @@
...
+    Date startDate =
+        Date.from(
+            LocalDate.of(Collections.min(availableYears), 1, 1)
+                .atStartOfDay(ZoneId.systemDefault())
+                .toInstant());
+
+    Date endDate =
+        Date.from(
+            LocalDate.of(Collections.max(availableYears), 12, 31)
+                .atStartOfDay(ZoneId.systemDefault())
```
gnespolino, line 74: are you sure it should be `atStartOfDay` ?

gnespolino (approved):

maikelarabori (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/AnalyticsPeriodCriteriaUtils.java
d-bernat: yep, for example 2027-12-31  will be translated into the sql as this: "< 2028-01-01"

PR #16958, opened 2024-03-27
fix: Analytics enrollment endpoint: remove the use of default period for empty enrollment/incident date [2.41.0-backport] [DHIS2-17103]

d-bernat:
**[NEED RCB APPROVAL]**

BACKPORT: The analytics API supports the period dimension, and the related query parameter is mandatory. New feature implementation removes this restriction for the enrollment/query entry point. When the period dimension is not included in the parameter set, the default rules are applied, wherein the start date and end date are equal to the maximum interval used for analytics export.


maikelarabori (approved):

gnespolino (approved):

janhenrikoverland:
Not a regression, been like this since 38, so will not get approved by RCB because it can wait for 2.41.1

PR #16959, opened 2024-03-27
fix: Tracker - Default Json response to browser only html requests [DHIS2-13014]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-13014

Currently, if we make a rest request to a tracker endpoint that supports multiple content types, such as events or tracked entities, the request doesn't default to JSON when it comes from a browser with a generic path like `/tracker/events`. It appears to consume the CSV endpoint, which is totally random; it could also be the gzip or zip endpoint. The need for defaulting to JSON is that other endpoints like CSV do not support pagination and could potentially run heavy requests to the server.

This issue happens because a browser uses an `Accept` header related to HTML and a null `Content-Type`.
We can see this using the `@RequestHeader("Accept")` in an endpoint. Here is an example of the `Accept` header from Chrome or Brave:

```
text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
```

This happens even if we have correctly set up the [ContentNegotationManager](https://github.com/dhis2/dhis2-core/blob/914433ffcd4e8c5e4e59e89bdeaf042c7e9d6a9d/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/WebMvcConfig.java#L222) which should default the content response to `application/json` by using a `FixedContentNegotiationStrategy`. 
However, this relates only to the Content-Type, and it is also useful when, for example, we use the suffix in the path (e.g.,`/event.csv`) but appears to be usless with the `Accept` header, which is free to be any value and seems to have precedence in defining the request. 

### Fix
One way to fix this is to apply the [headers](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/GetMapping.html#headers()) attribute in the GET mapping and set it to a sensible value to intercept the browser request. Then, we can use the spring `ResponseEntity` to set the response's content type because the browser request comes with a **null** `Content-Type`. We extend this to all tracker endpoints.

E2e test added. Manual tests were done with Chrome and Brave.

### Alternatives
I couldn't find any alternative to this. Could there be ways to configure the web layer, intercept the requests, or set a different relevance for some request attributes? I have yet to find any, but fixing this at a configuration level might be something we want if we already set a default JSON content type, although this seems related to the tracker endpoints because a user might want to download payloads for the export using a browser URL. Also, setting this as a general rule might affect other endpoints now or in the future, making it difficult to spot the problem.

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/export/TrackerExportTest.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/export/TrackerExportTest.java
@@ -47,6 +47,7 @@ import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.startsWith;
 import static org.hisp.dhis.helpers.matchers.MatchesJson.matchesJSON;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.google.gson.JsonObject;
 import io.restassured.http.Header;
@@ -572,7 +573,7 @@ public class TrackerExportTest extends TrackerApiTest {
             null,
             new Headers(new Header(HttpHeaders.ACCEPT, "text/html")));
 
-    assertEquals("application/csv; charset=UTF-8", response.getContentType());
+    assertTrue(response.getContentType().contains("application/csv"));
   }
 
   @Test
@@ -598,7 +599,7 @@ public class TrackerExportTest extends TrackerApiTest {
             null,
             new Headers(new Header(HttpHeaders.ACCEPT, "text/html")));
 
-    assertEquals("application/csv; charset=UTF-8", response.getContentType());
+    assertTrue(response.getContentType().contains("application/csv"));
   }
 
   @Test
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/export/TrackerExportTest.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/export/TrackerExportTest.java
@@ -76,6 +76,8 @@ import org.skyscreamer.jsonassert.JSONAssert;
  * @author Gintare Vilkelyte <vilkelyte.gintare@gmail.com>
  */
 public class TrackerExportTest extends TrackerApiTest {
+  private static final String DEFAULT_JSON_CONTENT_TYPE_WITH_HTML_REQUEST =
+      "%s do not default to application/json format when the Accept header is html";
   private static final String TE = "Kj6vYde4LHh";
 
   private static final String TE_POTENTIAL_DUPLICATE = "Nav6inZRw1u";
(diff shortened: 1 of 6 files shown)


enricocolasante (commented):
Looks good.
Maybe one missing test is to verify that when sending the accept header `text/html` and the URL ends in `.csv` the response is in csv format.

enricocolasante (approved):

teleivo (commented):
> This issue happens because a browser uses an Accept header related to HTML and a null Content-Type.

Might be a bit of a misunderstanding involved in that the `Content-Type` will only be set if the request/response actually sends content.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

So when you make a GET from the browser to one of these URLs there is no content and hence no `Content-Type` header.

[ContentNegotationManager](https://github.com/dhis2/dhis2-core/blob/914433ffcd4e8c5e4e59e89bdeaf042c7e9d6a9d/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/WebMvcConfig.java#L222) is involved in finding the content to be returned/which controller will do it. `Content-Type` header and `Accept` header and more play a role in finding the controller.

`defaultContentType` AFAIK is only used if there is no `Accept` header. So this cannot solve our issue as this mechanism is not supposed to interferE if there is an `Accept` header.

I think we should solve this for all endpoints. We need to fully understand all Spring options of which there are many. This to me is similar to us implementing gzip/zip support on endpoints individually.

lucaCambi77:
ok thanks @teleivo  considering the pressing release I think this could be a solution. As i said i haven't found an alternative and happy if there is a more general  solution for that and also the compression part

teleivo:
> ok thanks @teleivo considering the pressing release I thought this could be a solution. As i said i haven't found an alternative and happy if there is a more general solution for that and also the compression part

https://dhis2.atlassian.net/browse/DHIS2-13014 is an old bug.

I think we should fix 

> The new tracker `api/tracker/events` endpoint throws a 500 error when called without any parameters.

ASAP in all supported versions.

I see

> The default returning format should be JSON.

as a separate issue. We should take our time and find a solution spanning endpoints. Otherwise, this is something we have to do again and again. @enricocolasante what is your opinion?

jbee (commented):
I would be in favor of adding another endpoint method that is in particular for this default. In its implementation it can just delegate to the JSON method. This way it becomes more obvious that the mapping is a solution to match for _Accept_ headers that do not contain any of the actually supported output format. Also this could be seen as a client error. A client should ask with one of the supported _Accept_ header OR using a file extension that contains the same information. I am pretty sure in many of our API we do end up with "unsupported media type" if clients were to ask without being specific of what they want and there being more than one method matching the path. Even more generally we said we want to avoid mapping the same path to different endpoint. I guess in this case it is too late and the API is already public and used like this?

lucaCambi77:
@jbee yes this is currently in use, and also, as discussed with Mozafar, it would be out of the scope now to create another endpoint because the path extension is the expected behavior from the old tracker. I've created https://dhis2.atlassian.net/browse/DHIS2-17208 so there will be a chance to refine this in detail. I am not going to add anything to this PR as this is already a minimal change

jbee (approved):
As discussed on slack given the options this is a good solution.

The main concern is that the workaround is not easily understood - why it works or what issue it tries to work around but I hope we can recover that from the ticket and PR should it be needed.

PR #16965, opened 2024-03-27
fix: Tracker event query is slow

lucaCambi77:


enricocolasante (approved):
Looks good to me :)
It would be good to have @larshelge input on it

muilpp (approved):

PR #16973, opened 2024-03-28
fix: AbstractJdbcTableManager.swapTable  does not swap object Table name

vietnguyen:
### Issue
- Currently if User run Analytic with number of years != All then below error will be thrown

```
drop table if exists "analytics_2023" cascade; alter table "analytics_2023_temp" rename to "analytics_2023";

alter table "analytics_2023_temp" no inherit "analytics_temp"; alter table "analytics_2023_temp" inherit "analytics";

ERROR: relation "analytics_2023_temp" does not exist
```

- Issue is the table name in database is changed but the name of the Table object is not changed in method `AbstractJdbcTableManager.swapTable()`. Hence, the method `swapInheritance` takes in the old table name.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/model/AnalyticsTable.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/model/AnalyticsTable.java
@@ -31,6 +31,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import lombok.Getter;
+import org.apache.commons.lang3.StringUtils;
 import org.hisp.dhis.analytics.AnalyticsTableType;
 import org.hisp.dhis.commons.collection.UniqueArrayList;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/model/AnalyticsTable.java
```diff
@@ -252,9 +253,17 @@ public AnalyticsTablePartition getLatestTablePartition() {
         .orElse(null);
   }
 
+  // implement toString method
+
   @Override
   public String toString() {
-    return "[Table name: " + getName() + ", partitions: " + tablePartitions + "]";
```
vietnguyen, line 257: This causes StackOverFlow exception as the `tablePartitions` will trigger infinite loop calls to `AnalyticsTable.toString()` method.

gnespolino (approved):

maikelarabori (approved):

PR #16980, opened 2024-03-28
feat: OpenAPI annotation based description for endpoint, parameters, responses

jbee:
### Summary
Allows to annotate endpoints with `@OpenApi.Description` to provide a description text

When annotating...
* method => endpoint description
* parameter => request body and parameter description
* return type => success response description
* exception type => error response description

As before a description text can be provided using markdown files with descriptions from the annotation taking precedence. 
The thought behind this precedence order is that something occurring in code should yield the expected outcome (principle of least surprise). 

### Improved JSON formatting
Another improvement is the more readable JSON formatting
* no longer has unnecessary blank lines after object open
* arrays of strings formatted as single line (inline) array

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
@@ -360,7 +360,7 @@ public @interface OpenApi {
     Pattern pattern() default Pattern.DEFAULT;
   }
 
-  @Target(ElementType.TYPE_USE)
+  @Target({ElementType.TYPE_USE, ElementType.METHOD, ElementType.PARAMETER})
   @Retention(RetentionPolicy.RUNTIME)
   @interface Description {
 
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
@@ -360,6 +360,13 @@ public @interface OpenApi {
     Pattern pattern() default Pattern.DEFAULT;
   }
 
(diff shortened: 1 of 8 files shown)


teleivo:
> ### Summary
> Allows to annotate endpoints with `@OpenApi.Description` to provide a description text
> 
> When annotating...
> 
> * method => endpoint description
> * parameter => request body and parameter description
> * return type => success response description
> * exception type => error response description
> 
> ### Improved JSON formatting
> Another improvement is the more readable JSON formatting
> 
> * no longer has unnecessary blank lines after object open
> * arrays of strings formatted as single line (inline) array

How does that compare/interact with what we could do so far i.e. describe in markdown?

jbee:
> How does that compare/interact with what we could do so far i.e. describe in markdown?

I should have explained :D 
So the code level has higher priority. I thought that this is better this way around because it is in the code and the expectation should be that what you see is what you get when it comes to code. If the code does not define a description the description from the markdown files is used. I think both have their pros and cons so having both as an option should help to manage the information easier. But lets discuss this further in a meeting.

teleivo (approved):
Nice 😄 

* I tried it on a parameter ✅
* I also tried it on a field in a param class which we mostly use in tracker like https://github.com/dhis2/dhis2-core/blob/18966c38c1975f567bc2136cc6d80ed1b5c9b812/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParams.java#L67. It does not work here which you also did not advertise 😋 but it would be great

jbee:
>  It does not work here which you also did not advertise 😋 but it would be great

I overlooked that as a source point. Will look into that :+1:

stian-sandvold (approved):

jbee:
I will include these changes in a new PR #17166

PR #17006, opened 2024-04-01
fix: Validate event parameters together in TE endpoint [DHIS2-16927]

enricocolasante:
`eventOccurredAfter`, `eventOccurredBefore`, and `eventStatus` are dependent parameters and they should either be all defined or none.

Docs are going to be updated accordingly.

This dependency between parameters should be avoided. This fix is needed because it should be fixed for new release but we should find a better way to deal with it.


muilpp (approved):

teleivo (approved):
Please also update the OpenAPI spec in https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md#parametertrackedentityrequestparamseventoccurredafter

PR #17010, opened 2024-04-02
refactor: Uniform generation of resource tables [DHIS2-16705]

larshelge:
Makes the generation of resource tables uniform by first creating a list of resource tables and invoking each in a loop.

The individual table methods in `ResoureTableService` were not needed, as calling code always generate all resource tables. The only exception is data approval resource tables, which has its own method.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTableService.java
```diff
@@ -33,50 +33,11 @@
  * @author Lars Helge Overland
  */
 public interface ResourceTableService {
-  /** Generates a resource table containing the hierarchy graph for each OrganisationUnit. */
-  void generateOrganisationUnitStructureTable();
+  /** Generates resource tables. */
+  void generateResourceTables();
```
david-mackessy, line 37: As someone with little analytics table knowledge it's not clear to me why `generateDataApprovalResourceTables` has its own method. Maybe a javadoc would make it clearer? (spread some knowledge for reasons why its separate and whether its's called at different stages of the generate flow)

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/resourcetable/ResourceTableService.java
larshelge: These methods are only called and are only necessary if the "hide unapproved data in analytics" system setting is enabled. They are called from the data approval analytics table manager.

larshelge: Good point on the Javadoc. I will update it in the next PR.

david-mackessy: lovely 👍

PR #17041, opened 2024-04-04
fix: Continuous execution of resource and analytics table jobs

larshelge:


david-mackessy (commented):
What are the implications of this? Does this setting only take affect if the job is already scheduled? If so, will it be running continuously even though it may have been setup to run every 1 hour? (will that cause a strain on CPU/DB)
I know it's a very small code change but a PR description covering this kind of information would really help someone who may have to look at this PR in a few months time.

david-mackessy (approved):

PR #17052, opened 2024-04-05
fix: Send webhook notification at stage completion [DHIS2-17168]

zubaira:
[DHIS2-17168](https://dhis2.atlassian.net/browse/DHIS2-17168)

[DHIS2-17168]: https://dhis2.atlassian.net/browse/DHIS2-17168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


muilpp (approved):

teleivo (commented):
How is this implemented in new tracker? Is the bug only affecting old tracker?

How can this bug/change be verified with a test?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/importer/shared/postprocess/ProgramNotificationPostProcessor.java
```diff
@@ -63,16 +63,14 @@ public void process(
             new ProgramStageCompletionNotificationEvent(this, programStageInstance.getId()));
         applicationEventPublisher.publishEvent(
             new StageCompletionEvaluationEvent(this, programStageInstance.getUid()));
+        applicationEventPublisher.publishEvent(
```
enricocolasante, line 66: The fix is not so easy. Just moving the notification here will treat all the webhook events as if they have the notification trigger to COMPLETE and that is not an assumption that can be made.

enricocolasante (changes requested):
A more complete solution is needed in order to solve this issue and not creating different ones

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/importer/shared/postprocess/ProgramNotificationPostProcessor.java
zubaira: these will only trigger when event has linked web hook notifications and event status is `COMPLETED`.
Secondly web hook notifications are only suppose to work at event/enrollment completion.

enricocolasante (approved):

PR #17056, opened 2024-04-05
fix: job progress without errors as immutable wither

jbee:
The operation end in a 500 as a `Map.of()` does not support the `clear()` operation. To avoid this the errors are now removed in an immutable fashion using `withoutErrors()` wither.


david-mackessy (changes requested):
Bugfix should be accompanied with a test to stop any potential regression in the future. If it can't be tested then please add a comment explaining why.

larshelge (approved):

jbee:
Will add tests

david-mackessy (approved):

PR #17057, opened 2024-04-06
fix: Type-safe authority checking in endpoints

david-mackessy:
# Background
We currently use `@PreAuthorize` in many endpoints to pre-authorise a user and make sure they have the required authority to be able to access that method.
It looks like: `@PreAuthorize("hasRole('ALL') or hasRole('F_DATAVALUE_ADD')")`

# Potential for improvement
As shown above, we currently pass authorities in string form when using this annotation (using SpEL)
Disadvantages with current approach:
- no type-safety
- room for typing errors
- compiler won't catch typos
- if an authority is removed, the pre-authorise could still be left with non-existing authority
- if there's no authority test in place we won't know there's an issue until reported
- not as easy as it could be to see where an authority is used

Advantages with new approach:
- bringing type-safety to authority checking in this manner
- compiler won't allow non-existing authorities
- easier to see where the authority is used throughout codebase
- slightly easier to read (less noise with the repeated `hasRole`)
e.g. `@HasAnyAuthority(authorities = {ALL, F_INDICATOR_MERGE})`

# Implementation
Using AOP (Spring + AspectJ) we can create our own Aspect, using a join point to apply advice wherever we use `@HasAnyAuthority` (could name be better?). `@HasAnyAuthority` takes in  any number of `Authorities` enums and the Aspect will check if the current User has any of the passed-in Authorities.

3 deps are required for this change.
We already have 2 of the required deps:
- `spring-aop` from `spring-context`
- `aspectjrt` from `spring-security-aspects`

Needed to add `aspectjweaver` as Spring uses some of the classes from this jar during runtime.

# Performance
Tested the response times of `@PreAuthorize` v `@HasAuthorities` on `/api/indicators/merge`
- calling the endpoint without any of the required authorities
- no difference in response times noticed
- called endpoint 10 times each
- average response was ~112 ms each

`@PreAuthorize` is also AOP so no surprise with matching times.

# Testing
Updated 2 e2e tests (1 valid, 1 invalid) which call the updated endpoint.

# Docs
More info about [Spring AOP here](https://docs.spring.io/spring-framework/docs/5.3.33/reference/html/core.html#aop).


stian-sandvold (approved):

jbee (commented):
Generally a very good idea, I just would like to do this without aspects. Maybe spring has a hook we could use to effectively do this?

david-mackessy:
> Generally a very good idea, I just would like to do this without aspects. Maybe spring has a hook we could use to effectively do this?

I can look into that 👍

david-mackessy:
> Generally a very good idea, I just would like to do this without aspects. Maybe spring has a hook we could use to effectively do this?

Spring-only alternative: #17072

netroms (approved):

david-mackessy:
closing, as Spring-only alternative preferred #17072

PR #17072, opened 2024-04-08
fix: Type-safe authority checking in endpoints

david-mackessy:
Spring-only alternative approach to #17057, using an Interceptor to check if the annotation is present.
Annotation looks slightly different `@RequiresAuthority(anyOf = {F_INDICATOR_MERGE})`
The `ALL` authority is automatically included in the check.

No difference in response times noticed with this new interceptor v master
- tested 30 web requests for each, on `/api/indicators/merge`
- response time ~112ms


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+
+    if (!handlerMethod.hasMethodAnnotation(HasAuthority.class)) {
+      return true;
+    }
+
+    HasAuthority hasAuthority = handlerMethod.getMethodAnnotation(HasAuthority.class);
+    if (hasAuthority != null) {
+      // include 'ALL' authority in required authorities
+      List<Authorities> requiredAuthorities = new ArrayList<>(List.of(Authorities.ALL));
+      requiredAuthorities.addAll(Arrays.stream(hasAuthority.anyOf()).toList());
```
jbee, line 78:
```suggestion
      requiredAuthorities.addAll(List.of(hasAuthority.anyOf()));
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+      Authentication authentication = securityContext.getAuthentication();
+      Collection<? extends GrantedAuthority> userAuthorities = authentication.getAuthorities();
+
+      // check if user has any of the required authorities passed in
+      if (requiredAuthorities.stream()
+          .noneMatch(
+              reqAuth ->
+                  userAuthorities.stream()
+                      .anyMatch(userAuth -> reqAuth.name().equals(userAuth.getAuthority())))) {
+        throw new AccessDeniedException("Access is denied");
```
jbee, line 91: Should we name the list of authorities in the message?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/HasAuthority.java
```diff
@@ -0,0 +1,49 @@
...
+ * org.hisp.dhis.user.User} has any of the {@link Authorities} passed in.
+ *
+ * <p>{@link Authorities#ALL} is automatically added to the check, as having this Authority allows
+ * access to all methods by default. No need to pass {@link Authorities#ALL} in the arguments. See
+ * {@link AuthorityInterceptor}.
+ */
+@Documented
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface HasAuthority {
```
jbee, line 47:
```suggestion
public @interface RequiresAuthority {
```
I think we should name this more declarative

jbee (commented):
As I said I like this better than a AOP based solution. 
A few small things to discuss. Maybe we just talk about this in tomorrows meeting.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
david-mackessy: i'll update this

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
david-mackessy: Could do, whatever we decide to go with 👍

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/HasAuthority.java
david-mackessy: i'll update this

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/IndicatorControllerTest.java
```diff
@@ -93,23 +93,4 @@ void testInvalidMerge() {
...
-            {
-                "sources": ["Uid00000010"],
-                "target": "Uid00000012",
-                "deleteSources": true
-            }""")
-            .content(HttpStatus.FORBIDDEN);
-    assertEquals("Forbidden", mergeResponse.getString("httpStatus").string());
-    assertEquals("ERROR", mergeResponse.getString("status").string());
-    assertEquals("Access is denied", mergeResponse.getString("message").string());
-  }
```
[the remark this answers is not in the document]
david-mackessy: Yep, I think that's because of our test setup.
It was moved to e2e test.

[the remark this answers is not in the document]
david-mackessy: actually, I tried it again after updating the test config. It does work now. I'll add it back in here.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+    RequiresAuthority requiresAuthority =
+        handlerMethod.getMethodAnnotation(RequiresAuthority.class);
+    if (requiresAuthority != null) {
+      // include 'ALL' authority in required authorities
+      List<Authorities> requiredAuthorities = new ArrayList<>(List.of(Authorities.ALL));
+      requiredAuthorities.addAll(List.of(requiresAuthority.anyOf()));
+
+      // get user authorities
+      final SecurityContext securityContext = SecurityContextHolder.getContext();
+      Authentication authentication = securityContext.getAuthentication();
```
vietnguyen, line 82: `securityContext.getAuthentication()` can return null. Would be great to have a NPE check here.

netroms (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
david-mackessy: cheers, i'll add a check

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
david-mackessy: after team meeting we've decided to list missing authorities in these checks. Will update in separate PR after speaking with FE.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/AuthorityInterceptor.java
```diff
@@ -0,0 +1,100 @@
...
+      if (authentication == null)
+        throw new SessionAuthenticationException("Error trying to get user authentication details");
```
david-mackessy, line 85: @netroms This kind of ties in to what we would like to do in future (instead of returning null for auth errors). Does this look ok to you? I wanted to call this out so you're aware of it, to make sure we're dealing with these errors in a consistent manner.
Returning null is not an option in this part of the code, it requires a boolean answer & true (has required auth) or false (does not have required auth, we don't know as it may be a system issue?) are both invalid responses in this context.

jbee (approved):

PR #17104, opened 2024-04-10
chore: Refactor pre auth checks

david-mackessy:
# Main aim of PR
1. Swap all uses of `@PreAuthorize` with `@RequiresAuthority`, after [previous PR](https://github.com/dhis2/dhis2-core/pull/17072) approved for use.
2. Include missing authorities in the web response. New response looks like:
```json
{
    "httpStatus": "Forbidden",
    "httpStatusCode": 403,
    "status": "ERROR",
    "message": "Access is denied, requires one Authority from [F_PERFORM_ANALYTICS_EXPLAIN]"
}
```
This new message format needs to be discussed with the Frontend team. Planned for April 16th. (we already return forbidden responses with different message values)
- discussed with FE team at Platform meeting on Apr 16, no issues.

# Issues encountered during refactor
## Potential bug
see [PR comment](https://github.com/dhis2/dhis2-core/pull/17104#discussion_r1560811961)
## Scope of pre auth checks
`@PreAuthorize` was mainly used at the method level but there were a few instances where it was used at the class level:
- `DataAnalysisController` class level only
- `OutlierDetectionController` class level only
- `AnalyticsOutlierDetectionController` class & method level (mix)

This required a change to `@RequiresAuthority` so it could be used at class level also -> `ElementType.Type` added to `@Target`.
`AuthorityInterceptor` also required an update to be able to handle scenarios where the annotation is used at class level, method level or both. The behaviour implemented mirrors how Spring works in this regard.

| Scope | Behaviour |
| -------- | ------- |
| Method  | Applies to annotated method, ignoring class level annotation |
| Class | Applies to all endpoint methods in the class, unless method level annotations exist |

## Authorities not defined in `Authorities`
10 authorities have been moved into `Authorities`. These were scattered around the codebase in schema descriptors, services, models etc. They are:
- `F_METADATA_MANAGE`
- `F_LOCALE_DELETE`
- `F_MINMAX_DATAELEMENT_ADD`
- `F_LEGEND_SET_PUBLIC_ADD`
- `F_LEGEND_SET_PRIVATE_ADD`
- `F_LEGEND_SET_DELETE`
- `F_MOBILE_SENDSMS`
- `F_JOB_LOG_READ`
- `F_MOBILE_SETTINGS`
- `M_DHIS_WEB_APP_MANAGEMENT("M_dhis-web-app-management")`

## Authority format
There are 2 different formats used for Authorities:
1. constant values: `THIS_IS_AN_AUTHORITY`
4. mix of underscore with hyphen and upper/lower: `THIS_IS-an-authority` (invalid Java constant)
  - these are authorities from web apps (I think)

`Authorities` has been updated to allow use of either
  - Just define any required web app authority like `WEB_APP_AUTHORITY("WEB_APP-authority")`
  - `AuthorityInterceptor` will always perform the check against the string value

# Testing
- tests added for class & method level auth checks
- test added for web app authority
- unit tests

# Other Info
## Convenience methods added to allow auth checks with `Authorities`:
They all just call the existing, underlying String methods which are stable and working (no new logic).
PRD
- `UserDetails#isAuthorized(@Nonnull Authorities auth)`
- `User#isAuthorized(@Nonnull Authorities auth)` (already exists)
- `User#hasAnyAuth(@Nonnull Collection<Authorities> auths)`

TEST
- `DhisConvenienceTest#createAndAddUserWithAuth(Set<OrganisationUnit> organisationUnits, Set<OrganisationUnit> dataViewOrganisationUnits, Authorities... auths)`
- `DhisConvenienceTest#createAndAddUserWithAuth(String userName, OrganisationUnit orgUnit, Authorities... auths)`
- `DhisConvenienceTest#createAndAddAdminUser(Authorities... authorities)`
- `DhisConvenienceTest#createUserWithAuthority(String username, Authorities... authorities)`
- `DhisControllerTestBase#switchToNewUserWithAuthorities(String username, Authorities... authorities)`

## Potential future work
User authorities are saved as `Set<String>`
- this is useful in testing, we can pass any string as an authority
- potentially could look to tighten that up in future, to `Set<Authorities>`


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataapproval/DataApproval.java
```diff
@@ -54,13 +59,13 @@
  * @author Jim Grace
  */
 public class DataApproval implements Serializable {
-  public static final String AUTH_APPROVE = "F_APPROVE_DATA";
+  public static final String AUTH_APPROVE = F_APPROVE_DATA.toString();
```
david-mackessy, line 62: these strings are only used in tests. I had to draw the line somewhere as this work could go on much longer :) 
The PRD code uses `Authorities`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/MaintenanceController.java
```diff
@@ -68,6 +70,7 @@
 @OpenApi.Tags("system")
 @Controller
 @RequestMapping(value = MaintenanceController.RESOURCE_PATH)
+@RequiresAuthority(anyOf = F_PERFORM_MAINTENANCE)
```
david-mackessy, line 73: Added check at class level as only 2 methods don't require this authority and they each have their own checks.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/validation/ValidationController.java
```diff
@@ -107,7 +108,7 @@ public ValidationSummary validate(
   @RequestMapping(
       value = "/sendNotifications",
       method = {RequestMethod.PUT, RequestMethod.POST})
-  @PreAuthorize("hasRole('ALL') or hasRole('M_dhis-web-app-management')")
+  @RequiresAuthority(anyOf = Authorities.M_DHIS_WEB_APP_MANAGEMENT)
```
david-mackessy, line 111: Is this a potential bug?
Should this be `F_RUN_VALIDATION`?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserControllerUtils.java
```diff
@@ -118,11 +118,15 @@ public ObjectNode getUserDataApprovalWorkflows(User user) {
...
+        authorities.contains(Authorities.ALL.toString())
+            || authorities.contains(F_APPROVE_DATA.toString());
     boolean canApproveLowerLevels =
-        authorities.contains(AUTHORITY_ALL) || authorities.contains(AUTH_APPROVE_LOWER_LEVELS);
+        authorities.contains(Authorities.ALL.toString())
+            || authorities.contains(F_APPROVE_DATA_LOWER_LEVELS.toString());
     boolean canAccept =
-        authorities.contains(AUTHORITY_ALL) || authorities.contains(AUTH_ACCEPT_LOWER_LEVELS);
+        authorities.contains(Authorities.ALL.toString())
+            || authorities.contains(F_ACCEPT_DATA_LOWER_LEVELS.toString());
```
jbee, line 129: I feel like we should not get all authorities and make contains checks but instead to individual "isAuthorized" checks on `UserDetails` for the user.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/validation/ValidationController.java
jbee: :thinking: good catch - it feels off - needs investigation to see if we can restore the knowledge of the intent here

jbee (commented):
Looks good. Great you are following up on your previous work.

I commented on lots of little things to consider now that we are at it. Nothing that you changed/introduced API wise as far as I can tell, more something we can straighten out in the same process.

Mostly I want to avoid having to do `toString()` (on `Authorities` instance) in too many places  if we can avoid it by adding overloaded helper methods that take care of it. I marked some places but if you see others that have potential and aren't overkill go for it.

Internally the authorities being a set of strings seems to have opened up for bugs where we wrongly check `contains` with an enum? That part might need backporting fix.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/MaintenanceController.java
david-mackessy: I know, I thought the same myself. Repeating something 21 times when 1 will do seems a bit unreasonable.
I can change it back, let me know.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -322,7 +322,7 @@ public boolean canIssueUserRole(UserRole group, boolean canGrantOwnUserRole) {
 
     final Set<String> authorities = getAllAuthorities();
 
-    if (authorities.contains(UserRole.AUTHORITY_ALL)) {
+    if (authorities.contains(Authorities.ALL.toString())) {
```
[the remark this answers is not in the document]
david-mackessy: This was one of the areas where there was a hardcoded string in `UserRole` which was used to check auth in some places. It's been removed.
`public static final String AUTHORITY_ALL = "ALL";

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -233,6 +234,8 @@ static UserDetails createUserDetails(
 
   boolean isAuthorized(String auth);
 
+  boolean isAuthorized(Authorities auth);
```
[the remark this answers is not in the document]
david-mackessy: i'll update

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetailsImpl.java
```diff
@@ -99,6 +100,15 @@ public boolean isAuthorized(String auth) {
...
+    return auths.contains(Authorities.ALL.toString()) || auths.contains(auth);
+  }
+
+  @Override
+  public boolean isAuthorized(Authorities auth) {
+    if (auth == null) {
+      return false;
+    }
+    final Set<String> auths = getAllAuthorities();
+    return auths.contains(Authorities.ALL.toString()) || auths.contains(auth.toString());
```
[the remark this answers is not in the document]
david-mackessy: yep, i'll update

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/TrackedEntityServiceTest.java
```diff
@@ -165,7 +165,7 @@ protected void setUpTest() throws Exception {
     fileResourceService.asyncSaveFileResource(fileResource, "fileResource".getBytes());
 
     userService = _userService;
-    user = createAndAddAdminUser(AUTHORITY_ALL);
+    user = createAndAddAdminUser(Authorities.ALL.toString());
```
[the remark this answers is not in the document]
david-mackessy: added

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserControllerUtils.java
david-mackessy: yep, that can be much simpler. I'll update.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/MaintenanceController.java
jbee: Lets discuss it in tomorrows meeting

david-mackessy: we agreed at the team meeting to allow class & method level

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/SystemUser.java
```diff
@@ -149,6 +149,11 @@ public boolean isAuthorized(String auth) {
     return false;
   }
 
+  @Override
+  public boolean isAuthorized(Authorities auth) {
+    return false;
```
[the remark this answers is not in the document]
david-mackessy: will update in separate PR so can backport

jbee (approved):

vietnguyen24 (approved):

vietnguyen (approved):

PR #17115, opened 2024-04-11
feat: support filestore MinIO DHIS2-17174 [2.41.0]

teleivo:
backport of https://github.com/dhis2/dhis2-core/pull/17014 and https://github.com/dhis2/dhis2-core/pull/17123

Approved by the RCB ✅


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jclouds/JCloudsStore.java
```diff
@@ -0,0 +1,221 @@
...
+  @PreDestroy
+  public void cleanUp() {
+    blobStoreContext.close();
+  }
+
+  private static Location createLocation(String provider, String location) {
+    if (location == null) {
+      // some BlobStores allow specifying a location, such as US-EAST, where containers will exist.
+      // null will choose a default location.
+      return null;
```
david-mackessy, line 166: This will satisfy the scenario outlined in the comment. I wonder could it cause issues in other scenarios? If null is not intended but passed in and a null location is retrieved somewhere else later in the code. Might be an edge case but just flagging it.

david-mackessy (approved):

jbee (approved):

PR #17119, opened 2024-04-11
fix: reject job config enable when no schedule is set [DHIS2-16408]

jbee:
### Summary
Ideally it is not possible to enable a job configuration that cannot run due to lack of schedule setup.
This is the state a job is in after no longer being part of a queue.

### Automatic Testing
Added a integration test scenario

### Manual Testing
* create 2 jobs
* make a queue with the 2 jobs
* delete the queue again
* verify the job(s) cannot be enabled 
* edit the job and set a schedule
* verify the job can be enabled again

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
@@ -380,6 +380,13 @@ public class JobConfiguration extends BaseIdentifiableObject implements Secondar
         && queueName == null;
   }
 
+  public boolean isReadyToRun() {
+    return schedulingType == SchedulingType.ONCE_ASAP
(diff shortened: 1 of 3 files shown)


david-mackessy (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfiguration.java
```diff
@@ -380,6 +380,13 @@ public boolean isRunOnce() {
         && queueName == null;
   }
 
+  public boolean isReadyToRun() {
+    return schedulingType == SchedulingType.ONCE_ASAP
+        || schedulingType == SchedulingType.CRON && cronExpression != null
```
larshelge, line 385: Just checking, do we want to use `StringUtils.notBlank` or some light validation of the cron expression to ensure it is valid?

PR #17153, opened 2024-04-16
fix: make oidcuser use parent isAuthorized method

netroms:
## Summary
Adds a missing implementation of the `isAuthorized` and `hasAnyAuthority` on the OidcUser.

### Automatic test
UserRoleTest#testOidcUserIsAuthorizedCheck


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/oidc/DhisOidcUser.java
```diff
@@ -174,7 +174,7 @@ public boolean hasAnyAuthority(Collection<String> auths) {
 
   @Override
   public boolean isAuthorized(String auth) {
```
david-mackessy, line 176: does `hasAnyAuthority` need to be updated also?

jbee: I think so

david-mackessy (approved):

PR #17164, opened 2024-04-17
fix: Hibernate nativeQuery clear all Hibernate caches[DHIS2-16934]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-16934

### Issue
- `HouseKeepingJob` use `org.hibernate.query.NativeQuery` to update/delete records in `jobconfiguration` table.
- That causes Hibernate to invalidate all caches in `SessionFactory`. Which in result trigger multiple `select` queries to database when lazy collection are loaded. This is the default Hibernate behavior, more details [here](https://vladmihalcea.com/how-does-hibernate-query-cache-work/rl).
- The job runs every 20s, so basically there are no cache entry available for whole application.

### Fix
- In order to avoid cache from being invalidated we must use method `NativeQuery.addSynchronizedEntityClass(class)`. Hibernate will then only invalidate cached entries of specified class.

### Test
- Added new integration test in `HibernateQueryCacheTest`

### Todo
- Look for other places which use `NativeQuery` and make sure only related caches are invalidated.


jbee (approved):

david-mackessy (approved):
nice find

PR #17169, opened 2024-04-17
fix: Just call sideeffect threads without sending to a queue [DHIS2-17134]

enricocolasante:
In tracker importer, the handling of side effects was sending notifications to a queue and when consuming a separate thread was created.
To send a message through the queue the bundle must be serializable but in this case that was not true because it depended on an external library (rule-engine) that is developed in Kotlin Multiplatform and it does not follow Java standards that allow serialization to work fine (no args constructor and getter and setters).

The solution with less impact is removing the queue and creating the thread directly.
A test has been created and it must be of type `IntegrationTestBase` as the assertion requires DB data to be committed and it needs to wait for the data to be committed using `Awaitility` library.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/job/BaseMessageManager.java
```diff
@@ -27,52 +27,19 @@
...
-    String jobId = CodeGenerator.generateUid();
-    sideEffectDataBundle.setJobId(jobId);
-
-    messageManager.sendQueue(getTopic(), sideEffectDataBundle);
-
-    return jobId;
-  }
-
   public void executeJob(Runnable runnable) {
     taskExecutor.executeTask(runnable);
```
teleivo, line 43: Remove/inline the BaseMessageManager as it just delegates to the AsyncTaskExecutor without adding any other logic.

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_metadata_with_program_rules.json
```diff
@@ -134,6 +135,9 @@
       "lastUpdatedBy": {
         "id": "M5zQapPyTZI"
       },
+      "createdBy": {
+        "id": "M5zQapPyTZI"
```
teleivo, line 139: AFAIK created/update users and dates will be ignored as its set by some base crud store. Adding these in fixtures is thus confusing 😬

teleivo (commented):
> To send a message through the queue the bundle must be serializable but in this case that was not true because it depended on an external library (rule-engine) that is developed in Kotlin Multiplatform and it does not follow Java standards that allow serialization to work fine (no args constructor and getter and setters).

AFAIK we put the bundle as JSON into the queue. Why does that not work with rule engine? 
You said

> In master this will be fixed properly, not using API objects from external libraries and using Jobs instead of simple threads

but then in these MessageManagers we have

```
    JobConfiguration jobConfiguration =
        new JobConfiguration("", JobType.TRACKER_IMPORT_NOTIFICATION_JOB, bundle.getAccessedBy());
```

and the confusingly named `executeJob` which runs the runnable. Are we using the jobConfig as a container for data we use to the send the message without actually creating a job via the scheduler?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/job/BaseMessageManager.java
enricocolasante: Fixed

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_metadata_with_program_rules.json
enricocolasante: Fixed

enricocolasante:
> AFAIK we put the bundle as JSON into the queue. Why does that not work with rule engine? 
The bundle is made of our model and it is serializable. Rule-engine result is coming from rule-engine library and the model there is not following the Pojo structure that is needed to correctly serialize/deserialize. Of course this is a bad practice, to use the model of an external API directly in our code, but the idea of this PR for the patch branch is to make it work with the less change possible.


> You said 
> > In master this will be fixed properly, not using API objects from external libraries and using Jobs instead of simple threads
> 
> but then in these MessageManagers we have
> 
> ```
>     JobConfiguration jobConfiguration =
>         new JobConfiguration("", JobType.TRACKER_IMPORT_NOTIFICATION_JOB, bundle.getAccessedBy());
> ```
> 
> and the confusingly named `executeJob` which runs the runnable. Are we using the jobConfig as a container for data we use to the send the message without actually creating a job via the scheduler?

Exactly, here the jobConfig is really used just to carry some data that will be notified at some point.
The idea for master would be to use the scheduler

teleivo (approved):

muilpp (approved):

PR #17183, opened 2024-04-18
fix: add back CurrentUserGroupInfo cache

vietnguyen:
### Issue
- Sharing ACL check requires database query which joins User and UserGroup table. 
- In 2.40 and previous versions, the result of that query was cached for performance purpose.
- However the cache has been removed in 2.41 as a part for the CurrentUserService refactoring.

### Fix
- Add back the cache, make sure to invalidate it when userGroup or User is updated.

### Test
- We already have many sharing ACL test in `SharingControllerTest` and `AbstractCrudControllerTest`.

### DIFF
--- a/dhis-2/dhis-web-embedded-jetty/pom.xml
+++ b/dhis-2/dhis-web-embedded-jetty/pom.xml
@@ -72,10 +72,6 @@
       <artifactId>spring-web</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.springframework.security</groupId>
-      <artifactId>spring-security-core</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.springframework</groupId>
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-web-embedded-jetty/pom.xml
```diff
@@ -71,10 +71,6 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-web</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.springframework.security</groupId>
-      <artifactId>spring-security-core</artifactId>
-    </dependency>
```
vietnguyen, line 77: maven throws unused declare dependency for this.

netroms (approved):

david-mackessy (approved):

PR #17189, opened 2024-04-18
fix: tracker/relationships emits inaccessible program attributes  [ Dhis2-17090 ]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-17090

1) Apply the configurable relationship constraints for a relationship type. As discussed with @JoakimSM, the UI only shows program attributes as a constraint.
In the code, this is done in the mapper as an @AfterMapping operation. This way, it is more maintainable and follows the model logic directly in the mapper. The alternative to this is to seek every place where we fetch relationships (directly or indirectly) and apply the filter. 
2) Apply the logic done by @muilpp for the `canRead` method of tracked entity in the `DefaultTrackerAccessManager`. This is now valid for every check in place and consistent across the tracker ( therefore, also for relationships ).

@muilpp `canRead` for enrollment and event still validates the tracked entity. In this case, we validate against a specific program, so should it be okay, or do we need to apply the new logic here?

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.trackedentity;
 
 import java.util.List;
+import java.util.Set;
 import org.hisp.dhis.category.CategoryOptionCombo;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
@@ -82,7 +83,7 @@ public interface TrackerAccessManager {
 
   List<String> canDelete(UserDetails user, Event event, boolean skipOwnershipCheck);
 
-  List<String> canRead(UserDetails user, Relationship relationship);
+  Set<String> canRead(UserDetails user, Relationship relationship);
 
   List<String> canWrite(UserDetails user, Relationship relationship);
 
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
@@ -28,7 +28,6 @@
 package org.hisp.dhis.trackedentity;
 
 import java.util.List;
-import java.util.Set;
 import org.hisp.dhis.category.CategoryOptionCombo;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
@@ -83,7 +82,7 @@ public interface TrackerAccessManager {
 
   List<String> canDelete(UserDetails user, Event event, boolean skipOwnershipCheck);
 
-  Set<String> canRead(UserDetails user, Relationship relationship);
+  List<String> canRead(UserDetails user, Relationship relationship);
 
   List<String> canWrite(UserDetails user, Relationship relationship);
 
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
@@ -32,7 +32,10 @@ import static org.hisp.dhis.trackedentity.TrackerOwnershipManager.OWNERSHIP_ACCE
 import static org.hisp.dhis.trackedentity.TrackerOwnershipManager.PROGRAM_ACCESS_CLOSED;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.category.CategoryOption;
 import org.hisp.dhis.category.CategoryOptionCombo;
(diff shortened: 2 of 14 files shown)


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/deprecated/tracker/relationship/AbstractRelationshipService.java
```diff
@@ -472,7 +477,8 @@
   }
 
   private org.hisp.dhis.dxf2.deprecated.tracker.trackedentity.RelationshipItem
-      includeRelationshipItem(RelationshipItem dao, boolean uidOnly) {
+      includeRelationshipItem(
+          RelationshipItem dao, boolean uidOnly, RelationshipConstraint constraint) {
```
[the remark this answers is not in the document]
lucaCambi77: missing implementation in the old tracker, will fix soon

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -59,31 +63,83 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
...
+    Set<String> errors = new HashSet<>();
+
+    List<Program> tetPrograms =
+        programs.stream()
+            .filter(
+                p -> Objects.equals(p.getTrackedEntityType(), trackedEntity.getTrackedEntityType()))
+            .toList();
+
+    if (tetPrograms.isEmpty()) {
+      return Set.of("User has no access to any program");
```
muilpp, line 109: @lucaCambi77 I changed this to the error message above, I think it makes it clearer for the user to understand what's going on. Anyway, we are not using this error messages anywhere, we just check whether this method returns values in the collection or not, and then we throw a different error message.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -59,31 +63,83 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
...
 
     TrackedEntityType trackedEntityType = trackedEntity.getTrackedEntityType();
 
     if (!aclService.canDataRead(user, trackedEntityType)) {
-      errors.add(
+      return Set.of(
           "User has no data read access to tracked entity type: " + trackedEntityType.getUid());
     }
 
+    initializeTrackedEntityOrgUnitParents(trackedEntity);
```
muilpp, line 98: should we leave the TODO comment on top of this so we don't forget to find a permanent solution?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -59,31 +63,93 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
...
+    if (tetPrograms.isEmpty()) {
+      return Set.of("User has no access to any program");
+    }
+
+    for (Program program : tetPrograms) {
+      List<String> e = canRead(user, trackedEntity, program);
+      if (e.isEmpty()) {
+        return Set.of();
+      } else {
+        errors.addAll(e);
```
enricocolasante, line 117: Shouldn't we just say that user cannot access TE?
I can see that here there could be a lot of info that are not interesting for the user, potentially all the programs for the same TET that the user has no access.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
```diff
@@ -82,7 +83,7 @@ List<String> canWrite(
 
   List<String> canDelete(UserDetails user, Event event, boolean skipOwnershipCheck);
 
-  List<String> canRead(UserDetails user, Relationship relationship);
+  Set<String> canRead(UserDetails user, Relationship relationship);
```
enricocolasante, line 86: Why do we need to change this to a `Set`?
This whole interface is working with `List`, I would maintain consistency unless there is a pretty important reason

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
lucaCambi77: These errors are not shown to the user, now they are converted to a forbidden exception or will filter out relationships like in this case. I understand @muilpp wants to move to use error codes which will then help to move to a not found / forbidden logic as a response so I'll leave this job for later

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
lucaCambi77: When validating relationships, you can have the same error repeating, so i thought it made sense to have a set

muilpp:
> https://dhis2.atlassian.net/browse/DHIS2-17090
> 
>     1. Apply the configurable relationship constraints for a relationship type. As discussed with @JoakimSM, the UI only shows program attributes as a constraint.
>        In the code, this is done in the mapper as an @AfterMapping operation. This way, it is more maintainable and follows the model logic directly in the mapper. The alternative to this is to seek every place where we fetch relationships (directly or indirectly) and apply the filter.
> 
>     2. Apply the logic done by @muilpp for the `canRead` method of tracked entity in the `DefaultTrackerAccessManager`. This is now valid for every check in place and consistent across the tracker ( therefore, also for relationships ).
> 
> 
> @muilpp `canRead` for enrollment and event still validates the tracked entity. In this case, we validate against a specific program, so should it be okay, or do we need to apply the new logic here?

@lucaCambi77 The current validations in both enrollments and events should be enough as they are. As you mentioned, they rely on their associated programs to identify the owner. This design ensures we should never have the case where there's no program in either of them.

muilpp (approved):

enricocolasante (approved):

PR #17200, opened 2024-04-22
fix: Update validation when no program specified in TE endpoint [DHIS2-17236]

muilpp:
When requesting multiple tracked entities without a program in the context, the ownership validation is done differently than in https://github.com/dhis2/dhis2-core/pull/17189.
In this case the validation is done in the sql query, so this is a refactor of the org unit join where we validate ownership.

- When no program provided, we use the join to trackedentityprogramowner to check if the user has access to that particular TE/program pair, by first matching by the owner, or the registering org unit if no owner is found.
- Remove the org unit id and name from the select clause because we are not using it and they are not sortable either.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-17236

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
@@ -56,7 +56,6 @@ import org.hisp.dhis.common.IllegalQueryException;
 import org.hisp.dhis.common.OrganisationUnitSelectionMode;
 import org.hisp.dhis.common.QueryFilter;
 import org.hisp.dhis.common.hibernate.SoftDeleteHibernateObjectStore;
-import org.hisp.dhis.commons.collection.CollectionUtils;
 import org.hisp.dhis.commons.util.SqlHelper;
 import org.hisp.dhis.event.EventStatus;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
@@ -334,8 +333,6 @@ class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<Tracked
                 "TE.inactive",
                 "TE.potentialduplicate",
                 "TE.deleted",
-                "TE.orgunit_uid",
-                "TE.orgunit_name",
                 "TE.trackedentitytypeid"));
 
     // all orderable fields are already in the select. Only when ordering by enrollment date do we
@@ -409,9 +406,7 @@ class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<Tracked
                 "TE.lastupdatedatclient as lastupdatedatclient",
                 "TE.inactive as inactive",
                 "TE.potentialduplicate as potentialduplicate",
-                "TE.deleted as deleted",
-                "OU.uid as orgunit_uid",
-                "OU.name as orgunit_name"));
+                "TE.deleted as deleted"));
 
     for (Order order : params.getOrder()) {
       if (order.getField() instanceof String field) {
@@ -449,20 +444,6 @@ class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<Tracked
   private String getFromSubQueryTrackedEntityConditions(
       SqlHelper whereAnd, TrackedEntityQueryParams params) {
     StringBuilder trackedEntity = new StringBuilder();
-
-    if (params.hasTrackedEntityType()) {
-      trackedEntity
-          .append(whereAnd.whereAnd())
-          .append("TE.trackedentitytypeid = ")
-          .append(params.getTrackedEntityType().getId())
-          .append(SPACE);
-    } else if (!CollectionUtils.isEmpty(params.getTrackedEntityTypes())) {
-      trackedEntity
-          .append(whereAnd.whereAnd())
-          .append("TE.trackedentitytypeid IN (")
-          .append(getCommaDelimitedString(getIdentifiers(params.getTrackedEntityTypes())))
-          .append(") ");
-    }
 
     if (params.hasTrackedEntities()) {
       trackedEntity
@@ -611,11 +592,12 @@ class HibernateTrackedEntityStore extends SoftDeleteHibernateObjectStore<Tracked
 
     orgUnits
         .append(" INNER JOIN organisationunit OU ")
-        .append("ON OU.organisationunitid = ")
+        .append("ON OU.organisationunitid in ")
         .append(
+            // TODO What about event programs, do they have an owner or do I have to check all TE?
             params.hasProgram() && !skipOwnershipCheck(params)
-                ? "PO.organisationunitid "
-                : "TE.organisationunitid ");
+                ? "(PO.organisationunitid)"
+                : getOwnerOrgUnitsForAccessibleTrackedEntityPrograms(params));
 
     if (!params.hasOrganisationUnits()) {
       return orgUnits.toString();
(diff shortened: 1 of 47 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -166,6 +169,16 @@ private List<TrackedEntityType> filterAndValidateTrackedEntityTypes(User user)
     return trackedEntityTypes;
   }
 
+  private List<Program> getPrograms(Program program, User user) {
+    if (program == null) {
+      return programService.getAllPrograms().stream()
```
enricocolasante, line 174: Here we could already filter out event programs or even better we could create a method in `ProgramService` to get all tracker programs.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -140,16 +145,14 @@ public TrackedEntityQueryParams map(TrackedEntityOperationParams operationParams
     return params;
   }
 
-  private List<TrackedEntityType> getTrackedEntityTypes(
-      TrackedEntityType trackedEntityType, Program program, User user) throws BadRequestException {
+  private List<TrackedEntityType> getTrackedEntityTypes(Program program, User user)
```
enricocolasante, line 148: Based on the validation that we do to the parameters in the request, we should have 3 cases here:
- TET from the params
- TET from the Program (it should be validated already to be a tracker program, hence the TET cannot be null)
- TETs from the list of TrackedEntitites requested

We shouldn't have a case where we look at all the TET that the user can access

enricocolasante (approved):
There are just a couple of comments that are more for further improvement than to be fixed right now.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -632,6 +614,40 @@ private String getFromSubQueryJoinOrgUnitConditions(TrackedEntityQueryParams par
...
+            + "from trackedentitytype tet "
+            + "left join trackedentityprogramowner tpo on tpo.trackedentityid = te.trackedentityid "
+            + "left join program p on p.programid = tpo.programid "
+            + "and tet.trackedentitytypeid = te.trackedentitytypeid "
+            + "and p.trackedentitytypeid = te.trackedentitytypeid "
+            + "and p.programid in ("
+            + getCommaDelimitedString(getIdentifiers(params.getPrograms()))
+            + ")";
+
+    if (params.hasTrackedEntityType()) {
```
lucaCambi77, line 639: this query doesn't look good to me. the access plan is not promising see https://explain.dalibo.com/plan/2aac13cc0c98a97b
Also, i don't see why you need to start from the `trackedentytype` table if you already know it from the program and the te. I'll see how to improve the query or you can check it out as well

lucaCambi77 (changes requested):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
muilpp: I agree it would make more sense to have a method in `ProgramService` that provides us with tracker programs directly.
Thing is, as you know, that service has some methods that validate metadata access and some that don't.
As we discussed with Viet, it might make more sense to create our own tracker program service and make sure we just add methods in there that are metadata validated. I'd rather do that than continue adding methods in the current program service. That could be a fairly big refactor and it might need a team discussion first , so I'll leave it for another ticket.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
muilpp: Agreed with this one too, the problem here is that getting the list of accessible TET is the only way we have to validate the access to a tracked entity when no program nor TET are present in the request.
Refactoring this now could lead to changes in multiple places and I don't feel comfortable doing it so close to the release.
I'd rather do it in another ticket that just focuses on this part, and preferably, after the release.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
muilpp: After the refactor, the query plan looks better: https://explain.dalibo.com/plan/5g16549b8b050gbb

lucaCambi77: @muilpp ask @radnov to set up an instance with a performance database, use the https://github.com/dhis2/performance-tests-gatling to see performances. You can compare the result with the https://test.performance.dhis2.org/dev instance I've used recently for https://github.com/dhis2/dhis2-core/pull/16965

muilpp:
here are the results of the performance test run with Gatling.
comparing them with [Luca's performance test](https://github.com/dhis2/dhis2-core/pull/16965), there's a 50ms decrease in a call to `/tracker/trackedEntities` requesting a specific `trackedEntityType`, while requesting a program or multiple tracked entity UIDs, seem to be 20/30ms faster.

Screenshot from 2024-04-26 13-33-21
Screenshot from 2024-04-26 13-33-16
Screenshot from 2024-04-26 13-33-03

lucaCambi77:
> here are the results of the performance test run with Gatling. comparing them to [Luca's performance test](https://github.com/dhis2/dhis2-core/pull/16965), there's a 50ms decrease in a call to `/tracker/trackedEntities` requesting a specific `trackedEntityType`, while requesting a program or multiple tracked entity UIDs, seem to be 20/30ms faster.
> 
@muilpp I'm not sure why you are referring to my PR, as it is about events. I see that you have deployed this branch and compared it with a dev instance (both using a performance test database). So, those are the performance results of this branch? Or the ones in the dev instance? Can you add both? Otherwise, it is not possible to evaluate a difference.
In any case, comparing how fast a specific request is irrelevant because it can not be deterministic. Rather, we need to look at the overall request/response throughput and check the difference between the distribution of those performances.

teleivo (approved):

muilpp:
> > here are the results of the performance test run with Gatling. comparing them to [Luca's performance test](https://github.com/dhis2/dhis2-core/pull/16965), there's a 50ms decrease in a call to `/tracker/trackedEntities` requesting a specific `trackedEntityType`, while requesting a program or multiple tracked entity UIDs, seem to be 20/30ms faster.
> 
> @muilpp I'm not sure why you are referring to my PR, as it is about events. I see that you have deployed this branch and compared it with a dev instance (both using a performance test database). So, those are the performance results of this branch? Or the ones in the dev instance? Can you add both? Otherwise, it is not possible to evaluate a difference. In any case, comparing how fast a specific request is irrelevant because it can not be deterministic. Rather, we need to look at the overall request/response throughput and check the difference between the distribution of those performances.

updated the comment with the results from master too

PR #17205, opened 2024-04-22
fix: Fix TE ownership validation in Tracker importer [DHIS2-17251]

enricocolasante:
In the importer we need to validate access to Tracked Entity the same way we are now doing in the exporter.
We have 3 different cases in the importer:

- CREATE: Check capture scope of the user using registration unit
- UPDATE or DELETE: Check if the user has access to any TE/program pair, by first matching by the owner, or the registering org unit if no owner is found

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
@@ -159,7 +159,7 @@ class SecurityOwnershipValidator
         programService.getAllPrograms().stream()
             .filter(
                 p ->
-                    p.getTrackedEntityType() != null
+                    p.isRegistration()
                         && p.getTrackedEntityType()
                             .getUid()
                             .equals(te.getTrackedEntityType().getUid()))
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
@@ -104,7 +104,8 @@ class SecurityOwnershipValidator
     else {
       TrackedEntity te = bundle.getPreheat().getTrackedEntity(trackedEntity.getTrackedEntity());
       if (!trackerAccessManager.canWrite(UserDetails.fromUser(bundle.getUser()), te).isEmpty()) {
-        reporter.addError(trackedEntity, ValidationCode.E1003, te.getUid());
+        reporter.addError(
+            trackedEntity, ValidationCode.E1003, bundle.getUser().getUid(), te.getUid());
       }
     }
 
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
```diff
@@ -91,14 +95,18 @@ public void validate(
                 .getOrganisationUnit()
             : bundle.getPreheat().getOrganisationUnit(trackedEntity.getOrgUnit());
 
-    // If trackedEntity is newly created, or going to be deleted, capture
-    // scope has to be checked
-    if (strategy.isCreate() || strategy.isDelete()) {
+    // If trackedEntity is newly created, capture scope has to be checked
```
muilpp, line 98: can we either remove the comment or add that we also validate write access to the tracked entity type?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
```diff
@@ -91,14 +95,18 @@ public void validate(
...
+    if (strategy.isCreate()) {
+      checkTeiTypeWriteAccess(reporter, bundle, trackedEntity, trackedEntityType);
```
muilpp, line 100: rename the method to check**TeT**ypeWriteAccess

muilpp (commented):
lgtm, just a couple of details I've seen.

I might add a test or two where a TE has been transferred, to validate a user with access to the transferred org unit can delete that TE, and a user with access to the registration org unit only, can't.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
```diff
@@ -91,14 +95,18 @@ public void validate(
...
       checkOrgUnitInCaptureScope(reporter, bundle, trackedEntity, organisationUnit);
     }
-    // if its to update trackedEntity, search scope has to be checked
+    // if its to update or delete trackedEntity, write access has to be checked
```
muilpp, line 103: similar to the one above, I think we either expand on why write access has to be checked, or we remove the comment altogether.

enricocolasante:
> lgtm, just a couple of details I've seen.
> 
> I might add a test or two where a TE has been transferred, to validate a user with access to the transferred org unit can delete that TE, and a user with access to the registration org unit only, can't.

Added integration tests to check that update and delete for transferred TE use ownership and not the registration org unit.
Create was already covered by existing tests

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
```diff
@@ -145,15 +153,25 @@ private void checkOrgUnitInCaptureScope(
...
+            .toList();
+    if (!programsToCheck.isEmpty()
+        && programsToCheck.stream()
+            .noneMatch(
+                p ->
+                    trackerAccessManager
+                        .canWrite(UserDetails.fromUser(bundle.getUser()), te, p, false)
+                        .isEmpty())) {
+      reporter.addError(dto, ValidationCode.E1003, te.getUid());
     }
```
lucaCambi77: a question about this. Should this be a general rule for checking access to write a te? Could this be moved to the ACL directly as we did for `canRead` of a te?

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
enricocolasante: This is an outdated version of the code.
I already moved it to the ACL

lucaCambi77 (approved):
you moved it in another PR? not sure what you mean but i don't want to block

PR #17206, opened 2024-04-23
fix: Bad SQL syntax exception in /analytics/Event/aggregate[2.42-DHIS2-16235]

d-bernat:
The issue was pinpointed through query parameter validation, specifically within the period dimension. When the parameters are found to be invalid, the error message sent to the client should appear as follows:
 {
    "httpStatus": "Conflict",
    "httpStatusCode": 409,
    "status": "ERROR",
    "message": "Start and end dates or at least one period must be specified",
    "errorCode": "E7205"
}
The instantiation of the error message was incorrectly implemented, resulting in certain special cases overwriting existing instances of the error message. This issue has now been rectified.


gnespolino (approved):
good one

maikelarabori (approved):

PR #17223, opened 2024-04-24
fix: add back CurrentUserGroupInfo cache (41.0)

vietnguyen:
backport https://github.com/dhis2/dhis2-core/pull/17183

https://dhis2.atlassian.net/browse/DHIS2-16934

### Issue
- Sharing ACL check requires database query which joins User and UserGroup table.
- In 2.40 and previous versions, the result of that query was cached for performance purpose.
- However the cache has been removed in 2.41 as a part for the CurrentUserService refactoring.

### Fix
- Add back the cache, make sure to invalidate it when userGroup or User is updated.
### Test
- We already have many sharing ACL test in SharingControllerTest and AbstractCrudControllerTest.


stian-sandvold (approved):
Approved by RCB as it's part of a related performance fix (DHIS2-16934)

PR #17224, opened 2024-04-24
fix: only one concurrent user allowed

netroms:
## Summary
Fixes bug, only allowing one concurrent user session.
Caused by new login controller not setting the max session per user from the dhis.conf

### Automatic tests
N/A now, requires an E2E test.

### Jira
[DHIS2-17225](https://github.com/dhis2/dhis2-core/pull/17225)

[DHIS2-17225]: https://dhis2.atlassian.net/browse/DHIS2-17225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vietnguyen (approved):

david-mackessy (changes requested):
We try to add tests for all bug fixes. In the rare cases where it is not possible, the PR should include a comment advising why no tests have been included.

jbee (approved):

david-mackessy (approved):

PR #17235, opened 2024-04-25
fix: Event Sync job throws NPE [DHIS2-17233]

david-mackessy:
# Issue
`EventSynchronization` job fails with `NullPointerException` due to null `User` in `JdbcEventStore#isUserSearchScopeNotSet`
`EventSynchronization` can be run on its own or as part of the `MetadataSyncJob` (2 separate flows needed to be tested)

# Cause
Pre `v41`, job scheduling worked differently. Jobs were run with a `null` superuser. Depending how the job flow works, it could hit this `null` superuser.

[This PR](https://github.com/dhis2/dhis2-core/pull/15041) introduced this current bug on `v39` in `September 2023`. The introduction of `User` params in `JdbcEventStore` methods meant that an `EventSynchronization` job would now run into this `null` superuser issue.

# Fix
As part of the `EventSynchronization`, the `User` setup for system synchronisation will now be injected to perform the synchronisation.
- this is only done if there is no current user in the session
- at this part of the job flow, we already know that the server is setup correctly for synchronisation, so valid credentials are present (this part of the flow is not hit if synchronisation is not set up correctly)
- the `User` needs to be initialized and unproxied in order to avoid `LazyInitializationException`

A new Hibernate session is also required to be bound to the job thread.

These types of fixes have been required in previous, similar bugs, and is a known issue with regards to old job scheduling (not present in `v41`).

# Testing
## Automated
Test added to first, replicate the `NullPointerException` in the `EventSynchronization` flow, and secondly, now confirms its absense.

## Manual
Tested locally & with Instance Manager. Also created a [test guide](https://github.com/dhis2/wow-backend/blob/master/guides/testing/metadata_sync_testing.md) for easy testing of Metadata synchronization job.

# Other Info
## Bugs
Two other bugs were found during this investigation
1. Always throwing a `MetadataSyncServiceException` in `EventSynchronization#synchronizePage` which has been fixed, see [this comment](https://github.com/dhis2/dhis2-core/pull/17235#discussion_r1582786091)
2. The `lastExecutedStatus` of a job config shows as `FAILED` even when it was successfull. Bug created [here](https://dhis2.atlassian.net/browse/DHIS2-17292).

## QA Regression pack
I've reached out to Lina and asked to possibly include a test for synchronization from now on, using the guide created above

## Backend awareness
I'll bring up this issue as an example, to be aware when making changes in any code that gets executed in a job (pre `v41`), especially when a `User` is involved. Also, if any code used in a synchronization flow is changed, the manual test should be run to confirm no adverse effects.


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/sync/EventSynchronization.java
```diff
@@ -187,6 +239,7 @@ protected void synchronizePage(int page, EventSynchronisationContext context) {
           events.getEvents().stream().map(Event::getEvent).collect(Collectors.toList());
       log.info("The lastSynchronized flag of these Events will be updated: " + eventsUIDs);
       eventService.updateEventsSyncTimestamp(eventsUIDs, context.getStartTime());
+      return;
```
david-mackessy, line 242: this fixes a separate bug introduced in [this PR](https://github.com/dhis2/dhis2-core/pull/10402/files#diff-d4fc2728be8a0844b9c927cd5a58e1af6db9ac0bcae6edc35e7711d83956017a), whereby a `MetadataSyncServiceException` would always be thrown in `EventSynchronization#synchronizePage`.
A `return` is needed to avoid always throwing the exception.

david-mackessy:
Hi @enricocolasante , I've included the Tracker team in this PR review to make sure that the proposed changes look ok from Tracker point of view. Just in case I'm not aware of any unknown impact this change may have.
The important part from the Tracker point of view, is that the `User` set up to perform synchronizations will be injected to execute the `EventSynchronization`.  This will resolve the NPE seen in `JdbcEventStore#isUserSearchScopeNotSet`

enricocolasante (approved):

netroms (approved):

PR #17248, opened 2024-04-26
WIP: [DHIS2-17282]

larshelge:


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/setting/AnalyticsTableSettings.java
```diff
@@ -125,4 +130,50 @@ Database getAndValidateDatabase(String value) {
...
+  }
+
+  /**
+   * Converts the boolean enabled flag to a {@link Skip} value.
+   *
+   * @param enabled the boolean enabled flag.
+   * @return a {@link Skip} value.
+   */
+  Skip toSkip(boolean enabled) {
+    return enabled ? Skip.INCLUDE : Skip.SKIP;
```
maikelarabori, line 177: I haven't realized this Enum.
It sounds strange to include something part of a skip, ie: Skip.INCLUDE. In the same way, it looks weird to skip something twice, ie: Skip.SKIP.

Something like Skip.TRUE and Skip.FALSE could sound better...? Or Skip.YES, Skip.NO. These options are closer to the boolean flag `enabled` param on this method (which can be true or false).

Just wondering...

d-bernat (approved):
Hey @Lars, here is my test outcome, 17282, full export:
SL DB
1. idxs on: 01:32, 5695 indexes on analytics* tables
2. idxs off: 01:36, 5123 indexes on analytics* tables

HMIS TEST DB 
1. idxs on: 07:04, 15835 indexes on analytics* tables
2. idxs off: 04:19, 13189 indexes on analytics* tables

maikelarabori:
> Hey @Lars, here is my test outcome, 17282, full export: SL DB
> 
>     1. idxs on: 01:32, 5695 indexes on analytics* tables
> 
>     2. idxs off: 01:36, 5123 indexes on analytics* tables
> 
> 
> HMIS TEST DB
> 
>     1. idxs on: 07:04, 15835 indexes on analytics* tables
> 
>     2. idxs off: 04:19, 13189 indexes on analytics* tables

Nice feedback @d-bernat 👍

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/setting/AnalyticsTableSettings.java
larshelge: Yes we can change to Skip.TRUE and Skip.FALSE in a separate PR. The goal is just to avoid boolean parameters in general, because it is hard to read, and error prone, as if you have many boolean parameters, you can inadvertently replace them and get different behavior.

larshelge:
Okay, thanks for the feedback @d-bernat . The output from SL demo seems to be off, I need to have another look then.

PR #17263, opened 2024-04-29
test: Add test for SqlHelper

larshelge:
Refactors the `SqlHelper` class. Adds a unit test.


netroms (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/util/SqlHelper.java
```diff
@@ -47,8 +47,16 @@ public class SqlHelper {
 
   private boolean andInvoked = false;
 
-  public SqlHelper() {}
+  /** Constructor. */
+  public SqlHelper() {
+    this.includeSpaces = false;
```
david-mackessy, line 52: is this just to be explicit, setting it to false? It's already false by default.

david-mackessy (approved):

PR #17274, opened 2024-04-30
fix: Fix TE ACL to check capture scope on register OU [DHIS2-17251]

enricocolasante:
In the importer, we need to validate access to the Tracked Entity in the same way we are now doing in the exporter.
We have 3 different cases in the importer:

- CREATE: Check the capture scope of the user using the registration unit and validate Tracked Entity Type write access
- UPDATE: Check if the user has access to any TE/program pair, by first matching by the owner, or the registering org unit if no owner is found
- DELETE: Check the capture scope of the user using the registration unit and check if the user has access to any TE/program pair, by first matching by the owner, or the registering org unit if no owner is found

For CREATE and DELETE, we need to check that the registration org unit is in the capture scope.
For UPDATE, we are not changing the current behavior but we will need to discuss if it is required to check if the org unit is being updated and if so what to do.

Then, we are not checking the TE/Program Pair for CREATE as we know there isn't any at this moment but we need to check it for UPDATE and DELETE.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java
@@ -95,10 +95,15 @@ class SecurityOwnershipValidator
                 .getOrganisationUnit()
             : bundle.getPreheat().getOrganisationUnit(trackedEntity.getOrgUnit());
 
-    if (strategy.isCreate()) {
+    if (strategy.isCreate()){
       checkTeTypeWriteAccess(reporter, bundle, trackedEntity, trackedEntityType);
+    }
+
+    if (strategy.isCreate() || strategy.isDelete()) {
       checkOrgUnitInCaptureScope(reporter, bundle, trackedEntity, organisationUnit);
-    } else {
+    }
+
+    if(!strategy.isCreate()){
       TrackedEntity te = bundle.getPreheat().getTrackedEntity(trackedEntity.getTrackedEntity());
       if (!trackerAccessManager.canWrite(UserDetails.fromUser(bundle.getUser()), te).isEmpty()) {
         reporter.addError(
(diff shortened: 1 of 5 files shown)


lucaCambi77 (approved):

muilpp (approved):

muilpp:
if we don't have it yet, would it make sense to have a test to validate if fails if the TE was transferred, the user has capture scope to the registering org unit but has no access to the one it was transferred to?

enricocolasante:
> if we don't have it yet, would it make sense to have a test to validate if fails if the TE was transferred, the user has capture scope to the registering org unit but has no access to the one it was transferred to?

`shouldFailToDeleteWhenTEWasTransferredAndUserHasAccessToTransferredOrgUnitAndTEOUIsNotInCaptureScope` should be it.
The only strategy that can be tested is DELETE, as we do not check capture scope for UPDATE and CREATE imply that there are no ownerships, so we couldn't have transfer before.

Does it make sense?

muilpp:
> > if we don't have it yet, would it make sense to have a test to validate if fails if the TE was transferred, the user has capture scope to the registering org unit but has no access to the one it was transferred to?
> 
> `shouldFailToDeleteWhenTEWasTransferredAndUserHasAccessToTransferredOrgUnitAndTEOUIsNotInCaptureScope` should be it. The only strategy that can be tested is DELETE, as we do not check capture scope for UPDATE and CREATE imply that there are no ownerships, so we couldn't have transfer before.
> 
> Does it make sense?

I meant it the other way around, where the user is trying to delete a TE that has the registering org unit in his capture scope, but has no access to the org unit the TE was transferred to.

enricocolasante:
> > > if we don't have it yet, would it make sense to have a test to validate if fails if the TE was transferred, the user has capture scope to the registering org unit but has no access to the one it was transferred to?
> > 
> > 
> > `shouldFailToDeleteWhenTEWasTransferredAndUserHasAccessToTransferredOrgUnitAndTEOUIsNotInCaptureScope` should be it. The only strategy that can be tested is DELETE, as we do not check capture scope for UPDATE and CREATE imply that there are no ownerships, so we couldn't have transfer before.
> > Does it make sense?
> 
> I meant it the other way around, where the user is trying to delete a TE that has the registering org unit in his capture scope, but has no access to the org unit the TE was transferred to.

Now it is there.

JoakimSM (approved):

PR #17287, opened 2024-05-02
fix: Add write access validation for relationships in Tracker Importer [DHIS2-17252]

enricocolasante:
Relationships were not validating write access for `from` and `to` items.
Now when importing a relationship, the rules are the following:
- Create bidirectional relationship: `from` and `to` must have write access
- Create unidirectional relationship: `from` must have write access and `to` must have write access
- Delete bidirectional relationship: `from` and `to` must have write access
- Delete unidirectional relationship: `from` must have write access `to`doesn't need any access

Relationships cannot be updated.
As the `SecurityOwnershipValidator` is accessing the `from` and `to` items, it must be moved after the constraint validator that checks the existence and validity of items.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
@@ -37,7 +37,6 @@ import org.hisp.dhis.program.Event;
 import org.hisp.dhis.program.Program;
 import org.hisp.dhis.relationship.Relationship;
 import org.hisp.dhis.user.UserDetails;
-import org.springframework.transaction.annotation.Transactional;
 
 /**
  * @author Morten Olav Hansen <mortenoh@gmail.com>
@@ -85,7 +84,6 @@ public interface TrackerAccessManager {
 
   List<String> canWrite(UserDetails user, Relationship relationship);
 
-  @Transactional(readOnly = true)
   List<String> canDelete(UserDetails user, @Nonnull Relationship relationship);
 
   /**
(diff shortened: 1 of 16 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackerAccessManager.java
```diff
@@ -83,6 +85,9 @@ List<String> canReadProgramAndTrackedEntityType(
 
   List<String> canWrite(UserDetails user, Relationship relationship);
 
+  @Transactional(readOnly = true)
+  List<String> canDelete(UserDetails user, @Nonnull Relationship relationship);
+
```
lucaCambi77: why is @transactional in the interface?

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_basic_metadata.json
```diff
@@ -1598,6 +1687,118 @@
...
       "password": "Test123###...",
       "attributeValues": []
+    },
+    {
+      "dataViewOrganisationUnits": [
+        {
+          "id": "QfUVllTs6cZ"
+        }
+      ],
+      "teiSearchOrganisationUnits": [],
```
muilpp, line 1697: if it's not too much trouble right now, can we replace it to t**eS**earchOrganisationUnits

muilpp (approved):

lucaCambi77 (approved):

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_basic_metadata.json
enricocolasante: We should change it but I would do it in another PR so we don't backport it to 2.41.0

PR #17288, opened 2024-05-02
feat: DataElement merge [DHIS2-14429]

david-mackessy:
# Summary
This change adds the `DataElement` merge feature.
A new endpoint `POST` `/api/dataElement/merge` has been added.
New required merge authority added `F_DATA_ELEMENT_MERGE`.
All new store methods added use HQL where possible (native used when JSONB functions required)

## Details
This `DataElement` merge change can be roughly thought of as: 
- try to merge source `DataElement`s into a target `DataElement`
- this involves retrieving all metadata/data with any source `DataElement` reference
- performing some action on that metadata
- for data, either deleting all source `DataValue`s / `EventDataValue`s or merging, using the last updated data value and deleting the rest
- deleting or keeping source `DataElement`s
- deleting or keeping `DataValueAudit`s / `TrackedEntityDataValueChangeLog`s

A full view of the metadata/data involved can be seen in the [Jira issue](https://dhis2.atlassian.net/browse/DHIS2-14429) or docs link below.

# Testing
## Automated
- tests added for all new store methods
- integration tests added for each metadata/data impacted by the merge
    - testing both deletion/keeping of source `DataElements`
- e2e tests added
    - validation tests
    - DB constraint test
    - valid merge
    - required authority test

## Manual
Some performance testing was carried out using the SL DB. This was mainly focused on `DataValue` and `EventDataValue` which are the areas of the system with potentially thousands/millions of entries.

### DataValue
- delete sources
- use `LAST_UPDATED` data merge strategy
- 155k source `DataValue`s

test request
```json
{
    "sources": [
        "Dkapzovo8Ll",
        "JIazHXNSnFJ"
    ],
    "target": "TCfIC3NDgQK",
    "deleteSources": true,
    "dataMergeStrategy": "LAST_UPDATED"
}
```
- outcome: test completes successfully in 10 seconds

### EventDataValue
- delete sources
- use `LAST_UPDATED` data merge strategy
- 145k source `Event`s
- 1.1 million  source `EventDataValue`s

test request
```json
{
    "sources": [
        "K6uUAvq500H",
        "X8zyunlgUfM",
        "ebaJjqltK5N"
    ],
    "target": "iSxGIwA3ZrV",
    "deleteSources": true,
    "dataMergeStrategy": "LAST_UPDATED"
}
```
- outcome: test completes successfully in 1 minute 10 seconds
- the feedback received around event data values being merged is that this scenario is not very likely to occur, so this test was more of an exploratory test in nature, because there are no technical limitations preventing this from this occurring.

# Docs
`DataElement` merge feature [docs PR here](https://github.com/dhis2/dhis2-docs/pull/1419)
[`DataElement` merge info and decision trees](https://dhis2.atlassian.net/wiki/spaces/SOFTWARE/pages/479887362/Merging+-+data+audit+discussion)

# Other info
During testing, it was observed that many audit triggers are actioned which explains longer execution times, depending on metadata/data involved (observed in the `EventDataValue`s performance test.

Some `WARN` logs were observed during testing like so:
```text
13:08:54.030  WARN [nio-8080-exec-5] o.h.d.e.DefaultExpressionService         : Invalid string token 'REPORTING_RATE' at line:1 character:71 parsing expression '(#{UOlfIjgN8X6.Prlt0C1RF0s}+#{UOlfIjgN8X6.V6L425pT3A0})*(#{BfMAe6Itzgt.REPORTING_RATE}/100)'
```
After looking into this, it looks like this is just down to bad data in the SL DB. None of the `DataElement`s tested with were ever referenced in the warning logs output.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dataelement/DataElementController.java
```diff
@@ -27,21 +27,64 @@
...
+
   private final DataElementService dataElementService;
+  private final MergeProcessor dataElementMergeProcessor;
+
+  @ResponseStatus(HttpStatus.OK)
+  @RequiresAuthority(anyOf = F_DATA_ELEMENT_MERGE)
+  @PostMapping(value = "/merge", produces = APPLICATION_JSON_VALUE)
+  public @ResponseBody WebMessage mergeDataElements(@RequestBody MergeParams params)
+      throws ConflictException {
+    log.info("Data element merge received");
```
jbee, line 73: Left over debug line?

jbee (commented):
That was a lot of code to get though - looks really good generally.

That were a lot of new query methods. I am in favor of "hand writing" the specific queries we need. In this instance it makes me feel more that we are paying a high price for a very niche feature. This is just about to be introduced and I already want to remove it, not because it has any issue but because it consts a lot for what we get out of it. Luckily I don't have to decide on that level :joy:  but I want to raise awareness for a feature with to me has poor cost/benefit ratio so that we think extra hard next time.

I am a bit confused by the merge related interface that are data element specific at the same time. This to me indicates their role is local. I probably would put them inside the class there they are used to move the code that changes together closer together and to not "pollute" the global namespace, mostly to not confuse readers with concepts that are not relevant outside of an implementation detail. 

As usual great work with all the tests. I skipped a bit not reading every line in detail - I trust they are of the usual quality ;)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElementGroupStore.java
```diff
@@ -27,6 +27,11 @@
...
 package org.hisp.dhis.dataelement;
 
+import java.util.Collection;
+import java.util.List;
 import org.hisp.dhis.common.IdentifiableObjectStore;
 
-public interface DataElementGroupStore extends IdentifiableObjectStore<DataElementGroup> {}
+public interface DataElementGroupStore extends IdentifiableObjectStore<DataElementGroup> {
+
+  List<DataElementGroup> getByDataElement(Collection<DataElement> dataElements);
```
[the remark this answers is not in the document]
david-mackessy: Sounds good. The initial validation for merging is done with `UID`s only already. That is a big change from what we currently do. We'd have to see how Hibernate queries would look when searching with `UID`s as opposed to using its own managed DB `ID`. Maybe there's some way we could set that up? But yeah, we should look into that.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataelement/DataElementService.java
```diff
@@ -172,6 +172,14 @@ public interface DataElementService {
    */
   List<DataElement> getDataElementsByAggregationLevel(int aggregationLevel);
 
+  /**
+   * Get all {@link DataElement}s by string value of {@link org.hisp.dhis.common.UID}
+   *
+   * @param dataElements data element {@link org.hisp.dhis.common.UID}s
+   * @return {@link DataElement}s found
+   */
+  List<DataElement> getDataElementsByUid(List<String> dataElements);
```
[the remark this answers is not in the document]
david-mackessy: it makes use of the existing underlying method [here](https://github.com/dhis2/dhis2-core/blob/8bedfd7bc08dcd6abad6bf9e2a372dd7919b0416/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java#L749) which, although it has `uid` in its name, uses strings. It's a bit of a grey area, if the javadoc is changed to say string, then the method should probably say string as well. I think we just have to accept these small mental transformations given the path we've already gone down. I know we're trying to move to `UID` but we're a very longer way from getting there. Baby steps.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/CommonMergeHandler.java
```diff
@@ -0,0 +1,107 @@
...
+
+      // denominators
+      List<Indicator> denominatorIndicators =
+          indicatorService.getIndicatorsWithDenominatorContaining(UID.of(source.getUid()));
+
+      denominatorIndicators.forEach(
+          i -> {
+            String existingDenominator = i.getDenominator();
+            i.setDenominator(existingDenominator.replace(source.getUid(), target.getUid()));
+          });
```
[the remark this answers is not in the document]
david-mackessy: Yep, I thought about this. I went with the more explicit and targeted approach (1 for each). It's debatable which would be better.
Best case scenario is that the Indicators have the same DE refs in both fields.
Worst case scenario is they don't. It would depend how the implementation has set up their system. Seems like it comes down to deciding whether we want:
(i) less IO (1 DB call) with potentially more in-memory code execution vs 
(ii) more IO (2 DB calls) with less in-memory code execution

happy to change it to 1 DB call. Only testing with large quantities of data would show the more performant approach. WDYT?

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/SqlUtils.java
```diff
@@ -146,4 +150,27 @@ public static String castToNumber(String value) {
   public static String lower(String value) {
     return "lower(" + value + ")";
   }
+
+  /**
+   * Method which will return a valid like predicate when there are 1 or many like targets. See
+   * {@link SqlUtilsTest#multiLikeMultiTest()} for example.
```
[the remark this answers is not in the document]
david-mackessy: it doesn't actually use the import. It shows as red in the IDE but still allows to jump to the test file which is nice.

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/SqlUtils.java
```diff
@@ -146,4 +150,27 @@ public static String castToNumber(String value) {
...
+   *
+   * @param column column to search
+   * @param likeTargets like targets, can be 1 or many
+   * @return like predicate
+   */
+  public static String multiLike(String column, List<String> likeTargets) {
```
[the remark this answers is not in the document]
david-mackessy: They both sound and have the same meaning to me. This isn't a deal breaker for me:) I can change this.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dataelement/DataElementController.java
david-mackessy: nope. Purposely left in. The logging in our system could be much better. We should be logging more, especially when it's not something that will be spamming the logs. It's great to see what's going on in the logs. Nothing worse that doing something in the system, checking the logs, only to see nothing has changed.

david-mackessy:
> In this instance it makes me feel more that we are paying a high price for a very niche feature. This is just about to be introduced and I already want to remove it, not because it has any issue but because it consts a lot for what we get out of it. Luckily I don't have to decide on that level 😂 but I want to raise awareness for a feature with to me has poor cost/benefit ratio so that we think extra hard next time.
> 
I hear ya, the comment is probably more suited at the project level, in Jira, or at at meeting where features are initially accepted/taken on.

> I am a bit confused by the merge related interface that are data element specific at the same time. This to me indicates their role is local. I probably would put them inside the class there they are used to move the code that changes together closer together and to not "pollute" the global namespace, mostly to not confuse readers with concepts that are not relevant outside of an implementation detail.
> 
moved them 👍

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/CommonMergeHandler.java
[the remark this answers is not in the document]
jbee: I was just thinking doing it together is less hassle. 1 query, one update per indicator. Not because of performance but simplicity of operation. But this isn't very important, just a thought and down to personal preference.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/dataelement/DataElementController.java
jbee: Generally I agree we could think more about where more logging makes sense. Not sure I find controller layer a good choice. Feels like that info should be possible to get with the right log level settings from the general HTTP container logs that I would assume are made by jetty or tomcat. but lets not make this complicated. we keep it and talk about the general log strategy in one of our meetings.

jbee (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/DataElementMergeService.java
```diff
@@ -0,0 +1,226 @@
...
+  private final CommonMergeHandler commonMergeHandler;
+  private final MergeValidator validator;
+  private final DataElementMergeValidator dataElementMergeValidator;
+  private final EntityManager entityManager;
+  private ImmutableList<DataElementMergeHandler> metadataMergeHandlers;
+  private ImmutableList<MergeHandler> commonMergeHandlers;
+  private ImmutableList<DataElementDataMergeHandler> dataMergeHandlers;
+  private ImmutableList<DataElementAuditMergeHandler> auditMergeHandlers;
+
+  @Override
```
vietnguyen, line 82: Is it intentional that `validate()` method doesn't have `@Transactional(readOnly=true)` ?

netroms (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/DataElementMergeService.java
david-mackessy: We could add it here as a wrapper on the internal read only transactions that are already marked with `@Transactional(readOnly=true)`? What do you think?
The UID validation goes through and uses [this method](@Transactional(readOnly=true)) which is marked read only.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/handler/DataDataElementMergeHandler.java
```diff
@@ -0,0 +1,228 @@
...
+        "Handling "
+            + dataValues.size()
+            + " non duplicate data values. Add new DataValue entry (using target DataElement ref) and delete old source entry");
+
+    dataValues.forEach(
+        sourceDataValue -> {
+          DataValue copyWithNewDataElementRef =
+              DataValue.dataValueWithNewDataElement(sourceDataValue, target);
+
+          dataValueStore.addDataValue(copyWithNewDataElementRef);
```
vietnguyen, line 212: - Do we have an estimation of how many dataValues would be saved in this for loop ?
- If the number is too big, I think we could consider to have implement batch insert here.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/DataElementMergeService.java
```diff
@@ -0,0 +1,226 @@
...
+    log.info("Deleting source DataElements");
+    for (DataElement source : sources) {
+      mergeReport.addDeletedSource(source.getUid());
+      dataElementService.deleteDataElement(source);
+    }
+  }
+
+  @PostConstruct
+  private void initMergeHandlers() {
+    metadataMergeHandlers =
```
vietnguyen, line 152: - I wonder if the requirement has a use case for skipping some of the merge steps in this list. Maybe for next version.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/handler/DataDataElementMergeHandler.java
david-mackessy: yep, I thought about batching here. After seeing the basic performance tests results I decided not to go down that road. Tests were done with 155k duplicate `DataValue`s and the merge was finishing in ~10 seconds (details in PR description)
This might be a `let's see what feedback we get on this` situation and only implement batching if needed. Definitely worth mentioning this though.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/DataElementMergeService.java
david-mackessy: possibly, although it seems to be an all or nothing feature, especially if the source `DataElement`s are being deleted (no refs can be left behind in the system)

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/DataElementMergeService.java
```diff
@@ -0,0 +1,226 @@
...
+    log.info("Handling DataElement reference associations and merges");
+    dataMergeHandlers.forEach(h -> h.merge(sources, target, request));
+    metadataMergeHandlers.forEach(h -> h.merge(sources, target));
+    commonMergeHandlers.forEach(h -> h.merge(sources, target));
+    auditMergeHandlers.forEach(h -> h.merge(sources, request));
+
+    // a flush is required here to bring Hibernate into a consistent state, due to some of the
+    // above merge operations involving required native queries (queries involving JSONB). This
+    // eliminates possible stale state issues.
+    entityManager.flush();
```
vietnguyen, line 134: Maybe you can utilize `FlushMode` for each query in order to make sure the query will not cause stale data.

david-mackessy: hmm, possibly. So it is currently set up to flush (once) at the end of all merges, which seems to work nicely. All DB constraints are adhered to and the native queried data behaves as expected  (in tests anyway).
Do you think flushing for some/all queries would be better somehow?

vietnguyen: - So the default FlushMode is AUTO, meaning we don't know when hibernate will flush the session.
- Therefore, we can't know if calling flush explicitly in this point is good or bad, depends on when is the last time hibernate flush the session.
- Another factor to consider flushing explicitly is when you have a very big list of objects in the persistent context. In that case, we would want implement batch process.
- As this is a new feature, I think we can wait for the feedbacks and see if there is any issue with this one.

PR #17295, opened 2024-05-02
fix: tracker/relationships emits inaccessible program attributes [ Dhis2-17090 ] [2.38]

lucaCambi77:
backport of https://github.com/dhis2/dhis2-core/pull/17281
@muilpp not sure you want to go that far with ACL, otherwise we discard this


muilpp:
In my opinion I'd not do it. We would still need to adapt the tracked entity endpoints which would take a lot of work and from next week on 2.38 is not supported anymore. 
It's a pity because this back port is already done, but merging only this one would not be consistent because some endpoints would validate the TE ownership correctly and others would still validate the TE registering unit if no program is provided.
@enricocolasante what do you think about it?

lucaCambi77:
@muilpp we can still reopen this issue in case

PR #17331, opened 2024-05-06
fix: add special filters to organisation units with member counts [BETA-158]

jbee:
### Summary
Adds in special filter parameters `query`, `level`, `maxLevel`, `withinUserHierarchy` and `withinUserSearchHierarchy` to the special endpoint that returns the OU list with member counts (has special parameters `memberObject` and `memberCollection`.
This means these can now be used in combination.

This endpoint is different from the other special endpoints in that it does not replace the loading of the result list itself. Instead it only post-processes the list adding the member counts. Therefore this endpoint also needs to support the same switch between the generic filters and the OU special filters listed above. This logic is present in the overridden `getObjectList` that handles normal listing (`/api/organisationUnits`).

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/organisationunit/OrganisationUnitController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/organisationunit/OrganisationUnitController.java
@@ -314,36 +314,60 @@ public class OrganisationUnitController extends AbstractCrudController<Organisat
       getOrganisationUnitsWithMemberCount(
           @RequestParam String memberObject,
           @RequestParam String memberCollection,
+          @RequestParam(required = false) String query,
+          @RequestParam(required = false) Integer level,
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/organisationunit/OrganisationUnitController.java
```diff
@@ -483,21 +507,36 @@ public ResponseEntity<StreamingJsonRoot<OrganisationUnit>> getObjectList(
...
-
-          p.setParents(
-              withinUserHierarchy == Boolean.TRUE
-                  ? currentUser.getOrganisationUnits()
-                  : withinUserSearchHierarchy == Boolean.TRUE
-                      ? currentUser.getTeiSearchOrganisationUnitsWithFallback()
-                      : Set.of());
-
-          return organisationUnitService.getOrganisationUnitsByQuery(p);
-        });
```
jbee, line 500: FYI: This part only got extracted to be reused

PR #17368, opened 2024-05-09
chore: remove unused old tracker code DHIS2-16564

teleivo:
follow up after https://github.com/dhis2/dhis2-core/pull/17358


enricocolasante (approved):
Looks good and feels good.
One thing to discuss, it doesn't need to be now, is about `EnrollmentStatus`.
Analytics has Tracked as a dependency, and it makes sense to be that way, so does it make sense to have an `EnrollmentStatus` in analytics or tracker module should export its model somehow?

teleivo:
> One thing to discuss, it doesn't need to be now, is about `EnrollmentStatus`.
> Analytics has Tracked as a dependency, and it makes sense to be that way, so does it make sense to have an `EnrollmentStatus` in analytics or tracker module should export its model somehow?

Analytics has a TODO https://github.com/dhis2/dhis2-core/pull/17368/files#diff-ed56f91769d03826c739a97ac89b491abb02f6138d26b6d12dc9746788c8684dR33 on that one. Maikel is aware and they will fix their TODO after we merge. So the `EnrollmentStatus` should be removed.

muilpp (approved):

PR #17372, opened 2024-05-09
fix: Validate unenrolled TE before importing [DHIS2-17280]

muilpp:
When importing a new enrollment with the CREATE strategy, it's possible that there's no owner of the enrollment TE/program pair if the TE is also imported at the same time. Previously, we were not handling this case correctly because we were not validating the ownership. 
With this fix, we will retrieve the registering unit and validate whether the user has access to it.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
@@ -39,6 +39,7 @@ import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.program.Program;
 import org.hisp.dhis.security.Authorities;
 import org.hisp.dhis.security.acl.AclService;
+import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.trackedentity.TrackedEntityProgramOwnerOrgUnit;
 import org.hisp.dhis.trackedentity.TrackerOwnershipManager;
 import org.hisp.dhis.tracker.imports.TrackerImportStrategy;
@@ -75,16 +76,12 @@ class SecurityOwnershipValidator implements Validator<Enrollment> {
         strategy.isUpdateOrDelete()
             ? bundle.getPreheat().getEnrollment(enrollment.getEnrollment()).getProgram()
             : bundle.getPreheat().getProgram(enrollment.getProgram());
-    String trackedEntity =
+    TrackedEntity trackedEntity =
         bundle.getStrategy(enrollment).isDelete()
-            ? bundle
-                .getPreheat()
-                .getEnrollment(enrollment.getEnrollment())
-                .getTrackedEntity()
-                .getUid()
-            : enrollment.getTrackedEntity();
-    OrganisationUnit ownerOrgUnit =
-        getOwnerOrganisationUnit(preheat, enrollment.getTrackedEntity(), program);
+            ? bundle.getPreheat().getEnrollment(enrollment.getEnrollment()).getTrackedEntity()
+            // TODO If it's create, will the tracked entity be in the database?
+            : bundle.getPreheat().getTrackedEntity(enrollment.getTrackedEntity());
+    OrganisationUnit ownerOrgUnit = getOwnerOrganisationUnit(preheat, trackedEntity, program);
 
     checkEnrollmentOrgUnit(reporter, bundle, strategy, enrollment);
 
@@ -98,15 +95,16 @@ class SecurityOwnershipValidator implements Validator<Enrollment> {
       }
     }
 
-    checkWriteEnrollmentAccess(reporter, bundle, enrollment, program, ownerOrgUnit, trackedEntity);
+    checkWriteEnrollmentAccess(
+        reporter, bundle, enrollment, program, ownerOrgUnit, trackedEntity.getUid());
   }
 
   private OrganisationUnit getOwnerOrganisationUnit(
-      TrackerPreheat preheat, String teUid, Program program) {
+      TrackerPreheat preheat, TrackedEntity trackedEntity, Program program) {
     Map<String, TrackedEntityProgramOwnerOrgUnit> programOwner =
-        preheat.getProgramOwner().get(teUid);
+        preheat.getProgramOwner().get(trackedEntity.getUid());
     if (programOwner == null || programOwner.get(program.getUid()) == null) {
-      return null;
+      return trackedEntity.getOrganisationUnit();
     } else {
       return programOwner.get(program.getUid()).getOrganisationUnit();
     }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
@@ -32,7 +32,6 @@ import static org.hisp.dhis.tracker.imports.validation.ValidationCode.E1103;
 import static org.hisp.dhis.tracker.imports.validation.validator.TrackerImporterAssertErrors.USER_CANT_BE_NULL;
 
 import java.util.Map;
-import java.util.Optional;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -98,15 +96,39 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
...
+
+      return te.map(
+              entity -> {
+                TrackedEntity newEntity = new TrackedEntity();
+                newEntity.setUid(entity.getUid());
+                newEntity.setOrganisationUnit(
+                    bundle.getPreheat().getOrganisationUnit(entity.getOrgUnit()));
+                return newEntity;
+              })
+          .get();
```
muilpp, line 120: @enricocolasante is it fine to assume the tracked entity will always be present in the payload when not found in the database?

enricocolasante: Yes, we already checked that it exists in [MetaValidator](https://github.com/dhis2/dhis2-core/blob/5da7aa8724d65a3fda83c57afa7bba1962c3d66b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/MetaValidator.java#L55-L59)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/TrackedEntityImportValidationTest.java
```diff
@@ -292,7 +292,7 @@ void shouldFailToUpdateWhenUserHasAccessToRegistrationUnitAndTEWasTransferred()
     assertNoErrors(trackerImportService.importTracker(new TrackerImportParams(), trackerObjects));
     trackerObjects = fromJson("tracker/validations/enrollments_te_enrollments-data_2.json");
     TrackerImportParams trackerImportParams = new TrackerImportParams();
-    trackerImportParams.setUserId(USER_10);
+    trackerImportParams.setUserId(USER_8);
```
teleivo, line 295: hard to understand this change 😅 what makes this user "special"? could you describe it in a javadoc on the constant?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -75,16 +77,12 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
...
-            ? bundle
-                .getPreheat()
-                .getEnrollment(enrollment.getEnrollment())
-                .getTrackedEntity()
-                .getUid()
-            : enrollment.getTrackedEntity();
-    OrganisationUnit ownerOrgUnit =
-        getOwnerOrganisationUnit(preheat, enrollment.getTrackedEntity(), program);
+    TrackedEntity trackedEntity =
+        bundle.getStrategy(enrollment).isUpdateOrDelete()
```
teleivo, line 81: Can you move all of this into a `getTrackedEntity()`? I don't really care how you get it when reading this top level validate method. The way its now is a bit mixed with some bundle/preheat lookup in here and some in `getTrackedEntityWhenStrategyCreate`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -98,15 +96,39 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
...
 
-    checkWriteEnrollmentAccess(reporter, bundle, enrollment, program, ownerOrgUnit, trackedEntity);
+    checkWriteEnrollmentAccess(
+        reporter, bundle, enrollment, program, ownerOrgUnit, trackedEntity.getUid());
+  }
+
+  private TrackedEntity getTrackedEntityWhenStrategyCreate(
+      TrackerBundle bundle, Enrollment enrollment) {
+    TrackedEntity trackedEntity =
+        bundle.getPreheat().getTrackedEntity(enrollment.getTrackedEntity());
```
teleivo, line 106: I would return early if `trackedEntity` is non null. Then chain all the operations on the `bundle.findTrackedEntityByUid` into the return expression.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -98,15 +96,39 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
...
+
+    if (trackedEntity == null) {
+      Optional<org.hisp.dhis.tracker.imports.domain.TrackedEntity> te =
+          bundle.findTrackedEntityByUid(enrollment.getTrackedEntity());
+
+      return te.map(
+              entity -> {
+                TrackedEntity newEntity = new TrackedEntity();
+                newEntity.setUid(entity.getUid());
+                newEntity.setOrganisationUnit(
```
teleivo, line 116: Why is that mapping necessary?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EnrollmentSecurityImportValidationTest.java
```diff
@@ -280,4 +290,88 @@ void testUserHasNoAccessToProgramTeiType() throws IOException {
 
     assertHasOnlyErrors(importReport, ValidationCode.E1104);
   }
+
+  @Test
+  void shouldFailWhenTeNotEnrolledAndUserHasNoAccessToTeRegisteringOrgUnit() throws IOException {
+    clearSecurityContext();
```
teleivo, line 296: Can you think of a way to make this test setup more readable? Its really hard to see what the reason is why this test will fail the import and why the other one will make it pass 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
muilpp: Because the TrackedEntity object returned by findTrackedEntityByUid is not the one the validator expects, so I map it to the expected object.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/TrackedEntityImportValidationTest.java
muilpp: That's the problem of having all the test configuration in a json file.
What makes this user special is the org units in its search scope. I could add a javadoc, but if for any reason we change the org units of this user, I'm pretty sure we'll forget about the javadoc and it won't be consistent anymore.
To me, it would be better rely less on json and configure the setup programatically, that way it makes it's easier to understand settings like this one.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -100,27 +95,32 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
         reporter, bundle, enrollment, program, ownerOrgUnit, trackedEntity.getUid());
   }
 
+  private TrackedEntity getTrackedEntity(TrackerBundle bundle, Enrollment enrollment) {
+    return bundle.getStrategy(enrollment).isUpdateOrDelete()
```
teleivo, line 99: I think it can all just be in one method

teleivo (approved):

PR #17398, opened 2024-05-13
chore: adjust tests and previous filter endpoint permission

netroms:
### Summary
Adjusts some tests and minor controller refactoring related to impersonation. Also removes unused URL permissions in the security config.

Original PR: [#17342](https://github.com/dhis2/dhis2-core/pull/17342)

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
@@ -70,6 +70,12 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
+/**
+ * Controller for impersonating users. Most of the code is copied from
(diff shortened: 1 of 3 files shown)


vietnguyen (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
```diff
@@ -130,12 +129,23 @@ public ImpersonateUserResponse impersonateUser(
     }
   }
 
+  private void validateReq(HttpServletRequest request) throws ForbiddenException {
+    boolean enabled = config.isEnabled(ConfigurationKey.SWITCH_USER_FEATURE_ENABLED);
+    if (!enabled) {
+      throw new ForbiddenException("Forbidden, user not allowed to impersonate user");
+    }
+    if (!hasAllowListedIp(request.getRemoteAddr())) {
+      throw new ForbiddenException("Forbidden, user not allowed to impersonate user");
```
david-mackessy, line 138: could these `if` checks be combined if the outcome is exactly the same?
or, if the allow listed IP would have a different error message, then this would make more sense.

david-mackessy (approved):

PR #17405, opened 2024-05-13
fix: custom format for integer types [DHIS2-15841]

gnespolino:
Previously, when exporting a data visualizer table containing columns with integer values, the output would display numbers ending with a comma or dot. This issue arose from the usage of a custom number format in Excel, which works well for numbers with decimals but caused the issue.
This PR introduces a specific formatting mechanism for integer values so that the decimal part is skipped when the type of the numeric value is an integer.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
@@ -349,7 +349,7 @@ public class GridUtils {
       for (Object column : columns) {
         if (column != null && Number.class.isAssignableFrom(column.getClass())) {
           Cell cell = xlsRow.createCell(columnIndex++, CellType.NUMERIC);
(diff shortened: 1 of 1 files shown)


d-bernat (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
```diff
@@ -344,7 +349,11 @@ private static void toXlsInternal(
       for (Object column : columns) {
         if (column != null && Number.class.isAssignableFrom(column.getClass())) {
           Cell cell = xlsRow.createCell(columnIndex++, CellType.NUMERIC);
-          cell.setCellStyle(numberCellStyle);
+          if (column instanceof Integer || column instanceof Long || column instanceof BigInteger) {
```
maikelarabori, line 352: A method like: `isIntegerType(column)` would be easier to read.

maikelarabori (approved):

PR #17409, opened 2024-05-13
fix: change exit roles on impersonation

netroms:
### Summary
Fixes previous PR making the impersonateExit endpoint too strict. This change make sure the impersonated user is able to exit, without having to have either ALL or F_IMPERSONATE_USER authority. A special F_PREVIOUS_IMPERSONATOR_AUTHORITY is added to the impersonated user, making it possible to exit the impersonating mode.

### Automated test:
ImpersonateUserControllerTest#testImpersonateUserOKAndExit

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
@@ -63,7 +63,6 @@ import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.security.web.authentication.WebAuthenticationDetailsSource;
 import org.springframework.security.web.authentication.switchuser.AuthenticationSwitchUserEvent;
-import org.springframework.security.web.authentication.switchuser.SwitchUserAuthorityChanger;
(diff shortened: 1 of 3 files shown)


jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/ImpersonateUserController.java
```diff
@@ -125,18 +125,37 @@ public ImpersonateUserResponse impersonateUser(
...
+      return;
+    }
+    Collection<? extends GrantedAuthority> authorities = userDetails.getAuthorities();
+    for (GrantedAuthority authority : authorities) {
+      if (authority.getAuthority().equals(F_IMPERSONATE_USER.name())
+          || authority.getAuthority().equals(F_PREVIOUS_IMPERSONATOR_AUTHORITY.name())) {
+        return;
+      }
+    }
+    throw new ForbiddenException("Forbidden, requires authority [F_IMPERSONATE_USER]");
```
david-mackessy, line 149: looks like this method looks for either `F_IMPERSONATE_USER` or `F_PREVIOUS_IMPERSONATOR_AUTHORITY`, but only `F_IMPERSONATE_USER` is in the error response. Might be misleading for the user. Should `F_PREVIOUS_IMPERSONATOR_AUTHORITY` be included in the error response?

netroms: Added!

PR #17421, opened 2024-05-14
chore: cleanup dhis-api/dhis-service-core DHIS2-16564

teleivo:
follow up after https://github.com/dhis2/dhis2-core/pull/17368

We have deleted the old tracker controllers, sync and dxf2 services and stores. This PR now removes the unused code that was previously used by the tracker dxf2 services. This mostly means interfaces in the module dhis-api and implementations in the module dhis-service-core.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentService.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentService.java
@@ -34,7 +34,6 @@ import org.hisp.dhis.common.IllegalQueryException;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.user.User;
-import org.hisp.dhis.user.UserDetails;
 
 /**
  * @author Abyot Asalefew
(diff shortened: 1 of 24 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentService.java
```diff
@@ -142,14 +125,6 @@ public interface EnrollmentService {
    */
   List<Enrollment> getEnrollments(EnrollmentQueryParams params);
```
enricocolasante, line 126: Matbe not for this PR but this seems to be used only in tests and it should allow to remove `EnrollmentQueryParams` altogether

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EnrollmentService.java
teleivo: I thought finding unused code should be easy but it's not 😅 I haven't found a way to find unused code or code only used in tests on a larger scale in Intellij. I might have to go service by service and method by method 😬 

I will follow up with another PR.

PR #17424, opened 2024-05-14
fix: field filter parsing exist block [DHIS2-17391] (2.41.0)

jbee:
cherry-pick backport of #17422

OBSS! **Not** RCB approved!


jbee:
Not considered to be in the release version.

PR #17431, opened 2024-05-14
fix: Validate TE relationship items [DHIS2-16460]

muilpp:
When fetching TEs with relationships, if one of the items in a given relationship is not readable, we raise an exception.
Instead, we need to return all readable relationships and skip the ones that are not.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-16460


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -248,12 +248,9 @@ public TrackedEntity getTrackedEntity(
    *
    * @return the TE object if found and accessible by the current user
    * @throws NotFoundException if uid does not exist
-   * @throws ForbiddenException if TE registration org unit not in user's scope or not enough
-   *     sharing access
    */
   private TrackedEntity getTrackedEntity(
```
enricocolasante, line 252: Maybe it would make sense to explain in the name of the method or in the JavaDoc that this method is meant to be used only for relationships.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -248,12 +248,9 @@ public TrackedEntity getTrackedEntity(
...
-      String uid, TrackedEntityParams params, boolean includeDeleted)
-      throws NotFoundException, ForbiddenException {
+      String uid, TrackedEntityParams params, boolean includeDeleted) throws NotFoundException {
     TrackedEntity trackedEntity = trackedEntityStore.getByUid(uid);
     addTrackedEntityAudit(trackedEntity, CurrentUserUtil.getCurrentUsername());
     if (trackedEntity == null) {
```
muilpp, line 256: @enricocolasante should I also return null instead of throwing a `NotFoundException` here?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -263,7 +260,7 @@ private TrackedEntity getTrackedEntity(
 
     List<String> errors = trackerAccessManager.canRead(currentUser, trackedEntity);
     if (!errors.isEmpty()) {
-      throw new ForbiddenException(errors.toString());
+      return null;
     }
```
lucaCambi77: maybe `try catch` where needed and throw an exception otherwise?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1794,4 +1839,15 @@ private void makeProgramMetadataInaccessible(Program program) {
     program.getSharing().setPublicAccess(AccessStringHelper.DEFAULT);
     manager.update(program);
   }
+
+  private TrackedEntityOperationParams createOperationParams(
```
teleivo, line 1843: I would not create this helper as:
* all other tests lay out how the service is called which often shows what the test tests
* and it's not clear what operation params you are creating here 🤷🏻 

The setup of the data is much more in the way of test readability IMHO than the lines of calls to the params builder.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1381,11 +1381,56 @@ void shouldReturnTrackedEntityWithRelationshipsTei2Tei()
...
+    injectSecurityContextUser(superuser);
+    OrganisationUnit inaccessibleOrgUnit = createOrganisationUnit('D');
+    manager.save(inaccessibleOrgUnit);
+    makeProgramMetadataInaccessible(programB);
+    makeProgramMetadataInaccessible(programC);
+    trackerOwnershipManager.assignOwnership(
+        trackedEntityA, programA, inaccessibleOrgUnit, true, true);
+    TrackedEntityOperationParams operationParams = createOperationParams(orgUnitB, trackedEntityB);
+
+    injectSecurityContextUser(user);
```
teleivo, line 1396: This belongs to the setup block

teleivo (approved):

PR #17437, opened 2024-05-15
fix: ReportController throws exception for jasperreport

vietnguyen:
- Follow up on https://github.com/dhis2/dhis2-core/pull/17382.
- After downgrading JasperReport library, the report service still throws exception because of conflicts between JasperReport `6.20.0` and JFreeChart `1.5.x`.
- This PR downgrade JFreeChart to `1.0.19` which is compatible with JasperReport `6.20.0`.
- There are some JFreeChart methods naming changes fixed as a result of downgrading. 
- I have update the unit test to include full report design to make sure no exception thrown.


jbee (approved):

david-mackessy (approved):
the version downgrade (`1.5.x` -> `1.0.x`) seems like a big drop, are there any adverse consequences of that which we should be aware of? e.g. CVE or security/sonar issues

vietnguyen:
> the version downgrade (`1.5.x` -> `1.0.x`) seems like a big drop, are there any adverse consequences of that which we should be aware of? e.g. CVE or security/sonar issues

They jumped from 1.0.19 to 1.5.0, so I would say it is not really a big drop. More details https://github.com/jfree/jfreechart/tags.

david-mackessy:
> > the version downgrade (`1.5.x` -> `1.0.x`) seems like a big drop, are there any adverse consequences of that which we should be aware of? e.g. CVE or security/sonar issues
> 
> They jumped from 1.0.19 to 1.5.0, so I would say it is not really a big drop. More details https://github.com/jfree/jfreechart/tags.

It might not be a big drop version-wise but it's a big drop time-wise (1.0.19 was published in 2016!) If there are no issues with that then it's fine. It's unusual that such an old dep would have no issues. Maybe they're just really good :)

vietnguyen:
> > > the version downgrade (`1.5.x` -> `1.0.x`) seems like a big drop, are there any adverse consequences of that which we should be aware of? e.g. CVE or security/sonar issues
> 
> > 
> 
> > They jumped from 1.0.19 to 1.5.0, so I would say it is not really a big drop. More details https://github.com/jfree/jfreechart/tags.
> 
> 
> 
> It might not be a big drop version-wise but it's a big drop time-wise (1.0.19 was published in 2016!) If there are no issues with that then it's fine. It's unusual that such an old dep would have no issues. Maybe they're just really good :) 

I guess there could be security issues, will check later. Nevertheless, if we don't downgrade then we would need to remove the JasperReport function and also refactoring the PDF export function to use another library. That is in 42 road map.

PR #17441, opened 2024-05-15
fix: Add programs param if no program requested [DHIS2-17410][2.39]

muilpp:
The last [ACL fix](https://github.com/dhis2/dhis2-core/pull/17357/files#diff-f960125887353fe285946328173a4e612632e40c0bca283e9d30c824348764dbR560) broke one of the methods used by the metadata sync. This PR contains a fix for it and it introduces a couple of tests to make sure it won't happen again


enricocolasante (approved):

david-mackessy:
thanks @muilpp for the quick resolution, much appreciated 👍

PR #17454, opened 2024-05-16
feat: Create StartDate and EndDate types [DHIS2-16019]

enricocolasante:
We are adding 2 new web parameter types for date filters `StartDate` and `EndDate`.
In a lot of APIs we have filters to get entities before and after something and most of them are of type `DateTime`.

Everything works correctly if the client specifies the time together with the date but if only the date is passed then the `before` filter default to the beginning of the day filtering out all the entities with the date in that day.

Now
- `StartDate` passed without a time, defaults to the beginning of the day (some behavior already present in `DateTime` type)
- `EndDate` passed without a time, defaults to the end of the day

`StartDate` and `EndDate` are meant to be used only in the web layer.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/DateUtils.java
```diff
@@ -672,6 +675,28 @@ public static Date parseDate(String dateString) {
...
+    if (StringUtils.isEmpty(dateString)) {
+      return null;
+    }
+
+    try {
+      Date date = safeParseDateTime(dateString, ONLY_DATE_FORMATTER);
+      org.joda.time.LocalDateTime localDateTime = org.joda.time.LocalDateTime.fromDateFields(date);
+      return localDateTime.withMillisOfDay(localDateTime.millisOfDay().getMaximumValue()).toDate();
+    } catch (IllegalArgumentException e) {
+      // dateString has time defined
```
teleivo, line 695: is that the only reason for an IllegalArgumentException?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/StartDate.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * StartDate represents a lower limit date used to filter results in search APIs.
+ *
+ * <p>StartDate accepts date and time to be defined. If no time is defined, then the time at the
+ * beginning of the day is used by default.
+ *
+ * <p>This behavior, combined with {@link EndDate}, allows to correctly implement an interval search
+ * including start and end dates. startDate=2020-10-10&endDate=2020-10-12 will include anything
+ * between 2020-10-10T00:00:00.000 and 2020-10-12T23:59:59.999.
```
teleivo, line 41: Should we leave

`startDate=2020-10-10&endDate=2020-10-12 will include anything between 2020-10-10T00:00:00.000 and 2020-10-12T23:59:59.999.`

from these docs here as this is at the discretion of the endpoint itself?

Or do we "standardize" on this behavior, in that case it makes sense to keep it here. I am not sure its 100% clear right now with `include anything between` that the endpoints are included as well. So maybe we could change the wording or add `(represent a closed interval [startDate, endDate]`.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatesBindingTest.java
```diff
@@ -0,0 +1,112 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.webapi.controller;
+
+import static org.hamcrest.core.StringContains.containsString;
```
teleivo, line 30: Please use our https://github.com/dhis2/dhis2-core/blob/45307f6feacc87a2b918a67fa74b778178c11ef9/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/utils/Assertions.java#L180

https://github.com/dhis2/wow-backend/blob/master/guides/testing/testing.md#assertions

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -150,6 +152,13 @@ public CrudControllerAdvice() {
 
   @InitBinder
   protected void initBinder(WebDataBinder binder) {
+    binder.registerCustomEditor(
```
teleivo, line 155: I wanted to talk to you about Spring binding as I learned something while you were out for a bit. See https://github.com/dhis2/dhis2-core/pull/16673

I think we can simplify this. Our type needs to have a constructor or valueOf factory to turn a String into a Start/EndDate. We then do not need to register anything and rely on Springs conversion system. See PR description on how it works.

teleivo (commented):
> StartDate and EndDate are meant to be used only in the web layer.

So these are used to convert the HTTP query parameters into a `Date` which is what we use in the service/repo layer?

I think it would make sense to already use these types in some controller in this PR.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/DateUtils.java
```diff
@@ -672,6 +675,28 @@ public static Date parseDate(String dateString) {
...
+   * @return a date.
+   */
+  public static Date parseDateEndOfTheDay(String dateString) {
+    if (StringUtils.isEmpty(dateString)) {
+      return null;
+    }
+
+    try {
+      Date date = safeParseDateTime(dateString, ONLY_DATE_FORMATTER);
+      org.joda.time.LocalDateTime localDateTime = org.joda.time.LocalDateTime.fromDateFields(date);
```
muilpp, line 692: Could we do this without relying on an external library?
Here's [an example](https://stackoverflow.com/questions/10308356/how-to-obtain-the-start-time-and-end-time-of-a-day) with java8

enricocolasante: Done

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
enricocolasante: Yesss.
I was confused by the code as it wasn't really as I remembered :grimacing: 
Changed to use Springs conversion

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatesBindingTest.java
enricocolasante: I am using it inside the `mvcMock` framework.
I am not asserting anything directly.

muilpp (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParams.java
```diff
@@ -139,9 +141,9 @@ public class EventRequestParams implements PageRequestParams {
 
   private Date scheduledBefore;
 
-  private Date updatedAfter;
+  private StartDate updatedAfter;
```
teleivo, line 144: These types should get an OpenAPI description. I think @jbee is changing how descriptions can be added. Maybe create a task that we add a description once his changes are through.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/EndDate.java
```diff
@@ -40,14 +43,12 @@
  * search including start and end dates. startDate=2020-10-10&endDate=2020-10-12 will include
  * anything between 2020-10-10T00:00:00.000 and 2020-10-12T23:59:59.999.
  */
-@RequiredArgsConstructor
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter
```
teleivo, line 47: `Date` is a mutable type. Should we return a copy of the `date` to make this class immutable?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatesBindingTest.java
```diff
@@ -54,10 +54,7 @@ class DatesBindingTest {
 
   @BeforeEach
   public void setUp() {
-    mockMvc =
-        MockMvcBuilders.standaloneSetup(new BindingController())
-            .setControllerAdvice(new CrudControllerAdvice())
-            .build();
+    mockMvc = MockMvcBuilders.standaloneSetup(new BindingController()).build();
```
teleivo, line 57: Can you adopt asserting like we do in dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/OrderBindingTest.java. So instead of asserting on strings in the response body use the actual type.

Also we replaced the word `Criteria` with `RequestParams` in the web layer so I would not add new code even test code using the word `criteria`.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SpringBindingTest.java
```diff
@@ -52,10 +52,7 @@ class SpringBindingTest {
 
   @BeforeEach
   public void setUp() {
-    mockMvc =
-        MockMvcBuilders.standaloneSetup(new BindingController())
-            .setControllerAdvice(new CrudControllerAdvice())
-            .build();
+    mockMvc = MockMvcBuilders.standaloneSetup(new BindingController()).build();
```
teleivo, line 55: There is enum related code in the `CrudControllerAdvice`. Not sure what it does but I wonder if `CrudControllerAdvice` should stay in here or the code in `CrudControllerAdvice` relating enums should be removed if it has no influence. Either way maybe do this in a separate PR.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
```diff
@@ -139,8 +139,14 @@ public EventOperationParams map(EventRequestParams eventRequestParams)
             .occurredBefore(eventRequestParams.getOccurredBefore())
             .scheduledAfter(eventRequestParams.getScheduledAfter())
             .scheduledBefore(eventRequestParams.getScheduledBefore())
-            .updatedAfter(eventRequestParams.getUpdatedAfter())
-            .updatedBefore(eventRequestParams.getUpdatedBefore())
+            .updatedAfter(
+                eventRequestParams.getUpdatedAfter() != null
+                    ? eventRequestParams.getUpdatedAfter().getDate()
+                    : null)
```
jbee, line 145: This type of null unpacking will occur a lot so we should support it with a static helper function to improve readability . Also I wonder if the should have a `toDate()` method instead of `getDate()` to make it more transparent if the `Date` is actually a field called `date` or if we store the date internally in a different way. We might want to go `java.time` API except where we interface with code that still is on `Date`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/EndDateTime.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <p>This behavior, combined with {@link StartDateTime}, allows to correctly implement an interval
+ * search including start and end dates.
+ */
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter
+public class EndDateTime {
+  private final Date date;
+
+  public static EndDateTime valueOf(String date) {
```
jbee, line 50: Javadoc on the list of valid patterns for `date` would be nice. Is this `YYYYmmdd` only?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/StartDateTime.java
```diff
@@ -0,0 +1,58 @@
...
+ * beginning of the day is used by default.
+ *
+ * <p>This behavior, combined with {@link EndDateTime}, allows to correctly implement an interval
+ * search including start and end dates.
+ */
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public class StartDateTime {
+  private final Date date;
+
+  public static StartDateTime valueOf(String date) {
```
jbee, line 48: Not a strong opinion but in newer Java APIs this is usually just called `of`, e.g. `List.of`, `LocalDate.of`

jbee (approved):
Happy to see this being introduced. As far as I could see from a glance this should work nicely. Some suggestions on API details.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java
enricocolasante: It would be nice to have everything on `java.time`.
For now I am relying on `DateUtils` methods and most of them are using `java.util.Date`.
I will check in next PR how much work is to use `java.time` in here.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/StartDateTime.java
teleivo: `of` should work as well with Springs conversion system [as it looks for](https://github.com/spring-projects/spring-framework/blob/d81619addd6d9060c9d8dc454c359054e517ed98/spring-core/src/main/java/org/springframework/core/convert/support/ObjectToObjectConverter.java#L191-L197) either a constructor or a factory called `of`, `valueOf` or `from`.

(see https://github.com/dhis2/dhis2-core/pull/16673)

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/DatesBindingTest.java
```diff
@@ -0,0 +1,115 @@
...
+      actualStartDateTime = applyIfNotNull(params.getAfter(), StartDateTime::toDate);
+      actualEndDateTime = applyIfNotNull(params.getBefore(), EndDateTime::toDate);
+      return ok();
+    }
+  }
+
+  @NoArgsConstructor
+  @Data
+  private class Params {
+    private StartDateTime after;
```
teleivo, line 111: `start` and `end` would probably make it easier to read the test as one does not need to translate what `before` refers to.

teleivo (approved):

PR #17475, opened 2024-05-17
ci: add stale action to close old PRs

teleivo:


teleivo:
@stian-sandvold this is based on https://github.com/openmrs/openmrs-core/blob/master/.github/workflows/stale.yml which I added some years ago. I phrased it differently since we hardly receive outside contributions. Is there some link/communication channel I should add though? How would one contact someone in case they are not part of the core team? Just requesting a review might go unnoticed.

stian-sandvold:
Looks great @teleivo ,

as for outside contributors, I don't have a good solution there at the moment. Maybe we could make sure that we reach out to external contributors when we see them, and set up some kind of channel for them

stian-sandvold (approved):

maikelarabori (approved):

PR #17489, opened 2024-05-20
fix:Incorrect rule actions while rule-engine context loading[DHIS2-17414]

zubaira:
[DHIS2-17414](https://dhis2.atlassian.net/browse/DHIS2-17414)

[DHIS2-17414]: https://dhis2.atlassian.net/browse/DHIS2-17414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleStore.java
+++ b/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleStore.java
@@ -29,6 +29,7 @@ package org.hisp.dhis.programrule.hibernate;
 
 import java.math.BigInteger;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.persistence.EntityManager;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleStore.java
```diff
@@ -110,17 +112,52 @@ public List<ProgramRule> getProgramRulesByActionTypes(
...
-            + "AND (pr.programStage IS NULL OR ps.uid = :programStageUid )";
-
-    return getQuery(hql)
-        .setParameter("programId", program)
-        .setParameter("implementableTypes", types)
-        .setParameter("programStageUid", programStageUid)
-        .getResultList();
+    String hql =
+        """
+                SELECT distinct pr FROM ProgramRule pr JOIN FETCH pr.programRuleActions pra
```
enricocolasante, line 117: Could we use native sql here?
It would be easier to understand what is happening and we would have more control on this query.
We don't need the `FETCH` keyword here as we are getting the actions in the next query.
In this query we just need to filter out program rules with no actions of certain types

zubaira: converting this into native should not be an issue.
But I do think we need join with `ProgramRuleAction` to filter rules based on `ProgramRuleActionType`.

enricocolasante: We need the join, we do not need the fetch part

zubaira: fixed

muilpp (approved):

PR #17490, opened 2024-05-20
fix:(2.40) Incorrect rule actions while rule-engine context loading[DHIS2-17414]

zubaira:


zubaira:
bug does not exist in this version

PR #17494, opened 2024-05-20
feat: embedded Tomcat and web restructuring, part 1 of 2

netroms:
## Summary
This is part 1 of 2 PRs that remove the `dhis-web-portal` and `dhis-web-common-resources` modules and unify them into one new module, `dhis-web-server`.
This new module uses the `spring-boot-maven-plugin` to create an executable .war file that can be started like an executable jar and deployed as a regular .war file in an application server like Tomcat.
Part 2 of this PR duo only changes the location of `dhis-web-common-resources`. Because part 2 only contains static resource assets and not any Java files, it is kept separate to avoid polluting this PR review with 1000+ file movements.
The intention is also to merge the `dhis-web-apps` into `dhis-web-server`, when a new, better way of building the apps has been developed.

The new `Main` class in the `dhis-web-server` is the new embedded Tomcat's entry point and start class. This new code is mostly borrowed from the Spring Boot project and has not been optimized and thoroughly tested. It currently lacks many configuration options that a full production-ready server needs. This should be addressed in upcoming PRs.

Since this PR lacks the correct static resource location, which is being done in part 2, some images, CSS, etc., are missing when starting the server. 

The new embedded Tomcat server has not yet undergone a security audit. It needs to be investigated to determine whether its handling of temporary directories is compatible and safe with the DHIS2 server practices.

### Build types
* Docker build (Used by the API tests in the GitHub Actions pipeline)
* Executable .war file **(NEW)**
* Deployable .war
* Embedded server, start from main(), for use in the IDE


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/ContentNegotiationConfig.java
```diff
@@ -58,4 +60,15 @@ public void configureContentNegotiation(ContentNegotiationConfigurer config) {
...
+
+  @Override
+  public void addResourceHandlers(ResourceHandlerRegistry registry) {
+    registry
+        .addResourceHandler("/static/**")
+        .addResourceLocations("/resources/", "/dhis/")
+        .setCachePeriod(3600)
+        .resourceChain(true)
+        .addResolver(new PathResourceResolver());
+    // , "classpath:/dhis-web-apps/"
```
jbee, line 72: remove?

jbee (approved):
Looks good, as we said in the meeting we might want to add their `LICENSE` file to the folders where we use copies of spring files.

david-mackessy (approved):
Just add some details about what testing has been done for this change so people know what has/has not been checked 👍

vietnguyen (approved):

PR #17512, opened 2024-05-22
fix: Validate enrollment in relationships [DHIS2-16460]

muilpp:


enricocolasante (approved):

teleivo (approved):
@enricocolasante we need to create a follow up issue to find a sustainable solution to our dependency graph that relationships introduce in our services 😅 

As I don't think we would want to expose methods like we do now in the `EnrollmentService` for everyone to use.

PR #17516, opened 2024-05-22
fix: /tracker/trackedEntites?order=enrolledAt returns wrong order

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-16446

Add to the join with the enrollment the program ID if it is present in the request, so we exclude from ordering the enrollments that are not required

### Next
There might be more cleanup to do, so I've created another issue https://dhis2.atlassian.net/browse/DHIS2-17447

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
@@ -38,7 +38,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.io.IOException;
-import java.time.Instant;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


muilpp (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java
```diff
@@ -356,6 +362,65 @@ void shouldOrderTrackedEntitiesByEnrolledAtAsc()
     assertEquals(List.of("QS6w44flWAf", "dUE514NMOlo"), trackedEntities);
   }
 
+  @Test
+  void shouldExcludeEnrollmentWhoseProgramIsNotRequestedAndOrderTrackedEntitiesByEnrolledAt()
+      throws ForbiddenException, BadRequestException, NotFoundException {
+
+    String oldEnrollment = "nxP7UnKhomJ";
```
teleivo, line 369: Can you please put the test setup into tracker/event_and_enrollment.json
This is now the only test that does it differently in this test class.

teleivo (approved):

PR #17522, opened 2024-05-22
fix: Use new start and end date types in tracker exporter [DHIS2-16019]

enricocolasante:
- Make all relevant date parameters use new `StartDateTime` and `EndDateTime` types
- Format dates in queries using `yyyy-MM-dd HH:mm:ss.SSS` format including milliseconds
- Modified a simple e2e test to prove that the `EndDateTime` default time is set to end of the day

**Next PR**
- Update OpenApi

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParams.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParams.java
@@ -28,7 +28,6 @@
 package org.hisp.dhis.webapi.controller.tracker.export.event;
 
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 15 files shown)


muilpp (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParams.java
```diff
@@ -133,27 +132,27 @@ public class EventRequestParams implements PageRequestParams {
...
+  private EndDateTime scheduledBefore;
 
   private StartDateTime updatedAfter;
 
   private EndDateTime updatedBefore;
 
   private String updatedWithin;
 
-  private Date enrollmentEnrolledBefore;
+  private EndDateTime enrollmentEnrolledBefore;
```
teleivo, line 149: I think it would make sense to group and order fields into Start then End. The type helps a lot as I was always wondering what is before/after mean again for the date range 😅

teleivo (approved):

PR #17548, opened 2024-05-23
fix: ValueType in metadata [DHIS2-15194]

maikelarabori:
Some metadata items were returned with ValueType NUMBER, even though they were of TEXT type.
This happened because the default ValueType, for some reason, has been NUMBER by default AND we did not have specific handling for `ProgramTrackedEntityAttributeDimensionItem`.

A new condition for `ProgramTrackedEntityAttributeDimensionItem` was added, so we get the respective ValuType from this object, where applicable. The default type will remain NUMBER, to not affect existing features. I extracted the logic into a smaller method to make it more readable and isolated.

A new e2e test to cover this scenario was added.


gnespolino (approved):
lgtm

d-bernat (approved):

PR #17552, opened 2024-05-23
feat: manually revert a running job to non-running [DHIS2-17419] (2.41)

jbee:
Cherry-pick backport of

* #17461
* #17551

The 2nd is dependent on changes in the 1st so I put this in one PR.

There were small merge conflicts in 2 files where code was different in 2.41 and I just kept the code as is in 2.41

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfigurationStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobConfigurationStore.java
@@ -154,7 +154,7 @@ public interface JobConfigurationStore extends GenericDimensionalObjectStore<Job
   JobConfiguration getNextInQueue(@Nonnull String queue, int fromPosition);
 
   /**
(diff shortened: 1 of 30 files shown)


zubaira (commented):
@jbee is this manual cancellation procedure available in UI or just API?

jbee:
@zubaira just API for now

jbee:
I didn't get what was wrong with this backport so I reattempted it entirely in #17592

jbee:
@zubaira just to be absolutely clear, this is not a cancellation feature. 

We have (and had for a while) `/cancel` which is a cooperative cancellation, meaning if the code does not "hang" in some slow query or alike it will eventually cancel but if it does this doesn't help. 

The feature added here is called `/revert` and should be used on a job that is supposed to be dead (not running at all) but which has a state of being considered running which puts it in a dead end of sorts as it cannot transition out of RUNNING when it is in fact not running. Usually we have an auto-detection for this that should transition it out of running but that might not work for some reason and this should give users a manual way to transition such a job back.

PR #17566, opened 2024-05-24
fix: Return enrollments from specified program only [DHIS2-14300]

muilpp:
Previously, when requesting a single TE in the context of a program, we returned all TE enrollments regardless of the specific program. With this PR, we will now return only the enrollments for the specified program. If no program is specified, all enrollments will be returned.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
@@ -236,7 +236,8 @@ class DefaultTrackedEntityService implements TrackedEntityService {
       UserDetails userDetails = getCurrentUserDetails();
 
       trackedEntity =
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -366,23 +368,21 @@ private Set<RelationshipItem> getRelationshipItems(
...
-        }
-        enrollment.setEvents(events);
-        enrollments.add(enrollment);
-      }
-    }
-    return enrollments;
+      TrackedEntity trackedEntity, UserDetails user, boolean includeDeleted, Program program) {
+    return trackedEntity.getEnrollments().stream()
+        .filter(e -> program == null || program.getUid().equals(e.getProgram().getUid()))
+        .filter(e -> trackerAccessManager.canRead(user, e, false).isEmpty())
```
enricocolasante, line 374: Maybe it would make sense to swap ACL and delete check as the former is much slower performance-wise

enricocolasante (approved):

lucaCambi77 (approved):

PR #17572, opened 2024-05-24
fix: Return TEA from specified program only [DHIS2-14300]

muilpp:
Previously, when requesting a single TE in the context of a program, we returned all TEA regardless of the specific program. With this PR, we will now return only the attributes of the specified program + the attributes of the tracked entity type. If no program is specified, only attributes of that particular TET will be returned.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
@@ -345,7 +345,8 @@ class DefaultTrackedEntityService implements TrackedEntityService {
     if (params.isIncludeProgramOwners()) {
       result.setProgramOwners(trackedEntity.getProgramOwners());
     }
-    result.setTrackedEntityAttributeValues(getTrackedEntityAttributeValues(trackedEntity));
+
+    result.setTrackedEntityAttributeValues(getTrackedEntityAttributeValues(trackedEntity, program));
 
     return result;
   }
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1955,6 +1970,33 @@ void shouldFailWhenRequestingSingleTEAndNoAccessToTET() {
         exception.getMessage());
   }
 
+  @Test
+  void shouldReturnProgramAttributesWhenSingleTERequestedAndProgramSpecified()
```
teleivo, line 1974:
```suggestion
  void shouldReturnTrackedEntityTypeAndProgramAttributesWhenSingleTERequestedAndProgramSpecified()
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -386,9 +389,14 @@ private Set<Enrollment> getEnrollments(
   }
 
   private Set<TrackedEntityAttributeValue> getTrackedEntityAttributeValues(
-      TrackedEntity trackedEntity) {
+      TrackedEntity trackedEntity, Program program) {
     Set<TrackedEntityAttribute> readableAttributes =
-        trackedEntityAttributeService.getAllUserReadableTrackedEntityAttributes();
```
teleivo, line 391: Is this method still used or can we remove it?

teleivo (approved):

PR #17590, opened 2024-05-27
chore: rename psi to event DHIS2-16563

teleivo:


dhis-2/dhis-test-e2e/src/main/java/org/hisp/dhis/actions/metadata/RelationshipTypeActions.java
```diff
@@ -66,11 +66,11 @@ public String createRelationshipType(
   private JsonObject createRelationshipConstraint(String type, String id) {
     JsonObjectBuilder builder = new JsonObjectBuilder().addProperty("relationshipEntity", type);
     switch (type) {
-      case "PROGRAM_STAGE_INSTANCE":
+      case "EVENT":
```
enricocolasante, line 69: I think this code is unused in e2e tests.
If not, we should change the call to this method with the new parameter values

enricocolasante (approved):

lucaCambi77 (approved):

PR #17594, opened 2024-05-27
chore: remove unused tracker data idSchemes DHIS2-16564

teleivo:
Old tracker endpoints `/trackedEntities`, `/enrollments`, `/events`, `/relationships` have been removed. Tracker data does not support idSchemes other than UID.


david-mackessy (approved):
nice one @teleivo 👍
They look Tracker specific, so if no tests are failing, it seems ok to me.

jbee (approved):

PR #17597, opened 2024-05-27
fix: Throw exception from stage in tracker importer [DHIS2-17448]

enricocolasante:


david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/RecordingJobProgress.java
```diff
@@ -199,7 +199,7 @@ public void startingProcess(String description, Object... args) {
     observer.run();
 
     if (isCancelled()) {
-      throw new CancellationException();
+      throw newCancellationException();
```
teleivo, line 202: when I first read it I thought this is a typo and invalid Java 😅 missing a space in between new and the exception. Maybe rename it to `cancellationException()` anyone else has the same first thought.

teleivo (commented):
The code that uses the TrackerImportService method


```
  ImportReport importTracker(
      TrackerImportParams params, TrackerObjects trackerObjects, JobProgress jobProgress);
```

seems susceptible to the same issue or would we never pass in a `noop` JobProgress?

jbee (approved):

PR #17598, opened 2024-05-27
fix: Filter TEA depending on whether program in context [DHIS2-14300]

muilpp:
Previously, we were loading all accessible program attributes into memory when requesting multiple TEs within a program's context. With this fix, we will now load only the accessible attributes that belong to the specific program mentioned in the request.

Regarding tracked entity type attributes, we were loading all attributes regardless of the requested TEs. Now, we load only the type attributes for the specified TEs. This change also addresses an issue related to disabling an attribute that already has a value in the database while it remains enabled for a different tracked entity type. If both conditions were met, we incorrectly displayed the attribute in the result, even though it was not enabled for the type of the current tracked entity.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
```diff
@@ -216,8 +209,9 @@ String validateAttributeUniquenessWithinScope(
   /**
    * Get all {@link TrackedEntityAttribute} grouped by {@link Program}
    *
+   * @param program if present, it will return {@link TrackedEntityAttribute} from this program only
    * @return a Map, where the key is the {@link Program} and the values is a Set of {@link
    *     TrackedEntityAttribute} associated to the {@link Program} in the key
    */
-  Map<Program, Set<TrackedEntityAttribute>> getTrackedEntityAttributesByProgram();
+  Map<Program, Set<TrackedEntityAttribute>> getTrackedEntityAttributesByProgram(Program program);
```
enricocolasante, line 216: Should this then just return a Set instead of a Map?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeStore.java
```diff
@@ -84,8 +84,9 @@ public interface TrackedEntityAttributeStore
   /**
    * Fetches all {@link TrackedEntityAttribute} and groups them by {@link Program}
    *
+   * @param program if present, it will return {@link TrackedEntityAttribute} from this program only
    * @return a Map, where the key is the {@link Program} and the values is a Set of {@link
    *     TrackedEntityAttribute} associated to the {@link Program} in the key
    */
-  Map<Program, Set<TrackedEntityAttribute>> getTrackedEntityAttributesByProgram();
+  Map<Program, Set<TrackedEntityAttribute>> getTrackedEntityAttributesByProgram(Program program);
```
enricocolasante, line 91: Same here

enricocolasante (commented):
I am confused.
We should always return a set of attributes, either program attributes + tet attributes or only tet attributes.
Map<Program, Set<TrackedEntityAttribute>> doesn't really makes sense to me

muilpp:
> I am confused. We should always return a set of attributes, either program attributes + tet attributes or only tet attributes. Map<Program, Set> doesn't really makes sense to me

What you describe is what we do, but the map you are referring to only holds program attributes, the tracked entity type attributes are added separately. This is all done [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/aggregates/TrackedEntityAggregate.java#L295)

We still need to have this Map because if the user requests the enrollments of the tracked entities, we'll return the program attributes of all the programs the user might be enrolled in.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
```diff
@@ -170,13 +170,6 @@ List<TrackedEntityAttribute> getTrackedEntityAttributesByDisplayOnVisitSchedule(
    */
   List<TrackedEntityAttribute> getTrackedEntityAttributesDisplayInListNoProgram();
 
-  /**
-   * Get all attributes that user is allowed to read (through program and tracked entity type)
-   *
-   * @return a list of attributes
-   */
-  Set<TrackedEntityAttribute> getAllUserReadableTrackedEntityAttributes();
```
enricocolasante, line 178: Was this one never used?

enricocolasante (commented):
Still a little bit confused.
I am right saying that this PR is not changing any behavior in the end but it is just optimizing to not load in memory all the attributes from all the programs when the client specify the program?

muilpp:
> Still a little bit confused. I am right saying that this PR is not changing any behavior in the end but it is just optimizing to not load in memory all the attributes from all the programs when the client specify the program?

yep, you are right, reading the PR description that's not clear, I'll update it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
muilpp: no, it's not used, now we get the attributes from the type of the tracked entity directly.

enricocolasante (changes requested):
I just noticed now after the fix on the TET attributes that we are getting the attributes and put them in a cache.
We cannot then filter them by TET or event Program because next call to export could have different TET and Program and will find in the cache only attributes not related to the current call.
We either need to put all the attributes in the caches or make the caches use the TET/Program as a key

muilpp:
> I just noticed now after the fix on the TET attributes that we are getting the attributes and put them in a cache. We cannot then filter them by TET or event Program because next call to export could have different TET and Program and will find in the cache only attributes not related to the current call. We either need to put all the attributes in the caches or make the caches use the TET/Program as a key

You are right, but I'm not sure it's worth it to create a new cache key that holds the attributes of each TET/Program pair. At least not before we know how many potential combinations we might have. I think it'd be better to filter the attributes in a different way. Since this is meant to fix a corner case, I'll close this PR and create a separate ticket to fix the issue.

PR #17599, opened 2024-05-27
ci: send all Jenkins build failure notifications to backend channel

radnov:
Currently only unit/integration tests failures from the Build stage will lead to notifications in the Backend channel, while other failures will be sent to the Jenkins channel, which mostly nobody follows. A couple of times we’ve ended up in a situation where builds of `master` or other dev branches keep failing and people are not aware.

Note that this might end up being a bit spammy, as failures occur for various reasons, sometimes outside of our control (like intermitent connectivity issues, etc).

### DIFF
--- a/jenkinsfiles/canary
+++ b/jenkinsfiles/canary
@@ -130,7 +130,11 @@ pipeline {
     post {
         failure {
             script {
-                slack.sendFailureMessage()
+                slack.sendMessage(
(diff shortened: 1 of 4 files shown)


teleivo (approved):
Thank you @radnov !

Philip-Larsen-Donnelly (approved):

PR #17602, opened 2024-05-28
fix: (2.41.0) Notifications from program rules in old tracker

zubaira:


enricocolasante (changes requested):
What is the Jira issue linked to this?
Anything going into 2.41.0 must be RCB approved

Philip-Larsen-Donnelly:
This is approved by RCB. It is a regression that was found testing another approved one (https://dhis2.atlassian.net/browse/DHIS2-17414) - ideally should have it's own issue; but I think it's OK just to get it in now.

enricocolasante (approved):

PR #17606, opened 2024-05-28
feat: use en as DB locale default [DHIS2-17408]

jbee:
Similar to UI locale the DB uses a locale default (EN)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
+++ b/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
@@ -59,7 +59,7 @@ import org.hisp.dhis.sms.config.SmsConfiguration;
  */
 public enum SettingKey {
   UI_LOCALE("keyUiLocale", LocaleManager.DEFAULT_LOCALE, Locale.class),
-  DB_LOCALE("keyDbLocale", Locale.class),
+  DB_LOCALE("keyDbLocale", LocaleManager.DEFAULT_LOCALE, Locale.class),
   ANALYSIS_DISPLAY_PROPERTY(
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):

jbee:
Keeps failing `PdfFormControllerTest.testGetDataSetPdfForm:62` now - so must be related

vietnguyen:
I think the test `testGetDataSetPdfForm` can be removed because it was added to support the case when DB_Locale is null.
Now with this fix it will never be null hence the test is redundant.

dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
```diff
@@ -59,7 +59,7 @@
  */
 public enum SettingKey {
   UI_LOCALE("keyUiLocale", LocaleManager.DEFAULT_LOCALE, Locale.class),
-  DB_LOCALE("keyDbLocale", Locale.class),
+  DB_LOCALE("keyDbLocale", LocaleManager.DEFAULT_LOCALE, Locale.class),
```
vietnguyen, line 62: This fix does not satisfy a conner case:
- An instance has DB_Locale = null. User enter value in Chinese.  The expected behavior is the raw value ( in Chinese ) should be returned. However with this fix, the DB_Locale is English so the return value will be the translated value in English, which is taken from `translations` jsonb column. 
- If the English translation value does not exist then the raw value will be returned.

Please have a look at [this code](https://github.com/dhis2/dhis2-core/blob/b2f0b6375c8c7ce21264c4300015e4bca0292939/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseIdentifiableObject.java#L336)

jbee:
Closing in favor of new PR that uses UI locale in the job as a fallback

PR #17608, opened 2024-05-28

david-mackessy:
# Issue
Issue originally found during QA testing of metadata sync job. The following exception was being thrown:
```
PersistentObjectException: detached entity passed to persist: org.hisp.dhis.organisationunit.OrganisationUnitGroupSetDimension 
```

All `BaseAnalyticalObject`s have the property:
```
protected List<OrganisationUnitGroupSetDimension> organisationUnitGroupSetDimensions;
```

# Cause
The root cause was eventually found after many dead ends. It was observed during debugging that `EventChart` objects were triggering the issue after a session flush. This seemed strange though, as the database being used did not have any `EventChart` data in it. It turned out that the central instance was saving a metadata snapshot which included `EventChart`s and `EventReport`s.

The reason for this was because there were 35 valid `EventVisualization`s in the database, and during the saving of the snapshot, those 35 `EventVisualization`s were also being used to instantiate 35 `EventReport`s and 35 `EventChart`s.

In the `ObjectBundleHook`, all metadata schemas are returned and used in the metadata export flow.

# Fix
A filter has been added to remove the 2 schemas (`EventChart` and `EventReport`) from the metadata export flow.

# Testing
## Automated
Unit tests added for:
- metadata export service getting metadata (ensuring deprecated classes are removed)
- predicate

## Manual
Multiple manual tests during the debugging of this issue. Steps to reproduce are in the Jira issue.
Manually confirmed this fix works locally & in IM. IM screenshot below of remote sync settings showing synced version from central instance

Screenshot 2024-05-28 at 16 51 27


david-mackessy:
@maikelarabori Can you advise that this change looks ok from an Analytics perspective? `EventChart` and `EventReport` will no longer be included in the default metadata export flow. Any possible adverse side effects?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataExportService.java
```diff
@@ -209,6 +211,15 @@ public Map<Class<? extends IdentifiableObject>, List<? extends IdentifiableObjec
     return metadata;
   }
 
+  /**
+   * This predicate is used to filter out deprecated Analytics schemas, {@link EventChart} & {@link
+   * EventReport}.As they are no longer used ({@link EventVisualization} has replaced them), they
+   * should be removed from the metadata export flow. This was causing issues otherwise. See <a
+   * href="https://dhis2.atlassian.net/browse/BETA-177">Jira issue</a>
+   */
+  public static final Predicate<Schema> DEPRECATED_ANALYTICS_SCHEMAS =
```
jbee, line 220: I assume it can be private?
I also would make this a constant for `Set<Class<?>>` instead and use that set in the filter. Seems less overly specific.

jbee (approved):
Nice find! We just spoke about those classes and turns out the did strike again :cry: There is a lessen in here somewhere :grimacing:

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/DefaultMetadataExportService.java
david-mackessy: It could be private. I wanted to add tests for it. I don't see any issue with it being public. It doesn't reveal anything unwanted about the class.
I created the predicate so tests could be added to cover the logic applied. This seems more valuable than using an inline set in the filter.

PR #17629, opened 2024-05-29
fix: Set skip audit to default true for aggregate data exchange [DHIS2-13105]

larshelge:


dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeService.java
```diff
@@ -270,8 +270,8 @@ ImportOptions toImportOptions(AggregateDataExchange exchange) {
     if (notNull(request.getImportStrategy())) {
       options.setImportStrategy(request.getImportStrategy());
     }
-    if (notNull(request.getSkipAudit())) {
-      options.setSkipAudit(request.getSkipAudit());
+    if (notNull(request.isSkipAuditOrDefault())) {
+      options.setSkipAudit(request.isSkipAuditOrDefault());
     }
     if (notNull(request.getDryRun())) {
```
larshelge, line 276: I know the if check above is not necessary but just makes the code more consistent and pretty.

jbee: hm... I think more that it is misleading because this will always set the property in the target object, having an `if` just gives the wrong idea about this behaviour.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataexchange/aggregate/TargetRequest.java
```diff
@@ -60,4 +62,10 @@ public class TargetRequest implements Serializable {
 
   /** Indicates whether to do a dry run. */
   @JsonProperty private Boolean dryRun;
+
+  /** Returns whether to skip audit, with fallback to default value if not set. */
+  @JsonIgnore
+  public boolean isSkipAuditOrDefault() {
+    return ObjectUtils.firstNonNull(skipAudit, true);
```
jbee, line 69: Can't we make the `skipAudit` field be a `boolean` initialised to `true` ? That should have the same effect just will less code.

dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeService.java
larshelge: Fine, changed it now.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataexchange/aggregate/TargetRequest.java
larshelge: I have bad experiences with primitive booleans in data models. Then you have no way of knowing whether the user supplied a value or not. I am also unsure how this works for existing Postgres JSONB blobs, i.e. if existing blobs have no `skipAudit` value, we may get a nullpointer here, not sure. Just wanted to keep it consistent with objects everywhere.

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataexchange/aggregate/TargetRequest.java
david-mackessy: I know what you're saying :)
For any new models, I think I'd advocate for either:
- primitive boolean with safe default (obviously this might not suit every use case) or
- use an enum instead of true/false (richer model)
I would always try to avoid having `Boolean`.

PR #17649, opened 2024-05-30
feat: OpenAPI ignore type and file description sources

jbee:
### Summary
After today's meeting I got an idea on how to avoid the need of knowing or using the placeholder `{md}` directly and instead have additional properties in the `@OpenApi.Description` annotation.

The key is that internally this still uses the placeholder but users don't have to use it to compose the text from multiple sources. Instead, by default, the text is always composed from all possible sources:

1. annotation on a specific element (e.g. a parameter)
2. annotation on the type of the element
3. a markdown file 

User then can opt-out of this automatic composition using two new properties in the `@OpenApi.Description` annotation:

* `ignoreTypeDescription`
* `ignoreFileDescription` 

As this internally is still based on the placeholder to combine annotation based and file based sources users can even continue to use the placeholder directly in their description text to place it differently than it is compose otherwise.

The composition of sources uses this sequence:

1. annotation on a specific element (e.g. a parameter)
2. annotation on the type of the element
3. a markdown file

This is so the most specific text comes first. For example, the text on the specific parameter is first, followed by the text on the type of the parameter, finally followed by the text from the markdown files which is assumed to be general as well when used in combination.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParams.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParams.java
@@ -116,7 +116,6 @@ public class EnrollmentRequestParams implements PageRequestParams {
 
   private Boolean followUp;
 
-  @OpenApi.Description("and this")
   private StartDateTime updatedAfter;
 
   private String updatedWithin;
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParams.java
```diff
@@ -116,6 +116,7 @@ public class EnrollmentRequestParams implements PageRequestParams {
 
   private Boolean followUp;
 
+  @OpenApi.Description("and this")
```
enricocolasante, line 119: That was probably just to test it

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentRequestParams.java
jbee: Oh yes :joy:

teleivo (approved):

PR #17651, opened 2024-05-30
feat: "New" calculation of outliers in the "old" API [2.42-DHIS2-17481]

d-bernat:
The "new" calculation of the modified z-score is based on the formula:
`Modified z-score = 0.6745(xi – x̃) / MAD`
The "old" calculation uses standard deviation, which is not robust enough for skewed data distributions.
For more details please see https://dhis2.atlassian.net/browse/DHIS2-16952?focusedCommentId=192941


gnespolino (approved):

dhis-2/dhis-services/dhis-service-validation/src/main/java/org/hisp/dhis/outlierdetection/processor/ZscoreSqlStatementProcessor.java
```diff
@@ -166,34 +196,151 @@ public String getSqlStatement(OutlierDetectionRequest request) {
...
-        ? "and pe.startdate >= :" + DATA_START_DATE.getKey() + " "
-        : StringUtils.EMPTY;
-  }
+    String thresholdParam = THRESHOLD.getKey();
 
-  private String getDataEndDateClause(Date dataStartDate) {
-    return dataStartDate != null
-        ? "and pe.enddate <= :" + DATA_END_DATE.getKey() + " "
-        : StringUtils.EMPTY;
+    return " with t2 as (select dv.dataelementid as dataelementid,"
```
maikelarabori, line 212: I think it could be formatted differently. I will ping you soon to chat about it.

dhis-2/dhis-services/dhis-service-validation/src/main/java/org/hisp/dhis/outlierdetection/processor/ZscoreSqlStatementProcessor.java
```diff
@@ -166,34 +196,151 @@ public String getSqlStatement(OutlierDetectionRequest request) {
...
+        + " dv.sourceid as sourceid,"
```
maikelarabori, line 213: Also, we could use text block here, and replace variables. But in the end, it's better like this to make backport easier, until we have all releases supporting it properly.

maikelarabori (approved):
Just a minor suggestion...
Cheers.

PR #17652, opened 2024-05-30
ci: add workflow for requesting PR reviews based on changed files

radnov:
UPDATE: superseded by https://github.com/dhis2/dhis2-core/pull/17693

This workflow requests a PR review via the [Auto Request Review action](https://github.com/marketplace/actions/auto-request-review). The current configuration only specifies adding the DevOps team as reviewers based on a certain set of files that can cause CI/build related failures, but can be exptended beyond that for more use cases.

### DIFF
--- a/.github/reviewers.yml
+++ b/.github/reviewers.yml
@@ -1,0 +1,19 @@
+# See https://github.com/marketplace/actions/auto-request-review for more configuration options.
+files:
+  '.github/**':
+    - team:devops
+  'dhis-2/*.sh':
+    - team:devops
+  'dhis-2/dhis-web-apps/apps-to-bundle.json':
+    - team:devops
+  'docker/**':
+    - team:devops
+  'jenkinsfiles/**':
+    - team:devops
+  'docker-compose.yml':
+    - team:devops
+  'jib.yaml':
+    - team:devops
+
+options:
+  ignore_draft: false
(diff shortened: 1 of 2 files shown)


teleivo (commented):
Do we have access to any of
* https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#about-auto-assignment
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
? to avoid using a 3rd party action.

I am not sure if core members still use forks 🤷🏻 Would this would work for forks? https://github.com/necojackarc/auto-request-review?tab=readme-ov-file#working-with-forks

tonsV2 (approved):

radnov:
> Do we have access to any of
> 
> * https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#about-auto-assignment
> * https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
>   ? to avoid using a 3rd party action.
> 
> I am not sure if core members still use forks 🤷🏻 Would this would work for forks? https://github.com/necojackarc/auto-request-review?tab=readme-ov-file#working-with-forks

@teleivo I don't think we need auto-assigment for now, but reading the CODEOWNERS docs, seems like it could be used to cover the current use case implemented in this PR. Will test it in a separate PR and proceed from there. 

I think some people still use forks, but [CODEOWNERS should work with forks](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-forks). 🤔

PR #17660, opened 2024-05-31
fix: TEXT and LONG_TEXT with single quote (') [DHIS2-17129] [2.41]

lucaCambi77:
https://dhis2.atlassian.net/browse/DHIS2-17129

In the old tracker, it is possible to update the data element value using the put request for _/events/{uid}/{dataElementUid}_ and send a payload with the new data values, for example:
```json
{
  "event": "ce2BE0tNVYJ",
  "dataValues": [
    {
      "dataElement": "aRz66hGQRnQ",
      "value": "no",
      "providedElsewhere": false
    }
  ]
}
```
The endpoint had a concurrent issue fixed in https://github.com/dhis2/dhis2-core/pull/15793. That PR replaces the whole event's record update with an SQL statement using a Postgres native function to update only the data elements in the `eventdatavalues` column, so we no longer have concurrency issues.
However, the SQL statement happens to use single quotes to delimit the data element's JSON replacement, and if the data element value contains a single quote, the query breaks.
With this fix, we escape the single quote and update a test with that case.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/EventImportTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/EventImportTest.java
@@ -258,7 +258,7 @@ class EventImportTest extends TransactionalIntegrationTest {
   }
 
   @Test
-  void shouldUpdateEventDataValues_whenAddingDataValuesToEvent() throws IOException {
+  void shouldUpdateEventDataValuesWhenAddingDataValuesToEvent() throws IOException {
     InputStream is =
         createEventJsonInputStream(
             programB.getUid(),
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/EventImportTest.java
```diff
@@ -279,12 +279,12 @@ void shouldUpdateEventDataValues_whenAddingDataValuesToEvent() throws IOExceptio
     // add a new data value and update an existing one
 
     DataValue dataValueA = new DataValue();
-    dataValueA.setValue("10'000");
+    dataValueA.setValue("10'''000'''");
```
teleivo, line 282: Do we assert anywhere that our replace logic does not alter what the user wanted to store?

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/deprecated/tracker/EventImportTest.java
lucaCambi77: yes there are assertions later referring to the getValue

teleivo (approved):

PR #17683, opened 2024-06-03
fix: Load PR with all PRA and filter later [DHIS2-17470]

enricocolasante:
We were retrieving ProgramRules and filtering the linked ProgramRuleActions based on the type needed for the server. This was causing trouble as Hibernate was updating ProgramRules and deleting some of the filtered ProgramRuleActions.
In this PR we are not filtering out ProgramRuleActions anymore, we are retrieving the needed ProgramRules with all the ProgramRuleActions associated and then mapping only the required actions.

Clean up:
- Remove all the unnecessary `@Transactional` in `dhis-service-program-rule` module
- Remove the unused `RuleActionKey`s
- Remove the mapping to actions not used in Tracker

### DIFF
--- a/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/NotificationRuleActionImplementer.java
+++ b/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/NotificationRuleActionImplementer.java
@@ -43,7 +43,6 @@ import org.hisp.dhis.program.notification.ProgramNotificationTemplateService;
 import org.hisp.dhis.rules.models.RuleAction;
 import org.hisp.dhis.rules.models.RuleEffect;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/NotificationRuleActionImplementer.java
```diff
@@ -70,7 +69,6 @@ protected ExternalNotificationLogEntry createLogEntry(String key, String templat
     return entry;
   }
 
-  @Transactional(readOnly = true)
```
muilpp, line 73: why do we remove the transactional annotation?

enricocolasante: Because it was useless here.
This method is just calling a service that is already annotated.

muilpp (approved):

PR #17685, opened 2024-06-03
feat: adds configuration for primary leader node [DHIS2-16878]

jbee:
### Summary
Adds a new configuration `node.primary_leader` that when set to true/on the node will unconditionally set itself as the leader (as opposed to only setting its ID when no other value is present). This means (eventually) the node should always take over the leader role in a cluster as all other nodes will only set their ID in case no ID is set. This way other nodes only take over the leader role if the primary leader is in fact not able to renew (maintain) its leader role by setting its ID.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
+++ b/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
@@ -75,6 +75,12 @@ public enum ConfigurationKey {
   /** Node identifier, optional, useful in clusters. */
   NODE_ID("node.id", "", false),
 
+  /**
+   * When true, the node will unconditionally set its node ID during leader election causing it to
+   * win the election as long as it is alive.
+   */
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -75,6 +75,12 @@ public enum ConfigurationKey {
   /** Node identifier, optional, useful in clusters. */
   NODE_ID("node.id", "", false),
 
+  /**
+   * When true, the node will unconditionally set its node ID during leader election causing it to
+   * win the election as long as it is alive.
```
jimgrace, line 80: I know this isn't (yet) end user documentation but I would suggest using the word "cluster" somewhere just so that people not concerned with clusters will skip over it faster.

jimgrace (approved):

vietnguyen (approved):

jbee:
@jimgrace Any news on the testing of this?

PR #17694, opened 2024-06-04
fix: push analytics job locale fallback to UI locale [DHIS2-17408]

jbee:
DB locale can be undefined, therefore we fallback to the UI locale with has a default so it should always be defined.
Giving the DB locale a default might have unintended side effects so this more local solution seems the better way to make sure that we always send a locale.


vietnguyen (approved):

david-mackessy (approved):
a test covering the new behaviour would be nice, especially if it's a fix

jbee:
> a test covering the new behaviour would be nice, especially if it's a fix

yeah, it is difficult to test unless I make it a test about just a helper method using mocks at which point I feel I exclude the part of the code I would be less sure about. At that point I opt to not invest time in such an automated test and rather ask Hendrik to re-test full circle.

PR #17701, opened 2024-06-04
feat: @Maturity annotation for endpoints

jbee:
### Summary
A new annotation that can be used in controllers to indicate the maturity of endpoints (methods) and/or individual endpoint parameters. 

This is a first step for a long term goal for better documentation and the ability for the dev team to classify the API so that it is clear what conditions apply for users of the APIs.

For now the annotation is picked up by OpenAPI and rendered into the document as `x-maturity`. Such `x-` properties can be added to OpenAPI freely for additional information. It does not have any effect unless a tool specifically supports such additional properties. 

Besides OpenAPI the maturity information might prove useful in other contexts.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
```diff
@@ -0,0 +1,68 @@
...
+     * The API has not matured yet. It might be changed or removed later on. 3rd parties should not
+     * rely on it and use this at their own risk.
+     */
+    EXPERIMENTAL,
+    /**
+     * 3rd parties should never depend on this, it is for internal use only. This means apps
+     * supported by the DHIS project use this, but it is subject to change whenever and however the
+     * app see fit.
+     */
+    INTERNAL
```
maikelarabori, line 64: I'm just wondering if INTERNAL should be part of this enum.
Maybe I'm overthinking, but we could also have something like:

`enum Exposure {
  INTERNAL,
  PUBLIC
}`

For example:
an API can be INTERNAL, and EXPERIMENTAL (or STABLE). It could also be PUBLIC and BETA.

Also, I'm unsure if INTERNAL is a state of maturity itself.

maikelarabori (commented):
Nice initiative @jbee 👍 
I just left some comments. Have a look if they make any sense.

Cheers

netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
```diff
@@ -0,0 +1,68 @@
...
+ * @author Jan Bernitt
+ */
+@Target({ElementType.METHOD, ElementType.TYPE, ElementType.PARAMETER, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Maturity {
+
+  enum Degree {
+    /**
+     * The API is mature, 3rd parties can depend on it and expect that breaking changes will occur
+     * rarely and only after a deprecation period.
```
david-mackessy, line 51: > and only after a deprecation period
I would think that an API can have breaking changes without it being marked as deprecated. Just a natural progression.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
david-mackessy: I'd agree with Maikel with regard to `INTERNAL` not being a maturity state. It's more about who should use the API.

david-mackessy:
Nice initiative Jan. I like the idea of bringing more clarity/intention to the APIs 👍
Do you think that the enum comments would be enough to circulate these changes with the wider community?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
jbee: Indeed. We might want to have more than one class of "stable/mature" or just weaken the guarantee we give.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
jbee: Assuming we keep things in 1 dimension of classification this is mostly a problem of the naming. If we name the enum and constant differently we might find names that are (more) coherent. That would be my goal.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
```diff
@@ -0,0 +1,68 @@
...
+ * <p>Can also be used to mark endpoint parameters with their maturity level. In parameter objects
+ * the corresponding fields are annotated.
+ *
+ * @author Jan Bernitt
+ */
+@Target({ElementType.METHOD, ElementType.TYPE, ElementType.PARAMETER, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Maturity {
+
+  enum Degree {
```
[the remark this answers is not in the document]
maikelarabori: I thought about BETA because of the documentation in the EXPERIMENTAL: _It might be changed or "removed" later on_

In BETA we would not have the possibility to "remove" the API, we are just highlighting this is a brand-new API and should be used carefully.

But if the goal is to be a bit vague as @netroms said, I believe that BETA could be covered as EXPERIMENTAL.

maikelarabori (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
maikelarabori: What if we call this enum `Category`, for example? This is more generic and would capture any of those intentions.

enricocolasante: It seems to me that it is indeed just a naming issue right now.
With 3 different statuses we should be covered and their meaning are:
- Safe to use
- Try it but do not get too comfortable (it could become safe to use )
- Not safe to use (it will never become safe to use )

I would suggest instead of `INTERNAL` something like `UNSTABLE` but I am not sure because it gives the idea that could change to `STABLE` at some point.

jbee: Yes, I think @enricocolasante found a good summary and way to put it. So we need better naming to reflect this and ideally make that read coherent. 

`Category` is a good idea, except the name clash with our metadata category makes it a bit confusing. Might still be better.

jbee: > We could consider assigning a score instead where 1 is the least mature and 3 is the most mature.

Good thinking outside the box I accidentally created, but a score would give the idea of linearity or a degree to which something is given. This does not match well with the 3 classifications we proposed thus far which are different and not by any means on some axis except that they are mutually exclusive.

stian-sandvold: I agree we should not use "INTERNAL", as one of our core ideas with the API is that it should all be public and consumable by others, and any app, script or interaction could be recreated by external people.

Switching it over to unstable sounds a lot better, but it's important that we use any non-stable maturity only in special cases, so we don't end up marking everything as unstable or experimental.

I would argue that experimental APIs are APIs where we dont have the complete implementation done (IE. we only implemented parts of the API, since we had to add partial support for something). Likewise, unstable APIs would be APIs that are new, and might not have any adoption in our core apps, and are subject to be changed before transitioning to stable.

amcgee:
This is a good initiative.  I think `EXPERIMENTAL`, `STABLE`, and `DEPRECATED` could be a simple, easy-to-understand, and useful set of initial maturity states.

```mermaid
flowchart LR
   EXPERIMENTAL --> removed
   EXPERIMENTAL --> STABLE
   STABLE --> DEPRECATED
   DEPRECATED --> removed
```

maikelarabori (approved):

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Maturity.java
```diff
@@ -0,0 +1,108 @@
...
+  ElementType.FIELD,
+  ElementType.ANNOTATION_TYPE
+})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Maturity {
+
+  enum Classification {
+    /**
+     * The API is stable and rarely subject to change. Change management occurs. Once an API is
+     * declared stable it does not change to another classification.
```
david-mackessy, line 57: This text may be a bit misleading to some. It might not technically move to another `Classification` value but it could still move to `Deprecated`. This needs to be clear, whether that's in the code or the external publication or both. No harm mentioning it here I think.

PR #17708, opened 2024-06-05
feat: Support DS/PR OUs param in aggregate [DHIS2-17485]

maikelarabori:
This PR enabled support for new syntax for the `ou` param, in `filter` and `dimension` (for the `/analytics` endpoint). Examples:
```
/api/analytics?dimension=ou:DS-lyLU2wR22tC
/api/analytics?dimension=ou:PR-lxAQ7Zs9VYR
/api/analytics?filter=ou:DS-lyLU2wR22tC
/api/analytics?filter=ou:PR-lxAQ7Zs9VYR
```
With the syntax above, we take into consideration all Org. Units associated with programs and data sets in `filter` and `dimension`, ie: `ou:DS-lyLU2wR22tC` and `ou:PR-lxAQ7Zs9VYR`

When the user doesn't have access to any of the Org. Units associated with the program or data set, an error message is raised through a Conflict/409 message (code E7120), as it already happens in the existing analytics flow.

The following request illustrates a full example:
```
http://localhost:8080/dhis/api/42/analytics?dimension=dx:FQ2o8UBlcrS,pe:LAST_5_YEARS,ou:PR-q04UBOqq3rp;DS-BfMAe6Itzgt&displayProperty=NAME&includeNumDen=true&skipMeta=true&skipData=false&relativePeriodDate=2022-10-01
```


gnespolino (approved):

vietnguyen (approved):
Would be great to have unit tests for new methods in the Store class to avoid hidden bug when object model is changed.

maikelarabori:
> Would be great to have unit tests for new methods in the Store class to avoid hidden bug when object model is changed.

Done 👍

PR #17709, opened 2024-06-05
chore: velocity and velocity-tools upgraded to latest version

gnespolino:
**This PR will upgrade Velocity to the latest version.**

### Reasons:

1. `commons-lang` is a forbidden dependency (we should use `commons-lang3` instead).
2. The currently used Velocity version (from 2010) still relies on `commons-lang`.
3. It would be safer to upgrade Velocity (and Velocity-Tools) to the latest version to ensure we have the most recent (and secure) version and to eliminate `commons-lang` from the classpath.

It would be helpful if someone from @dhis2/platform-backend could investigate any potential regressions.

Cheers

### DIFF
--- a/dhis-2/dhis-services/dhis-service-reporting/pom.xml
+++ b/dhis-2/dhis-services/dhis-service-reporting/pom.xml
@@ -117,7 +117,7 @@
     </dependency>
     <dependency>
       <groupId>org.apache.velocity</groupId>
-      <artifactId>velocity</artifactId>
+      <artifactId>velocity-engine-core</artifactId>
     </dependency>
     <dependency>
       <groupId>org.hibernate</groupId>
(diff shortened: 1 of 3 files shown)


enricocolasante (approved):

maikelarabori (approved):

david-mackessy:
I would suggest that this change be linked with a new Jira issue so it can go through the QA process. AFAIK it's used for template rendering. My guess is that QA would be most likely to catch any issues through the UI.

A [quick search](https://github.com/search?q=repo%3Adhis2%2Fdhis2-core+velocity&type=code) shows that it's used in the following areas:
- emails
- reports
- messages
- push analysis
- grid?

Do @stian-sandvold @jbee @vietnguyen @netroms have any more info or suggestions?

david-mackessy (approved):

vietnguyen (approved):

vietnguyen:
@gnespolino 
Approved as I did a quick test by create a new instance using this PR and the report are generated properly in the Report App.
However, it make sense to create a jira issue and forward to QA team for more testing.

PR #17721, opened 2024-06-06
chore: Use formatted console logs for easy reading

david-mackessy:
During the transition to use Tomcat embedded and some module restructuring, the console logging seems to have been impacted.

With this change, it will go from this:
Screenshot 2024-06-06 at 09 46 45

to this (which was the previous config):
Screenshot 2024-06-06 at 09 51 08


david-mackessy:
closing as ansi causes problems in production (breaks parsers, enlarges log files and escapes some text). Will look for alternative local dev solution.

PR #17723, opened 2024-06-06
fix: Add templateid column in program rule action [DHIS2-17515]

zubaira:
[DHIS2-17515](https://dhis2.atlassian.net/browse/DHIS2-17515)

**Description**
This pull request addresses an issue where `ProgramNotificationTemplate` objects could be deleted without checking for references in `ProgramRuleAction`. Such deletions led to `ProgramRuleAction` lacking an associated `ProgramNotificationTemplate`, causing potential issues in program rule execution.

**Changes Made**

 - Removed `templateUid` column and introduced `ProgramNotificationTemplate` reference.
 - During the transition period we will keep getter/setter for both.
 - Add flyway script to add new column and migrate existing records
 - Added e2e test for `ProgramRuleAction`
 - Changes are backward compatible

**Impact**
This fix ensures that all `ProgramRuleAction` retain their associated `ProgramNotificationTemplate`, maintaining data integrity and preventing potential issues with program rule actions. 

[DHIS2-17515]: https://dhis2.atlassian.net/browse/DHIS2-17515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/action/validation/NotificationProgramRuleActionValidator.java
```diff
@@ -64,26 +64,22 @@ public ProgramRuleActionValidationResult validate(
...
-              .getByUid(programRuleAction.getTemplateUid());
+      pnt = programRuleAction.getNotificationTemplate();
     }
 
     if (pnt == null) {
       log.debug(
           String.format(
               "ProgramNotificationTemplate id: %s for program rule: %s does not exist",
-              programRuleAction.getTemplateUid(), rule.getName()));
+              programRuleAction.getNotificationTemplate().getUid(), rule.getName()));
```
enricocolasante, line 74: This is a NullPointerException.
And no failing test pointing this issue means that we are not testing the case where the notification template is not present

enricocolasante (changes requested):

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/action/validation/NotificationProgramRuleActionValidator.java
zubaira: good catch @enricocolasante. After this change, i think we won't be needing this part of the code. 
I will add test as well to verify this change.

dhis-2/dhis-services/dhis-service-program-rule/src/test/java/ProgramRuleActionValidatorTest.java
```diff
@@ -120,4 +126,47 @@ void testHideOptionInvalid() {
...
+    NotificationProgramRuleActionValidator notificationValidator =
+        new NotificationProgramRuleActionValidator();
+    ProgramRuleActionValidationResult result =
+        notificationValidator.validate(programRuleAction, validationContext);
+    assertTrue(result.isValid());
+  }
+
+  @Test
+  void testInValidRuleActionSendMessage() {
+    ProgramRuleAction programRuleAction = new ProgramRuleAction();
```
enricocolasante, line 156: `ProgramRuleAction` shouldn't be of type `SENDMESSAGE`

dhis-2/dhis-services/dhis-service-program-rule/src/test/java/ProgramRuleActionValidatorTest.java
```diff
@@ -120,4 +126,47 @@ void testHideOptionInvalid() {
...
+    programRule.getProgramRuleActions().add(programRuleAction);
+
+    validationContext.setProgramRule(programRule);
+    validationContext.setNotificationTemplate(null);
+
+    NotificationProgramRuleActionValidator notificationValidator =
+        new NotificationProgramRuleActionValidator();
+    ProgramRuleActionValidationResult result =
+        notificationValidator.validate(programRuleAction, validationContext);
+    assertFalse(result.isValid());
```
enricocolasante, line 170: Could you also assert the ErrorCode to be sure you are getting the right error

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
```diff
@@ -134,7 +134,7 @@ public class ProgramRuleAction extends BaseIdentifiableObject implements Metadat
    *   <li>sendmessage
    * </ul>
    */
-  private String templateUid;
+  private ProgramNotificationTemplate notificationTemplate;
```
enricocolasante, line 137: Will this affect Maintenance App?
They were sending a uid before, how is it working now?

zubaira: yes and we do have a ticket for that. 
https://dhis2.atlassian.net/browse/DHIS2-17534

enricocolasante: This is a breaking change.
If we want to proceed this way, we need to coordinate with Platform team and we cannot just leave the app broken for the time being.
Why cannot we leave this field as a `templateUid` and change the underneath logic to validate it and store the actual `ProgramNotificationTemplate` in the DB?

enricocolasante (changes requested):
Api tests are failing because of the contract change?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
zubaira: this is exactly what I discussed previously and we agreed to have a foreign key instead of a validation. 
But foreign key is the right solution to go with. if its breaking change then we need to coordinate with the platform team.

enricocolasante: I still don't get it why we need to change the API.
That field could be annotated with @JsonIgnore, we convert the templateUid to a ProgramNotificationTemplate and everything works in the backend without changing the API

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
```diff
@@ -326,8 +338,25 @@ public String getTemplateUid() {
     return templateUid;
   }
 
-  public void setTemplateUid(String programNotificationTemplate) {
-    this.templateUid = programNotificationTemplate;
+  public void setTemplateUid(String templateUid) {
```
enricocolasante, line 341: What happens if both templateUid and programNotificationTemplate are defined and they have different values?
Who is the winner?

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/programrule/hibernate/ProgramRuleAction.hbm.xml
```diff
@@ -43,6 +43,9 @@
 
     <property name="templateUid" column="templateuid"/>
```
enricocolasante, line 44: Why do we need to keep templateUid column?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleActionObjectBundleHook.java
```diff
@@ -69,6 +70,21 @@ public void validate(
     }
   }
 
+  @Override
+  public void preCreate(ProgramRuleAction object, ObjectBundle bundle) {
```
enricocolasante, line 74: Why are we getting ProgramNotificationTemplate from the DB?
Should be already in ProgramRuleAction at this point?

enricocolasante (changes requested):
I still have doubts on the real need to change the API in order to fix the issue.

Where in the code are we validating that the uid for the template is valid and is present in the DB?

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramRuleActionObjectBundleHook.java
zubaira: No. It will be `ProgramNotificationTemplate` object with only uid.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
zubaira: `programNotificationTemplate` will be ignored if it's there.

zubaira: we are not changing API.

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/programrule/hibernate/ProgramRuleAction.hbm.xml
zubaira: its still in the ProgramRuleAction. It will be removed after 2.42.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
```diff
@@ -134,6 +137,18 @@ public class ProgramRuleAction extends BaseIdentifiableObject implements Metadat
...
+  /**
+   * The program notification template uid that will be triggered by the rule action. Used for:
+   *
+   * <ul>
+   *   <li>sendmessage
+   * </ul>
+   *
+   * @deprecated use {@link #notificationTemplate} instead
+   */
+  @Deprecated(since = "2.43")
```
enricocolasante, line 151: Why do we need to introduce `notificationTemplate` as an API field?
Would everything work fine now that we changed the DB to use the foreing key?
Can we have only `templateUid` as part of the API as before and make `notificationTemplate` @JsonIgnore?

enricocolasante (commented):
Can you add a good description to the PR so anyone that want to review knows what all this is about?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
zubaira: This is exactly what's happening

zubaira: > Why do we need to introduce `notificationTemplate` as an API field?
> 
> Would everything work fine now that we changed the DB to use the foreing key?
> 
> Can we have only `templateUid` as part of the API as before and make `notificationTemplate` @JsonIgnore?

Can we make it a foreign key without having it as field ?

enricocolasante: You are mixing 2 different approaches.
1. We can annotate `notificationTemplate` with @JsonIgnore, hence not using it in the API and use it in the DB and remove `templateuid` column. This approach do not involve any deprecation. The API remains the same and it will not be changed in the future.
2. For a transition period we accept either fields in the API `templateUid` and `notificationTemplate`, validating that if both are sent we are sending an error back, and specifing on `templateUid` when it is going to be removed with @Deprecated(forRemoval = true, since = "2.43"). The DB part is the same as the other, `templateuid` column is removed.

enricocolasante: We need to decide which one to use and follow just one approach

enricocolasante (approved):
The API and the DB part look good to me.
I am adding the platform team as a reviewer to check the metadata importer part

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
```diff
@@ -134,6 +137,15 @@ public class ProgramRuleAction extends BaseIdentifiableObject implements Metadat
...
+  @JsonIgnore private ProgramNotificationTemplate notificationTemplate;
+
+  /**
+   * The program notification template uid that will be triggered by the rule action. Used for:
+   *
+   * <ul>
+   *   <li>sendmessage
+   * </ul>
+   */
   private String templateUid;
```
david-mackessy, line 149: Could you use `@JsonSetter` & `@JsonGetter` on the relevant methods and get rid of this property completely? It doesn't look like it serves a purpose & muddies the Object (2 properties that do the same thing but this isn't used anywhere that I can see, unless I'm missing something).
I see the `templateuid` column is being removed from the DB also which is good, so it would make sense to remove it here also I think.
Jackson should be flexible enough to keep the same API, while using the new `notificationTemplate` property.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleAction.java
```diff
@@ -322,12 +334,29 @@ public void setDataElement(DataElement dataElement) {
...
+  }
+
+  public void setTemplateUid(String templateUid) {
+    if (templateUid == null) {
+      notificationTemplate = null;
+      return;
+    }
+
+    this.templateUid = templateUid;
+    ProgramNotificationTemplate template = new ProgramNotificationTemplate();
```
david-mackessy, line 349: Is there a possibility that creating a new `ProgramNotificationTemplate` here and only setting its `UID` could override an existing `ProgramNotificationTemplate` with the same `UID`, which already had an existing valid `messageTemplate` for example? (when objects are saved to the DB)
So `messageTemplate` could have been something before & now it has been overridden with `null`?
A test would confirm whether this is a potential issue or not (if that condition is not already checked)

david-mackessy: that test would also confirm whether the API is working as expected after the changes introduced, specifically the `ProgramRuleAction` `get`/`set` `templateUid` (can't see any API tests in the PR, unless existing tests cover the change)

david-mackessy (approved):

PR #17746, opened 2024-06-10
fix: broken authority listing in the user admin

netroms:
## Summary
After removal of Struts in 2.42, the app and schema authority listing system was also removed. This PR replaces the Struts pre-removal listing system with a new much simpler one, that directly lists the app authorities from the app manager and the schema service.

### Automatic tests
AuthoritiesControllerTest#testGetAllAuthorities() 

### Manual test
1. Navigate to the user admin app
2. Validate that all metadata authorities is listed
3. Validate that all installed app authorities is listed
4. Validate that all bundled app authorities is listed
5. Validate that all system authorities is listed

### Jira
[DHIS2-17441](https://dhis2.atlassian.net/browse/DHIS2-17441)

[DHIS2-17441]: https://dhis2.atlassian.net/browse/DHIS2-17441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/AuthoritiesController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/AuthoritiesController.java
@@ -30,16 +30,23 @@ package org.hisp.dhis.webapi.controller.security;
 import static java.util.Collections.singletonMap;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.servlet.http.HttpServletResponse;
+import org.apache.commons.lang3.StringUtils;
+import org.hisp.dhis.appmanager.AndroidSettingsApp;
 import org.hisp.dhis.appmanager.App;
+import org.hisp.dhis.appmanager.AppManager;
 import org.hisp.dhis.common.DhisApiVersion;
 import org.hisp.dhis.common.OpenApi;
 import org.hisp.dhis.i18n.I18n;
 import org.hisp.dhis.i18n.I18nManager;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/AuthoritiesController.java
```diff
@@ -79,14 +101,35 @@ public Map<String, List<Map<String, String>>> getAuthorities(HttpServletResponse
...
+            });
+    authorities.add(AndroidSettingsApp.AUTHORITY);
+    return authorities;
+  }
+
+  private List<String> getBundledAppsAuthorities() {
+    List<String> authorities = new ArrayList<>();
+    Set<String> bundledApps = AppManager.BUNDLED_APPS;
+    for (String app : bundledApps) {
+      String key = "M_dhis-web-" + app;
```
jbee, line 121: I wonder if this logic should be the same or based on the same code as the one in `App#getSeeAppAuthority`

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/AuthoritiesController.java
david-mackessy: Yep, best to get the authority from the App itself (if possible) rather than create it outside of the App.

david-mackessy (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/AuthoritiesController.java
netroms: I think we should possibly refactor the AppManager code first. I'm not sure that we want to continue using a hard-coded static list of bundled apps.

PR #17823, opened 2024-06-18
fix: throws exception for invalid dates [DHIS2-17539]

gnespolino:
Before this PR, a request containing an invalid date could pass through and get processed anyway.
This PR changes this behavior, returning an error when an invalid date is passed.


d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DimensionalObjectProducer.java
```diff
@@ -258,7 +259,9 @@ private void addDailyPeriods(
       dimensionalKeywords.addKeyword(
           isoPeriodHolder.getIsoPeriod(), join(" - ", startDate, endDate));
       periods.add(periodToAdd);
+      return;
     }
+    throw new IllegalQueryException("Invalid period: " + isoPeriodHolder.getIsoPeriod());
```
maikelarabori, line 264: We should always prioritize throwing and exception with an ErrorCode, ie:
```
throw new IllegalQueryException(new ErrorMessage(ErrorCode.E7611, isoPeriodHolder.getIsoPeriod()));
```

maikelarabori (commented):
Thanks @gnespolino, just a minor improvement required. 👍

maikelarabori (approved):

PR #17835, opened 2024-06-19
fix: Skip mandatory data value validation for UPDATE [DHIS2-17560]

enricocolasante:
Event update was failing when mandatory data values were not present in the payload even though they were already present in the DB.
Now tracker importer is validating mandatory data values only when creating an event.
Same logic is applied to `SetMandatoryField` rule action that is dynamically adding a mandatory check on a data value.

~~To resume, when strategy is create:~~
~~- If a mandatory data element is not provided, a validation error is returned~~
~~- If a mandatory data element is removed (set to null), a validation error is return. This is covered even if it doesn't really make sense to remove a data value when creating an event.~~

~~When strategy is update:~~
~~- If a mandatory data element is not provided, validation passes~~
~~- If a mandatory data element is removed (set to null), a validation error is return.~~

Statuses that allow to define data values are `ACTIVE`, `COMPLETED` or `VISITED`.
Statuses that DO NOT allow to define data values are `SCHEDULE`, `SKIPPED `or `OVERDUE`.

To resume, when strategy is create:
- If a mandatory data element is not provided and the status allows data values, a validation error is returned
- If a mandatory data element is removed (set to null) and the status allows data value, a validation error is returned.

When strategy is update:
- If a mandatory data element is not provided and the status is being updated between a status that do not allow data values to one that allows them, a validation error is returned.
- If a mandatory data element is not provided and the status is being updated between a status that allow data values to another one that allows them (also the same exact status), validation passes.
- If a mandatory data element is removed (set to null) and the status allows data value, a validation error is return.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -99,24 +100,31 @@ public static List<Note> validateNotes(
...
   }
 
-  public static List<MetadataIdentifier> validateMandatoryDataValue(
+  public static List<MetadataIdentifier> validateDeletionMandatoryDataValue(
       ProgramStage programStage, Event event, List<MetadataIdentifier> mandatoryDataElements) {
-    List<MetadataIdentifier> notPresentMandatoryDataElements = Lists.newArrayList();
-
     if (!needsToValidateDataValues(event, programStage)) {
-      return notPresentMandatoryDataElements;
+      return List.of();
```
muilpp, line 106: out of the scope of this PR, but why do we skip the validation if the event status is `SCHEDULE`?

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
enricocolasante: A `SCHEDULE` event is created at some point in time but it is meant to happen later at some point, so when it is created it is kind of `empty` and it doesn't make sense to require mandatory data values at that point.

muilpp: and in that case, are covered if I create and event this way, and then I provide the values later on?
what I mean by that is, at the moment I provide the values, do we treat it as an import strategy `CREATE` again to make sure all mandatory fields are set?

enricocolasante: Oh no, you are right.
At the moment if you create a schedule event without the mandatory datavalues and later you just change the status of the event then you can have an empty mandatory datavalues were they are meant to be present.

Will need to check how we manage the transition from a status to another

lucaCambi77 (approved):

muilpp (changes requested):
waiting for the scheduled events implementation

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -99,24 +100,48 @@ public static List<Note> validateNotes(
...
+      return false;
+    }
+
+    if (bundle.getStrategy(event).isCreate()) {
+      return true;
     }
 
-    return notPresentMandatoryDataElements;
+    EventStatus savedStatus = bundle.getPreheat().getEvent(event.getUid()).getStatus();
+    return EventStatus.STATUSES_WITHOUT_DATA_VALUES.contains(savedStatus)
```
muilpp, line 143: do the event statuses `OVERDUE` and `SKIPPED` also need to be considered here? or `SCHEDULE` would be enough?

enricocolasante: Yes they do. They have the exact same behaviour as `SCHEDULE` events, so we need to check them to be sure we are in a consistent state

muilpp (approved):

PR #17840, opened 2024-06-19
fix: Check owner when getting TE with superuser [DHIS2-16981]

muilpp:
Transferred enrollments were not showing up at the correct org unit when using a superuser, the cause was we were always using the registering org unit of a TE to validate whether that particular user could access the enrollment or not. This meant an enrollment would show up only if the registering org unit was selected, regardless of whether that enrollment was transferred or not.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-16981


enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerOwnershipManagerTest.java
```diff
@@ -426,6 +443,21 @@ void shouldFindTrackedEntityWhenTransferredToAccessibleOrgUnit()
...
+  void shouldFindTrackedEntityWhenTransferredToAccessibleOrgUnitAndSuperUser()
+      throws ForbiddenException, BadRequestException, NotFoundException {
+    transferOwnership(trackedEntityA1, programA, organisationUnitB);
+    superUser.setOrganisationUnits(Set.of(organisationUnitB));
+    userService.updateUser(superUser);
+    TrackedEntityOperationParams operationParams = createOperationParams(superUser, null);
+
+    List<TrackedEntity> trackedEntities = trackedEntityService.getTrackedEntities(operationParams);
+
+    assertContainsOnly(
```
teleivo, line 456: We should adopt some of the practices from other tests like https://github.com/dhis2/dhis2-core/blob/8cf7036ab8ef79b8830071d289d77eb271de9355/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/OrderAndPaginationExporterTest.java#L354-L357 to make them as readable.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -594,13 +594,9 @@ private String getFromSubQueryJoinOrderByAttributes(TrackedEntityQueryParams par
    * Generates an INNER JOIN for program owner. This segment is only included if program is
    * specified or user is not super.
```
teleivo, line 595: > This segment is only included if program is
   * specified or user is not super.

Is this sentence still true?

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
muilpp: nope, that part is not true anymore

PR #17870, opened 2024-06-24
fix: Deprecate old change log endpoints [DHIS2-16741]

muilpp:
Following up on the work done in [change logs](https://github.com/dhis2/dhis2-core/pull/16628), we need to deprecate the old endpoints in favor of the new ones.


teleivo (approved):
Can you please already add this to https://github.com/dhis2/dhis2-releases/blob/master/releases/2.42/ReleaseNote-2.42.md#tracker. Best we do this as we make changes so we are ready whenever we release is 😋

enricocolasante (approved):

PR #17871, opened 2024-06-24
fix: Filter pra types to get DE and TEA [DHIS2-17243]

enricocolasante:
I am asking @dhis2/platform-backend review because I am changing a store method and using `nativeSynchronizedQuery`

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleStore.java
@@ -44,12 +44,14 @@ public interface ProgramRuleStore extends IdentifiableObjectStore<ProgramRule> {
    */
   List<ProgramRule> get(Program program);
 
-  List<String> getDataElementsPresentInProgramRules();
-
-  List<String> getTrackedEntityAttributesPresentInProgramRules();
-
-  List<ProgramRule> getProgramRulesByActionTypes(Program program, Set<ProgramRuleActionType> types);
+  List<ProgramRule> getProgramRulesByActionTypes(
+      Program program, Set<ProgramRuleActionType> actionTypes);
 
   List<ProgramRule> getProgramRulesByActionTypes(
-      Program program, Set<ProgramRuleActionType> serverSupportedTypes, String programStageUid);
+      Program program, Set<ProgramRuleActionType> actionTypes, String programStageUid);
+
+  List<String> getDataElementsPresentInProgramRules(Set<ProgramRuleActionType> actionTypes);
+
+  List<String> getTrackedEntityAttributesPresentInProgramRules(
+      Set<ProgramRuleActionType> actionTypes);
 }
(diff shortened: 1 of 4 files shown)


david-mackessy:
Hi @enricocolasante, one of the concerns about using the native query, and it's mentioned in [the javadoc](https://github.com/dhis2/dhis2-core/blob/188ee0a3d0a8ff509caed402e338a7c25bbcbd44/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/hibernate/HibernateGenericStore.java#L366), is that it should only target the same table of the store, in this case `programruleaction`.
Because these 2 updated queries use joins (`dataelement` & `trakedentityattribute`), my understanding is that this will cause Hibernate to flush all `DataElement` or `TrackedEntityAttribute` from the cache. This could potentially have negative impact on the query.
If there was a performance test for this endpoint it would help guide a decision.

enricocolasante:
> Hi @enricocolasante, one of the concerns about using the native query, and it's mentioned in [the javadoc](https://github.com/dhis2/dhis2-core/blob/188ee0a3d0a8ff509caed402e338a7c25bbcbd44/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/hibernate/HibernateGenericStore.java#L366), is that it should only target the same table of the store, in this case `programrule`. Because these 2 updated queries use joins (`dataelement` & `trakedentityattribute`), and also the main select is to `programruleaction`, my understanding is that this will cause Hibernate to flush all `DataElement`, `TrackedEntityAttribute` & `ProgramRuleAction` from the cache before executing the query. This could potentially have negative impact on the query. If there was a performance test for this endpoint it would help guide a decision.

Changed to use `nativeQuery`

david-mackessy:
Also, any fix should usually be accompanied with a test, proving the absence of the bug being fixed.

jbee:
FYI: If you look at the implementation of `nativeSynchronizedQuery` it might become more clear what the "synchronized" part entails. It adds the table of the store as a table to "synchronize" with, meaning "make sure all changes are flushed and caches cleared". If you use a join the proper usage for a native query is to add the joined tables to the query as well as those tables also need to be "in sync" to give you the results you want. At least most likely this is what you want. 

As David pointed out, if you use HQL that part is done for you. But there are other reasons to prefer SQL over HQL so I also think it is wrong to advocate for HQL purely because hibernate is asking you to do this manually. One could even get the idea they did that on purpose to make users dependent. Because they could have easily offered an option to run the SQL on the same basis as the HQL. It is not that this would be hard to figure out based on the hibernate mappings.

enricocolasante:
Thanks for the comments, it was a quite confusing PR as it is extracted from a bigger one just to try to make it smaller but it was very clarifying for me. I thought I was confused about the different query option that we have but I found out that I was more than expected.
I just went with @david-mackessy suggestion on this one and I am using hql and everything is working as expected

david-mackessy (approved):
thanks for adding the tests :)

PR #17882, opened 2024-06-25
chore: Sonar - remove unused code

david-mackessy:
# Sonar issue
While investigating some sonar issues, while looking deeper into the git history, it looks like the removed code in this PR has been replaced with `existsByUser` in the same class. It was part of [this feature](https://dhis2.atlassian.net/browse/DHIS2-13470) but it looks like it may have been left behind by mistake when a change was required by the feature.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/DefaultIdentifiableObjectManager.java
```diff
@@ -1253,6 +1253,9 @@ private <T extends IdentifiableObject> List<T> findByUser(
         schema.getPersistedProperty(BaseIdentifiableObject_.LAST_UPDATED_BY) != null;
 
     UserDetails currentUserDetails = UserDetails.fromUser(user);
+    if (currentUserDetails == null)
+      throw new IllegalArgumentException(
```
jbee, line 1257: Hm.. maybe we should throw an exception linked to authority that also causes a redirect to the login? @netroms any thoughts?

david-mackessy: well, the `user` param in this method doesn't necessarily have to be the logged in `user`. Not sure if that would be the correct thing to do here. It looks like this method can take any `user` (doesn't have to be logged in) and finds objects linked them.

jbee: AFAIK any user in DHIS2 is a "logged in" user. There are only a few special APIs one may use as a "unidentified" user, like the login page. Lets see what MS has to say :)

david-mackessy: had a chat with @netroms, he's happy to just delete the method as well as it's not used. So I'll update the PR to just remove this method completely.

jbee (approved):

PR #17890, opened 2024-06-26
fix: Add event validation when updating status [DHIS2-17658]

enricocolasante:
This is the first PR to fix https://dhis2.atlassian.net/browse/DHIS2-17658.
We were missing the validation on event status update.
Some statuses are meant to be used for events that will be handled at a later point in time ('SCHEDULE', 'SKIPPED', 'OVERDUE') and they do not allow data values to be defined and other statuses are meant to be used for events that are being worked on ('ACTIVE', 'COMPLETED', 'VISITED').
Updating from the first ones to the second ones is always allowed while going back from the second ones to the first ones does not make sense and it has to be blocked.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
@@ -53,7 +53,9 @@ public enum EventStatus {
     return value;
   }
 
-  public static Set<EventStatus> ALLOW_DATA_VALUES_STATUSES = Set.of(ACTIVE, VISITED, COMPLETED);
+  public static final Set<EventStatus> ALLOW_DATA_VALUES_STATUSES =
+      Set.of(ACTIVE, VISITED, COMPLETED);
 
-  public static Set<EventStatus> NO_ALLOW_DATA_VALUES_STATUSES = Set.of(SCHEDULE, SKIPPED, OVERDUE);
+  public static final Set<EventStatus> NO_ALLOW_DATA_VALUES_STATUSES =
+      Set.of(SCHEDULE, SKIPPED, OVERDUE);
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
@@ -53,9 +53,9 @@ public enum EventStatus {
     return value;
   }
 
-  public static final Set<EventStatus> ALLOW_DATA_VALUES_STATUSES =
+  public static final Set<EventStatus> STATUSES_WITH_DATA_VALUES =
       Set.of(ACTIVE, VISITED, COMPLETED);
 
-  public static final Set<EventStatus> NO_ALLOW_DATA_VALUES_STATUSES =
+  public static final Set<EventStatus> STATUSES_WITHOUT_DATA_VALUES =
       Set.of(SCHEDULE, SKIPPED, OVERDUE);
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
@@ -157,6 +157,7 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
+  E1316("Event `{0}` cannot transition from status `{1}` to status `{2}.`"),
 
   /* Relationship */
   E4000("Relationship: `{0}` cannot link to itself"),
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
@@ -157,7 +157,7 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
-  E1316("Event `{0}` cannot transition from status `{1}` to status `{2}.`"),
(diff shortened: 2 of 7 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
```diff
@@ -52,17 +53,9 @@ public int getValue() {
     return value;
   }
 
-  public static EventStatus fromInt(int status) {
-    for (EventStatus eventStatus : EventStatus.values()) {
-      if (eventStatus.getValue() == status) {
-        return eventStatus;
-      }
-    }
+  public static final Set<EventStatus> ALLOW_DATA_VALUES_STATUSES =
```
muilpp, line 56: wouldn't it make more sense to name it `ALLOWED_STATUSES`?
we are not referencing event data values here, are we?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
```diff
@@ -52,17 +53,9 @@ public int getValue() {
...
+      Set.of(ACTIVE, VISITED, COMPLETED);
 
-    throw new IllegalArgumentException();
-  }
-
-  public static boolean isExistingEvent(EventStatus status) {
-    return status != null && (COMPLETED.equals(status) || VISITED.equals(status));
-  }
+  public static final Set<EventStatus> NO_ALLOW_DATA_VALUES_STATUSES =
```
muilpp, line 59: I'd call this one `DISALLOWED_DATA_VALUES_STATUSES`, and in case the comment above makes sense, just `DISALLOWED_STATUSES`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
muilpp: Anyway, is this used somewhere?

enricocolasante: `ALLOWED_STATUSES` is not true. These are the statuses that allow to have data values defined.
`ALLOWED_STATUSES` gets you the idea that those statuses are not allowed at all, which is not true.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
enricocolasante: I think not allowed and disallowed are not synonyms, but we could ask a native speaker about it.
The idea here is the same as above, these statuses are valid, the just do not permit to define data values.
Maybe because this was extracted from a bigger PR is not clear, but we will add another validation that is simply going to fail if the event has a status in `NO_ALLOW_DATA_VALUES_STATUSES` and `dataValues` is not empty

dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
teleivo: How about `STATUSES_ALLOWING_DATA_VALUES` or `STATUSES_WITH_DATA_VALUES`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/event/EventStatus.java
teleivo: `STATUSES_DISALLOWING_DATA_VALUES` or `STATUSES_WITHOUT_DATA_VALUES`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
```diff
@@ -157,6 +157,7 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
+  E1316("Event `{0}` cannot transition from status `{1}` to status `{2}.`"),
```
teleivo, line 160: Why is there a gap in error codes?

Should we phrase it as `No event can transition...` as its not specific to that event. The uid of the event is returned in the object containing the message right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataStatusValidator.java
```diff
@@ -0,0 +1,62 @@
...
+    org.hisp.dhis.program.Event savedEvent = bundle.getPreheat().getEvent(event.getUid());
+
+    if (checkInvalidStatusTransition(savedEvent.getStatus(), event.getStatus())) {
+      reporter.addError(event, E1316, savedEvent.getStatus(), event.getStatus());
+    }
+  }
+
+  private boolean checkInvalidStatusTransition(EventStatus fromStatus, EventStatus toStatus) {
+    return switch (fromStatus) {
+      case VISITED, ACTIVE, COMPLETED ->
```
teleivo, line 52: These are `ALLOW_DATA_VALUES_STATUSES`, should we not use that then? like you do with `NO_ALLOW_DATA_VALUES_STATUSES`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataStatusValidator.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.hisp.dhis.tracker.imports.validation.ValidationCode.E1316;
+
+import org.hisp.dhis.event.EventStatus;
+import org.hisp.dhis.tracker.imports.TrackerImportStrategy;
+import org.hisp.dhis.tracker.imports.bundle.TrackerBundle;
+import org.hisp.dhis.tracker.imports.domain.Event;
+import org.hisp.dhis.tracker.imports.validation.Reporter;
+import org.hisp.dhis.tracker.imports.validation.Validator;
+
+class DataStatusValidator implements Validator<Event> {
```
teleivo, line 40: `StatusUpdateValidator` is maybe more accurate?

muilpp (approved):

JoakimSM (approved):

PR #17891, opened 2024-06-26
feat(DHIS2-17506): enable configuring a section display options

flaminic:


david-mackessy:
Hi @flaminic , could a description be added to this PR, giving some info around why the change is needed, the approach being taken and any future plans regarding display option properties? The [PR guide](https://github.com/dhis2/wow-backend/blob/master/guides/git_pull_requests.md) is handy to be aware of.
Is the new field being returned as expected?
Ideally there would be a test included for the new behaviour, I can help with that if needs be.

flaminic:
> Hi @flaminic , could a description be added to this PR, giving some info around why the change is needed, the approach being taken and any future plans regarding display option properties? The [PR guide](https://github.com/dhis2/wow-backend/blob/master/guides/git_pull_requests.md) is handy to be aware of. Is the new field being returned as expected? Ideally there would be a test included for the new behaviour, I can help with that if needs be.

Hey, I hope now it's better :) 
I think so, tested it locally using postman and both post and get work well. But I will keep the PR in draft for a bit longer, until I fully tested it with the new UI. 
I would love some help with the tests!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/DataSet.java
```diff
@@ -778,6 +780,16 @@ public void setStyle(ObjectStyle style) {
     this.style = style;
   }
 
+  @JsonProperty
+  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
+  public String getDisplayOptions() {
```
jbee, line 785: I think it is correct to type this `String` as the backend has no interest in "understanding" the JSON so there is no use in putting in effort to parse and render the JSON content. However, documentation and developers will benefit from pointing out what this is for OpenAPI. We can do this by annotating it `@OpenApi.Property(ObjectNode.class)` (jackson's `ObjectNode`).

Could you also please add that to the same getter in `Section` in this PR? Thanks!

jbee (approved):
Looks good.

As David said we usually try to add at least a test for something surfacing in API but since this is purely for the UI I don't see a need to test this as part of the backend as it should be tested as part of the app using it. 

I might have already said it in the PR for `Section` that I don't think it is such a good idea adding UI concerns to the backend model. Especially since you want JSON there is the datastore which is a perfect fit to use for settings like these. But that boat has sailed, I guess. I vaguely remember a conversation and that the solution should only be "temporary"? Or it is just me hoping it is :joy: 

I guess what we have no good support for yet is datastore values that are "auto-cleaned" on deletion of a metadata object. We maybe should build that feature so there are no excuses to not use the datastore :)

flaminic:
> Looks good.
> 
> As David said we usually try to add at least a test for something surfacing in API but since this is purely for the UI I don't see a need to test this as part of the backend as it should be tested as part of the app using it.
> 
> I might have already said it in the PR for `Section` that I don't think it is such a good idea adding UI concerns to the backend model. Especially since you want JSON there is the datastore which is a perfect fit to use for settings like these. But that boat has sailed, I guess. I vaguely remember a conversation and that the solution should only be "temporary"? Or it is just me hoping it is 😂
> 
> I guess what we have no good support for yet is datastore values that are "auto-cleaned" on deletion of a metadata object. We maybe should build that feature so there are no excuses to not use the datastore :)

Awesome, thanks Jan. 
eheh i see what you mean re the datastore. In this case, i prioritized coherence with the section work. But we did mention with @kabaros that the work on the new maintenance app will give us an opportunity to clean a few of these rendering fields. One idea was to group them all in display options, but also, i think this is a good opportunity to evaluate the different options :)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dataset/DataSet.java
flaminic: Sure thing

david-mackessy (approved):

jbee:
@flaminic we discussed the datastore feature I suggested today in the platform backend meeting and one reason to keep this information in the metadata objects themselves is that often people want this to transfer with the matadata object. So if you were to export a dataset you want the system that imports it to also get the display options. Thus it is better to have it as part of it so this information always stays together. We said that going forward what we will aim for is to consolidate all display related settings in one JSON(B) field like the one added in this PR.

flaminic:
> @flaminic we discussed the datastore feature I suggested today in the platform backend meeting and one reason to keep this information in the metadata objects themselves is that often people want this to transfer with the matadata object. So if you were to export a dataset you want the system that imports it to also get the display options. Thus it is better to have it as part of it so this information always stays together. We said that going forward what we will aim for is to consolidate all display related settings in one JSON(B) field like the one added in this PR.

Amazing thanks @jbee. Tagging @Birkbjo as he might be interested in the above comment.

PR #17914, opened 2024-06-28
fix: User analysis org is refered to user capturing org in /analytics[2.42-DHIS2-13498]

d-bernat:
The user analysis org set up in user management is not used in data analysis app.


d-bernat:
Do not add run_analytics_tests label, the e2e tests are failing! It will be added after new e2e db export file integration to build pipeline.

maikelarabori (approved):

PR #17916, opened 2024-06-28
chore: remove dhis-api EventService.createEvent DHIS2-17677

teleivo:
follow up after https://github.com/dhis2/dhis2-core/pull/17873

## Big picture

## This PR

`EventService.createEvent` was only used in tests and thus belongs in a test helper. `DhisConvenienceTest` base class already has a `createEvent` that fulfilled most of what the tests relied on.

The

`programStage.getGeneratedByEnrollmentDate()`

was always `false` so the param `enrollmentDate` was never used.

Set the default category option combo in the `DhisConvenienceTest`. Store
the event via the `IdentifiableObjectManager` as we do for many other test
fixtures created via the `DhisConvenienceTest` methods.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/predictor/EventPredictionServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/predictor/EventPredictionServiceTest.java
@@ -64,7 +64,6 @@ import org.hisp.dhis.program.AnalyticsType;
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.EnrollmentService;
 import org.hisp.dhis.program.Event;
-import org.hisp.dhis.program.EventService;
 import org.hisp.dhis.program.Program;
 import org.hisp.dhis.program.ProgramIndicator;
 import org.hisp.dhis.program.ProgramIndicatorService;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/predictor/EventPredictionServiceTest.java
```diff
@@ -257,16 +260,13 @@ public void setUpTest() {
         enrollmentService.enrollTrackedEntity(
             trackedEntity, program, dateMar20, dateMar20, orgUnitA);
     enrollmentService.addEnrollment(enrollment);
-    Event stageInstanceA =
-        eventService.createEvent(enrollment, stageA, dateMar20, dateMar20, orgUnitA);
-    Event stageInstanceB =
-        eventService.createEvent(enrollment, stageA, dateApr10, dateApr10, orgUnitA);
+    Event stageInstanceA = createEvent(stageA, enrollment, orgUnitA);
```
enricocolasante, line 263: While we are at it, can we rename `stageInstanceA` to `eventA`?

enricocolasante (approved):

muilpp (approved):

PR #17934, opened 2024-07-01
fix: Validate no datavalues for some event statuses [DHIS2-17658]

enricocolasante:
Following on https://github.com/dhis2/dhis2-core/pull/17890.
Event statuses `OVERDUE`, `SKIPPED` and `SCHEDULE` do not allow to define data values. This PR is adding a validation to verify this.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
@@ -157,6 +157,7 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
+  E1315("Status`{0}` does not allow to define data values."),
   E1316("No event can transition from status `{0}` to status `{1}.`"),
 
   /* Relationship */
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataValuesValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataValuesValidator.java
@@ -42,6 +42,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.hisp.dhis.dataelement.DataElement;
+import org.hisp.dhis.event.EventStatus;
 import org.hisp.dhis.fileresource.FileResource;
 import org.hisp.dhis.program.ProgramStage;
 import org.hisp.dhis.program.ProgramStageDataElement;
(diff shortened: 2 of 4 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataValuesValidator.java
```diff
@@ -80,10 +81,20 @@ public void validate(Reporter reporter, TrackerBundle bundle, Event event) {
       validateDataValue(reporter, bundle, dataElement, dataValue, programStage, event);
     }
 
+    if (EventStatus.STATUSES_WITHOUT_DATA_VALUES.contains(event.getStatus())) {
```
teleivo, line 84: I would say keep the style of the methods below and move the if inside like in https://github.com/dhis2/dhis2-core/blob/f9d763cf8f351d3c22b9eab11f815fa6847cfc76/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DataValuesValidator.java#L150

so it reads

```java
validateDataValuesSetInAllowedEventStatus(reporter, event);
validateMandatoryDataValues(reporter, bundle, event);    validateDataValueDataElementIsConnectedToProgramStage(reporter, bundle, event, programStage);
```

We could also shorten the method names as we are in the `DataValuesValidator`. So for the new one maybe `validateEventStatus`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
```diff
@@ -157,6 +157,7 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
+  E1315("Status `{0}` does not allow to define data values."),
```
teleivo, line 160: Should we print a static list of event status that do not allow setting data values? As in give the user all the information needed to fix their problem.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/events/EventsDataValueValidationTests.java
```diff
@@ -126,11 +126,7 @@ public void shouldValidateWhenNoDataValue(String validationStrategy, String even
   }
 
   @ParameterizedTest
-  @CsvSource({
-    "ON_COMPLETE,ACTIVE",
-    "ON_UPDATE_AND_INSERT,SCHEDULE",
-    "ON_UPDATE_AND_INSERT,SKIPPED"
-  })
+  @CsvSource({"ON_COMPLETE,ACTIVE"})
```
teleivo, line 129: What does this do?

enricocolasante: This test was checking the that mandatory values in `SCHEDULE` and `SKIPPED` statuses were not validated but now we are even more restrictive so those tests cannot be reproduced anymore.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/tracker/imports/validation/ValidationCode.java
```diff
@@ -157,7 +157,8 @@ public enum ValidationCode {
   E1312("Referral events need to have both sides of a relationship."),
   E1313(
       "Event {0} of an enrollment does not point to an existing tracked entity. The data in your system might be corrupted"),
-  E1315("Status `{0}` does not allow to define data values."),
+  E1315(
+      "Status `{0}` does not allow to define data values. Statuses that allow to define data values are: {1}"),
```
teleivo, line 161:
```suggestion
      "Status `{0}` does not allow defining data values. Statuses that do allow defining data values are: {1}"),
```

teleivo (approved):

PR #17935, opened 2024-07-01
fix: Remove indicator from section on deletion

larshelge:
Removes indicator from data set sections when indicator is deleted.

Cleans up code related to `SectionService` and `SectionStore`. Updates Javadoc. Removes unused methods.

Added integration test.


david-mackessy (commented):
We try to have any fix accompanied with a test to show that the issue has been resolved. Can a test be added?
Is this issue separate from the indicator merge function? (just curious, as indicator refs are removed from sections when being deleted as part of a merge)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/indicator/IndicatorServiceTest.java
```diff
@@ -281,4 +281,22 @@ void testNumeratorTranslation() {
     assertEquals(numeratorTranslated, indicatorA.getDisplayNumeratorDescription());
     assertEquals(denominatorTranslated, indicatorA.getDisplayDenominatorDescription());
   }
+
+  @Test
+  void testRemoveIndicator() {
```
david-mackessy, line 286: It doesn't look like this actually tests the new code added, whether an `Indicator` has been removed from a `Section` or not. I pulled the branch code, commented out the new addition to the `SectionDeletionHandler` and this test still passes.
Should there not be a `Section` with `Indicator`s which should then be checked after the `Indicator` is deleted?

larshelge: I can add it. It just becomes a bit too arbitrary. There is a very high number of associations between data set and other entities, and all other types of entities for that matter, and we have almost no test coverage for deletion of associations. So we could add this particular case, but it will still give a very low confidence overall, so just wondering if it is worth it.

david-mackessy: I don't think it's arbitrary at all. A good test confirms expected behaviour & confirms the presence of a bug if there's a regression.
If this change is in fact a fix (Indicator deletion prohibited when associated with Section) then surely we'd want to make sure that this scenario can't happen again?
This topic arises very frequently in our codebase whereby a simple test would have prevented a regression in behaviour.

larshelge: Okay no problem. Will update the test.

Moving the work to a new PR due to the signed commit requirement:

https://github.com/dhis2/dhis2-core/pull/19015

PR #17946, opened 2024-07-02
fix: cast numeric type in SqlLiteral [DHIS2-13925]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-13925
### Issue
- User create programIndicator with expression "22/10"
- Expected result returned by `api/analytic` is "2.2". However "2" is returned.

### Fix
- The Expression parser doesn't have `::numeric` cast in the sql query. Therefore, a select statement `select 22/10" in postgresql will always return "2". We need to cast either "22" or "10" to `numeric` so postgreSQL will not ignore the decimal.
- The expected sql query generated should be `select 22::numeric/10::numeric`
- I have add a new `NumericLiteral` which do `numeric` cast if the expression contains a math equation which matches regex pattern `"^[0-9+\\-*/]+$"`

### Test
- Successfully run analytics table job on my local.
- Fixed all tests that throw error after code change .


jimgrace (commented):
Looks good, thanks Viet

maikelarabori (approved):

vietnguyen:
Closing this PR can't be merged because it contains unsigned commit.
Create a new one here https://github.com/dhis2/dhis2-core/pull/20162

PR #17978, opened 2024-07-04
fix: Remove dhis-service-program-rule module [DHIS2-17706]

enricocolasante:
It seems a big PR but it is just moving files around:
- Metadata classes are moved to `dhis-service-core` into package `programrule`
- Data classes are moved to `dhis-service-tracker` into package `programrule`
- `dhis-service-program-rule` has been removed and so all the pom files were updated

I was trying to break the PR down but then mvn dependency analyzer was giving an hard time, so I did it in just one step.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramRuleActionController.java
```diff
@@ -39,8 +39,8 @@
 import org.hisp.dhis.i18n.I18nManager;
 import org.hisp.dhis.program.ProgramIndicator;
 import org.hisp.dhis.programrule.ProgramRuleAction;
-import org.hisp.dhis.programrule.engine.ProgramRuleEngine;
 import org.hisp.dhis.rules.models.RuleValidationResult;
+import org.hisp.dhis.tracker.imports.programrule.engine.ProgramRuleEngine;
 import org.hisp.dhis.webapi.controller.AbstractCrudController;
```
teleivo, line 44: Should we move these 2 controllers closer/into a tracker package? This controller/event package still contains things that have nothing to do with events 😅

teleivo (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/ProgramRuleActionController.java
enricocolasante: Yes but we need to be careful...
These controllers are also metadata controllers, you can create and update program rules with them.
In my head the endpoints to get the description of an expression shouldn't live here at all, and then they will make sense in tracker.
We already had some discussion on who should be responsible for metadata and if we start moving them into tracker I am just scared that we will end up with a lot of responsibility :sweat_smile: 
But anyway, they are misplaced here and they should be moved

teleivo: Got it, thank you. Let's think about it some more then 😅

muilpp (approved):

PR #17980, opened 2024-07-04
fix: Remove WebHookService [DHIS2-17638]

enricocolasante:
Removing `DefaultTrackerNotificationWebHookService`.
This was before used in old tracker. At some point we will need to make it work in tracker as well but most likely we do not need all the sending application event complexity


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/ProgramNotificationListener.java
```diff
@@ -44,30 +40,8 @@
 public class ProgramNotificationListener {
   private final ProgramNotificationService programNotificationService;
 
-  private final TrackerNotificationWebHookService trackerNotificationWebHookService;
-
   @TransactionalEventListener(fallbackExecution = true)
   public void onEnrollment(ProgramEnrollmentNotificationEvent event) {
```
enricocolasante, line 44: @teleivo This is still used somewhere down the line in the SMS part.
It should be remove when we fix it there

muilpp (approved):

teleivo (approved):

PR #17982, opened 2024-07-04
fix: Include capture scope in search scope [DHIS2-17708]

muilpp:
It currently possible for users to configure org unit scopes such that the capture scope and search scope are entirely independent. Consequently, the tracker must adapt to this configuration and assume that users can search within their capture scope, even if the scopes are unrelated.

The goal of this PR is to address this scenario by creating a new set to hold the org unit values for both scopes. This set will reside in the `UserDetails` class, and will be propagated throughout the system from there.

There are still remaining usages of the search scope org units Set. Both platform and analytics teams will be updated on this issue so they can determine if it's convenient for them to use the new one as well. The specific usages are in:
- `DefaultUserService.handleUserQueryParams` and `DefaultUserService.validateUserCreateOrUpdateAccess`
- `UserDeletionHandler.deleteOrganisationUnit`
- `OrganisationUnitController.getObjectList`
- `MetadataOrgUnitMergeHandler.mergeUsers` and `MetadataOrgUnitMergeHandler.splitUsers`
- `DefaultDataQueryService.getUserOrgUnits`

Ticket: https://dhis2.atlassian.net/browse/DHIS2-17708


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -159,6 +162,12 @@ static UserDetails createUserDetails(
               ? setOfIds(user.getTeiSearchOrganisationUnitsWithFallback())
               : (searchOrgUnitUids.isEmpty() ? orgUnitUids : searchOrgUnitUids);
 
+      Set<String> userEffectiveSearchOrgUnitIds =
+          Stream.of(userOrgUnitIds, userSearchOrgUnitIds)
+              .filter(Objects::nonNull)
+              .flatMap(Collection::stream)
```
jbee, line 168: This is a bit odd way to get where `Stream.concat` would get you, like you did in the `User`, but work too I guess :)

jbee (approved):
Looks good!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
muilpp: That's how I originally did it, but sonar complained about a possible NPE, that's why I changed it

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/User.java
```diff
@@ -773,33 +774,21 @@ public Set<OrganisationUnit> getDataViewOrganisationUnitsWithFallback() {
...
-  public OrganisationUnit getTeiSearchOrganisationUnitWithFallback() {
-    return hasTeiSearchOrganisationUnit() ? getTeiSearchOrganisationUnit() : getOrganisationUnit();
-  }
-
   /** Returns the tei search organisation units or organisation units if not exist. */
   public Set<OrganisationUnit> getTeiSearchOrganisationUnitsWithFallback() {
     return hasTeiSearchOrganisationUnit() ? teiSearchOrganisationUnits : organisationUnits;
   }
 
+  /** Returns all search organisation units and organisation units. */
```
teleivo, line 786: I think this javadoc would benefit from the explanation you give in the PR or issue. Maybe

`Users capture scope org units can be outside of the search scope org units. The effective search org units represent their union.
`
Then maybe something about when this should be used over the other methods. The other related org unit methods could also link to this one in their javadoc.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
```diff
@@ -179,7 +179,7 @@ private OrganisationUnit validateRequestedOrgUnit(String orgUnitUid, User user)
     }
 
     if (!organisationUnitService.isInUserHierarchy(
```
teleivo, line 181: Can `user.isInUserEffectiveSearchOrgUnit` be used here and similar usages later on?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapper.java
```diff
@@ -105,8 +105,7 @@ public EnrollmentQueryParams map(EnrollmentOperationParams operationParams)
   private void mergeOrgUnitModes(
       EnrollmentOperationParams operationParams, User user, EnrollmentQueryParams queryParams) {
     if (user != null && operationParams.getOrgUnitMode() == ACCESSIBLE) {
-      queryParams.addOrganisationUnits(
-          new HashSet<>(user.getTeiSearchOrganisationUnitsWithFallback()));
+      queryParams.addOrganisationUnits(new HashSet<>(user.getEffectiveSearchOrganisationUnits()));
```
teleivo, line 108: Can we remove the `new HashSet` in this and similar calls?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerOwnershipManagerTest.java
```diff
@@ -537,6 +537,22 @@ void shouldNotFindTrackedEntityWhenProgramSuppliedAndUserIsNotOwner()
...
+    programA.setAccessLevel(AccessLevel.OPEN);
+    programService.updateProgram(programA);
+    userB.setOrganisationUnits(Set.of(organisationUnitA));
+    userB.setTeiSearchOrganisationUnits(Set.of(organisationUnitB));
+    userService.updateUser(userB);
+    injectSecurityContext(UserDetails.fromUser(userB));
+
+    assertEquals(
+        trackedEntityA1,
+        trackedEntityService.getTrackedEntity(
```
teleivo, line 552: Why is this in the TrackerOwnershipManagerTest? I think this should be part of our TE service test

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapper.java
muilpp: We could, but first we'd need to replace `User` by `UserDetails`.
That kind of user is present in multiple locations in our validators and I'm not sure that's a refactor we want to do in this ticket.

teleivo (approved):

PR #18010, opened 2024-07-08
chore: remove dhis-api deleteEvent DHIS2-17638

teleivo:


muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/DeleteEventSMSListener.java
```diff
@@ -101,7 +98,7 @@ protected SmsResponse postProcess(IncomingSms sms, SmsSubmission submission, Use
       throw new SMSProcessingException(SmsResponse.INVALID_EVENT.set(subm.getEvent()));
     }
 
-    apiEventService.deleteEvent(event);
+    identifiableObjectManager.delete(event);
```
enricocolasante, line 101: That should use tracker importer at some point.
Should we put a TODO just to remember?

enricocolasante (approved):

PR #18016, opened 2024-07-08
chore: remove dhis-api EventService DHIS2-17638

teleivo:
* inline `saveEventDataValuesAndSaveEvent` into the sms code. The easiest to get that done is through duplication as the class hierarchy would not allow us inheriting from the same base class.

We have a meeting to discuss the sms code today. We will then make a plan and refactor that code to do its crud via the tracker importer instead of this code and the manager.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/RegisterSMSListener.java
```diff
@@ -142,4 +166,104 @@ protected void register(
         sms.getOriginator(),
         INFO);
   }
+
+  public void saveEventDataValuesAndSaveEvent(
```
enricocolasante, line 170: Could this be private?

enricocolasante (approved):

muilpp (approved):

PR #18023, opened 2024-07-08
feat: Users data capture orgunit should be within their data view hierarchy

jason-p-pickering:
When a data view organisation unit has been specified for  a user, all data capture organisation units should fall within the hierarchy of that data view organisation unit. This situation can occur when the data view organisation unit has been set at a lower level of the hierarchy or different branch of the hierarchy. In this case, the user might not be able to view their data which they have entered. 

Note that users who have NO data view organisation unit specified, default to their data capture unit. Thus, this integrity check only applies to users who have a data capture orgunit specified, which is not within their data view hierarchy.


dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/users/users_capture_ou_not_in_data_view_ou.yaml
```diff
@@ -0,0 +1,121 @@
...
+  GROUP BY u.userinfoid
+  HAVING array_length(ARRAY_AGG(dop) FILTER (WHERE dop IS NOT NULL), 1) > 0
+  ) x on x.userinfoid = ui.userinfoid
+  INNER JOIN organisationunit ou on ou.path = any(x.invalid_data_orgunits)
+  GROUP BY ui.uid, ui.username;
+details_id_type: users
+severity: SEVERE
+introduction: >
+  Users who can enter data should be able to view their own data. This situation occurs when a user has 
+  a data capture organisation unit unit which is not within the data view organisation unit hierarchy.
```
jbee, line 114:
```suggestion
  a data capture organisation unit which is not within the data view organisation unit hierarchy.
```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/users/users_capture_ou_not_in_data_view_ou.yaml
```diff
@@ -0,0 +1,121 @@
...
+      SELECT path AS dop
+      FROM unnest(u.data_orgunits_paths) AS path
+      WHERE NOT EXISTS (
+          SELECT 1
+          FROM unnest(COALESCE(v.view_orgunit_paths, '{}')) AS view_path
+          WHERE path LIKE view_path || '%'
+      )
+  ) invalid_paths ON true
+  GROUP BY u.userinfoid
+  HAVING array_length(ARRAY_AGG(dop) FILTER (WHERE dop IS NOT NULL), 1) > 0
```
jbee, line 67: I this the filter that only finds cases where at least one data view association has been made so the "none=all" logic does not apply?

jbee (approved):
I think it looks good (apart from a typo) but it is a lengthy SQL to grasp :sweat_smile:

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/users/users_capture_ou_not_in_data_view_ou.yaml
jason-p-pickering: Not quite. That should filter out cases where the capture path is not like any of the data view paths. This bit `INNER JOIN user_view_orgunits v ON u.userinfoid = v.userinfoid` uses an inner join (as opposed to a left join), so if there are no view organisation units, then basically all of their data capture orgunits are valid, and no results can be returned for that particular user.

jbee (changes requested):
As Jason told me the SQL is very slow. As a check in real use it would not be a problem but as a part of the test suite it is. Lets see if we can make that SQL fast(er). Just from the kind of thing it wants to find I would assume this should be a quite fast check :thinking:

jbee (approved):

PR #18031, opened 2024-07-09
chore: Remove enrollmentService.getEnrollment [DHIS2-17712-2]

muilpp:
### Big picture

### This PR

Moves the method getEnrollment(uid) from the api module to the service tracker module.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -71,6 +70,11 @@ class DefaultEnrollmentService
 
   private final EnrollmentOperationParamsMapper paramsMapper;
 
+  @Override
+  public Enrollment getEnrollment(String uid) throws ForbiddenException, NotFoundException {
```
teleivo, line 74: Can you please use our UID type like in the EventService. If its easier you can do it in a separate PR.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -75,7 +75,10 @@ public TrackerTypeReport deleteEnrollments(TrackerBundle bundle) {
 
       Entity objectReport = new Entity(TrackerType.ENROLLMENT, uid);
 
-      Enrollment enrollment = enrollmentService.getEnrollment(uid);
+      Enrollment enrollment = manager.get(Enrollment.class, uid);
```
teleivo, line 78: I know I did not for events but I am wondering if we should use the exporter service here @enricocolasante ?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EnrollmentSMSListener.java
```diff
@@ -185,23 +190,27 @@ protected SmsResponse postProcess(IncomingSms sms, SmsSubmission submission, Use
     // TODO: Unsure about this handling for enrollments, this needs to be
     // checked closely
     Enrollment enrollment;
-    boolean enrollmentExists = enrollmentService.enrollmentExists(enrollmentid.getUid());
+    boolean enrollmentExists = apiEnrollmentService.enrollmentExists(enrollmentid.getUid());
     if (enrollmentExists) {
-      enrollment = enrollmentService.getEnrollment(enrollmentid.getUid());
+      try {
+        enrollment = enrollmentService.getEnrollment(enrollmentid.getUid());
```
teleivo, line 196: You need to pass the user in. This does not have the user set in the context like in an HTTP request context.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/RelationshipSMSListener.java
```diff
@@ -175,10 +175,13 @@ private RelationshipItem createRelationshipItem(
         break;
 
       case PROGRAM_INSTANCE:
-        Enrollment progInst = enrollmentService.getEnrollment(objId.getUid());
-        if (progInst == null) {
+        Enrollment progInst;
```
teleivo, line 178:
```suggestion
        Enrollment enrollment;
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/EnrollmentServiceTest.java
```diff
@@ -293,13 +300,14 @@ void shouldNoteDeleteNoteWhenDeletingEnrollment() {
...
-    enrollmentService.addEnrollment(enrollmentA);
+    apiEnrollmentService.addEnrollment(enrollmentA);
 
     assertNotNull(enrollmentService.getEnrollment(enrollmentA.getUid()));
 
-    enrollmentService.deleteEnrollment(enrollmentA);
+    apiEnrollmentService.deleteEnrollment(enrollmentA);
 
-    assertNull(enrollmentService.getEnrollment(enrollmentA.getUid()));
+    Assertions.assertThrows(
```
teleivo, line 309: use a static import

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
muilpp: I'll do it in the following PR

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
muilpp: I think we should use the exporter here, the thing is using the service-tracker `enrollmentService`, we run the validations and map the Enrollment to a new object, which is then used by this class to delete the enrollment, and fails because the session already contains a different Enrollment object with the same id.
Long story short, we need to refactor this logic and it might need its own PR.

teleivo: thank you for checking 😄

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
enricocolasante: I think we need to discuss on how to proceed here.
I am not sure we should use the exporter service here. `deleteEnrollments` method should never be exposed and it should only be used by our importer that already did all the needed validations to check if the enrollment is there and if it can be deleted by the user (and also if the events can be deleted).

Here in the end we should just update the enrollment to `deleted=false` and all the lastUpdated info.

enricocolasante (approved):
Just a comment on the next steps when we delete

PR #18038, opened 2024-07-09
feat: Users data capture orgunit should be within their TEI search hierarchy

jason-p-pickering:
When a users data capture orgunit is not within their TEI search hierarchy, users may not be able to search for TEIs which they have entered. While chatting with @muilpp , it seems that the behavior in version 42+ is that users can search in both their capture OU as well as their TEI search OU. Thus, this integrity check should only be target in versions <41.


dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReaderTest.java
```diff
@@ -56,7 +56,7 @@ void testReadDataIntegrityYaml() {
 
     List<DataIntegrityCheck> checks = new ArrayList<>();
     readYaml(checks, "data-integrity-checks.yaml", "data-integrity-checks", CLASS_PATH);
-    assertEquals(72, checks.size());
+    assertEquals(73, checks.size());
```
jbee, line 59: Be aware that if we merge both your PRs this will break. We should make this a more relaxed condition

jbee (approved):

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReaderTest.java
jason-p-pickering: Roger that. Will fix in the merge but would be good to able to handle this independently.

PR #18049, opened 2024-07-10
fix: [WIP]Set invalid COC to default for sms code to fix upgrade failure [DHIS2-15169]

zubaira:
[DHIS2-15169](https://dhis2.atlassian.net/browse/DHIS2-15169)

[DHIS2-15169]: https://dhis2.atlassian.net/browse/DHIS2-15169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbee (commented):
Have you tried if this script can be run multiple times or succeeds in case any of the changed or added  columns and FKs already are changed or exist? I would think this would fail this script. It should be possible to write it so it does not fail in those scenarios.

zubaira:
Closing the issue for now until we find a way to skip the previous (faulty) Flyway migration.

PR #18067, opened 2024-07-11
chore: Remove enrollmentService.enrollmentExists [DHIS2-17712-3]

muilpp:
### Big picture

### This PR

Removes the method `enrollmentExists(uid)` from the api module as we can achieve the same result by using the method `getEnrollment` from the service tracker module.

We need to discuss how to deal with not found enrollments, since it might be because the enrollment truly doesn't exist, or because the user has no metadata access to it. Depending on the reason, we'll need to react differently (I guess).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EnrollmentSMSListener.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EnrollmentSMSListener.java
@@ -188,28 +188,30 @@ public class EnrollmentSMSListener extends EventSavingSMSListener {
 
     TrackedEntity te = teService.getTrackedEntity(teUid.getUid());
 
-    // TODO: Unsure about this handling for enrollments, this needs to be
-    // checked closely
-    Enrollment enrollment;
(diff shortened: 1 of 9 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EnrollmentSMSListener.java
```diff
@@ -188,28 +188,30 @@ protected SmsResponse postProcess(IncomingSms sms, SmsSubmission submission, Use
...
-      } catch (ForbiddenException | NotFoundException e) {
-        throw new SMSProcessingException(SmsResponse.INVALID_ENROLL.set(enrollmentid.getUid()));
-      }
-      // Update these dates in case they've changed
+    Enrollment enrollment = null;
+    try {
+      enrollment =
+          enrollmentService.getEnrollment(enrollmentid.getUid(), UserDetails.fromUser(user));
       enrollment.setEnrollmentDate(enrollmentDate);
       enrollment.setOccurredDate(incidentDate);
```
teleivo, line 196: please rename the var to occuredDate

teleivo (approved):

PR #18068, opened 2024-07-11
fix: Validate mandatory data values also on UPDATE [DHIS2-17560]

enricocolasante:
We were validating mandatory data values when creating or when transitioning from a status that does not allow to define data values to one that allows it. This is an issue in some cases:
- When a program rule is evaluated to true based on some temporal factor. A data value could then be not mandatory at the moment of creation/transition, but become mandatory at a later stage.
- When metadata is changed. A DataElement is now mandatory.
- When the programStage validation strategy is `ON_COMPLETE` and the status was `ACTIVE` and become `COMPLETE`. Becasue it is not a `CREATE` we were skipping validation here. This is a really common use case that we were missing.

The solution is to not consider anymore if the import is a CREATE or an UPDATE but always validate mandatory data values (when data values need to be validated, based on the status and the programStage validation strategy) based on all the data values (DB + payload).

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/programrule/executor/event/SetMandatoryFieldExecutorTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/programrule/executor/event/SetMandatoryFieldExecutorTest.java
@@ -190,9 +190,9 @@ class SetMandatoryFieldExecutorTest extends DhisConvenienceTest {
     when(preheat.getIdSchemes()).thenReturn(TrackerIdSchemeParams.builder().build());
     when(preheat.getDataElement(DATA_ELEMENT_UID.getValue())).thenReturn(dataElement);
     when(preheat.getProgramStage(MetadataIdentifier.ofUid(programStage))).thenReturn(programStage);
-    Event event = getEventWithMandatoryValueNOTSet(savedStatus);
+    Event event = getEventWithMandatoryValueNOTSet(newStatus);
     when(preheat.getEvent(event.getUid()))
-        .thenReturn(eventWithMandatoryValue(event.getUid(), newStatus));
+        .thenReturn(eventWithMandatoryValue(event.getUid(), savedStatus));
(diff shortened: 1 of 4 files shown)


muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/programrule/executor/event/SetMandatoryFieldExecutorTest.java
```diff
@@ -184,13 +185,14 @@ void shouldFailValidationWhenMandatoryFieldIsDeletedAndStrategyIsUpdate() {
...
-  void shouldPassValidationWhenMandatoryFieldIsNotPresentAndStrategyIsUpdate(
+  void shouldPassValidationWhenMandatoryFieldIsNotPresentInPayloadButPresentInDB(
       EventStatus savedStatus, EventStatus newStatus) {
     when(preheat.getIdSchemes()).thenReturn(TrackerIdSchemeParams.builder().build());
     when(preheat.getDataElement(DATA_ELEMENT_UID.getValue())).thenReturn(dataElement);
     when(preheat.getProgramStage(MetadataIdentifier.ofUid(programStage))).thenReturn(programStage);
     Event event = getEventWithMandatoryValueNOTSet(savedStatus);
-    when(preheat.getEvent(event.getUid())).thenReturn(event(event.getUid(), newStatus));
+    when(preheat.getEvent(event.getUid()))
+        .thenReturn(eventWithMandatoryValue(event.getUid(), newStatus));
```
teleivo, line 195: Should that not be `savedStatus` as the preheat has the events from the DB 🤔 
and `newStatus` should be set on the `event` and its reference in the `bundle`.

enricocolasante: Yes :)
I changed and everything is still working because we are transitioning between statuses that allow data values, so swapping the saved and new didn't change the result.

teleivo (approved):

PR #18076, opened 2024-07-12
chore: Backport new data integrity checks from master

jason-p-pickering:
Adding the following integrity checks to this version: 
-Users capture OU within data view OU

-Users capture OU within TEI search OU

-Users with no roles

-User roles with no users

-User roles with no authorities


david-mackessy (approved):
nice work 👍

PR #18078, opened 2024-07-12
chore: Remove method enrollmentExistIncludingDeleted [DHIS2-17712]

muilpp:
### Big picture

### This PR

Removes the method enrollmentExistsIncludingDeleted(uid) from the api module, since it's only used in tests and we can achieve the same result by using the method getEnrollments from the service tracker module.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
@@ -28,8 +28,6 @@
 package org.hisp.dhis.maintenance;
 
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ALL;
-import static org.hisp.dhis.utils.Assertions.assertIsEmpty;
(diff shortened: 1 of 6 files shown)


enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
```diff
@@ -466,4 +486,12 @@ private boolean eventExistsIncludingDeleted(String uid) {
         jdbcTemplate.queryForObject(
             "select exists(select 1 from event where uid=?)", Boolean.class, uid));
   }
+
+  private EnrollmentOperationParams createEnrollmentParamsIncludingDeleted(String enrollmentUid) {
+    return EnrollmentOperationParams.builder()
```
teleivo, line 491: If you look at the code using this you can see that all usages actually want to do is to check if `enrollmentExistsIncludingDeleted`. So create `enrollmentExistsIncludingDeleted` like

```java 
private boolean enrollmentExistsIncludingDeleted(Enrollment enrollment)
```

This will make the tests more readable.

teleivo (approved):

PR #18079, opened 2024-07-12
fix: Do not use deletionHandlers for tracker data [DHIS2-17743]

enricocolasante:
`DeletionHandlers` are not meant to be used for tracker data, so I am removing all the logic related just to tracker data. (It still make sense to block the deletion of a metadata if there is data linked to it).

The only way to delete any tracker data is using `TrackerObjectDeletionService`. This service is meant to be used after all the validations to check if the user can delete the entities are run.

At the moment only `TrackerImportService` is using `TrackerObjectDeletionService`, after this PR also `DeduplicationService` will use it. This is the first discussion that we will need to have. `DeduplicationService` should use the importer or can be trusted that all the relevant validations were already run? 
At the moment, make `DeduplicationService` use importer will need quite a bit of work as it is doing at least one update and one deletion and we will need to coordinate it.

Signature of `TrackerObjectDeletionService` was changed, now it is accepting a list of uids to delete. This removes the need of conversion between models and now the user info are retrieved from the context. This is ok for deduplication feature but we will need to check if we arrive to use this service for SMS.

The delete methods can return a `NotFoundException`. `DeduplicationService` is catching that exception and raise a generic 500. I am not sure if we need to think about creating a proper  deduplication exception here or it is better to wait and see how we will change the `DeduplicationService`. (If it is going to use the importer, the exception handling will be completely different).

**Next steps**
- Check what is really needed to be deleted when we are soft deleting entities.
- If we agree that `TrackerObjectDeletionService` is and must be the only place where we delete tracker objects, we could implement here our stores (our queries to delete) removing the need to use the `IdentifiableObjectManager` to delete and to get the object.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/TrackedEntityDataValueChangeLogDeletionHandler.java
```diff
@@ -44,14 +44,9 @@ public class TrackedEntityDataValueChangeLogDeletionHandler extends DeletionHand
...
     whenDeleting(DataElement.class, this::deleteDataElement);
-    whenDeleting(Event.class, this::deleteEvent);
   }
 
   private void deleteDataElement(DataElement dataElement) {
     trackedEntityDataValueAuditService.deleteTrackedEntityDataValueChangeLog(dataElement);
   }
-
-  private void deleteEvent(Event event) {
-    trackedEntityDataValueAuditService.deleteTrackedEntityDataValueChangeLog(event);
```
muilpp, line 55: when deleting an event we were deleting its change logs too, but I don't see us doing it in the importer.
https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java#L112

am I missing something or is it intentional?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -168,7 +169,7 @@ public void setUpTest() {
     attributeService.addTrackedEntityAttribute(filtG);
 
     User user = createUserWithAuth("testUser");
-    user.setTeiSearchOrganisationUnits(Set.of(organisationUnit));
```
teleivo, line 171: I assume this was an accident, as this happens to me with Intellij 😁 `Set.of` is easier to read than just `of`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/SoftDeleteHibernateObjectStore.java
```diff
@@ -53,7 +52,6 @@ public SoftDeleteHibernateObjectStore(
 
   @Override
   public void delete(@Nonnull SoftDeletableObject object) {
-    publisher.publishEvent(new ObjectDeletionRequestedEvent(object));
```
teleivo, line 56: Are you sure no one else but us was listening to this event?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -60,43 +66,49 @@ public class DefaultTrackerObjectsDeletionService implements TrackerObjectDeleti
...
       Enrollment enrollment = manager.get(Enrollment.class, uid);
       if (enrollment == null) {
         throw new NotFoundException(Enrollment.class, uid);
       }
-      enrollment.setLastUpdatedByUserInfo(bundle.getUserInfo());
+      enrollment.setLastUpdatedByUserInfo(userInfoSnapshot);
+
+      List<String> events =
+          enrollment.getEvents().stream()
+              .filter(event -> !event.isDeleted())
```
teleivo, line 92: Are these `enrollment.getEvents()` coming from the payload? If so are we safe from users setting `deleted` on an event in the payload?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DefaultDeduplicationService.java
```diff
@@ -166,8 +170,11 @@ private void merge(DeduplicationMergeParams params)
         original, duplicate, mergeObject.getTrackedEntityAttributes());
     potentialDuplicateStore.moveRelationships(original, duplicate, mergeObject.getRelationships());
     potentialDuplicateStore.moveEnrollments(original, duplicate, mergeObject.getEnrollments());
-
-    potentialDuplicateStore.removeTrackedEntity(duplicate);
+    try {
+      trackerObjectDeletionService.deleteTrackedEntities(List.of(duplicate.getUid()));
+    } catch (NotFoundException e) {
+      throw new RuntimeException("Could not find TrackedEntity: " + duplicate.getUid());
```
muilpp, line 176: can this NFE even happen?
I see in the controller we are getting the original and duplicate TEs from the DB, and in case they are not found, we already throw an exception.
https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java#L173

dhis-2/dhis-test-e2e/src/main/resources/config.properties
```diff
@@ -1,6 +1,6 @@
 # property pointing to running DHIS2 instance.
 # must include /api
-instance.url=http://localhost:8080/api
+instance.url=http://localhost:8080/dhis_war/api
```
teleivo, line 3: is that a mistake?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/TrackedEntityDataValueChangeLogDeletionHandler.java
muilpp: nevermind, I just saw you add in this PR the change log deletion while deleting the event

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
enricocolasante: These events are from the DB, the delete methods are only getting the ids of the entities to delete and then we retrieving them from the DB.

dhis-2/dhis-test-e2e/src/main/resources/config.properties
enricocolasante: Yes :)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackedEntityServiceTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/SoftDeleteHibernateObjectStore.java
enricocolasante: Yes.
This was fired to check what to do when a tracker entity was deleted and now all the methods like `deleteTrackerEntity` and `deleteEnrollment` in deletion handlers are removed.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DefaultDeduplicationService.java
enricocolasante: I don't think it will ever happen as well.
But as I am not sure if we should use the importer here instead, I didn't want to propagate the exception all the way up and just swallowing the exception seemed too much to me

muilpp (approved):

teleivo (approved):

PR #18086, opened 2024-07-15
test: improve ability to override base dhis.conf in tests DHIS2-16557

teleivo:
## Why

We have too many base test classes that do not provide some common functionality. They seem to exist because configuring our tests is complicated and our base test classes and configs are not well structured and named. Our configurations should be composable so they can be combined to serve a very specific test scenario.

## What

* Module `dhis-support-test` and package/src `/src/main/java/org/hisp/dhis/config` only contains reusable test configurations after this PR.
* `dhis.conf` files are now suffixed with `dhis.conf` as at first site its not clear that this is what they are.
* Classes annotated with Springs `@Configuration` are suffixed with `Configuration`: `H2DhisConfiguration` or `PostgresDhisConfiguration`. They are responsible for setting up the `DhisConfigurationProvider` bean.
* Classes implementing the `DhisConfigurationProvider` are suffixed with `ConfigurationProvider`: `H2DhisConfigurationProvider` or `PostgresDhisConfigurationProvider`
* @Configuration` that are only used by a single test class are moved to a static inner class inside the test class
* `dhis.conf` files only used by a single test are moved into the module the test lives in
* Delete base test class `HibernateCacheBaseTest` and merge it with the only tests `HibernateQueryCacheTest` using it

Regarding base class `HibernateCacheBaseTest` and `HibernateQueryCacheTest`:

The main difference in `dhis.conf` that is needed for this test is

```
diff dhis-2/dhis-support/dhis-support-test/src/main/resources/hibernateCacheTest.conf dhis-2/dhis-support/dhis-support-test/src/main/resources/postgresTestConfig.conf
< hibernate.cache.use_query_cache=true
< hibernate.cache.use_second_level_cache=true
---
> hibernate.cache.use_query_cache=false
> hibernate.cache.use_second_level_cache=false
```

This can now easily be done and the diff is visible if you look at the `HibernateQueryCacheTest.DhisConfiguration`.

## Spring Configuration and base class specifics

`@ContextConfiguration` is inherited by default. We can thus inherit the config of a base class like `IntegrationTestBase` and override what should be different. In the case of `HibernateQueryCacheTest` that is the `DhisConfigurationProvider` bean as we want to configure the `'hibernate.cache'`.

Allow overriding `IntegrationTestBase.before`. This allows us to reuse `IntegrationTestBase` while changing the BeforeEach setup behavior.

Allow creating a `TestDhisConfigurationProvider` with properties or properties from a file. There is no need for this class to be abstract.


maikelarabori (approved):

david-mackessy (commented):
looks nice 👍

jbee (commented):
It is already merged but I wanted to have a look anyway - really nice improvement :+1:

PR #18100, opened 2024-07-16
feat: Users with invalid usernames

jason-p-pickering:


david-mackessy (approved):
Do current/newer versions prevent invalid usernames? Older ones allow?
What scenario would this check be used in?
Would it mainly be in older systems where they are looking to upgrade? Or are there more?
There's a possibility that username validation could be out of sync with this check, maybe put a comment where the username validation is done to remind that if changed, this check should also be changed?

jason-p-pickering:
> Do current/newer versions prevent invalid usernames? 
Yes, the WebAPI defintely does block the creation of new users with invalid usernames. 

>Older ones allow? 
Yes, but not clear exactly when this restriction was implemented. Maybe @netroms knows.

>What scenario would this check be used in? Would it mainly be in older systems where they are looking to upgrade? Or >are there more? 

I have tried a few different scenarios with updating users with "invalid" user names, and things do seem to work. I was a bit surprised that a PUT operation worked, but it seemed to. While it does seem to work now, perhaps it will not always work Thus, a warning to admins now that they might need to sort it out. 

Another scenario I could think of is one we came across with users names with emojis and invisible UNICODE characters in their user name in an actual system. This does not seem ideal, and while it does seem to work with these users, it might not be best practice. 

>There's a possibility that username validation could be out of sync with this check, maybe put a comment where the username validation is done to remind that if changed, this check should also be changed?
Good idea.  Will update

dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
jason-p-pickering, line 1: Note to self and code reviewers (@dhis2/platform-backend ) that we need to be sure to include the translations for new checks.

jbee (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/ValidationUtils.java
```diff
@@ -75,6 +75,11 @@
  * @author Lars Helge Overland
  */
 public class ValidationUtils {
+  // Note that a data integrity check replicates this pattern in a slightly different way.
+  // If this pattern is changed, the data integrity check should be updated as well.
+  // See
+  // dhis-service-administration/src/main/resources/data-integrity-checks/users/users_with_invalid_usernames.yaml
+  // for details.
```
david-mackessy, line 82: I think this might help someone in the future hopefully, thanks 👍

PR #18101, opened 2024-07-16
test: move duplicated code into DhisControllerTestBase DHIS2-16557

teleivo:
builds on
* https://github.com/dhis2/dhis2-core/pull/18086
* https://github.com/dhis2/dhis2-core/pull/18099

## Main Change

All test classes extending `DhisControllerTestBase` had almost the same code in them. The main difference was
* either use H2 or Postgres which can simply be steered via the `@ActiveProfiles("test-h2") `or `@ActiveProfiles("test-postgres")` annotations
* use a different `dhis.conf` which can be steered via a custom `DhisConfigurationProvider` written as a static inner class

Move that common code into `DhisControllerTestBase`. This makes it clear how child test classes differ and what they are for.

Inline a couple of base classes that are only used by one test like `AppControllerTest` and `ImpersonateUserControllerTest`.

## Small Changes

There might be better places for the helpers. Now that we have less code to compare and sift through this will hopefully become obvious.

* Move helpers like `await`, `switchToUserWithOrgUnitDataView` into the only tests using them. We can find a better place for them if they are truly generic and should be easier to discover. That being said base classes should not be a dumping ground for all the things.
* Rely on https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/BeforeEach.html behavior. A base class `@BeforeEach` will run before the child classes. It is way easier to understand and control when things run if we allow devs to specify this annotation in their own tests. This might even lead to fewer base classes.

## Main Learnings

These are some things we should internalize and rely on

* https://docs.spring.io/spring-framework/reference/testing/annotations/integration-spring/annotation-contextconfiguration.html is inherited by default. We can put H2/PostgresDhisConfiguration components together in a base class. Then use `@ActiveProfiles("test-h2") `or `@ActiveProfiles("test-postgres")` annotations to pick one over the other in a child class. https://docs.spring.io/spring-framework/reference/testing/testcontext-framework/ctx-management/env-profiles.html We can also disable inheritance if needed.
* https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/BeforeEach.html behavior. A base class `@BeforeEach` will run before the child classes.

## Next

* Explore removing the dedicated `DhisControllerConvenienceTest` and `DhisControllerIntegrationTest` for an annotation. Tests need to extend `DhisControllerTestBase` which configures the Spring context and one of these pairs

```java
@ActiveProfiles("test-h2")
@IntegrationH2Test
```

```java
@ActiveProfiles("test-postgres")
@IntegrationTest
```

* It looks as if the user related setup code has some left overs from the `UserDetails` refactoring. @netroms please have a look/reach out to me so we clean that up.


jbee (approved):
Great!

david-mackessy (approved):

PR #18105, opened 2024-07-17
chore: Remove getEnrollments(List<String>) [DHIS2-17712]

muilpp:
### Big picture

### This PR

Moves the method `getEnrollments(List<String>)` from the api module, to the service-tracker module. This method is just used by the deduplication service.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
@@ -27,6 +27,8 @@
  */
 package org.hisp.dhis.tracker.deduplication;
 
+import static org.hisp.dhis.security.Authorities.ALL;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -36,16 +38,19 @@ import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.common.CodeGenerator;
 import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.commons.collection.ListUtils;
+import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.organisationunit.OrganisationUnitService;
 import org.hisp.dhis.program.Enrollment;
-import org.hisp.dhis.program.EnrollmentService;
 import org.hisp.dhis.relationship.Relationship;
 import org.hisp.dhis.relationship.RelationshipItem;
 import org.hisp.dhis.relationship.RelationshipService;
 import org.hisp.dhis.relationship.RelationshipType;
+import org.hisp.dhis.security.Authorities;
 import org.hisp.dhis.security.acl.AclService;
 import org.hisp.dhis.trackedentity.TrackedEntity;
+import org.hisp.dhis.trackedentity.TrackerAccessManager;
 import org.hisp.dhis.trackedentityattributevalue.TrackedEntityAttributeValue;
+import org.hisp.dhis.tracker.export.enrollment.EnrollmentService;
 import org.hisp.dhis.user.CurrentUserUtil;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.user.UserDetails;
@@ -65,6 +70,7 @@ public class DeduplicationHelper {
   private final EnrollmentService enrollmentService;
 
   private final UserService userService;
+  private final TrackerAccessManager trackerAccessManager;
 
   public String getInvalidReferenceErrors(DeduplicationMergeParams params) {
     TrackedEntity original = params.getOriginal();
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
```diff
@@ -58,6 +59,15 @@ List<Enrollment> getEnrollments(EnrollmentOperationParams params)
   Page<Enrollment> getEnrollments(EnrollmentOperationParams params, PageParams pageParams)
       throws BadRequestException, ForbiddenException;
 
+  /**
+   * Returns a list of existing Enrollments from the provided UIDs
+   *
+   * @param uids Enrollment UIDs to check
+   * @return Enrollment list
+   */
+  List<Enrollment> getEnrollments(@Nonnull List<String> uids, UserDetails currentUser)
```
teleivo, line 68:
```suggestion
  List<Enrollment> getEnrollments(@Nonnull List<String> uids, UserDetails user)
```

Can you transition it to `List<UID>` please.

Regarding the javadoc: I would remove the `@param` and `@return` unless we have something useful to say about them. I would mention the user as in https://github.com/dhis2/dhis2-core/blob/79cb6c1eda5b3c22dab9862ff251e02ce2a8585d/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventService.java#L68-L72

teleivo, line 68: @muilpp and @enricocolasante will this not be called from an HTTP request? If so the current user should be set in the auth context and we do not need to pass it explicitly.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
```diff
@@ -274,7 +283,8 @@ public String getUserAccessErrors(
       return "Missing data write access to one or more Relationship Types.";
     }
 
-    List<Enrollment> enrollments = enrollmentService.getEnrollments(mergeObject.getEnrollments());
```
teleivo, line 277: Can we use the existing `getEnrollments(EnrollmentOperationParams params)` instead of creating a new method?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
muilpp: shouldn't we always get it from the context, even in those cases where it's provided in the request?

teleivo: That is something we have to discuss and settle on. Right now with SMS its not in the auth context so you cannot get it that way. Deduplication AFAIK is only triggered via HTTP, hence you should not have to pass the user.

I would also hope that you can reuse the existing service method to get some enrollments.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
muilpp: I thought about that, but for deduplication, we need access to all requested enrollments. If any are missing, we should raise an exception. The current method just skips any that aren't accessible, which isn't ideal. 
I could compare the requested enrollments to the final list to find any missing ones, but that means creating a new exception instead of using the ones from the access manager, which I'd rather stick with.
That being said, I'm not convinced by either solution.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
muilpp: With the exception of SMS, I think now it makes more sense to always get it from the context, before there was a call to the database, but now if we use `UserDetails` there's none anymore.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
teleivo: ok, lets just not forget about looking into this 😬

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
enricocolasante: I agree we should use a `List<UID>` but to do so we need to validate the uids that are strings in the `DeduplicationController`. We can either explicitly validate them or we can change the `MergeObject` to use `List<UID>` instead of `List<String>`.

In general I agree to use `UID` wherever it makes sense, and this means that if we validated a string to be a `UID` than it makes sense to never go back to string.

enricocolasante: We need to find our way to proceed but if we know how the method is used, I would not pass the user as a parameter and get it from the context.

enricocolasante (approved):
I have a couple of comments that can also be solved or discussed in later PRs.
Anyway, we should take a moment to evaluate the whole deduplication feature and check if it make sense to have its own stores or if it should use the importer for any change to the DB. In the end it would be 2 calls to the importer, one CREATE_AND_UPDATE to move stuff from duplicate to original and one DELETE to remove the duplicate.

teleivo (approved):

PR #18107, opened 2024-07-17
test: Cleanup auth related base test classes DHIS2-16557

teleivo:
builds on
* https://github.com/dhis2/dhis2-core/pull/18086
* https://github.com/dhis2/dhis2-core/pull/18099
* https://github.com/dhis2/dhis2-core/pull/18101

Cleanup auth related base test classes (only used by one test 🤷🏻 ). They differ from regular `DhisControllerConvenienceTest` tests
* in their context config. They use `WebMvcConfig.class` and not `MvcTestConfiguration.class, WebTestConfiguration.class`
* add some security filter to `MockMvc`

## Learning

We cannot move the `@ActivateProfiles("test-h2")` into the `@IntegrationH2Test` (same for postgres) because of https://github.com/spring-projects/spring-framework/issues/26145. This would have allowed us to delete the `DhisControllerIntegrationTest` and `DhisControllerConvenienceTest` classes 😢.

## Next

* Merge `DhisMockMvcControllerTest` and `DhisControllerTestBase` as we only use `DhisControllerTestBase`.
* Can we create a common integration test class that the controller tests are based on? Web API tests should need what the service based integration tests need with some extra web stuff.
* Can we move the H2 function registration into the H2DhisConfiguration? Right now these two are separated but should be closely related.
* `DhisControllerWithApiTokenAuthTest` needs more work as tests fail if I let `DhisControllerTestBase` do the user setup. Needs help from @netroms.


jbee (approved):
:tada:

david-mackessy (approved):

PR #18109, opened 2024-07-17
fix: Remove clear session log error in tests

enricocolasante:


dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/BaseSpringTest.java
```diff
@@ -75,11 +75,6 @@ protected void tearDownTest() throws Exception {}
   protected void nonTransactionalAfter() throws Exception {
     clearSecurityContext();
     tearDownTest();
-    try {
-      dbmsManager.clearSession();
```
teleivo, line 79: should this https://github.com/dhis2/dhis2-core/blob/358c5de15ee58b383abafee5df06477e02cc3bdc/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/test/integration/TransactionalIntegrationTest.java#L82 also be removed?

enricocolasante: I think so

vietnguyen (approved):

teleivo (approved):

PR #18110, opened 2024-07-17
fix: embedded Jetty failing to start

netroms:
## Summary
Fixes a regression that caused the embedded Jetty to fail during startup.
A profile is applied when starting Jetty to selectively add the request event listener.

### Automatic test
AuthTest (will start embedded Jetty on an available port and run login tests)

### Manual test
* Start the embedded Jetty server
* Observe you can log in


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DhisWebApiWebAppInitializer.java
```diff
@@ -100,6 +100,9 @@ public static void setupServlets(
     dispatcher.setLoadOnStartup(1);
     dispatcher.addMapping("/*");
 
+    //    RequestContextListener requestContextListener = new RequestContextListener();
+    //    context.addListener(requestContextListener);
```
david-mackessy, line 104: can these be removed?

jbee (approved):

david-mackessy (approved):

PR #18113, opened 2024-07-17
feat: add max upload file size validation

netroms:
## Summary
This PR adds a upload file size hard limit. There is a 2MB limit enforced in the UI, this adds the limit check to the back-end too. The new limit can be configured in dhis.conf with the `max.file_upload_size` setting. The default is 10MB.

### Automatic tests
FileResourceControllerTest

### Manual test
* Try to upload a file bigger than 3MB
* Observe you are not allowed to upload

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
@@ -83,8 +83,6 @@ public class FileResourceUtils {
   @Autowired private final JobRunner jobRunner;
   @Autowired private final FileResourceService fileResourceService;
 
-  public static final long MAX_FILE_UPLOAD_SIZE_BYTES = 3_000_000;
-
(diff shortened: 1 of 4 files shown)


vietnguyen (approved):

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/FileResourceUtils.java
```diff
@@ -282,7 +284,7 @@ private static void validateFileExtension(String fileName, List<String> validExt
     }
   }
 
-  private static void validateFileSize(@Nonnull MultipartFile file, long maxFileSizeInBytes) {
+  public static void validateFileSize(@Nonnull MultipartFile file, long maxFileSizeInBytes) {
```
david-mackessy, line 287: we could probably remove the file size param & just use the constant every time, unless other callers pass different sizes for different reasons?

david-mackessy: can see it's called in another PR with a different size. Can ignore that comment 👆

PR #18128, opened 2024-07-18
test: move NoOpFlywayConfiguration into dhis-support-test DHIS2-16557

teleivo:
so we do not include it into the dhis.war and risk using it in production.

No need to specify the bean name if the method has the expected name.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/config/FlywayConfig.java
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/config/FlywayConfig.java
@@ -34,7 +34,6 @@ import javax.sql.DataSource;
 import org.flywaydb.core.Flyway;
 import org.flywaydb.core.api.Location;
 import org.flywaydb.core.api.configuration.ClassicConfiguration;
-import org.hisp.dhis.db.migration.helper.NoOpFlyway;
 import org.hisp.dhis.external.conf.DhisConfigurationProvider;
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/config/FlywayConfig.java
```diff
@@ -49,7 +48,7 @@ public class FlywayConfig {
 
   private static final String FLYWAY_MIGRATION_FOLDER = "org/hisp/dhis/db/migration";
 
-  @Bean(value = "flyway", initMethod = "migrate")
+  @Bean(initMethod = "migrate")
   @Profile("!test-h2")
```
enricocolasante, line 52: Will we or can we already remove `@Profile("!test-h2")` here?

enricocolasante (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/config/FlywayConfig.java
teleivo: I also want to get rid of it but want to understand our whole config better first. Make the `WebTestConfiguration` only build on a config from the service integration tests.

(I think it could already work as I define `NoOpFlywayConfig` before `FlywayConfig` in `WebTestConfiguration`.)

d-bernat (approved):

muilpp (approved):

PR #18132, opened 2024-07-18
fix: Adds missing translations for new integrity checks

jason-p-pickering:
We were missing a few translations for some data integrity checks which have been added recently


david-mackessy (approved):

jbee (approved):
Merci!

PR #18133, opened 2024-07-18
fix: MinIO - Retrieving and updating an App work correctly [DHIS2-17197]

david-mackessy:
# Issues when using MinIO for storage
1. Endpoint: `GET` `/dhis-web-{app}/index.html`
    Retrieving an `App` to display in the UI results in a blank page (`200` response with `Content-Length=0`)

2. Endpoint: `POST` `api/appHub/{appId}`
    When updating an `App` to a different version, the process failed with errors:
```text
Failed to install app: error deleting directory apps/capture_100.70.9// in containerName: dhis2
Object name contains unsupported characters
```

# Causes
1. `Content-Length=0` when resource is of type `URL` (`File` works fine)
3. extra `/` in path. The blob storage does not need to call `JCloudStore#deleteDirectory`. The code to remove a blob is already there to use.

# Fixes
1. Resolve the `Content-Length` when the resource is a `URL`  (`File` resource behaviour is unchanged)
2. Branched the existing code for deleting directories/blobs which was used for all storage types, into
    a. File (deletes the filesystem directory & its contents)
    b. Blob (removes the blob, no longer calls `JCloudStore#deleteDirectory` which was the origin of the error) 

# Testing
## Automated
Added MinIO integration tests
- install and update `App`
- get `URL` resource content length

Added some extra provider unit tests also.

## Manual
Tested using the following scenario
- Install Capture `App` (`latest`)
- Install Capture `App` (`100.70.6`)
- Verify Capture `App` opens (index.html received with `Content-Length` greater than 0)
- previous Capture `App` version now removed from storage

Tested using MinIO and default filesystem storage.

# Info
UrlConnection uses caching by default.

Add MinIO TestContainer config.
To use just add `@ContextConfiguration(classes = {MinIOConfiguration.class})` to your integration test.

# Next
File system deletion of directory uses a deprecated method (`BlobStore#deleteDirectory`). Not a quick fix, will look to see if we can use something else in a separate PR chore.


dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/config/MinIOConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Use this configuration for tests relying on MinIO storage running in a Docker container. e.g. add
+ * to test class like `@ContextConfiguration(classes = {MinIODhisConfiguration.class})`
+ *
+ * @author david mackessy
+ */
+@Configuration
+public class MinIOConfiguration {
```
jbee, line 41: Hm... is a separate configuration the way to go? Wouldn't you use profiles to switch between this and the alternatives? If so I'd think this can be part of a general configuration.

jbee (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/AppOverrideFilter.java
```diff
@@ -145,7 +145,11 @@ private void serveInstalledAppResource(
         response.setContentType(mimeType);
       }
 
-      response.setContentLength((int) resource.contentLength());
+      // we need to handle scenarios when the Resource is a File (knowing the content length)
+      // or when it's URL (not knowing the content length and having to make a call, e.g. remote web
+      // link in AWS S3/MinIO) - otherwise content length can be set to 0 which causes issues at
+      // the front-end, returning an empty body.
+      response.setContentLength(appManager.getUriContentLength(resource));
       response.setHeader("ETag", etag);
```
vietnguyen, line 153: - Correct me if I'm wrong but if this `AppOverrideFilter` is triggered for every requests to the server as defined below, then the `getUricontentLength` would be called for every single request.  Do we have a cache mechanism for this value ? Maybe `UrlConnection.useCaches` ? 
```
context
        .addFilter("AppOverrideFilter", new DelegatingFilterProxy("appOverrideFilter"))
        .addMappingForUrlPatterns(null, true, "/*");
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -395,6 +395,22 @@ public Resource getAppResource(App app, String pageName) throws IOException {
...
 
+  /**
+   * @param resource resource to check content length
+   * @return the content length or -1 (unknown size) if exception caught
+   */
+  @Override
+  public int getUriContentLength(Resource resource) {
+    try {
+      URLConnection urlConnection = resource.getURL().openConnection();
+      return urlConnection.getContentLength();
```
vietnguyen, line 406: There is a `getContentLong()` method. Are we using `int` on purpose ?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/AppOverrideFilter.java
david-mackessy: Nice callout. The URL connection uses caching by default 👍
I can update the flow so that we only make the connection if it's not a File though, which might be better (and what the current behaviour is)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
david-mackessy: As with most things, you're hoping the initial implementation chose `int` for a reason :) 
I'd hope the App would never need to return a resource bigger than `Integer.MAX_VALUE` (~2GB)

vietnguyen (approved):

jbee (approved):

dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/config/MinIOConfiguration.java
david-mackessy: small change made so `MinIOConfigurationProvider` is independant. This extra config is still needed.

PR #18138, opened 2024-07-19
fix: Add test for changelogs when event is deleted [DHIS2-17743]

enricocolasante:
I just added a test to confirm that `EventChangeLogService.getEventChangeLog` is correctly validating if the event is deleted.
Deprecated `TrackedEntityDataValueChangeLogService.getTrackedEntityDataValueChangeLogs` is returning changelogs of soft deleted events. 
This is why we need to remove changelogs from the DB when we soft delete an event.
We can stop removing changelogs when soft deleting an event when the deprecated service is removed.


muilpp:
If I understand this correctly, as soon as we remove `TrackedEntityDataValueChangeLogService.getTrackedEntityDataValueChangeLogs`, we won't delete the change logs of a deleted event?

enricocolasante:
> If I understand this correctly, as soon as we remove `TrackedEntityDataValueChangeLogService.getTrackedEntityDataValueChangeLogs`, we won't delete the change logs of a deleted event?

Of a soft deleted event.
Right now when soft deleting we are removing some stuff linked to an entity and we are not removing some others. It is quite confusing.

To be honest, I hope we could get rid of soft deletion itself pretty soon so we will just delete the entity and everything linked to it

muilpp (approved):

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -139,7 +139,7 @@ public TrackerTypeReport deleteEvents(List<String> events) {
       deleteRelationships(relationships);
 
       // This is needed until deprecated method
-      // TrackedEntityDataValueChangeLogService.getTrackedEntityDataValueChangeLogs is not removed.
+      // TrackedEntityDataValueChangeLogService.getTrackedEntityDataValueChangeLogs is  removed.
```
teleivo, line 146: Double whitespace

teleivo (approved):

PR #18139, opened 2024-07-19
chore: Remove method getByPrograms(List<Program>) [DHIS2-17712]

muilpp:
This PR continues the work of moving tracker-related code out of the API module. Specifically it:

- Removes the `getByPrograms` method from `EnrollmentStore` and its implementation. The logic to get event programs from the database is now a private method in `EnrollmentSupplier`, since we need to access all enrollments regardless of user authorization and don't need/want this method in our exporters.
- Removes the `countEnrollments` method from `EnrollmentStore` and makes it private in its implementation, as it's only used internally.
- Removes a unit test from `EnrollmentSupplierTest` because it can't be properly mocked anymore. The same behavior is tested in `EventImportValidationTest.testTrackerAndProgramEventUpdateSuccess`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
@@ -28,17 +28,14 @@
 package org.hisp.dhis.tracker.imports.preheat.supplier;
 
 import java.util.List;
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.persistence.EntityManager;
 import javax.persistence.TypedQuery;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.program.Enrollment;
-import org.hisp.dhis.program.Program;
-import org.hisp.dhis.program.ProgramStore;
-import org.hisp.dhis.program.ProgramType;
 import org.hisp.dhis.tracker.imports.domain.TrackerObjects;
 import org.hisp.dhis.tracker.imports.preheat.TrackerPreheat;
+import org.hisp.dhis.tracker.imports.preheat.mappers.EnrollmentMapper;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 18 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
```diff
@@ -73,4 +70,13 @@ public void preheatAdd(TrackerObjects trackerObjects, TrackerPreheat preheat) {
           });
     }
   }
+
+  private List<Enrollment> getEnrollmentsByProgram(List<Program> programs) {
+    TypedQuery<Enrollment> query =
+        entityManager.createQuery(
+            "FROM Enrollment e WHERE e.deleted = false AND e.program IN (:programIds)",
```
teleivo, line 77: Can we settle on using lowercase in our SQL 😋 it's more comfortable to type and does not look like screaming 😂

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
```diff
@@ -60,11 +59,9 @@ public void preheatAdd(TrackerObjects trackerObjects, TrackerPreheat preheat) {
     if (programsWithoutRegistration.isEmpty()) {
       programsWithoutRegistration = programStore.getByType(ProgramType.WITHOUT_REGISTRATION);
     }
+
     if (!programsWithoutRegistration.isEmpty()) {
-      List<Enrollment> enrollments =
-          DetachUtils.detach(
```
teleivo, line 65: Is that not still needed? or are we not getting a proxied enrollment with the `EntityManager`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
```diff
@@ -73,4 +70,13 @@ public void preheatAdd(TrackerObjects trackerObjects, TrackerPreheat preheat) {
           });
     }
   }
+
+  private List<Enrollment> getEnrollmentsByProgram(List<Program> programs) {
```
enricocolasante, line 74: This is not strictly related to this PR but could we make the query just get all the enrollments with programs without registration instead of getting the programs and passing the ids?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EnrollmentSupplier.java
enricocolasante: Agreed. I think we still need to detach the enrollments.

muilpp: yep, nice catch!

enricocolasante (approved):

teleivo (approved):

PR #18142, opened 2024-07-19
fix: repeated fields in TE query [DHIS2-17745]

gnespolino:
This PR addresses fields that are sometimes repeated in TE analytics queries.
Moreover, it fixes the JSON aggregation query in TE analytics queries, which unnecessarily uses non-analytics tables.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/query/Field.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/query/Field.java
@@ -69,7 +69,7 @@ public class Field extends BaseRenderable {
   /** virtual fields won't be added to the select clause */
   @With @Getter private final boolean virtual;
 
-  // a cached version of the rendered field
+  // A cached version of the rendered field.
   private String renderedField;
 
   public Field asVirtual() {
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/query/Field.java
```diff
@@ -63,6 +69,9 @@ public class Field extends BaseRenderable {
   /** virtual fields won't be added to the select clause */
   @With @Getter private final boolean virtual;
 
+  // a cached version of the rendered field
```
maikelarabori, line 72:
```suggestion
  // A cached version of the rendered field.
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/context/sql/RenderableSqlQuery.java
```diff
@@ -141,7 +145,20 @@ private String select() {
   }
 
   private List<Field> nonVirtualSelectFields() {
-    return selectFields.stream().filter(f -> !f.isVirtual()).toList();
+    return selectFields.stream()
+        .filter(Predicate.not(Field::isVirtual))
+        // we need to filter out fields that have already been rendered;
```
maikelarabori, line 150:
```suggestion
        // We need to filter out fields that have already been rendered.
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/tei/query/context/sql/RenderableSqlQuery.java
```diff
@@ -141,7 +145,20 @@ private String select() {
...
+        // it might happen that a field can be added multiple times (for example
```
maikelarabori, line 151:
```suggestion
        // It might happen that a field can be added multiple times (for example
```

maikelarabori (approved):
Looks good, just minor comments.
Please, add a unit test for the duplicated fields as we discussed. 
Thanks!

d-bernat (approved):

PR #18144, opened 2024-07-20
feat: OpenAPI HTML rendering and API

jbee:
### Summary
Adds HTML rendering capabilities to to the OpenAPI package and an API that makes DHIS OpenAPI available as HTML documents.

This is essentially a new product that allows to turn an OpenAPI JSON specification to a stand alone HTML document viewer. This is indeed a static transformation function. 

The second part is using this function to offer a DHIS2 specific OpenAPI API that allows to view DHIS2 OpenAPI specification which is generated by the server itself.

### Work items
This is new product development so this keeps track of work items identified and provided a summary of sorts to document the individual steps taken.

#### Generation work items
- [x] fix: `$path` should not have `/api/...` part in the name, e.g. not `apidataElement` but `dataElements`
- [x] feat: add `@OpenApi.Kind` annotation and grouping component schemas by kind
- [x] QoI: Gist object view should be typed `@EntityType` not `object`
- [x] QoI: Gist object list view should be typed `@EntityType[]` not `object`
- [x] fix: Gist `fields` which now uses `ParameterNames` strangely sometimes substitutes `@EntityType` to the wrong type :thinking: 
- [x] feat: new annotation `@OpenApi.Shared.Inline` can be used to exclude parameters from a shared object (this is needed to fix the issue with `ParameterNames` because the type of such a field is varying)
- [x] fix: component schema list should not contain `JsonTypedAccessStore` (e.g. `JobConfiguration` domain)
- [x] Qoi: model "unspecified JSON" as `any`
- [x] fix: analysis should never extract schemas for spring views and alike
- [x] feat: adds `object` to `@OpenApi.Response` and use it for `JsonRoot` usages
- [x] refactor: replace `asProperty` in `@OpenApi.Param` with a similar `object` attribute

#### Rendering work items
- [x] feat: URL parameter to specify the maximum number of enum values up until which the values are inlined in schema breakdown
- [x] QoI: objects with single property show special signature (wrapper)
- [x] QoI: objects with header + list show special signature (envelope)
- [x] HTML-escape all property names and attribute names where needed   
- [x] test: write a unit test for escape and skip HTML utils 
- [x] QoI: add response mime+type summary 
- [x] QoI: show HTTP method count breakdown in package group summary
- [x] fix: markdown parsing - support lists
- [x] test: more test for markdown processing
- [x] feat: add a "with source" link to the operation package/domain level
- [x] chore: introduce a params object for the general OpenAPI controller params
- [x] feat: simple cache for the full document
- [x] feat: ETag for OpenAPI controller (all media types) + parameter to bypass during development

#### UI work items (JS/client side stuff)
- [x] feat: add UI hotkeys for the URL hash part as a form of navigation history
- [x] feat: add UI back-references in JS that finds all usages of the component schema

### Automated Tests
New unit tests were added for the HTML escape and strip utilities as well as the markdown parsing and conversion to HTML.
Added one controller test for HTML generation just to do one full circle test.

### Manual Tests
The main OpenAPI rendering is best tested manually as there are countless details.

The fill document can be viewed at `/openapi/openapi.html`
Partial documents can be generated using the `domain` parameter, e.g. `/openapi/openapi.html?domain=DataElement`.

### Next
- [ ] QoI: allow specifying the key's schema of map objects as `x-keys` (they are strings but e.g. are required to be numbers etcetera) 
- [ ] fix: consider read-only quality of object properties (is there a JSON schema property?, exclude properties from input schema, mark read-only properties in UI etcetera)  
- [ ] refactor: make `@OpenApi.Document#group` a  `String` that also can have a description
- [ ] refactor: undo `@OpenApi.Shared#pattern` in favor of always using `name` directly
- [ ] feat: add a `@OpenApi.Since` annotation to allow stating when an operation was added
- [ ] feat: add `@OpenApi.Team` annotation and propagate the info in the model
- [ ] feat: add UI support for teams information
- [ ] feat: markup "keywords" - support map to specify a link target (hash)
- [ ] fix: `@OpenApi.Shared.Inline`d properties pick up descriptions from .md in addition to the annotated one
- [ ] fix: a parameter annotated as `Period` becomes a `UID_null` string type ;(
- [ ] feat: add UI OpenAPI tags as filters
- [ ] fix: `AnalyticsDimensionsPagingWrapper` becomes `?` because it has a generic (make it object?)
- [ ] feat: allow big slices of the API via a single selector, e.g. tags like "tracker" gives all tracker API operations
- [ ] feat: domain navigation drop-down to jump to other single domain pages
- [ ] chore: once tags are used as filters no longer add the group as tag to the JSON spec
- [ ] feat: make the operations accessible as "guest" user (no login)?!
- [ ] feat: default to `String` for mediaTypes that are not JSON if the method return type is `void`
- [ ] feat: introduce a `@OpenApi.Type` annotation to put on a type to alias it to another type? This would e.g. be useful to alias an interface to the base class implementation where we have the properties declared. the alternative would be to annotate all endpoints using interfaces individually which is more brittle and more work and more code


vietnguyen (approved):

david-mackessy (approved):
nice work 👍

PR #18157, opened 2024-07-22
chore: Remove getters by entity from RelationshipService [DHIS2-17713]

enricocolasante:
## Big picture

## This PR
Remove `getRelationshipBy` getters from `RelationshipService` that were used in tests and in `TrackerObjectDeletionService`, use `RelationshipStore` from tracker instead.
We cannot use relationship exporter here as we need to get and delete relationships that can potentially be inaccessible to the user. In the importer there is a validation to check the `CASCADE` authority that allows to delete everything under the entity being deleted.


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/EnrollmentServiceTest.java
```diff
@@ -65,6 +66,11 @@
  */
 @Transactional
 class EnrollmentServiceTest extends PostgresIntegrationTestBase {
+  private static final UID ENROLLMENT_A_UID = UID.of(CodeGenerator.generateUid());
```
teleivo, line 69: are we only accessing these uids as strings in here? can't see it form the diff 🤔 if yes can we keep them as strings to get rid of the `getValue()` we do all the time in here

teleivo (approved):

muilpp (approved):

PR #18161, opened 2024-07-22
chore: Remove get by uids from RelationshipService [DHIS2-17713]

enricocolasante:
## Big picture

## This PR

Remove `Optional<Relationship> getRelationshipByRelationship(Relationship relationship);
` as it was not used anymore.
Move `List<Relationship> getRelationships(@Nonnull List<String> uids);` to tracker.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
@@ -30,6 +30,7 @@ package org.hisp.dhis.tracker.export.relationship;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import javax.annotation.Nonnull;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.feedback.ForbiddenException;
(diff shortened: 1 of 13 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -90,6 +91,16 @@ public Relationship getRelationship(String uid) throws ForbiddenException, NotFo
     return map(relationship);
   }
 
+  @Override
+  public List<Relationship> getRelationships(@Nonnull List<String> uids)
```
teleivo, line 95: Will we use our UID type in a separate PR?

Should we create a task for fetching relationships for uids all at once instead of one at a time?

teleivo (approved):

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
enricocolasante: Yes and yes.
I think when I manage to remove `RelationshipService` and `RelationshipStore` from the core, then I will review our tracker service and store.

PR #18163, opened 2024-07-22
chore: Create event program service and store [DHIS2-17712]

muilpp:

- Removes the `hardDeleteEnrollment` method from `EnrollmentService` and `EnrollmentStore`.
- Creates a new service and store called `EventProgramService` and `EventProgramStore`, respectively. Their goal is to handle CRUD operations on event programs.
- Ensures that `EventProgramEnrollmentDeletionHandler` deletes enrollments of event programs only.


muilpp:
will be done here: https://github.com/dhis2/dhis2-core/pull/18166

PR #18164, opened 2024-07-22
chore: Remove get relationship by type from RelationshipService [DHIS2-17713]

enricocolasante:
## Big picture

## This PR

`getRelationshipsByRelationshipType` was only used in deletion handlers and it was replaced with a SQL query, that is also ore performant as it is not getting all the relationships of a specific relationshipType.


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/relationship/RelationshipDeletionHandlerTest.java
```diff
@@ -0,0 +1,87 @@
...
+@Transactional
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class RelationshipDeletionHandlerTest extends PostgresIntegrationTestBase {
+  private RelationshipType relationshipType;
+
+  private RelationshipType deletableRelationshipType;
+
+  private Relationship relationship;
+
+  @Autowired private IdentifiableObjectManager identifiableObjectManager;
```
teleivo, line 54:
```suggestion
  @Autowired private IdentifiableObjectManager manager;
```

is usually what we name it

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/relationship/RelationshipDeletionHandlerTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Autowired private DbmsManager dbmsManager;
```
teleivo, line 56: that is unused

teleivo (approved):

teleivo (approved):

muilpp (approved):

PR #18166, opened 2024-07-22
chore: Remove hardDeleteEnrollment [DHIS-17712]

muilpp:
- Removes the hardDeleteEnrollment method from EnrollmentService and EnrollmentStore.
- Removes the ceremony around deleting an event program enrollment.


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/EnrollmentDeletionHandler.java
```diff
@@ -60,15 +58,10 @@ private DeletionVeto allowDeleteProgram(Program program) {
...
-      while (iterator.hasNext()) {
-        Enrollment enrollment = iterator.next();
-        iterator.remove();
-        enrollmentService.hardDeleteEnrollment(enrollment);
-      }
+    if (program.isRegistration()) {
+      return;
     }
+
+    delete("delete from enrollment where programid = :id", Map.of("id", program.getId()));
```
teleivo, line 65: this is deleting dummy enrollments right?
is it correct that other enrollments into a program with registration are only deleted by trackers TrackerObjectDeletionService?

I probably got all of the above wrong which suggests we should document this at least in a comment here.

teleivo (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/EnrollmentDeletionHandler.java
muilpp: you've got it right, but I can add a comment anyway

teleivo: 😅 yes please

teleivo (approved):

muilpp:
The api test fails to build the docker image continuously, for some reason, the same PR [here](https://github.com/dhis2/dhis2-core/pull/18176) works, so I close this one

PR #18168, opened 2024-07-22
chore: Remove getIncludingDeleted from RelationshipStore [DHIS2-17713]

enricocolasante:
## Big picture

## This PR

Implement `getByUidsIncludeDeleted` directly in `RelationshipStrategy`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/RelationshipStrategy.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/RelationshipStrategy.java
@@ -52,12 +52,8 @@ public class RelationshipStrategy extends HibernateGenericStore<Relationship>
     implements ClassBasedSupplierStrategy {
 
   public RelationshipStrategy(
-      EntityManager entityManager,
-      JdbcTemplate jdbcTemplate,
-      ApplicationEventPublisher publisher,
-      Class<Relationship> clazz,
-      boolean cacheable) {
-    super(entityManager, jdbcTemplate, publisher, clazz, cacheable);
+      EntityManager entityManager, JdbcTemplate jdbcTemplate, ApplicationEventPublisher publisher) {
+    super(entityManager, jdbcTemplate, publisher, Relationship.class, false);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/RelationshipStrategy.java
```diff
@@ -27,30 +27,42 @@
...
+    value = org.hisp.dhis.tracker.imports.domain.Relationship.class,
+    mapper = RelationshipMapper.class)
+public class RelationshipStrategy extends HibernateGenericStore<Relationship>
+    implements ClassBasedSupplierStrategy {
+
+  public RelationshipStrategy(
+      EntityManager entityManager,
+      JdbcTemplate jdbcTemplate,
+      ApplicationEventPublisher publisher,
+      Class<Relationship> clazz,
```
teleivo, line 58: Is this argument always the same for this class? if so can we inline that and remove one constructor param

teleivo (approved):
is this covered by a test?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/RelationshipImportTest.java
```diff
@@ -107,4 +108,20 @@ void successUpdateRelationships() throws IOException {
...
+    TrackerObjects trackerObjects = fromJson("tracker/relationships.json");
+    trackerImportService.importTracker(trackerImportParams, trackerObjects);
+
+    manager.delete(manager.get(Relationship.class, "Nva3Xj2j75W"));
+
+    trackerImportParams.setImportStrategy(TrackerImportStrategy.CREATE_AND_UPDATE);
+    ImportReport importReport =
+        trackerImportService.importTracker(trackerImportParams, trackerObjects);
+    assertHasError(importReport, ValidationCode.E4017);
+    assertThat(importReport.getStats().getIgnored(), is(2));
```
teleivo, line 125: why is the number 2?
we decided to stick to JUnit assertions instead of hamcrest https://github.com/dhis2/wow-backend/blob/master/guides/testing/testing.md#assertions

PR #18171, opened 2024-07-22
chore: Remove deleteEnrollment [DHIS-17712]

muilpp:

- Removes the `deleteEnrollment` method from `EnrollmentService`. We can use the manager to achieve the same result.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -113,7 +113,11 @@ public TrackerTypeReport deleteEnrollments(List<String> enrollments) throws NotF
       TrackedEntity te = enrollment.getTrackedEntity();
       te.setLastUpdatedByUserInfo(userInfoSnapshot);
 
-      apiEnrollmentService.deleteEnrollment(enrollment);
+      enrollment.setStatus(EnrollmentStatus.CANCELLED);
```
teleivo, line 116: Is something similar needed for events which also has a status?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -113,7 +113,11 @@ public TrackerTypeReport deleteEnrollments(List<String> enrollments) throws NotF
...
+      enrollment.setDeleted(true);
+      enrollment.setLastUpdated(new Date());
+      manager.update(enrollment);
```
teleivo, line 119: Why are you not using `manager.delete` like we do for the other entities?

enricocolasante: I think this is too many steps here.
We can add the `delete` method to our `EnrollmentStore` and make it change the status and call the super.delete

enricocolasante: And then we should use manager.delete here and in the tests.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
muilpp: The PR to refactor the events deletion is this one: https://github.com/dhis2/dhis2-core/pull/18010/files
Doesn't look like we were using the status before either, I don't know why though.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/DefaultTrackerObjectsDeletionService.java
```diff
@@ -113,7 +114,8 @@ public TrackerTypeReport deleteEnrollments(List<String> enrollments) throws NotF
       TrackedEntity te = enrollment.getTrackedEntity();
       te.setLastUpdatedByUserInfo(userInfoSnapshot);
 
-      apiEnrollmentService.deleteEnrollment(enrollment);
+      enrollmentStore.deleteEnrollment(enrollment);
```
enricocolasante, line 117: You could also use `manager.delete` here. In the end it will call the exact same method.

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
```diff
@@ -260,7 +262,8 @@ void testDeleteSoftDeletedEnrollmentWithAProgramMessage()
     manager.save(enrollment);
     programMessageService.saveProgramMessage(message);
     assertNotNull(manager.get(Enrollment.class, enrollment.getUid()));
-    apiEnrollmentService.deleteEnrollment(enrollment);
+
+    trackerObjectDeletionService.deleteEnrollments(List.of(enrollment.getUid()));
```
teleivo, line 266: is this what we want @enricocolasante ? should we not simply use the `manager` here? who cares about the `status`. I would rather not add a dependency from here to the deletionService.

enricocolasante: The `status` will be correctly updated using the `manager` as in the end the correct store will be used.
The difference is that using `TrackerObjectDeletionService` will delete enrollments and all linked events and relationships. 
Anyway I think it is better to use the `manager` and manually delete any linked entity.

teleivo (approved):

PR #18178, opened 2024-07-23
test: move injectAdminUser into SpringIntegrationTestBase DHIS2-17768

teleivo:
This change is important so I can make the `ControllerTestBase` extend the `SpringIntegrationTestBase`.

To everyone:

The well known user with username `admin` (UID `M5zQapPyTZI`) and role `Superuser` (UID `yrB6vc5Ip3r`)
* is now created and injected into the Spring security context by [SpringIntegrationTestExtension](https://github.com/dhis2/dhis2-core/blob/19979149e4dfd56d5fffe6018f6f32de576ce2a0/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/junit/SpringIntegrationTestExtension.java#L184) 
* you can access it in integration tests via [getAdminUser](https://github.com/dhis2/dhis2-core/blob/408a73fcb5e36be647d2ca3cacec90cc371c2244/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/SpringIntegrationTestBase.java#L49)
* you can also inject it back into the security context if you changed user using [injectAdminIntoSecurityContext](https://github.com/dhis2/dhis2-core/blob/408a73fcb5e36be647d2ca3cacec90cc371c2244/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/SpringIntegrationTestBase.java#L52)

**For the future:** every test should create a dedicated user so the tests are less brittle

To each team:

review only changes to your tests and the TestBase if you like. You might find a `// TODO(DHIS2-17768 teamname)` I added to a test were I think you tested under the wrong assumption. Please take the time to fix this after we merged the changes.

## To team platform

I added some `TODO(DHIS2-17768 platform)`s
* disabled a few test as you will be way faster in fixing its setup

Note that gist controller tests assert using some magic numbers. I decremented the ones that caused test failures as the `ControllerTestBase` now only creates the above `admin` and not also another super user.

## To team tracker

I changed the UIDs and unique constraint names in our json fixtures. So our tests now have their own `trackeradmin` user that does the metadata import. I did it so I could keep the organization unit setup as it is. We should probably remove most authorities from this user so we test with a non-superuser.

Search for `TODO(DHIS2-17768 tracker)`
* disabled one test class - I hope someone else can take on enabling it again 😫


jbee (approved):

enricocolasante (approved):

jbee:
:tada:

PR #18179, opened 2024-07-23
chore: Remove method getWithScheduledNotifications [DHIS2-17712]

muilpp:

- Removes the `getWithScheduledNotifications` method from `EnrollmentStore`.
- Creates a new method in `ProgramNotificationService` to get the aforementioned enrollments. It's marked as deprecated to find a better way to do it when we are done with this epic. This method should be private.


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/EnrollmentStoreTest.java
```diff
@@ -244,16 +247,16 @@ void testGetWithScheduledNotifications() {
     // Queries
     List<Enrollment> results;
     // A
-    results = enrollmentStore.getWithScheduledNotifications(a1, today);
+    results = programNotificationService.getEnrollmentsWithScheduledNotifications(a1, today);
```
teleivo, line 250: not sure what is tested here but should this not move as well?

muilpp: yes, this should be moved and some point and the method be made private

enricocolasante: I think `testGetWithScheduledNotifications` could be moved to `ProgramNotificationServiceTest`.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationServiceTest.java
```diff
@@ -367,4 +376,67 @@ void eventDataValuesSearchStringTest() {
...
+    Date today = cal.getTime();
+    cal.add(Calendar.DATE, 1);
+    Date tomorrow = cal.getTime();
+    cal.add(Calendar.DATE, -2);
+    Date yesterday = cal.getTime();
+    cal.add(Calendar.DATE, -6);
+    Date aWeekAgo = cal.getTime();
+    // Enrollments
+    Enrollment enrollmentA = new Enrollment(today, tomorrow, trackedEntityX, programA);
+    enrollmentStore.save(enrollmentA);
```
teleivo, line 424: you should not introduce dependencies to our stores as we want to make them private. can you not save using the manager?

enricocolasante (approved):

teleivo (approved):

PR #18187, opened 2024-07-24
chore: Move note service into tracker module [DHIS2-17794]

enricocolasante:
`Note` persisting was working only because there was a bug in the code 
[`if (Objects.isNull(preheat.getNote(note.getUid()))) {
    entityManager.persist(note);
}`](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java#L94-L96)
`preheat.getNote` is returning an Optional, so the object is never null and the persist is never performed.

The actual persisting is happening when persisting the event.

This PR is removing all the unnecessary store and service for `Note`, remove the buggy `peristNotes` from the `AbstractTrackerPersister` and move the still needed store in tracker module.
We still need (for now?) an empty `HibernateNoteStore` in order to be able to use the manager to get and persist notes in the tests.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/note/HibernateNoteStore.java
```diff
@@ -39,22 +38,13 @@
 /**
  * @author David Katuscak
  */
-@Repository("org.hisp.dhis.trackedentitycomment.NoteStore")
-public class HibernateNoteStore extends HibernateIdentifiableObjectStore<Note>
-    implements NoteStore {
+@Repository("org.hisp.dhis.tracker.export.note.NoteStore")
+public class HibernateNoteStore extends HibernateIdentifiableObjectStore<Note> {
```
teleivo, line 42: can we make this package private?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/NoteStrategy.java
```diff
@@ -27,32 +27,60 @@
...
+            .flatMap(Collection::stream)
+            .filter(CodeGenerator::isValidUid)
+            .collect(Collectors.toSet());
+
+    preheat.addNotes(getExistingNotes(uids));
+  }
+
+  private Set<String> getExistingNotes(Set<String> uids) {
+    Set<String> notes = new HashSet<>();
+    List<List<String>> uidsPartitions = Lists.partition(Lists.newArrayList(uids), 20000);
```
teleivo, line 72: can we use `List.of` instead of the google lib

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -97,7 +97,7 @@ public static List<Note> validateNotes(
       {
         // If a note having the same UID already exist in the db, raise
         // warning, ignore the note and continue
-        if (isNotEmpty(note.getNote()) && preheat.getNote(note.getNote()).isPresent()) {
+        if (isNotEmpty(note.getNote()) && preheat.noteExists(note.getNote())) {
```
teleivo, line 100: not sure if we stick to it but usually we use `is/has` prefixes for such methods, maybe `preheat.hasNote`?

teleivo (approved):

muilpp (approved):

PR #18188, opened 2024-07-24
chore: Remove getEnrollments(Program) [DHIS2-17712]

muilpp:

- Removes the getEnrollments(Program) method from the `EnrollmentService`.
- Creates a new service and store to deal with event programs enrollments. These two are meant to be used as a temporary solution until event programs are completely separated from tracker programs. At that point, both classes should be removed.

As a side note, we've found out the `CopyService` is actually copying enrollments from tracker programs. For now, we don't change its logic, but that's something worth of a discussion when the appropriate people are back from holidays.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEventProgramEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEventProgramEnrollmentService.java
@@ -38,7 +38,8 @@ public class DefaultEventProgramEnrollmentService implements EventProgramEnrollm
   private final EventProgramEnrollmentStore eventProgramEnrollmentStore;
 
   @Override
-  // TODO The CopyService is currently copying enrollments from tracker programs. We need to discuss
+  // TODO(tracker) The CopyService is currently copying enrollments from tracker programs. We need
+  // to discuss
   // if that's correct. Seems to make more sense to copy enrollments from event programs only.
   public List<Enrollment> getEnrollments(Program program) {
     return eventProgramEnrollmentStore.get(program);
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEventProgramEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEventProgramEnrollmentService.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2004-2024, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/EnrollmentServiceTest.java
```diff
@@ -242,11 +244,11 @@ void testGetEnrollmentsByProgram() {
     manager.save(enrollmentA);
     manager.save(enrollmentB);
     manager.save(enrollmentD);
-    List<Enrollment> enrollments = apiEnrollmentService.getEnrollments(programA);
+    List<Enrollment> enrollments = eventProgramEnrollmentService.getEnrollments(programA);
```
enricocolasante, line 247: As the previous PR, I think it make sense to create `EventProgramEnrollmentServiceTest` to test this method.
This test class should be removed when this Jira task is done, but we will need more time to solve the program/tracker event separation.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultEventProgramEnrollmentService.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service("org.hisp.dhis.program.DefaultEventProgramEnrollmentService")
+public class DefaultEventProgramEnrollmentService implements EventProgramEnrollmentService {
+
+  private final EventProgramEnrollmentStore eventProgramEnrollmentStore;
+
+  @Override
+  // TODO The CopyService is currently copying enrollments from tracker programs. We need to discuss
```
teleivo, line 41: Please use `TODO(tracker)` or `TODO(issue nr)` as such todos are hard to find and todos without owners have even less chances of getting done 😁

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/EventProgramEnrollmentServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+    user.setTeiSearchOrganisationUnits(Set.of(organisationUnitA, organisationUnitB));
+    user.setOrganisationUnits(Set.of(organisationUnitA));
+
+    injectSecurityContextUser(user);
+  }
+
+  @Test
+  void testGetEnrollmentsByProgram() {
+    List<Enrollment> enrollments = eventProgramEnrollmentService.getEnrollments(programA);
+    assertEquals(2, enrollments.size());
```
teleivo, line 113: please use `assertContainsOnly` or `assertEquals(List.of(), enrollments)` if order matters

this style of assertions are a pain if they fail on CI which is why we should stop writing them this way 😬

teleivo (changes requested):

teleivo (approved):

PR #18196, opened 2024-07-25
refactor: Remove unused methods and add Mapper and validation

zubaira:
**Description:** 
This PR primarily focuses on refactoring the `ProgramMessage` functionality, which involves updates to the controller, services, and repository layers.
Identified a bug in `DELETE /api/messages/{uid}`. Fixing bug in this PR because correct deletion is pre-requisite for e2e tests.

**Changes:**

-   **Controller Updates:**
    
    -   `ProgramMessageController` now extends from `AbstractFullReadOnlyController`.
    -   Added `PUT` and `DELETE` endpoints to `ProgramMessageController`.
-   **Service Updates:**
    
    -   Refactored `ProgramMessageService` to remove unused methods.
-   **Validation and Mapping:**
    
    -   Created parameter validation logic.
    -   Implemented mappers for `ProgramMessage`.
-   **Tests:**
    
    -   Added new end-to-end (e2e) tests for `ProgramMessage`.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParamsMapper.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @author Zubair Asghar
+ */
+@Component
+@RequiredArgsConstructor
+public class ProgramMessageRequestParamsMapper {
+  private final IdentifiableObjectManager manager;
+
+  @Transactional(readOnly = true)
+  public ProgramMessageQueryParams ValidateAndMap(ProgramMessageRequestParams params)
```
enricocolasante, line 55:
```suggestion
  public ProgramMessageQueryParams map(ProgramMessageRequestParams params)
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParamsMapper.java
```diff
@@ -0,0 +1,109 @@
...
+      throws ConflictException, BadRequestException {
+    if (params == null) {
```
enricocolasante, line 57: `params` will never be null, this check can be removed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParamsMapper.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * @author Zubair Asghar
+ */
+@Component
+@RequiredArgsConstructor
+public class ProgramMessageRequestParamsMapper {
+  private final IdentifiableObjectManager manager;
+
+  @Transactional(readOnly = true)
```
enricocolasante, line 54: If you want to follow the pattern that we have for tracker exporters would be good to create an intermediate `ProgramMessageOperationParams` and then when mapping to `ProgramMessageQueryParams` you can access the DB to check if enrollment or event exists.

We should avoid accessing the DB in the web layer.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParamsMapper.java
```diff
@@ -0,0 +1,109 @@
...
+
+    Enrollment enrollment = null;
+    Event event = null;
+
+    if (enrollmentUid != null) {
+      enrollment =
+          Optional.ofNullable(manager.get(Enrollment.class, entity.getValue()))
+              .orElseThrow(
+                  () ->
+                      new IllegalQueryException(
```
enricocolasante, line 85: Would it make sense to be `NotFoundException` instead?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParams.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.NoArgsConstructor;
+import org.hisp.dhis.common.UID;
+import org.hisp.dhis.program.message.ProgramMessageStatus;
+
+/**
+ * @author Zubair Asghar
+ */
+@Data
+@NoArgsConstructor
+public class ProgramMessageRequestParams {
```
enricocolasante, line 42: It would be good to carry over the `@Open.API` annotations that were present in the endpoint before

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageController.java
```diff
@@ -0,0 +1,125 @@
...
+    for (ProgramMessage programMessage : batch.getProgramMessages()) {
+      programMessageService.validatePayload(programMessage);
+    }
+
+    return programMessageService.sendMessages(batch.getProgramMessages());
+  }
+
+  @RequiresAuthority(anyOf = F_MOBILE_SENDSMS)
+  @PutMapping(value = "/{uid}", produces = APPLICATION_JSON_VALUE)
+  public WebMessage updateProgramMessage(
```
enricocolasante, line 112: Before we were allowing changing every single field of a `ProgramMessage`, now just the `status`, are we sure nobody was updating other fields?

Anyway, we should either maintain the other endpoint and deprecate it or agree on introducing a breaking change

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageController.java
```diff
@@ -0,0 +1,125 @@
...
+      @PathVariable String uid, ProgramMessageUpdateRequest updatedRequest)
+      throws NotFoundException {
+    programMessageService.updateProgramMessage(uid, updatedRequest);
+    return ok(format("ProgramMessage with id %s updated", uid));
+  }
+
+  @RequiresAuthority(anyOf = F_MOBILE_SENDSMS)
+  @DeleteMapping(value = "/{uid}", produces = APPLICATION_JSON_VALUE)
+  public WebMessage deleteProgramMessage(@PathVariable String uid) throws NotFoundException {
```
enricocolasante, line 121:
```suggestion
  public WebMessage deleteProgramMessage(@PathVariable UID uid) throws NotFoundException {
```

We could directly use `UID` type so we are sure the format is correct

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/message/ProgramMessageUpdateRequest.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Zubair Asghar
+ */
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/** User input when updating {@link org.hisp.dhis.program.message.ProgramMessage}. */
+@Data
+@NoArgsConstructor
+public class ProgramMessageUpdateRequest {
```
enricocolasante, line 40: This should live in the web layer, not in `dhis-api` module

enricocolasante (changes requested):
I think it would be easier to review and much more clear to understand if this PR was broken down in more PRs.

At least we need one PR where we are just removing unused methods and then one or more PRs for all the other changes.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageController.java
zubaira: yes, ProgramMessage is an internal object created by `notificationServiceand`  and used by `messageSender`.
I hardly see any use case of changing `PrograMessage` fields other than `status`.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageRequestParamsMapper.java
zubaira: absolutely but it will change the API response.

zubaira:
Divided this pull request into several smaller ones, so I am closing this original PR.

PR #18197, opened 2024-07-25
fix: Add capture namespace protection [DHIS2-17718](2.40)

enricocolasante:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/CaptureApp.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * The main purpose (so far) of the {@link CaptureApp} component is to establish the protected
+ * {@link #NAMESPACE} in the {@link DatastoreService} so that only the app can write to it using a
+ * role having the {@link #AUTHORITY}.
+ */
+@Component
+public class CaptureApp {
+  public static final String NAMESPACE = "capture";
+
+  public static final String AUTHORITY = "F_capture";
```
eirikhaugstulen, line 44: Do you think we can change the authority to something like this?

```suggestion
  public static final String AUTHORITY = "F_CAPTURE_DATASTORE_UPDATE";
```

jbee: I guess can is really this should be the same as the one added in the other files.

.github/workflows/deploy-instance.yml
eirikhaugstulen, line 1: @jbee I guess this file should be reverted before merging?

eirikhaugstulen (approved):
👍

PR #18198, opened 2024-07-25
Fix disabled data integrity unit tests

jason-p-pickering:


teleivo (approved):
Is there a way to make this less prone to failure only because one user in our base test setup is created? 🤔 the expected number looks magical to me 😅

jason-p-pickering:
So basically these tests are adding new users. Previously there were two users for some reason, now there is just one. So maybe the better approach would be to actually determine how many existing users are present prior to adding any new ones? That might help to future proof against any changes in the base test classes. What do you think?

jbee (approved):

PR #18207, opened 2024-07-25
fix: Improve test clarity for user data integrity checks

jason-p-pickering:
Based on feedback on an earlier PR, it was suggested to clarify the intent of the tests, and specifically how the various percentages were being calculated.


jbee (commented):
You cannot (successfully) manually format the code as it is auto-formatted or expected to be in the auto-formatted format. The way to improve readability with that is usually to extract local variables so that intermediate results get a explaining name and the lines break down more. Also sometimes it might help to force a like break using `//` at the end of a line

david-mackessy (approved):

jbee (approved):

PR #18212, opened 2024-07-26
fix: (2.41) Notifications from program rules in old tracker (#17602)

enricocolasante:


zubaira (commented):
i think we might also need authentication for 
[this](https://github.com/dhis2/dhis2-core/blob/2eedec6092c80a17dbe01c2196700b2fa6fd9d15/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/RuleActionScheduleMessageImplementer.java#L107) as well.

zubaira (approved):

PR #18217, opened 2024-07-26
fix: Fix pending tests for tracker

enricocolasante:
Remove `@Disabled` annotation from `LastUpdateImportTest`. Different tests are injecting different users and all of them were actually assuming the admin user to be injected when starting the test, depending on the order that was not true for some of them.
To fix it I added a `@BeforeEach` annotated method that inject the admin user.

Then `testDisplayIncidentDateTrueButDateValueIsInvalid` and `testWrongScheduledDateString` were removed as they are not integration tests. They were "testing" that a not parseable payload would return an error.


muilpp (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EnrollmentImportValidationTest.java
```diff
@@ -119,14 +118,6 @@ void testPreheatOwnershipForSubsequentEnrollment() throws IOException {
             });
   }
 
-  @Test
-  void testDisplayIncidentDateTrueButDateValueIsInvalid() {
-    // TODO(DHIS2-17768 tracker) what are we testing here? should we not call the importer?
-    assertThrows(
-        IOException.class,
-        () -> fromJson("tracker/validations/enrollments_error-displayIncident.json"));
```
teleivo, line 127: is this still something we should test, just not here?

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EnrollmentImportValidationTest.java
enricocolasante: Yes, but this is not just tracker related and the error raised by this is not clear.
I think it can be discussed in the Thursady meeting.
i will add a point in the agenda

PR #18238, opened 2024-07-30
fix: use schema.getPersistedProperties() in resetNonOwnerProperties method

vietnguyen:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AbstractCrudController.java
```diff
@@ -923,18 +887,11 @@ private boolean isCompatibleWith(String type, MediaType mediaType) {
     return false;
   }
 
-  protected Patch diff(HttpServletRequest request) throws IOException, BadRequestException {
```
vietnguyen, line 926: This method is not in use.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/patch/DefaultPatchService.java
```diff
@@ -1,455 +0,0 @@
...
-import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Transactional;
-
-/**
- * @author Morten Olav Hansen <mortenoh@gmail.com>
- */
-@Service
-@RequiredArgsConstructor
-@Transactional // TODO not sure if this can be completely readonly
-public class DefaultPatchService implements PatchService {
```
vietnguyen, line 60: Moved to `dhis-service-dxf2` module

PR #18241, opened 2024-07-30
chore: Remove cyclic dependency in TrackedEntityService [DHIS2-17714]

enricocolasante:
## Big picture

## This PR

`TrackedEntityService` is depending on `TrackedEntityProgramOwnerService` and `TrackerOwnershipManager` that are depending on `TrackedEntityService` itself.
In all the methods that are using `TrackedEntityService` trackedEntity uid (or even id) is passed as a parameter and then the entity is retrieved but most of the time the caller already has the entity and other times we can just use the uid to extract the information from the DB. In particular this was changed in `HibernateProgramTempOwnerStore` joining with the `trackedentity` table.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
@@ -65,7 +65,6 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
   private final ProgramTempOwnershipAuditService programTempOwnershipAuditService;
   private final ProgramTempOwnerService programTempOwnerService;
   private final ProgramOwnershipHistoryService programOwnershipHistoryService;
-  private final TrackedEntityService trackedEntityService;
   private final DhisConfigurationProvider config;
   private final UserService userService;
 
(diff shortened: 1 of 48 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
```diff
@@ -349,14 +345,9 @@ private boolean hasTemporaryAccessWithUid(
 
     return tempOwnerCache.get(
         getTempOwnershipCacheKey(trackedEntityUid, program.getUid(), user.getUid()),
-        s -> {
-          TrackedEntity trackedEntity = trackedEntityService.getTrackedEntity(trackedEntityUid);
-          if (trackedEntity == null) {
```
teleivo, line 354: just to double-check if the te does not exist the count should be returned as 0, no NPEs right?

teleivo (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
enricocolasante: Yes, this is just checking if it is present in the cache or not

muilpp (approved):

PR #18243, opened 2024-07-30
chore: remove unused asciidoc from dhis-test-web-api DHIS2-17819

teleivo:
The `doc` profile does not work anymore. These docs are not hosted anywhere. The git history shows these files were here since the beginning of this git repos history. They were never changed.

Our docs are built from https://github.com/dhis2/dhis2-docs and we document our APIs using OpenAPI now.


david-mackessy (approved):
I've no idea about this feature, but it sounds like you have a good grip of it.

teleivo:
> I've no idea about this feature, but it sounds like you have a good grip of it.

Its documentation that might have been used at some point. You can look at the content of the deleted asciidoc files

david-mackessy:
> > I've no idea about this feature, but it sounds like you have a good grip of it.
> 
> Its documentation that might have been used at some point. You can look at the content of the deleted asciidoc files

yep, seems redundant now.

PR #18244, opened 2024-07-30
test: Update recaptcha test assertions

david-mackessy:
2 recaptcha tests started failing. They actually call out to a google API and get the response, which is what is causing the tests to fail. The response has changed from google.
Response changed from `invalid-input-secret` to `invalid-input-response`

Initial fix is to keep the assertion more vague. 

Next: Will look to move these 2 tests so we can mock the response from google.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/UserAccountControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/UserAccountControllerTest.java
@@ -54,6 +54,7 @@ import org.hisp.dhis.test.message.FakeMessageSender;
 import org.hisp.dhis.test.web.HttpStatus;
 import org.hisp.dhis.test.webapi.H2ControllerIntegrationTestBase;
 import org.hisp.dhis.test.webapi.json.domain.JsonUser;
+import org.hisp.dhis.test.webapi.json.domain.JsonWebMessage;
 import org.hisp.dhis.user.User;
 import org.hisp.dhis.util.DateUtils;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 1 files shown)


vietnguyen (approved):

vietnguyen (approved):

zubaira (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/UserAccountControllerTest.java
```diff
@@ -411,15 +412,15 @@ void selfRegInvalidRecaptchaInput() {
...
-        "Recaptcha validation failed: [invalid-input-secret]",
+    JsonWebMessage response =
         POST(
                 "/auth/registration",
                 renderService.toJsonAsString(getRegParamsWithRecaptcha("secret", RegType.SELF_REG)))
-            .content(HttpStatus.BAD_REQUEST));
+            .content(HttpStatus.BAD_REQUEST)
+            .as(JsonWebMessage.class);
+
+    assertTrue(response.getMessage().contains("Recaptcha validation failed"));
```
teleivo, line 422: We have https://github.com/dhis2/dhis2-core/blob/92f7623624bb1e695318f139e73578fd4ac51707/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/utils/Assertions.java#L180 (or assertStartsWith) which give a better assertion message 😄 see https://github.com/dhis2/wow-backend/blob/master/guides/testing/testing.md#assertions

david-mackessy: cool, that's much nicer, I'll use that from now on 👍

PR #18252, opened 2024-07-30
fix: Add capture namespace protection [DHIS2-17718]

enricocolasante:


jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/CaptureApp.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * The main purpose (so far) of the {@link CaptureApp} component is to establish the protected
+ * {@link #NAMESPACE} in the {@link DatastoreService} so that only the app can write to it using a
+ * role having the {@link #AUTHORITY}.
+ */
+@Component
+public class CaptureApp {
+  public static final String NAMESPACE = "capture";
+
+  public static final String AUTHORITY = "F_capture";
```
jbee, line 44: I think this should be `F_CAPTURE_DATASTORE_UPDATE`, no?

PR #18260, opened 2024-07-31
chore: Remove get TE methods from TrackedEntityService [DHIS2-17714]

enricocolasante:
## Big picture

## This PR

- Remove `getTrackedEntity(long id)` method from `TrackedEntityService`. Use `manager.get` instead. In `TrackedEntityRegistrationSMSListener`, `getTrackedEntity(String uid)` was used instead as it is auditing the read. We need to have a discussion about this type of auditing and understand when and what we want to really audit.
- Remove `trackedEntityExists(String uid)` method from `TrackedEntityService`. Use `manager.exists` instead.
- Remove `trackedEntityExistsIncludingDeleted(String uid)` method from `TrackedEntityService`. This is then implemented directly in some tests and in `TrackedEntiityStrategy` class.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/TrackerEntityStrategy.java
```diff
@@ -27,37 +27,60 @@
...
-      // Add to preheat
       preheat.putTrackedEntities(
           DetachUtils.detach(
               this.getClass().getAnnotation(StrategyFor.class).mapper(), trackedEntities));
     }
   }
+
+  private List<TrackedEntity> getIncludingDeleted(List<String> uids) {
+    List<TrackedEntity> trackedEntities = new ArrayList<>();
+    List<List<String>> uidsPartitions = Lists.partition(Lists.newArrayList(uids), 20000);
```
teleivo, line 71: as in the other strategy please use google's collection API only when Javas does not exist/is not good enough

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/TrackedEntityRegistrationSMSListener.java
```diff
@@ -142,16 +142,17 @@ protected void postProcess(
...
-      trackedEntityId =
-          trackedEntityService.createTrackedEntity(trackedEntity, patientAttributeValues);
+      trackedEntityService.createTrackedEntity(trackedEntity, patientAttributeValues);
     } else {
       sendFeedback("No TrackedEntityAttribute found", senderPhoneNumber, WARNING);
     }
 
     enrollTrackedEntity(
-        trackedEntityService.getTrackedEntity(trackedEntityId), program, orgUnit, occurredDate);
+        trackedEntityService.getTrackedEntity(trackedEntity.getUid()),
```
teleivo, line 152: In the PR description you say

> getTrackedEntity(String uid) was used instead as it is auditing the read.

do you mean you are using `getTrackedEntity(String uid)` instead of `getTrackedEntity(long)`?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/TrackerEntityStrategyTest.java
```diff
@@ -1,73 +0,0 @@
-/*
- * Copyright (c) 2004-2022, University of Oslo
```
teleivo, line 2: Is the strategy tested somewhere?

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/TrackedEntityRegistrationSMSListener.java
enricocolasante: Yes. We will eventually remove also the other method, but we need to discuss about the read auditing part and I didn't want to do it just yet.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/TrackerEntityStrategyTest.java
enricocolasante: Not a unit test but it is covered by several integration tests.
A unit test now would be weird as we should assert on the SQL to verify that the strategy is working.

muilpp (approved):

PR #18272, opened 2024-08-01
chore: remove unneeded code DHIS2-17821

teleivo:
* `@DependsOn` is not needed if the dependency is explicit via the code like a parameter to a bean annotated method
* Setting the bean name is not necessary if the method has the same name
* Use one mechanism to wire up the bean: either parameters (which I prefer) or fields


david-mackessy (approved):
this makes it much easier to reason about 👍

enricocolasante (approved):

PR #18278, opened 2024-08-02
refactor: remove unused code from program message service/store [DHIS2-17759]

zubaira:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/DefaultProgramMessageService.java
```diff
@@ -267,37 +235,37 @@ public void validateQueryParameters(ProgramMessageQueryParams params) {
...
+      violations.add("Organisation unit does not exist");
     }
 
-    if (violation != null) {
-      log.info("Message validation failed: " + violation);
-
-      throw new IllegalQueryException(violation);
+    if (!violations.isEmpty()) {
+      String violationMessage = String.join(", ", violations);
+      log.info("Message validation failed: " + violationMessage);
```
teleivo, line 267: Why would you log this? How will this exception be handled?

When logging we should always think about who this log entry is for and what they should do about it. In this case I think a user should know that what they did was not valid and what they can do to fix it. Logging this will not help a user. It will just cause bigger log files operators have to store/seek through when looking for errors they should take action on.

enricocolasante (approved):

PR #18279, opened 2024-08-02
refactor: use tracker-export pattern for program message[DHIS2-17759]

zubaira:
**Description:** This PR primarily focuses on refactoring the `ProgramMessage` model. It aims to adopt the same export pattern that we have used for other tracker entities for example `Enrollment`.

**Changes:**
-   **Controller Updates:**
      - The GET method now uses `ProgramMessageRequestParams` for user input.
      - A mapper is added to validate and map user input to `ProgramMessageOperationParams`.

-   **Service layer validation and mapping:**
    - Created a mapper to validate and map `ProgramMessageOperationParams` to `ProgramMessageQueryParams` for use in the repository.
-   **Tests:**
	- Created unit tests for both mapper classes.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/ProgramMessageOperationParamMapper.java
```diff
@@ -0,0 +1,114 @@
...
+        ObjectUtils.firstNonNull(
+            enrollment, Optional.ofNullable(event).map(Event::getEnrollment).orElse(null));
+
+    if (entity == null) {
+      throw new IllegalQueryException("Enrollment or Event has to be provided");
+    }
+
+    List<Program> programs = programService.getCurrentUserPrograms();
+
+    User currentUser = userService.getUserByUsername(CurrentUserUtil.getCurrentUsername());
```
muilpp, line 106: we should use `CurrentUserUtil` instead of the `userService`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/ProgramMessageOperationParamMapper.java
```diff
@@ -0,0 +1,114 @@
...
+    if (currentUser != null && !programs.contains(entity.getProgram())) {
+      throw new IllegalQueryException(
+          String.format(
+              "User:%s does not have access to the required program:%s",
```
muilpp, line 110: if the `currentUser` is null, should we provide a more accurate error message?
in this case it's not that the user has no access, it's that something went wrong and we couldn't retrieve the user, right?

zubaira: is it possible to have `currentUser` as null? 
Even in async mode there will be `systemUser` available.

muilpp: I guess this is a question for @netroms

PR #18282, opened 2024-08-02
chore: production code should not be test aware DHIS2-17821

teleivo:


teleivo:
This is currently blocked because of https://github.com/dhis2/dhis2-core/pull/18270#discussion_r1734056805

PR #18290, opened 2024-08-05
refactor: program message extend read only crud controller[DHIS2-17759]

zubaira:
**Description:** This PR primarily focuses on refactoring the ProgramMessageController. 

**Changes:**
 -   **Controller Updates:**
	- `ProgramMessageController` now extends `AbstractFullReadOnlyController`.
	- `PUT` and `DELETE` methods are explicitly added.

 -   **Services**
    -   Moved `ProgramMessageImport` to web layer
 -   **Tests:**
	- Added e2e test for program message

<span style="color:red">**Breaking Change**</span>
PR will restrict the update of program message by only allowing changes in program message status. Other fields do not need an update.
The rationale behind this is that `ProgramMessage` is an intermediate object created by `ProgramNotificationService` and used by `MessageSender`.  The only thing that the user can be interested in changing is its `status`.

Here are the fields
 - `Enrollment` (foreign key)
 - `Event` (foreign key)
 - `recipients`. (message destination. once a message is sent there is no point in changing its destination)
 - `text` (same as above)
 - `subject` ( same as above)
 - `processedDate` (auto generated)
 - `notificationTemplate` (foreign key)
 - `deliveryChannels` ( one message is sent through SMS or EMAIL then no need to this to be changed )


enricocolasante (commented):
Maybe it would make sense to have a meeting on this one where we try to understand how the feature is working now and what we are trying to achieve here.
I am confused an in my mind a `ProgramMessageController` should write anything at all, the messages are sent by jobs, why the user can change the status?

PR #18295, opened 2024-08-06
refactor: User ForbiddenException for top level ACL checks

david-mackessy:
# Background
We have several different ways to represent unauthorized errors:
- `ForbiddenException` (DHIS code)
- `AccessDeniedException` (Spring code, extended by the 4 exceptions below)
  - `CreateAccessDeniedException` (DHIS code)
  - `ReadAccessDeniedException` (DHIS code)
  - `UpdateAccessDeniedException` (DHIS code)
  - `DeleteAccessDeniedException` (DHIS code)

# Changes in this PR
Our agreed approach is to throw the checked exception `ForbiddenException` at the top level of the system. This  includes `Controller`s & `Service`s. This allows us to clearly indicate that this exception must be thought about and handled as an API consumer (web or internal services).

At any other lower level e.g. `Store`, we can throw the unchecked `AccessDeniedException`. Because there are a lot of ACL checks at the lower levels, we cannot throw a checked exception here as it would force most of the system to change  method signatures and this is not desirable.

Delete custom exceptions
- `CreateAccessDeniedException`
- `ReadAccessDeniedException`
- `UpdateAccessDeniedException`
- `DeleteAccessDeniedException`

A few pom dep changes (2 x removals, 1 x addition).

# Info
- We handle `ForbiddenException` [here](https://github.com/dhis2/dhis2-core/blob/7faf2987a68d85e03656a6d4b4a659a2f1e2c2c4/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java#L193)
- We handle `AccessDeniedException` [here](https://github.com/dhis2/dhis2-core/blob/7faf2987a68d85e03656a6d4b4a659a2f1e2c2c4/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java#L437)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ForbiddenException.java
```diff
@@ -40,7 +40,7 @@
 @Getter
 @Accessors(chain = true)
 @OpenApi.Response(status = FORBIDDEN, value = WebMessageResponse.class)
-public final class ForbiddenException extends Exception implements Error {
+public final class ForbiddenException extends RuntimeException implements Error {
```
jbee, line 43: Please revert to a checked exception and propagate that in the method signatures as `throws ForbiddenException`

jbee (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ForbiddenException.java
david-mackessy: `ForbiddenException` kept as checked exception. We discussed this & the general approach will be to throw checked exceptions at the top level (controllers, services) and throw unchecked exception (`AccessDeniedException`) at lower levels (e.g. store).

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dataapproval/DataApprovalWorkflowServiceTest.java
```diff
@@ -172,7 +171,7 @@ void testSaveWorkFlowWithAuthority() {
   void testSaveWorkFlowWithoutAuthority() {
     createUserAndInjectSecurityContext(false, null);
     assertThrows(
-        CreateAccessDeniedException.class,
+        AccessDeniedException.class,
```
jbee, line 174: Note that `assertThrows` does include sub-classes so if this still would throw a `CreateAccessDeniedException` this assert would also pass. If you want to assert an exact exception type use `assertThrowsExactly`. I always thought the naming on those got it wrong because I feel the default one with the short name should have been the exact match and then there should be one called `assertThrowsAssignable` or so if you want to include any subtypes.

jbee (approved):
Came out nice for what we have to work with :)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dataapproval/DataApprovalWorkflowServiceTest.java
david-mackessy: `CreateAccessDeniedException` will never be thrown as it has been deleted. It's now either `ForbiddenException` or `AccessDeniedException` depending on the level where it's thrown.

vietnguyen (approved):

PR #18297, opened 2024-08-06
chore: Change how Period is saved in some tests

david-mackessy:
I've seen these tests fail sometimes in PRs, never locally tough.
Transient exception when Period not saved, even though they are saved!
This change uses the period service to save them to see if it improves.


jbee (approved):

vietnguyen:
If this doesn't work then you could try with `PeriodService.reloadPeriod`.

vietnguyen (approved):

PR #18298, opened 2024-08-06
chore: Return enrollments of event programs only [DHIS2-17712]

muilpp:
The `CopyService` should only replicate metadata when copying a program, never the actual data. 
This means it should copy enrollments in event programs, where the enrollment serves as a placeholder, but should not copy enrollments in tracker programs, as these are considered data.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EventProgramEnrollmentService.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EventProgramEnrollmentService.java
@@ -36,12 +36,15 @@ import java.util.List;
  */
 public interface EventProgramEnrollmentService {
 
-  /** Returns a list of enrollments in the given program. */
-  List<Enrollment> getEnrollments(Program program);
-
   /**
    * Returns the list of enrollments in the given program, only if the program is of type {@link
    * ProgramType#WITHOUT_REGISTRATION}
+   */
+  List<Enrollment> getEnrollments(Program program);
+
+  /**
+   * Returns the list of enrollments in the given program and status, only if the program is of type
+   * {@link ProgramType#WITHOUT_REGISTRATION}
    */
   List<Enrollment> getEnrollments(Program program, EnrollmentStatus enrollmentStatus);
 }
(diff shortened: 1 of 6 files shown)


david-mackessy (approved):
Looks good, could a comment be added (probably in the copy service where the enrollments are retrieved) explaining why we only want a certain type of enrollment? Just in case someone looks at this in the future and doesn't know the difference between enrollment types. That bit of knowledge won't be lost then.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/EventProgramEnrollmentService.java
```diff
@@ -36,12 +36,15 @@
...
    * Returns the list of enrollments in the given program, only if the program is of type {@link
    * ProgramType#WITHOUT_REGISTRATION}
    */
+  List<Enrollment> getEnrollments(Program program);
+
+  /**
+   * Returns the list of enrollments in the given program and status, only if the program is of type
+   * {@link ProgramType#WITHOUT_REGISTRATION}
+   */
   List<Enrollment> getEnrollments(Program program, EnrollmentStatus enrollmentStatus);
```
ameenhere, line 49: I am wondering if the actual method names could/should be changed to make it self-explanatory? 
 If the method only returns enrollments for event programs, the name "getEnrollments(..)" is misleading right? The javadoc might be overlooked by a consumer. 

The interface is EventProgramEnrollmentService though. So that may suffice. What are your thoughts?

david-mackessy: the method name reflecting what it's actually doing sounds like a good improvement to me 👍

muilpp: In my mind, naming the interface `EventProgramEnrollment` was meant to indicate that any method under it would deal exclusively with event programs. I didn't want to make the name redundant by calling each method `getEventProgramEnrollments`. I added the Javadoc in case someone is using it and doesn't know the difference between event and tracker programs.

Anyway, this service should be temporary, since the long-term idea is to separate tracker programs from event programs so that they will be different models. Once we achieve that, this service will no longer be needed.

ameenhere: Sounds good to me. 👍

ameenhere (approved):

PR #18314, opened 2024-08-08
chore: upgrade to spring 6

vietnguyen:
### Goal
- Upgrade spring to version 6.1.12. 
- Hibernate is kept at version 5.6.15.Final. No upgrade. 
- Dependencies upgrade:
  - [Servlet API 6](https://jakarta.ee/specifications/servlet/6.0/)
  - [Spring Security 6.3](https://docs.spring.io/spring-security/reference/whats-new.html)
  - [Ehcache 3](https://www.ehcache.org/documentation/3.0/getting-started.html) - Use for Hibernate second level cache

### Notes
- `SocketUtils` is removed from spring utils package. So I copied the class to our project as a temporary solution.
- Support for Apache HttpClient has been [removed in Spring Framework 6.0](https://github.com/spring-projects/spring-framework/issues/28925), immediately replaced by `org.apache.httpcomponents.client5:httpclient5`
- Spring 6 [Introduce HttpStatusCode interface ](https://github.com/spring-projects/spring-framework/issues/28214).  `Response.getStatusCode()` return the interface `HttpStatusCode` instead of enum `HttpStatus` like before. This breaks many existing codes.
- In `WebhookHandler` the method `httpClient.setTimeOut()` has been removed. Need to replace by setting `SocketConfig.setSoTimeout`
- https://github.com/spring-projects/spring-framework/issues/29559 - This causes some of the bean overriding in test config stop working.
- https://github.com/spring-projects/spring-framework/issues/31130 - This change causes current code with `@EventListener` stop working.
- MultiPart changes required
  - Spring 6 `MultipartResolver` changed from using `CommonsMultipartResolver` to `StandardServletMultipartResolver` ([change](https://github.com/dhis2/dhis2-core/pull/18314/files#diff-ba454979438951fa6f3bb824bfe3662be41a35cb664016b2f356f9df11bdc68cR162) already included by Viet)
  - Seems like 1 more change was required to get `FileStoreTest#shouldStoreFileResourceInExternalStore` passing
  - Before the change it was failing with a `400`  with message `Required part 'file' is not present.`. The servlet could not recognise that there was a file submitted as part of the multipart request.
  - searched and found spring docs ([1](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-methods/multipart-forms.html)) ([2](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-servlet/multipart.html)) that recommend adding one more piece of config (setting multipart config on dispatch servlet)
  - tested locally and file upload stored in file system directory `dhis2_home/files/icon`
  - able to save & retrieve the file from system

### Spring-security changes
- `builder.clientAuthenticationMethod(ClientAuthenticationMethod.BASIC);` -> `builder.clientAuthenticationMethod(ClientAuthenticationMethod.CLIENT_SECRET_BASIC);`
- [@EnableGlobalMethodSecurity](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#migration-enableglobalmethodsecurity) is replaced by [@EnableMethodSecurity](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_12__Fix_id_column_type.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE programruleaction ALTER COLUMN notificationtemplateid TYPE BIGINT;
```
vietnguyen, line 1: new jpa schema validator complain about this column has type `integer` in database but `long` in Java

dhis-2/dhis-support/dhis-support-hibernate/pom.xml
```diff
@@ -36,20 +36,40 @@
...
+        <exclusion>
+          <groupId>org.hibernate</groupId>
+          <artifactId>hibernate-core</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
-      <groupId>com.vladmihalcea</groupId>
-      <artifactId>hibernate-types-52</artifactId>
-      <version>${hibernate-types.version}</version>
```
vietnguyen, line 52: `com.vladmihalcea` is replaced by `io.hypersistence`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/DhisWebApiWebSecurityConfig.java
```diff
@@ -453,7 +452,7 @@ private void configureCspFilter(
   }
 
   private void configureCorsFilter(HttpSecurity http) {
-    http.addFilterBefore(CorsFilter.get(), BasicAuthenticationFilter.class);
+    http.addFilterBefore(DhisCorsFilter.get(), BasicAuthenticationFilter.class);
```
vietnguyen, line 455: Fix error 
```
Bean named 'corsFilter' is expected to be of type 
'org.springframework.web.filter.CorsFilter' but was actually of type org.hisp.dhis.webapi.filter.CorsFilter
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/DhisWebApiWebSecurityConfig.java
```diff
@@ -419,7 +419,6 @@ public <O extends BasicAuthenticationFilter> O postProcess(O filter) {
         .and()
         ////////////////////
         .sessionManagement()
-        .requireExplicitAuthenticationStrategy(true)
```
vietnguyen, line 422: Fix below error 
```
Caused by: java.lang.IllegalStateException: Invalid configuration that explicitly sets requireExplicitAuthenticationStrategy to true but implicitly requires it due to the following properties being set: [sessionCreationPolicy = ALWAYS, maximumSessions = 10]
	at org.spr
```

jbee (approved):
Things to look at after merge (team tasks):

* Using `default` implementation of new methods in `MessageSender` seems off as their default implementation looks more like a stub 
* (as discussed) have a look at job execute now (likely simplification is possible now)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/message/MessageSender.java
```diff
@@ -71,4 +74,14 @@ ListenableFuture<OutboundMessageResponseSummary> sendMessageBatchAsync(
...
+
+  default List<OutboundMessage> getMessagesByEmail(String recipient) {
+    return List.of();
+  }
+
+  default void clearMessages() {}
+
+  default List<OutboundMessage> getAllMessages() {
+    return List.of();
+  }
```
david-mackessy, line 86: wondering what these new default methods are for? are they used anywhere?
should a MessageSender just send messages, or get messages as well? :)

david-mackessy: i've seen it used in the FakeMessageSender while looking through the rest of the PR. This seems like a code smell if it's only used for tests.

david-mackessy (approved):
great work

dhis-2/dhis-api/src/main/java/org/hisp/dhis/message/MessageSender.java
vietnguyen: This is currently a workaround for fixing the issue of overriding a bean in test environment. As `@Qualifer` doesn't work for that purpose.  I can give an explanation in next week meeting.
Basically in test we have `FakeMessageSender` `SmsMessageSender` and `EmailMessageSender`.
We want the `FakeMessageSender` to override both `SmsMessageSender` and `EmailMessageSender`.  
But `FakeMessageSender` has specific methods `getAllMessages()`, which then can't be casted back to the others messageSenders.

vietnguyen: In Spring 5 you can override a bean by declaring same bean name. 
But in spring 6, it doesn't work anymore. Need to implement same interface.

PR #18315, opened 2024-08-08
feat: OpenAPI HTML rendering - improvments [DHIS2-17812]

jbee:
This is a follow up on #18144 where open ends are tidied up.

### Improvements Declaration
Ways in which programmers can declare OpenAPI details has become more capable
- [x] chore: replace `@OpenApi.Shared#pattern()` with `name()` (pattern wasn't worth the extra alternative)
- [x] refactor: make @OpenApi.Document#group a String that also can have a description 
- [x] feat/fix: default values are now given as JSON in annotations or converted property from Java values for 3rd party annotation where the `String` value had that semantic, overall this makes default values more correct, consistent and allows specifying complex default values for any type as long as it can be expressed in JSON
- [x] fix: `@OpenApi.Description` should take effect for object properties (declarations on fields and methods)
- [x] feat: `@OpenApi.Kind` auto-applies to enclosed (inner) classes when used on top level type

### Improvements Generation
Ways in which the generated OpenAPI JSON document has become better
- [x] feat: extraction of OpenAPI schemas from `JsonValue` types, in particular sub-types of  `JsonObject` 
- [x] fix: logic of which properties to include in presence of `@OpenApi.Propertiy` to avoid the pitfall that adding it on one field makes all now non annotated fields be no longer included
- [x] qoi: provided detailed types for the job configuration errors API (also as a test case for the `JsonValue` support) 
- [x] qoi: added a regex `pattern` info to schema for `Locale` 
- [x] qoi: OpenAPI JSON endpoint use specific type (not `object`)
- [x] feat: new parameter `extendedRefs` to ignore effects of `@JsonSerialize(as=...)` and `@JsonSerialize(contentAs=...)` (this is in essence what DHIS2-17684 asked for)
- [x] fix: `JsonMap` support wasn't generating proper object declaration 
- [x] feat: include annotation based required authorities as `x-auth` property of an operation 

### Improvements Rendering
Ways in which the generated OpenAPI HTML is more complete or correct
- [x] fix: object parameters with descriptions and a `$ref` type now render the parameter description
- [x] feat: show required authorities of an operation 

### Next
Tasks that were identified (not necessarily of high priority)
- [ ] fix: support `oneOf` via `@OpenApi.Property`
- [ ] perf: `Class` based schemas should be cachable in a static map (maybe)
- [ ] feat: HTML generation summary with timings and some counts


david-mackessy (approved):
nice work

vietnguyen (approved):

PR #18316, opened 2024-08-08
chore: Remove getTrackedEntity(UID) [DHIS2-17714]

muilpp:
## Big picture

## This PR

- Removes the method `getTrackedEntity(String uid)` from `TrackedEntityService`. 
- It's replaced by the manager in all the tests, and by the `getTrackedEntity` method from the tracker module `TrackedEntityService` in the service layer. There are still usages of the manager in the service, they are explained below.
-  The use of the manager in the `OperationParamsValidator` needs to be revisited. I'm not sure if it is sufficient or if we need to use the tracker service to run the appropriate validations. Same goes for the `DeduplicationController`.
- In `DefaultProgramMessageService` and `DeliveryChannelStrategy` we still use the manager because they both live in the service core module, so they have no access to the tracker module. These two services will audit the log directly using the `TrackedEntityChangeLogService`. Despite its name, this service actually generates audit logs, not change logs. However, this is only a temporary solution, as the service belongs to the tracker module and should be relocated to the appropriate module. Once we move it, neither of these two services will have access to it anymore.
- `DefaultTrackerObjectsDeletionService` is also using the manager, because the proper validations are already run before we reach that point, it's not necessary to run them again.
- The error messages of the SMS feature need to be rethought. Once we are able to mock the gateway server and test it properly we can redo that part as well.

Please note this is a very long PR, but there are a lot of changes that consist of a method signature refactor only.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -186,11 +191,14 @@ public TrackedEntity validateTrackedEntity(String trackedEntityUid, User user)
       return null;
     }
 
-    TrackedEntity trackedEntity = trackedEntityService.getTrackedEntity(trackedEntityUid);
+    // TODO(tracker) Are these validations enough? Should we check for ownership too?
```
enricocolasante, line 194: We should probably unify this method with the one present in RelationshipOperationParamsMapper when solving this TODO

zubaira (approved):

PR #18317, opened 2024-08-08
chore: Copy enrollments of event programs only [DHIS2-17712] [2.41]

muilpp:
back port of https://github.com/dhis2/dhis2-core/pull/18298

I can't use the interface we do in master as it does not exist in v41, so I just added a guard clause for this particular case.


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/ProgramControllerIntegrationTest.java
```diff
@@ -88,76 +76,6 @@ public void testSetup() throws JsonProcessingException {
     POST("/metadata", WebClient.Body("program/create_program.json")).content(HttpStatus.OK);
   }
 
-  @Test
```
david-mackessy, line 91: instead of removing the test, could it be updated so it still provides value and confirms at least one of the expected behaviours? (either copying enrollments or not copying enrollments)

muilpp: sure, I can test tracker program enrollments are not copied.
Unfortunately I can't test event program enrollments are because they are not returned in the api response.

muilpp: I'll add this test in master as well.

david-mackessy: nice one 👍

david-mackessy (approved):

PR #18330, opened 2024-08-12
feat: email verification

netroms:
# Summary
Adds a verified email field to the User object, to facilitate for email verification.
The users email will be verified when the user has received an email with a verification link and clicked it. When the link is clicked, the user's verified email field will be set to the same as the email field.
The email verified field is unique in contrast to the original email field, hence many can have the same email field value, but only one can have the verified email.

Technical non-obvious choices made:

-  This is the first iteration of this feature to facilitate specifically for 2FA emails. 
- Having two email fields, instead of only one and a boolean saying verified/not verified.
   - In order support both unique emails and non-unique emails for backward compatibility, only one, the `verifiedEmail` is unique.
- Another column `emailverificationtoken` is used for storing the verification token.
- An account is not "locked" during/after verification email is sent/requested. You can send as many verification emails as you want, until the latest is clicked/verified and the state is changed to verified. You can change email and re-verify, by first changing the non-unique email, then resend the verification and verify the new one. There is no automation to this procedure in order to be backward compatible and verification has to be triggered/requested manually trough the API for now.
-  In this first iteration of this feature, we only intend to use this as a prerequisite for enabling 2FA with email, i.e. you can't enable 2FA with email, unless your email is verified first. 
- No restrictions is added to the functionality of other features based on verified vs non-verified emails.
- You can try to verify any email you want, you can change your email after verification and try to verify the new one.
- When changing your non-unique email, you are automatically not having a verified email, until you verify the new one.
- System wide enforcement of verified emails is not a part of this task, but can be added in a later iteration.
- We are only forcing users to verify their email if and ONLY when they want to enable 2FA with email.
- There is no automatic/background cleanup of expired tokens, they will just be overwritten next time you try to send a new verification email, and set to NULL on successful verification.

## Automatic tests
AccountControllerTest

## Manual test
There is no UI for this feature yet.
To test manually. Do the following API calls on an instance with a valid SMTP server configured:
1. Make sure your test user has en email set, that you control. Use this user when calling the API.
2. Call  POST `/api/users/USER_ID/sendEmailVerification`
3. Open your email and click the link or call: GET `/api/users/USER_ID/verifyEmail?token=YOUR_TOKEN_RECEIVED_IN_THE_LINK`
4. Observe that the user now has the verifiedEmail field set: GET `/api/users/USER_ID`

### JIRA:
[DHIS2-16853](https://dhis2.atlassian.net/browse/DHIS2-16853)

[DHIS2-16853]: https://dhis2.atlassian.net/browse/DHIS2-16853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
```diff
@@ -619,4 +619,20 @@ public void setActiveLinkedAccounts(
...
+  @Override
+  public User getUserByVerificationToken(String token) {
+    Query<User> query =
+        getSession().createQuery("from User u where u.verifiedEmail like :token", User.class);
+    query.setParameter("token", token + "%");
+    return query.uniqueResult();
+  }
+
+  @Override
+  public User getVerifiedEmail(String email) {
```
david-mackessy, line 632: Should the method name be `getUserByVerifiedEmail`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
```diff
@@ -619,4 +619,20 @@ public void setActiveLinkedAccounts(
...
+    Query<User> query =
+        getSession().createQuery("from User u where u.verifiedEmail like :email", User.class);
```
david-mackessy, line 634: is there a reason why `like` is being used here over `=`?
do we prevent using `%` in an email?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -1516,4 +1520,84 @@ public boolean canDataRead(IdentifiableObject identifiableObject) {
...
   }
+
+  @Override
+  @Transactional
+  public boolean sendEmailVerificationToken(User user, String requestUrl) {
+    String token = CodeGenerator.getRandomSecureToken();
+
+    String encodedToken = token + "|" + (System.currentTimeMillis() + EMAIL_TOKEN_EXPIRY_MILLIS);
+    user.setVerifiedEmail(encodedToken);
+    updateUser(user);
```
vietnguyen, line 1531: Should we update `User` after email is sent successfully ?

david-mackessy (changes requested):
I have a suggestion & some questions (I appreciate they might not all be relevant for this piece of work but if i don't ask them now I'll forget them) 🙂

### Suggestion
Maybe we shouldn't store the verification token in the verified email address
- looking at the DB in isolation without knowing about this feature won't make any sense (mix of emails and tokens)
- data potentially left in this table that won't make sense
- possible bugs whereby we may check if a verified email already has a value & presume it's an email (but it's a token)

probably better to use a verifyEmailToken column or something to that effect (keep them separate)

### Questions
Is a user account `locked`/`unusable` until their email is verified?
- what does this flow look like?
- do we want to restrict a user from performing any actions until verified?

US point 4 says `User can choose to use existing email OR a new one`
what does `or a new one` mean here?
- does it mean a user is allowed to set a new email for themselves or to just provide a new email for this feature?
- either way is it the same user email field that's used for this feature?

Is there value in having the verified email as a string value?
Does this feature now prevent new users from using the same email address?
- why not use a boolean?
- are we trying to get to a point where only 1 user is tied to 1 email?
- should we try to push that back to the actual email instead, so we don't allow the system to enter a state that we don't want (multiple users with the same email but only 1 is verified and the rest left unverified)? (for new users)

What is the status of all existing users?
- will they all be unverified?
- will they be forced to verify?
- do we want to get to a state where all users with emails are verified?

What happens if `/account/verifyEmail?token={token}` is called multiple times?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: I don't think that adds much value. The different responses from email sending is not really 100% reliable anyway. It's worse that we end up NOT saving the token and maybe the email is sent, because the error messages is not correct. For example, it can be waiting in the transport and get sent later.
Worst case it is saved and not sent, is that nothing happens, and the user can try to send it again.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -1516,4 +1520,89 @@ public boolean canDataRead(IdentifiableObject identifiableObject) {
...
+    }
+
+    user.setEmailVerificationToken(null);
+    user.setVerifiedEmail(user.getEmail());
+    updateUser(user);
+    return true;
+  }
+
+  @Override
+  @Transactional
```
david-mackessy, line 1592: should this be a `readOnly` transaction?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -1516,4 +1520,89 @@ public boolean canDataRead(IdentifiableObject identifiableObject) {
...
+  public User getUserByVerificationToken(String token) {
+    return userStore.getUserByVerificationToken(token);
+  }
+
+  @Override
+  @Transactional
```
david-mackessy, line 1598: should this be a `readOnly` transaction?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -1516,4 +1520,89 @@ public boolean canDataRead(IdentifiableObject identifiableObject) {
...
+  public boolean isEmailVerified(User user) {
+    return user.getEmail().equals(user.getVerifiedEmail());
+  }
+
+  @Override
+  @Transactional
```
david-mackessy, line 1604: should this be a `readOnly` transaction?

david-mackessy (approved):
nice, just a few transactions comments

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -1516,4 +1520,89 @@ public boolean canDataRead(IdentifiableObject identifiableObject) {
...
+  public String generateAndSetNewEmailVerificationToken(User user) {
+    String token = CodeGenerator.getRandomSecureToken();
+    String encodedToken = token + "|" + (System.currentTimeMillis() + EMAIL_TOKEN_EXPIRY_MILLIS);
+    user.setEmailVerificationToken(encodedToken);
+    updateUser(user);
+    return token;
+  }
+
+  @Override
+  @Transactional
```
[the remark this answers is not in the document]
jbee: the message sending might write to DB

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
jbee: it should

jbee (approved):
You should also annotate those service methods that don't need `@Transactional` with `@NonTransactional` to indicate that you have considered the transactional question with that outcome.

PR #18333, opened 2024-08-12
fix: Don't find un-logged-in openId [DHIS2-17784]

jimgrace:


jbee (approved):
Looks good to me by this really should have approval from @netroms for agreeing with the semantics assumed by the code.

netroms (approved):

PR #18334, opened 2024-08-12
fix(programRuleAction): add propertyType annotation to templateUid

Birkbjo:
Fixes https://dhis2.atlassian.net/browse/DHIS2-17800

There were some internal changes to `templateUid` made in https://github.com/dhis2/dhis2-core/issues/17723. However that change made `templateUid` appear as a `REFERENCE` in the schema, but it's exposed as a string in the API, and the reference is handled internally.


zubaira (approved):

david-mackessy (approved):
if @zubaira is happy with the change and it still provides the functionality they were looking for, then great 👍

PR #18345, opened 2024-08-15
fix: Validate program/orgUnit combo when transferring [DHIS2-17880]

muilpp:
When transferring the ownership of a te/program combination, we need to validate the org unit where the ownership is transferred to, is associated with the selected program. If it's not, we need to fail and let the user know.
While I was at it, I removed the `skipAccessValidation` and `createIfNotExists` flags, because they were always set to false in our production code.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-17880


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerOwnershipManagerTest.java
```diff
@@ -506,6 +514,19 @@ void shouldFindTrackedEntityWhenTransferredToInaccessibleOrgUnitIfSuperUser()
...
+  @Test
+  void shouldNotTransferOwnershipWhenOrgUnitNotAssociatedToProgram() {
+    OrganisationUnit notAssociatedOrgUnit = createOrganisationUnit('C');
+    organisationUnitService.addOrganisationUnit(notAssociatedOrgUnit);
+    Exception exception =
+        assertThrows(
+            AccessDeniedException.class,
+            () -> transferOwnership(trackedEntityA1, programA, notAssociatedOrgUnit));
+    assertEquals(
+        "User does not have access to change ownership for the entity-program combination",
```
ameenhere, line 526: This sounds like an access error, which is true if the hasAccess() method fails.

Does it make sense to have a different error along the lines of "The program is not associated/supported in this org unit" and have that return specifically when the new validation you added fails?

muilpp: Yes, it probably makes more sense to return the error you mention, I'll change it.

ameenhere (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerAccessManagerTest.java
```diff
@@ -261,7 +262,8 @@ void checkAccessPermissionForEnrollmentInClosedProgram() {
         "User has no create access to organisation unit:");
     enrollment.setOrganisationUnit(orgUnitA);
     // Transferring ownership to orgUnitB. user is no longer owner
-    trackerOwnershipManager.transferOwnership(te, programA, orgUnitB, true, true);
+    trackerOwnershipManager.assignOwnership(trackedEntityA, programA, orgUnitA, false, true);
```
enricocolasante, line 265:
```suggestion
    trackerOwnershipManager.assignOwnership(trackedEntity, programA, orgUnitA, false, true);
```

PR #18346, opened 2024-08-15
fix: Validate transfer program/orgUnit combo [DHIS2-17880][2.41]

muilpp:
backport of https://github.com/dhis2/dhis2-core/pull/18345

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
@@ -43,6 +43,7 @@ import org.hisp.dhis.organisationunit.OrganisationUnitService;
 import org.hisp.dhis.program.Program;
 import org.hisp.dhis.program.ProgramOwnershipHistory;
 import org.hisp.dhis.program.ProgramOwnershipHistoryService;
+import org.hisp.dhis.program.ProgramService;
 import org.hisp.dhis.program.ProgramTempOwner;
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerOwnershipManager.java
```diff
@@ -134,31 +140,37 @@ public void transferOwnership(
...
-                  program,
-                  entityInstance,
-                  teProgramOwner.getOrganisationUnit(),
-                  teProgramOwner.getLastUpdated(),
-                  teProgramOwner.getCreatedBy());
-          programOwnershipHistoryService.addProgramOwnershipHistory(programOwnershipHistory);
-          trackedEntityProgramOwnerService.updateTrackedEntityProgramOwner(
-              entityInstance, program, orgUnit);
-        }
+      if (teProgramOwner != null && !teProgramOwner.getOrganisationUnit().equals(orgUnit)) {
```
ameenhere, line 163: I just realized something regarding this merging of conditions here. 

Isnt the below `else if (createIfNotExists)` applicable only when  `teProgramOwner!=null` and should not be applicable even if `!teProgramOwner.getOrganisationUnit().equals(orgUnit)`

Or am I misunderstanding something?

muilpp: you're right, I removed that flag from master since it's always set to false, but I didn't remove it in the back ports, nice catch!

ameenhere (approved):

PR #18360, opened 2024-08-19
feat: mobileApplicationNotification [DHIS2-17630]

jimgrace:
See [DHIS2-17630](https://dhis2.atlassian.net/browse/DHIS2-17630).

This is a hook needed by PEPFAR (and conceivably others) to put custom HTML/JavaScript into the v40 mobile login page. They've moved their v40 implementation to a single sign-on system and they use system setting `keyApplicationNotification` to redirect non-mobile users to the sign-on system. The problem is that the login page for mobile users can't also redirect mobile users to the sign-on system, so mobile users effectively cannot log in. This feature/fix allows PEPFAR (or anyone else) to supply custom HTML/JavaScript for the mobile login page.

Note that this is only needed for v40, since v41 and following use a login app.

[DHIS2-17630]: https://dhis2.atlassian.net/browse/DHIS2-17630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
+++ b/dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
@@ -67,6 +67,7 @@ public enum SettingKey {
   APPLICATION_TITLE("applicationTitle", "DHIS 2", String.class, false, true),
   APPLICATION_INTRO("keyApplicationIntro", true),
   APPLICATION_NOTIFICATION("keyApplicationNotification", true),
+  MOBILE_APPLICATION_NOTIFICATION("keyMobileApplicationNotification", false),
   APPLICATION_FOOTER("keyApplicationFooter", true),
   APPLICATION_RIGHT_FOOTER("keyApplicationRightFooter", true),
   FLAG("keyFlag", "dhis2", String.class),
--- a/dhis-2/dhis-web/dhis-web-commons-resources/src/main/webapp/dhis-web-commons/security/login_mobile.vm
+++ b/dhis-2/dhis-web/dhis-web-commons-resources/src/main/webapp/dhis-web-commons/security/login_mobile.vm
@@ -45,6 +45,12 @@
 <div class="header-box" align="center"><a href="account.action">Create an account</a></div>
 #end
 
+#if ( $keyMobileApplicationNotification )
+<div id="mobileNotificationArea">
+	$!{keyMobileApplicationNotification}
+</div>
+#end
+
 </form>
 </div>
 </div>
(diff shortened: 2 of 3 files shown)


dhis-2/dhis-services/dhis-service-setting/src/main/java/org/hisp/dhis/setting/SettingKey.java
```diff
@@ -67,6 +67,7 @@ public enum SettingKey {
   APPLICATION_TITLE("applicationTitle", "DHIS 2", String.class, false, true),
   APPLICATION_INTRO("keyApplicationIntro", true),
   APPLICATION_NOTIFICATION("keyApplicationNotification", true),
+  MOBILE_APPLICATION_NOTIFICATION("keyMobileApplicationNotification", false),
```
jbee, line 70: I assume the name is picked to be in line the the one above. But if their purpose is to inject random HTML/JS into the login page both names are really poor ones and should be reconsidered. Also this seems a pretty big security risk to me if you can just put anything in this setting and it is rendered "as is" into the login. Again, I see how this maybe just follows the pattern of the other setting but still this might not be a good idea to add. Has this been discussed in the security meeting?

jimgrace: @jbee Regarding the setting name: There are several such settings: APPLICATION_TITLE, APPLICATION_INTRO, APPLICATION_NOTIFICATION,  APPLICATION_FOOTER, and APPLICATION_RIGHT_FOOTER. PEPFAR says they only need a mobile variant of APPLICATION_NOTIFICATION to put in the code they need. Of course we could make mobile variants of all the others too. I was operating under the principle of doing only what is required. I assume that we've not been changing the names of existing keys for backwards compatibility -- for example, not renaming APPLICATION_FOOTER to APPLICATION_LEFT_FOOTER after APPLICATION_RIGHT_FOOTER was added, even though in the System Settings UI the prompts for these two settings are called "Application left-side footer" and "Application right-side footer". (A Flyway patch would be needed if the setting names change in the database.) PEPFAR doesn't care what this new setting is called; if you have another suggestion I'm happy to go with that.

Regarding security: What would you suggest as an alternative? These settings have been there for a long time and I know that many implementations make use of them to customize the look and feel of their installations. In SL Demo, the APPLICATION_FOOTER is "`Learn more at <a href="http://www.dhis2.org">dhis2.org</a>`". Changing the system settings requires a special privilege which I hope implementations would severely restrict. I didn't bring this to the security meeting but you are welcome to if you think it should be. I had hoped this would be a timely fix because mobile logins currently don't work for all of PEPFAR's DATIM system.

PEPFAR considers this to be a temporary solution. Once they move to v42 they intend to use a custom login app so they won't need this setting.

netroms: We have a strict CSP policy on the login page on 2.40.
using script-src with a nonce. This prevents injection of inline Javascript and CSS, and potential "hidden" stealing of password etc. Would be nice to confirm, once merged.

netroms (approved):

PR #18361, opened 2024-08-19
task: Use CurentUserDetails instead of User [DHIS2-17912]

muilpp:
This is a refactor of the tracked entity services, where any usage of the `User` class is replaced by `UserDetails`.
https://dhis2.atlassian.net/browse/DHIS2-17912

Currently, the implementation of `CurrentUserUtil.getCurrentUserDetails` can return null, @netroms is working on a fix where that particular method will return an unchecked exception instead of null.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -659,10 +683,12 @@ private String getDescendantsQuery(TrackedEntityQueryParams params) {
     orgUnits.append("AND (");
 
     for (OrganisationUnit organisationUnit : params.getOrgUnits()) {
-
-      OrganisationUnit ou = organisationUnitStore.getByUid(organisationUnit.getUid());
-      if (ou != null) {
-        orgUnits.append(orHlp.or()).append("OU.path LIKE '").append(ou.getPath()).append("%'");
+      if (organisationUnit != null) {
```
ameenhere, line 686: Confused about this not null check. The previous code fetched something from db before checking null

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapperTest.java
```diff
@@ -147,7 +151,12 @@ void shouldMapWithoutFetchingNullParamsWhenParamsAreNotSpecified()
     EnrollmentOperationParams operationParams =
         EnrollmentOperationParams.builder().orgUnitMode(ACCESSIBLE).build();
 
-    mapper.map(operationParams);
+    try (MockedStatic<CurrentUserUtil> userUtilMockedStatic = mockStatic(CurrentUserUtil.class)) {
```
enricocolasante, line 154: Could you remove the static mocks and fix the injection of the user in the setup?
I think the user is injected before
`user.setTeiSearchOrganisationUnits(Set.of(orgUnit1, orgUnit2));
    user.setOrganisationUnits(Set.of(orgUnit2));`

And then you can probably not find the search and capture scope that you need

ameenhere (approved):

PR #18362, opened 2024-08-19
feat: make getCurrentUserDetails() throw exception instead of returning NULL

netroms:
# Summary
The primary update is that getCurrentUserDetails() and getCurrentUsername() now throw exceptions when no user is logged in. In certain rare scenarios, such as when a user is not logged in yet in a public end-point, it’s recommended to first check using hasCurrentUser(). For background processes it's recommended to use the SystemUser as the authenticated user.

Previously, these methods returned NULL, which could indicate a background system process or be used as a convenient fallback when no user was set in the context, allowing tests to proceed as if they were run by a superuser. However, this introduced ambiguity in cases where NULL might represent a genuinely unauthenticated user, such as in public endpoints. This ambiguity increased code complexity by requiring frequent NULL checks and posed potential security risks due to unclear intent.

The `@Nonnull` annotations was also added to the methods, now making it clearer this is not intended to return NULL.
Some refactoring of the most low-hanging cases where `@Nonnull` is discovered by the static code analysis was refactored. There is still many cases left especially in the ACL related code, this will be addressed in follow up refactoring PRs.

## Automatic tests
CurrentUserUtilTest# testCurrentUsernameThrowsException
CurrentUserUtilTest# testCurrentUserDetailsThrowsException

## Manual tests
N/A


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseIdentifiableObject.java
```diff
@@ -420,8 +420,11 @@ public void setFavorites(Set<String> favorites) {
   @JsonProperty
   @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
   public boolean isFavorite() {
+    if (!CurrentUserUtil.hasCurrentUser()) {
+      return false;
+    }
     UserDetails user = CurrentUserUtil.getCurrentUserDetails();
-    return user != null && favorites != null && favorites.contains(user.getUid());
+    return favorites != null && favorites.contains(user.getUid());
```
jbee, line 427: The `favorites != null` part should be lifted up as a condition for early exit. In that case we do not even have to access the current user.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -47,12 +47,13 @@ private CurrentUserUtil() {
...
    * @return the current user's username
    */
+  @Nonnull
   public static String getCurrentUsername() {
     Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
     if (authentication == null
         || !authentication.isAuthenticated()
         || authentication.getPrincipal() == null) {
-      return null;
+      throw new RuntimeException("No authenticated user found");
```
jbee, line 56: maybe lets use `IllegalStateException` to be a bit more specific

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
```diff
@@ -147,8 +147,8 @@ public void prunePeriods() {
   @Override
   @Transactional
   public boolean pruneData(OrganisationUnit organisationUnit) {
-    if (CurrentUserUtil.getCurrentUsername() == null
-        || !CurrentUserUtil.getCurrentUserDetails().isSuper()) {
+    CurrentUserUtil.getCurrentUsername();
```
jbee, line 150: I think this line can be removed now?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
```diff
@@ -166,8 +166,8 @@ public boolean pruneData(OrganisationUnit organisationUnit) {
   @Override
   @Transactional
   public boolean pruneData(DataElement dataElement) {
-    if (CurrentUserUtil.getCurrentUsername() == null
-        || !CurrentUserUtil.getCurrentUserDetails().isSuper()) {
+    CurrentUserUtil.getCurrentUsername();
```
jbee, line 169: same here

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/event/data/AbstractAnalyticsServiceTest.java
```diff
@@ -72,7 +73,7 @@
  * @author Luciano Fiandesio
  */
 @ExtendWith(MockitoExtension.class)
-class AbstractAnalyticsServiceTest {
+class AbstractAnalyticsServiceTest extends TestBase {
```
jbee, line 76: I think it gets messy when mixing a mockito test with spring stuff. Maybe you can copy what you need instead or make a `MockitoTestBase`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataapproval/hibernate/HibernateDataApprovalStore.java
```diff
@@ -284,7 +284,7 @@
     User currentUser = userService.getUserByUsername(currentUsername);
 
     final String strArrayUserGroups =
-        (currentUser == null || CollectionUtils.isEmpty(currentUser.getGroups()))
+        (CollectionUtils.isEmpty(currentUser.getGroups()))
```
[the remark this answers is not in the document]
jbee: Adding an annotation should now help with this?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataelement/DataElementDefaultDimensionPopulator.java
```diff
@@ -107,9 +110,28 @@ public void executeInTransaction() {
...
+    context.setAuthentication(authentication);
+    SecurityContextHolder.setContext(context);
+  }
+
+  private static void clearContext() {
+    SecurityContext context = SecurityContextHolder.getContext();
+    if (context != null) {
+      SecurityContextHolder.getContext().setAuthentication(null);
+    }
+    SecurityContextHolder.clearContext();
```
jbee, line 135: This feels like code that does exist elsewhere. I get we need this as `static` so if we have not yet created such helpers as static methods these should be put somewhere so we can reuse this code

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/startup/ConfigurationPopulator.java
```diff
@@ -74,4 +90,20 @@ private void checkSecurityConfiguration() {
...
+    context.setAuthentication(authentication);
+    SecurityContextHolder.setContext(context);
+  }
+
+  private static void clearContext() {
+    SecurityContext context = SecurityContextHolder.getContext();
+    if (context != null) {
+      SecurityContextHolder.getContext().setAuthentication(null);
+    }
+    SecurityContextHolder.clearContext();
```
jbee, line 107: Indeed, here is at least one more usage of the same code. I think this is also similar or the same as I added in the `DefaultAuthenticationService` so even that might want to call the same static helper methods.

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/scheduling/RecordingJobProgressTest.java
```diff
@@ -53,13 +56,19 @@
  *
  * @author Jan Bernitt
  */
-class RecordingJobProgressTest {
+class RecordingJobProgressTest extends TestBase {
```
jbee, line 59: This is a pure JUnit test, for these I would also prefer we don't mix in a base class that suggests spring when it is not involved. We rather should call a static method that is directly imported.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/deduplication/DeduplicationServiceTest.java
```diff
@@ -163,6 +168,13 @@ void shouldBeAutoMergeable()
           PotentialDuplicateForbiddenException,
           ForbiddenException,
           NotFoundException {
+    SystemUser actingUser = new SystemUser();
+    Authentication authentication =
+        new UsernamePasswordAuthenticationToken(actingUser, "", actingUser.getAuthorities());
+    SecurityContext context = SecurityContextHolder.createEmptyContext();
+    context.setAuthentication(authentication);
+    SecurityContextHolder.setContext(context);
```
jbee, line 176: These lines look familiar :D I think this is the static helper again I pointed out or one that also should exist.

dhis-2/dhis-services/dhis-service-validation/src/main/java/org/hisp/dhis/validation/hibernate/HibernateValidationResultStore.java
```diff
@@ -290,11 +290,11 @@ private String getQueryRestrictions(ValidationResultQuery query) {
    * @return String to add restrictions to the HQL query.
    */
   private String getUserRestrictions(SqlHelper sqlHelper) {
-    UserDetails currentUserDetails = CurrentUserUtil.getCurrentUserDetails();
-    if (currentUserDetails == null || currentUserDetails.isSuper()) {
+    //    !CurrentUserUtil.hasCurrentUser() ||
```
jbee, line 293: remove?

dhis-2/dhis-services/dhis-service-validation/src/test/java/org/hisp/dhis/validation/ValidationResultStoreHqlTest.java
```diff
@@ -79,7 +81,7 @@
  *
  * @author Jan Bernitt
  */
-class ValidationResultStoreHqlTest {
+class ValidationResultStoreHqlTest extends TestBase {
```
jbee, line 84: Again, plain JUnit tests should not extend a base but import static methods

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapperTest.java
```diff
@@ -134,7 +133,7 @@ class EventRequestParamsMapperTest {
   public void setUp() throws ForbiddenException, NotFoundException, BadRequestException {
     User user = new User();
 
-    when(userService.getUserByUsername(CurrentUserUtil.getCurrentUsername())).thenReturn(user);
+    when(userService.getUserByUsername(null)).thenReturn(user);
```
jbee, line 136: This is surprising - it clearly is not what it was - is this intentional?

jbee (changes requested):
This looks really good. 
Very nicely removes a lot of annoying null checks.
Thanks for adding all the `@Nonnull` annotations.

It does need some cleanup with extracting some static helpers for reuse and importing static methods in plain JUnit and mockito tests instead of inheriting a base class.

Our `TestBase` has just too much spring-iness in it that is very confusing in the context of mockito or plain JUnit tests. I would like to avoid dragging in those details as they would need to be considered and deemed irrelevant. Also some stuff in our base class might not even work when not being in a spring based test.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseIdentifiableObject.java
jbee: I mean make a `if (favorites == null) return false`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataapproval/hibernate/HibernateDataApprovalStore.java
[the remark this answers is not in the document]
netroms: This whole part should be refactored to use UserDetails later

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -74,12 +77,13 @@ public static String getCurrentUsername() {
...
    *     unauthenticated
    */
+  @Nonnull
   public static UserDetails getCurrentUserDetails() {
     Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
     if (authentication == null
         || !authentication.isAuthenticated()
         || authentication.getPrincipal() == null) {
-      return null;
+      throw new RuntimeException("No authenticated user found");
```
jbee, line 86: you missed this one to replace with `IllegalStateException`

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/FileResourceControllerMockTest.java
```diff
@@ -37,18 +37,21 @@
...
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.mock.web.MockHttpServletResponse;
 
 @ExtendWith(MockitoExtension.class)
-class FileResourceControllerMockTest {
+class FileResourceControllerMockTest extends TestBase {
```
jbee, line 54: Missed this one to not extend

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/SharingControllerTest.java
```diff
@@ -53,7 +56,7 @@
  * @author Volker Schmidt
  */
 @ExtendWith(MockitoExtension.class)
-class SharingControllerTest {
+class SharingControllerTest extends TestBase {
```
jbee, line 59: and this

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataelement/DataElementOperandControllerTest.java
```diff
@@ -97,7 +99,7 @@
  */
 @MockitoSettings(strictness = Strictness.LENIENT)
 @ExtendWith(MockitoExtension.class)
-class DataElementOperandControllerTest {
+class DataElementOperandControllerTest extends TestBase {
```
jbee, line 102: and some more, maybe do a search ;)

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/service/GeoFeatureServiceMockTest.java
```diff
@@ -156,4 +169,12 @@ private OrganisationUnit createOrgUnitWithCoordinates() throws IOException {
     ou.setGeometry(new GeometryJSON().read(POINT));
     return ou;
   }
+
+  private static void injectUserInSecurityContext(UserDetails user) {
```
jbee, line 173: I think you now have a static helper for this or is it different?

jbee (approved):
You missed a couple but I approve assuming you change those as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/CurrentUserUtil.java
```diff
@@ -177,15 +170,11 @@ public static void setUserSetting(UserSettingKey key, Serializable value) {
...
-          userSettings.put(key, value);
-        } else {
-          userSettings.remove(key);
-        }
-      }
+    Map<String, Serializable> userSettings = currentUser.getUserSettings();
+    if (value != null) {
+      userSettings.put(key, value);
+    } else {
+      userSettings.remove(key);
```
david-mackessy, line 177: I know this is not part of this PR but seeing this seems a little odd.
The method's intention is to set a user setting, but if the value passed is null we remove the key. This sounds like a hidden side effect.

david-mackessy (approved):
great to see this

PR #18386, opened 2024-08-20
feat: Support XLSX in Analytics downloads[2.42-DHIS2-13907]

d-bernat:
**The old XLS format** has limitations on the number of rows (65,536) and columns (256) it can support. Additionally, it does not support newer features such as data validation, advanced data visualization, or more complex functions that were introduced in later versions of Excel.
**The new XLSX format** supports a much larger number of rows (1,048,576) and columns (16,384). It also allows for advanced features like conditional formatting, improved charting tools, complex formulas, and better integration with other applications and platforms.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
@@ -279,25 +279,24 @@ public class GridUtils {
 
   /** Writes a XLS (Excel workbook) representation of the given Grid to the given OutputStream. */
   public static void toXls(Grid grid, OutputStream out) throws IOException {
-    Workbook workbook = new HSSFWorkbook();
-
-    String sheetName =
-        filenameEncode(StringUtils.defaultIfEmpty(grid.getTitle(), XLS_SHEET_PREFIX + 1));
-
-    toXlsInternal(
-        grid,
-        workbook.createSheet(sheetName),
-        createHeaderCellStyle(workbook),
-        createCellStyle(workbook));
-
-    workbook.write(out);
-    workbook.close();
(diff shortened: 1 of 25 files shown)


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
```diff
@@ -293,6 +294,23 @@ public static void toXls(Grid grid, OutputStream out) throws IOException {
     workbook.close();
   }
 
+  /** Writes a XLSX (Excel workbook) representation of the given Grid to the given OutputStream. */
+  public static void toXlsx(Grid grid, OutputStream out) throws IOException {
+    Workbook workbook = new XSSFWorkbook();
+
+    String sheetName =
```
maikelarabori, line 301: The lines 301 to 311 are same as lines 284 to 294.
I suggest we extract this common code into a separate private method and reuse this new method on both places.

maikelarabori (commented):
It looks good in general. 👍 
I left one minor implementation suggestion, please take a look.

In addition, there are other Analytics download endpoints: AnalyticsController, TrackedEntityAnalyticsController and AnalyticsOutlierDetectionController classes. Would be important to make changes on those classes too...

Some e2e tests would be cool as well. You can take TrackedEntityQueryDownloadTest as example.

Thx!

maikelarabori (approved):

d-bernat:
> It looks good in general. 👍 I left one minor implementation suggestion, please take a look.
> 
> In addition, there are other Analytics download endpoints: AnalyticsController, TrackedEntityAnalyticsController and AnalyticsOutlierDetectionController classes. Would be important to make changes on those classes too...
> 
> Some e2e tests would be cool as well. You can take TrackedEntityQueryDownloadTest as example.
> 
> Thx!

Done

gnespolino (approved):

PR #18387, opened 2024-08-20
chore: Remove updateTrackedEntityLastUpdated [DHIS2-17714]

muilpp:
## Big picture

## This PR

- Removes the method `updateTrackedEntityLastUpdated`.
- Replaces the last usages of `TrackedEntityStore` with the manager.
- Removes `TrackedEntityService` and `TrackedEntityStore` since they are no longer used.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/DefaultTrackerBundleService.java
```diff
@@ -137,18 +137,39 @@ public void postCommit(TrackerBundle bundle) {
...
   private void updateTrackedEntitiesLastUpdated(TrackerBundle bundle) {
-    if (!bundle.getUpdatedTrackedEntities().isEmpty()) {
-      try {
-        trackedEntityService.updateTrackedEntityLastUpdated(
-            bundle.getUpdatedTrackedEntities(),
-            new Date(),
-            mapper.writeValueAsString(bundle.getUserInfo()));
-      } catch (JsonProcessingException e) {
-        throw new PersistenceException(e);
+    Set<String> updatedTrackedEntities = bundle.getUpdatedTrackedEntities();
```
enricocolasante, line 140: This is now run outside any transaction.
Before we had a @Transactional annotation on the `updateTrackedEntityLastUpdated` method in the old service, now we need have it somewhere.
We cannot annotate a private method with @Transactional so we either annotate `postCommit` method or we move this code to our `TrackedEntityService` and mark the method there.

This code with annotation shouldn't work so the fact that we have no failing tests seems to me that we are missing some test cases.

muilpp: That's a good point.
I added the `@Transactional` annotation on `postCommit`, I rather not have the `TrackedEntityService` do anything else than reading data.

There were some failing tests in the [LastUpdateImportTest](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/LastUpdateImportTest.java) that got fixed, even without the `@Transactional` annotation. I ran a manual test by spinning up a local instance and updating a TE, and the `lastUpdated` field was updated as expected.

As far as I know, the session object opens a connection to the database, and it lets you perform operations even without a transaction, but this opens the door to multiple problems in multi threading environments.
Postgres does also start a transaction by default, I'm not sure if we overrode that configuration, but that could be another reason why this worked.

Based on this, I think the tests are fine. Do you agree?

enricocolasante (approved):

ameenhere (approved):

PR #18397, opened 2024-08-21
chore: Remove ownership methods [DHIS2-17714]

muilpp:
## Big picture

## This PR

- Removes the methods `transferOwnership`, `assignOwnership` and `grantTemporaryOwnership` from org.hisp.dhis.trackedentity.TrackerOwnershipManager, they are moved into the tracker module.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -66,7 +66,7 @@
 class DefaultEnrollmentService implements EnrollmentService {
   private final EnrollmentStore enrollmentStore;
 
-  private final TrackerOwnershipManager trackerOwnershipAccessManager;
+  private final ApiTrackerOwnershipManager trackerOwnershipAccessManager;
```
enricocolasante, line 69: That can use the tracker one not the API

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -65,7 +65,7 @@
 class SecurityOwnershipValidator implements Validator<Enrollment> {
 
   @Nonnull private final AclService aclService;
-  @Nonnull private final TrackerOwnershipManager ownershipAccessManager;
+  @Nonnull private final ApiTrackerOwnershipManager ownershipAccessManager;
```
enricocolasante, line 68: Same here

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityOwnershipValidator.java
```diff
@@ -79,7 +79,7 @@ class SecurityOwnershipValidator implements Validator<org.hisp.dhis.tracker.impo
       "Event {} has no organisation unit assigned, so we skip user validation";
 
   @Nonnull private final AclService aclService;
-  @Nonnull private final TrackerOwnershipManager ownershipAccessManager;
+  @Nonnull private final ApiTrackerOwnershipManager ownershipAccessManager;
```
enricocolasante, line 82: Same here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -62,7 +62,7 @@
 public class DefaultTrackerAccessManager implements TrackerAccessManager {
 
   private final AclService aclService;
-  private final TrackerOwnershipManager ownershipAccessManager;
+  private final ApiTrackerOwnershipManager ownershipAccessManager;
```
enricocolasante, line 65: As this is the only place outside tracker where we are using `TrackerOwnershipManager`, wouldn't be easier to move this class in this same PR and we avoid the need to create  `ApiTrackerOwnershipManager`

muilpp: I can't move this class into tracker because it's used by the `DefaultTrackedEntityChangeLogService`, which is also outside our module, and it's in turn used by `DefaultProgramMessageService` and `DeliveryChannelStrategy`, which are also outside. This is one of the things I'd like discuss.

enricocolasante (approved):

ameenhere (approved):

PR #18398, opened 2024-08-21
fix: Set CSP header for attachments [DHIS2-17310](2.41)

enricocolasante:
Fixing `trackedEntityInstances/{teUid}/{attributeUid}/file` and `trackedEntityInstances/{teUid}/{attributeUid}/image` endpoints

- Adding attachment parameter in Content-Disposition response header
- Set CSP header as it is defined in `csp.header.value` configuration key


dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/deprecated/tracker/TrackedEntityControllerTest.java
```diff
@@ -89,19 +112,92 @@ public void setUp() throws BadRequestException, IOException {
...
             new TrackedEntityInstanceStrategyImpl(
                 trackedEntityInstanceSyncStrategy, trackedEntityInstanceAsyncStrategy),
-            userService);
+            userService,
+            config);
 
     mockMvc = MockMvcBuilders.standaloneSetup(controller).build();
+  }
+
+  @Test
```
ameenhere, line 127: Is the only difference between both these tests is the content type "image/png" vs "file/png"? If yes, could we use ParameterizedTest instead?

enricocolasante: Not just this.
`/image` needs to mock the input stream 
`when(fileResourceService.getFileResourceContent(fileResource))
        .thenReturn(new FileInputStream(file));`
 while `/file` doesn't need it

ameenhere (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/deprecated/tracker/TrackedEntityInstanceController.java
```diff
@@ -603,6 +608,8 @@ private void validateFileResource(FileResource fileResource, TrackedEntityAttrib
   private void setHttpResponse(HttpServletResponse response, FileResource fileResource) {
     response.setContentType(fileResource.getContentType());
     response.setContentLengthLong(fileResource.getContentLength());
-    response.setHeader(HttpHeaders.CONTENT_DISPOSITION, "filename=" + fileResource.getName());
+    response.setHeader(
+        HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=" + fileResource.getName());
```
JoakimSM, line 612: Not sure about adding `attachment` to images. Would be a regression at least.

enricocolasante: @JoakimSM It makes sense. I changed it back.

JoakimSM: Thanks, looks good now!

PR #18411, opened 2024-08-22
chore: Move trackedEntityAttributeValue service and store into tracker [DHIS2-17714]

enricocolasante:
## Big picture

## This PR

- Move `TrackedEntityAttributeValueService`, `DefaultTrackedEntityAttributeValueService` and `HibernateTrackedEntityAttributeValueStore` into tracker.
- Remove `TrackedEntityAttributeValueStore` interface as the store is only used by the service.
- Change `TrackedEntityAttributeValueDeletionHandler` to directly call the DB
- Remove unused methods


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentityattributevalue/TrackedEntityAttributeValueDeletionHandler.java
```diff
@@ -40,21 +39,19 @@
...
   protected void register() {
     whenVetoing(TrackedEntityAttribute.class, this::allowDeleteTrackedEntityAttribute);
   }
 
   private DeletionVeto allowDeleteTrackedEntityAttribute(TrackedEntityAttribute attribute) {
-    return attributeValueService.getCountOfAssignedTrackedEntityAttributeValues(attribute) == 0
-        ? ACCEPT
-        : VETO;
+    String sql =
+        "select 1 from trackedentityattributevalue where trackedentityattributeid = :id limit 1";
```
muilpp, line 54: do we need to add "limit 1" if `trackedentityattributeid` is the PK?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/trackedentityattributevalue/HibernateTrackedEntityAttributeValueStore.java
```diff
@@ -38,15 +37,14 @@
...
-import org.hisp.dhis.trackedentityattributevalue.TrackedEntityAttributeValueStore;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Repository;
 
 /**
  * @author Abyot Asalefew
  */
-@Repository("org.hisp.dhis.trackedentityattributevalue.TrackedEntityAttributeValueStore")
+@Repository
```
muilpp, line 47: even though most of the time we have only one implementation of a service and a store, we usually give it a name, I took it almost as a convention in this project... so this is just to understand why in this case the class name is enough

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentityattributevalue/TrackedEntityAttributeValueDeletionHandler.java
enricocolasante: Yes, this is a one-to-many relationship.
For a `TrackedEntityAttribute` we can have multiple `TrackedEntityAttributeValue`s, and we just want to check that it exists

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/trackedentityattributevalue/HibernateTrackedEntityAttributeValueStore.java
enricocolasante: No reason, I was following what I did for the service.
Putting it back

muilpp (approved):

ameenhere (approved):

PR #18425, opened 2024-08-23
chore: Move Changelog service to tracker module [DHIS2-17835]

muilpp:
The audit and changelog features are currently mixed together, making it hard to tell them apart since they use the same controller, and some services and stores are misnamed. Plus, there are different types of audits and changelogs, and not all are connected to the tracker. Some of them don't seem to be used at all.
This PR starts cleaning things up.

Here's what it does:
- Renames everything related to audits from `TrackedEntityChangeLog*` to `TrackedEntityAudit*`.
- Moves `TrackedEntityAuditService` and `TrackedEntityAuditStore` to the tracker module and sets up temporary `ApiTrackedEntityAuditService` and `ApiTrackedEntityAuditStore` versions because `DeliveryChannelStrategy` and `DefaultProgramMessageService` still need to use the `addTrackedEntityAudit` method.
- Deprecates all methods in `TrackedEntityAuditService`, since no one seems to be using them.

Ticket: https://dhis2.atlassian.net/issues/DHIS2-17835


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MaintenanceControllerTest.java
```diff
@@ -33,15 +33,15 @@
...
+import org.hisp.dhis.test.webapi.PostgresControllerIntegrationTestBase;
 import org.junit.jupiter.api.Test;
 
 /**
  * Tests {@link MaintenanceController} using (mocked) REST requests.
  *
  * @author Jan Bernitt
  */
-class MaintenanceControllerTest extends H2ControllerIntegrationTestBase {
+class MaintenanceControllerTest extends PostgresControllerIntegrationTestBase {
```
enricocolasante, line 44: Why is this change needed?

enricocolasante (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MaintenanceControllerTest.java
muilpp: It's actually not needed in this PR, but there's [this following PR](https://github.com/dhis2/dhis2-core/pull/18426) that continues the work on this topic, and it's needed there. Maybe we can leave it as it is, since it will be merged anyway.

zubaira (approved):

PR #18426, opened 2024-08-23
chore: Move TrackerAccessManager into tracker module [DHIS2-17835]

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
```diff
@@ -57,6 +66,45 @@ public Page<EventChangeLog> getEventChangeLog(
         eventUid, operationParams.getOrder(), pageParams);
   }
 
+  @Override
+  @Transactional(readOnly = true)
+  public List<TrackedEntityDataValueChangeLog> getTrackedEntityDataValueChangeLogs(
+      TrackedEntityDataValueChangeLogQueryParams params) {
```
teleivo, line 72: TrackedEntityDataValueChangeLogQueryParams says 

https://github.com/dhis2/dhis2-core/blob/34da9f9b0e1160dc7d8800ba3e0d7d6c5be25d4b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityDataValueChangeLogQueryParams.java#L44

is it audit or changelog? 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -45,6 +48,21 @@ Page<EventChangeLog> getEventChangeLog(
...
       throws NotFoundException, ForbiddenException;
 
+  /**
+   * @deprecated use {@link EventChangeLogService#getEventChangeLog} instead
+   */
+  @Deprecated(since = "2.41")
+  List<org.hisp.dhis.trackedentitydatavalue.TrackedEntityDataValueChangeLog>
+      getTrackedEntityDataValueChangeLogs(TrackedEntityDataValueChangeLogQueryParams params);
+
+  void addTrackedEntityDataValueChangeLog(
```
teleivo, line 58: Should we not also deprecate these 3 methods?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EventSavingSMSListener.java
```diff
@@ -318,7 +318,7 @@ private void createAndAddAudit(EventDataValue dataValue, DataElement dataElement
             dataValue.getProvidedElsewhere(),
             ChangeLogType.CREATE);
 
-    dataValueAuditService.addTrackedEntityDataValueChangeLog(dataValueAudit);
+    eventChangeLogService.addTrackedEntityDataValueChangeLog(dataValueAudit);
```
teleivo, line 321: I assume this can be removed when use the tracker importer, right? As somewhere in the importer we should be auditing changes to TEs, ...

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ChangeLogController.java
```diff
@@ -538,6 +535,7 @@ public RootNode getTrackedEnityInstanceChangeLog(
   }
 
   @GetMapping("trackedEntity")
+  @Deprecated(since = "2.42", forRemoval = true)
```
teleivo, line 538: Why is this one deprecated and which endpoint should be used instead?

teleivo (commented):
Is it true that we are going to remove `/api/audits/trackedEntityDataValue` (presumably in v43)?

As audit is not changelog, do we then still have audit related endpoints for tracker?

Why do we merge `TrackedEntityDataValueChangeLogService` into our changelog service if we are going to remove all its methods? 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
muilpp: This is change log, I'll remove that comment.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
muilpp: we don't have a replacement for the other ones yet, and I'm not sure we actually need to replace them at all.
this one was deprecated because of the new change log feature, which returns the data in a different format, but the other methods (to save and delete change logs) are still useful the way they are.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/EventSavingSMSListener.java
muilpp: Yes, the importer should end up doing this, I'll change that variable and method name, since they are still called audit...

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/ChangeLogController.java
muilpp: This endpoint returns audit entries for the given TE. These entries can either be READ or SEARCH, depending on the type of request the user made to access a particular TE.

There are a few issues related to this. The first is that some of the READ entries are not created by a user accessing a TE but are instead generated by system processes or validations. These entries provide no value to the end user.
The second issue is related to security. Currently, there is no ACL validation on this endpoint, allowing a user to access the audit of a TE that is not within their search scope.
Moreover, neither Android nor Frontend are using this endpoint.

For all these reasons, we are deprecating this endpoint with no replacement in sight until we come up with a better design. There are ongoing discussions regarding the security issue, depending on the outcome, we might remove this endpoint completely.

muilpp:
**Is it true that we are going to remove /api/audits/trackedEntityDataValue (presumably in v43)?**
Yes, we've got the new method to get event change logs in the [EventExportController](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportController.java#L326)

**As audit is not changelog, do we then still have audit related endpoints for tracker?**
The idea is not to have tracker audit related endpoints anymore. At least until we come with a better design, that provides value to the end user and has no security flaws. More on that [here](https://github.com/dhis2/dhis2-core/pull/18426#discussion_r1732342593)

**Why do we merge TrackedEntityDataValueChangeLogService into our changelog service if we are going to remove all its methods? 🤔**
We are not removing all those methods, the only one we will remove is `getTrackedEntityDataValueChangeLogs`, the rest stay.

enricocolasante (approved):

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
muilpp: that one we can deprecate, I'll do it in the following PR

PR #18430, opened 2024-08-23
chore: Move changelog into tracker [DHIS2-17714]

enricocolasante:
## Big picture

## This PR
- Move `TrackedEntityAttributeValueChangeLogService` and the store into tracker.
- Move `HibernateTrackedEntityDataValueChangeLogStore` into tracker.
- Delete `TrackedEntityAttributeDataValueChangeLogStore` interface as it is only used by the service in the same package.
- `TrackedEntityAttributeValueChangeLogStore` cannot be removed as it is used in `HibernatePotentialDuplicateStore`.
- Merge `TrackedEntityAttributeChangeLogService` and `TrackedEntityChangeLogService` in one service, since both have to do with change logs in tracked entities.
- Make `dhis-service-administration` module depend on `dhis-service-tracker` module.


teleivo (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
```diff
@@ -171,7 +171,7 @@ public boolean pruneData(DataElement dataElement) {
       return false;
     }
 
-    manager.delete(dataElement);
```
ameenhere, line 174: Curious about what this used to do, did it actually delete the metadata? 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -64,7 +65,57 @@ public class DefaultTrackedEntityChangeLogService implements TrackedEntityChange
...
+    // we do not have ACL API to check TE attributes.
+
+    Set<String> allUserReadableTrackedEntityAttributes =
+        trackedEntityAttributeService
+            .getAllUserReadableTrackedEntityAttributes(CurrentUserUtil.getCurrentUserDetails())
+            .stream()
+            .map(IdentifiableObject::getUid)
+            .collect(Collectors.toSet());
+
+    return attributeValueChangeLogs.stream()
```
zubaira, line 98: what if `allUserReadableTrackedEntityAttributes` is empty? can we return early ?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/DefaultMaintenanceService.java
enricocolasante: This was wrongly introduced a few PRs ago
https://github.com/dhis2/dhis2-core/commit/420326b1da4b1bd0ff4da6a8e96afdf66a871ba7#diff-10acfa7c6114045775d7939221d3d89bb04ecbf87cff67d6e35a8ca91fba9c1f
We shouldn't delete the `dataElement` here, only the attached values

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
enricocolasante: Maybe.
I am just moving some code in this PR, I would prefer not to touch anything at the moment

zubaira (approved):
LGTM

PR #18444, opened 2024-08-26
fix: only run generic search query if no result list present yet [DHIS2-17952] (2.40)

jbee:
based on cherry-pick from #18442 and manual merge conflict resolve. 

I opted to simply not backport the test scenario added as a lot of setup has changed around tests which makes it hard to backport tests.


david-mackessy (approved):

jbee:
I got confused with versions - 2.40 did not have the issue

PR #18447, opened 2024-08-26
fix: only run generic search query if no result list present yet [DHIS2-17952] (2.41.1)

jbee:
cherry-pick backport of #18442


Philip-Larsen-Donnelly (approved):
Approved as backport

david-mackessy (approved):

PR #18454, opened 2024-08-27
fix(DHIS2-13294): return app details from Apps install endpoint

kabaros:
implements [DHIS2-18027](https://dhis2.atlassian.net/browse/DHIS2-18027). 
needed for the fix for [DHIS2-13294](https://dhis2.atlassian.net/browse/DHIS2-13294) 

After manually installing an app (`POST /apps`), we wanted to be able to redirect the user to the app just installed. But since the API only returned an empty 204, the FE had no way to know the ID and other info of the installed App. This backend change exposes this information for the frontend return a 200 with the App object.

The ticket is relevant mainly when manually installing, but I also updated the install from AppHub to return the App object for consistency (and because we can make use of that info to avoid having an extra call to get the list of apps as we do now)

[DHIS2-13294]: https://dhis2.atlassian.net/browse/DHIS2-13294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-18027]: https://dhis2.atlassian.net/browse/DHIS2-18027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
@@ -294,6 +294,7 @@ public class DefaultAppManager implements AppManager {
     App installedApp = new App();
     if (appHubId == null) {
       installedApp.setAppState(AppStatus.NOT_FOUND);
+      return installedApp;
     }
 
     try {
@@ -301,6 +302,7 @@ public class DefaultAppManager implements AppManager {
       if (versionJson == null || versionJson.isEmpty()) {
         log.info(String.format("No version found for id %s", appHubId));
         installedApp.setAppState(AppStatus.NOT_FOUND);
+        return installedApp;
       }
       JsonString downloadUrlNode = JsonMixed.of(versionJson).getString("downloadUrl");
       if (downloadUrlNode.isUndefined()) {
@@ -308,6 +310,7 @@ public class DefaultAppManager implements AppManager {
             String.format(
                 "No download URL property found in response for id %s: %s", appHubId, versionJson));
         installedApp.setAppState(AppStatus.NOT_FOUND);
+        return installedApp;
       }
       String downloadUrl = downloadUrlNode.string();
       URL url = new URL(downloadUrl);
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
@@ -273,35 +273,41 @@ public class DefaultAppManager implements AppManager {
   }
 
   @Override
-  public AppStatus installApp(File file, String fileName) {
+  public App installApp(File file, String fileName) {
     App app = jCloudsAppStorageService.installApp(file, fileName, appCache);
+
+    log.info(
+        String.format(
+            "Installed App with ID %s (status: %s)", app.getAppHubId(), app.getAppState()));
 
     if (app.getAppState().ok()) {
       appCache.put(app.getKey(), app);
       registerDatastoreProtection(app);
     }
 
-    return app.getAppState();
+    return app;
   }
 
   @Override
-  public AppStatus installApp(UUID appHubId) {
+  public App installApp(UUID appHubId) {
+
+    App installedApp = new App();
     if (appHubId == null) {
-      return AppStatus.NOT_FOUND;
+      installedApp.setAppState(AppStatus.NOT_FOUND);
     }
 
     try {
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AppHubControllerTest.java
```diff
@@ -69,4 +71,20 @@ void testGetAppHubApiResponse() {
         "404 Not Found: \"{\"statusCode\":404,\"error\":\"Not Found\",\"message\":\"Not Found\"}\"",
         GET("/appHub/v37/test").content(HttpStatus.NOT_FOUND));
   }
+
+  @Test
+  void testAppHubInstallResponseContainsAppInfo() {
+    JsonArray apps = GET("/appHub").content();
+    JsonObject firstApp = apps.getObject(0).getArray("versions").getObject(0);
```
kabaros, line 78: I am not sure this is a good test .. but I couldn't find other tests that install apps to follow. 

It gets the information of the app to install by querying the API then installing the first app version it finds .. I think this should always work but feels a bit hacky (also need some pointer about the best way to query the JSON returned) .

david-mackessy: Hi Mozafar, it looks ok and should be reliable.
if you're looking to get a more explicit test, you could use an app/version which you know exists rather than getting and installing the first. Then explicitly asserting on version and versionId values.

There is the [JSON web domain](https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-support/dhis-support-test/src/main/java/org/hisp/dhis/test/webapi/json/domain) which might help with the readability and feel of the test as well (having dedicated JsonApp object to work with, with desired methods like getVersion etc.)

There is also [this class](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/test/webapi/Assertions.java) for some general-purpose web response assertions.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -273,35 +273,41 @@ public App getApp(String key, String contextPath) {
...
+  public App installApp(UUID appHubId) {
+
+    App installedApp = new App();
     if (appHubId == null) {
-      return AppStatus.NOT_FOUND;
+      installedApp.setAppState(AppStatus.NOT_FOUND);
     }
 
     try {
       String versionJson = appHubService.getAppHubApiResponse("v2", "appVersions/" + appHubId);
```
amcgee, line 300: I think this will send a request to `appVersions/null` if the appHubId is not found.  We need to keep the early-exit we had with the `return` above

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
```diff
@@ -153,21 +153,23 @@ public ResponseEntity<List<App>> getApps(@RequestParam(required = false) String
...
 
-    if (!status.ok()) {
-      String message = i18nManager.getI18n().getString(status.getMessage());
+    if (!appStatus.ok()) {
+      String message = i18nManager.getI18n().getString(installedApp.getAppState().getMessage());
 
       throw new WebMessageException(conflict(message));
     }
+
+    return ResponseEntity.ok(installedApp);
```
amcgee, line 172: Should this be `200` or still `201` since the app entity was created...?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppHubController.java
```diff
@@ -81,16 +80,19 @@ public String getAppHubApiResponse(@PathVariable String apiVersion, HttpServletR
...
+    App app = appManager.installApp(versionId);
+    AppStatus status = app.getAppState();
 
     if (!status.ok()) {
       String message = i18nManager.getI18n().getString(status.getMessage());
 
       throw new ConflictException(message);
     }
+
+    return ResponseEntity.ok(app);
```
amcgee, line 96: Same question as above - should this be `200` or `201`?

amcgee (changes requested):
Added some comments - I think the early exit needs to be maintained

amcgee (approved):

amcgee:
LGTM!

PR #18458, opened 2024-08-27
fix: index html caching

KaiVandivier:
This should fix these two issues, which both seem to be caused by HTML files being cached and served by the browser:
* [DHIS2-17782](https://dhis2.atlassian.net/browse/DHIS2-17782)
* [DHIS2-17804](https://dhis2.atlassian.net/browse/DHIS2-17804)

It seems that in v42, responses for HTML files are missing the `Cache-Control: no-cache, no-store, max-age=0, must-revalidate` headers that previous versions added to responses -- these changes reintroduce them

**To do**

- [ ] Consider HTML files to handle -- currently handling all HTML files; could instead handle only `index.html` and `plugin.html` for routes like `/dhis-web-` and `/api/apps/`, for example. Consider other HTML files, like custom reports or custom data entry forms
- [ ] Think about what the implications might be of doing this for non-app html files (or html files that aren't index or plugin).
- [ ] Need to think about where this should be placed... are bundled apps served with this servlet?
- [ ] Why did this stop working in v42?  Possibly due to restructuring to support embedded tomcat and/or removal of struts for bundled app serving
- [x] Rename this filter; maybe reuse the app override filter

**Testing**

The change can be tested by opening up apps and checking the response headers on HTML files -- they should include the `Cache-Control` headers above. _Other_ requests should not all have the same headers -- a few do, like `/api/me/dashboard` and `api/system/info`, but not all. NB: If opening up apps at a URL that hosted a previous v42 instance, you made need to get past previously cached versions of `index.html`, which can be done by reloading multiple times, hard reloading, or using the `Disable Cache` option in the Network panel of dev tools

Here are some examples of response headers after the changes:
Screenshot 2024-08-27 at 2 16 10 PM
Screenshot 2024-08-27 at 2 15 54 PM
Screenshot 2024-08-27 at 2 15 35 PM
Screenshot 2024-08-27 at 2 23 35 PM
Screenshot 2024-08-27 at 2 23 28 PM
6)

[DHIS2-17782]: https://dhis2.atlassian.net/browse/DHIS2-17782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-17804]: https://dhis2.atlassian.net/browse/DHIS2-17804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/HttpNoCacheFilter.java
```diff
@@ -0,0 +1,66 @@
...
+import org.hisp.dhis.webapi.utils.ContextUtils;
+import org.springframework.http.HttpMethod;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+/**
+ * Filter which enforces no cache for HTML pages like index pages to prevent stale versions being
+ * rendered in clients.
+ *
+ * @author Lars Helge Overland
```
amcgee, line 46:
```suggestion
 * @author Kai Vandivier
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DhisWebApiWebAppInitializer.java
```diff
@@ -139,6 +139,21 @@ public static void setupServlets(
...
+      headers using ContextUtils.setNoStore(response).  Think about what the implications might
+      be of doing this for non-app html files (or html files that aren't index or plugin).
+
+      Need to think about where this should be placed... are bundled apps served with this servlet?
+
+      Why did this stop working in v42?  Possibly due to restructuring to support embedded tomcat
+      and/or removal of struts for bundled app serving
+
+      Consider renaming this filter if it is very specific to app paths, or reusing the app override filter
+    */
```
amcgee, line 152:
```suggestion
    /* Intercept index.html, plugin.html, and other html requests to inject no-cache
      headers using ContextUtils.setNoStore(response).
    */
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/HttpNoCacheFilter.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.web.filter.OncePerRequestFilter;
+
+/**
+ * Filter which enforces no cache for HTML pages like index pages to prevent stale versions being
+ * rendered in clients.
+ *
+ * @author Kai Vandivier
+ */
+@Component
+public class HttpNoCacheFilter extends OncePerRequestFilter {
```
amcgee, line 49: Let's rename this to `HtmlFileNoCacheFilter` or something similar to make sure it reflects the behavior

amcgee (commented):
Looking good!  Few minor comments as discussed.

@netroms could you take a quick look at this to make sure we're doing the filtering in the right place?  Thanks!

amcgee (approved):
This change looks good to me!  Tested here https://dev.im.dhis2.org/pr-18458

netroms (approved):

jbee (changes requested):
No can't do it like this. I know of HTML we do want to cache. So the headers need to be added to the specific endpoints, not always for all HTML responses.

amcgee:
> No can't do it like this. I know of HTML we do want to cache. So the headers need to be added to the specific endpoints, not always for all HTML responses.

@jbee which HTML are you thinking of?

We discussed making the regex more targeted, we can do that if there are html files elsewhere that should be cached.  It is generally always problematic to cache application root html files, and this was the previous behavior.  @netroms and I also discussed that we need to address static asset caching headers more holistically so this will be updated before 42.

jbee:
@amcgee I was thinking about the `/openapi/openapi.html` and others in the `/openapi` path. But you are also allowed to use `/api/dataElements/openapi.html` and so forth.

jbee (approved):

KaiVandivier:
Thanks @jbee!

Regarding the Maps app, I tried switching the order of operations in the filter as recommended:
```java
chain.doFilter(request, response);

if (m.find() && HttpMethod.GET.matches(request.getMethod())) {
  ContextUtils.setNoStore(response);
}
```

But still the headers don't seem to get applied to the Maps app 🤔 any other ideas? @amcgee 

(This screenshot is from before changing the order, but it still looks the same after -- included for illustration)
new-regex-maps-not-working

jbee:
@KaiVandivier in that case I would revert the change of order again. Send me a message on slack so we can talk about the request this does and I maybe can help debugging it.

KaiVandivier:
The Maps HTML seems to be working fine on the IM deployment; maybe there was something weird on my local instance 🤔 

So I think this is nearly done; just discussing one last thing with Austin

Screenshot 2024-11-04 at 18 23 04

amcgee (approved):

PR #18473, opened 2024-08-28
fix: processing sms and test tracker event deletion DHIS2-17729

teleivo:
* Fixes bug: [UserDetails refactoring](https://github.com/dhis2/dhis2-core/pull/15727) broke SMS processing as SMS processing was done in another thread that does not have the `currentUser` set. The processing code uses the manager which requires the `currentUser` to be set.
* Process incoming SMS via jobs instead to fix the above bug. Every incoming sms will result in a job executed as the user sending the sms.
* Override `MessageSender` for Email and `SmsMessageSender` with `FakeMessageSender` so we can assert on error sms responses.
* Test that we can delete a tracker event via an sms and that we respond with an sms if the event cannot be deleted.

The previous in-memory queue of unprocessed incoming sms is replaced by a queue of jobs (one per sms). We do not have any numbers on the sms usage. Android can use the tracker sms to send data to DHIS2 in case users are offline. According to @zubaira sms are costly and are therefore not that likely to be used frequently. We can optimize later if it turns out that we should rather batch sms processing into a single job.

## Next

* Enable the 2 disabled `SmsOutboundControllerTest`s again. Untangle `Email/SmsMessageSender`. Most code is either dedicated to email or sms yet depends on `MessageSender` but then nugdes Spring to inject the email or sms sender. This does not make sense. I also could not get the test configuration to only override one or the other which is why I had to temporarily skip these 2 tests.
* Use tracker importer in all tracker sms listeners.
* Add more tests for sms processing. There are lots of classes involved in sms processing (deeper hierarchy) with indirection.
* In a future task: make `OutboundMessage` a `@Value` class
* In a future task: introduce generic error/warning `SmsResponse` codes so we can more easily map tracker errors to a response sms. Right now `https://github.com/dhis2/sms-compression/blob/1cecf7d97cc4b80788fd3b4a17a0de09a53ed96a/src/main/java/org/hisp/dhis/smscompression/SmsResponse.java#L36` are very specific. Tracker has its own status codes and each import can have multiple errors we need to be able to map to a single short sms.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/parameters/SmsInboundProcessingJobParameters.java
```diff
@@ -25,16 +25,22 @@
...
+import org.hisp.dhis.scheduling.JobParameters;
 
-public interface MessageQueue {
-  void put(IncomingSms message);
+@Getter
+@Setter
+@NoArgsConstructor
+public class SmsInboundProcessingJobParameters implements JobParameters {
+  @JsonProperty(required = true)
+  private long smsId;
```
jbee, line 41: This should be UID since job parameters are observable in the API. Then also add `@OpenApi.Property({UID, ?})` - have a look at other parameter classes that have ID parameters. `?` because I am not sure what the object type for SMS is :D

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
```diff
@@ -77,7 +76,8 @@ public StyleManager styleManager(
 
   @Bean("org.hisp.dhis.outboundmessage.OutboundMessageService")
   public DefaultOutboundMessageBatchService defaultOutboundMessageBatchService(
-      SmsMessageSender smsMessageSender, EmailMessageSender emailMessageSender) {
+      @Qualifier("smsMessageSender") MessageSender smsMessageSender,
+      @Qualifier("emailMessageSender") MessageSender emailMessageSender) {
```
jbee, line 80: I am confused. Didn't we discuss to use specific interfaces? This is like the opposite. Maybe I misunderstood?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/job/InboundSmsProcessingJob.java
```diff
@@ -25,72 +25,70 @@
...
+        (SmsInboundProcessingJobParameters) config.getJobParameters();
+    progress.startingProcess("Process incoming SMS with id {}", params.getSmsId());
 
-    while (message != null) {
-      log.info("Received SMS: " + message.getText());
+    IncomingSms sms = incomingSmsService.get(params.getSmsId());
+    if (sms == null) {
+      progress.failedProcess("No incoming SMS found for id {}", params.getSmsId());
+      return;
+    }
```
jbee, line 70: Usually I would advice on using stages in the process to execute the steps. But since this is very basic and we might want to keep it light weight it may be a good case where process level tracking is all we need. 

If you want to use stage level the work like `incomingSmsService.get` should be run via `runStage`. Look at usages of `JobProgress.nonNullStagePostCondition`. First the service all gets run via `runStage`. Then the `nonNullStagePostCondition` makes sure the result is non-null. This would happen as any exception in `runStage` would be caught and `null` would be returned. Wrapping that in `nonNullStagePostCondition` makes sure such issues fail eagerly as opposed to lazily at the next `startingStage` call. Obviously when you make the loading a stage you also need an `startingStage` before the `runStage` part.

If you decide to use stages the second part of finding and receiving via listener would be another stage.

jbee (approved):
Almost :) Just switch the ID for the job parameters to a UID `String`.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
teleivo: See first next step. I had to touch many files so the PR would not be as easy to review. I also could not get it to work. On the one hand most components either need email/sms but there are a couple that tie them together like `DefaultMessageService` and `DefaultOutboundMessageBatchService`. I need to analyze this again and figure out how we can make them replaceable in specific tests.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/job/InboundSmsProcessingJob.java
teleivo: That sounds to complicated for me right now 😅

enricocolasante (approved):

PR #15611, opened 2023-11-06
ci: add script and pipeline for rebuilding images [skip ci] [DEVOPS-294]

radnov:

PR #18483, opened 2024-08-29
ci: refactor docker rolling tags logic [skip ci]

radnov:
The logic in the `build-docker-image.sh` script (added with https://github.com/dhis2/dhis2-core/pull/15611) was relying on having the version being released in the `stable.json`.  However, in the [stable pipeline](https://github.com/dhis2/dhis2-core/blob/master/jenkinsfiles/stable) the `stable.json` is updated with the new version after the WAR and Docker image artefacts are already built.


Philip-Larsen-Donnelly (approved):
LGTM

PR #18491, opened 2024-08-30
fix: CI surefire report plugin issue 2.41

netroms:
## Summary
Fixes broken CI pipeline due to possible this surefire plugin issue: https://issues.apache.org/jira/projects/SUREFIRE/issues/SUREFIRE-2257

Upgrading to 3.5.1 was not possible yet, but should be done once it is released. In other words this PR should be reversed, when it is released.
Temporary solution is to downgrade version manually in the CI pipeline script [.github/workflows/run-tests.yml]

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -96,7 +96,7 @@ jobs:
           path: coverage.tar
           retention-days: 5
       - name: Generate surefire aggregate report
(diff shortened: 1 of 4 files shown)


gnespolino (approved):

radnov (approved):
LGTM, given that it's a temp solution.

PR #18508, opened 2024-09-03
fix: Load events to fix program rule variables calculation [DHIS2-17951]

enricocolasante:
`DefaultProgramRuleService` was wrongly trying to get all the events of an enrollment from the `TrackerPreheat` but they are not loaded in there and they need to be retrieved using `EventService`. Hence, the calculation of the correct dataValue based on the previous event was not working as previous event was not found unless it was present in the payload.

This change exposed some issues that we had in tests as some users didn't have org units in either the search scope and the capture scope and this is an invalid configuration that was getting validated by the event exporter.

Furthermore, when we inject different users in different tests, we need to make sure that we know which user is injected if we do not explicitly do so in the test. So in tests where we inject different users in tests we need to add a `@BeforeEach` method to inject the default user.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
```diff
@@ -202,10 +212,22 @@ private List<TrackedEntityAttributeValue> getAttributes(
...
-            .filter(e -> e.getEnrollment().getUid().equals(enrollmentUid))
-            .filter(e -> bundle.findEventByUid(e.getUid()).isEmpty());
+    Stream<Event> events;
+    try {
+      events =
+          eventService
+              .getEvents(
+                  EventOperationParams.builder()
+                      .eventParams(EventParams.TRUE)
+                      .orgUnitMode(ACCESSIBLE)
```
teleivo, line 222: Who is the currentUser/who set it when this is executed?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
```diff
@@ -202,10 +212,22 @@ private List<TrackedEntityAttributeValue> getAttributes(
...
+                      .enrollments(Set.of(enrollmentUid))
+                      .build())
+              .stream()
+              .filter(e -> e.getEnrollment().getUid().equals(enrollmentUid))
```
teleivo, line 226: Why is this necessary? 🤔 should we not only get events for this enrollment as you put that into the params?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -157,8 +164,6 @@ void shouldReturnChangeLogsWhenDataValueIsCreated() throws NotFoundException, Fo
   @Test
   void shouldReturnChangeLogsWhenDataValueIsDeleted()
       throws NotFoundException, IOException, ForbiddenException {
-    testAsUser("M5zQapPyTZI");
```
teleivo, line 160: Is this method still needed? should we remove it

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -119,9 +130,62 @@ void shouldNotImportWithWarningWhenAttributeWithSameValueIsAssignedByAssignRule(
...
+    assertHasOnlyWarnings(importReport, E1308);
+
+    Event event = manager.get(Event.class, "D9PbzJY8bZZ");
+
+    List<String> eventDataValues =
+        event.getEventDataValues().stream()
+            .filter(dv -> dv.getDataElement().equals("DATAEL00002"))
+            .map(dv -> dv.getValue())
+            .toList();
+    Assertions.assertContainsOnly(List.of("THIRD"), eventDataValues);
```
teleivo, line 156: add static import

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -119,9 +130,62 @@ void shouldNotImportWithWarningWhenAttributeWithSameValueIsAssignedByAssignRule(
...
+
+    ImportReport importReport = trackerImportService.importTracker(params, trackerObjects);
+    assertHasOnlyWarnings(importReport, E1308);
+
+    Event event = manager.get(Event.class, "D9PbzJY8bZZ");
+
+    List<String> eventDataValues =
+        event.getEventDataValues().stream()
+            .filter(dv -> dv.getDataElement().equals("DATAEL00002"))
+            .map(dv -> dv.getValue())
```
teleivo, line 154: a method reference should work right?

dhis-2/dhis-test-integration/src/test/resources/tracker/programrule/event_with_last_date.json
```diff
@@ -0,0 +1,79 @@
+{
+  "importMode": "COMMIT",
```
teleivo, line 2: This file is parsed into `TrackerObjects` so should it not only contain

```java
public class TrackerObjects implements Serializable {
  @JsonProperty @Builder.Default
  private final List<TrackedEntity> trackedEntities = new ArrayList<>();
  @JsonProperty @Builder.Default private final List<Enrollment> enrollments = new ArrayList<>();
  @JsonProperty @Builder.Default private final List<Event> events = new ArrayList<>();
  @JsonProperty @Builder.Default private final List<Relationship> relationships = new ArrayList<>();
}
```

?

This would make our files easier to read as we mostly put in fields that are ignored anyway like `createdAt`, ... 😅

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
```diff
@@ -119,9 +130,62 @@ void shouldNotImportWithWarningWhenAttributeWithSameValueIsAssignedByAssignRule(
     assertHasOnlyWarnings(importReport, E1310);
   }
 
+  @Test
+  void shouldImportEventWithLastDateAndCorrectlyAssignPreviousEventDataValue() throws IOException {
```
teleivo, line 134: It's hard to see from this test what we are testing and what state is changing in the process 😅 

Why is it that you expect THIRD here and SECOND in the next test? I would like to see the difference in input/config that causes that inside the test.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
enricocolasante: This is using current user.
Exporters uses current user to perform validations.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
enricocolasante: It is not, copy and paste code :grimacing:

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
enricocolasante: it is still used to inject other users in the test

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
enricocolasante: Yes, that should have been caught by Sonar :thinking:

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
enricocolasante: Done

dhis-2/dhis-test-integration/src/test/resources/tracker/programrule/event_with_last_date.json
enricocolasante: Done

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/programrule/ProgramRuleAssignActionTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
teleivo: but the program rules run as part of the import job which sets the current user to the user making the HTTP request, right?

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/DefaultProgramRuleService.java
enricocolasante: Yes yes, program rules run as part of import job, so they are using the same user.

PR #18516, opened 2024-09-04
fix: runStage null to non-null or cancel via exception (2.41)

jbee:
backport based on cherry-pick of #17846 followed by checkout of `JobProgress` and `RecordingJobProgress` from `master`.


jbee:
I make a different attempt

PR #18517, opened 2024-09-04
fix: runStage null to non-null or cancel via exception (2.41)

jbee:
backport of #17846


jbee:
giving up on backport

PR #18536, opened 2024-09-05
test: eliminate verbose error logging in org.hisp.dhis.webapi.control…

cjmamo:
…ler.OpenApiControllerTest.testGetOpenApiDocument_CodeGeneration


dhis-2/dhis-test-web-api/src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- NOTE: configuration changes will only have an effect AFTER recompiling dhis-support-test
```
teleivo, line 2: that comment is not true here as changes to this config will immediately affect the test runs in this module

teleivo (approved):

PR #18539, opened 2024-09-05
fix: Clean up completed data when status is not completed [DHIS2-16122]

enricocolasante:
`completedAt` and `completedBy` fields were correctly populated when an event transitioned to `COMPLETED` status but they were not cleaned when an event transitioned from `COMPLETED` to `ACTIVE`. (An event can only transition from `COMPLETE` to another status accepting data values as implemented in [StatusUpdateValidator](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/StatusUpdateValidator.java#L50-L59))


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/EventImportTest.java
```diff
@@ -0,0 +1,160 @@
...
+      throws IOException, ForbiddenException, NotFoundException {
+    TrackerImportParams params = TrackerImportParams.builder().userId(importUser.getUid()).build();
+    TrackerObjects trackerObjects = fromJson("tracker/te_enrollment_event.json");
+    trackerObjects.getEvents().get(0).setStatus(EventStatus.COMPLETED);
+
+    ImportReport importReport = trackerImportService.importTracker(params, trackerObjects);
+
+    assertNoErrors(importReport);
+
+    Event event = eventService.getEvent(UID.of("D9PbzJY8bJO"));
```
teleivo, line 82: Can you get the id from the `trackerObjects.getEvents().get(0)` to make sure/show that the event that was completed is the one you assert on?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/EventImportTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    assertNull(event.getCompletedBy());
+    assertNull(event.getCompletedDate());
+  }
+
+  @Test
+  void shouldDeleteCompletedDataWhenUpdatingAnEventWithStatusActive()
```
teleivo, line 107: I think you should merge this with the first test as you should assert that `completed` fields are set before than setting status to active and asserting they are not.

teleivo (commented):
Do we not yet have a test class where these can go? I am just surprised.

Looking at `EventTrackerConverterService` it is mapping Events. To me the permutations you test here should and can be unit tested. This logic does not benefit from these time consuming tests 😅  Having one/two that go all the way is ok but we should already have those 😅

enricocolasante:
> Do we not yet have a test class where these can go? I am just surprised.
> 
> Looking at `EventTrackerConverterService` it is mapping Events. To me the permutations you test here should and can be unit tested. This logic does not benefit from these time consuming tests 😅 Having one/two that go all the way is ok but we should already have those 😅

I added the unit tests for this change.
Our integration tests are not well organized, I couldn't find a test just for importing events.
As you were saying, I think that we should have a couple of import tests that go all the way for all the entity types and then move most of the integration tests to unit tests https://dhis2.atlassian.net/browse/DHIS2-16554

teleivo (approved):

PR #18567, opened 2024-09-10
fix: Fix unit test asserting on async log entry

enricocolasante:


teleivo (approved):
Have you tried changing the implementation to see the tests fail? Just to make sure

david-mackessy (approved):

enricocolasante:
> Have you tried changing the implementation to see the tests fail? Just to make sure

Yes it is failing.`assertInstanceOf` is catching this

PR #18568, opened 2024-09-10
fix: Test nullable org unit code for new rule-engine version [DHIS2-17715]

enricocolasante:
Bump `rule-engine` version.
New `rule-engine` allows to create an event or an enrollment with a `null` organisation unit code.
In this PR I added the related tests and refactor the whole test class.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -94,7 +94,7 @@
  */
 @Slf4j
 @RequiredArgsConstructor
-@Service("org.hisp.dhis.tracker.imports.programrule.engine.ProgramRuleEntityMapperService")
+@Service
```
teleivo, line 97: I think this is our convention https://github.com/dhis2/wow-backend/blob/master/guides/spring_conventions.md#bean-id

teleivo (approved):

ameenhere (approved):

PR #18580, opened 2024-09-11
fix: Stop server startup when no server base url [DHIS2-17980]

david-mackessy:
# Issue
Issue raised during development of the Push Analytics service (hasn't been raised before).
An invalid URL value was being returned when the `server.base.url` dhis conf is not set:
`"icon": "nullapi/icons/ambulance_positive/icon.svg"`

It raises the question `Should the dhis.conf setting 'server.base.url' be mandatory on startup?`

# Info
This config setting is used by at least 5 features (email, messaging, analytics, push analytics, interpretation). The assumption is that implementations already set this config in order for these features to work. Play instances already have this value set. A call to https://play.im.dhis2.org/dev/api/system/info?fields=instanceBaseUrl confirms this.

This change just makes this config mandatory and will make it easier to detect issues on startup.
An error will be logged on startup if it is not set:
```
08:09:10.896 ERROR [main] o.h.d.w.t.TomcatStarter: Error starting Tomcat context. Exception: java.lang.IllegalStateException. Message: dhis.conf config `server.base.url` requires a value on startup
```

Discussed with the team and we're happy to put this in for `42` but won't be backporting.

# Testing
Automated test added to ensure it's set on startup.
Other tests fail if it is missing also.


larshelge (approved):

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/DefaultDhisConfigurationProvider.java
```diff
@@ -102,6 +102,13 @@ public void init() {
     this.properties.setProperty(
         "connection.dialect", "org.hisp.dhis.hibernate.dialect.DhisPostgresDialect");
 
+    // ---------------------------------------------------------------------
+    // If mandatory config is not present stop startup
+    // ---------------------------------------------------------------------
+    String serverBaseUrl = this.properties.getProperty(ConfigurationKey.SERVER_BASE_URL.getKey());
```
teleivo, line 108: Just a thought: I assume there are other required configs. Could we declare that in `ConfigurationKey` with a `boolean required` and fail startup if any of these required ones without a default have no value? We might awkwardly fail at some later point when a user triggers code reading the config key for the first time 😅

david-mackessy: Yep, that's a good idea for a new feature. This PR has fallen through the cracks. Hope to get back to this smaller fix before the release.

larshelge (approved):

larshelge:
A few conflicts need resolve @david-mackessy .

david-mackessy:
closed until further discussion, to avoid merging accidentally.

PR #18594, opened 2024-09-13
feat: Allow for fractional expiry days in datasets

jason-p-pickering:
https://dhis2.atlassian.net/browse/DHIS2-18075


jbee (changes requested):
You are missing a flyway DB migration to update the column in dataset to allow floating point values.

david-mackessy (approved):
WOW flyway versioning doc should also be updated - https://github.com/dhis2/wow-backend/blob/master/coordination/flyway_versioning.md

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/Period.java
```diff
@@ -75,7 +74,7 @@ public static boolean isDateInTimeFrame(
     }
     ZoneId zoneId = ZoneId.systemDefault();
     Function<Date, ZonedDateTime> toZonedDate =
-        date -> ZonedDateTime.ofInstant(date.toInstant(), zoneId).with(LocalTime.MIN);
```
jbee, line 78: Have you checked that all callers now do want more than full day resolution? I think I added this method some time back to evaluate period related logic and periods are always full days.

jason-p-pickering: @jbee I have added another method which mirrors this one, but considers exact dates with times instead.

jason-p-pickering:
> You are missing a flyway DB migration to update the column in dataset to allow floating point values.

Added the flyway script.

jbee (approved):

PR #18597, opened 2024-09-13
feat: Support fractional timeliness days

jason-p-pickering:
The original issue was described in https://dhis2.atlassian.net/browse/DHIS2-9507. 

The use cases is basically when timeliness needs to calculated from a time other than midnight. In certain cases, in order to be considered timely, reports should be received at 12 noon. This effectively means 0.5 days after the end of the period. 

This PR adds support for fractional days in  for data set reporting timeliness calculations. While handling fractional days is maybe not great, it is expected that most of the time, users will continue to use whole days. However, in cases where timeliness should be based on some other time (say 10 AM or 5 PM), then an administrator should be able to figure how what that fraction is (10/24 or 17/24).


d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcCompletenessTableManager.java
```diff
@@ -256,7 +256,12 @@ private String getPartitionClause(AnalyticsTablePartition partition) {
 
   private List<AnalyticsTableColumn> getColumns() {
     String idColAlias = "concat(ds.uid,'-',ps.iso,'-',ous.organisationunituid,'-',ao.uid) as id ";
-    String timelyDateDiff = "cast(cdr.date as date) - ps.enddate";
+    int secondsInDay = (int) (24L * 60 * 60);
```
maikelarabori, line 259: We could extract it into a constant or move it to DateUtils.java (see DAYS_IN_YEAR and MS_PER_DAY, for example)

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcCompletenessTableManager.java
```diff
@@ -256,7 +256,12 @@ private String getPartitionClause(AnalyticsTablePartition partition) {
...
+
+    // The timeliness clock can only start ticking at the
+    String timelyDateDiff =
+        format(
+            "(EXTRACT(EPOCH FROM (cdr.date - ps.enddate)) - {}) / {}", secondsInDay, secondsInDay);
```
maikelarabori, line 264:
```suggestion
            "(extract(epoch from (cdr.date - ps.enddate)) - {}) / {}", secondsInDay, secondsInDay);
```

maikelarabori: Even though I would prefer uppercase for SQL/Functions, our internal standard uses lowercase.

maikelarabori (commented):
Looks good @jason-p-pickering. I just left minor comments for your review.

maikelarabori (approved):

PR #18598, opened 2024-09-13
fix: fix broken redirect when missing trailing slash

netroms:
## Summary
After the removal of Struts and restructuring of how resources are served in 2.42, the server's ability to redirect a resource URL without ending slash to one with, is no longer present.
This PR adds an interceptor to make sure resource URLs/apps are redirected properly with an appending slash.

### Automatic tests:
AuthTest#testRedirectMissingEndingSlash()

### Manual tests:
1. Log in as normal
2. Manually enter a URL like this: http://localhost:8080/dhis-web-dashboard
3. Observe you get redirected to: http://localhost:8080/dhis-web-dashboard/

### Jira:
[DHIS2-17493](https://dhis2.atlassian.net/browse/DHIS2-17493)

[DHIS2-17493]: https://dhis2.atlassian.net/browse/DHIS2-17493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/TrailingSlashInterceptor.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/TrailingSlashInterceptor.java
@@ -1,9 +1,36 @@
+/*
+ * Copyright (c) 2004-2024, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/interceptor/TrailingSlashInterceptor.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.hisp.dhis.webapi.mvc.interceptor;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+@Component
+public class TrailingSlashInterceptor implements HandlerInterceptor {
```
jbee, line 36: I feel an interceptor isn't the right way to solve this. We want this handling for a particular endpoint or few endpoints, don't we? This is affecting all endpoints and feels like a big risk of messing it up for other endpoints that have other ideas about slashes. Also interceptors are very in-transparent and hard to discover and understand mechanism that we really only should use when there is no better way.

jbee (approved):

vietnguyen (approved):

PR #18622, opened 2024-09-19
ci: fix build docker image script for non-latest major versions [skip ci]

radnov:
The major version of the version to be built was incorrectly compared to the latest major version we have in the `stable.json` file even if the version to be built was not the latest major version.

This was tested locally by the following command and with modifying the script to exit early without actually building anything, but just listing the rolling and immutable tags that would be created:
```
$ echo "42.0.0 41.2.0 41.1.0 40.5.0 40.4.2 40.4.0 39.5.1" | xargs -n 1 -I {} ./dhis-2/build-docker-image.sh -t "{}"

New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags.
Immutable tag:
42.0.0-20240919T143148Z
Rolling tags:
42.0.0 2.42.0.0 42.0 42 2.42.0 2.42

New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags.
Immutable tag:
41.2.0-20240919T143148Z
Rolling tags:
41.2.0 2.41.2.0 41.2 41 2.41.2 2.41

New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags.
Immutable tag:
41.1.0-20240919T143148Z
Rolling tags:
41.1.0 2.41.1.0 41.1 41 2.41.1 2.41

New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags.
Immutable tag:
40.5.0-20240919T143148Z
Rolling tags:
40.5.0 2.40.5.0 40.5 40 2.40.5 2.40

New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags.
Immutable tag:
40.4.2-20240919T143149Z
Rolling tags:
40.4.2 2.40.4.2 40.4 40 2.40.4 2.40

Immutable tag:
40.4.0-20240919T143149Z
Rolling tags:
40.4.0 2.40.4.0

New patch for a non-latest minor version, creating M.m tags.
Immutable tag:
39.5.1-20240919T143149Z
Rolling tags:
39.5.1 2.39.5.1 39.5 2.39.5
```

Please take a look at the example DHIS2 versions used for the test command above and let me know if there's any other case I might be missing. 🙏

### DIFF
--- a/dhis-2/build-docker-image.sh
+++ b/dhis-2/build-docker-image.sh
@@ -60,7 +60,7 @@ function create_rolling_tags() {
   )"
 
   if [[ "$major" -gt "$latest_major_version" ]] || \
-     [[ "$minor" -gt "$latest_minor_version" && "$major" -eq "$latest_major_version" ]] || \
+     [[ "$minor" -gt "$latest_minor_version" ]] || \
      [[ "$patch" -ge "$latest_patch_version" && "$minor" -eq "$latest_minor_version" ]]; then
     echo "New major, new minor or new/existing patch for the latest minor version, creating M.m and M tags."
     rolling_tags+=(


amcgee:
I'm not sure I follow the logic here...?  It looks like this will only create the alias tags for the "branch" of the latest major/minor versions... but we also want to create, for example, tag `40` and `40.6` when we publish `40.6.0` even though the "latest" release in stable.json is `41.1.0`... maybe I'm mis-reading the shell script?

Definitely agree with trying to create some tests for this!

radnov:
> I'm not sure I follow the logic here...? It looks like this will only create the alias tags for the "branch" of the latest major/minor versions... but we also want to create, for example, tag `40` and `40.6` when we publish `40.6.0` even though the "latest" release in stable.json is `41.1.0`... maybe I'm mis-reading the shell script?
> 
> Definitely agree with trying to create some tests for this!

I think you are misreading it indeed (not that it's easy to read, though 😬).

In the test I shared above I used a previous version of the stable.json that did not have the 40.5 version, so the test case would be equal to using 40.6 with the current version of the stable.json.

teleivo:
I thought we dropped the 2. prefix 🤔 
Can we drop the 2.x variants at least for 42 and 41?

radnov:
> I thought we dropped the 2. prefix 🤔 Can we drop the 2.x variants at least for 42 and 41?

Good question. @Philip-Larsen-Donnelly what/where was the plan for dropping the prefix? I can't remember.

teleivo:
> > I thought we dropped the 2. prefix 🤔 Can we drop the 2.x variants at least for 42 and 41?
> 
> Good question. @Philip-Larsen-Donnelly what/where was the plan for dropping the prefix? I can't remember.

It's gone from https://dhis2.org/downloads/ for all major versions. Lets do it for Docker as well 🥳

Philip-Larsen-Donnelly (approved):

PR #18624, opened 2024-09-19
ci: revert to using the frontend plugin to bundle apps

teleivo:
temporary revert

- https://github.com/dhis2/dhis2-core/pull/18612
- https://github.com/dhis2/dhis2-core/pull/18617

for now. For some reason the webapps are cloned but not packaged into the `dhis.war` in CI.


radnov (approved):
LGTM

PR #18632, opened 2024-09-20
ci: merge maven module dhis-web-app with dhis-web-server

teleivo:
We want to reduce the number of maven modules and simplify our build.

This PR merges the maven module `dhis-web-app` into `dhis-web-server`. `dhis-web-app` was [bundling (`git clone`)](https://github.com/dhis2/dhis2-core/blob/cb0dad1ff6d886c2a8ab1f6c5e6028cac60af5de/dhis-2/dhis-web-apps/scripts/bundle-apps.js#L35) the [web apps](https://github.com/dhis2/dhis2-core/blob/cb0dad1ff6d886c2a8ab1f6c5e6028cac60af5de/dhis-2/dhis-web-apps/apps-to-bundle.json#L2) into a war. That war was then extracted again into the `dhis.war` in the `dhis-web-server` module. That is unnecessarily complicated.

- `mvn package` builds `dhis.war` for application servers like Tomcat
  - add jib profiles if you also want to build it into a Docker image
- `mvn package --activate-profiles embedded` builds an executable `dhis.war` with embedded Tomcat

## Improvements

* Expect Node.js to be present on a machine building DHIS2 (dev or CI). Installed on [GitHub](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#nodejs) and [Jenkins](https://github.com/dhis2-sre/jenkins/blob/6dc024d083a81403c732e6364d35b4ebc007e203/agents/ec2/jdk17/cloud-config.yaml#L51). That fixes the frequent build failures `dhis-web-apps: Could not download Node.js: Could not download https://nodejs.org/dist/v16.13.2/node-v16.13.2-linux-x64.tar.gz: 
Premature end of Content-Length delimited message body (expected: 32,399,497; received: 2,269,184)`.
* Don't exclude modules when running unit, integration, ... tests like `-pl -dhis-web-server` as they might have tests that are then just not run. As it happened with a `dhis-web-server` test.
* Remove the `jetty` profile and embedded jetty as we agreed on only using tomcat.
* Make the embedded Tomcat profile clearer by calling it `embedded` and use it in `./dhis-2/run-api.sh`
* Use `pluginManagement` in the `dhis-web-server/pom.xml` to define default version and config for plugins to reuse it in different profiles.
* The default `mvn package` should build the `dhis.war` we ship. That is a `dhis.war` that can be deployed into a Tomcat. If you want to use the embedded Tomcat add `--activate-profiles embedded`.
* Use long CLI args in our build scripts as it's easier to remember what they are for.

## Tests

* deployed this PR using the instance manager: that uses the Docker image built in the api test workflow on GitHub
* `./dhis-2/build-dev.sh` which builds the Docker image and ran it locally
* build `dhis.war` locally via `mvn clean package` and run it locally in Tomcat
- build `dhis.war` on Jenkins and run it locally in Tomcat
- build and run embedded Tomcat `mvn clean package -f dhis-2/pom.xml -DskipTests=true --activate-profiles embedded` then `DHIS2_HTTP_PORT=8084 DHIS2_HOME=~/.local/dhis2 java -jar 
~/code/dhis2/core/dhis-2/dhis-web-server/target/dhis.war`
- ran `./dhis-2/run-api.sh -d ~/.local/dhis2 -p 8084` which uses embedded Tomcat
- tested the Jenkins pipeline changes with @radnov before merging

## Questions for later

* why do we have 2 `velocity.properties`?
* why do we include tomcat classes and libraries in our `dhis.war` if we do not build the embedded Tomcat?
* should we go with a more maven like folder layout https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html static content should go into `src/main/webapp`


radnov (approved):
LGTM

netroms (approved):

PR #18659, opened 2024-09-24
chore: Harmonize user management in tracker [DHIS2-18001]

enricocolasante:
Harmonize user management in validator phase of tracker import.
- Get `UserDetails` from `CurrentUserUtil.getCurrentUserDetails()`.
- Pass `UserDetails` as a parameter to private methods.
- Remove all `checkNotNull` preconditions as they are already verified by the logic of validation in the importer (eg. when validating events in [DateValidator](https://github.com/dhis2/dhis2-core/blob/df865c5763672e0f72a2b1e53bce186f449c8275/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidator.java#L86) we know `Program` is not null as it was already validated in [MandatoryFieldsValidator](https://github.com/dhis2/dhis2-core/blob/df865c5763672e0f72a2b1e53bce186f449c8275/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/MandatoryFieldsValidator.java#L66). And from the order defined in [EventValidator](https://github.com/dhis2/dhis2-core/blob/df865c5763672e0f72a2b1e53bce186f449c8275/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/EventValidator.java#L57) we know that the validation will stop if the `Program` is null before validating the dates.)

**NEXT**
- Harmonize exporters


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -75,7 +73,6 @@ private ValidationUtils() {
 
   public static void validateGeometry(
       Reporter reporter, TrackerDto dto, Geometry geometry, FeatureType featureType) {
```
teleivo, line 75: Add `@Nonnull` to help us remember and help static analysis. Here and where applicable.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
```diff
@@ -71,15 +69,15 @@ class SecurityOwnershipValidator implements Validator<Enrollment> {
   public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollment) {
     TrackerImportStrategy strategy = bundle.getStrategy(enrollment);
     TrackerPreheat preheat = bundle.getPreheat();
-    User user = bundle.getUser();
+    UserDetails user = CurrentUserUtil.getCurrentUserDetails();
```
teleivo, line 72: Not sure if this call is expensive. Might also be a matter of style but now that we call this instead of passing it into methods we might as well call this in private methods 🤷🏻 If your not unit testing a function it saves a param of which we have many.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
enricocolasante: Done.

ameenhere (approved):

PR #18660, opened 2024-09-24
chore: Remove userid from tracker import params [DHIS2-18001]

enricocolasante:
`TrackerImportService` is always executed by the current user and can be retrieved using `CurrentUserUtil.getCurrentUserDetails()`. No matter if it runs in a `Job`, the current user is always correctly set.

This PR removes the `userId` field from the `TrackerImportParams` used to set the `User` for the `TrackerImportService` execution.
Instead, the `User` object is always created from `CurrentUserUtil.getCurrentUserDetails()`.

**NEXT**
- Harmonize exporters

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
@@ -82,7 +82,7 @@ public class DefaultTrackerImportService implements TrackerImportService {
   @IndirectTransactional
   public ImportReport importTracker(
       TrackerImportParams params, TrackerObjects trackerObjects, JobProgress jobProgress) {
-    User user = trackerUserService.getUser(params.getUserId());
+    User user = trackerUserService.getCurrentUser();
 
     jobProgress.startingStage("Running PreHeat");
     TrackerBundle trackerBundle =
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportParams.java
@@ -43,9 +43,6 @@ import org.hisp.dhis.tracker.imports.bundle.TrackerBundleMode;
 @NoArgsConstructor
 @AllArgsConstructor
 public class TrackerImportParams implements JobParameters {
-  /** User uid to use for import job. */
-  @JsonProperty private String userId;
-
   /** Should import be imported or just validated. */
   @JsonProperty @Builder.Default
   private final TrackerBundleMode importMode = TrackerBundleMode.COMMIT;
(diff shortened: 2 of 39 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EventImportValidationTest.java
```diff
@@ -89,9 +90,14 @@ void setUp() throws IOException {
             params, fromJson("tracker/validations/enrollments_te_enrollments-data.json")));
   }
 
+  @AfterEach
```
ameenhere, line 93: Just curious. Why do we rely on AfterEach instead of BeforeEach?

enricocolasante: Because I forgot to change it.

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -82,7 +82,7 @@ private PersistenceReport commit(TrackerImportParams params, TrackerBundle track
   @IndirectTransactional
   public ImportReport importTracker(
       TrackerImportParams params, TrackerObjects trackerObjects, JobProgress jobProgress) {
-    User user = trackerUserService.getUser(params.getUserId());
+    User user = trackerUserService.getCurrentUser();
```
muilpp, line 85: In the next iteration of this epic, we will replace instances of the `User` class such as this one, with `UserDetails`, correct?

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
enricocolasante: Yes, that is the idea.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportParams.java
```diff
@@ -43,9 +43,6 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class TrackerImportParams implements JobParameters {
-  /** User uid to use for import job. */
-  @JsonProperty private String userId;
```
teleivo, line 47: JobParameters are stored as JSONB in the DB. They usually need a migration. Does this change not need one?

enricocolasante: I didn't think about it but probably this is ok.
This `userId` was set when mapping `ImportRequestParams` to `TrackerImportParams` and the first one didn't have a `userId`.
I am going to double check how the `JobParameters` object is created and if this change can cause any issue.

PR #18669, opened 2024-09-25
ci: run all test on mvn test by default

teleivo:
Before we had to add `--activate-profiles -default` which excludes the `default` profile to run all tests. That is confusing. The default is now that we run all types of tests by default. This way we cannot miss any tests in a pipeline due to forgetting to remove the default profile. So with these changes

* `mvn test` - runs all tests
* `mvn test --activate-profiles unit-test` - only runs unit tests
* `mvn test --activate-profiles integration-test` - only runs postgres integration tests annotated with `@Integration`
* `mvn test --activate-profiles integration-h2-test` - only runs H2 integration tests annotated with `@IntegrationH2`

On Jenkins we simply do `mvn clean install` which runs all tests and creates the war. We can do that since the machines running Jenkins have more resources than the ones on GH. On GitHub we already ran each of these types of tests in a concurrent job. It should now be easier to understand as each category has its own maven profile.

* `AuthTest` was not properly classified as an integration test. Now it will run with all and in the `integration-test` profile.
* Disable the jacoco test coverage agent when not running tests.
* Use the same version for the `maven-surefire-report-plugin` as the `maven-surefire-plugin` running tests and producing xml files the report plugin should be able to parse.
* Remove unused `dev` profile in `pom.xml`.
* Use long form options in scripts so they are easier to read without a man page/help.

## Cross-check

I made sure the number of tests when no profile is active and each profile is run individually adds up.

### Method

Run all tests and each category on its own

* all `mvn clean test --file ./dhis-2/pom.xml`
* unit `mvn clean test --file dhis-2/pom.xml --activate-profiles unit-test`
* integration `mvn clean test --file dhis-2/pom.xml --activate-profiles integration-test`
* integration h2 `mvn clean test --file dhis-2/pom.xml --activate-profiles integration-h2-test`

from the same [commit](https://github.com/dhis2/dhis2-core/tree/ae2685320bb330b1d44e7c3a10375c804f41187f).

For each of the runs capture the build log and the surefire aggregate report

    mvn surefire-report:report-only -Daggregate=true --file ./dhis-2/pom.xml

Get the totals for all and each category via build log and surefire aggregate reports. I am only posting the aggregate report numbers as they are the same as in the build logs.

| Profile              | Tests | Errors | Failures | Skipped | Success Rate | Time    |
|----------------------|-------|--------|----------|---------|--------------|---------|
| unit-test            | 4477  | 0      | 0        | 4       | 99.9%        | 182.6 s |
| integration-test     | 3606  | 0      | 0        | 109     | 97.0%        | 664.3 s |
| integration-h2-test  | 1397  | 0      | 0        | 8       | 99.4%        | 291.2 s |
|----------------------|-------|--------|----------|---------|--------------|---------|
| **No profile (all tests)** | **9480** | **0**    | **0**    | **121**  | **98.7%**  | **1222 s**  |

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,8 +21,8 @@ jobs:
           java-version: 17
           distribution: temurin
           cache: maven
-      - name: Test core
-        run: mvn clean verify --threads 2C --batch-mode --no-transfer-progress --update-snapshots --file ./dhis-2/pom.xml
+      - name: Run unit tests
+        run: mvn clean verify --threads 2C --batch-mode --no-transfer-progress --update-snapshots --file ./dhis-2/pom.xml --activate-profiles unit-test
         timeout-minutes: 30
       - name: Report coverage to codecov
         uses: codecov/codecov-action@v3
@@ -65,7 +65,7 @@ jobs:
           distribution: temurin
           cache: maven
       - name: Run integration tests
-        run: mvn clean verify --threads 2C --batch-mode --no-transfer-progress -Pintegration --update-snapshots --file ./dhis-2/pom.xml
+        run: mvn clean verify --threads 2C --batch-mode --no-transfer-progress --update-snapshots --file ./dhis-2/pom.xml --activate-profiles integration-test
         timeout-minutes: 30
       - uses: actions/upload-artifact@v4
         name: Upload test logs on failure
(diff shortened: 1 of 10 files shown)


jbee (approved):
I like it. :+1:

david-mackessy (approved):

enricocolasante (approved):

radnov (approved):
Nice work! 👏

PR #18677, opened 2024-09-26
fix: Remove user field from tracker preheat [DHIS2-18001]

enricocolasante:
Remove user fields from `TrackerPreheat`.
`NotesConverterService` needs `TrackerUserService` to correctly create a `Note` object. This is a temporary solution that will be fixed in upcoming PRs.

**NEXT**
- Harmonize exporters


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
```diff
@@ -201,10 +202,10 @@ private Event from(
       result.setUid(!StringUtils.isEmpty(event.getEvent()) ? event.getEvent() : event.getUid());
       result.setCreated(now);
       result.setStoredBy(event.getStoredBy());
-      result.setCreatedByUserInfo(preheat.getUserInfo());
+      result.setCreatedByUserInfo(UserInfoSnapshot.from(CurrentUserUtil.getCurrentUserDetails()));
```
teleivo, line 205: I am not sure how some of these classes are currently tested. Changes like this make them less easy to unit test. Not sure what the strategy is but if we want to unit test such transformations we should pass the user into the function.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
```diff
@@ -377,7 +378,10 @@ private void delete(
             : entityManager.merge(trackedEntityAttributeValue));
 
     logTrackedEntityAttributeValueHistory(
-        preheat.getUsername(), trackedEntityAttributeValue, trackedEntity, ChangeLogType.DELETE);
+        CurrentUserUtil.getCurrentUsername(),
```
muilpp, line 381: I guess it's just a matter of preference, but why import this method statically?

muilpp (approved):

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterServiceTest.java
```diff
@@ -80,7 +82,10 @@ class EventTrackerConverterServiceTest extends TestBase {
 
   private static final Date today = new Date();
 
-  private final NotesConverterService notesConverterService = new NotesConverterService();
+  @Mock public TrackerUserService trackerUserService;
```
ameenhere, line 85: I thought Mocking was banned from today 😆

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/converter/EventTrackerConverterService.java
enricocolasante: Maybe we can discuss this and agree on the best way to use the user and have testable code.
At the moment, we need to inject the user in the tests.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
enricocolasante: Yeah, I can add static imports

PR #18678, opened 2024-09-26
fix: Remove user fields from tracker bundle [DHIS2-18001]

enricocolasante:
This is last PR for the importer.

- Remove `TrackerUserService`.
- Remove `user` and `userInfo` fields from `TrackerBundle`
- Adding a `CurrentUserSupplier` that will add current user to Preheat and it will be used to populate the user info for `Note` object. (Unfortunately we have a few different ways to represent and store a user in the DB. `Note` needs a `User` object instead of the `UserInfoSnapshot` object used by tracker objects)

**NEXT**
- Harmonize exporters


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -82,12 +79,10 @@ private PersistenceReport commit(TrackerImportParams params, TrackerBundle track
   @IndirectTransactional
   public ImportReport importTracker(
       TrackerImportParams params, TrackerObjects trackerObjects, JobProgress jobProgress) {
-    User user = trackerUserService.getCurrentUser();
```
teleivo, line 85: I think we should assert that there is a current user at the start of our import. It would be a programmer error not to have it set. We should rather fail at the start of the import than later down the line.

teleivo (approved):
I think we should discuss when to pass in the user as a param vs calling getCurrent... after the standup. You are changing lots of files that we need to change again if we settle on a different strategy 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
muilpp: Generally I would agree, but I think this case is an exception.
Since we are not validating user input, I think we should assume it's always present and use it as such.

muilpp (approved):

PR #18683, opened 2024-09-27
fix: use tracker importer in SimpleEventSMSListener DHIS2-17729

teleivo:
As a reference [this is](https://github.com/dhis2/dhis2-core/blob/998f7cc3c5a1271fd7742adf1253521f8e66e234/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/listener/SimpleEventSMSListener.java#L98) the `SimpleEventSMSListener` processing logic before we made our changes to sms processing. You can see that users were able to create an event in an event program. The logic of creating the dummy enrollment into the event program is done by the metadata importer.

This is how the [SimpleEventSmsSubmission](https://github.com/dhis2/sms-compression/blob/1cecf7d97cc4b80788fd3b4a17a0de09a53ed96a/src/main/java/org/hisp/dhis/smscompression/models/SimpleEventSmsSubmission.java#L43) sms is encoded.

* Use the tracker importer in `SimpleEventSMSListener`.
* Replace mock based test with integration test.

## Next

These are the tracker related compression based SMS we need to adapt so they use the importer:

- `ENROLLMENT`        => `EnrollmentSMSListener` ✅ (https://github.com/dhis2/dhis2-core/pull/18595)
- `SIMPLE_EVENT`      => `SimpleEventSMSListener` (this PR)

- `PROGRAM_STAGE_DATAENTRY_PARSER` => `ProgramStageDataEntrySMSListener`
- `EVENT_REGISTRATION_PARSER` => `SingleEventListener`


enricocolasante (approved):
It is not really tracker importer that creates the dummy enrollment, unfortunately it is metadata importer when creating the program. What we have in tracker importer is a preprocessor that will assign the correct dummy enrollment from the event program.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/SimpleEventSMSListener.java
```diff
@@ -98,95 +84,27 @@ public SimpleEventSMSListener(
...
-            subm.getCoordinates());
-    if (!errorUIDs.isEmpty()) {
-      return SmsResponse.WARN_DVERR.setList(errorUIDs);
-    } else if (subm.getValues() == null || subm.getValues().isEmpty()) {
-      // TODO: Should we save the event if there are no data values?
-      return SmsResponse.WARN_DVEMPTY;
-    }
-
-    return SmsResponse.SUCCESS;
+    // TODO(DHIS2-18003) we need to map tracker import report errors/warnings to an sms
```
enricocolasante, line 106: One other point in favor of separating program events and tracker events API is that in this listener if the user send a tracker program, the validation error from the tracker importer will be that the enrollment is missing, instead of wrong program type.

muilpp (approved):

PR #18707, opened 2024-09-30
fix: Exclude default category types [DHIS2-17564]

david-mackessy:
# Issue
The exclude defaults request param feature is not working (default types are being returned)
The intention of this feature is to exclude any default category types from payloads:
- `Category`
- `CategoryCombo`
- `CategoryOption`
- `CategoryComboOption`

## Current behaviour
So when sending a request like `GET` `/api/dataElements?defaults=EXCLUDE`, the returned payload includes default `categoryCombos` in any `DataElement` which has one.
e.g.
```
{
"dataElements": [
        {
            "name": "Low birth weight in community",
            "id": "Ce6kAvaHsxR",
            "categoryCombo": {
                "name": "default",
                "id": "bjDvmb4bfuf"
            }
        }
]}
```
## Expected behaviour
The returned payload should still include all `DataElement`s, but any default `categoryCombo` should not be present (should be removed)
e.g.
```
{
"dataElements": [
        {
            "name": "Low birth weight in community",
            "id": "Ce6kAvaHsxR"
        }
]}
```

# Cause
It seems that there are 2 FieldFilterServices ([old](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/fieldfilter/DefaultFieldFilterService.java) & [new](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java))
The old one has logic for excluding defaults but it seems to have been partially phased out.
The new FieldFilterService had no logic for excluding defaults. This new one seems to be the main field filter service.

# Fix
The surface area for this type of issue is very large. The `AbstractFullReadOnlyController` & `MetadataExportService` have been updated so that we now parse the query param `defaults=EXCLUDE` and ensure it is passed to the appropriate place to decide if defaults should be excluded. This seems to satisfy the majority of calls. If there are edge cases, they are hidden well and they can be fixed when raised.

# Tests
Could not find any existing tests for this feature which is why this bug probably went unnoticed for a while.

## Automated
There are 2 types of EXCLUDE default processing:
1. requests made to the defaults category types themselves:
  - `/api/categories`
  - `/api/categoryCombo`
  - `/api/categoryOption`
  - `/api/categoryOptionCombo`
  - these endpoints were working correctly (QueryService handles this) but had no tests. Tests have been added for each.
2. requests made for other metadata which contain category types:
  - `/api/dataElements?fields=categoryCombos` will contain `categoryCombo` fields
  - `/api/metadata?fields=id,name,categoryCombos[id,name]&dataElements=true` will contain `categoryCombo` fields
  - `/api/categoryCombos?fields=categories` will contain `category` fields
  - `/api/categories?fields=categoryOptions` will contain `categoryOptions` fields
  - `/api/categoryCombos?fields=categoryOptionCombos` will contain `categoryOptionCombo` fields
  - these endpoints were broken. Tests have been added for each of these.


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/StreamingJsonRoot.java
```diff
@@ -39,11 +40,14 @@
...
+@AllArgsConstructor
 @JsonRootName(value = "root", namespace = DxfNamespaces.DXF_2_0)
 public class StreamingJsonRoot<T> {
   private final Pager pager;
 
   private final String wrapperName;
 
   private final FieldFilterParams<T> params;
+
+  private boolean excludeDefaults;
```
david-mackessy, line 52: The possibility of including this value as a field filter was explored but I think this just makes things even more complicated and messy. Keeping them as separate things makes changes easier & the features easier to reason about.

netroms (approved):

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
```diff
@@ -251,10 +276,45 @@ private <T> void toObjectNodes(
...
+   * <p>resulting output: <br>
+   * <code>
+   *   {
+   *      "name": "dhis2"
+   *   }
+   * </code>
+   *
+   * @param objectNode object node to process on
+   */
+  private void removeEmptyObjects(ObjectNode objectNode) {
```
vietnguyen, line 307: For this, maybe we can turn on [JsonInclude.Include.NON_EMPTY](https://github.com/dhis2/dhis2-core/blob/3f202cc74b9ceb72b17beb257fc7fb3d34bd55fa/dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/config/JacksonObjectMapperConfig.java#L165). 
Can discuss later.

vietnguyen (approved):

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java
david-mackessy: I actually tried using that in several places but it didn't work for some reason.
Used it on the mapper `.setSerializationInclusion(JsonInclude.Include.NON_EMPTY)`
and also tried at the class & field level annotation. Strange.

PR #18717, opened 2024-10-02
fix: Prevent import of duplicate analytics tables hooks

jason-p-pickering:
Original issue is described in https://dhis2.atlassian.net/browse/DHIS2-18148. 

Basically two equivalent analytics tables hook (same phase, table type, and SQL expression) can be imported if they have different names. It seems to make no sense to execute the exact same SQL twice for the same phase and resource table type so we should prevent users from importing (perhaps mistakenly) the same analytics hook more than one time.


jbee (commented):
This looks good except this is not the correct way to do this validation. We cannot change an entity object's equals/hashCode contract to de-duplicate on another semantic of equality than the database one. 

That said you are very close to the code needed. Just scrap the hashCode change and move your equals implementation to the `AnalyticsTableHookObjectBundleHook` where it is a static method taking two objects like 
```java
static boolean equal(AnalyticsTableHook one, AnalyticsTableHook other) { ... }
```

jason-p-pickering:
Thank you @jbee for the feedback and guidance. I have reworked the code based on your comments. I have also incorporated feedback from @david-mackessy regarding the tests. Everything seems green now, so please have another look when you get a chance.

david-mackessy (approved):

maikelarabori (approved):

jbee (approved):

PR #18718, opened 2024-10-02
fix: remove username validation to allow old usernames on login

netroms:
## Summary
Remove username validation on login to allow for pre 2.38 style username. Stricter username constraint was introduced in 2.38 and was only enforced on user creation and update. In 2.41 we introduced a new login page that utilize a new back-end controller and logic to validate the username before it is looked up in the database. This fix just removes this check as it is not really necessary.

### Automatic test
AuthenticationControllerTest#testSuccessfulLoginWithOldUsername()

### Manual test
1. Start a new 2.37 instance.
2. Create a user with a username that is not allowed in 2.38+, for example "Üsername".
3. Upgrade the instance to 2.41.
4. Try to log in with the "Üsername" username.
5. Observe you get logged in.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/security/AuthenticationControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/security/AuthenticationControllerTest.java
@@ -58,6 +58,21 @@ class AuthenticationControllerTest extends AuthenticationApiTestBase {
   @Autowired private SessionRegistry sessionRegistry;
 
   @Test
+  void testSuccessfulLoginWithOldUsername() {
+    User adminUser = userService.getUserByUsername("admin");
+    adminUser.setUsername("Üsername");
+    userService.updateUser(adminUser);
+
+    JsonLoginResponse response =
+        POST("/auth/login", "{'username':'Üsername','password':'district'}")
+            .content(HttpStatus.OK)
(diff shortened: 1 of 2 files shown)


vietnguyen (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/security/AuthenticationControllerTest.java
```diff
@@ -57,6 +57,21 @@ class AuthenticationControllerTest extends AuthenticationApiTestBase {
   @Autowired SystemSettingManager systemSettingManager;
   @Autowired private SessionRegistry sessionRegistry;
 
+  @Test
+  void testSuccessfulLoginWithOldUsername() {
+    User adminUser = userService.getUserByUsername("admin");
+    adminUser.setUsername("Üsername");
```
david-mackessy, line 63: does the system allow you to set a username to an invalid username? is this a separate bug?

david-mackessy (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/security/AuthenticationControllerTest.java
netroms: This is only possible manually like I do, not trough the API.

PR #18719, opened 2024-10-02
fix: use tracker importer in ProgramStageDataEntrySMSListener DHIS2-17729

teleivo:
As a reference [this is](https://github.com/dhis2/dhis2-core/blob/998f7cc3c5a1271fd7742adf1253521f8e66e234/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/listener/ProgramStageDataEntrySMSListener.java#L116) the `ProgramStageDataEntrySMSListener` processing logic before we made our changes to sms processing. You can see that users were able to enroll and create an event in a tracker program.

This is how an sms for an sms command can look like

```
{
  "text": "visit s=yes,c=no,h=14",
  "originator": "+233223232"
}
```

The exact details of the format depend on how you configure the command you create
* https://docs.dhis2.org/en/use/user-guides/dhis-core-version-241/maintaining-the-system/mobile.html?h=sms+2.41#sms-command-for-program-stage-data-entry
* https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/sms.html#webapi_sms_commands

**Important**: this type of SMS is peculiar. For all SMS the SMS originator is used to find the user making whatever action is associated with the type of SMS. We try to find the TE by the phone number attribute value = originator. The SMS processing job is run as the user that sent the SMS. So with this SMS the authenticated user is also the TE we are looking for. An error is returned if more than 1 TE has the same number.

Change summary

* Use the tracker importer in `ProgramStageDataEntrySMSListener`.
* Replace `parallelStream` to search TE by phone number using `stream`. The former would require us to set the `currentUser` on each thread and clean up afterwards. Not sure who uses this feature so make the easy change for now. We should rather make one query to look for any phone number attribute at once.
* Make the `TrackerEventSMSTest` non-transactional as finding the TE is done in multiple threads. `@Transactional` tests means our test data is not visible.
* Use paged versions of our APIs as we need to fetch 1+ entities to find and validate that there is at most one entity.
* `QueryOperator.LIKE` was used to find the TE via the originator phone number. Use `EQ` instead as you do not want to enroll someone else!
* Remove unused code.

## Next

These are the tracker related compression based SMS we changed so they use the importer:

- `ENROLLMENT`        => `EnrollmentSMSListener` ✅ (https://github.com/dhis2/dhis2-core/pull/18595)
- `SIMPLE_EVENT`      => `SimpleEventSMSListener` ✅ (https://github.com/dhis2/dhis2-core/pull/18683)

- `PROGRAM_STAGE_DATAENTRY_PARSER` => `ProgramStageDataEntrySMSListener` (this PR)
- `EVENT_REGISTRATION_PARSER` => `SingleEventListener` (https://github.com/dhis2/dhis2-core/pull/18708)


enricocolasante (approved):

zubaira (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/ProgramStageDataEntrySMSListener.java
```diff
@@ -187,20 +231,16 @@ private boolean hasMoreThanOneEntity(List<TrackedEntity> trackedEntities) {
...
-    queryFilter.setOperator(QueryOperator.LIKE);
+    queryFilter.setOperator(QueryOperator.EQ);
     queryFilter.setFilter(sms.getOriginator());
 
     return TrackedEntityOperationParams.builder()
         .filters(Map.of(attribute.getUid(), List.of(queryFilter)))
-        .programUid(program.getUid())
-        .organisationUnits(
-            ous.stream().map(BaseIdentifiableObject::getUid).collect(Collectors.toSet()))
+        .trackedEntityTypeUid(program.getTrackedEntityType().getUid())
```
ameenhere, line 242: If the phone number we use to search for a TE, can it be either a program TEAttribute or a TETypeAttribute?  In cases where the phone number is a program tea, does this filter on TET here block us from finding the TE with the phone number?  Probably not. Just curious.

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/sms/ProgramStageDataEntrySMSListener.java
teleivo: This SMS can also enroll a TE. We can not find the TE if we query using the program 😅

PR #18756, opened 2024-10-04
fix: use tomcat 10 for dev

vietnguyen:


david-mackessy (approved):

tonsV2 (approved):

radnov:
We need to update the `tomcat` base image version in the stable pipeline, too:
https://github.com/dhis2/dhis2-core/blob/2389bebcc0b151253d85529040f866865f9cfaf7/jenkinsfiles/stable#L38
https://github.com/dhis2/dhis2-core/blob/2389bebcc0b151253d85529040f866865f9cfaf7/jenkinsfiles/stable#L73

PR #18761, opened 2024-10-04
chore: cleanup and simplify web test API

jbee:
### Summary
I was looking into ways to enable us to use the same API we use in integration tests in the e2e. I concluded that the easiest to do now and also to maintain in this case would be making a copy as e2e and core are otherwise independent modules. 
Therefore, before I made a copy I thought I try to simplify and cleanup the API so we have less code and complexity that is copied. This PR is that first step of simplifying the API.

* removed some unused methods
* use of `record` classes for data
* removed the `WebSnippet` concept (was only unused in the initial examples I made)
* removed the feature to state the content type as prefix of the value, e.g. `plain/text:hello` now must be `ContentType("plain/text"), Body("hello")` (was too special and unclear and collided with other features)
* made POST/PUT/PATCH/DELETE with JSON body specific methods to allow `json5` language annotation
* added overload for POST/PUT/PATCH with `Path` in case a file becomes the body
* some renames (want to do more but those will affect many files so I do them in another PR)


david-mackessy (approved):
One of the things I like about the e2e tests is the use of RestAssured for response body assertions. Are you thinking of removing it?

netroms (approved):

jbee:
> One of the things I like about the e2e tests is the use of RestAssured for response body assertions. Are you thinking of removing it?

Not removing but discouraging it, yes. It is one of the reasons why e2e is hard to maintain as finding errors with that type of assertions is very time-consuming compared to basic JUnit asserts we use in integration. I understand that these are powerful and quick to write but that isn't a good point of optimization. We need something that is easy to debug/understand and thus easy to maintain. Apart from that RestAssured also has some performance issues that become relevant i a few endpoints. As we have many existing tests I don't see that we have resources to change them but I at least want to enable us to use the API we use in integration in future tests.

david-mackessy:
cool, look forward to trying it out 👍

PR #18776, opened 2024-10-07
fix: No data returned for data element details/operands as filter [2.42-DHIS2-18182]

d-bernat:
The Analytics API is not returning data for DataElementOperand dimensions passed in the filter parameter. The missing functionality needs to be added to the codebase.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/handler/DataHandler.java
```diff
@@ -845,6 +820,140 @@ private void addDataElementOperandValues(DataQueryParams params, Grid grid, Tota
...
+  /**
+   * Delivers a DataQueryParams instance for DataElementOperands. The dimension and filter operands
+   * are added to the query parameters.
+   *
+   * @param params the {@link DataQueryParams}.
+   * @param operands the collection of {@link DataElementOperand}.
+   * @param totalType the {@link TotalType}.
+   * @return mapped DataQueryParams
+   */
+  private DataQueryParams getOperandDataQueryParams(
```
gnespolino, line 832: I agree, all repeated code should be isolated in separate method

gnespolino (approved):

maikelarabori (approved):

PR #18788, opened 2024-10-09
fix: Fix flaky unit tests missing authenticated user

enricocolasante:
Failing tests were missing the authenticated user and `CurrentUserUtil.getCurrentUserDetails` was throwing an exception.
They are flaky tests because depending on the order of tests, some setup was setting the user and not cleaning up the context.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackerImporterServiceTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackerImporterServiceTest.java
@@ -40,6 +40,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.hisp.dhis.scheduling.JobProgress;
 import org.hisp.dhis.scheduling.RecordingJobProgress;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackerImporterServiceTest.java
```diff
@@ -85,6 +86,8 @@ public void setUp() {
         new DefaultTrackerImportService(
             trackerBundleService, validationService, trackerPreprocessService);
 
+    injectSecurityContext(user);
```
jbee, line 89: I think we added a static version specifically so mock based tests do not have to extend `TestBase` for this

jbee (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackerImporterServiceTest.java
enricocolasante: Done

teleivo (approved):

PR #18790, opened 2024-10-09
feat: add CSRF token feature

netroms:
## Summary
Introduces config option to enable CSRF tokens in Spring security. Historically this feature has been disabled due to the high cost of implementing it, caused by the diverse ecosystem of DHIS2 apps. Since 2.42 all apps are SPA apps and all bundled apps are using a common library for requests, this feature now seems more realistic to be implemented.
This PR serves as a development preview of the feature, and is disabled by default. The idea is that it can be enabled when app developers start integrating/testing it into their apps.

The PR implements the reference code for the SPA version of CSRF tokens, explained here: [https://docs.spring.io/spring-security/reference/servlet/exploits/csrf.html#csrf-integration-javascript-spa](https://docs.spring.io/spring-security/reference/servlet/exploits/csrf.html#csrf-integration-javascript-spa)

The server will set a HttpOnly=false cookie called XSRF-TOKEN, which the client needs to extract and put into a header called `X-XSRF-TOKEN` for all POST, PUT, PATCH and DELETE requests.

### Automatic tests
N/A yet.

### Manual testing
1. Enable the feature in the dhis.conf with: `http.security.csrf.enabled = true`
2. Install ModHeader Chrome extension or similar to "fake" the CSRF token header.
3. Do any requests to the server and fetch the XSRF-TOKEN cookie value that should be set.
4. Add a header called: `X-XSRF-TOKEN`, in the ModHeader extension, use the value fetched in step 3.
5. Observe you don't get a 503 forbidden when trying to do a POST, PUT, PATCH and DELETE.


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -667,7 +667,10 @@ public enum ConfigurationKey {
   SWITCH_USER_ALLOW_LISTED_IPS(
       "switch_user_allow_listed_ips", "localhost,127.0.0.1,[0:0:0:0:0:0:0:1]", false),
 
-  MAX_FILE_UPLOAD_SIZE_BYTES("max.file_upload_size", Integer.toString(10_000_000), false);
+  MAX_FILE_UPLOAD_SIZE_BYTES("max.file_upload_size", Integer.toString(10_000_000), false),
+
+  /** CSRF feature. Enable or disable the feature. */
+  CSRF_ENABLED("csrf.enabled", Constants.OFF, true);
```
jbee, line 673: Should we use a more namespaced name like `http.security.csrf.enabled` ? or something like it - maybe the other configurations we have in that area already establish a common prefix?

jbee (approved):

david-mackessy (approved):

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
netroms: I like that, will use that. Better late than never, we haven't really done that before, but why not start.

PR #18802, opened 2024-10-10
fix: Catch JCloud exceptions again [DHIS2-18212]

jimgrace:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/jclouds/JCloudsStore.java
```diff
@@ -149,13 +151,32 @@ private Properties configureOverrides(String provider, String endpoint) {
...
   public void init() {
     Location location = createLocation(fileStoreConfig.provider, fileStoreConfig.location);
-    blobStoreContext.getBlobStore().createContainerInLocation(location, fileStoreConfig.container);
 
-    log.info(
-        "File store configured with provider: '{}', container: '{}' and location: '{}'.",
-        fileStoreConfig.provider,
-        fileStoreConfig.container,
-        fileStoreConfig.location);
+    try {
```
david-mackessy, line 155: I think a comment should be added above the try/catch explaining the crucial part it plays for AWS usage.
Without it, something similar could happen in the future.
We have tests for filesystem & MinIO usage but none for AWS, so a detrimental change to the AWS flow is not easily evident to us.

PR #18807, opened 2024-10-11
fix: Move converters into the bundle package [DHIS2-18093]

enricocolasante:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -0,0 +1,348 @@
...
+    dbEnrollment.setLastUpdated(now);
+    dbEnrollment.setLastUpdatedByUserInfo(UserInfoSnapshot.from(user));
+    dbEnrollment.setCreatedAtClient(DateUtils.fromInstant(enrollment.getCreatedAtClient()));
+    dbEnrollment.setLastUpdatedAtClient(DateUtils.fromInstant(enrollment.getUpdatedAtClient()));
+
+    Date enrollmentDate = DateUtils.fromInstant(enrollment.getEnrolledAt());
+    Date occurredDate = DateUtils.fromInstant(enrollment.getOccurredAt());
+
+    dbEnrollment.setEnrollmentDate(enrollmentDate);
+    dbEnrollment.setOccurredDate(occurredDate != null ? occurredDate : enrollmentDate);
```
muilpp, line 141: what's the difference between this one and `enrollmentDate`? and why do we set it to `enrollmentDate` if `occurredDate` is null?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/RelationshipPersister.java
```diff
@@ -47,23 +48,22 @@
...
     super(reservedValueService, trackedEntityChangeLogService);
-    this.relationshipConverter = relationshipConverter;
   }
 
   @Override
   protected org.hisp.dhis.relationship.Relationship convert(
       TrackerBundle bundle, Relationship trackerDto) {
-    return relationshipConverter.from(bundle.getPreheat(), trackerDto, bundle.getUser());
+    if (bundle.getStrategy(trackerDto) == TrackerImportStrategy.UPDATE) {
+      return null;
```
muilpp, line 63: why do we not map the object in case of UPDATE?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -0,0 +1,348 @@
...
+      switch (dbEnrollment.getStatus()) {
+        case ACTIVE -> {
+          dbEnrollment.setCompletedDate(null);
+          dbEnrollment.setCompletedBy(null);
+        }
+        case COMPLETED -> {
+          dbEnrollment.setCompletedDate(now);
+          dbEnrollment.setCompletedBy(user.getUsername());
+        }
+        case CANCELLED -> {
```
muilpp, line 161: unrelated to this PR, but I've seen that when the enrollment is `CANCELLED`, the related event(s) are not modified, so it could happen that an event is `ACTIVE` while the parent enrollment is `CANCELLED`, does that make sense?

muilpp (approved):

teleivo:
> * `EnrollmentStatus` default is `ACTIVE`. Added default value in the view and the model and removed it from the mapper.

Why is a default needed in both places?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
enricocolasante: Now the there is less code and it is easier to understand this logic I created https://dhis2.atlassian.net/browse/DHIS2-18224 to reason about this kind of stuff.
I am not sure what is the logic here

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
enricocolasante: Same as before, I just didn't change the logic, but in https://dhis2.atlassian.net/browse/DHIS2-18224 we should try to understand if it makes sense or not.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/RelationshipPersister.java
enricocolasante: Relationships cannot be updated.
If this is the case, we are ignoring the object in the payload and send a warning.
Ivo created a bug https://dhis2.atlassian.net/browse/DHIS2-18110 where we should improve the way we set the ignored objects to always provide an error/warning. Here it is happening but it is not enforced by the code.

enricocolasante:
> > * `EnrollmentStatus` default is `ACTIVE`. Added default value in the view and the model and removed it from the mapper.
> 
> Why is a default needed in both places?

Yeah, I don't think it is.
I started adding the default in the view and then I realized that I needed in the service, because tracker importer endpoint is not the only client, but then I didn't remove it from the view.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -0,0 +1,348 @@
...
+          preheat.getCategoryOptionCombo(event.getAttributeOptionCombo()));
+    } else {
+      dbEvent.setAttributeOptionCombo(preheat.getDefault(CategoryOptionCombo.class));
+    }
+
+    dbEvent.setGeometry(event.getGeometry());
+
+    EventStatus currentStatus = event.getStatus();
+    EventStatus previousStatus = dbEvent.getStatus();
+
```
teleivo, line 218: like above if you make a paragraph (no empty lines) in between all the status logic its easier to read this logic

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -0,0 +1,348 @@
...
+    return dbEvent;
+  }
+
+  public static @Nonnull Relationship map(
+      @Nonnull TrackerPreheat preheat,
+      @Nonnull org.hisp.dhis.tracker.imports.domain.Relationship relationship,
+      @Nonnull UserDetails user) {
+    RelationshipType relationshipType =
+        preheat.getRelationshipType(relationship.getRelationshipType());
+    RelationshipItem fromItem = new org.hisp.dhis.relationship.RelationshipItem();
```
teleivo, line 279: declare vars as close to their usage. easier to read and added benefit of sometimes narrowing their scope

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -0,0 +1,348 @@
...
+
+    return dbRelationship;
+  }
+
+  private static @Nonnull Note map(
+      @Nonnull TrackerPreheat preheat,
+      @Nonnull org.hisp.dhis.tracker.imports.domain.Note note,
+      @Nonnull UserDetails user) {
+    Note dbNote = new Note();
+    dbNote.setAutoFields();
```
teleivo, line 332: why are we using `setAutoFields` on notes and not other entities?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
```diff
@@ -127,23 +128,24 @@ public TrackerTypeReport persist(EntityManager entityManager, TrackerBundle bund
         //
         // Save or update the entity
         //
-        if (isNew(bundle.getPreheat(), trackerDto)) {
+        if (bundle.getStrategy(trackerDto) == TrackerImportStrategy.CREATE) {
```
teleivo, line 131: why is https://github.com/dhis2/dhis2-core/blob/112bd6fb4102b07d3a737528e48711503b3d10c7/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerImportStrategy.java#L44 not considered here? 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
```diff
@@ -127,23 +128,24 @@ public TrackerTypeReport persist(EntityManager entityManager, TrackerBundle bund
...
           entityManager.persist(convertedDto);
           typeReport.getStats().incCreated();
           typeReport.addEntity(objectReport);
           updateAttributes(
               entityManager, bundle.getPreheat(), trackerDto, convertedDto, bundle.getUser());
         } else {
-          if (isUpdatable()) {
+          if (trackerDto.getTrackerType() == TrackerType.RELATIONSHIP) {
```
teleivo, line 138: how does this influence the bug we've been seeing with someone trying to import events and only getting an ignored count with no validation errors in the report?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EnrollmentPersister.java
```diff
@@ -163,16 +152,15 @@ protected TrackerType getType() {
 
   @Override
   protected void persistOwnership(TrackerPreheat preheat, Enrollment entity) {
-    if (isNew(preheat, entity.getUid())) {
```
teleivo, line 166: Why is the `isNew` deleted? Was this wrong?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
enricocolasante: At this point in the importer `CREATE_AND_UPDATE` is not relevant anymore.
In [StrategyPreProcessor](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preprocess/StrategyPreProcessor.java) we are assigning the correct strategy to every single entity in the bundle, so here it can only be `CREATE`, `UPDATE` or `DELETE`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java
enricocolasante: Here is a case where we are increasing the ignored stats. In my mind if we are fixing the bug making the stats an invariant that will force us to declare an error/warning at the moment of increasing the ignored stat, it will improve the code here as well.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EnrollmentPersister.java
enricocolasante: It was not wrong but the logic to make it work with relationships was really complicated.
I put it back but changed the implementation to use the strategy in `TrackerBundle` instead of checking the `TrackerPreheat`

enricocolasante:
> Is the `TrackerObjectsMapper` only used by the persisters? if so move it their and make package private

Unfortunately it is used in here https://github.com/dhis2/dhis2-core/blob/7f59931e9b44d2471e6a4651d5ff796d1896e0e0/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/relationship/SecurityOwnershipValidator.java#L63.

I would like to remove it at some point but it is out of the scope of this PR.

enricocolasante:
> > > * `EnrollmentStatus` default is `ACTIVE`. Added default value in the view and the model and removed it from the mapper.
> > 
> > 
> > Why is a default needed in both places?
> 
> Yeah, I don't think it is. I started adding the default in the view and then I realized that I needed in the service, because tracker importer endpoint is not the only client, but then I didn't remove it from the view.

It is actually needed in both places.
If I don't add it in the view, than the `null` will be propagated.
If I don't set it in the service, than any other client will not have a default value.

teleivo (approved):

PR #18834, opened 2024-10-16
ci: migrate to docker compose v2 CLI in stable and dev pipelines [skip ci]

radnov:


teleivo (approved):
can you also update https://github.com/dhis2/dhis2-core/blob/be67efbf523f453d89275779771c5acbd14020f8/.github/workflows/run-api-analytics-tests.yml#L75

`docker compose` is available as its used before that line in https://github.com/dhis2/dhis2-core/blob/be67efbf523f453d89275779771c5acbd14020f8/.github/workflows/run-api-analytics-tests.yml#L68

david-mackessy (approved):

PR #18843, opened 2024-10-16
feat: Add embedded properties to Dashboard [DHIS2-18239]

larshelge:
This PR adds a object `EmbeddedDashboard` properties to the `Dashboard` entity meant to represent an embedded dashboard.

* `provider` refers to the provider of embedded dashboards, and can be extended to Power BI, Tableau and the like later.
* `id`  refers to the identifier of the embedded dashboard defined by the external provider.
* `options` refers to customization options from the embedded dashboard provider, and can be extended later.

Adds a system setting for enabling Superset embedded dashboards.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dashboard/Dashboard.java
```diff
@@ -67,11 +67,35 @@ public class Dashboard extends BaseNameableObject implements MetadataObject {
   /** Allowed filter dimensions (if any) which may be used for the dashboard. */
   private List<String> allowedFilters = new ArrayList<>();
 
+  /**
+   * Provider of embedded dashboards, is never <code>null</code>. The default value is {@link
+   * EmbeddedProvider#NONE} which refers to standard DHIS 2 dashboards. Other values indicate that
+   * this dashboard is sourced from an external provider and embedded in DHIS 2.
+   */
+  private EmbeddedProvider embeddedProvider;
```
maikelarabori, line 75: We could have an embedded class, like `Embedded`, that encapsulated the following:

```
private String id;
private Provider provider;
private Options options;
```

or something similar, if possible. So, we can groups all embedded thing into a parent single object.

maikelarabori: Also, could we have many embedded providers? If so, we could have a Set<Embedded> of embedded(s).

larshelge: Yes good point, was thinking of how to use an embedded class (meaning @Embedded Hibernate annotation/type), I am just unsure how that will work with our metadata importer, as I don't think we use embedded classes anywhere else.

We could use a class and persist as JSONB. That way we can encapsulate it, though that makes it harder to upgrade with Flyway later.

larshelge: There can only be one embedded "item". If you want many embedded dashboards, you create more DHIS 2 dashboards.

larshelge: I have modified it to use an `EmbeddedDashboard` class with JSONB storage now @maikelarabori .

gnespolino (approved):

maikelarabori (approved):

larshelge:
I moved this PR to a new one due to the signed commit requirement in master.

https://github.com/dhis2/dhis2-core/pull/19016

larshelge:
Merged the new PR now. Closing this PR.

https://github.com/dhis2/dhis2-core/pull/19016

PR #18857, opened 2024-10-17
chore: Clean up tracker domain objects [DHIS2-18222]

enricocolasante:
This PR is removing all fields that are not used in tracker importer. Most of them they were there because our view model, that in the web layer is mapped to tracker model, is used for input and output and contains read-only fields that are not necessary in tracker importer.

In particular:
- Tracker model is flat (tracker objects are not nested, it is flattened in the controller [BodyConverter](https://github.com/dhis2/dhis2-core/blob/40689f3ad9d62a5c8e83091182793993913c9702/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/BodyConverter.java#L42) ), so all the nesting entities can be safely removed.
- All system-created fields are not needed in tracker model (createdAt, createdBy, etc...)
- `deleted` flag cannot be modified and importer is only dealing with not deleted objects.

***Next steps***
`DataValue` fields are going to be removed in https://dhis2.atlassian.net/browse/DHIS2-18223


teleivo:
> Most of them they were there because our view model, that in the web layer is mapped to tracker model, is used for input and output and contains read-only fields that are not necessary in tracker importer.

maybe to clarify: our view models are fairly new and are used as the types for import/export. The view models originated from the domain models. We likely forgot to delete those domain model fields when we created the view models.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Event.java
```diff
@@ -96,10 +86,6 @@ public class Event implements TrackerDto, Serializable {
 
   @JsonProperty private User assignedUser;
```
teleivo, line 87: Can users set `completedBy/At`?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/JobConfigurationRunErrorsControllerTest.java
```diff
@@ -174,7 +174,7 @@ void testGetJobRunErrors_ListIncludeInput() throws InterruptedException {
     // language=JSON
     String expected =
         """
-      {"trackedEntities":[{"trackedEntity":"sHH8mh1Fn0z","trackedEntityType":{"idScheme":"UID","identifier":"nEenWmSyUEp"},"orgUnit":{"idScheme":"UID","identifier":"DiszpKrYNg7"},"inactive":false,"deleted":false,"potentialDuplicate":false,"relationships":[],"attributes":[],"enrollments":[]}],"enrollments":[],"events":[],"relationships":[]}""";
```
teleivo, line 177: Does this change affect jobs? If so should we add a migration to remove unnecessary data? Or would such jobs be removed after a while anyway

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Event.java
enricocolasante: I asked this myself and then forgot to go forward with that.
The `completedBy` in my head shouldn't be set by the user, you shouldn't be able to do stuff on behalf of others.
The `createdAt` could be something that the user needs to set. I asked @vgarciabnz how it is expected to work when Android is synching.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/JobConfigurationRunErrorsControllerTest.java
enricocolasante: I never know :grimacing: 
@jbee Can you help us here?

muilpp (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/JobConfigurationRunErrorsControllerTest.java
jbee: Tracker import jobs are removed a day after they have been running. I am not quite sure what changed with regards to the job but I think you removed something which should be fine. If the job was stored with more parameter data than it later executes this still works.

enricocolasante: Thanks @jbee .

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Event.java
enricocolasante: I had a chat with Victor another with Ameen and Joakim. `completedAt` could be set by the user so I will put it back for `Enrollment` and leave it as it is for `Event`.
I am going to do it in next PR though

PR #18871, opened 2024-10-18
chore: Clean up data values persistence [DHIS2-18223]

enricocolasante:
The flow of persisting data values was quite complicated and error prone.
Now we are dealing with data values in `Event` in the same way we are dealing with attributes in `Enrollment` and `TrackedEntity`. We expect only the changed values (`null` value represent a `DELETE`).

- `dataValues` field is not mapped in `TrackerObjectsMapper`
- In `EventPersister`, for every `dataValue` in the payload we decide if it is a new value, if it is an update or a deletion and we deal with `FileResource`s and `ChangeLog` accordingly.
- `createdAt`, `updatedAt`, `createdBy` and `updatedBy` fields follow the same logic as implemented for all tracker objects.
- It was possible for the client to set `createdAt` for a `dataValue`. This is a bug. https://dhis2.atlassian.net/browse/DHIS2-18252. A failing e2e test, solved by fixing this [file](dhis-2/dhis-test-e2e/src/test/resources/tracker/importer/events/events.json), was showing this issue.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,94 @@ private void handleDataValues(
...
           DataElement dataElement = preheat.getDataElement(dv.getDataElement());
-          checkNotNull(
-              dataElement,
-              "Data element should never be NULL here if validation is enforced before commit.");
 
           // EventDataValue.dataElement contains a UID
           EventDataValue eventDataValue = dataValueDBMap.get(dataElement.getUid());
 
-          ValuesHolder valuesHolder = getAuditAndDateParameters(eventDataValue, dv);
+          if (isNewDataValue(eventDataValue, dv)) {
```
muilpp, line 176: LGTM, the only thing is that this mapper method got very long, I'd split it in different methods, maybe each change log type condition (CREATE, UPDATE and DELETE) could be done different method

enricocolasante: Done

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,118 @@ private void handleDataValues(
     payloadDataValues.forEach(
         dv -> {
           DataElement dataElement = preheat.getDataElement(dv.getDataElement());
-          checkNotNull(
-              dataElement,
-              "Data element should never be NULL here if validation is enforced before commit.");
 
           // EventDataValue.dataElement contains a UID
```
teleivo, line 173: is this comment still needed?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,118 @@ private void handleDataValues(
...
           EventDataValue eventDataValue = dataValueDBMap.get(dataElement.getUid());
```
teleivo, line 174: In the TrackerObjectsMapper you prefix the var representing the value from the DB with `db`.

The `event` prefix isn't all that useful in the event persister. Its hard to understand what `eventDataValue` vs `dv` is.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,118 @@ private void handleDataValues(
...
-      eventChangeLogService.addTrackedEntityDataValueChangeLog(valueAudit);
-    }
+      String userName,
+      DataElement de,
+      Event event,
+      String value,
+      boolean providedElsewhere,
+      ChangeLogType changeLogType) {
+
+    TrackedEntityDataValueChangeLog valueAudit = new TrackedEntityDataValueChangeLog();
```
teleivo, line 274:
```suggestion
    TrackedEntityDataValueChangeLog changeLog = new TrackedEntityDataValueChangeLog();
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,118 @@ private void handleDataValues(
...
+      EntityManager entityManager,
+      TrackerPreheat preheat) {
+    if (dataElement.isFileType()) {
+      unassignFileResource(entityManager, preheat, event.getUid(), eventDataValue.getValue());
+    }
+
+    event.getEventDataValues().remove(eventDataValue);
   }
 
   private void logTrackedEntityDataValueHistory(
```
teleivo, line 266:
```suggestion
  private void logDataValueChange(
```

or `addTrackedEntityDataValueChangeLog` as the service name (although I am unsure thats a good name).

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -172,63 +169,118 @@ private void handleDataValues(
...
+      EventDataValue eventDataValue,
+      Event event,
+      DataElement dataElement,
+      EntityManager entityManager,
+      TrackerPreheat preheat) {
+    if (dataElement.isFileType()) {
+      unassignFileResource(entityManager, preheat, event.getUid(), eventDataValue.getValue());
+    }
+
+    event.getEventDataValues().remove(eventDataValue);
```
teleivo, line 263: Can we rely on the equals implementation? I was bitten a few times by our equals implementations. Just wanted to double-check 😅

teleivo:
> expect only the changed values (`null` value represent a `DELETE`)

and an empty string as well right? Here it looks like we are checking for blank, while for attributes we check for empty. Not that we have to support that but both data values and attributes should behave the same way.

https://github.com/dhis2/dhis2-core/blob/accad7214119afaf3de7d702821216a2ba8c82de/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java#L322

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
muilpp: This adds a change log record for an event. I would prefer not to include the word `TrackedEntity` in the method name, as I believe it makes it difficult to distinguish from the attributes change log.

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
teleivo: definitely 👍🏻 

I guess this service method is just an old one with old names. The type itself is also `TrackedEntityDataValueChangeLog`.

The service is `EventChangeLogService`, how about
* `addDataValueChangeLog()`
* `add` might be good enough as well as you have the context from the service/parameter type on what changes you are tracking

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
enricocolasante: In this case it is just checking the `dataElement` https://github.com/dhis2/dhis2-core/blob/accad7214119afaf3de7d702821216a2ba8c82de/dhis-2/dhis-api/src/main/java/org/hisp/dhis/eventdatavalue/EventDataValue.java#L120 so it works for us.

dhis-2/dhis-test-e2e/src/test/resources/tracker/importer/events/events.json
```diff
@@ -117,8 +117,6 @@
           "providedElsewhere": false
         },
         {
-          "updatedAt": "2019-03-04T15:12:59.209",
```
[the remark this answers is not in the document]
enricocolasante: I am not sure.
Maybe we can find a way to have a `jsonMapper` just for tests in the integration tests that is going to be strict and fail if there are unknown fields.
In e2e tests that wouldn't work because we are using the view model and those fields are actually used in the output.
Anyway, I cannot think of something easy to implement.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
muilpp: sounds good to me
there's [this ticket](https://dhis2.atlassian.net/browse/DHIS2-17835) that's about splitting change log from audit, we can tackle all these naming issues there

enricocolasante:
> > expect only the changed values (`null` value represent a `DELETE`)
> 
> and an empty string as well right? Here it looks like we are checking for blank, while for attributes we check for empty. Not that we have to support that but both data values and attributes should behave the same way.
> 
> https://github.com/dhis2/dhis2-core/blob/accad7214119afaf3de7d702821216a2ba8c82de/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/AbstractTrackerPersister.java#L322

I agree.
Maybe even at API level we should agree on one way, like we are always considering a blank string as an empty string.
We can ask in the backend meeting if anyone has any input on that.

PR #18875, opened 2024-10-18
fix: Safely serialize JSON map [DHIS2-18253] [2.40]

larshelge:
This PR makes the `JsonAttributeValueBinaryType` more reliable against bad data in the `attributevalues` JSONB column in metadata tables like `organisationunit`.

A null key in the JSONB column will lead to an exception from Jackson.

```
com.fasterxml.jackson.databind.JsonMappingException: Null key for a Map not allowed in JSON
```


netroms (approved):

jbee (approved):
FYI: This is no longer an issue in 2.42 after my latest attribute cleanup

PR #18881, opened 2024-10-21
chore: Validate completedAt for Event and Enrollment [DHIS2-18222]

enricocolasante:
This PR fixes the logic around completion of enrollments and events.

***Enrollments***
A saved enrollment follows these rules:
- When an enrollment has status `COMPLETED`, fields `completedAt` and `completedBy` must be populated
- When an enrollment has status  `CANCELLED`, fields `completedAt` must be populated and `completedBy` must be null
- When an enrollment has status  `ACTIVE`, fields `completedAt` and `completedBy` must be null

***Events***
A saved event follows these rules:
- When an event has status `COMPLETED`, fields `completedAt` and `completedBy` must be populated
- When an event has status that is not `COMPLETED`, fields `completedAt` and `completedBy` must be null

***Fields logic***

`completedBy` is always calculated by the system, if required by the status, and it will be set to current user.
`completedAt` is set by the client or calculated by the system if not present, only when the status required it. The use case for it is Android sync where an event/enrollment is completed on day X and the sync happens on day Y, we still want `completedAt` to be X.

`completedBy` and `completedAt` are set only if the status is changing. Otherwise, the value for those fields present in the DB will not be updated.

If the status do not require `completedAt` field to be set but the client set it, a validation error is returned.


teleivo:
> completedAt

Completed events/enrollments can transition to other states again right? Does this logic you describe then apply in the same way if an event was previously completed?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/DateValidatorTest.java
```diff
@@ -168,4 +174,39 @@ void testFailOnMissingOccurredAtDate() {
...
+    validator.validate(reporter, bundle, enrollment);
+
+    assertHasError(reporter, enrollment, E1052);
+  }
+
+  @ParameterizedTest
+  @EnumSource(
+      value = EnrollmentStatus.class,
+      names = {"COMPLETED", "CANCELLED"})
+  void shouldValidateWhenCompletedAtIsPresentAndStatusAcceptCompletedAt(EnrollmentStatus status) {
```
teleivo, line 198:
```suggestion
  void shouldPassWhenCompletedAtIsPresentAndStatusAcceptCompletedAt(EnrollmentStatus status) {
```

is opposed to the `shouldFail` above (not sure if other validator tests already have a naming scheme/word we can reuse).

enricocolasante:
> > completedAt
> 
> Completed events/enrollments can transition to other states again right? Does this logic you describe then apply in the same way if an event was previously completed?

Yes, I updated the description to state that the completed fields are updated only if there is a transition.
So, if a completed event is updating one data element and it is not setting the `completedAt` field, the old `completedAt` previously calculated by the system will still be in the DB and it will not be updated.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/DateValidator.java
```diff
@@ -55,13 +57,22 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
...
-      reporter.addError(enrollment, E1023, enrollment.getOccurredAt());
+      reporter.addError(enrollment, E1023);
+    }
+
+    validateCompletedDateIsSetOnlyForSupportedStatus(reporter, enrollment);
+  }
+
+  private void validateCompletedDateIsSetOnlyForSupportedStatus(
+      Reporter reporter, Enrollment enrollment) {
+    if (enrollment.getCompletedAt() != null && EnrollmentStatus.ACTIVE == enrollment.getStatus()) {
```
teleivo, line 68: Maybe unlikely but still what happens if we add another enrollment status? This would then allow setting the `completedAt`. Is that a good behavior? Or would this be caught by a test. You use a different strategy for events (check for allowed status/here check for disallowed status).

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidator.java
```diff
@@ -70,25 +70,28 @@ public void validate(Reporter reporter, TrackerBundle bundle, Event event) {
...
+  private void validateCompletedDateIsSetOnlyForSupportedStatus(Reporter reporter, Event event) {
+    if (event.getCompletedAt() != null && EventStatus.COMPLETED != event.getStatus()) {
+      reporter.addError(event, E1051, event, event.getStatus());
+    }
+  }
+
   private void validateExpiryDays(
       Reporter reporter, Event event, Program program, UserDetails user) {
-    if (user.isAuthorized(Authorities.F_EDIT_EXPIRED.name())) {
+    if (user.isAuthorized(Authorities.F_EDIT_EXPIRED.name()) || event.getCompletedAt() == null) {
```
teleivo, line 86:
```suggestion
    if (event.getCompletedAt() == null || user.isAuthorized(Authorities.F_EDIT_EXPIRED.name())) {
```

I assume `isAuthorized` needs to check some collection which can be avoided if the date is null.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/DateValidator.java
enricocolasante: You are right, I am changing the logic here

muilpp (approved):

PR #18899, opened 2024-10-23
ci: keep war/Docker image Tomcat version in sync

teleivo:
We are only building/testing against one version of Tomcat on master since https://github.com/dhis2/dhis2-core/pull/18787

* remove the Tomcat matrix remnants from the jenkins pipeline (used before https://github.com/dhis2/dhis2-core/pull/18787)
* undo move of jib/tomcat version related properties done in https://github.com/dhis2/dhis2-core/pull/17040 back to the dhis-web-server/pom.xml. Like any other var its best to keep its scope minimal. These properties are not used anywhere else. Having them in the root makes it complex to override and understand where the actual values are coming from in the build of the war/Docker image.
* remove unused plugin `cargo-maven2-plugin`
* keep version of our embedded tomcat and Docker base image in sync by reusing the `tomcat.version` property

Tests
* `./dhis-2/run-api.sh` - running embedded Tomcat
* `./dhis-2/build-dev.sh` - building and running Docker image locally
* deployed PR using IM

### DIFF
--- a/dhis-2/dhis-web-server/pom.xml
+++ b/dhis-2/dhis-web-server/pom.xml
@@ -15,7 +15,13 @@
   <properties>
     <rootDir>../</rootDir>
     <build.time>${maven.build.timestamp}</build.time>
-    <tomcat.version>10.1.28</tomcat.version>
+    <tomcat.version>10.1.30</tomcat.version>
+    <jib.version>3.4.3</jib.version>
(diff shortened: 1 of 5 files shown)


radnov (approved):
Nice improvement! Looks good to me. 👍

enricocolasante (approved):

PR #18902, opened 2024-10-23
chore: Use UID in event exporter [DHIS2-17790]

enricocolasante:

Use UID in EventService and propagate the change until the generic stores.
Use UID in EventOperationParams and propagate until the generic stores and until tracker stores when building the queries.
EventRequestParams are already using UIDs.
EventOperationParams are now using UIDs. `dataElementFilters` and `attributeFilters` are going to be changed in next PR.
EventQueryParams is using hibernate object to build the query. A few fields are using UID and they are passed to the store and the value is used to build the query.

There is a quick fix in `UID` object to make tests work with `SystemUser`. It will be fix soon in https://dhis2.atlassian.net/browse/DHIS2-18296

Next steps*

Harmonize UID in Filters
Harmonize UID in TrackedEntity
Harmonize UID in Relationship
Harmonize UID in OperationsParamsValidator
Harmonize UID in deduplication package

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
@@ -131,14 +131,12 @@ class EventOperationParamsMapperTest {
 
   @BeforeEach
   public void setUp() {
-    String orgUnitId = "orgUnitId";
-    OrganisationUnit orgUnit = createOrgUnit("orgUnit", orgUnitId);
+    OrganisationUnit orgUnit = createOrgUnit("orgUnit");
     orgUnit.setChildren(
-        Set.of(
-            createOrgUnit("captureScopeChild", "captureScopeChildUid"),
-            createOrgUnit("searchScopeChild", "searchScopeChildUid")));
+        Set.of(createOrgUnit("captureScopeChild"), createOrgUnit("searchScopeChild")));
 
     User testUser = new User();
+    testUser.setUid(CodeGenerator.generateUid());
     testUser.setUsername("test");
     testUser.setOrganisationUnits(Set.of(orgUnit));
     user = UserDetails.fromUser(testUser);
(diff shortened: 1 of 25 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
```diff
@@ -420,25 +418,27 @@ void shouldMapOrgUnitWhenProgramProvidedAndRequestedOrgUnitInSearchScope(
...
 
     when(organisationUnitService.getOrganisationUnit(searchScopeChildOrgUnit.getUid()))
         .thenReturn(searchScopeChildOrgUnit);
 
     EventOperationParams operationParams =
         eventBuilder
-            .programUid(program.getUid())
-            .orgUnitUid(searchScopeChildOrgUnit.getUid())
+            .program(UID.of(program))
+            .orgUnit(UID.of(searchScopeChildOrgUnit))
```
teleivo, line 441: Would it make the API easier to use if we allowed passing things implementing the UidObject interface and then create the UIO ourselves? or do we then risk of not getting validated UIDs as the UID type throws if its an invalid UID while the interface might not be able guarantee that given that its

```java
public interface UidObject {
  /**
   * @return external unique ID of the object as used in the RESTful API
   */
  @JsonProperty(value = "id")
  String getUid();
}
```

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventOperationParamsMapperTest.java
enricocolasante: It would make it easier for tests for sure but in the implementation, at least from the exporter endpoints, at this point we do not have objects but just the UIDs passed as parameters.
For some specific cases like, currentUser, that could be useful.
The interface does not guarantee that the UID is valid, but we are kind of living with that assumption anyway.
I am gonna check how it would look like in the way you described.
For example, for the SMS part, would be easier to create the params.

PR #18903, opened 2024-10-23
feat: scope based global navigation [DHIS2-18266]

jbee:
### Summary
Provides scopes as a universal concept for global navigation.

The PR mainly provides the machinery behind the menu shown below

DHIS2-API-2-42-10-23-2024_07_27_PM

### Scopes + Classifiers
The main new concept are scopes. Each controller class is associated with a key-value set of classifiers. 
The `@OpenAPI.Document.classifiers()` annotation is used to add more classifiers to a controller class.
Based on these users can then compose a scope by specifying the classifiers to include. 

In the URL this is now given as a single parameter `scope`. A scope value has a classifier name and value; for example, `scope=entity.DataElement`. The `.` is used as separator as most other symbols are URL escaped making it hard to read if added to the parameters using JS.

While the scope entirely depends on the classifiers some classifiers are pre-filled from other properties for convenience but they can be overridden by specifying them explicitly.  

### Usage (Developers)
To classify a controller it is annotated like this
```java
@OpenApi.Document(classifiers = {"team:platform", "purpose:operation"})
```
To add a new way of classifying controller the only work is to add a classifier value. For example, when adding `"utilization:frequent"` a new row would appear in the _Scope_ menu shown above that has this option and when selected the annotated controller would be included in the scope.

### Next
I left a **TODO** on purpose to rename a property as that will affect many controllers. That PR can also be used to add more `classifiers` values to most controller. In this PR I only used it on one to verify that it works. 

### Automatic Testing
Existing tests were adjusted to use the new `scope` URL parameter instead of the now obsolete `domain` and `path`.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/OpenApi.java
@@ -163,19 +163,6 @@ public @interface OpenApi {
     String value();
   }
 
-  @Inherited
-  @Retention(RetentionPolicy.RUNTIME)
-  @Target(ElementType.TYPE)
-  @interface Team {
-
-    /**
-     * Names are case-insensitive, e.g. "Tracker" is the same as "tracker"
-     *
-     * @return name of the team that supports the annotated element
-     */
-    String value();
-  }
-
   /**
    * Can be used to annotate endpoint methods to constraint which concrete {@link EntityType}s will
    * support the annotated endpoint method.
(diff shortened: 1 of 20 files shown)


david-mackessy (approved):
Are you going for `team` over `domain`?
Domain is more universal whereas team is more of an internal term used by us.

jbee:
@david-mackessy 
> Are you going for team over domain?

I was going for team. One important group of users of the OpenAPI documentation are developers in and around the project. I think for them it is useful to have a team selector.

PR #18906, opened 2024-10-24
chore: Make SystemUser return valid uid [DHIS2-18296]

enricocolasante:
This PR is making interface `UserDetails` extending the `UidObject` interface and making `SystemUser` return a valid uid string for method `getUid`.
I choose the string `XXXXXSystem` as uid for `SystemUser` but I am open to suggestions.
I just remove from `UID` object the quick fix I introduced in a previous [PR](https://github.com/dhis2/dhis2-core/pull/18902) to make tests pass.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
@@ -35,11 +35,13 @@ import java.util.stream.Stream;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import org.hisp.dhis.common.IdentifiableObject;
+import org.hisp.dhis.common.UidObject;
 import org.hisp.dhis.security.Authorities;
 import org.hisp.dhis.user.UserDetailsImpl.UserDetailsImplBuilder;
 import org.springframework.security.core.GrantedAuthority;
 
-public interface UserDetails extends org.springframework.security.core.userdetails.UserDetails {
+public interface UserDetails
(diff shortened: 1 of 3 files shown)


david-mackessy (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -195,6 +197,7 @@ static UserDetails createUserDetails(
 
   boolean isSuper();
 
+  @Override
   String getUid();
```
teleivo, line 201: We could only be certain that this is a valid UID if this had the type UID. Is that something we could do later on?

teleivo (approved):

david-mackessy:
Maybe check for usage of the string value `system-process` throughout the system also, in case any logic has been implemented, expecting that string value.

enricocolasante:
> It seems that there are usages of the string value `system-process` throughout the system. They will probably need updating also I presume? (I stumbled across 1 such check by chance just now)

I think all of those usages are for the username, that is `system-process` as well. I checked and I couldn't find any usage for the UID.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
enricocolasante: I agree.
I am creating the UIDType for Hibernate in another PR, maybe one day we will be able to change the `BaseIdentifiableObject.uid` type to UID.

PR #18911, opened 2024-10-25
fix: (2.40) Incorrect follow-up value while saving program working list[DHIS2-18279]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-18279

This PR refactors the `ProgramStageQueryCriteria` class to change the `followUp` field from a `boolean` primitive to the `Boolean` wrapper type. This update allows `followUp` to hold three possible states: true, false, and null.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/JsonProgramStageQueryCriteria.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/JsonProgramStageQueryCriteria.java
@@ -49,7 +49,7 @@ public interface JsonProgramStageQueryCriteria extends JsonObject {
     return get("eventOccurredAt").as(JsonDatePeriod.class);
   }
 
-  default boolean getFollowUp() {
+  default Boolean getFollowUp() {
     return getBoolean("followUp").booleanValue();
   }
 }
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/programstageworkinglist/JsonProgramStageQueryCriteria.java
```diff
@@ -49,7 +49,7 @@ default JsonDatePeriod getEventOccurredAt() {
     return get("eventOccurredAt").as(JsonDatePeriod.class);
   }
 
-  default boolean getFollowUp() {
+  default Boolean getFollowUp() {
     return getBoolean("followUp").booleanValue();
```
ameenhere, line 53: This booleanValue() function returns a primitive boolean though. Not sure how this default function gets consumed in the end though.

zubaira: Firstly, this will only be used in tests. Secondly, there are three possible cases:

`followUp = null` (in this case, it will be ignored by JSON).
`followUp` is not provided (it will also be ignored by JSON).
`followUp = true/false` (the provided value will be used as-is).

ameenhere (approved):

PR #18912, opened 2024-10-25
chore: Use UID in relationship exporter [DHIS2-17790]

enricocolasante:

Use UID in RelationshipService and propagate the change until the generic stores.

Next steps*

Harmonize UID in TrackedEntity
Harmonize UID in OperationsParamsValidator
Harmonize UID in deduplication package


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -76,12 +77,12 @@ public Page<Relationship> getRelationships(
   }
 
   @Override
-  public Relationship getRelationship(@Nonnull String uid)
+  public Relationship getRelationship(@Nonnull UID relationshipUid)
```
teleivo, line 80:
```suggestion
  public Relationship getRelationship(@Nonnull UID uid)
```

teleivo (approved):

PR #18926, opened 2024-10-28
test: get one event via path and query param is identical DHIS2-18304

teleivo:
* isolate `EventsExportChangeLogsControllerTest` into own test class as its needs full control over changes
* make [EventsExportControllerByIdTest](https://github.com/dhis2/dhis2-core/blob/cd8e90ed664e51f6fb975b5c7afe896044056b23/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerByIdTest.java#L96) a postgres controller test. Before this change we could test the `/events/{uid}` code using H2 and Postgres. `/events?events={uid}` can only be tested using Postgres. Now that we use the same code we just need one type of test.

- [ ] can I use test lifecycle class?
- [ ] can we move `EventsExportControllerH2Test` into `EventsExportControllerTest` as well? Overriding the context is costly as Spring cannot cache the application context. Can we merge this test. Especially if we made its test lifecycle class that should be cheap.

## Behavior Changes

The following lists differences in responses from

* `/events/{uid}` - later referred to as `path`
* `/events?events={uid}` - later referred to as `query`

1. this is actually a bug

```sh
 diff path query                                                                                                                                                                              *+ am/rebase main
92c92
<       "storedAt": "2024-10-28T08:03:40.130",
---
>       "storedAt": "2024-10-28T00:00:00.000",
```

https://github.com/dhis2/dhis2-core/blob/ae5feb05372127b74b0538e5ea4052f9cc55a4c0/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java#L398

should already have used `resultSet.getTimestamp`.

2. there is a difference in rounding

```sh
diff --context=4 path/A01VPwDohRp.json query/A01VPwDohRp.json                                                                                                                                *+ am/rebase main 
*** path/A01VPwDohRp.json       2024-10-25 15:33:51.207614780 +0200
--- query/A01VPwDohRp.json      2024-10-25 15:33:51.299614704 +0200
***************
*** 18,26 ****
    "geometry": {
      "type": "Point",
      "coordinates": [
        -12.088856936816127,
!       7.6601569266967475
      ]
    },
    "dataValues": [
      {
--- 18,26 ----
    "geometry": {
      "type": "Point",
      "coordinates": [
        -12.088856936816127,
!       7.660156926696748
      ]
    },
    "dataValues": [
      {
```


teleivo:
Closing this in favor of https://github.com/dhis2/dhis2-core/pull/18943. Unfortunately, making signed commits mandatory was done without still allowing force pushes to our own branches. Re-creating a new PR is easiest for me until we fix our GH settings.

PR #18928, opened 2024-10-28
feat: add a brute force protection on the 2fa/disable endpoint (#18155)

netroms:
(cherry picked from commit 405f4881db73c424dba5299d3ad05ab6cb5f4330)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
@@ -32,6 +32,7 @@ import static java.time.ZoneId.systemDefault;
 import static java.time.ZonedDateTime.now;
 import static org.hisp.dhis.common.CodeGenerator.isValidUid;
 import static org.hisp.dhis.common.IdentifiableObjectUtils.getUids;
+import static org.hisp.dhis.security.DefaultSecurityService.LOGIN_MAX_FAILED_ATTEMPTS;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
```diff
@@ -755,18 +759,42 @@ public void enableTwoFa(User user, String code) {
...
+
   @Transactional
   @Override
   public void disableTwoFa(User user, String code) {
     if (user.getSecret() == null) {
       throw new IllegalStateException("Two factor is not enabled, enable first");
     }
 
+    if (twoFaDisableIsLocked(user.getUsername())) {
+      throw new IllegalStateException("Too many failed attempts, try again later");
```
david-mackessy, line 787: could we be more helpful to the user & advise how long they need to wait?

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/DefaultUserService.java
netroms: Well, a little bit of the point is to avoid attacks, so that would also help the attackers to script it. Classic security vs. convenience dilemma.

david-mackessy: By not providing that info it does not prevent attacks, only possibly delays them. We're not really preventing anyone obtaining that info given we're open source.

PR #18931, opened 2024-10-28
fix: don't expose password in logs on invalid input (#18156)

netroms:
* fix: don't expose password in logs on invalid input

(cherry picked from commit 3f53d8a4c3f951158c75d955ed0a76a5b03f8316)


samr874:
@netroms  what is the update regading this issue?? is it still underchecking?

jbee (approved):

PR #18945, opened 2024-10-29
chore: Use UID in Deduplication [DHIS2-17790]

enricocolasante:

Use UID in DeduplicationService and propagate the change.
Use UID in PotentialDuplicateCriteria.
Use UID in PotentialDuplicate. This is a Hibernate object so in order to be persisted a new `UIDUserType` was created.
Add @JsonValue annotation on `UID` `toString` method in order to be serialized as a String.

Next steps*

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/UID.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/UID.java
@@ -30,6 +30,7 @@ package org.hisp.dhis.common;
 import static java.util.stream.Collectors.toUnmodifiableSet;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.List;
@@ -67,6 +68,7 @@ public final class UID implements Serializable {
   }
 
   @Override
+  @JsonValue
   public String toString() {
     return value;
   }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
@@ -29,16 +29,14 @@ package org.hisp.dhis.tracker.deduplication;
 
 import static org.hisp.dhis.security.Authorities.ALL;
 
-import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
-import org.hisp.dhis.common.CodeGenerator;
 import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.common.UID;
-import org.hisp.dhis.commons.collection.ListUtils;
 import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.organisationunit.OrganisationUnitService;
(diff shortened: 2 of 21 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/UID.java
```diff
@@ -67,6 +68,7 @@ private UID(String value) {
   }
 
   @Override
+  @JsonValue
```
teleivo, line 71: > A typical usage is that of annotating toString() method so that returned String value is used as the JSON serialization; and if deserialization is needed, there is matching constructor or factory method annotated with [JsonCreator](https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonCreator.html) annotation.
https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonValue.html

Should we also add a `@JsonCreator` or how is a `UID` currently serialized?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelper.java
```diff
@@ -78,44 +76,28 @@ public String getInvalidReferenceErrors(DeduplicationMergeParams params) {
     MergeObject mergeObject = params.getMergeObject();
 
     /*
-     * Step 1: Make sure all uids in mergeObject is valid
+     * Step 1: Make sure all references objects exists in duplicate
```
teleivo, line 79:
```suggestion
     * Step 1: Make sure all reference objects exists in duplicate
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DefaultDeduplicationService.java
```diff
@@ -64,8 +65,8 @@ public PotentialDuplicate getPotentialDuplicateById(long id) {
 
   @Override
   @Transactional(readOnly = true)
-  public PotentialDuplicate getPotentialDuplicateByUid(String uid) {
-    return potentialDuplicateStore.getByUid(uid);
+  public PotentialDuplicate getPotentialDuplicateByUid(UID uid) {
```
teleivo, line 68: `Nullable/Nonnull`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicate.java
```diff
@@ -60,32 +58,28 @@ public class PotentialDuplicate extends BaseIdentifiableObject {
...
 
-  public PotentialDuplicate(String original, String duplicate) {
+  public PotentialDuplicate(UID original, UID duplicate) {
     this.original = original;
     this.duplicate = duplicate;
   }
 
   @JsonProperty
   @JacksonXmlProperty
-  @Property(value = PropertyType.IDENTIFIER, required = Property.Value.TRUE)
```
teleivo, line 70: Is this information automatically provided by the type being UID? As I think this was exposed in the schema

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
```diff
@@ -211,7 +209,7 @@ private void checkDbAndRequestStatus(
               + DeduplicationStatus.MERGED.name());
   }
 
-  private PotentialDuplicate getPotentialDuplicateBy(String id) throws NotFoundException {
+  private PotentialDuplicate getPotentialDuplicateBy(UID id) throws NotFoundException {
```
teleivo, line 212:
```suggestion
  private PotentialDuplicate getPotentialDuplicateBy(UID uid) throws NotFoundException {
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/UID.java
enricocolasante: We have it here https://github.com/dhis2/dhis2-core/blob/2401fff481cfa8c8c31c2a1651348c3f9a231371/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/UID.java#L78

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicate.java
enricocolasante: But we are not using the schema here for anything, I am not sure to be honest, but I think that was not required. Anyway, the required property was not working.
I am asking @dhis2/platform-backend for a review on this PR, to especially check on this line and the `UIDUserType`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/UniqueAttributesSupplier.java
```diff
@@ -208,7 +208,7 @@ private List<UniqueAttributeValue> getDuplicatedUniqueValuesInPayload(
     }
 
     // If an attribute and value appear in 2 or more tracked entities
-    // it means that is not unique
+    // it means that is not uniqueF
```
teleivo, line 211:
```suggestion
    // it means that is not unique
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
```diff
@@ -209,11 +209,12 @@ private void checkDbAndRequestStatus(
...
 
-  private PotentialDuplicate getPotentialDuplicateBy(UID id) throws NotFoundException {
-    return Optional.ofNullable(deduplicationService.getPotentialDuplicateByUid(id))
+  private PotentialDuplicate getPotentialDuplicateBy(UID uid) throws NotFoundException {
+    return Optional.ofNullable(deduplicationService.getPotentialDuplicateByUid(uid))
         .orElseThrow(
             () ->
-                new NotFoundException("No potentialDuplicate records found with id '" + id + "'."));
+                new NotFoundException(
+                    "No potentialDuplicate records found with id '" + uid + "'."));
```
teleivo, line 217:
```suggestion
new NotFoundException(uid)
```

Should be enough

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicate.java
david-mackessy: My understanding of how the schemas work is that the line removed here would only ever have worked if there was a schema for `PotentialDuplicate`. And it looks like that does not exist. So this seems fine to me. As long as you're seeing expected behaviour. If there's a test covering that situation where no `PotentialDuplicate` can be missing this property, even better.

david-mackessy (approved):

PR #18948, opened 2024-10-29
ci: remove building/testing with tomcat 8.5 [2.41]

radnov:
Backport of https://github.com/dhis2/dhis2-core/pull/18899

Requested review from everyone that has been involved in discussing this one way or another, mostly for visibility.

### DIFF
--- a/jenkinsfiles/rebuild-image
+++ b/jenkinsfiles/rebuild-image
@@ -40,11 +40,12 @@ pipeline {
                         env.VERSIONS_TO_REBUILD.tokenize(" ").each { version ->
                             def majorVersion = version.split('\\.')[0].toInteger()
                             if (majorVersion >= 42) {
-                                env.BASE_IMAGE = "tomcat:10.0-jre17"
+                                // THIS MUST be kept in sync with the jib.from.image in ../dhis-2/dhis-web-server/pom.xml
+                                env.BASE_IMAGE = "tomcat:10.1.30-jre17"
                             } else if (majorVersion == 41) {
-                                env.BASE_IMAGE = "tomcat:9.0-jre17"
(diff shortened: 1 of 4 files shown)


jenkinsfiles/rebuild-image
```diff
@@ -40,11 +40,12 @@ pipeline {
                         env.VERSIONS_TO_REBUILD.tokenize(" ").each { version ->
                             def majorVersion = version.split('\\.')[0].toInteger()
                             if (majorVersion >= 42) {
-                                env.BASE_IMAGE = "tomcat:10.0-jre17"
+                                // THIS MUST be kept in sync with the jib.from.image in ../dhis-2/dhis-web-server/pom.xml
+                                env.BASE_IMAGE = "tomcat:10.1.30-jre17"
                             } else if (majorVersion == 41) {
```
teleivo, line 45: remove `else if (majorVersion == 41)` as it sets the same version as the `else` branch.

does this pipeline even need to know about other versions? I mean if this one rebuilds images on this branch thats all it needs to know right? if `rebuild-image` on master is the one thats used this file should not exist 🤔

teleivo (approved):

jenkinsfiles/rebuild-image
radnov: The difference is that 41 is using `*-jre17` and lower versions are using `*-jre11`.

I backported the `rebuild-image` to all supported branches initially, which might have been a bit of an oversight. However, I think that was before releasing 41 (hence `master` was 41). Once we release 42 (current `master`), the file will be on the 42 dev branch and on `master`, so I suppose we either have to keep deleting it with releases or keep backporting it to lower versions branches as is and making sure it doesn't drift.

david-mackessy (approved):

PR #18985, opened 2024-10-31
feat: OpenAPI page info header [DHIS2-18266]

jbee:
### Summary
Adds a header to the HTML rendering that shows the scope and some statistics.

Features
* links to switch view to one of the currently selected scopes
* links to remove one scope from the current selection
* link to view the JSON source of the current document
* link to view the YAML source of the current document
* link to add JSON source information to the current document
* link to remove all scopes (full document)

Bugfix: This also contains a important bugfix for shared parameters. With the stats it became clear that there way to many shared parameters. I found that parameters would be added multiple times, once for each endpoint for which they were used. This wasn't recognized before as this leads to duplicate keys in the JSON structure which still works as expected. Just that the document was way bigger than necessary.

Example of the header
DHIS2-API-2-42-10-31-2024_03_46_PM


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/ApiExtractor.java
```diff
@@ -553,14 +521,17 @@ private void extractParams(Api.Endpoint endpoint, Class<?> paramsObject) {
     String sharedName = getSharedName(paramsObject, shared, null);
     if (sharedName != null) {
       Map<Class<?>, List<Api.Parameter>> sharedParameters = api.getComponents().getParameters();
+      boolean addShared = !sharedParameters.containsKey(paramsObject);
```
jbee, line 524: FYI: without this check the same params class would add all its parameters each time the params class occurred in an endpoint method resulting in 18K parameters instead of ~250. Woopsy! :sweat_smile:

david-mackessy (approved):
I've only just noticed the hardcoded css & JS code in Java strings! I'm sure there's a "good" reason for this 😄 Care to share?

jbee:
> I've only just noticed the hardcoded css & JS code in Java strings! I'm sure there's a "good" reason for this 😄 Care to share?

That is really just a relic from the prototyping age. The benefit is that you can change Java, CSS and JS all in the same file, recompile and reload. Makes for fast turn around. But it is about time to move them to their own files. Will do that in one of the next OpenAPI PRs.

netroms (approved):

larshelge (approved):

PR #18986, opened 2024-10-31
 feat: Migrate tracker change logs into new table [DHIS2-18117]

muilpp:
This PR adds a script to create and migrate data from `trackedentitydatavalueaudit` to `eventchangelog`, with CRUD support.

The migration script uses a window function to pull current and previous values by audit type, similar to the first change log iteration. Corner cases, as agreed with front end:

    1. Null created date: Defaults to 1970-01-01.
    2. Unset `createdby`: Defaults to '--'; front end will handle display.
    3. Unset `eventid` or `dataelementid`: Stays in old table.
    4. Non-standard change types (not `CREATE`, `UPDATE`, `DELETE`): Stays in old table.
    5. Primary key constraint: `trackedentitydatavalueauditid` is always non-null, so `eventchangelogid` is too.
 
 This migration script has been tested against SL, Sri Lanka (pending results), and Laos databases.
 
 Hibernate defers persistence, initially inserting a null in the foreign key attribute until the referenced data is available. This worked when the old table allowed null values in the `eventid` attribute, but it’s now problematic since nulls aren’t allowed in the new table. To solve this:
 	1. New events -> Data values update after event persistence. If not, the foreign key constraint fails.
 	2. Updates/deletes -> Data values update before event persistence. If not, deleted values are not updated in the event object and consequently, not removed.
 
Additional changes:
 - Removed the field `providedElsewhere`, as agreed [here](https://dhis2.slack.com/archives/GA4HSQ1HV/p1727273794812869)
 - Removed the test class `TrackedEntityDataValueChangeLogTest`, it's no longer necessary.
 
 Left to do:
 - Remove old endpoint, service and store.
 - Add occurredDate, scheduledDate and geometry
 - Migrate attribute change logs


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -264,22 +266,15 @@ private void deleteDataValue(
 
   private void logDataValueChange(
       String userName,
-      DataElement de,
+      DataElement dataElement,
+      String eventProperty,
```
enricocolasante, line 270: Can we remove this parameter and hardcode it in the method?

enricocolasante (commented):
Is the old table ever updated?
In the old table there are going to remain only invalid data (null eventid) and not changelog data (READ, SEARCH). Is there an endpoint to still access these data?

I am just trying to understand why we are not deleting the table once and for all

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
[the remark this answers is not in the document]
muilpp: For now it's always assigned to the same value, but in the next PR I'll need it, that's why I added it as a parameter.

muilpp:
> Is the old table ever updated? In the old table there are going to remain only invalid data (null eventid) and not changelog data (READ, SEARCH). Is there an endpoint to still access these data?
> 
> I am just trying to understand why we are not deleting the table once and for all

We agreed in [this thread](https://dhis2.slack.com/archives/C05S1NTHBJ9/p1727438767825559?thread_ts=1727426983.066279&cid=C05S1NTHBJ9) to keep the non migratable data in the old table. 
Once I'm done with the change logs, there will no longer be an endpoint to access this data, we'll simply leave it there in case they might need it, as we’re not sure of its potential value to them. It will be up to each implementation to decide what to do with that table.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -55,16 +57,31 @@ Page<EventChangeLog> getEventChangeLog(
...
   @Deprecated(since = "2.42")
   int countTrackedEntityDataValueChangeLogs(TrackedEntityDataValueChangeLogQueryParams params);
 
   void deleteTrackedEntityDataValueChangeLog(Event event);
 
+  void deleteEventChangeLog(Event event);
+
   void deleteTrackedEntityDataValueChangeLog(DataElement dataElement);
 
+  @Transactional
```
ameenhere, line 82: Is this Transactional annotation needed here?

muilpp: nope, it's already in the implementation

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
```diff
@@ -264,22 +266,15 @@ private void deleteDataValue(
 
   private void logDataValueChange(
```
ameenhere, line 267: Should we rename this method to better reflect what its doing? I am assuming it does not only log data value changes? So something like `logEventChange()` ? or similar?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_24__Migrate_event_change_logs.sql
```diff
@@ -0,0 +1,60 @@
...
+    previousvalue varchar(50000) null,
+    changelogtype varchar(100) not null,
+    created timestamp not null,
+    createdby varchar(255) not null,
+    constraint eventchangelog_pkey primary key (eventchangelogid),
+    constraint fk_eventchangelog_dataelementid foreign key (dataelementid) references dataelement(dataelementid),
+    constraint fk_eventchangelog_eventid foreign key (eventid) references event(eventid)
+);
+
+create index if not exists index_eventchangelog_programstageinstanceid on eventchangelog using btree (eventid);
```
ameenhere, line 21: "programstageinstanceid" ? why not "eventid" ?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_24__Migrate_event_change_logs.sql
```diff
@@ -0,0 +1,60 @@
...
+            lag (t.value) over (partition by t.eventid, t.dataelementid order by t.created desc) as currentchangelogvalue,
+            t.value AS previouschangelogvalue,
+            e.eventdatavalues -> d.uid  ->> 'value' as currentvalue
+     from trackedentitydatavalueaudit t
+              join event e using (eventid)
+              join dataelement d using (dataelementid)
+     order by t.trackedentitydatavalueauditid) cl
+where cl.audittype in ('CREATE', 'UPDATE', 'DELETE')
+and not exists (
+    select 1 from eventchangelog ecl where ecl.eventchangelogid = cl.trackedentitydatavalueauditid
```
ameenhere, line 52: I did not understand why this clause is needed? This is a new table right. Maybe I am missing something?

muilpp: Any other statement in this script will either not be executed or will have the same outcome if run multiple times, except for this insert statement. This is my way of making the insert idempotent, maybe it's a bit over the top but it guarantees I'm not inserting the same change log twice.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EventPersister.java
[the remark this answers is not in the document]
teleivo: best inline this method as it just calls the eventService, then any discussions here are mute 😋

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -268,6 +264,29 @@ void shouldReturnChangeLogsWhenDataValueIsCreatedUpdatedAndDeleted()
         () -> assertCreate(dataElement, "15", changeLogs.getItems().get(2)));
   }
 
+  @Test
+  void shouldReturnOnlyUserNameWhenUserDoesNotExistInDatabase()
+      throws ForbiddenException, NotFoundException {
+    Event event = getEvent("QRYjLTiJTrA");
+    String dataElementUid = event.getEventDataValues().iterator().next().getDataElement();
+    DataElement dataElement = manager.get(DataElement.class, dataElementUid);
+    User fakeUser = new User();
```
teleivo, line 273: Use `deletedUser` rather than fake one as that is what this is supporting.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
```diff
@@ -52,6 +53,20 @@ public class DefaultEventChangeLogService implements EventChangeLogService {
   private final HibernateTrackedEntityDataValueChangeLogStore trackedEntityDataValueChangeLogStore;
   private final TrackerAccessManager trackerAccessManager;
 
+  @Transactional
+  @Override
+  public void addEventChangeLog(
```
teleivo, line 58: Are we always calling this with either a `dataElement` or an `eventProperty` (the other arg being null)? If so providing two methods make a clearer API. The repo underneath can stay the same. Having many string args like this is awkward in terms of programmer errors. If you put it in the wrong order you got a bug.

teleivo (commented):
1.

> Removed the field providedElsewhere, as agreed here

You only removed the field from the event changelogs right? As we still store it as part of an events data value.

2. Why is

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackedEntityDataValueChangeLogTest.java

no longer necessary?

3. 

> Remove old endpoint, service and store.

Which old endpoint?

4.

> This migration script has been tested against SL, Sri Lanka (pending results), and Laos databases.

What checks did you to against that data? Always provide the method you use with numbers so we can find holes in it. Like in a scientific experiment 😋 

5.

> Hibernate defers persistence, initially inserting a null in the foreign key attribute until the referenced data is available. This worked when the old table allowed null values in the eventid attribute, but it’s now problematic since nulls aren’t allowed in the new table. To solve this:
> 1. New events -> Data values update after event persistence. If not, the foreign key constraint fails.
> 2. Updates/deletes -> Data values update before event persistence. If not, deleted values are not updated in the event object and consequently, not removed.
> 
> While using JDBC, and because Hibernate doesn't commit the data to the table until the end of the transaction, I need to flush the data, so the FK constraints pass. The data is not committed to the table, so it allows roll back if necessary

I don't fully get this.

> Hibernate doesn't commit the data to the table until the end of the transaction

Is "hibernate committing" different than Postgres committing? As https://www.postgresql.org/docs/current/sql-commit.html will commit and close the transaction.

>  1. New events -> Data values update after event persistence. If not, the foreign key constraint fails.

To me this reads as if we are committing the event to the DB, then committing the data value, so 2 transactions, right?

muilpp:
> > Removed the field providedElsewhere, as agreed here
> 
> You only removed the field from the event changelogs right? As we still store it as part of an events data value.

Yes, in this PR I'm talking about the changelog scope only

> 
>     2. Why is
> 
> 
> dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/TrackedEntityDataValueChangeLogTest.java
> 
> no longer necessary?

Because we don't persist data to `trackedentitydatavalueaudit` anymore, so the test won't find data to work with.

> 
> 
> > Remove old endpoint, service and store.
> 
> Which old endpoint?

/api/audits/trackedEntityDataValue

> 
> 
> > This migration script has been tested against SL, Sri Lanka (pending results), and Laos databases.
> 
> What checks did you to against that data? Always provide the method you use with numbers so we can find holes in it. Like in a scientific experiment 😋

This is what I'm doing:
1. After the migration, I check if the total number of records in `eventchangelog` is the same as the ones in `trackedentitydatavalueaudit`
2. If it's not, I check that the expected migrated records (the ones with a valid audit type, non-null eventid and non-null dataelementid) are present in the new table
So far with this two validations it worked in SL, Laos and a small database in Sri Lanka. I'm waiting for the results of a bigger database where there are supposed to be 4 million change log records.

Then, to have a bit of context I check how many `createdby` and/or `created` attributes are not set and how many `eventid` and/or `dataelementid` are null.

> 
> 
> > Hibernate defers persistence, initially inserting a null in the foreign key attribute until the referenced data is available. This worked when the old table allowed null values in the eventid attribute, but it’s now problematic since nulls aren’t allowed in the new table. To solve this:
> > 
> > 1. New events -> Data values update after event persistence. If not, the foreign key constraint fails.
> > 2. Updates/deletes -> Data values update before event persistence. If not, deleted values are not updated in the event object and consequently, not removed.
> > 
> > While using JDBC, and because Hibernate doesn't commit the data to the table until the end of the transaction, I need to flush the data, so the FK constraints pass. The data is not committed to the table, so it allows roll back if necessary
> 
> I don't fully get this.
> 
> > Hibernate doesn't commit the data to the table until the end of the transaction
> 
> Is "hibernate committing" different than Postgres committing? As https://www.postgresql.org/docs/current/sql-commit.html will commit and close the transaction.
> 
> > 1. New events -> Data values update after event persistence. If not, the foreign key constraint fails.
> 
> To me this reads as if we are committing the event to the DB, then committing the data value, so 2 transactions, right?

Everything happens in one transaction, so there's just one commit right before the transaction is terminated.
The issue here JDBC can't see the data Hibernate persists because it's not committed in the database. At the moment JDBC comes into the picture, Hibernate just holds the data in its session, that's why it's not visible to anyone else. To make it visible to JDBC, I call `flush`, which synchronizes the session with the database.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -55,16 +56,30 @@ Page<EventChangeLog> getEventChangeLog(
   List<TrackedEntityDataValueChangeLog> getTrackedEntityDataValueChangeLogs(
       TrackedEntityDataValueChangeLogQueryParams params);
 
+  @Deprecated(since = "2.42")
```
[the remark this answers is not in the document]
muilpp: since we decided [we will remove the endpoint](https://dhis2.slack.com/archives/C05S1NTHBJ9/p1730292188352939?thread_ts=1730290544.389879&cid=C05S1NTHBJ9), it makes sense to remove everything underneath too, unless some other service is using this, which I think it's not the case

teleivo:
> Then, to have a bit of context I check how many `createdby` and/or `created` attributes are not set and how many `eventid` and/or `dataelementid` are null.

That bit is crucial 👌🏻 If you don't get any "problematic" data in your tests you have not tested it can deal with it.

> Everything happens in one transaction, so there's just one commit right before the transaction is terminated.
The issue here is JDBC can't see the data Hibernate persists because it's not committed in the database yet. At the point when JDBC is involved, Hibernate just holds the data in its session, that's why it's not visible to anyone else. To make it visible to JDBC, I call flush, which synchronizes the session with the database.

I think you/we need to become more precise with our language and understand our tools like hibernate better 😅 

> Everything happens in one transaction, so there's just one commit right before the transaction is terminated.

I am not sure this is true.

> JDBC can't see the data Hibernate persists because it's not committed in the database yet

explains why. Hibernate likely has its own transaction and JDBC is not part of it. The transaction is what isolates the changes/make them invisible to anyone outside of it. Hence, there will be 2 transactions, right? One for Hibernate and one for JDBC.

Understanding that well is important as doing work in separate transactions means that we open ourselves up for data races. It's not guaranteed that no other transaction changes data in between the two. How unlikely that may be.

muilpp:
> > Then, to have a bit of context I check how many `createdby` and/or `created` attributes are not set and how many `eventid` and/or `dataelementid` are null.
> 
> That bit is crucial 👌🏻 If you don't get any "problematic" data in your tests you have not tested it can deal with it.

I know, that's why I want to run this script in a bigger instance in Sri Lanka. So far, it worked on 3 different databases, but doesn't mean it won't fail. I can't test it in all implementations, but I'm covering all the cases I could think of.

> 
> > Everything happens in one transaction, so there's just one commit right before the transaction is terminated.
> > The issue here is JDBC can't see the data Hibernate persists because it's not committed in the database yet. At the point when JDBC is involved, Hibernate just holds the data in its session, that's why it's not visible to anyone else. To make it visible to JDBC, I call flush, which synchronizes the session with the database.
> 
> I think you/we need to become more precise with our language and understand our tools like hibernate better 😅
> 
> > Everything happens in one transaction, so there's just one commit right before the transaction is terminated.
> 
> I am not sure this is true.
> 
> > JDBC can't see the data Hibernate persists because it's not committed in the database yet
> 
> explains why. Hibernate likely has its own transaction and JDBC is not part of it. The transaction is what isolates the changes/make them invisible to anyone outside of it. Hence, there will be 2 transactions, right? One for Hibernate and one for JDBC.
> 
> Understanding that well is important as doing work in separate transactions means that we open ourselves up for data races. It's not guaranteed that no other transaction changes data in between the two. How unlikely that may be.

You’re right. As we discussed on the call, the theory says that when using the same data source they should be in the same transaction. However, that’s not the case here. Even though JDBC and Hibernate are using the same data source, they still operate in separate transactions. In a test where the JDBC query fails, I've seen that changes made by Hibernate aren’t rolled back. I want to understand why they don’t interact as expected, but for now, I don’t feel comfortable combining them. Since the importer is using Hibernate, I’ll stick to it for the change log feature as well

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  public Page<EventChangeLogDto> getEventChangeLogs(
+      UID event, List<Order> order, PageParams pageParams) {
+
+    String hql =
+        String.format(
+            """
+        select new org.hisp.dhis.tracker.export.event.EventChangeLogDto (de.uid, ecl.currentValue, ecl.previousValue, ecl.changeLogType, ecl.created, ecl.createdBy, ui.firstName, ui.surname, COALESCE(ui.username, ecl.createdBy), ui.uid)
+        from EventChangeLog ecl
```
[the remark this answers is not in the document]
enricocolasante: Maybe I missed some discussion and you have a specific reason for it, but why do we have an `EventChangelogDTO` type at all?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/EventChangeLog.java
```diff
@@ -0,0 +1,45 @@
...
+public record EventChangeLog(
+    @JsonProperty User createdBy,
+    @JsonProperty Date createdAt,
+    @JsonProperty String type,
+    @JsonProperty Change change) {
+
+  public record Change(@JsonProperty DataValueChange dataValue) {}
+
+  public record DataValueChange(
+      @JsonProperty String dataElement,
```
enricocolasante, line 42:
```suggestion
      @JsonProperty UID dataElement,
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -0,0 +1,136 @@
...
+    entityManager.unwrap(Session.class).save(eventChangeLog);
+  }
+
+  public Page<EventChangeLogDto> getEventChangeLogs(
+      UID event, List<Order> order, PageParams pageParams) {
+
+    String hql =
+        String.format(
+            """
+        select new org.hisp.dhis.tracker.export.event.EventChangeLogDto (de.uid, ecl.currentValue, ecl.previousValue, ecl.changeLogType, ecl.created, ecl.createdBy, ui.firstName, ui.surname, COALESCE(ui.username, ecl.createdBy), ui.uid)
```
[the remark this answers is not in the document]
muilpp: that's a mistake to always fetch `createdBy`, it should only be in case no match is found in the user table.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_25__Migrate_event_change_logs.sql
```diff
@@ -0,0 +1,60 @@
...
+    previousvalue varchar(50000) null,
+    changelogtype varchar(100) not null,
+    created timestamp not null,
+    createdby varchar(255) not null,
+    constraint eventchangelog_pkey primary key (eventchangelogid),
+    constraint fk_eventchangelog_dataelementid foreign key (dataelementid) references dataelement(dataelementid),
+    constraint fk_eventchangelog_eventid foreign key (eventid) references event(eventid)
+);
+
+create index if not exists index_eventchangelog_eventid on eventchangelog using btree (eventid);
```
ameenhere, line 21: Can we also add the necessary indexes based on our query usage? I remember we are going to sort by created? So an index (or multiple index) to support combinations of filtering and sorting?

muilpp: We will support sorting by date, user, and data element/property, and will also allow filtering by these same attributes. I think it makes sense to refactor the existing index to a composite on `(eventid, created DESC)` since this is the default query. Even when sorting or filtering by other attributes this index will still be used because we will always join the query by `eventid`.

I'm not sure about adding more indexes, as this depends entirely on user parameters. Covering other possible combinations would mean indexing everything except the values and `changeLogType`, which could increase overhead and impact performance. 

Do you agree?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -55,16 +56,29 @@ Page<EventChangeLog> getEventChangeLog(
...
   void addTrackedEntityDataValueChangeLog(
       TrackedEntityDataValueChangeLog trackedEntityDataValueChangeLog);
 
+  void addDataValueChangeLog(
+      Event event,
+      DataElement dataElement,
+      String currentValue,
```
enricocolasante, line 66: Maybe not so important but I would expect previous to go before than current as it is done in the view object.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventChangeLogMapper.java
```diff
@@ -0,0 +1,64 @@
...
+  @Mapping(target = "surname", source = "createdBy.surname")
+  User mapUser(org.hisp.dhis.tracker.export.event.EventChangeLog eventChangeLog);
+
+  @Mapping(target = "dataElement", source = "dataElement.uid")
+  @Mapping(target = "previousValue", source = "previousValue")
+  @Mapping(target = "currentValue", source = "currentValue")
+  DataValueChange mapDataValueChange(
+      org.hisp.dhis.tracker.export.event.EventChangeLog eventChangeLog);
+
+  default UID mapStringToUID(String value) {
```
enricocolasante, line 58: You can remove this default method and use `UIDMapper` in the @Mapper annotation

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogDto.java
```diff
@@ -0,0 +1,65 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.export.event;
+
+import java.util.Date;
+import org.hisp.dhis.changelog.ChangeLogType;
+import org.hisp.dhis.common.UID;
+
+public record EventChangeLogDto(
```
enricocolasante, line 34: Is this used?

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
```diff
@@ -0,0 +1,43 @@
...
+      <type name="org.hibernate.type.EnumType">
+        <param name="enumClass">org.hisp.dhis.changelog.ChangeLogType</param>
+        <param name="useNamed">true</param>
+        <param name="type">12</param>
+      </type>
+    </property>
+
+    <property name="created" type="timestamp" not-null="true"/>
+
+    <property name="createdByUsername" column="createdby" type="string"/>
```
enricocolasante, line 37: Is this used?

ameenhere (approved):

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
muilpp: yes, it is used when creating a new entry and when fetching the results it's used as a condition for the join to the user table

enricocolasante (approved):

PR #18988, opened 2024-10-31
refactor: Reduce functional code complexity [DHIS2-18306]

maikelarabori:
This PR aims to reduce functional code that could be simplified targeting code reusage and readability.
The DataHandler has to be split into different classes, at some point, but it was not in scope for this change.
I also added unit tests for the code changed.


larshelge (approved):

larshelge:
Great effort, great to reduce stream code and the added tests.

luciano-fiandesio (approved):

PR #19008, opened 2024-11-01
feat: email-based authentication codes for MFA/2FA

netroms:
## Summary
This adds 2FA login with a code sent via email as a new feature.

NOTE: This PR removes the enforced 2FA enrollment with TOTP functionality and does not address this for email 2FA either. The requirements for enforced enrollment needs to be addressed in a separate JIRA ticket. Enforced enrollment for TOTP was never used, so it's safe to remove.

TODO:
* Add rate limiting to email 2FA sending
* Do sending of 2FA email async so we don't block UI
* Enforced enrollment

### Automatic tests
* TwoFactorControllerTest
* UserServiceTest
* AuthenticationControllerTest
* LoginTest (E2E)

### Manual tests
* Enable 2FA email in system settings: POST api/42/systemSettings/email2FAEnabled?value=true
* Set up a working email SMTP server in settings
* Verify your email: POST api/account/sendEmailVerification (email with code will be sent to your verified email)
* Enroll in email 2FA: POST api/2fa/enrollEmail2FA
* Enable email 2FA with code sent to your email: POST api/2fa/enable with {code:CODE_SENT_TO_EMAIL}
* Try to login, code should be sent to email
* Now login should show 2FA code input, enter the code and submit
* Observe you are logged in

### Jira:
[DHIS2-13334](https://dhis2.atlassian.net/browse/DHIS2-13334)

[DHIS2-13334]: https://dhis2.atlassian.net/browse/DHIS2-13334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -195,21 +195,30 @@ public enum ErrorCode {
...
+  E3029("User must start 2FA enrollment first"),
   E3030(
       "User cannot update their own user's 2FA settings via this API endpoint, must use /2fa/enable or disable API"),
-  E3031("Two factor authentication is not enabled"),
+  E3031("User has not enabled 2FA"),
   E3032("User `{0}` does not have access to user role"),
   E3040("Could not resolve JwsAlgorithm from the JWK. Can not write a valid JWKSet"),
   E3041("User `{0}` is not allowed to change a user having the ALL authority"),
   E3042("Too many failed disable attempts. Please try again later"),
+  E3043("User has not a verified email, please verify your email first before you enable 2FA"),
```
david-mackessy, line 213: this doesn't read correctly. Instead of `User has not a verified email`, we could use `User does not have a verified email`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -195,21 +195,30 @@ public enum ErrorCode {
...
+  E3044("TOTP 2FA is not enabled"),
+  E3045("Email based 2FA is not enabled in the system settings"),
+  E3046("TOTP 2FA is not enabled in the system settings"),
+  E3047("User is not in TOTP 2FA enrollment mode"),
+  E3048("User has not email 2FA enabled"),
```
david-mackessy, line 218: this doesn't read correctly. Instead of `User has not email 2FA enabled`, we could use `User does not have email 2FA enabled`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/twofa/TwoFactorAuthService.java
```diff
@@ -0,0 +1,316 @@
...
+/**
+ * @author Morten Svanæs
+ */
+@Slf4j
+@Service
+@AllArgsConstructor
+public class TwoFactorAuthService {
+
+  public static final String TWO_FACTOR_AUTH_REQUIRED_RESTRICTION_NAME = "R_ENABLE_2FA";
+  public static final long TWOFA_EMAIL_CODE_EXPIRY_MILLIS = 900_000; // 15 minutes
```
david-mackessy, line 74: can be an int as within the int range

dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
```diff
@@ -1798,6 +1798,16 @@ email_verify_1_2nd_paragraph=Please click the link below to verify the email add
...
 verify_email_subject=Verify email address
 
+#-- 2FA code email -------------------------------------------------------------#
+
+email_2fa_subject=Your Two-Factor Authentication Code for
+email_2fa_1_greeting=Dear
+email_2fa_1_1st_paragraph_application_title=You have requested a two-factor authentication code to log into your account.
+email_2fa_1_2nd_paragraph=Your two-factor authentication code:
+email_2fa_1_3rd_paragraph=Please use this code within 15 minutes to log into your account.
+email_2fa_1_4rd_paragraph=If you did not request this code, please contact your system administrator immediately to secure your account.
```
david-mackessy, line 1808: `4rd` should be `4th`

dhis-2/dhis-services/dhis-service-core/src/main/resources/twofa_email_body_template_v1.vm
```diff
@@ -0,0 +1,15 @@
...
+
+$object.i18n.getString(
+    "email_2fa_1_1st_paragraph_application_title" )
+
+$object.i18n.getString( "email_2fa_1_2nd_paragraph" )
+${object.code}
+
+$object.i18n.getString( "email_2fa_1_3rd_paragraph" )
+
+$object.i18n.getString( "email_2fa_1_4rd_paragraph" )
```
david-mackessy, line 11: this will need updating after updating the i18n property above. `4rd` to `4th`

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AccountController.java
```diff
@@ -568,6 +576,16 @@ public void verifyEmail(@RequestParam String token) throws ConflictException {
     }
   }
 
+  @GetMapping("/listSessions")
```
david-mackessy, line 579: Any auth needed to call this or is this something we want open for all users?

dhis-2/dhis-test-e2e/pom.xml
```diff
@@ -36,9 +36,38 @@
...
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+      <version>${spring.version}</version>
+    </dependency>
```
david-mackessy, line 70: brining spring into the e2e seems like overkill. Does it give us something RestAssured does not have?

david-mackessy (approved):
some small comments left, good job 👍

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/SessionController.java
```diff
@@ -0,0 +1,119 @@
...
+    List<SessionInformation> allSessions = new ArrayList<>();
+    List<User> allUsers = userService.getAllUsers();
+    for (User user : allUsers) {
+      allSessions.addAll(
+          sessionRegistry.getAllSessions(userService.createUserDetails(user.getUid()), true));
+    }
+    return allSessions;
+  }
+
+  @PostMapping(value = "/{username}/invalidate", produces = APPLICATION_JSON_VALUE)
```
jbee, line 92: Same as below, isn't using `@DeleteMapping("/{username}")` more fitting?
Also it does not `produces = APPLICATION_JSON_VALUE`. Like a typical delete it doesn't return anything, which should use `@ResponseStatus(HttpStatus.NO_CONTENT)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/SessionController.java
```diff
@@ -0,0 +1,119 @@
...
+    if (userDetails != null) {
+      List<SessionInformation> allSessions = sessionRegistry.getAllSessions(userDetails, false);
+      allSessions.forEach(SessionInformation::expireNow);
+    }
+  }
+
+  @PostMapping(value = "/invalidateAll", produces = APPLICATION_JSON_VALUE)
+  @RequiresAuthority(anyOf = ALL)
+  public void invalidateAllSessions() {
+    invalidateAllUserSessions();
```
jbee, line 106: As far as I can see this is only called here so why even have another method?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/SessionController.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private void invalidateAllUserSessions() {
+    List<User> allUsers = userService.getAllUsers();
+    for (User user : allUsers) {
```
jbee, line 111: This seems like a unnecessary cumbersome way to invalidate all sessions. The service called from the controller should have this as a dedicated operation as it likely can just do something similar to a  `Map.clear()` to perform it instead of fetching and looping stuff here.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/TwoFactorController.java
```diff
@@ -74,142 +70,125 @@
...
+   * Enrolls the user in TOTP 2FA and generates a QR code for the user to scan.
+   *
+   * @throws IOException The QR code could not be generated.
+   * @throws ConflictException The user is already enrolled in 2FA.
+   */
+  @OpenApi.Response(byte[].class)
+  @GetMapping(
+      value = {"/qrCode"},
+      produces = APPLICATION_OCTET_STREAM_VALUE)
+  @ResponseStatus(HttpStatus.ACCEPTED)
```
jbee, line 135: These were my initial thoughts below but I think you just moved the endpoint here and I guess we have to keep it like this for backwards compatibility?

> `ACCEPTED` seems an odd choice for a status as the endpoints task seems to be to respond with a QR code which it always does so there is nothing left or unfinished that would make `ACCEPTED` the better choice or am I missing something? In terms of 2FA the process is not complete but the work of this endpoint is complete.
> Since we call `enrollTOTP2FA` that to me looks like a modification of sorts so `GET` seems the wrong HTTP method. Isn't this more of a `PUT` or `POST`?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -545,21 +552,39 @@ public void unexpireUser(@PathVariable("uid") String uid) throws Exception {
     setExpires(uid, null);
   }
 
+  @PostMapping("/{uid}/invalidateSessions")
+  @ResponseStatus(value = HttpStatus.NO_CONTENT)
+  @RequiresAuthority(anyOf = ALL)
+  public void invalidateSessions(@PathVariable("uid") String uid) {
```
jbee, line 558: maybe have these session related endpoints in the `/sessions` controller?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DhisWebApiWebAppInitializer.java
```diff
@@ -95,6 +95,12 @@ private DhisConfigurationProvider getConfig() {
   public static void setupServlets(
       ServletContext context, AnnotationConfigWebApplicationContext webApplicationContext) {
 
+    context
+        .addFilter(
+            "SpringSessionRepositoryFilter",
+            new DelegatingFilterProxy("springSessionRepositoryFilter"))
```
jbee, line 101: Does it matter that the strings here are not the same? One is lower case S and one upper.

jbee (changes requested):
I saw you moved more (internal) APIs from using a user's UID to the username which I think is a good thing. As we discussed it seems to me that the username is our identifier when it comes to questions of authentication. :+1: 

The are a handful or so details to correct and some suggestions about the REST API endpoints.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/SessionController.java
netroms: I know, but it's because the internals of the Redis based session store lack a way of looping through all sessions.

david-mackessy: just for reference, the most users in a system I've heard of is 10k. Depending how often this method is called, it could impact the system.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/CodeGenerator.java
```diff
@@ -143,6 +144,28 @@ public static char[] generateSecureRandomCode(int codeSize) {
...
+    return Arrays.copyOf(bytes, length);
+  }
+
+  /**
+   * Generates a string of random numeric characters.
+   *
+   * @param length the number of characters in the code.
+   * @return the code.
+   */
+  public static char[] generateSecureRandomNumber(int length) {
```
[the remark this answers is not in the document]
netroms: It's actually just a little bit security related, trying to not use strings, which is "cached" in the JVMs.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DhisWebApiWebAppInitializer.java
netroms: This is just naming convention for filters.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/security/SessionController.java
netroms: I don't think this is going to be an issue, the only use-case right now is the LoginTest to invalidate sessions after each test. It's also only possible to use by ALL authority.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/twofa/TwoFactorType.java
```diff
@@ -32,10 +32,10 @@
...
 public enum TwoFactorType {
   NONE,
-  TOTP,
-  EMAIL,
-  ENROLLING_TOTP,
-  ENROLLING_EMAIL;
+  TOTP, // Enabled state, using TOTP
+  EMAIL, // Enabled state, using email
+  ENROLLING_TOTP, // Enrolling state, using TOTP
+  ENROLLING_EMAIL; // Enrolling state, using email
```
david-mackessy, line 38: if these are states then it might be better to have that represented more clearly  in the type name and states, without needing the comments. Suggestion:
```java
enum TwoFactorState {
  NOT_ENABLED,
  EMAIL_ENABLED,
  TOTP_ENABLED,
  EMAIL_ENROLLED,
  TOTP_ENROLLED;
}
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/twofa/TwoFactorAuthService.java
```diff
@@ -0,0 +1,347 @@
...
+
+  /**
+   * Enroll user in email-based 2FA authentication.
+   *
+   * @param username The user that is being enrolled.
+   */
+  @Transactional
+  public void enrollEmail2FA(@Nonnull String username) throws ConflictException {
+    User user = userService.getUserByUsername(username);
+    if (user == null) {
```
larshelge, line 114: Not a big deal or necessarily need to change. I have a preference for splitting validation logic and business logic into separate methods. You could have a method `ErrorCode validateEnrollEmail2Fa(..)` that returns null if validation is success and then an `ErrorCode` if it failed. Then potentially throw the exception. That makes it easier to unit test the validation in isolation. Applies further down as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/twofa/TwoFactorType.java
netroms: I like that idea.

vietnguyen (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_29__add_2FA_type_userinfo.sql
```diff
@@ -0,0 +1,8 @@
+-- DHIS2-13334: Add twofactortype column to userinfo table
+ALTER TABLE userinfo
+    ADD COLUMN IF NOT EXISTS twofactortype character varying(50);
+
+-- Set all existing users to have the default twofactortype TOTP, if the secret column is not empty.
+UPDATE userinfo
+SET twofactortype = 'TOTP'
+WHERE secret IS NOT NULL;
```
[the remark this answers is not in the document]
larshelge: Agree that db schema must match Hibernate and Java annotation schema. When setting all null rows to 'TOTP' then setting column not null is safe.

```
alter table userinfo alter column twofactortype set not null;
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/twofa/TwoFactorAuthService.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    if (isInvalid2FACode(user, code)) {
+      throw new ForbiddenException(ErrorCode.E3023);
+    }
+
+    setEnabled2FA(user.getUsername(), CurrentUserUtil.getCurrentUserDetails());
+  }
+
+  public void setEnabled2FA(@Nonnull String username, @Nonnull UserDetails actingUser) {
```
jbee, line 165: I think this is meant to be `private`?

jbee (commented):
There a some comments left 
- on visibility and TX-annotations in services
- on which HTTP method and response code to use for the new endpoints (session controller)

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_29__add_2FA_type_userinfo.sql
```diff
@@ -0,0 +1,8 @@
+-- DHIS2-13334: Add twofactortype column to userinfo table
+ALTER TABLE userinfo
+    ADD COLUMN IF NOT EXISTS twofactortype character varying(50);
+
+-- Set all existing users to have the default twofactortype TOTP, if the secret column is not empty.
+UPDATE userinfo
+SET twofactortype = 'TOTP'
```
david-mackessy, line 7: does this value need to match the updated enum value?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/twofa/TwoFactorAuthService.java
```diff
@@ -0,0 +1,316 @@
...
+  private final I18nManager i18nManager;
+
+  /**
+   * Verify the 2FA code for the user.
+   *
+   * @param user The user
+   * @param code The 2FA code
+   * @return true if the code is invalid, false if the code is valid.
+   */
+  public boolean isInvalid2FACode(@Nonnull User user, @Nonnull String code)
```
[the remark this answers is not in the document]
netroms: Changed to private

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
netroms: They are moved now.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/twofa/TwoFactorAuthService.java
netroms: Removed method

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_29__add_2FA_type_userinfo.sql
netroms: Thanks, updated.

jbee (approved):

PR #19013, opened 2024-11-01
chore: Add ability to count SQL queries in tests

david-mackessy:
Adds an optional data source proxy for use in tests when we want to confirm the expected number of queries to the DB.

To use in a test:
- just add it on the test class: `@ContextConfiguration(classes = {QueryCountDataSourceProxy.class})`
- reset the counter before the condition being checked: `SQLStatementCountValidator.reset();`
- use one of the asserts to confirm expected count: `assertDeleteCount(1);`

When used, it just enhances the original data source by adding a query counter listener.


jbee (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/merge/category/CategoryOptionMergeProcessorTest.java
```diff
@@ -239,6 +245,29 @@ void catOptComboRefsReplacedSourcesNotDeletedTest() throws ConflictException {
...
+    // when
+    MergeParams mergeParams = getMergeParams();
+    mergeParams.setSources(Set.of(UID.of(coSource1A)));
+    mergeParams.setDeleteSources(true);
+
+    SQLStatementCountValidator.reset();
+    mergeProcessor.processMerge(mergeParams);
+
+    // then
+    assertDeleteCount(1);
```
vietnguyen, line 268: - Could we add another check here, such as `mergeReport.sourcesDeleted == 1` ? In case the library you just added doesn't work correctly in the future maybe cause of referenced libraries get upgraded.

vietnguyen (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/merge/category/CategoryOptionMergeProcessorTest.java
david-mackessy: sure, i'll add a call to get the object as well. The inclusion in the report wouldn't guarantee that it's deleted either.

PR #19015, opened 2024-11-01
fix: Cannot delete indicator if associated with section [DHIS2-18339]

larshelge:


netroms (approved):

mortenoh (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodService.java
```diff
@@ -282,6 +282,16 @@ public interface PeriodService {
    */
   PeriodType reloadPeriodType(PeriodType periodType);
 
+  /**
+   * Returns a PeriodType with the given {@link PeriodTypeEnum} value.
+   *
+   * @param periodType the {@link PeriodTypeEnum}.
+   * @return the PeriodType with the given name, or null if no match.
+   */
+  default PeriodType getPeriodType(PeriodTypeEnum periodType) {
```
jbee, line 291: @larshelge `default` methods cannot be used if the called method is `@Transactional` as this bypasses the proxy magic.

PR #19017, opened 2024-11-01
feat: Add skipAnalytics to ProgramTrackedEntityAttribute [DHIS2-16454]

zubaira:
This PR introduces a Flyway migration to add a new boolean column, `skipAnalytics`, to the `program_attributes` table, mapped to the `ProgramTrackedEntityAttribute` object. This migration is made idempotent, ensuring compatibility across environments where the column may already exist, preventing redundant attempts to add it.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramTrackedEntityAttribute.java
```diff
@@ -68,6 +68,8 @@ public class ProgramTrackedEntityAttribute extends BaseIdentifiableObject
 
   private Boolean searchable = false;
 
+  private boolean skipIndividualAnalytics;
```
larshelge, line 71: I am wondering, would it be better to just call this `skipAnalytics` ? 

I think I get the purpose of just skipping this very attribute in analytics export, vs. the entire program.

For `ProgramStageDataElement` we also have `skipAnalytics`, so this will make this consistent.

I general this is a very nice new feature/option which is great for avoiding exposure of sensitive data.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_23__Add_skipIndividualAnalytics_column_to_program_attributes.sql
```diff
@@ -0,0 +1,8 @@
+
+-- Step 1: Add the new column with a default value of false
+ALTER TABLE program_attributes
+    ADD COLUMN IF NOT EXISTS skipanalytics boolean DEFAULT FALSE;
+
+-- Step 2: Update existing rows to ensure they all have the value false
```
enricocolasante, line 6: Why is this step 2 necessary?
If there was no column before, than it just do nothing.
If there was a column before and some values were `true`, why are we changing them to `false`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramTrackedEntityAttribute.java
ameenhere: @larshelge  This flag does not behave the same way as `skipAnalytics` in `ProgramStageDataElement` 
The intention of this new flag `skipIndividualAnalytics` is to allow aggregation to happen for these attributes, but blocking these attributes from being fetched individually through analytics/query api (from Line Listing App, Maps App) and so on. The attribute values will still be consumed for aggregate analytics.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_23__Add_skipIndividualAnalytics_column_to_program_attributes.sql
zubaira: I added it as a safeguard statement, but after testing, I realized we don't actually need it.

ameenhere (approved):

ameenhere:
Approved it. But like Enrico mentioned, please remove the second step in the flyway script, and also sync with master 👍

enricocolasante (approved):

PR #19046, opened 2024-11-04
fix: non-standard object list queries [DHIS2-17200]

jbee:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -250,6 +251,14 @@ static UserDetails createUserDetails(
 
   void setId(Long id);
 
+  default boolean canIssueUserRole(UserRole role, boolean canGrantOwnUserRole) {
```
jbee, line 254: FYI: An ID based version of the existing `canIssueUserRole` in `User`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/DefaultJpaQueryParser.java
```diff
@@ -137,43 +149,43 @@ private Restriction getRestriction(
...
       case "null" -> Restrictions.isNull(path);
       case "!null" -> Restrictions.isNotNull(path);
       case "empty" -> Restrictions.isEmpty(path);
       default -> throw new QueryParserException("`" + operator + "` is not a valid operator.");
     };
   }
 
+  @Nonnull
+  @SuppressWarnings("rawtypes")
+  private Collection parseValues(Property property, Class<?> valueType, Object arg) {
```
jbee, line 163: FYI: duplication extracted to this method

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/DefaultJpaQueryParser.java
```diff
@@ -137,43 +149,43 @@ private Restriction getRestriction(
...
+            ? parseValue(Collection.class, property.getItemKlass(), arg)
+            : parseValue(Collection.class, valueType, arg);
+
+    if (values == null || values.isEmpty()) {
+      throw new QueryParserException("Invalid argument `" + arg + "` for in operator.");
+    }
+    return values;
+  }
+
+  private Collection<Disjunction> getDisjunctionsFromCustomMentions(
```
jbee, line 175: FYI: logic extracted from outside of the engine to be inside as it operates on the 3 parts on a filter

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/JpaCriteriaQueryEngine.java
```diff
@@ -108,68 +111,11 @@ public List<T> query(Query query) {
     CriteriaQuery<T> criteriaQuery = builder.createQuery(klass);
     Root<T> root = criteriaQuery.from(klass);
 
-    if (query.isEmpty()) {
```
jbee, line 111: FYI: This path was eliminated as it should be equal to the "else" path except for the starting point being an empty conjunction.

Note that this path did not add orders because `isEmpty` is the absence of filters and orders. In the refactored code the effect is the same but the orders are always added when non-empty.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/JpaCriteriaQueryEngine.java
```diff
@@ -215,36 +183,26 @@ public long count(Query query) {
...
     criteriaQuery.where(predicate);
 
-    if (!query.getOrders().isEmpty()) {
-      criteriaQuery.orderBy(
-          query.getOrders().stream()
-              .map(
-                  o ->
-                      o.isAscending()
-                          ? builder.asc(root.get(o.getProperty().getName()))
-                          : builder.desc(root.get(o.getProperty().getName())))
```
jbee, line 239: FYI: I think using `getName()` instead of `getFieldName()` was a bug - new code uses a common extracted method for fetch and count query

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -152,97 +149,81 @@ public class UserController extends AbstractCrudController<User> {
...
-    }
-
-    if (options.isManage()) {
-      params.setCanManage(true);
-      params.setAuthSubset(true);
+      queryParams.addOrganisationUnit(organisationUnitService.getOrganisationUnit(ou));
     }
 
-    boolean hasUserGroupFilter = filters.stream().anyMatch(f -> f.startsWith("userGroups."));
-    params.setPrefetchUserGroups(hasUserGroupFilter);
```
jbee, line 178: FYI: This is no longer needed because the query will only result in a list of UIDs

david-mackessy (approved):

jbee:
Thir PR was merged but somehow it does not show as merged, therefore I am closing it now

PR #19058, opened 2024-11-05
fix: broken run-api.sh environment variable check

netroms:
### Summary
Fixes a check for the `DHIS2_HOME` environment variable in the bash script `run_api.sh`.
The script was using `-v` when it should have been using `-n`, `-v` means to check if a variable is set, but can not be used with the `[[` syntax, `-n` should be used instead, this checks if a variable is not empty.

### DIFF
--- a/dhis-2/run-api.sh
+++ b/dhis-2/run-api.sh
@@ -21,7 +21,7 @@ DHIS2_PORT=9090
 SKIP_COMPILE=0
 
 # Set DHIS2 home directory to DHIS2_HOME_DIR env variable if set
-if [[ -v "$DHIS2_HOME" ]]; then
+if [[ -n "$DHIS2_HOME" ]]; then
   DHIS2_HOME_DIR=$DHIS2_HOME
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):

dhis-2/run-api.sh
```diff
@@ -21,7 +21,7 @@ DHIS2_PORT=9090
 SKIP_COMPILE=0
 
 # Set DHIS2 home directory to DHIS2_HOME_DIR env variable if set
-if [[ -v "$DHIS2_HOME" ]]; then
+if [[ -n "$DHIS2_HOME" ]]; then
   DHIS2_HOME_DIR=$DHIS2_HOME
```
teleivo, line 25: Setting defaults like

DHIS2_HOME="${DHIS2_HOME:-/opt/dhis2}"

is more succinct

teleivo (approved):

mortenoh (approved):

PR #19062, opened 2024-11-05
chore: Use UID for tracked entity in tracker importer [DHIS2-17790]

enricocolasante:
***Big Picture***
At the moment, tracker objects uids are represented as strings because the are validated in the importer and they can represent an invalid UID.
We are now changing this and tracker objects are going to use UID object instead so the validation of UIDs is happening when binding the body object in the API.
This allows the importer to not worry anymore on the UIDs for tracker objects as they now are valid and not null (this is guaranteed by the `BodyConverter` that is initializing all the UIDs).

***What is in this PR***
This PR is changing the importer domain for `TrackedEntity`.
I am also removing 2 out of 3 TODOs introduced in the `Event` PR as the logic there was correct, still over-complicated and confusing, but correct.

***What is next***
- Handle TODO left in `Event` PR
- Change importer model for `Relationship`
- Change importer model for `Note`
- Change API view model


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EnrollmentPersister.java
```diff
@@ -73,7 +73,7 @@ protected void updateAttributes(
         entityManager,
         preheat,
         enrollment.getAttributes(),
-        preheat.getTrackedEntity(enrollmentToPersist.getTrackedEntity().getUid()),
+        preheat.getTrackedEntity(UID.of(enrollmentToPersist.getTrackedEntity())),
```
teleivo, line 76: I am not counting so don't know but it feels like we have lots of changes where I know need todo `UID.of`. This sometimes makes sense if we don't want to deal with errors of invalid `UID`s at this point. If that does not matter we should make it easier to pass around our types and not have to do `UID.of` so often. This is not only related to the `preheat` but any API.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/TrackerPreheat.java
```diff
@@ -537,19 +536,20 @@ public void putEnrollmentsWithoutRegistration(String programUid, Enrollment enro
...
   }
 
   private void addProgramOwner(
-      String teUid, String programUid, TrackedEntityProgramOwnerOrgUnit tepo) {
+      UID teUid, String programUid, TrackedEntityProgramOwnerOrgUnit tepo) {
     programOwner.computeIfAbsent(teUid, k -> new HashMap<>()).put(programUid, tepo);
   }
 
-  public void addProgramOwner(String teUid, String programUid, OrganisationUnit orgUnit) {
+  public void addProgramOwner(UID teUid, String programUid, OrganisationUnit orgUnit) {
```
teleivo, line 548:
```suggestion
  public void addProgramOwner(UID te, String program, OrganisationUnit orgUnit) {
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/AttributeValidator.java
```diff
@@ -196,7 +195,7 @@ private void validateMandatoryAttributes(
                     enrollment.getEnrollment()));
   }
 
-  private MetadataIdentifier getOrgUnitUidFromTei(TrackerBundle bundle, String teUid) {
+  private MetadataIdentifier getOrgUnitUidFromTei(TrackerBundle bundle, UID teUid) {
```
teleivo, line 198:
```suggestion
  private MetadataIdentifier getOrgUnitUidFromTei(TrackerBundle bundle, UID te) {
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/ReportSummaryIntegrationTest.java
```diff
@@ -93,28 +91,6 @@ void testStatsCountForOneCreatedAndOneUpdatedTE() throws IOException {
     assertEquals(0, trackerImportTeReport.getStats().getDeleted());
   }
 
-  @Test
-  void testStatsCountForOneCreatedAndOneUpdatedTEAndOneInvalidTE() throws IOException {
```
teleivo, line 97: Why is that not needed anymore?

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/ReportSummaryIntegrationTest.java
enricocolasante: We cannot have an invalid UID at this point anymore.
We will have some web tests to check this when we will change the view model

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/persister/EnrollmentPersister.java
enricocolasante: I will check this kind of stuff at the end of the changes.
We will need to review the SMS part as well.
One day, hopefully, everything in DHIS2 will be UID and we will not have this problem anymore :grimacing:

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/bundle/ReportSummaryIntegrationTest.java
teleivo: This test was not about an invalid UID. It just used an invalid UID to test the report. So this should not be removed but the TE made invalid in some other way.

enricocolasante: You are right :sweat_smile: 
Putting it back also for other entities

PR #19065, opened 2024-11-05
Trigger external apps test

adeldhis2:


teleivo (commented):
What is the purpose of this? Can you please add a PR description/jira link.

We are moving away from Jenkins for dhis2-core. So I don't think we should add any new Jenkins pipelines.

PR #19077, opened 2024-11-06
chore: Use UID for relationship in tracker importer [DHIS2-17790]

enricocolasante:
***Big Picture***
At the moment, tracker objects uids are represented as strings because the are validated in the importer and they can represent an invalid UID.
We are now changing this and tracker objects are going to use UID object instead so the validation of UIDs is happening when binding the body object in the API.
This allows the importer to not worry anymore on the UIDs for tracker objects as they now are valid and not null (this is guaranteed by the `BodyConverter` that is initializing all the UIDs).

***What is in this PR***
This PR is changing the importer domain for `Relationship`.

***What is next***
- Remove obsolete `getStringUid` method from `TrackerDto` interface
- Change importer model for `Note`
- Change API view model
- - Handle TODO left in `Event` PR


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/TrackerPreheat.java
```diff
@@ -483,13 +483,13 @@ public RelationshipType getRelationshipType(MetadataIdentifier id) {
     return get(RelationshipType.class, id);
   }
 
-  public Relationship getRelationship(String relationshipUid) {
+  public Relationship getRelationship(UID relationshipUid) {
```
teleivo, line 486:
```suggestion
  public Relationship getRelationship(UID relationship) {
```

teleivo (approved):

PR #19078, opened 2024-11-06
chore: Delete app hub controller tests as they call a live service

david-mackessy:
Delete 3 AppHubController tests.

These tests have failed sometimes in our CI build (when AppHub service is down).
They call the live App Hub service (https://apps.dhis2.org/api).
We shouldn't rely on an external service in our automated build tests.

These tests should probably be part of release testing. Stian has reached out to Lina about this to see if they could be included somewhere more appropriate.


jbee (approved):

larshelge:
Maybe just remove the tests entirely? Now I think they will just be disabled forever. Not a big deal.

larshelge (approved):

david-mackessy:
> Maybe just remove the tests entirely? Now I think they will just be disabled forever. Not a big deal.

Yep, I'll remove the class 👍

PR #19081, opened 2024-11-06
chore: Use UID in view model in tracker importer [DHIS2-17790]

enricocolasante:


dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/config/UIDStdSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class UIDStdSerializer extends StdSerializer<UID> {
+  public UIDStdSerializer() {
+    super(UID.class);
+  }
+
+  @Override
+  public void serialize(UID uid, JsonGenerator gen, SerializerProvider provider)
+      throws IOException {
+    gen.writeString(uid == null ? "" : uid.toString());
```
teleivo, line 44: Is that what we do for other types?
Why not turn `null` into JSON `null`?

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/events/EventsTests.java
```diff
@@ -136,6 +137,57 @@ void eventsImportNewEventsFromFile(String fileName, String contentType) throws E
...
+            .replacePropertyValuesWith("event", "invalid_uid")
+            .get();
+
+    ApiResponse response =
+        trackerImportExportActions.post(
+            "",
+            ContentType.JSON.toString(),
+            obj,
+            new QueryParamsBuilder()
+                .addAll("dryRun=false", "eventIdScheme=UID", "orgUnitIdScheme=UID"));
```
teleivo, line 154: IdSchemes are only supported for metadata. `eventIdScheme` does not exist. Is it even necessary to add these params? UID is the default

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportControllerTest.java
```diff
@@ -61,6 +62,22 @@ void shouldSucceedWhenAllValidParametersArePassed() {
             .content(HttpStatus.OK));
   }
 
+  @Test
+  void shouldReturnBadRequestWhenThereIsAnInvalidUidInThePayload() {
```
teleivo, line 66: Why the split of tests into e2e and web-api? Only using one would be less surprising unless there is a good reason. I prefer the web-api ones.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportControllerTest.java
```diff
@@ -61,6 +62,22 @@ void shouldSucceedWhenAllValidParametersArePassed() {
...
+        "Bad Request",
+        400,
+        "ERROR",
+        "JSON parse error: Cannot construct instance of `org.hisp.dhis.common.UID`, problem: UID must be an alphanumeric string of 11 characters starting with a letter.",
+        POST(
+                "/tracker?async=false&reportMode=FULL"
+                    + "&validationMode=SKIP"
+                    + "&atomicMode=OBJECT"
+                    + "&skipRuleEngine=true",
+                Path.of("tracker/import_invalid_uid_payload.json"))
```
teleivo, line 77: Now that multi-line strings are supported favor that. This reduces the scope of the test setup. We have so many files 😅 Hard to know what's not used anymore, needed by what, ...

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/CrudControllerAdvice.java
```diff
@@ -513,6 +514,18 @@ public WebMessage illegalArgumentExceptionHandler(IllegalArgumentException ex) {
     return badRequest(ex.getMessage());
   }
 
+  /**
+   * Handles {@link RuntimeJsonMappingException} and logs the stack trace to standard error. {@link
```
teleivo, line 518:
```suggestion
   * Handles {@link RuntimeJsonMappingException} and logs the stack trace. {@link
```

where its logged depends on the effective log config.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/imports/BodyConverter.java
```diff
@@ -246,8 +246,8 @@ private List<Enrollment> extractEnrollments(TrackedEntity trackedEntity) {
    * @param uid the uid to check and update
    * @return a valid uid
    */
-  private String updateReference(String uid) {
-    return StringUtils.isEmpty(uid) ? CodeGenerator.generateUid() : uid;
+  private UID updateReference(UID uid) {
```
teleivo, line 249: Can we then add `@Nonnull` on our domain type fields?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/UIDMapper.java
```diff
@@ -30,10 +30,13 @@
...
-@Deprecated(forRemoval = true)
+/**
+ * Converts a string representation of a UID to a {@link UID} object. If the provided string is
+ * {@code null}, the method returns {@code null}. Otherwise, it creates a new {@link UID} instance
+ * using {@code UID.of}.
+ */
 @Mapper
 public interface UIDMapper {
-
   default UID fromString(String stringUid) {
```
teleivo, line 40:
```suggestion
  default UID fromString(String uid) {
```

add the nonnull/nullable annotations as this ones public

teleivo (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerImportControllerTest.java
enricocolasante: Yeah, this split was weird.
I will remove the json test here and just have the web-api one but I will leave the csv as an e2e, we don't have, at least now, an easy way to test that in the web-api.

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/config/UIDStdSerializer.java
enricocolasante: Other types are not consistent, of course :grimacing: 
The serializer used in the CSVMapper are serializing to`null`, I am going to do the same.
What do you think @david-mackessy?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/CsvEventService.java
```diff
@@ -202,6 +199,7 @@ public List<Event> read(InputStream inputStream, boolean skipFirst)
           dataValue.getProvidedElsewhere(),
           dataValue.getDataElement(),
           dataValue.getValue(),
+          dataValue.getGeometry(),
```
ameenhere, line 202: Was this intentional?
I don't see this used inside the if construct unlike the remaining ones.

ameenhere (approved):

dhis-2/dhis-support/dhis-support-commons/src/main/java/org/hisp/dhis/commons/jackson/config/UIDStdSerializer.java
enricocolasante: That is true when we are using the UID as the identifier of the object but there are places where we are using UIDs for fields that are nullable for example
`@JsonProperty
  @OpenApi.Property({UID.class, TrackedEntity.class})
  private UID trackedEntity;`
   in Event it can be null if it is an event program.

PR #19091, opened 2024-11-07
perf: improves CSV exports when too many dimensions combinations exists [DHIS2-18377][2.40]

gnespolino:
This PR improves the CSV Export when the combinations of dimensions explode into huge numbers.

Before this change, when exporting a CSV, all possible rows would be generated, just to check if they had a value in the query result. While the result is correct, the approach seemed to not consider performance implications when the cartesian product of all dimensionalItems is big.

This PR improves it by generating only the rows that exist in the query result.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,96 @@ private Grid getAggregatedDataValuesTableLayout(
...
+   * used to render the grid.
+   *
+   * @param grid the grid
+   * @param columnsDimensionItemsByDimension a map of dimension items by dimension for columns
+   * @param rowsDimensionItemsByDimension a map of dimension items by dimension for rows
+   * @param columnDimensionIds the column dimension ids
+   * @param rowDimensionIds the row dimension ids
+   * @return the alternative grid items
+   */
+  public static AlternativeGridItems getAlternateGridItems(
```
maikelarabori, line 314: This method can be simplified to work only with a Map<String, List<DimensionalItemObject>> and a List List<String>.

So, you can call it `getDimensionItemsFromMap`, or something similar.

Then the consumers will do:

`Map<String, List<DimensionalItemObject>> columnsDimensionItemsByDimension = getDimensionItemsFromMap(...);`

`Map<String, List<DimensionalItemObject>> rowsDimensionItemsByDimension = getDimensionItemsFromMap(...);`

And we eliminate the class `AlternativeGridItems`

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,96 @@ private Grid getAggregatedDataValuesTableLayout(
 
     Map<String, Object> valueMap = AnalyticsUtils.getAggregatedDataValueMapping(grid);
 
+    AlternativeGridItems alternativeGridItems =
```
maikelarabori, line 274: See my comment below about removing the need for `AlternativeGridItems`.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
maikelarabori, line 314: Please add unit tests that reach all different conditions used in this method.

maikelarabori (commented):
Thanks for the great perf. improvement @gnespolino.
I left some comments, trying to simplify the code, and make it easier to follow (I hope).

Also, do not forget to add tests, please.

d-bernat (approved):

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,96 @@ private Grid getAggregatedDataValuesTableLayout(
...
+   * @return the alternative grid items
+   */
+  public static AlternativeGridItems getAlternateGridItems(
+      Grid grid,
+      Map<String, List<DimensionalItemObject>> columnsDimensionItemsByDimension,
+      Map<String, List<DimensionalItemObject>> rowsDimensionItemsByDimension,
+      List<String> columnDimensionIds,
+      List<String> rowDimensionIds) {
+
+    Set<List<DimensionalItemObject>> alternateRows = new HashSet<>();
```
luciano-fiandesio, line 321: Since this is a performance related improvement, would it make sense to set an initial capacity for these two HashSet(s)?

gnespolino: in this case, it shouldn't change much, I used a set just to avoid duplicates so the initial capacity wouldn't be determined in advance

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,96 @@ private Grid getAggregatedDataValuesTableLayout(
...
+        new ArrayList<>(alternateRows), new ArrayList<>(alternateColumns));
+  }
+
+  private static DimensionalItemObject findItem(
+      Map<String, List<DimensionalItemObject>> rowsDimensionItemsByDimension, String value) {
+    return rowsDimensionItemsByDimension.values().stream()
+        .flatMap(List::stream)
+        .filter(dio -> dio.getDimensionItem().equals(value))
+        .findFirst()
+        .orElse(null);
```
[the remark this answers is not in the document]
gnespolino: yeah good one, I've been looking at it as well. I think `Optional` will be better in the end

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,79 @@ private Grid getAggregatedDataValuesTableLayout(
...
+          alternateItem[indexInColumn] = findValueInDimensionItem(dimensionItemsByDimension, value);
+        }
+      }
+      alternateItems.add(Arrays.stream(alternateItem).collect(Collectors.toList()));
+    }
+    return new ArrayList<>(alternateItems);
+  }
+
+  private static DimensionalItemObject findValueInDimensionItem(
+      Map<String, List<DimensionalItemObject>> rowsDimensionItemsByDimension, String value) {
```
maikelarabori, line 337:
```suggestion
      Map<String, List<DimensionalItemObject>> dimensionalItems, String value) {
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,79 @@ private Grid getAggregatedDataValuesTableLayout(
...
+    return rowsDimensionItemsByDimension.values().stream()
+        .flatMap(List::stream)
+        .filter(dimensionalItemObject -> dimensionalItemObject.getDimensionItem().equals(value))
+        .findFirst()
+        .orElseThrow(() -> new IllegalQueryException("Dimensional item not found"));
```
maikelarabori, line 342: We have to define and use an `ErrorCode` for this exception.

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/DefaultAnalyticsServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.hisp.dhis.common.DimensionalItemObject;
+import org.hisp.dhis.common.Grid;
+import org.hisp.dhis.common.GridHeader;
+import org.junit.jupiter.api.Test;
+
+class DefaultAnalyticsServiceTest {
+
+  @Test
+  void testGetGridItems() {
+
```
maikelarabori, line 48: Blank line

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/DefaultAnalyticsServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+    Grid grid = mock(Grid.class);
```
maikelarabori, line 49: We can create a ListGrid object and set all the needed data.
In this case, we should not use a mock, especially because there are no dependencies involved...

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/DefaultAnalyticsServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+    List<List<DimensionalItemObject>> gridItems =
+        DefaultAnalyticsService.getGridItems(grid, dimensionItemsByDimension, dimensionIds);
+
+    assertEquals(grid.getRows().size(), gridItems.size());
+
+    for (int i = 0; i < grid.getRows().size(); i++) {
+      assertEquals(grid.getWidth() - 1, gridItems.get(i).size());
+    }
+
+    Set<String> fromGrid =
```
maikelarabori, line 106: Could we simplify it? This is the kind of code we would like to avoid, as discussed days ago...

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
```diff
@@ -255,10 +271,79 @@ private Grid getAggregatedDataValuesTableLayout(
...
+   * @param grid the grid
+   * @param dimensionItemsByDimension a map of dimension items by dimension
+   * @param dimensionIds the dimension ids
+   * @return the alternative grid items
+   */
+  static List<List<DimensionalItemObject>> getGridItems(
+      Grid grid,
+      Map<String, List<DimensionalItemObject>> dimensionItemsByDimension,
+      List<String> dimensionIds) {
+
```
maikelarabori, line 313: Blank line

dhis-2/dhis-services/dhis-service-analytics/src/test/java/org/hisp/dhis/analytics/data/DefaultAnalyticsServiceTest.java
gnespolino: The grid is indeed a "dependency" of the tested method. However I'll apply your suggestion

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DefaultAnalyticsService.java
gnespolino: I think sometimes a blank line makes the code more readable

maikelarabori (approved):

PR #19094, opened 2024-11-07
feat: Merge both test metadata json files [DHIS2-14968]

muilpp:
Following up on [this slack tread](https://dhis2.slack.com/archives/C05S1NTHBJ9/p1730884308213749), I'm merging both test metadata json files so we reuse the same one in all our integration test.

While at it, I've cleaned it up a bit. I've removed all:
- empty arrays
- json fields set to false
- empty json objects
- `publicAccess` fields, as sharing doesn't use that one anymore


ameenhere (approved):
LGTM!

My only concern is, is it worth keeping all the "false" and "[]" properties for brevity?

enricocolasante (approved):

muilpp:
> LGTM!
> 
> My only concern is, is it worth keeping all the "false" and "[]" properties for brevity?

In my opinion that's just noise, but maybe @enricocolasante or @teleivo have something to say about it

PR #19101, opened 2024-11-08
Prevent import of duplicate default COCs

jason-p-pickering:
This PR attempts to solve two problems when importing  an alternative default Category option combos through the metadata API. This PR should prevent the import of a duplicated, non-standard default category option combination. 

Some time ago, this was hard-coded into the system, but it is still possible to import another default COC with a different UID. This creates all sorts of analysis and configuration issues. A separate integrity check has also been created to identify this situation in existing systems. Various tests which utilized the alternative default COCs needed to be updated. 

Some tests needed to be updated as part of this PR. Some integration and E2E tests were utilizing an alternative default COC, and these have now been updated to use the standard default COC instead.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -66,6 +66,8 @@ public enum ErrorCode {
   E1119("{0} already exists: `{1}`"),
   E1120("Update cannot be applied as it would make existing data values inaccessible"),
 
+  E1122("Category option combo {0} already exists for category combo {1}"),
```
david-mackessy, line 69: i updated this as I have another PR using `1121`. To avoid conflict.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
```diff
@@ -0,0 +1,76 @@
...
+  private final CategoryService categoryService;
+
+  private void checkDuplicateCategoryOptionCombos(
+      CategoryOptionCombo categoryOptionCombo, Consumer<ErrorReport> addReports) {
+
+    CategoryCombo categoryCombo = categoryOptionCombo.getCategoryCombo();
+
+    List<CategoryOptionCombo> categoryOptionCombos =
+        categoryService.getAllCategoryOptionCombos().stream()
+            .filter(coc -> coc.getCategoryCombo().equals(categoryCombo))
```
david-mackessy, line 54: I think the issue might be here. After filtering, `categoryOptionCombos` is empty (no match) because calling equals on `categoryCombo` uses `BaseIdentifiableObject#equals` because `categoryCombo` doesn't have it's own `equals` method implemented.
You'll have to get that working as you want so there is something to process on in the next step below.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EventImportValidationTest.java
jason-p-pickering, line 1: @enricocolasante I cannot get this test to return the 1099 error code. Maybe we can touch base on this.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    List<CategoryOptionCombo> categoryOptionCombos =
+        categoryService.getAllCategoryOptionCombos().stream()
+            .filter(
+                coc ->
+                    coc.getCategoryCombo()
+                        .getUid()
+                        .equals(categoryOptionCombo.getCategoryCombo().getUid()))
+            .toList();
```
jbee, line 93: This looks like it should use a dedicated method only loading COC with the tagret CC ID instead to avoid loading a big number of DB objects into memory. 

Also I wonder if this logic is correct. Don't you want to check if any other CC has the same categories but is not the same CC?

jbee:
As discussed the basic idea of this PR is not wrong but to really validate duplicates correctly we need to consider all effects of the import context as a whole which is difficult to do in the bundle hooks. This needs some more thinking about where we could have such a overall consistency validation.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
jason-p-pickering: Resolving this comment as this method has been removed from the PR

jbee (approved):

david-mackessy (approved):

PR #19106, opened 2024-11-08
fix: Prohibit changing DataElement valueType when it has data values [DHIS2-18152]

david-mackessy:
# Issue
Issue reported as an analytics failure, after changing the `valueType` of a `DataElement`. e.g. `NUMBER` -> `TEXT`.

# Cause
Analytics can't process correctly as it expects a specific `valueType` when processing the data

# Fix (API feature really)
From the `DataElement` API perspective, this is a new feature. Checking if changing a `DataElement`s `valueType` should be allowed or not.
Bundle hook check added for `DataElement`:
When validating a `DataElement` check if the `valueType` is being changed:
  - if the `valueType` is being changed, check if `DataElement` is associated with any `DataValue`
    - if there are `DataValue`s associated with `DataElement`, prohibit the change
    - if there are no `DataValue`s associated with `DataElement`, allow the change
  - if the `valueType` is not being changed, no action taken

# Testing
## Automated
Add 3 web API tests
- prohibit `valueType` change through `/dataElements` API
- prohibit `valueType` change through `/metadata` API
- allow `valueType` change when no `DataValue`s exist for `DataElement`

# Manual
tested on IM using existing metadata


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DataElementControllerTest.java
```diff
@@ -133,4 +139,309 @@ void dataElementsExcludingDefaultCatComboTest() {
...
+
+    // try update data element with new value type (TEXT)
+    JsonWebMessage validationErrorMsg =
+        assertWebMessage(
+            "Conflict",
+            409,
+            "ERROR",
+            "One or more errors occurred, please see full details in import report.",
+            PUT(
+                    "/dataElements/DeUid000001",
```
david-mackessy, line 236: added a test for both the `/dataElements` API and the `/metadata` API to cover different ways we allow updates and also in case the implementation of one diverges from the other.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/datavalue/DataValueStore.java
```diff
@@ -184,4 +185,12 @@ DataValue getDataValue(
...
    */
   boolean dataValueExists(CategoryCombo combo);
+
+  /**
+   * Checks if any data values exist for the provided {@link DataElement} {@link UID}.
+   *
+   * @param dataElementUid the {@link DataElement} {@link UID} to check
+   * @return true, if any values exist, otherwise false
+   */
+  boolean dataValueExistsForDataElement(String dataElementUid);
```
larshelge, line 195: Maybe we can just call the variable `uid` in these cases, it is implicit from the method name, and it makes the code a bit more elegant?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/hibernate/HibernateDataValueStore.java
```diff
@@ -267,6 +267,15 @@ public boolean dataValueExists(CategoryCombo combo) {
         .isEmpty();
   }
 
+  @Override
+  public boolean dataValueExistsForDataElement(String uid) {
+    return !getQuery("select 1 from DataValue dv where dv.dataElement.uid = :uid")
```
jbee, line 272: Not sure it is needed since we use `select 1` but adding `limit 1` would make sure this query terminates as soon as 1 match is found.

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/hibernate/HibernateDataValueStore.java
david-mackessy: `setMaxResults(1)` is the HQL equivalent, which is used.

vietnguyen (approved):

PR #19112, opened 2024-11-09
feat: Add not-null constraint for column trackedentitytypeid in trackedentity table [DHIS2-15066]

zubaira:
[DHIS2-15066](https://dhis2.atlassian.net/browse/DHIS2-15066)

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
@@ -1,0 +1,3 @@
+-- Add not null constraint to trackedentitytypeid column in trackedentity table
+ALTER TABLE trackedentity
+    ALTER COLUMN trackedentitytypeid SET NOT NULL;
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/trackedentity/hibernate/TrackedEntity.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/trackedentity/hibernate/TrackedEntity.hbm.xml
@@ -58,7 +58,7 @@
                      foreign-key="fk_trackedentityinstance_organisationunitid" not-null="true" lazy="false"/>
 
         <many-to-one name="trackedEntityType" class="org.hisp.dhis.trackedentity.TrackedEntityType"
-                     column="trackedentitytypeid" foreign-key="fk_trackedentityinstance_trackedentitytypeid" lazy="false"/>
+                     column="trackedentitytypeid" foreign-key="fk_trackedentityinstance_trackedentitytypeid" lazy="false" not-null="true"/>
     </class>
 
     <sql-query name="updateTrackedEntitiesLastUpdated">update trackedentity set lastUpdated = :lastUpdated, lastupdatedbyuserinfo = CAST(:lastupdatedbyuserinfo as jsonb) WHERE uid in :trackedEntities</sql-query>


enricocolasante (commented):
We are having discussion about the approach here (that is the same for organisationunitid column in event and enrollment) and we are trying to avoid to create dummy data that are likely entering the DB and never being deleted.

I think we should fix this the same way in all the places where we found it.
As soon as we have a strategy we should go back to this PR and see what we should do.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelperTest.java
```diff
@@ -521,14 +521,14 @@ private List<Enrollment> getEnrollments() {
   }
 
   private TrackedEntity getTrackedEntityA() {
-    TrackedEntity te = createTrackedEntity(organisationUnitA);
+    TrackedEntity te = createTrackedEntity(organisationUnitA, createTrackedEntityType('I'));
     te.setTrackedEntityType(trackedEntityTypeA);
```
enricocolasante, line 525: setTrackedEntityType can be removed

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/deduplication/DeduplicationHelperTest.java
```diff
@@ -521,14 +521,14 @@ private List<Enrollment> getEnrollments() {
...
 
     return te;
   }
 
   private TrackedEntity getTrackedEntityB() {
-    TrackedEntity te = createTrackedEntity(organisationUnitB);
+    TrackedEntity te = createTrackedEntity(organisationUnitB, createTrackedEntityType('F'));
     te.setTrackedEntityType(trackedEntityTypeB);
```
enricocolasante, line 532: setTrackedEntityType can be removed

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerAccessManagerTest.java
```diff
@@ -155,12 +155,11 @@ void setUp() {
...
-    TrackedEntity femaleB = createTrackedEntity(orgUnitB);
-    trackedEntityA.setTrackedEntityType(trackedEntityType);
-    trackedEntityB.setTrackedEntityType(trackedEntityType);
+    trackedEntityA = createTrackedEntity(orgUnitA, trackedEntityType);
+    TrackedEntity trackedEntityB = createTrackedEntity(orgUnitB, trackedEntityType);
+    TrackedEntity femaleA = createTrackedEntity(orgUnitA, trackedEntityType);
+    TrackedEntity femaleB = createTrackedEntity(orgUnitB, trackedEntityType);
+
     femaleA.setTrackedEntityType(trackedEntityType);
     femaleB.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 164: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidatorTest.java
```diff
@@ -320,7 +322,8 @@ private TrackedEntity teWithDeleteEnrollments() {
   }
 
   private TrackedEntity teWithEnrollments() {
-    TrackedEntity trackedEntity = createTrackedEntity(organisationUnit);
+    TrackedEntity trackedEntity =
+        createTrackedEntity(organisationUnit, createTrackedEntityType('R'));
     trackedEntity.setUid(TE_ID.getValue());
     trackedEntity.setEnrollments(Sets.newHashSet(new Enrollment()));
     trackedEntity.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 329: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidatorTest.java
```diff
@@ -311,7 +312,8 @@ private TrackedEntity teWithDeleteEnrollments() {
     Enrollment enrollment = new Enrollment();
     enrollment.setDeleted(true);
 
-    TrackedEntity trackedEntity = createTrackedEntity(organisationUnit);
+    TrackedEntity trackedEntity =
+        createTrackedEntity(organisationUnit, createTrackedEntityType('B'));
     trackedEntity.setUid(TE_ID.getValue());
     trackedEntity.setEnrollments(Sets.newHashSet(enrollment));
     trackedEntity.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 319: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidatorTest.java
```diff
@@ -299,7 +299,8 @@ void shouldFailWhenUpdateTEAndUserHasNoWriteAccess() {
   }
 
   private TrackedEntity teWithNoEnrollments() {
-    TrackedEntity trackedEntity = createTrackedEntity(organisationUnit);
+    TrackedEntity trackedEntity =
+        createTrackedEntity(organisationUnit, createTrackedEntityType('Z'));
     trackedEntity.setUid(TE_ID.getValue());
     trackedEntity.setEnrollments(Sets.newHashSet());
     trackedEntity.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 306: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityOwnershipValidatorTest.java
```diff
@@ -486,7 +486,8 @@ void verifySuccessEventValidationWhenEventHasNoOrgUnitAssigned() {
   }
 
   private TrackedEntity teWithNoEnrollments() {
-    TrackedEntity trackedEntity = createTrackedEntity(organisationUnit);
+    TrackedEntity trackedEntity =
+        createTrackedEntity(organisationUnit, createTrackedEntityType('E'));
     trackedEntity.setUid(TE_ID.getValue());
     trackedEntity.setEnrollments(Sets.newHashSet());
     trackedEntity.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 493: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidatorTest.java
```diff
@@ -389,7 +389,8 @@ void verifyValidationFailsForEnrollmentDeletionAndUserWithoutTrackedEntityTypeRe
   }
 
   private TrackedEntity teWithNoEnrollments() {
-    TrackedEntity trackedEntity = createTrackedEntity(organisationUnit);
+    TrackedEntity trackedEntity =
+        createTrackedEntity(organisationUnit, createTrackedEntityType('C'));
     trackedEntity.setUid(TE_ID.getValue());
     trackedEntity.setEnrollments(Sets.newHashSet());
     trackedEntity.setTrackedEntityType(trackedEntityType);
```
enricocolasante, line 396: setTrackedEntityType

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/DataRelationsValidatorTest.java
```diff
@@ -263,7 +263,7 @@ private TrackedEntityType trackedEntityType(String uid, char uniqueChar) {
   }
 
   private TrackedEntity trackedEntity(UID uid, TrackedEntityType type, OrganisationUnit orgUnit) {
-    TrackedEntity te = createTrackedEntity(orgUnit);
+    TrackedEntity te = createTrackedEntity(orgUnit, createTrackedEntityType('R'));
     te.setUid(uid.getValue());
     te.setTrackedEntityType(type);
```
enricocolasante, line 268: setTrackedEntityType

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
```diff
@@ -0,0 +1,39 @@
...
+    RAISE NOTICE 'Inconsistencies found: %', inconsistent_records_count;
+
+    FOR results IN
+    SELECT programid, trackedentityid FROM enrollment WHERE trackedentityid IN (
+        SELECT trackedentityid
+        FROM trackedentity
+        WHERE trackedentitytypeid IS NULL
+    )
+    LOOP
+    UPDATE trackedentity
```
enricocolasante, line 18: Here we shouldn't update any record.
We should only apply the constraint.
The we can suggest such a script in the migration notes

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
```diff
@@ -0,0 +1,39 @@
...
+    )
+    WHERE trackedentityid = results.trackedentityid;
+
+    RAISE NOTICE 'Updated trackedentity: programid = %, trackedentityid = %', results.programid, results.trackedentityid;
+    END LOOP;
+    SELECT COUNT(*) INTO inconsistent_records_count FROM trackedentity WHERE trackedentitytypeid IS NULL;
+
+    IF inconsistent_records_count = 0 THEN
+        ALTER TABLE trackedentity ALTER COLUMN trackedentitytypeid SET NOT NULL;
+    ELSE
```
enricocolasante, line 32: This logic is a little bit redundant.
You can apply the constraint and if it fails then you raise the exception.

enricocolasante (changes requested):

ameenhere:
Before we merge this, can we make sure we update the dev db snapshot to have this "fixing scripts" run? So that we don't break dev again for Phil to fix later.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
zubaira: there is a lower possibility of errors if we handle the task ourselves rather than leaving it to the user. Therefore, we should aim to complete as much of the migration process as possible and leave as little as necessary for the user to manage.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/trackedentity/TrackerAccessManagerTest.java
zubaira: fixed

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
```diff
@@ -0,0 +1,40 @@
...
+inconsistent_records_count INTEGER;
+BEGIN
+
+SELECT COUNT(*) INTO inconsistent_records_count FROM trackedentity WHERE trackedentitytypeid IS NULL;
+
+IF inconsistent_records_count > 0 THEN
+        RAISE NOTICE 'Inconsistencies found: %', inconsistent_records_count;
+
+    FOR results IN
+    SELECT programid, trackedentityid
```
ameenhere, line 13: Suggested changes
1. You can fetch the trackedEntityTypeId directly in this query by joining with program table so that it is already present in results
2. Include another where clause so that you also check TrackedentityTypeid is not null (from the enrollment table)
3. In the update section, you can directly use results.trackedEntityTypeId

What do you think?

zubaira: used set based operation and removed looping logic.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
enricocolasante: For this kind of process that deal with data we need to be much more clear with the user and very very careful on any step.
Here you are fixing inconsistent data in a correct way, getting the tracked entity type from any underling enrollment but this is not the only issue that could happen and this SQL is running in a transaction (even without specifically start the transaction as done here, flyway will wrap this in a transaction), so if there is any inconsistent data that cannot be fixed in the way you provided, no data is fixed.
Than the user is sent to the guide and there we are not providing any explanation of what we attempted in the migration and we are not providing any option of fixing the data, only a script to delete tracked entitities and with all the enrollments and events.

To resume:
If the user has a number x of "fixable" tracked entities and one that is not automatically fixable, we are deleting ALL the tracked entities with inconsistent data.

enricocolasante (changes requested):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_26__Add_not_null_constraint_to_trackedentity.sql
```diff
@@ -0,0 +1,35 @@
...
+            SELECT 1
+            FROM enrollment
+                     JOIN program ON enrollment.programid = program.programid
+            WHERE enrollment.trackedentityid = trackedentity.trackedentityid
+        );
+
+    ALTER TABLE IF EXISTS trackedentity ALTER COLUMN trackedentitytypeid SET NOT NULL;
+ELSE
+        RAISE NOTICE 'No inconsistencies found, trackedentitytypeid is already populated.';
+        ALTER TABLE IF EXISTS trackedentity ALTER COLUMN trackedentitytypeid SET NOT NULL;
```
ameenhere, line 29: I did not understand why do we have two ALTER COLUMN SET NOT NULL on the same table-column combination?

zubaira: If the initial migration fails due to invalid data, it will raise an exception, prompting the user to correct or remove the problematic data. Once the issues are resolved, the same migration can be reapplied, and the ELSE condition will enforce the NOT NULL constraint on the column.

ameenhere: Then, can we move the ALTER TABLE to outside the IF ELSE condition, since it is executed at the end anyway?

ameenhere:
Any outstanding review comments for this? @enricocolasante  @zubaira  (Couldn't go through it completely)

enricocolasante:
> Any outstanding review comments for this? @enricocolasante @zubaira (Couldn't go through it completely)

I think here we are done or almost done, but there is more refining to do in the release notes.
And I think it makes sense to merge both at the same time to be sure we are in sync between the migration and the guide.

ameenhere (approved):

enricocolasante (commented):
Looks good to me.
Before merging we should test the migration on our play server to make sure we are not breaking anything in our environment.
@ameenhere si going to test is, right?
We should wait to merge this PR until the test is performed and eventual inconsistency are fixed in SL database

ameenhere:
Agreed. I'll try to use the release note doc to fix the play dbs first and update here. Not sure when I will get to it, but surely this week. And then we can merge this. 👍

enricocolasante (approved):

PR #19122, opened 2024-11-11
fix: NPE when delete eventVisualization

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-18330
### Issue
- Use got NPE when trying to delete an EventVisualization in Line Listing App.
- During metadata import preheat, Java Reflection is used to add properties to collections without null check.

### Fix
- Added null check and also unit test

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/preheat/DefaultPreheatService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/preheat/DefaultPreheatService.java
@@ -766,7 +766,13 @@ public class DefaultPreheatService implements PreheatService {
           }
 
           objects.forEach(
-              o -> list.addAll(ReflectionUtils.invokeMethod(o, property.getGetterMethod())));
+              o -> {
+                Collection<Object> propertyValue =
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/preheat/DefaultPreheatService.java
```diff
@@ -766,7 +766,13 @@ private void collectScanTargets(Map<Class<?>, List<?>> targets) {
           }
 
           objects.forEach(
-              o -> list.addAll(ReflectionUtils.invokeMethod(o, property.getGetterMethod())));
+              o -> {
+                Collection<Object> propertyValue =
+                    ReflectionUtils.invokeMethod(o, property.getGetterMethod());
+                if (!org.apache.commons.collections4.CollectionUtils.isEmpty(propertyValue)) {
```
david-mackessy, line 772: we have our own `CollectionUtils.isNotEmpty` method that does the same thing.
Always better to use our own instead of the dep.

david-mackessy (approved):

larshelge (approved):

PR #19143, opened 2024-11-13
refactor: Use period resource tables for analytics tables

larshelge:
Refactors `JdbcAnalyticsTableManager` to use resource tables instead of metadata entity tables. This to reduce the number of tables which have to be replicated in an analytics database.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/AnalyticsTableUpdateParams.java
```diff
@@ -166,6 +166,15 @@ public Date getFromDate() {
     return earliest;
   }
 
+  /**
+   * Indicates whether a from date exists.
+   *
+   * @return true if a from date exists.
+   */
+  public boolean hasFromDate() {
```
maikelarabori, line 174: It would sound better if we called it `hasInitialDate`, `hasStartDate`, or something like that.

maikelarabori (approved):
It looks good, just a small suggestion for review. Thx!

PR #19154, opened 2024-11-14
refactor: Further merge refactor

david-mackessy:
After working on a recent merge refactor, it became clear that it could be simplified even further.

There is potential to further simplify the merge code by removing the `MergeProcessor` which is the high-level conductor of a merge. It has 1 default method, `processMerge`. Within that, it calls each `MergeService` methods:
- `validate`
- `merge` (`@Transactional`)

This default method could be moved into the `MergeService` interface. It just means that the transactional `merge` method cannot be invoked by itself.
To get around this, the `@Transactional` annotation can be moved to the default `processMerge` method, which keeps everything working as expected.

Implications of proposed change:
- the transaction now includes validations + merge (previously just merge (handling of references))
- an abstraction layer is removed which:
  - should help simplify how merges work
  - removes code


jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/merge/MergeService.java
```diff
@@ -39,6 +40,27 @@
...
+    MergeReport mergeReport = new MergeReport();
+
+    MergeRequest mergeRequest = validate(mergeParams, mergeReport);
+    if (mergeReport.hasErrorMessages())
+      throw new ConflictException("Merge validation error").setMergeReport(mergeReport);
+
+    MergeReport report = merge(mergeRequest, mergeReport);
+    if (report.hasErrorMessages())
+      throw new ConflictException("Merge error").setMergeReport(mergeReport);
+
```
vietnguyen, line 60: There are two variables `mergeReport` and `report`. Seems there should only be one ?

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/merge/MergeService.java
david-mackessy: yes, essentially the same, but can get rid of one to keep it clearer.
Nice spot.

PR #19193, opened 2024-11-18
fix: Fix enrollment and event tables [DHIS2-12600]

enricocolasante:
In the DB `enrollment` and `event` tables were allowing null values for `organisationunitid` column even though the API is validating that organisation unit is not null for `Enrollment` and `Event` when importing.

The PR is changing the DB with a flyway migration to force `organisationunitid` column to be not null.

If the migration fails, it displays a message detailing how to detect and fix the issue.

The migration notes will be published together with the release notes, a PR can be found [here](https://github.com/dhis2/dhis2-releases/pull/195).

***Changes needed after having an organisation unit for placeholder enrollment***
- When creating the placeholder enrollment an organisation unit must be assigned to it. We are now assigned one of the root organisation unit to it.
- In the enrollment exporter we now need to filter out all placeholder enrollments. Before they were filtered out because organisation unit was null.


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/ProgramObjectBundleHook.java
```diff
@@ -133,6 +136,8 @@ private void addProgramInstance(Program program) {
       enrollment.setProgram(program);
       enrollment.setStatus(EnrollmentStatus.ACTIVE);
       enrollment.setStoredBy("system-process");
+      enrollment.setOrganisationUnit(
+          organisationUnitService.getRootOrganisationUnits().iterator().next());
```
ameenhere, line 140: This looks like there could be multiple root org units? I guess it still does not matter that we arbitrarily chose one of the root org unit.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
```diff
@@ -0,0 +1,61 @@
+-- Delete invalid events linked to invalid enrollments.
+delete from event e where e.enrollmentid in (
```
ameenhere, line 2: What about foreign key constraints of the eventid existing in other tables?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityOwnershipValidator.java
```diff
@@ -304,9 +300,7 @@ private void checkEventOrgUnitWriteAccess(
       OrganisationUnit eventOrgUnit,
       boolean isCreatableInSearchScope,
       UserDetails user) {
-    if (eventOrgUnit == null) {
-      log.warn(ORG_UNIT_NO_USER_ASSIGNED, event.getUid());
-    } else if (isCreatableInSearchScope
+    if (isCreatableInSearchScope
```
muilpp, line 303: maybe I'm missing something...
but [here](https://github.com/dhis2/dhis2-core/pull/19193/files#diff-f19b9c82388c030b253a1dc12bdd536b1ac900bb398162a5aca6bd81c9cfb894R99) we validate the user capture scope if the strategy is not an UPDATE.
since this method is only called when the strategy is not update, do we need to validate the org unit scope again?

enricocolasante: This whole validator is quite messy and difficult to follow.
We already have a bug to double check that everything in here is correct.
https://dhis2.atlassian.net/browse/DHIS2-17335

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
enricocolasante: :fearful: 
I completely forgot any of this and I got lucky in the SL database when applying this.
I am going to add all the necessary checks on other tables, I will probably split this migration in 2 then.

teleivo:
What is a placeholder enrollment?
We used the term dummy enrollment AFAIK so far for programs without registration. Is that the same or a different enrollment?

> 1. 1.1 If any of the Event OU inside the enrollment is  not null, update the same as Enrollment OU

Can the events have different OU? If so, what consequence can it have to pick one?

> 1.2 If all Event OUs are null and the enrollment is a placeholder (link to a program WITHOUT_REGISTRATION) then use the root OU.

What is the root OU?

> After step 1, for all Event OU nulls, update the Enrollment OU as the event OU

Do you mean `update the event OU with the enrollment OU`?

enricocolasante:
> What is a placeholder enrollment? We used the term dummy enrollment AFAIK so far for programs without registration. Is that the same or a different enrollment?

Yes, a placeholder enrollment is the dummy enrollment

> 
> > 1. 1.1 If any of the Event OU inside the enrollment is  not null, update the same as Enrollment OU
> 
> Can the events have different OU? If so, what consequence can it have to pick one?
> 

Here we are guessing a good OU. This will come together with a big warning in the release notes, if they have a lot of enrollment/events with null OU, they should check the data beforehand and understand what they want to do with them. Our solution is a best effort solution in order to make the system consistent and remove the possibility to have inconsistent data in the DB

> > 1.2 If all Event OUs are null and the enrollment is a placeholder (link to a program WITHOUT_REGISTRATION) then use the root OU.
> 
> What is the root OU?

It is a OU with no parent, hence a OU with hierarchy level 1

> 
> > After step 1, for all Event OU nulls, update the Enrollment OU as the event OU
> 
> Do you mean `update the event OU with the enrollment OU`?

Yes :)

> 
> Deleting data that is not ours always feels scary 😨 Are we certain we are not deleting anything that might still be of value? Can we move the data somewhere so it could be analyzed/recovered/made valid? If not we should at least add a big notice to the release notes with a sql query to run beforehand to retrieve the data that would be deleted.

We will have a gib warning on the release notes about it.
Anyway, tracker data without an attached tracked entity is just lost and meaningless data that shouldn't ever be in the DB.
I added that part in the migration because we had some dirty data in SL database that needed to be fixed to make the migration work, but "I hope" that wouldn't be the case in most of DBs out there.
We are going to have a warning in the release notes and maybe also a query to make them check the invalid data before upgrading.

muilpp:
>     For Enrollment OU nulls
>     1.1 If any of the Event OU inside the enrollment is not null, update the same as Enrollment OU
>     1.2 If all Event OUs are null and the enrollment is a placeholder (link to a program WITHOUT_REGISTRATION) then use the root OU.
>     1.3 For tracker enrollments that still have null OUs, use the TEI OU as Enrollment OU
> 
>     After step 1, for all Event OU nulls, update the Enrollment OU as the event OU
> 

If an enrollment or event doesn't have an org unit, are we currently showing it to the user? If we are not, maybe it'd be better to not update the org unit at all and keep on not showing it. Maybe move it to another table and let the implementation decide what to do with that data, something similar to what we've done with change logs.
Again, assuming we are currently not showing it, could it happen that by updating the org unit we end up showing the same program stage twice in the same event?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
```diff
@@ -0,0 +1,37 @@
+DO $$
+DECLARE dummyOrgUnitId bigint;
+DECLARE dummyOrgUnitUid varchar(11);
+BEGIN
+    select coalesce((select max(organisationunitid) + 1 from organisationunit), 1) into dummyOrgUnitId;
+    select generate_uid() into dummyOrgUnitUid;
+
+    while (select count(*) from organisationunit where uid = dummyOrgUnitUid) > 0 loop
```
ameenhere, line 8: Maybe replace by count(1) or count(organisationunitid) or maybe even count(uid) since we are not interested in any of the actual columns. To make it pure index scan based.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
```diff
@@ -0,0 +1,37 @@
...
+        select generate_uid() into dummyOrgUnitUid;
+    end loop;
+
+    insert into organisationunit
+    (organisationunitid, name, code, parentid, shortname, openingdate, created, lastupdated, uid, hierarchylevel)
```
ameenhere, line 13: I am thinking of several things here.
1. Is there some unique constraint on the name or code? Maybe not.
2. I am thinking if hierarchylevel=99 will have any consequences? Since it will also not have a parent id, it will simply be an orphan. 
3. Functionally we might be reliant on the path property, in this case path is null which we might have not handled?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
zubaira: Will this be executed for all instances, regardless of whether they have the inconsistency? Would it be more sensible to first check for any inconsistencies before proceeding with the procedure?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
```diff
@@ -0,0 +1,18 @@
+DO $$
```
zubaira, line 1: Would it make sense to first check for inconsistencies and only execute the procedure if any are found, thereby avoiding running it on all instances?

enricocolasante: This procedure is doing nothing else than remove null values if there is any.
It is a procedure because we need to catch the exception and raise a understandable message.

ameenhere (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
```diff
@@ -0,0 +1,18 @@
...
+    where en.organisationunitid is null
+      and en.programid in (select programid from program where type = 'WITHOUT_REGISTRATION');
+
+    alter table if exists enrollment alter column organisationunitid set not null;
+
+    alter table if exists event alter column organisationunitid set not null;
+
+EXCEPTION
+  WHEN not_null_violation THEN
+    RAISE EXCEPTION 'There are inconsistent data in your DB. Please check https://github.com/dhis2/dhis2-releases/blob/master/releases/2.42/migration-notes.md#null-organisation-unit to have more information on the issue and to find ways to fix it. Detailed error message: %', SQLERRM;
```
teleivo, line 15:
```suggestion
    RAISE EXCEPTION 'There is inconsistent data in your DB. Please check https://github.com/dhis2/dhis2-releases/blob/master/releases/2.42/migration-notes.md#null-organisation-unit to have more information on the issue and to find ways to fix it. Detailed error message: %', SQLERRM;
```

The URL better not change 😅 maybe a permalink is better, downside is its going to point to outdated docs if you do change its content.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
teleivo: @zubaira 

is 

```sql
    where en.organisationunitid is null
      and en.programid in (select programid from program where type = 'WITHOUT_REGISTRATION');
```

not limiting/finding the inconsistencies?

teleivo (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_28__Make_ou_event_and_enrollment_not_null.sql
zubaira: @teleivo 
What I meant was something like this, but I don't believe it will yield much benefit.

```
BEGIN
    SELECT 
        (SELECT COUNT(*) 
         FROM enrollment 
         WHERE organisationunitid IS NULL 
           AND programid IN (SELECT programid FROM program WHERE type = 'WITHOUT_REGISTRATION'))
        +
        (SELECT COUNT(*) 
         FROM event 
         WHERE organisationunitid IS NULL)
        INTO total_nulls;

    IF total_nulls > 0 THEN
   -- the rest of the procedure goes here
```

PR #19214, opened 2024-11-19
feat: report metadata without identifier for idScheme DHIS2-14968

teleivo:
Report any metadata that does not have an identifier for the requested `idScheme`.

For example

`GET http://localhost:8080/api/tracker/events/FV4JCI73wO2?idScheme=ATTRIBUTE:Y1LUDU8sWBR`

gets response

```http
HTTP/1.1 422 

{
  "httpStatus": "Unprocessable Entity",
  "httpStatusCode": 422,
  "status": "ERROR",
  "message": "Not all metadata has an identifier for the requested idScheme. Either change the requested idScheme or add the missing identifiers to the metadata.",
  "devMessage": "Following metadata listed using their UIDs is missing identifiers for the requested idScheme:\nProgramStage[ATTRIBUTE:Y1LUDU8sWBR]=A03MvHHogjR\nCategoryOptionCombo[ATTRIBUTE:Y1LUDU8sWBR]=HllvX50cXC0\nCategoryOption[ATTRIBUTE:Y1LUDU8sWBR]=xYerKDKCefk\nProgram[ATTRIBUTE:Y1LUDU8sWBR]=IpHINAT79UW\nOrganisationUnit[ATTRIBUTE:Y1LUDU8sWBR]=DiszpKrYNg8"
}
```

if you process the `devMessage` via `| jq -r .devMessage` it looks like

```
Following metadata listed using their UIDs is missing identifiers for the requested idScheme:
ProgramStage[ATTRIBUTE:Y1LUDU8sWBR]=A03MvHHogjR
CategoryOptionCombo[ATTRIBUTE:Y1LUDU8sWBR]=HllvX50cXC0
CategoryOption[ATTRIBUTE:Y1LUDU8sWBR]=xYerKDKCefk
Program[ATTRIBUTE:Y1LUDU8sWBR]=IpHINAT79UW
OrganisationUnit[ATTRIBUTE:Y1LUDU8sWBR]=DiszpKrYNg8
```

We show at most 20 UIDs to shield against requests to retrieve all events. The following is an example (capped at 2 to illustrate) how it looks

```
Following metadata listed using their UIDs is missing identifiers for the requested idScheme:
ProgramStage[ATTRIBUTE:Y1LUDU8sWBR]=CWaAcQYKVpq,edqlbukwRfQ,...
CategoryOptionCombo[ATTRIBUTE:Y1LUDU8sWBR]=HllvX50cXC0
CategoryOption[ATTRIBUTE:Y1LUDU8sWBR]=xYerKDKCefk
Program[ATTRIBUTE:Y1LUDU8sWBR]=VBqh0ynB2wv,qDkgAbB5Jlk,...
OrganisationUnit[ATTRIBUTE:Y1LUDU8sWBR]=DiszpKrYNg8,g8upMTyEZGZ,...
```

Status 422 was the most fitting I found for this use case

> The HTTP 422 Unprocessable Content client error response status code indicates that the server understood the content type of the request content, and the syntax of the request content was correct, but it was unable to process the contained instructions.

Clients that receive a 422 response should expect that repeating the request without modification will fail with the same error.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

* make `MetadataMapper` generic so it can map all the metadata we support while picking the correct `idScheme`
* extract common CSV related code in the controllers

## Limitations DataValues

We cannot report data elements! Reason is that `EventDataValues.dataElement` is of type `String` and not a `MetadataObject` like the other metadata. `EventDataValues.dataElement` is set to the identifier in the requested `idScheme`. We could insert the UID if we don't have another identifier. We would not have a reliable way to tell that the String is an identifier in the correct idScheme or a UID. So right now data values will not show up if we don't have an identifier for the requested idScheme. To fix that we would either need to make `EventDataValues.dataElement` a `DataElement` (basically undoing https://dhis2.atlassian.net/browse/DHIS2-4477) or make it a `MetadataIdentifier` which carries the `idScheme` information.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
@@ -18,6 +18,25 @@
       "trackedEntityTypeAttribute": true,
       "unique": true,
       "valueType": "TEXT"
+    },
+    {
+      "id": "i57a0734128",
+      "categoryOptionAttribute": true,
+      "categoryOptionComboAttribute": true,
+      "dataElementAttribute": true,
+      "name": "no metadata should have an attribute value for this attribute",
+      "organisationUnitAttribute": true,
+      "programAttribute": true,
+      "programStageAttribute": true,
+      "relationshipTypeAttribute": true,
+      "sharing": {
+        "owner": "PQD6wXJ2r5j",
+        "public": "rw------"
+      },
+      "trackedEntityAttributeAttribute": true,
+      "trackedEntityTypeAttribute": true,
+      "unique": true,
+      "valueType": "TEXT"
     }
   ],
   "categories": [
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/TrackerIdSchemeParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/TrackerIdSchemeParams.java
@@ -38,6 +38,7 @@ import lombok.NoArgsConstructor;
 import org.hisp.dhis.category.CategoryOption;
 import org.hisp.dhis.category.CategoryOptionCombo;
 import org.hisp.dhis.common.IdentifiableObject;
+import org.hisp.dhis.common.MetadataObject;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.program.Program;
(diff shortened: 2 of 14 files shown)


dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
```diff
@@ -18,6 +18,25 @@
       "trackedEntityTypeAttribute": true,
       "unique": true,
       "valueType": "TEXT"
+    },
+    {
+      "id": "i57a0734128",
+      "categoryOptionAttribute": true,
+      "categoryOptionComboAttribute": true,
+      "dataElementAttribute": true,
+      "name": "idScheme related attribute no metadata should have",
```
muilpp, line 27: does it make sense to rename this to
`idScheme related attribute without metadata indentifier`?

muilpp (approved):

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
teleivo: That name proposal does not sound correct to me 🤔 The attribute itself has no metadata identifier. It's that other metadata like a program can have an attribute value for that attribute i.e. an identifier for the idScheme `ATTRIBUTE:i57a0734128`.

muilpp: ok, it's just when I read `idScheme related attribute no metadata should have`, I don't know that that means

ameenhere (approved):

PR #19217, opened 2024-11-19
refactor: Add sort key to Table [DHIS2-16705]

larshelge:
Adds a sort key list property to `Table`. The sort key is applicable to Doris and ClickHouse analytics databases. The sort key is used to sort rows in tables, which is the primary approach for achieving performance in column oriented data warehouses.

The sort key can be used in `DorisSqlBuilder` # `createTable` method for the `key` clause, to provide a meaningful key for tables which don't have a primary key, such as the aggregate data value table and completeness table.


luciano-fiandesio (commented):
@larshelge can we wait until later today to merge this PR, because I did something similar yesterday.

larshelge:
I guess we could merge now and then I can help with reconciling the DHIS2-16705 branch @luciano-fiandesio ?

maikelarabori (approved):

PR #19238, opened 2024-11-20
debug: experimental getPath on OrganisationUnit

netroms:


jbee:
What this experiment has shown us is that not re-computing `path` on access is a important step we have to take to avoid performance, consistency and stale data issues. We know that `path` should just be a normally fetched property fetched "as is" from the DB column. However, looking into on how to maintain consistency for `path` in the DB it became clear that this isn't an easy task. It will need further investigation to find a proper solution that improves performance while also delivering consistency and make it unnecessary to manually trigger a full table updating of the "path" because the parent references have changed.

PR #19239, opened 2024-11-20
fix: Org Unit split being assigned to incorrect users [DHIS2-18423]

david-mackessy:
# Issue
Org Unit split feature is assigning the split org units to incorrect users (sometimes all the users in the DB in certain circumstances)

# Cause
After creating the `UserQueryParams` with the source `OrganisationUnit`, that `OrganisationUnit` was being overridden to use the `User`'s `OrganisationUnit`s carrying out the split.

In cases where the `User` either had no `organisationUnits`, `dataViewOrganisationUnits`, or `teiSearchOrganisationUnits`, then all the `User`s in the DB would be returned from the query, resulting in all `User`s getting the split `OrganisationUnit`s.

Fix avoids using:
- `UserService#handleUserQueryParams` which was overriding the source `OrganisationUnit` already set
-  `UserStore#getUsers` which tries to generate a handcrafted SQL query taking into consideration a multitude of options & checks using `UserQueryParams` (this looks too complex and tries to accommodate too many things). This resulted in returning all `User`s in the DB

# Fix
Use a dedicated query to fetch the `User`s to be actioned.
Fewer select queries are generated as a result of the fix, and is not proportionate to the number of users found (old query was triggering 1 extra query per user found - manual testing).
Warning added to `UserQueryParams` property.

# Testing
## Automated
2 integration tests added
- 1 for Org Unit split service: checking that the correct users get assigned the split org units
- 1 for the store: checking the generated select query count & retrieved users

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
@@ -75,6 +75,11 @@ public class UserQueryParams {
 
   @ToString.Include private UserInvitationStatus invitationStatus;
 
+  /**
+   * WARNING: If this field has an empty collection, it can potentially result in all users being
+   * returned from the database. Check the generated query to make sure you're seeing expected
+   * queries and results.
+   */
   private Set<OrganisationUnit> organisationUnits = new HashSet<>();
 
   private Set<UserGroup> userGroups = new HashSet<>();
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
@@ -75,11 +75,7 @@ public class UserQueryParams {
 
   @ToString.Include private UserInvitationStatus invitationStatus;
 
-  /**
-   * WARNING: If this field has an empty collection, it can potentially result in all users being
-   * returned from the database. Check the generated query to make sure you're seeing expected
-   * queries and results.
-   */
+  /** If empty, no matching condition will be generated */
   private Set<OrganisationUnit> organisationUnits = new HashSet<>();
 
   private Set<UserGroup> userGroups = new HashSet<>();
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
```diff
@@ -75,6 +75,11 @@ public class UserQueryParams {
 
   @ToString.Include private UserInvitationStatus invitationStatus;
 
+  /**
+   * WARNING: If this field has an empty collection, it can potentially result in all users being
+   * returned from the database. Check the generated query to make sure you're seeing expected
+   * queries and results.
+   */
```
jbee, line 82: I think this phrasing is a bit misleading. This is one of many potential filters that when empty does not filter away anything. That sounds like expected behaviour to me as long as it is clear that empty means "not set" and not "match against an empty set". I think the comment we leave here should make that clear.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
```diff
@@ -891,4 +893,15 @@ void validateTwoFactorUpdate(boolean before, boolean after, User userToModify)
...
+  /**
+   * Method that retrieves all {@link User}s that have an entry for the {@link OrganisationUnit} in
+   * the given table
+   *
+   * @param orgUnitTable table to search (one of 'organisationUnits', 'dataViewOrganisationUnits',
+   *     'teiSearchOrganisationUnits')
+   * @param uid {@link OrganisationUnit} {@link UID} to match on
+   * @return matching {@link User}s
+   */
+  List<User> getUsersWithOrgUnit(@Nonnull String orgUnitTable, @Nonnull UID uid);
```
jbee, line 906: I don't think we should have names of tables escape the store. That is one of the main points to have a store abstraction. Maybe use an enum? 
While these 3 properties are backed by mapping tables in the DB in the object model these are properties so the wording should be accordingly.

david-mackessy: enum added

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserQueryParams.java
david-mackessy: I've updated the comment, keeping it broad enough to still help people when debugging. (this would have saved me some time debugging).
If it's still not acceptable, please suggest an alternative.

netroms (approved):

jbee (approved):

PR #19243, opened 2024-11-20
feat: Implement sorting on event change logs [DHIS2-18012]

muilpp:
This PR introduces sorting change logs by creation date, username, data element, or event property.

Unlike other endpoints, this one allows sorting by only one element at a time. If the chosen element is not the creation date, a secondary sort will be applied by creation date to ensure the change logs are presented in a timely order.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
-import java.util.List;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
@@ -48,7 +47,7 @@ public class EventChangeLogOperationParams {
     // does not support. The repeated order field and private order method prevent access to order
     // via the builder.
     // Order should be added via the orderBy builder methods.
-    private EventChangeLogOperationParamsBuilder order(List<Order> order) {
+    private EventChangeLogOperationParamsBuilder order(Order order) {
       return this;
     }
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
@@ -44,8 +44,6 @@ public class EventChangeLogOperationParams {
 
   public static class EventChangeLogOperationParamsBuilder {
 
-    private Order order;
-
     // Do not remove this unused method. This hides the order field from the builder which Lombok
     // does not support. The repeated order field and private order method prevent access to order
     // via the builder.
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
@@ -27,7 +27,6 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
-import java.util.ArrayList;
 import java.util.List;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
@@ -41,11 +40,11 @@ import org.hisp.dhis.tracker.export.Order;
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class EventChangeLogOperationParams {
 
-  private List<Order> order;
+  private Order order;
 
   public static class EventChangeLogOperationParamsBuilder {
 
-    private final List<Order> order = new ArrayList<>();
+    private Order order;
 
     // Do not remove this unused method. This hides the order field from the builder which Lombok
     // does not support. The repeated order field and private order method prevent access to order
@@ -56,7 +55,7 @@ public class EventChangeLogOperationParams {
     }
 
     public EventChangeLogOperationParamsBuilder orderBy(String field, SortDirection direction) {
-      this.order.add(new Order(field, direction));
+      this.order = new Order(field, direction);
       return this;
     }
   }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
@@ -50,6 +50,10 @@ import org.springframework.stereotype.Repository;
 @Repository("org.hisp.dhis.tracker.export.event.HibernateEventChangeLogStore")
 public class HibernateEventChangeLogStore {
   private static final String COLUMN_CHANGELOG_CREATED = "ecl.created";
+  private static final String COLUMN_CHANGELOG_USER = "ecl.createdByUsername";
+  private static final String COLUMN_CHANGELOG_DATA_ELEMENT = "d.uid";
+  private static final String COLUMN_CHANGELOG_PROPERTY = "ecl.eventProperty";
+
   private static final String DEFAULT_ORDER =
       COLUMN_CHANGELOG_CREATED + " " + SortDirection.DESC.getValue();
 
@@ -61,7 +65,11 @@ public class HibernateEventChangeLogStore {
    * map to a view model. This mapping is not necessary for change logs.
    */
   private static final Map<String, String> ORDERABLE_FIELDS =
-      Map.ofEntries(entry("createdAt", COLUMN_CHANGELOG_CREATED));
+      Map.ofEntries(
+          entry("createdAt", COLUMN_CHANGELOG_CREATED),
+          entry("username", COLUMN_CHANGELOG_USER),
+          entry("dataElement", COLUMN_CHANGELOG_DATA_ELEMENT),
+          entry("property", COLUMN_CHANGELOG_PROPERTY));
 
   private final EntityManager entityManager;
 
@@ -73,8 +81,7 @@ public class HibernateEventChangeLogStore {
     entityManager.unwrap(Session.class).save(eventChangeLog);
   }
 
-  public Page<EventChangeLog> getEventChangeLogs(
-      UID event, List<Order> order, PageParams pageParams) {
+  public Page<EventChangeLog> getEventChangeLogs(UID event, Order order, PageParams pageParams) {
 
     String hql =
         String.format(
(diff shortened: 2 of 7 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
```diff
@@ -56,7 +53,7 @@ private EventChangeLogOperationParamsBuilder order(List<Order> order) {
     }
 
     public EventChangeLogOperationParamsBuilder orderBy(String field, SortDirection direction) {
```
teleivo, line 55: is

```
    private EventChangeLogOperationParamsBuilder order(List<Order> order) {
```

above still needed?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -73,8 +81,7 @@ public void addEventChangeLog(EventChangeLog eventChangeLog) {
     entityManager.unwrap(Session.class).save(eventChangeLog);
   }
 
-  public Page<EventChangeLog> getEventChangeLogs(
-      UID event, List<Order> order, PageParams pageParams) {
+  public Page<EventChangeLog> getEventChangeLogs(UID event, Order order, PageParams pageParams) {
```
teleivo, line 84: please add the nullable/nonnull annotations

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
             assertPropertyCreate("geometry", "(-11.419700, 8.103900)", geometryChangeLogs.get(1)));
   }
 
-  private void updateDataValue(String event, String dataElementUid, String newValue)
-      throws IOException {
-    TrackerObjects trackerObjects = fromJson("tracker/event_and_enrollment.json");
+  private static Stream<Arguments> provideOrderParamsDesc() {
```
teleivo, line 462: there is one SortDirection.ASC in the args

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
...
+    return Stream.of(
+        Arguments.of(
```
teleivo, line 464: All the cases have

```java
            Arrays.asList("25", "20", "15"),
            Arrays.asList("20", "15", null)),
```

so they should not be a parameter as they are always the same. Right now I had to figure that out which is clear if you put them as is into the test.

I hope this would also make it easy to compare the expectations with the inverse (reverse order) test of `shouldSortChangeLogsWhenOrderingByCreatedAtAsc`.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
...
+            "username",
+            SortDirection.ASC,
+            Arrays.asList("25", "20", "15"),
+            Arrays.asList("20", "15", null)));
+  }
+
+  @ParameterizedTest
+  @MethodSource("provideOrderParamsDesc")
+  void shouldSortChangeLogsByCreatedAtDescWhenOrderingByDateOrUsername(
```
teleivo, line 483: Is this test supposed to assert the secondary sort on createdAt? as the current/previousValues which I assume are the expected ones are all the same for every test in here.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
...
+      String field,
+      SortDirection sortDirection,
+      List<String> currentValues,
+      List<String> previousValues)
+      throws ForbiddenException, NotFoundException {
+    EventChangeLogOperationParams params =
+        EventChangeLogOperationParams.builder().orderBy(field, sortDirection).build();
+    Event event = getEvent("QRYjLTiJTrA");
+    String dataElementUid = event.getEventDataValues().iterator().next().getDataElement();
```
teleivo, line 492: You could create a helper as I think you are doing this often, maybe `getFirst`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
...
+        () ->
+            assertDataElementUpdate(
+                "GieVkTxp4HH", previousValues.get(1), currentValues.get(1), changeLogs.get(1)),
+        () -> assertDataElementCreate("GieVkTxp4HH", currentValues.get(2), changeLogs.get(2)));
+  }
+
+  @Test
+  void shouldSortChangeLogsWhenOrderingByCreatedAtAsc()
+      throws ForbiddenException, NotFoundException {
+
```
teleivo, line 515: We don't put newlines at the start of methods

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -463,9 +459,225 @@ void shouldReturnEventPropertiesChangeLogWhenExistingGeometryPointPropertyDelete
...
+    assertNumberOfChanges(3, changeLogs);
+    assertAll(
+        () -> assertDataElementCreate(dataElementUid, "15", changeLogs.get(0)),
+        () -> assertDataElementUpdate(dataElementUid, "15", "20", changeLogs.get(1)),
+        () -> assertDataElementUpdate(dataElementUid, "20", "25", changeLogs.get(2)));
+  }
+
+  @Test
+  void shouldSortChangeLogsWhenOrderingByCreatedAtDesc()
```
teleivo, line 537: is that not already part of `shouldSortChangeLogsByCreatedAtDescWhenOrderingByDateOrUsername` element 1 🤔

teleivo (commented):
Right now `EventChangeLogServiceTest` has test lifecycle method. That means you'll create a new instance of the test class for every test which also means we pay the setup cost for every test. Did you check how long this test takes to run?

We have `OrderAndPaginationExporterTest` is annotated with  `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` for this reason. All our read focused tests should use that IMHO. They can all share the same setup and we pay less cost in build time.

dhis-2/dhis-test-integration/src/test/resources/tracker/event_and_enrollment.json
```diff
@@ -888,7 +888,16 @@
             "idScheme": "UID",
             "identifier": "GieVkTxp4HH"
           },
-          "value": "14",
+          "value": "15",
+          "providedElsewhere": false
+        },
+        {
+          "created": "2022-04-22T06:00:38.339",
```
enricocolasante, line 895:
```suggestion
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -159,14 +166,21 @@ public void deleteEventChangeLog(Event event) {
     entityManager.createQuery(hql).setParameter("event", event).executeUpdate();
   }
 
-  private static String sortExpressions(List<Order> order) {
-    if (order.isEmpty()) {
+  private static String sortExpressions(Order order) {
```
enricocolasante, line 169: Any particular reason why this is `static`?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
teleivo: Why should it not be? 🤔 Functions should be `static`.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
muilpp: I agree with the first statement, after the last refactor they all ended up being the same, so it makes no sense to have those arguments anymore.
However, the reason why I couldn't combine it with `shouldSortChangeLogsWhenOrderingByCreatedAtAsc`, is because I want to test the `changeLogType`. Unless I do something very ugly in the test, I couldn't come up with a solution to test the four cases together.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
muilpp: Yes. The username is always the same, so this case will test the secondary sort.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventChangeLogServiceTest.java
```diff
@@ -65,15 +54,17 @@
...
 import org.junit.jupiter.api.Test;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.core.io.ClassPathResource;
+import org.springframework.transaction.annotation.Transactional;
 
-class EventChangeLogServiceTest extends PostgresIntegrationTestBase {
+@Transactional
```
teleivo, line 66: TrackerTest is already transactional.

muilpp:
> Right now `EventChangeLogServiceTest` has test lifecycle method. That means you'll create a new instance of the test class for every test which also means we pay the setup cost for every test. Did you check how long this test takes to run?
> 
> We have `OrderAndPaginationExporterTest` is annotated with `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` for this reason. All our read focused tests should use that IMHO. They can all share the same setup and we pay less cost in build time.

I moved the sorting related tests into `OrderEventChangeLogTest`, which extends `TrackerTest`.
`EventChangeLogServiceTest` now extends `TrackerTest` too, but does modify the data, so I've annotated with `@Transactional`. Even though it means we lose visibility in the database, it rolls back to the original state after every method is run, .
The execution time of both classes is much better now.

PR #19258, opened 2024-11-21
feat: Doris analytics event table population [DHIS-16705]

luciano-fiandesio:
### PR Overview

Various fixes to `JdbcEventAnalyticsTableManager` to allow the `analytics_event_*` table generation process to work with Apache Doris.

#### Add a `year` column for declarative partitioning

The new `year` column is required for declarative partitioning in Doris. It uses the existing `yearly` column to infer the `year`.

`CAST(SUBSTRING(dps.yearly, 1, 4) AS INTEGER)`

#### Fix OU subqueries

Organization Unit related subqueries are not working in Doris, when inserting data into the event tables.
This is an example of a invalid subquery:

```
select
   ou.uid
from
   "organisationunit" ou
where
   ou.uid = (
	select
	    eventdatavalues #>> '{S33cRBsnXPo, value}'
	from
	   "event"
	where
	   eventid = ev.eventid )) as "S33cRBsnXPo",
```

The fix removed the subselect and the event condition.
The `eventid` `where` condition is redundant, since `ev`  already represents the row from the event table that we are working with.
The new query works both on Doris and Postgres:

```
select
  ou.uid
from
  "organisationunit" ou
where
  ou.uid = ev.eventdatavalues #>> '{S33cRBsnXPo, value}') as "S33cRBsnXPo",
```

### Fix maplegend subqueries

When processing the `select` statement for data elements with legend set, the current Postgres-based engine generates a query that is invalid in Doris/MySql:

```
(
select
  l.uid
from
  pg_dhis.public.`maplegend` l inner join pg_dhis.public.`event` on
l.startvalue <= cast(JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) as bigint)
	and l.endvalue > cast(JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) as bigint)
	and l.maplegendsetid = 906836
	and eventid = ev.eventid
	and JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) regexp '^(-?[0-9]+)(\.[0-9]+)?$') as qrur9Dvnyt5_TiOkbpGEud4,
cast(SUBSTRING(dps.yearly, 1, 4) as INTEGER)
``` 

The subquery references columns from the "outer" query (`ev.eventid`). This is not supported in MySQL. MySQL does not allow you to reference columns from the outer query directly within a subquery in the SELECT clause.

The fix moves the `eventid = ev.eventid` condition from the join clause to a where clause.

```
select
  l.uid
from
  pg_dhis.public.`maplegend` l
inner join 
  pg_dhis.public.`event` 
on
  l.startvalue <= cast(JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) as bigint)
and 
  l.endvalue > cast(JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) as bigint)
and 
  l.maplegendsetid = 906836
and 
 JSON_UNQUOTE(JSON_EXTRACT(eventdatavalues, '$.qrur9Dvnyt5.value')) regexp '^(-?[0-9]+)(\.[0-9]+)?$'
where eventid = ev.eventid
) as qrur9Dvnyt5_TiOkbpGEud4,
```

The new query works both in Doris and Postgres.
@larshelge @maikelarabori I did some testing and moving the eventid condition into the `where` clause doesn't seem to affect the data, but please take a look if you can think of any potential problem.

### Add a `jsonExtractNested` method

Add a method to all `SqlBuilder` implementations, to extract a nested (`a.b.c.`) value from a JSON/JSONB column


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
```diff
@@ -461,10 +462,23 @@ tea, getNumericClause(), getDateClause(), true))
...
 
     return filterDimensionColumns(columns);
   }
 
+  @Override
+  protected AnalyticsTableColumn getPartitionColumn() {
+    return AnalyticsTableColumn.builder()
+        .name("year")
+        .dataType(INTEGER)
+        .selectExpression("CAST(SUBSTRING(dps.yearly, 1, 4) AS INTEGER)")
```
larshelge, line 478: Can we use resource table instead? Just feels more consistent with existing code and natural.

larshelge:
Looks really good! Nice work.

larshelge (approved):

maikelarabori:
Looks good to me, just waiting for the analytics e2e to run so that we can merge it. Thanks!

PR #19266, opened 2024-11-22
chore: let COC.isDefault look at its own name only DHIS2-14968

teleivo:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOptionCombo.java
```diff
@@ -162,7 +164,7 @@ public void removeAllCategoryOptions() {
 
   @Override
   public boolean isDefault() {
-    return categoryCombo != null && DEFAULT_NAME.equals(categoryCombo.getName());
+    return DEFAULT_CATEGORY_COMBO_NAME.equals(name);
```
jimgrace, line 167: I suggest `DEFAULT_NAME.equals(name)` because you're testing a CatOptionCombo name, not a CatCombo name. (They're both named "default" but it's better form to test against the one that's defined for this object.)

Note that the line you're replacing was also not great for the opposite reason; it was comparing a CatCombo name with the default CatOptionCombo name.

dhis-2/dhis-api/src/test/java/org/hisp/dhis/category/CategoryOptionComboTest.java
```diff
@@ -145,7 +145,7 @@ void hasDefault() {
 
   @Test
   void testIsDefault() {
-    categoryComboA.setName(CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME);
+    optionComboA.setName(CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME);
```
jimgrace, line 148: Likewise, I suggest `optionComboA.setName(CategoryOptionCombo.DEFAULT_NAME)`. You're setting a CatOptionCombo name so it's better form to use the CatOptionCombo default name.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOptionCombo.java
```diff
@@ -27,6 +27,8 @@
  */
 package org.hisp.dhis.category;
 
+import static org.hisp.dhis.category.CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME;
```
jimgrace, line 30: (not needed if you accept the suggest change below)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOptionCombo.java
teleivo: I used it as that is what is used to find it in https://github.com/dhis2/dhis2-core/blob/2313c47c91d9229ad0659b0ed5818d9ceaad40ba/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/DefaultCategoryService.java#L583

I am making your suggested change 👌🏻 but then also changing `DefaultCategoryService.getDefaultCategoryOptionCombo`.

jbee (approved):
As Jim mentioned there is code using the wrong constant for the string "default" because they all are the same but good that you cleaned that up as well. It is a bit confusing and I assume was just a copy&paste or autocomplete mistake made in the past.

jason-p-pickering:
Hmm. Is this the only way to solve this problem? I see that all of the tests pass, but I do worry about changing this for this particular reason. What happens when there are multiple default COCs? I have been working on a PR here to prevent import of new/duplicate COCs,  and we have some integrity checks around this as well, but we do know of systems which either have a default CC/CO/COC which are not the hard-coded ones which we currently have, or which have actual duplicate default COCs (i.e multiple "default" COCs) which are associated with the default CC.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOptionCombo.java
jimgrace: Very good, it would be right to change that too. Functionally it doesn't matter because all the names are "default", but if you do a code search for all uses of `CategoryOptionCombo.DEFAULT_NAME`, you should find all the tests for the CatOptionCombo default name. Likewise, if you search for all uses of `CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME` you should fine all the tests for the CatCombo default name, etc.

jbee:
> What happens when there are multiple default COCs? 

@jason-p-pickering It is a valid concern but, let me point out this line to you https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/DefaultCategoryService.java#L583 

As you can see the service is picking the default COC by the name "default" and it has for a long time. So the issue you worry about is already a reality for a long time. This change only makes the COC class logic be in line with this service logic and arguably with the logic of C, CC `isDefault()`.

jason-p-pickering (approved):
Fingers crossed! :-D

dhis-2/dhis-api/src/test/java/org/hisp/dhis/category/CategoryOptionComboTest.java
```diff
@@ -145,13 +145,13 @@ void hasDefault() {
...
   @Test
   void testIsDefault() {
-    categoryComboA.setName(CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME);
+    optionComboA.setName(CategoryOptionCombo.DEFAULT_NAME);
     assertTrue(optionComboA.isDefault());
   }
 
   @Test
   void testIsNotDefault() {
-    categoryComboA.setName(CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME + "x");
```
jimgrace, line 154: This was correct before the change. For a `CategoryCombo` you should use `CategoryCombo.DEFAULT_CATEGORY_COMBO_NAME`.

jimgrace (changes requested):

jimgrace (approved):
Looks good now!

PR #19268, opened 2024-11-22
feat: Add filtering capabilities in events change log [DHIS2-18012]

muilpp:
This PR introduces filtering for change logs. These filters differ slightly from the ones we currently use:

- They accept only one field at a time.
- They support only the equals operator.
- Filtering is limited to username, data element, or property.

Another key difference is that while our current filters are [UID based](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventRequestParamsMapper.java#L112), these new filters are string based.

Although the only supported operator is `eq`, I’ve retained the filter format [field]:[operator]:[filter] used in other endpoints. This way, if we need to support additional operators in the future, it won’t require a breaking change.

`validateFilter(String, Set<String>)` is used only for event change logs, but it will be used for attributes as soon as we refactor those change logs, that's why I added it to `RequestParamsValidator`

Note: ACL validation for data elements is not yet implemented. I will address this in a separate ticket once we decide how data sharing validation on data elements should be handled.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
@@ -27,10 +27,12 @@
  */
 package org.hisp.dhis.tracker.export.event;
 
+import java.util.Map;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
+import org.hisp.dhis.common.QueryFilter;
 import org.hisp.dhis.common.SortDirection;
 import org.hisp.dhis.tracker.export.Order;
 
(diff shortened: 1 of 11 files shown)


larshelge (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -73,6 +73,12 @@ public class HibernateEventChangeLogStore {
           entry("dataElement", COLUMN_CHANGELOG_DATA_ELEMENT),
           entry("property", COLUMN_CHANGELOG_PROPERTY));
 
+  private static final Map<String, String> FILTERABLE_FIELDS =
```
enricocolasante, line 76: In here we could somehow put the type information, we know that "dataElement" is a UID (the key instead of a String could be a Pair<String, Class<?>>), so then we could validate invalid UID format in the `validateFilter` method.
I think it would be more consistent in behavior as we are not accepting invalid UIDs in filters in other APIs.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
```diff
@@ -40,20 +42,30 @@
 public class EventChangeLogOperationParams {
 
   private Order order;
+  private Map.Entry<String, QueryFilter> filterEntry;
```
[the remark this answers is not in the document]
muilpp: Yeah, I just didn't want to use an external library, but I agree conceptually probably makes more sense to use `Pair` instead of `Map.Entry`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogOperationParams.java
```diff
@@ -40,20 +42,30 @@
...
-    // Order should be added via the orderBy builder methods.
+    // Do not remove these unused methods. They hide the order and filter fields from the builder
+    // which Lombok
+    // does not support.
+    // They should be added via their respective orderBy and filterBy builder methods.
     private EventChangeLogOperationParamsBuilder order(Order order) {
       return this;
     }
 
+    private EventChangeLogOperationParamsBuilder filterMap(Pair<String, QueryFilter> filterMap) {
```
enricocolasante, line 57:
```suggestion
    private EventChangeLogOperationParamsBuilder filter(Pair<String, QueryFilter> filter) {
```

enricocolasante (approved):

PR #19294, opened 2024-11-26
fix: use generated user name column for make query DB only [DHIS2-17200]

jbee:
### Summary
Makes the `User` `name` property a generated column that is auto-composed from `firstName` and `surname` in/by the DB.

This makes `name` a persisted property whereby `query=Xyz` also is a pure DB filter for `User`s. 
This is important to avoid the different terms in the OR clause generated from it to be split up between JPA and in-memory query because splitting would not result in the correct result as a filter like (A OR B OR C) AND X changes to (A OR B) AND X AND C (the OR bracket being created from the `query` parameter).

As a bonus this also means `name` is now a usable property in `filter` when using the Gist API.

This PR also corrects the HBM mapping where `firstName` and `surname`. They were incorrectly marked as nullable when they always have been `not null` in the DB. That means these properties changed in the schema from being optional to now be `required`. This change caused a couple of tests to fail as they were not providing the required fields.

This PR also fixes a bug in the "filter that always are false" detection. They only result in an empty list of filters are combined AND so this extra condition was added.

As a further improvement `query` will only add the `id:eq:<query>` filter part if the `<query>`  string has 11 characters as it is otherwise impossible for this filter to match.

### Automatic Testing
Existing tests where adjusted. A new test scenario was added to verify that `name` can be used as Gist API filter.

### Manual Testing 
* filter users by name `/api/users?filter=name:like:xyz` (compare results  to older versions)
* filter by query `/api/users?query=name:like:xyz` (compare results  to older versions)
* update `firstName` and/or `surname` of a user and check the change is automatically reflected in their `name`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
@@ -134,7 +134,12 @@ public final class Restrictions {
   }
 
   public static Disjunction query(Schema schema, String query) {
-    return or(schema, eq("id", query), eq("code", query), ilike("name", query, MatchMode.ANYWHERE));
+    Restriction name = ilike("name", query, MatchMode.ANYWHERE);
+    Restriction code = eq("code", query);
+    if (query.length() != 11) return or(schema, code, name);
+    // only a query with length 11 has a chance of matching a UID
+    Restriction id = eq("id", query);
+    return or(schema, id, code, name);
   }
 
   public static Disjunction or(Schema schema, Criterion... filters) {
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
@@ -133,6 +133,13 @@ public final class Restrictions {
     return new Restriction(path, new EmptyOperator<>());
   }
 
+  /**
+   * Builds a PR group to match the query string against id, code and name.
+   *
+   * @param schema of the root entity
+   * @param query the query string value used the URL {@code query} parameter
+   * @return OR group with the filters for the query string
+   */
   public static Disjunction query(Schema schema, String query) {
     Restriction name = ilike("name", query, MatchMode.ANYWHERE);
     Restriction code = eq("code", query);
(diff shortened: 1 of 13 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
```diff
@@ -134,7 +134,12 @@ public static Restriction isEmpty(String path) {
   }
 
   public static Disjunction query(Schema schema, String query) {
-    return or(schema, eq("id", query), eq("code", query), ilike("name", query, MatchMode.ANYWHERE));
+    Restriction name = ilike("name", query, MatchMode.ANYWHERE);
+    Restriction code = eq("code", query);
+    if (query.length() != 11) return or(schema, code, name);
```
david-mackessy, line 139: The name `query` is a bit confusing here for me. It sounds like it's the query param being checked for a length of 11. `query` to me implies it's the whole query string.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_32__User_name_as_generated_column.sql
```diff
@@ -0,0 +1,11 @@
...
+Adds the "name" column to users as a computed column based on first and last name
+to move named based filters into the DB.
+This is important as the web API "query" search is a combination of multiple columns
+one of which is name. Such combinations must stay all in DB to be correct.
+
+Since both source columns are not null the name can and should also be not null
+and the computation does not need to handle null.
+*/
+alter table userinfo
+    add column if not exists name character varying(321) not null generated always as ( firstname || ' ' || userinfo.surname ) stored;
```
david-mackessy, line 11: any reason the table name prefix is used for `surname` and not `firstname` when generating the value? it looks a little funny.

david-mackessy (approved):

david-mackessy:
Not sure if we have any metrics, but it would be nice to know if this has a positive performance impact and by how much, or if it mainly simplifies things for us in code/query.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
jbee: It is called `query` in the URL and I am trying to avoid requiring mental mapping by calling the value for the parameter the same name hoping that this makes it easy to follow once there is a bit of background knowledge about the parameter itself.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_32__User_name_as_generated_column.sql
jbee: That must have been auto-complete insert I didn't notice. I'll remove the prefix.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/Restrictions.java
jbee: I added javadoc

PR #19297, opened 2024-11-26
fix: [DHIS2-15272] Fix program expiryDays check with periodType for events

ameenhere:
There exists a configuration for setting a expiryPeriodType and expiryDays for a program. 

As per the documentation

> Expiry period type & Expiry days 
The expiry days defines for how many days after the end of the previous period, an event can be edited. The period type is defined by the expiry period type. This means that when the specified number of expiry days has passed since the end date of the previous period, the events from that period are locked.If you set the expiry type to "Monthly" and the expiry days to "10" and the month is October, then you can't add or edit an event to October after the 10th of November.

The validation that existed ignored the expiryDays part of the configuration.

This PR fixes this by making sure the current date is always before the last day of period in which the event took place  plus the expiryDays count. Meaning if the event occuredAt October and the expiryDays is 10, then the validation is updated so that it verifies currentDate is before ( October 31+10 days = Nov 10) 

Some additional changes done

- Renamed some of the validator private methods to better reflect the business/functional terminology.
- Indirectly this has fixed allowing Scheduling events into the future (which earlier was not possible if expiryPeriodType was configured)


larshelge (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidatorTest.java
```diff
@@ -246,6 +246,24 @@ void testEventIsNotValidWhenDateBelongsToExpiredPeriod() {
     assertHasError(reporter, event, E1047);
   }
 
+  @Test
+  void testEventIsValidWhenScheduledDateBelongsToFuturePeriod() {
```
muilpp, line 250: Some time ago, we decided to use the convention for our test names: `should[ExpectedResult]When[Action]`

muilpp:
This is what our AI told me when I asked what happens if I set a program's expiry days to 0.
Screenshot from 2024-11-26 20-21-23

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidatorTest.java
ameenhere: Thanks @muilpp ! Good catch on the expiryDays to 0 meaning indefinite. I'll update the code and add a test case for that scenario as well!

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidatorTest.java
```diff
@@ -246,12 +246,48 @@ void testEventIsNotValidWhenDateBelongsToExpiredPeriod() {
...
+
+    // when
+    validator.validate(reporter, bundle, event);
+
+    // then
+    assertIsEmpty(reporter.getErrors());
+  }
+
+  @Test
+  void shouldFailValidationForEventWhenScheduledDateBelongsToFuturePeriod() {
```
enricocolasante, line 268: Either the name of the test or the assertion is wrong here.
AFAIK the name seems correct but we should get an error than in the reporter.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidatorTest.java
muilpp: @ameenhere, I didn't know how the `exipiryDays` work, so I read our docs, but I couldn't find anything about setting its value to 0, so I asked AI. That said, I'm not sure we should rely on it, wouldn't be the first time it lies. Just wanted to mention it here before we make any decision about it.

ameenhere: I do remember expiryDays set to 0 is indeed the default which means "there is no expiry". 

But I was hoping that expiryDays is only (and always) applicable if a "expiryPeriodType" is set to something. 
Meaning if expiryPeriodType is null, then that means there is no expiry (whatever the expiryDays is configured as). 

But if expiryPeriodType is a specific periodType, then expiryDays would be used (even if it is 0) to limit the number days after the end of the period to lock the event. This would also make it possible to actually say, "If the period has ended, lock it!" . This was my assumption. But I realize that may not be existing behaviour.

For resolving the original bug atleast, I think it is safe that I do not change the existing behaviour. Whether or not we should change the behaviour to my assumption, we can decide later. So reverted the if clause change.

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidatorTest.java
ameenhere: I think the name was incorrect. ScheduledDate should be allowed for future periods. I have changed the name to reflect that is should pass.

PR #19318, opened 2024-11-27
fix: [DHIS2-15272] Handle final expiring day for expiryPeriodType and expiryDays

ameenhere:
Additional handling for the final expiring day.

If expiryPeriodType is Monthly. And expiryPeriodDays is 10. 
For an event occured in October 20th (for example), we would want to be able to edit the event upto and including November 10th. 

Since the period end date of the event is calculated as 202X-10-31 00:00:00 and adding 10 days will take it to 202X-11-10 00:00:00. If now() is after this, we say the event is locked to edit. But this means the event will be locked during the daytime of 10th November while user expectation is to be able to edit it even during the final day (until 23:59:59). To include the final 24 hours, we add one more day to the expiry check logic.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/DateValidator.java
@@ -115,8 +115,11 @@ class DateValidator implements Validator<Event> {
 
     if (eventPeriod
         .getEndDate()
-        .toInstant()
-        .plus(ofDays(program.getExpiryDays()))
+        .toInstant() // This will be 00:00 time of the period end date.
+        .plus(
+            ofDays(
+                program.getExpiryDays()
+                    + 1L)) // Extra day added to account for final 24 hours of expiring day
         .isBefore(Instant.now())) {
       reporter.addError(event, E1047, event);
     }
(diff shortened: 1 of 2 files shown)


muilpp (approved):

jason-p-pickering:
Wouldn't adding one more day to the expiry day itself achieve the same effect, and it could simply be documented as such? 

You may also want to consider to implement fractional expiry days as we have done on the aggregate side, which allows admins much finer grained control over exactly when a given dataset (or in this case program) should expire, since they may not always want it to expire at midnight. https://github.com/dhis2/dhis2-core/pull/18597

enricocolasante (approved):

ameenhere:
Thanks @jason-p-pickering 

> Wouldn't adding one more day to the expiry day itself achieve the same effect, and it could simply be documented as such?

I think having it behave like this by default is more intuitive and also straightforward in terms of documenting. As a user, when I configure expiryDays is 10, I would not expect that the event gets locked at midnight on the 9th day.

> What happens for instance when the expiry days are set to zero? 

Currently it behaves the same as in aggregate. There is no expiry if expiry days are set to zero.

> You may also want to consider to implement fractional expiry days as we have done on the aggregate side, which allows admins much finer grained control over exactly when a given dataset (or in this case program) should expire, since they may not always want it to expire at midnight.

This is a good point and I see the value in it. I'll create a separate jira feature issue for it.

jason-p-pickering:
OK, but then how is this going to behave between tracker and aggregate? If the expiry days are set to 5 on a program and 5 on an aggregate data set, will they expire at the same time?

PR #19320, opened 2024-11-27
fix: OpenAPI - object list extraction [DHIS2-17200]

jbee:
### Summary
The CRUD `getObjectList` method has 2 major complications for the OpenAPI analysis

* it uses a type variable as parameter object
* it has a "virtual" response type that gets substituted based in the current value of `@EntityType`

Type variable parameter objects were not supported before so a simple version was added that only supports this particular case.

The response broke due to too many properties escaping the parameter type analysis as it would pick up a property from the annotated field (which was renamed) and another one from the getter (which lacked the renaming annotation and therefore wasn't ignored). Therefore the de-duplication is now based on the Java property name, not the name the property gets after considering potential renames from an annotation.

I also added OpenAPI descriptions to all special parameters.

### Shared Parameter and Inheritance
Previously there was no support for inheritance which worked but would result in each subclass repeating (duplicating) the parameter properties defined in supertypes. With the changes in this PR a shared parameter will use the declaring class simple name as the "namespace" if no explicit name is defined via `@OpenApi.Shared(name="SchemaName")`.

### Automatic Testing
Added a few new OpenAPI tests to make sure the added or fixed features keep working

### Manual Testing
* check e.g. `/api/openapi/openapi.html?scope=entity:Attribute&skipCache=true&source=true&scope=entity:OrganisationUnit#OrganisationUnit.getObjectList` , there should be plenty of parameters with description, the return type should be summarized as `<object{#,organisationUnits:array[OrganisationUnit]}>`
* check e.g. `/api/openapi/openapi.json?scope=entity:OrganisationUnit` if you know OpenAPI speak

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
@@ -304,6 +304,8 @@ public class HibernateUserStore extends HibernateIdentifiableObjectStore<User>
       hql += hlp.whereAnd() + " u.selfRegistered = true ";
     }
 
+    // TODO(JB) shoundn't UserInvitationStatus.NONE match "u.invitation = false" and null mean no
+    // filter at all?
     if (UserInvitationStatus.ALL.equals(params.getInvitationStatus())) {
       hql += hlp.whereAnd() + " u.invitation = true ";
     }
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
@@ -152,21 +152,64 @@ public class UserController
   @Data
   @EqualsAndHashCode(callSuper = true)
   public static final class GetUserObjectListParams extends GetObjectListParams {
+    @OpenApi.Description(
+        "Limits results to users with the given phone number (shorthand for `filter=phoneNumber:eq:{value}`)")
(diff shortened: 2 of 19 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/user/hibernate/HibernateUserStore.java
```diff
@@ -304,6 +304,8 @@ private Query<?> getUserQuery(UserQueryParams params, List<String> orders, Query
       hql += hlp.whereAnd() + " u.selfRegistered = true ";
     }
 
+    // TODO(JB) shoundn't UserInvitationStatus.NONE match "u.invitation = false" and null mean no
```
jbee, line 307: FYI: I left this intentionally for another PR

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -217,7 +255,12 @@ private UserQueryParams toUserQueryParams(GetUserObjectListParams params) {
     res.setInvitationStatus(params.getInvitationStatus());
     res.setUserOrgUnits(params.isUserOrgUnits());
     res.setIncludeOrgUnitChildren(params.isIncludeChildren());
-    res.setOrgUnitBoundary(params.getOrgUnitBoundary());
+    String ou = params.getOu();
```
jbee, line 258: FYI: moving this code here is in a way a bugfix so that the `orgUnitBoundary` parameter can be used and is not always overridden by the effect of `addOrganisationUnit`.

vietnguyen (approved):

david-mackessy (approved):

PR #19330, opened 2024-11-28
Fix: Prioritise payload/bundled events over database events for enrolment[DHIS2-17806]

zubaira:
This PR fixes an issue where events linked to an enrollment were being merged from both the database and the tracker bundle payload without prioritizing the payload events. This caused outdated or incorrect events from the database to potentially overwrite the more relavant payload events.

- Added logic to prioritize events from the tracker bundle payload over events fetched from the database.
- Combined events efficiently using maps to ensure uniqueness and prioritize payload events when UIDs conflict.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/DefaultProgramRuleService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/programrule/DefaultProgramRuleService.java
@@ -34,7 +34,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.collections4.ListUtils;
 import org.hisp.dhis.dxf2.events.event.EventQueryParams;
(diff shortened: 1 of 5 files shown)


enricocolasante (commented):
Are you sure this was already working fine in 2.39?
It seems to me that there is the same code in 2.39 and I am not sure why it would work there.

zubaira:
> Are you sure this was already working fine in 2.39? It seems to me that there is the same code in 2.39 and I am not sure why it would work there.

yes, its working on 2.39. Can be tested using Child Program in 2.39.7
https://play.im.dhis2.org/stable-2-39-7

enricocolasante:
> > Are you sure this was already working fine in 2.39? It seems to me that there is the same code in 2.39 and I am not sure why it would work there.
> 
> yes, its working on 2.39. Can be tested using Child Program in 2.39.7 https://play.im.dhis2.org/stable-2-39-7

But then, why the same code is working on 2.39 and not working in 2.40?

zubaira:
> > > Are you sure this was already working fine in 2.39? It seems to me that there is the same code in 2.39 and I am not sure why it would work there.
> > 
> > 
> > yes, its working on 2.39. Can be tested using Child Program in 2.39.7 https://play.im.dhis2.org/stable-2-39-7
> 
> But then, why the same code is working on 2.39 and not working in 2.40?

I’m not entirely sure why this behavior differs. A rough guess would be that it might be due to changes in the underlying implementation or the libraries being used. We can revisit and further investigate the reasons behind why this worked in 2.39

enricocolasante (approved):

muilpp (approved):

PR #19342, opened 2024-11-29
fix:Adjust SQL for data integrity check of COCs without names

jason-p-pickering:
This check identifies COCs without names, but in the user interface, we use the name (which is of course blank). This creates issues with seeing anything useful in the UI. 

The SQL has been adjusted so that we use the string array of category options instead. This should hopefully be more reliable, but if the category options also have blank names, then it will still of course be a problem. We may need to consider some fallback in the UI? 

image


jbee (approved):

david-mackessy (approved):
Category Options require a name (DB constraint also), so hopefully your concern about empty CO names is just a very small corner case.

PR #19346, opened 2024-11-29
task: Remove old event change log endpoint [DHIS2-18477]

muilpp:
Removes `/api/audits/tackedEntityDataValue`, as now we've got `/api/tracker/events/{event}/changeLogs`.
It would have been first deprecated in 2.42, but because of the migration to the new table, it doesn't make sense to still expose that endpoint, as it points to the old database table that's no longer used.

With it, I'm removing the service and store along with a test class too.
I'm also renaming the `ChangeLogController` to `AuditController`, as it's more appropriate given what it does.


enricocolasante (approved):
We should add this info about what was removed in the release notes for v42

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -51,17 +50,6 @@ Page<EventChangeLog> getEventChangeLog(
       UID event, EventChangeLogOperationParams operationParams, PageParams pageParams)
       throws NotFoundException, ForbiddenException;
 
-  /**
-   * @deprecated use {@link EventChangeLogService#getEventChangeLog} instead
-   */
-  @Deprecated(since = "2.41")
```
teleivo, line 57: Are you going to remove the `TrackedEntity` from the remaining method names in here?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/merge/dataelement/DataElementMergeServiceTest.java
```diff
@@ -2546,8 +2555,9 @@ void dataValueAuditMergeDeleteTest() throws ConflictException {
   // --------------------------------------
   @Test
   @DisplayName(
-      "TrackedEntityDataValueChangeLog with references to source DataElements are not changed or deleted when sources not deleted")
-  void trackedEntityDataValueChangeLogMergeTest() throws ConflictException {
+      "TrackedEntityChangeLogs with references to source DataElements are not changed or deleted when sources not deleted")
+  void trackedEntityChangeLogMergeTest()
```
teleivo, line 2559: You renamed the test but kept the tracked entity in here even though the test seems to test `EventChangeLog...` things. Should the test not be named after `EventChangeLog` somehow? 😅

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AuditController.java
```diff
@@ -466,11 +374,11 @@ public RootNode getDataApprovalAudit(
   }
 
   /**
-   * @deprecated use {@link #getTrackedEnityChangeLog} instead.
+   * @deprecated use {@link #getTrackedEntityAudit} instead.
    */
   @Deprecated(since = "2.41")
   @GetMapping("trackedEntityInstance")
-  public RootNode getTrackedEnityInstanceChangeLog(
+  public RootNode getTrackedEntityInstanceChangeLog(
```
teleivo, line 381: Please rename the method as well. Why would we still call this change log 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
muilpp: The methods `deleteTrackedEntityDataValueChangeLog` and `deleteTrackedEntityDataValueChangeLog` I want to keep because there could potentially be an entry in the old table holding a foreign key to an event, so we would need to delete it as we've been doing so far.

I thought it made sense to keep the same name, so the distinction between the new and old logic is clear.

teleivo (approved):

PR #19362, opened 2024-12-02
chore: propagate SqlBuilder to Antlr Expression visitor [DHIS-16705]

luciano-fiandesio:
## Description

This PR attempts at propagating the `SqlBuilder` into the Antlr parser, so that parser can generate SQL expression compatible with the underlying analytics database.

### Changes

- ~~Configure the `ExpressionService` Spring Bean into a `ServiceConfig`, so that the `SqlBuilder` can be injected explicitly.~~
- Modified `DefaultProgramIndicatorService` so that the expression map is now built externally (see `ExpressionMapBuilder`). This allow to use the new `withSqlBuilder` method on the expressions in the map and pass the `SqlBuilder` to each expression.
- Add a "fluent" `withSqlBuilder` method to `ProgramExpressionItem`.
- Use a custom builder for `CommonExpressionVisitor` to make possible to initialize the existing `DefaultStatementBuilder` with `SqlBuilder`
- Add a new `cast` function to `SqlBuilder` to hanlde SQL casting

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
@@ -29,7 +29,18 @@ package org.hisp.dhis.config;
 
 import java.util.HashMap;
 import java.util.Map;
+import org.hisp.dhis.cache.CacheProvider;
 import org.hisp.dhis.common.DeliveryChannel;
+import org.hisp.dhis.common.DimensionService;
+import org.hisp.dhis.common.IdentifiableObjectManager;
+import org.hisp.dhis.constant.ConstantService;
+import org.hisp.dhis.db.SqlBuilderProvider;
+import org.hisp.dhis.db.sql.SqlBuilder;
+import org.hisp.dhis.expression.DefaultExpressionService;
+import org.hisp.dhis.expression.Expression;
+import org.hisp.dhis.expression.ExpressionService;
+import org.hisp.dhis.hibernate.HibernateGenericStore;
+import org.hisp.dhis.i18n.I18nManager;
 import org.hisp.dhis.i18n.ui.resourcebundle.DefaultResourceBundleManager;
 import org.hisp.dhis.i18n.ui.resourcebundle.ResourceBundleManager;
 import org.hisp.dhis.message.MessageSender;
@@ -44,6 +55,11 @@ import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
  */
 @Configuration("coreServiceConfig")
 public class ServiceConfig {
+
+  @Bean
+  public SqlBuilder sqlBuilder(SqlBuilderProvider provider) {
+    return provider.getSqlBuilder();
+  }
 
   @Bean("taskScheduler")
   public ThreadPoolTaskScheduler threadPoolTaskScheduler() {
@@ -70,5 +86,26 @@ public class ServiceConfig {
   @Bean("org.hisp.dhis.i18n.ui.resourcebundle.ResourceBundleManager")
   public ResourceBundleManager resourceBundleManager() {
     return new DefaultResourceBundleManager();
+  }
+
+  @Bean("org.hisp.dhis.expression.ExpressionService")
+  public ExpressionService expressionService(
+      @Qualifier("org.hisp.dhis.expression.ExpressionStore")
+          HibernateGenericStore<Expression> expressionStore,
+      ConstantService constantService,
+      DimensionService dimensionService,
+      IdentifiableObjectManager idObjectManager,
+      I18nManager i18nManager,
+      CacheProvider cacheProvider,
+      SqlBuilder sqlBuilder) {
+
+    return new DefaultExpressionService(
+        expressionStore,
+        constantService,
+        dimensionService,
+        idObjectManager,
+        i18nManager,
+        cacheProvider,
+        sqlBuilder);
   }
 }
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
@@ -29,18 +29,7 @@ package org.hisp.dhis.config;
 
 import java.util.HashMap;
 import java.util.Map;
-import org.hisp.dhis.cache.CacheProvider;
 import org.hisp.dhis.common.DeliveryChannel;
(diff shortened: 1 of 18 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
```diff
@@ -71,4 +87,25 @@ public DefaultOutboundMessageBatchService defaultOutboundMessageBatchService(
   public ResourceBundleManager resourceBundleManager() {
     return new DefaultResourceBundleManager();
   }
+
+  @Bean("org.hisp.dhis.expression.ExpressionService")
```
larshelge, line 91: I am wondering if this definition is actually unnecessary. Since `SqlBuilder` is now a bean, you can now inject it into DefaultExpressionService as usual with (Lombok) constructor injection.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
```diff
@@ -146,6 +147,10 @@ public class DefaultProgramIndicatorService implements ProgramIndicatorService {
 
   private final Cache<String> analyticsSqlCache;
 
+  private final SqlBuilder sqlBuilder;
+
+  public static ImmutableMap<Integer, ExpressionItem> PROGRAM_INDICATOR_ITEMS;
```
larshelge, line 152: It is a little bit of an anti-pattern to initialize `static` variables in constructors in singletons. Since this is a Spring singleton bean, the constructor is guaranteed to be invoked once. It is more natural to make it consistent with other dependencies, and instead use this:

```
private final Map<Integer, ExpressionItem> programIndicatorItems;
```

luciano-fiandesio: Sorry, my bad, this part is actually no longer like this. The problem is that this `PROGRAM_INDICATOR_ITEMS` variable is only used by Unit Tests and not by the runtime. But it can no longer be static, since we need to pass the SqlBuilder. Fixing the unit tests, this will change too.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/ClickHouseSqlBuilder.java
```diff
@@ -228,6 +229,15 @@ public String jsonExtractNested(String json, String... expression) {
     return String.format("JSONExtractString(%s, '%s')", json, path);
   }
 
+  @Override
+  public String cast(String column, DataType dataType) {
+    return switch (dataType) {
+      case NUMERIC -> "toDecimal64(" + column + ", 8)"; // 8 decimal places precision
```
larshelge, line 235: Use `String.format` over String concatenation.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisSqlBuilder.java
```diff
@@ -231,6 +232,15 @@ public String jsonExtractNested(String column, String... expression) {
     return String.format("json_unquote(json_extract(%s, '%s'))", column, path);
   }
 
+  @Override
+  public String cast(String column, DataType dataType) {
+    return switch (dataType) {
+      case NUMERIC -> "CAST(" + column + " AS DECIMAL)";
```
larshelge, line 238: Use `String.format` over String concatenation. Use lower-case SQL by default.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/PostgreSqlBuilder.java
```diff
@@ -249,6 +250,16 @@ public String jsonExtractNested(String column, String... expression) {
     return String.format("%s #>> '{%s}'", column, String.join(", ", expression));
   }
 
+  @Override
+  public String cast(String column, DataType dataType) {
+    return column
+        + switch (dataType) {
+          case NUMERIC -> "::numeric";
```
larshelge, line 257: Should we include the `column` argument so that these methods can be called consistently, as the Doris and ClickHouse methods include the column?

```
column + "::numeric";
```

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
```diff
@@ -287,6 +288,16 @@ public interface SqlBuilder {
    */
   String jsonExtractNested(String json, String... expression);
 
+  /**
+   * Generates a database-specific SQL casting expression for the given column or expression.
```
larshelge, line 292: We don't have to say "database-specific" in the Javadoc, as everything in `SqlBuilder` is database specific.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/PostgreSqlBuilder.java
luciano-fiandesio: The `column` argument is there? `return column...`

larshelge:
Looks good! Great to see these incremental improvements.

maikelarabori (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/config/ServiceConfig.java
luciano-fiandesio: Yes, good catch, removed.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/PostgreSqlBuilder.java
larshelge: Sorry my bad.

larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
larshelge: Understand. Yes good to get it out of the implementation if it is only used in tests.

PR #19365, opened 2024-12-02
fix: Remove redundant org unit UID lookup in analytics [DHIS2-16705]

larshelge:
It appears that event and enrollment analytics table SQL statements are making an unnecessary lookup to the `organisationunit` table to retrieve the org unit `uid` field. The `uid` is already available in the `trackedentityattributevalue` table `value` column. In other words, there is no point in looking up the `uid` using the `uid` as argument.

There appears to be separate code paths and methods in `getColumnForOrgUnitDataElement`and `getColumnForOrgUnitTrackedEntityAttribute` for retrieving the `name` and `geomentry`, for which a lookup is necessary.


luciano-fiandesio (commented):
It looks good, but I thins @maikelarabori and @gnespolino should double-check, because they know this part much better than I do.

PR #19372, opened 2024-12-03
fix: [DHIS2-18551] (2.39) Fix access check when registering TEI

ameenhere:
When registering a new TEI, there is no context of a program. But the old Tracker API goes ahead and checks program ownership for “all the programs in the system”. This itself was only part of the problem.

There is a cache used to store programOwnerOu for a TEI-Program combination. For TE creation, when the TE is yet to be created, it will not have an id (the primary key column which is obtained after a TE is persisted). So during the time of TE creation, the entity object will always have an id of 0. This 0 is used as part of the key to cache programOwners. This means any subsequent TEI registration will also have 0 as the id and will pick up the  "wrong owner OU" from the cache(the first one that got into the cache).

So irrelevant access checks + incorrect cache hits was the root cause of the original issue.

This PR fixes this by making sure programOwnership is not checked when registering a new TEI and rather simply check if the user has TE.orgUnit in their Capture Scope and if the user has Data Write access to the TE.TEType

This fix is consistent with the new Tracker API, in master. In v42(master), [when creating a new TE](https://github.com/dhis2/dhis2-core/blob/39b5511344023394c1099048e1e49659f15628bb/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java#L94), it only checks if the user has dataWrite access to the TeType and nothing of ownership is checked. It has separate access check flows when the TE is being created vs when the TE is being updated.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
@@ -188,11 +188,15 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
     List<String> errors = new ArrayList<>();
     if (!aclService.canDataWrite(user, trackedEntity.getTrackedEntityType())) {
       errors.add(
-              "User has no data write access to tracked entity type: " + trackedEntity.getTrackedEntityType().getUid());
+          "User has no data write access to tracked entity type: "
+              + trackedEntity.getTrackedEntityType().getUid());
     }
 
-    if ( trackedEntity.getOrganisationUnit() != null && !organisationUnitService.isInUserHierarchy(user,  trackedEntity.getOrganisationUnit())) {
-      errors.add("User has no write access to organisation unit: " +  trackedEntity.getOrganisationUnit().getUid());
+    if (trackedEntity.getOrganisationUnit() != null
+        && !organisationUnitService.isInUserHierarchy(user, trackedEntity.getOrganisationUnit())) {
+      errors.add(
+          "User has no write access to organisation unit: "
+              + trackedEntity.getOrganisationUnit().getUid());
     }
 
     return errors;
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
@@ -173,6 +173,31 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
     return canWrite(user, trackedEntity, programService.getAllPrograms());
   }
 
+  /**
+   * Check if the user can create the TE. For a regular user, they must have data write permissions
+   * to tracked entity type metadata as well as Capture Access to the Org Unit.
+   *
+   * @return No errors if a user has write access to the TrackedEntityType and to the OrgUnit
+   */
+  @Override
+  @Transactional(readOnly = true)
+  public List<String> canCreate(User user, TrackedEntityInstance trackedEntity) {
+    if (user == null || user.isSuper() || trackedEntity == null) {
+      return List.of();
+    }
+    List<String> errors = new ArrayList<>();
+    if (!aclService.canDataWrite(user, trackedEntity.getTrackedEntityType())) {
+      errors.add(
+              "User has no data write access to tracked entity type: " + trackedEntity.getTrackedEntityType().getUid());
+    }
+
+    if ( trackedEntity.getOrganisationUnit() != null && !organisationUnitService.isInUserHierarchy(user,  trackedEntity.getOrganisationUnit())) {
+      errors.add("User has no write access to organisation unit: " +  trackedEntity.getOrganisationUnit().getUid());
+    }
+
+    return errors;
+  }
+
   private List<String> canWrite(
       User user, TrackedEntityInstance trackedEntity, List<Program> programs) {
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
```diff
@@ -173,6 +173,35 @@ public List<String> canWrite(User user, TrackedEntityInstance trackedEntity) {
...
+  /**
+   * Check if the user can create the TE. For a regular user, they must have data write permissions
+   * to tracked entity type metadata as well as Capture Access to the Org Unit.
+   *
+   * @return No errors if a user has write access to the TrackedEntityType and to the OrgUnit
+   */
+  @Override
+  @Transactional(readOnly = true)
+  public List<String> canCreate(User user, TrackedEntityInstance trackedEntity) {
+    if (user == null || user.isSuper() || trackedEntity == null) {
```
muilpp, line 185: If I remember correctly, user was null when running internal jobs, but after the last refactor by the platform team, that's no longer the case

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/dxf2/events/TrackedEntityInstanceServiceTest.java
```diff
@@ -603,6 +607,45 @@ void testSavePerson() {
     assertEquals(ImportStatus.SUCCESS, importSummary.getStatus());
   }
 
+  @Test
+  void shouldFailWhenRegisteringPersonOutsideCaptureScope() {
+
```
muilpp, line 612: Usually, we don't include a line break between the method name and its code, and we format the test to follow the Arrange-Act-Assert pattern.

muilpp (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
ameenhere: This PR is for 2.39, and I also see other methods in the class doing the same user == null check. 
Unsure if the refactor you talk about exists in 2.39.

ameenhere: Ah for some reason, your second comment just got in. Yes Agreed :D

teleivo: That a `null` user means `canCreate` == true does not feel right. If we assume that the user must be authenticated when this is called we should not check for null. Better an NPE than unauthorized access

zubaira:
> When registering a new TEI, there is no context of a program. But the old Tracker API goes ahead and checks program ownership for “all the programs in the system”. This itself was only part of the problem.
> 
> There is a cache used to store programOwnerOu for a TEI-Program combination. For TE creation, when the TE is yet to be created, it will not have an id (the primary key column which is obtained after a TE is persisted). So during the time of TE creation, the entity object will always have an id of 0. This 0 is used as part of the key to cache programOwners. This means any subsequent TEI registration will also have 0 as the id and will pick up the "wrong owner OU" from the cache(the first one that got into the cache).
> 
> So irrelevant access checks + incorrect cache hits was the root cause of the original issue.
> 
> This PR fixes this by making sure programOwnership is not checked when registering a new TEI and rather simply check if the user has TE.orgUnit in their Capture Scope and if the user has Data Write access to the TE.TEType
> 
> This fix is consistent with the new Tracker API, in master. In v42(master), [when creating a new TE](https://github.com/dhis2/dhis2-core/blob/39b5511344023394c1099048e1e49659f15628bb/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/trackedentity/SecurityOwnershipValidator.java#L94), it only checks if the user has dataWrite access to the TeType and nothing of ownership is checked. It has separate access check flows when the TE is being created vs when the TE is being updated.

This is great, @ameenhere. Just for clarification:
Once the cache is polluted, will the error occur even if the same user tries to register another TEI, or does it only happen when a different user attempts to register a TEI?

teleivo (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/trackedentity/DefaultTrackerAccessManager.java
muilpp: It doesn't feel right, but there's a job in previous versions run by a null user, which is why we need to skip this validation. However, it might be worth checking if that job needs to create a TE.

ameenhere: I agree it does not feel right. Which is why in later versions, there was some work done to fix the user to never being null. 
But in 2.39, I feel this may be safer. To follow the same approach as is done for canUpdate , canDelete, and canRead.

ameenhere:
> Once the cache is polluted, will the error occur even if the same user tries to register another TEI, or does it only happen when a different user attempts to register a TEI?

My assumption is, if the user was able to register one TEI, he will be able to register any number of TEIs (Since the cached entry of ownerOU falls into his scope). If the user was not able to register a TEI because of an incorrect ownerOU (polluted cache), he will not be able to register any TEIs until the polluted cache is cleared.

PR #19373, opened 2024-12-03
task: Rename eventProperty to eventField [DHIS2-18549]

muilpp:
As agreed on slack, I'm renaming `eventProperty` to `eventField`, to align with our current behavior in the API.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
@@ -133,7 +133,7 @@ public class DefaultEventChangeLogService implements EventChangeLogService {
   }
 
   private <T> void logIfChanged(
-      String fieldName,
+      String field,
       Function<Event, T> valueExtractor,
       Function<T, String> formatter,
       Event currentEvent,
@@ -148,27 +148,27 @@ public class DefaultEventChangeLogService implements EventChangeLogService {
 
       EventChangeLog eventChangeLog =
           new EventChangeLog(
-              event, null, fieldName, currentValue, newValue, changeLogType, new Date(), userName);
+              event, null, field, currentValue, newValue, changeLogType, new Date(), userName);
 
       hibernateEventChangeLogStore.addEventChangeLog(eventChangeLog);
     }
   }
 
   private ChangeLogType getChangeLogType(String oldValue, String newValue) {
-    if (isNewField(oldValue, newValue)) {
+    if (isFieldCreated(oldValue, newValue)) {
       return ChangeLogType.CREATE;
-    } else if (isUpdateField(oldValue, newValue)) {
+    } else if (isFieldUpdated(oldValue, newValue)) {
       return ChangeLogType.UPDATE;
     } else {
       return ChangeLogType.DELETE;
     }
   }
 
-  private boolean isNewField(String originalValue, String payloadValue) {
+  private boolean isFieldCreated(String originalValue, String payloadValue) {
     return originalValue == null && payloadValue != null;
   }
 
-  private boolean isUpdateField(String originalValue, String payloadValue) {
+  private boolean isFieldUpdated(String originalValue, String payloadValue) {
     return originalValue != null && payloadValue != null;
   }
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
@@ -99,7 +99,7 @@ public class DefaultEventChangeLogService implements EventChangeLogService {
 
   @Override
   @Transactional
-  public void addPropertyChangeLog(
+  public void addFieldChangeLog(
       @Nonnull Event currentEvent, @Nonnull Event event, @Nonnull String username) {
     logIfChanged(
         "occurredAt", Event::getOccurredDate, this::formatDate, currentEvent, event, username);
@@ -133,7 +133,7 @@ public class DefaultEventChangeLogService implements EventChangeLogService {
   }
 
   private <T> void logIfChanged(
-      String propertyName,
+      String fieldName,
       Function<Event, T> valueExtractor,
       Function<T, String> formatter,
       Event currentEvent,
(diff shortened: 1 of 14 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
```diff
@@ -133,7 +133,7 @@ public Set<Pair<String, Class<?>>> getFilterableFields() {
   }
 
   private <T> void logIfChanged(
-      String propertyName,
+      String fieldName,
```
ameenhere, line 136: Is simply "field" OK here?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
```diff
@@ -148,34 +148,27 @@ private <T> void logIfChanged(
...
-    } else if (isUpdateProperty(oldValue, newValue)) {
+    } else if (isUpdateField(oldValue, newValue)) {
       return ChangeLogType.UPDATE;
     } else {
       return ChangeLogType.DELETE;
     }
   }
 
-  private boolean isNewProperty(String originalValue, String payloadValue) {
+  private boolean isNewField(String originalValue, String payloadValue) {
```
ameenhere, line 167: Thoughts about

isFieldCreated()
OR
isFieldNew()

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
```diff
@@ -148,34 +148,27 @@ private <T> void logIfChanged(
...
     return originalValue == null && payloadValue != null;
   }
 
-  private boolean isUpdateProperty(String originalValue, String payloadValue) {
+  private boolean isUpdateField(String originalValue, String payloadValue) {
```
ameenhere, line 171: Thoughts about

isFieldUpdated()

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_33__Rename_event_change_log_property_to_field.sql
```diff
@@ -0,0 +1,3 @@
+-- DHIS2-18549
+
+alter table eventchangelog rename eventproperty to eventfield;
```
ameenhere, line 3: Since we have not released v42, can we include this change to the original migration script and remove this script?

Instead of first creating something else and then altering it?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventChangeLogService.java
teleivo: I don't think it even needs `field`. This is about whether this is a create, update or delete which we determine based on the values. How about `isCreate`, `isUpdate`, `isDelete`

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_33__Rename_event_change_log_property_to_field.sql
teleivo: What you are proposing would cause issues on play dev as the we would change the checksum of a migration that ran already. We would need to fix that manually AFAIK.

teleivo (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_33__Rename_event_change_log_property_to_field.sql
muilpp: this is not released, but it's been in master for a few days now.
I guess some people will have pulled from it already, if I use the original migration script, won't it create a conflict in Flyway for these people?

ameenhere: It would create some conflicts yes. For devs who already have applied the old migration, they will have to either manually fix it or start from an older snapshot. 
I doubt it would be an issue in play dev, as that always uses a db snapshot from a while ago. Possibly v41 db snapshot. We also have the flyway repair option in dhis config to make the fixes slightly easier. For master play dev, I think we do have flyway repair always set to true (I think). 

Main intention for this suggestion is to avoid these developmental evolution "within a release" forced onto production migrations. 
It is better we take some developer convenience hit, rather than having more steps in the migration process than needed.  Imagine the table already having millions of entries after the original migration, and in this second migration, we do something extra in that large table which could have been combined in the original migration.


Either way, my feeling is to fix (and communicate to devs how to fix) the conflicts for this. What do you think? I could be convinced otherwise :D

muilpp: I get the concern about adding more steps to the migration, but isn’t it pretty much transparent to the implementations? They’d just update to v42, and Flyway would handle the updates automatically. Whether it’s all in one script or split into a few shouldn’t really matter to them.

I see your point of view and would agree if the second script was performing an operation on the data, such as adding a new column to an existing index or modifying the data in some significant way. However, in this case, it's simply renaming a column. Even if the table contains millions of entries, the execution time should stay the same.

Would you agree?

ameenhere: To clarify, I was not suggesting to add a "Alter table rename..." to the original migration script. Rather change the create table script to have the correct column names the first time. 

> Whether it’s all in one script or split into a few shouldn’t really matter to them.

It will matter to them if we do split migrations would cost more time right? It may not be significant for this case though. Since the renaming column is potentially a constant time operation. But that constant time is still an additional time. If the original query, when creating the table the first time, if it had the right column names, this renaming wouldn't have been necessary. 

In this case though, assuming (and hoping) the additional time is very very small, let us go with this separate step 👍 😄

ameenhere (approved):

PR #19374, opened 2024-12-03
feat: limits for category combinations [DHIS2-18521]

jbee:
### Summary
Adds 3 configurations for new limits around categories that are validated on create and update.

* `metadata.categories.max_options`: The maximum number of category options in a single category
* `metadata.categories.max_per_combo`: The maximum number of categories per category combo
* `metadata.categories.max_combinations`: The maximum for the product of the number of options in a category combo

### Automatic Testing
New controller tests added for each of the validations.

### Manual Testing
Test limit max categories in a combo:

* goto maintenance app
* create a new category combo
* enter a name, select 6 categories and press save
* check a error message is shown that explains that a combo can only have 5 categories

Test limit max options per category:

* goto maintenance app
* edit an existing category
* select more than 31 options (move to right) and press save
* check a error message is shown that explains that a category can only have 31 options or less

Test limit max combinations:

* goto maintenance app
* find or create 3 categories with each have at least 8 (or more) options
* create a category combo with these 3 categories and press save
* check a error message is shown that explains that a category combo can only have more than 500 combinations


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/CategoryComboDeletionHandler.java
```diff
@@ -58,8 +59,10 @@ private DeletionVeto allowDeleteCategory(Category category) {
   }
 
   private void deleteCategoryOptionCombo(CategoryOptionCombo categoryOptionCombo) {
-    for (CategoryCombo categoryCombo : categoryService.getAllCategoryCombos()) {
```
jbee, line 61: FYI: I am not sure why we would loop over all CCs when there is just one linked to the COC which needs unlinking. Also I added a check if the un-linking isn't already done because if this was called as a consequence of deleting the CC it has and so the update should not be called again on the CC.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -69,6 +69,10 @@ public enum ErrorCode {
 
   E1122("Category option combo {0} cannot be associated with the default category combo"),
   E1125("Category option combo {0} contains options not associated with category combo {1}"),
+  E1126("Category combo {0} cannot combine more than {1} categories, but had: {2}"),
+  E1127("Category {0} cannot have more than {1} options, but had: {2} "),
+  E1128("Category combo {0} cannot have more than {1} combinations, bud requires: {2}"),
```
david-mackessy, line 74: maybe this message can mirror the other 2?
`..... cannot have more than x combinations, but had: y"

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/CategoryComboDeletionHandler.java
david-mackessy: If not needed, this is very nice. That could be very costly.
An existing/new test would confirm that.

david-mackessy (approved):
This sounds like a good idea, to have default limits for some items.
I'm wondering how this will play out in real use cases.
If a user tries to create more than the default, they will get the error message. Is the intention then that they should rethink how they are modelling things, but will most likely just update the config and get the higher limit they are looking for.
So it may just be an extra hoop they need to jump through (might be seen as annoying).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
jbee: It is worded differently because it is different. The combinations do not exist (yet) and when the validation fails we also don't create them, so I thought wording it "but had" is confusing if you know what is going on.

jbee:
> ...but will most likely just update the config and get the higher limit they are looking for.
> So it may just be an extra hoop they need to jump through (might be seen as annoying).

Yes, this usually should trigger a conversation. First to see if the use case really is what they were trying to model but if so they should ask the admins to increase the limit in the config. A good example is if you work with a extensive code standard that has thousands of codes that is a valid use case. But using the limits you can basically prevent mixing such category with another category by setting the max-options and max-combinations to same value.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/CategoryComboDeletionHandler.java
jbee: yeah, I only have my subjective observation from test and debugging but I found this because an deletion attempt would just spin my PC out of control so I had to kill the server, then step through it to see what was going on, which was when I found this. I then redid the deletion and it still was super slow but the server was at least responsive.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
david-mackessy: ok cool. There is a typo in `bud`. Should be `but`.

PR #19380, opened 2024-12-04
Make D2 dates related functions SqlBuilder aware (DHIS-16705)

luciano-fiandesio:
## Summary  
This PR enhances the integration of the `SqlBuilder` with the Antlr parser context and updates D2 functions to utilize the newly introduced `dateDifference` function.

## Changes  
- Introduced a `dateDifference` function within the `SqlBuilder` for improved date manipulation.
- Updated `D2*Between` functions to incorporate the `SqlBuilder` for uniform processing.
- Added unit tests for all modified expression classes.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
+++ b/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
@@ -312,12 +312,12 @@ public interface SqlBuilder {
    *
    * @param startDate the start date expression (can be a date literal or a column reference)
    * @param endDate the end date expression (can be a date literal or a column reference)
-   * @param datePart the unit of time to calculate the difference in (e.g., DAYS, MONTHS, YEARS)
+   * @param dateUnit the unit of time to calculate the difference in (e.g., DAYS, MONTHS, YEARS)
    * @return a String containing the database-specific SQL expression for calculating the date
    *     difference
-   * @see DatePart
+   * @see DateUnit
    */
-  String dateDifference(String startDate, String endDate, DatePart datePart);
+  String dateDifference(String startDate, String endDate, DateUnit dateUnit);
 
   // Statements
 
@@ -444,7 +444,7 @@ public interface SqlBuilder {
    */
   String dropCatalogIfExists();
 
-  enum DatePart {
+  enum DateUnit {
     DAYS,
     MONTHS,
     MINUTES,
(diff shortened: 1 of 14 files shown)


dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/SqlBuilder.java
```diff
@@ -422,4 +443,12 @@ public interface SqlBuilder {
    * @return a drop catalog if exists statement.
    */
   String dropCatalogIfExists();
+
+  enum DatePart {
```
maikelarabori, line 447: Call it DateUnit, maybe?

maikelarabori (approved):
Looks good 👍 
Minor suggestion added.
Thx!

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
```diff
@@ -126,7 +126,7 @@ public DefaultProgramIndicatorService(
     this.analyticsSqlCache = cacheProvider.createAnalyticsSqlCache();
     this.sqlBuilder = sqlBuilder;
 
-    this.programIndicatorItems = new ExpressionMapBuilder(sqlBuilder).getExpressionItemMap();
+    this.programIndicatorItems = new ExpressionMapBuilder().getExpressionItemMap();
```
gnespolino, line 129: I’m curious — was there a particular reason we opted not to use dependency injection in this case?

gnespolino (approved):
Good job, just a minor comment

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
jimgrace: I think it's OK. In `DefaultExpressionService` these maps are still generated in the service (for all the expression types that are not program indicators), so this is not consistent with that. But all this ANLR code will be ripped out when we transition to Jan B's new parser. (Did you know about that?) I'm OK with keeping it as a separate class or putting it back where it was.

jimgrace (approved):
Some code will also need to change at some point under `org.hisp.dhis.parser.expression` if Doris follows MySQL syntax rather than Postgres. Classes I've noticed in a quick review that will need changing are:
`FunctionContains`
`FunctionContainsItems`
`OperatorMathPower`

PR #19388, opened 2024-12-04
feat: Remove old TEA change log [DHIS2-18477]

muilpp:
This PR removes the old TEA change log endpoint, service, and store.
While working on this, I realized a few things:

- Initially, my idea was to keep the TEA change log delete method, similar to what we did for event change logs. This way, when a TE is deleted, we could also delete all related change logs from the new table and the old one. However, the problem with this approach is that if the implementation decides to drop the old table, the project will no longer compile because the Hibernate mapping won't find that table. While this compilation error could be resolved by using JDBC instead, it would lead to runtime errors, which doesn't improve the situation. For this reason, I decided to remove the method altogether, meaning no change logs will be deleted from the old TEA change log table anymore. The same applies to event change logs. These two tables are completely the responsibility of the implementation.
- I'm using `plainValue` for TEA instead of `value` to ensure that if the value is encrypted, the encrypted value doesn’t end up in the change log table.
- When manually merging two entities, the deduplication service was adding a `DELETE` change log entry on behalf of the duplicated TE. That TE was then deleted along with all its related change logs, including the one just created. I removed the step where that `DELETE` change log entry is created.
- `TrackedEntityAuditService` and `TrackedEntityAuditStore` should not be deprecated. This was probably due to confusion caused by the inconsistent naming between "change log" and "audit."
- `DefaultTrackedEntityAttributeValueService.updateTrackedEntityAttributeValue` was only used in tests, so I am removing it as well.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deprecated/audit/TrackedEntityAuditService.java
```diff
@@ -36,7 +36,6 @@
 /**
  * @author Abyot Asalefew Gizaw abyota@gmail.com
  */
-@Deprecated(since = "2.42", forRemoval = true)
```
enricocolasante, line 39: If this is not deprecate, should be moved outside `deprecated` package?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AuditController.java
```diff
@@ -246,76 +243,6 @@ public RootNode getAggregateDataValueAudit(
     return rootNode;
   }
 
-  /**
```
enricocolasante, line 249: I think we need to add this removal somewhere in the release-notes

enricocolasante (commented):
I always get confused when we talk about audit.
Is TrackedEntityAudit really used by someone? Maybe we can ask one more time, hopefully last time, to Morten to be sure it is needed.
It seems to me that it is just auditing when something it is DELETED and randomly auditing when something is READ.

muilpp:
> I always get confused when we talk about audit. Is TrackedEntityAudit really used by someone? Maybe we can ask one more time, hopefully last time, to Morten to be sure it is needed. It seems to me that it is just auditing when something it is DELETED and randomly auditing when something is READ.

I don't know if someone is using the audit feature, I'll contact Morten to clarify that. Until then, I think it should not be deprecated.

enricocolasante (approved):

ameenhere (approved):

PR #19393, opened 2024-12-05
chore: remove old data integrity report API

jbee:
Since the data integrity app now uses the new API the old API can be removed. 

They have been deprecated for removal since 2.38.

It would be quite cumbersome to patch the job parameters for jobs that use type `REPORT` so I kept the enum constant but it will now be an alias for `SUMMARY`.


david-mackessy (approved):
lovely

vietnguyen (approved):

PR #19397, opened 2024-12-06
fix: [DHIS2-18527] (2.40) Fix deleted relationship getting returned in event

ameenhere:
In event fetching API ( `api/tracker/event/{uid}` ) , when `includeRelationship` is set to true. The service fetches all relationships without respecting the `deleted` flag. 

This PR corrects this.

In collection event fetching API ( `api/tracker/events=event={uid}` ) this issue does not exist.


muilpp (approved):
just curious... it is possible to request deleted event relationships?
I know in other endpoints there's the parameter `includeDeleted`, not sure if it exists here too. If it does, I wonder if this logic would effectively disable that parameter.

ameenhere:
> just curious... it is possible to request deleted event relationships? I know in other endpoints there's the parameter `includeDeleted`, not sure if it exists here too. If it does, I wonder if this logic would effectively disable that parameter.

@muilpp Good point Marc! I had forgotten about that and did not find any references for includeDeleted in the relationships area. It could be that that parameter never respected in the relationship side of things so far. I'll create a separate jira issue to track/test that 👍  Thanks!

PR #19399, opened 2024-12-06
feat: Separate ChangeLogType between audit and changelog [DHIS2-18583]

muilpp:
This PR introduces a new enum to centralize values needed for tracker auditing while simplifying the enum used in change logs to include only `CREATE`, `UPDATE`, and `DELETE`. All usages of these enums have been updated accordingly.

Additionally, obsolete change log entities and their Hibernate mappings, which should have been removed in the [previous PR](https://github.com/dhis2/dhis2-core/pull/19388), are now cleaned up.
Change log deletion handlers have also been renamed for consistency, using either `EventChangeLog` or `TrackedEntityChangeLog` where appropriate.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/DefaultDataValueService.java
```diff
@@ -143,7 +143,7 @@ public boolean addDataValue(DataValue dataValue) {
       if (config.isEnabled(CHANGELOG_AGGREGATE)) {
         DataValueAudit dataValueAudit =
             new DataValueAudit(
-                dataValue, dataValue.getValue(), dataValue.getStoredBy(), ChangeLogType.CREATE);
+                dataValue, dataValue.getValue(), dataValue.getStoredBy(), TrackerAuditType.CREATE);
```
enricocolasante, line 146: We are using `TrackerAuditType` here but are we sure `DataValue` is part of tracker?
To be honest I am not sure what is this `DataValue` but `DataValueService` is in core and I cannot see it living in tracker module.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/datavalueset/DefaultDataValueSetService.java
```diff
@@ -933,7 +933,7 @@ private void saveDataValueDelete(
                 internalValue,
                 existingValue.getValue(),
                 context.getStoredBy(dataValue),
-                ChangeLogType.DELETE);
+                TrackerAuditType.DELETE);
```
enricocolasante, line 936: Here is the same, `DataValueSet` is not part of tracker so it seems a stretch that audit is prefixed with tracker

enricocolasante (commented):
I would probably rename `TrackerAuditType` to simply `AuditType` as it is used by entities outside tracker.

muilpp:
> I would probably rename `TrackerAuditType` to simply `AuditType` as it is used by entities outside tracker.

That was my initial thought, but there's already an `AuditType` enum used by the system audit. This enum defines the same values, but it also includes one called `SECURITY`, which is why I didn’t want to use it.

Based on [the docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/audit.html), this audit logs changes to aggregate data values (which are based on data elements) as well as read and search operations on tracked entities.

Therefore, I added the suffix "Tracker" to it. My intention was not to imply that it belongs to the tracker, but rather to emphasize that it audits tracker objects.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/audit/TrackerAuditType.java
```diff
@@ -25,34 +25,16 @@
...
-  private List<TrackedEntity> trackedEntities = new ArrayList<>();
-
-  private List<ChangeLogType> auditTypes = new ArrayList<>();
-
-  private Pager pager;
-
-  public boolean hasPager() {
-    return pager != null;
-  }
+public enum TrackerAuditType {
```
ameenhere, line 34: Let us find a generic name for this. AuditOperationType ? AuditOperation? Something else?

ameenhere (approved):
PR Looks Good! The only suggestion of removing the Tracker prefix from the enum name and making it generic.

enricocolasante (approved):

PR #19409, opened 2024-12-09
fix: make system-settings test time-zone independent

jbee:
System settings use local server zone when formatting the date string so one cannot expect a UTC (neutral) date and time therefore this is compared against a local date time which also uses the system default time zone.


david-mackessy (approved):

jimgrace (approved):
Excellent, thanks. I ran the test and that fixed it!

PR #19410, opened 2024-12-09
fix: login fallback error

Chisomchima:
Implements [DHIS2-18585](https://dhis2.atlassian.net/browse/DHIS2-18585)
---

#### Description:

This pull request addresses an issue with the login page where the dynamic base URL was not considering the URL context path (e.g., `/dev`, `/prod`). Without this fix, the application attempted to send the authentication request to the wrong endpoint (`/api/auth/login` instead of `/dev/api/auth/login`), leading to a failure when hosted under specific context paths like `/dev`.

#### Changes:

- **Base URL Calculation**:
  - Updated the `getBaseUrl()` function to account for the context path (e.g., `/dev`, `/dev-2-41`, etc.) in the URL.
  - The context path is now correctly appended to the API endpoint, ensuring that requests like `POST /api/auth/login` are sent to `/dev/api/auth/login` when the app is hosted under `/dev`.
  - If the page URL ends with `login.html`, that part is correctly stripped off while preserving the context path.

#### Problem Addressed:
- Before this change, when the app was hosted under a context path (e.g., `/dev/login.html`), the login form submitted requests to `http://localhost:8080/api/auth/login` instead of the expected `/dev/api/auth/login`.
- This caused authentication requests to fail for applications hosted in a subdirectory or versioned environment.

---

[DHIS2-18585]: https://dhis2.atlassian.net/browse/DHIS2-18585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-web-api/src/main/resources/org/hisp/dhis/webapi/servlet/login.html
```diff
@@ -321,9 +325,26 @@ <h1>Log in</h1>
     </div>
 
     <script>
+      // This function will dynamically determine the base URL of the current page
+      function getBaseUrl() {
```
Chisomchima, line 329: I added the `getBaseUrl ` function to dynamically get base url depending on the version, If you think there's a better way to do this from the server, I'm open to that conversation

Chisomchima: I have used a relative path instead of the `getBaseUrl ` function

tomzemp (approved):
lgtm 👍

PR #19415, opened 2024-12-10
fix: Unable to save outbound sms during async operation [DHIS2-18586]

zubaira:
[DHIS2-18586](https://dhis2.atlassian.net/browse/DHIS2-18586)

Async operation tries to save outboundSMS and it fails due to currentUser not set. This PR injects currentUser using AuthenticationService.

[DHIS2-18586]: https://dhis2.atlassian.net/browse/DHIS2-18586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsMessageSender.java
```diff
@@ -292,7 +286,12 @@ private void handleResponse(OutboundMessageResponse status, OutboundSms sms) {
       sms.setStatus(OutboundSmsStatus.FAILED);
     }
 
-    outboundSmsService.save(sms);
+    try {
```
enricocolasante, line 289: Who is calling this method?
Why the user is not present?
This doesn't seem to be running in another thread, if I am wrong, who is starting the thread?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsMessageSender.java
```diff
@@ -122,7 +123,7 @@ public OutboundMessageResponse sendMessage(
   public Future<OutboundMessageResponse> sendMessageAsync(
       String subject, String text, String footer, User sender, Set<User> users, boolean forceSend) {
     OutboundMessageResponse response = sendMessage(subject, text, footer, sender, users, forceSend);
-    return new AsyncResult<>(response);
+    return CompletableFuture.completedFuture(response);
```
enricocolasante, line 126: This is doing nothing, it is not running sendMessage in a different thread and it is not running it in parallel.
Would it make sense to just return the response?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/sms/config/SmsMessageSender.java
zubaira: https://github.com/dhis2/dhis2-core/blob/fd2b4211e66c501e1360bdc6bae7edb3341b2a4a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/message/DefaultMessageService.java#L491

ameenhere (approved):

enricocolasante (approved):

PR #19423, opened 2024-12-10
fix: Validate data sharing on program stage [DHIS2-17594]

muilpp:
When accessing `tracker/trackedEntities/{uid}`, data sharing is not applied to the program stage, which may result in events that are inaccessible to the current user being returned.

A fix has been implemented in the tracked entity service, similar to the approach used for enrollments and events, to address this issue.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
@@ -399,8 +399,8 @@ class DefaultTrackedEntityService implements TrackedEntityService {
                   e.getEvents().stream()
                       .filter(
                           event ->
-                              trackerAccessManager.canRead(user, event, false).isEmpty()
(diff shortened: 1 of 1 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -397,7 +397,10 @@ private Set<Enrollment> getEnrollments(
             e -> {
               Set<Event> filteredEvents =
                   e.getEvents().stream()
-                      .filter(event -> includeDeleted || !event.isDeleted())
+                      .filter(
+                          event ->
+                              trackerAccessManager.canRead(user, event, false).isEmpty()
```
ameenhere, line 402: I am thinking of the performance implication here. 
Can we atleast move this as the second clause in this condition? 
Meaning swapping the conditions on either sides of &&.

muilpp: makes sense, thanks

ameenhere (approved):

PR #19430, opened 2024-12-10
feat: Create note endpoints in tracker [DHIS2-17579]

enricocolasante:
Added endpoint to create a `note` for `enrollments` and `events`:
- `POST tracker/enrollments/{uid}/note`
- `POST tracker/events/{uid}/note`

A new service and a new store using JDBC are created.
The validation applied are:
- Verify that event/enrollment exists
- Verify that user has read access to event/enrollment
- Verify that field `value` for the `note` is not empty
- Verify that a `note` with the same `uid` does not already exist

***Inconsistency between the 2 endpoints***
To check that the event/enrollment exists and that the user has all the right permissions we are using services (`EnrollmentService.getEnrollment(uid)` and `EventService.getEvent(uid)`) but they do behave differently when an entity without the right permission is accessed:
- `EnrollmentService` throws a 403
- `EventService` throws a 404

This is part of a biggest discussion and it will not be fixed here

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/note/hibernate/Note.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/note/hibernate/Note.hbm.xml
@@ -11,7 +11,13 @@
     <id name="id" column="noteid">
       <generator class="native" />
     </id>
-    &identifiableProperties;
+
+    <property name="uid" column="uid" unique="true" not-null="true" />
+
+    <property name="created" type="timestamp" not-null="true" update="false" />
+
+    <many-to-one name="lastUpdatedBy" class="org.hisp.dhis.user.User"
+                 column="lastupdatedby" foreign-key="fk_lastupdateby_userid" />
 
     <property name="noteText" column="notetext" type="text" />
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -132,7 +132,6 @@ class JdbcEventStore {
        n.created as note_created,\
        n.creator as note_creator,\
        n.uid as note_uid,\
-       n.lastupdated as note_lastupdated,\
        userinfo.userinfoid as note_user_id,\
        userinfo.code as note_user_code,\
        userinfo.uid as note_user_uid,\
@@ -466,8 +465,6 @@ class JdbcEventStore {
                 noteLastUpdatedBy.setSurname(resultSet.getString("note_user_surname"));
                 note.setLastUpdatedBy(noteLastUpdatedBy);
               }
-
-              note.setLastUpdated(resultSet.getTimestamp("note_lastupdated"));
 
               event.getNotes().add(note);
               notes.add(resultSet.getString("note_id"));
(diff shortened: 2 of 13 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/note/hibernate/Note.hbm.xml
```diff
@@ -11,7 +11,13 @@
     <id name="id" column="noteid">
       <generator class="native" />
     </id>
-    &identifiableProperties;
+
```
muilpp, line 14: why do we need this mapping if we use JDBC?

muilpp:
Just for me to understand, why don't we use the importer to create notes as we do for other tracker objects?

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/note/hibernate/Note.hbm.xml
enricocolasante: We still need it because tracker importer is adding notes that are defined in the enrollments/events and we are persisting using hibernate there.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/JdbcNoteStore.java
```diff
@@ -0,0 +1,121 @@
...
+    jdbcTemplate.update(sql, Map.of("event", event.getValue(), "noteId", noteId));
+  }
+
+  boolean exists(@Nonnull UID note) {
+    Integer count =
+        jdbcTemplate.queryForObject(
+            "select count(1) from note where uid = :uid",
+            Map.of("uid", note.getValue()),
+            Integer.class);
+    return count == null || count > 0;
```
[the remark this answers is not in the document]
enricocolasante: I wrongly inverted the logic here.
Fixed

enricocolasante:
> @enricocolasante, just for me to understand, why don't we use the importer to create notes as we do for other tracker objects?

We agreed with frontend that would make sense to have an easier way for them to create a single note.
Using the importer for that adds a lot of unnecessary data that they should pass together with the note as we do not allow patch, so they would need to send a payload with the whole enrollment/event just to add a note.
And as well, it is not so good for performance, because we need to validate all that unnecessary data anyway

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -132,7 +132,6 @@ class JdbcEventStore {
        n.created as note_created,\
        n.creator as note_creator,\
        n.uid as note_uid,\
-       n.lastupdated as note_lastupdated,\
```
ameenhere, line 135: Are notes immutable? Is that why this is done?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_34__Create_sequence_for_note_table.sql
```diff
@@ -0,0 +1,5 @@
+create sequence if not exists note_id_sequence;
```
ameenhere, line 1: we usually name sequences as "tablename_sequence"
So maybe this can be "note_sequence" ?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
enricocolasante: Yes, they are immutable

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_34__Create_sequence_for_note_table.sql
enricocolasante: Done

ameenhere (approved):

PR #19438, opened 2024-12-11
fix: Nepali calendar support

abyot:
- ISO period names should match Nepali calendar. Baisakh 2081 is 208101 not 202404
- Fix to weekly periods start and end dates


dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/impl/NepaliCalendar.java
```diff
@@ -144,25 +145,36 @@ public int isoWeek(DateTimeUnit dateTimeUnit) {
 
   @Override
   public int week(DateTimeUnit dateTimeUnit) {
-    return isoWeek(dateTimeUnit);
+    /*
+     * https://en.m.wikipedia.org/wiki/ISO_week_date
+     */
+    int week = (10 + getDayOfYear(dateTimeUnit) - isoWeekday(dateTimeUnit)) / 7;
```
maikelarabori, line 151: Could you add constants to the integers here?
Also, a Javadoc would be nice. The Wikipedia website can be left as a reference.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/impl/NepaliCalendar.java
```diff
@@ -144,25 +145,36 @@ public int isoWeek(DateTimeUnit dateTimeUnit) {
...
+    return week;
   }
 
   @Override
   public int isoWeekday(DateTimeUnit dateTimeUnit) {
-    DateTime dateTime =
-        toIso(dateTimeUnit).toJodaDateTime(ISOChronology.getInstance(DateTimeZone.getDefault()));
-    return dateTime.getDayOfWeek();
+    /*
+     * calculating week day from calendar dateTimeUnit is best managed
```
maikelarabori, line 163:
```suggestion
     * Calculating week day from calendar dateTimeUnit is best managed
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/impl/NepaliCalendar.java
```diff
@@ -475,6 +500,39 @@ private DateInterval toDayIsoInterval(DateTimeUnit dateTimeUnit, int offset, int
   private static final Map<Integer, int[]> CONVERSION_MAP = new HashMap<>();
```
maikelarabori, line 500: Can you add a comment, so we know until which year this is valid?
I belive that at some point we need to add more elements to the map...?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/period/PeriodType.java
```diff
@@ -570,7 +570,7 @@ public Period createPeriod(DateInterval dateInterval) {
     final DateTimeUnit from = cal.toIso(dateInterval.getFrom());
     final DateTimeUnit to = cal.toIso(dateInterval.getTo());
 
-    return new Period(this, from.toJdkDate(), to.toJdkDate(), getIsoDate(from));
+    return new Period(this, from.toJdkDate(), to.toJdkDate(), getIsoDate(dateInterval.getFrom()));
```
maikelarabori, line 573: Is this change intentional? I'm afraid it could break existing code relying on the old way...

abyot: Yes, intentional. That is why the "old way" was returning Gregorian ISO names for non-gregorian calendars. See the problem description in the jira ticket.

maikelarabori:
Hi @abyot, would you have any updates on this PR? Did you have you have time to review the comments?

Thanks!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/impl/NepaliCalendar.java
abyot: yes, we need to update the mapping table when we get new information.

abyot:
> Hi @abyot, would you have any updates on this PR? Did you have any time to review the comments?
> 
> Thanks!

now updated. thanks.

maikelarabori (approved):

gnespolino (approved):

maikelarabori:
Thanks for the update @abyot.
Please, do not forget to apply the same changes to branches 41 and 42. 👍

PR #19447, opened 2024-12-11
feat: add optionset to dataItems endpoint [DHIS2-18451]

gnespolino:
This pull request adds OptionSet items in `dataItems` endpoint


maikelarabori (commented):
Looks good @gnespolino.
Now that we have e2e tests, it would be nice to add a few more covering different queries and filters.

gnespolino:
> Looks good @gnespolino. Now that we have e2e tests, it would be nice to add a few more covering different queries and filters.

adding an E2E for this is something bigger than the ticket itself. So either just have an integration test, that I might add to this PR, or let's do the e2e in a dedicated ticket.

maikelarabori (approved):

gnespolino:
Added one e2e in the end

PR #19470, opened 2024-12-15
fix: Simplify created by display name for event analytics [DHIS2-18417]

larshelge:
In event and enrollment analytics table statements, we have 64 lines of SQL code to produce two fields - `createdbydisplayname` and `lastupdatedbydisplayname`. IMHO this is too much complexity and code to produce two fields. It is hard to reason about, makes the update slower, and makes it hard to port to Doris/ClickHouse. 

I have been doing some research on event/tracker DHIS 2 instances, and I could see that typically either `createdbyuserinfo` is not set at all, or is fully set with all properties (first name, username, username). I have not seen a partial value yet. If some implementations have a partial `userinfo` object for enrollments/events, we should fix that once with a SQL statement, and/or fix the issue with tracker that causes this in the first place. That is better than "fixing" it every time analytics table update runs.


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/DisplayNameUtils.java
```diff
@@ -64,83 +46,9 @@ public static String getDisplayName(
     String surname = extractJsonValue(sqlBuilder, tablePrefix, originColumn, "surname");
     String firstName = extractJsonValue(sqlBuilder, tablePrefix, originColumn, "firstName");
     String username = extractJsonValue(sqlBuilder, tablePrefix, originColumn, "username");
+    String expression = sqlBuilder.concat(surname, "', '", firstName, "' ('", username, "')'");
```
maikelarabori, line 49: I agree that the logic for getting the display name is not great, but it prints the expected values from the user's perspective.

With this change, if the surname is blank, we will end up with `, first name (username)`, and if the username is blank, with `surname, first name ()`, etc. It might look like a bug.

maikelarabori (approved):
Looks good, I just left one comment for review.
Thx!

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/util/DisplayNameUtils.java
larshelge: Yes that is correct. However, I have been doing some research on event/tracker DHIS 2 instances, and I could see that typically either `createdbyuserinfo` is not set at all, or is fully set with all properties (first name, username, username). I have not seen a partial value yet. If some implementations have a partial `userinfo` object for enrollments/events, we should fix that once with a SQL statement, and/or fix the issue with tracker that causes this in the first place. That is better than "fixing" it every time analytics table update runs.

IMHO we are covering up a data integrity issue which should be fixed at the source, not be corrected every time analytics tables are updated.

janhenrikoverland: In the case where it is not set at all the api now prints `, ()`

PR #19472, opened 2024-12-15
fix: Use SqlBuilder for if-then-else logic [DHIS2-18417]

larshelge:
Moves the conditional if-then-else logic behind `SqlBuilder` to allow for differences across analytics databases.


gnespolino (commented):
Don't we have multiple choices "case"? Would be nice supporting those as well

larshelge:
We have multi-if statements in some places but I plan to create another overloaded method for that.

gnespolino (approved):

maikelarabori (approved):

PR #19492, opened 2024-12-16
feat: Support option set in DV object [DHIS2-18370]

maikelarabori:
Adds a new object (`OptionSetItem`) to the Visualization object, so users can save the option set along with the option set items they wish to be saved as part of their Visualization. The aggregation type is also required as the items can be aggregated or disaggregated. It depends on the users' choice.


gnespolino (commented):
would be nice to have e2e test for this

gnespolino (approved):

PR #19512, opened 2024-12-17
feat: Sort data elements and event fields as single unit [DHIS2-18012]

muilpp:
When it comes to sorting event change logs, we need to treat data elements and event fields as a single entity, filtering stays the same.
Sort data elements by form name instead of UID, as that's how the UI shows them.
Still need to take care of translations, I'll do that in a separate ticket.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
@@ -64,14 +64,14 @@ public class HibernateEventChangeLogStore {
 
   /**
    * Event change logs can be ordered by given fields which correspond to fields on {@link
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -57,23 +57,21 @@ public class HibernateEventChangeLogStore {
   private static final String COLUMN_CHANGELOG_USER = "ecl.createdByUsername";
   private static final String COLUMN_CHANGELOG_DATA_ELEMENT = "d.uid";
   private static final String COLUMN_CHANGELOG_FIELD = "ecl.eventField";
-
+  private static final String ORDER_DATA_ITEM_EXPRESSION =
```
enricocolasante, line 60: I think we can remove `dataItem` word from the code and use `change` instead

enricocolasante (approved):

ameenhere (approved):

PR #19521, opened 2024-12-18
feat: Add option to exclude ALL auth from check

david-mackessy:
This change allows us to exclude the `ALL` authority from an endpoint authority check.
The use for this is probably very low right now, but given it's such a small change and it has no impact on the current behaviour, it seems like a nice option to be able to choose. It's a good topic for conversation at least.

We currently allow the use of `ALL` to do anything in the system. A recent conversation during a merge demo brought up the topic of the `ALL` authority. Given that users know about the `ALL` authority, it is reported that a lot of users are granted the `ALL`  so they can just use the system without hassle and use features they possibly should not be allowed to.

There are no endpoints that use this feature in this change, so no tests can be added yet.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/RequiresAuthority.java
```diff
@@ -37,9 +37,11 @@
  * Annotation that takes in any number of {@link Authorities}. This allows us to check if a {@link
  * org.hisp.dhis.user.User} has any of the {@link Authorities} passed in.
  *
- * <p>{@link Authorities#ALL} is automatically added to the check, as having this Authority allows
- * access to all methods by default. No need to pass {@link Authorities#ALL} in the arguments. See
- * {@link AuthorityInterceptor}.
+ * <p>{@link Authorities#ALL} is automatically added to the check by default, as having this
```
jbee, line 40: I think this  documentation is too much about how it works and to little about what the goal is on a semantic level and why.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/RequiresAuthority.java
```diff
@@ -50,4 +52,6 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface RequiresAuthority {
   Authorities[] anyOf();
+
+  boolean excludeAllAuth() default false;
```
jbee, line 56: This should have javadoc explaining what it means to use `true`. I am actually not really sure even after reading some code. Is this basically deactivating the `ALL` superuser for being specially privileged and allowed to do anything?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/RequiresAuthority.java
david-mackessy: I've updated the javadocs in this file. Let me know if it's any better/clearer.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/security/RequiresAuthority.java
david-mackessy: I've updated the javadocs in this file. Let me know if it's any better/clearer.

david-mackessy:
discussed at platform backend team meeting, feedback was that we might tackle the ALL auth in the future (how it's used system-wide), so best not to introduce divergences in how that works for now.

PR #19551, opened 2024-12-20
fix: Always add default order as last order param in tracker exporters [DHIS2-18659]

enricocolasante:
Always add default order as a last order parameter to have a consistent order between results.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
@@ -279,7 +279,7 @@ class HibernateEnrollmentStore extends SoftDeleteHibernateObjectStore<Enrollment
       orderJoiner.add(
           order.getField() + " " + (order.getDirection().isAscending() ? "asc" : "desc"));
     }
-    return " order by " + orderJoiner;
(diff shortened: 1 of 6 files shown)


muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
```diff
@@ -279,7 +279,7 @@ private static String orderBy(List<Order> orders) {
       orderJoiner.add(
           order.getField() + " " + (order.getDirection().isAscending() ? "asc" : "desc"));
     }
-    return " order by " + orderJoiner;
+    return " order by " + orderJoiner + ", " + DEFAULT_ORDER;
```
ameenhere, line 282: I guess there is no consequence if the same DEFAULT_ORDER is also part of the orderJoiner?

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
enricocolasante: Default order, aka entity id, it is not part of orderable fields, so it cannot ever be part of the orderJoiner here

PR #19553, opened 2024-12-21
fix: [DHIS2-18690] Re-add TESchemaDescriptor to avoid in-memory query planner (2.41)

ameenhere:
Calle had complaned a simple enrollment api caused an OutOfMemory heap space error.  On investigating, the culprit was the removal of `TrackedEntityInstanceSchemaDescriptor` removal. This meant the SchemaService would identify TrackedEntity as a "non persisted" entity and cause the queryPlanner in DefaultQueryService to use npQuery (non-persisted in-memory query) instead. 

This meant, whenever enrollment was done, as part of `prepareCaches` it loaded all the TEs  (8 million in Calle's case) into the memory and did an in-memory filtering for the single TEI it was interested in. Thereby causing the OutOfMemory issue.

This PR simply re-introduces the SchemaDescriptor back as TrackedEntitySchemaDescriptor. This is rightfully removed in v42 and not required there. However the old API still exists in v41 and relied on the existence of the SchemaDescriptor. Instead of changing any other logics, re-introduced it back which fixes the issue.


ameenhere:
Converting the PR to draft. @zubaira  Rightly pointed out that ProgramInstance and ProgramStageInstance also had SchemaDescriptor which was removed in v41. Need to investigate this before the PR is considered complete.

enricocolasante (approved):

PR #19574, opened 2024-12-30
feat: password validation regexp in settings

netroms:
## Summary
This introduces a new system setting called `passwordValidationPattern`, this is a Regexp version of the internal password rule set, to be used in clients, most important the standard web UI.
Since the password validation is configurable in the back-end, (currently only max and min password length), the rule aka. Regexp needs to be provided dynamically from the back-end.
This new property along with the min and max length properties are also exposed in the `LoginConfigController`, for easy access in non logged in environments, like the signup page.

### Fix
This PR also fixes an issue with the current back-end password rules not checking for a lower-case character present, this is enforced in the UI, but not in the back-end. To align this, a new password rule was added, called `LowerCasePatternValidationRule.java`. Since this rule was expected and always enforced in the UI, I see this as safe to add.

### Automatic tests
SystemSettingsTest.java
UserTest.java
LoginConfigControllerTest.java
UserAccountControllerTest.java
### Manual test
1. Call API endpoint /systemSettings
2. Observe there is a `passwordValidationPattern` property present and that the value is: `^(?=.*[A-Z])(?=.*[a-z])(?=.*\d)(?=.*[\W_])[A-Za-z\d\W_]{8,72}$`
3. Call API endpoint /loginConfig
4. Observe the same property is present along with `minPasswordLength` and `maxPasswordLength`

#### JIRA: [DHIS2-18574](https://dhis2.atlassian.net/browse/DHIS2-18574)

[DHIS2-18574]: https://dhis2.atlassian.net/browse/DHIS2-18574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
```diff
@@ -745,4 +745,15 @@ default boolean isHideUnapprovedDataInAnalytics() {
...
   }
+
+  /**
+   * @return A regex pattern string that enforces the current password validation rules
+   */
+  default String getPasswordValidationPattern() {
+    return asString(
+        "passwordValidationPattern",
+        String.format(
+            "^(?=.*[A-Z])(?=.*[a-z])(?=.*\\d)(?=.*[\\W_])[A-Za-z\\d\\W_]{%d,%d}$",
```
jbee, line 756: That regex seems too specific. With this a upper case letter must always be before a lower case one, both before a digit and all of them before a non-word character. All of that must be followed by the min-max number of letters, gitis, and non-word characters. That seems illogical. Why use `.*` in several places allowing anything but then at the end be so specific? I would assume `\W` basically excludes what you then included again with `A-Za-z\\d` so effectively that will be very close or similar to `.`. 

I think what you wanted to say is something entirely different where it is min-max number of characters for which at least one is a lower letter, one an upper letter, one a digit, and one a special character.

jbee (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
jbee: For example: `+1aaaaAA00` does not match. That has 8 letters but as I explained you basically ask for 4+min-max. But considering that, even `+1aaaaAA0012345676` would not match.

jbee (approved):

PR #19601, opened 2025-01-07
feat: PI Disaggregation objects [DHIS2-18745]

jimgrace:
Creating objects for PI Disaggregation -- work in pogress

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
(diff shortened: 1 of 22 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
```diff
@@ -0,0 +1,104 @@
...
+ * option combination to be returned as a program indicator dimension.
+ *
+ * @author Jim Grace
+ */
+@NoArgsConstructor
+@JacksonXmlRootElement(localName = "optionMappings", namespace = DxfNamespaces.DXF_2_0)
+public class ProgramCategoryOptionMapping implements EmbeddedObject {
+
+  /** The database internal identifier for this Object. */
+  private int id;
```
vietnguyen, line 51: The `id` column is mapped as `bigint` in database so I think it should be `long` java type.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/JobScheduler.java
```diff
@@ -76,7 +76,7 @@ public class JobScheduler implements Runnable, JobRunner {
    *   <li>Leaves as much time as possible otherwise to get through the work of triggering jobs
    * </ul>
    */
-  private static final int LOOP_SECONDS = 20;
+  private static final int LOOP_SECONDS = 20000000;
```
jbee, line 79: ???

jimgrace: @jbee Ha, that's temporary! When I'm debugging I sometimes find it very helpful to watch my Postgres log in real time to see what's happening in the database. (I log all statements all the time and rotate the log by day of week so it doesn't fill up my disk.) When I don't need the job scheduler (which is nearly all the time), I change this value so it's not dumping in a bunch of queries three times a minute. It make the log *much* easier to follow!

PR #19611, opened 2025-01-07
fix: user query parameter must include username [DHIS2-18748]

jbee:
### Summary
When using the `query` parameter with `User` this should not run the standard implementation. 
Therefore `query` needs to be considered a "special filter" and excluded for users.

This got overlooked when reworking the special filters in #19046 since the `query` parameter basically has a name clash.

### Manual Testing 
Using it to find `system` user by `username` once with other special filters and once without

* `/api/users?userOrgUnits=true&includeChildren=true&query=system`
* `/api/users?query=system`
* check both URLs result in a list containing the system user (in SL this is the only match)

### Automatic Testing
Added a controller test scenario that searches with `query=system` to match on the `username`.


vietnguyen (approved):

jbee:
> Can a test be added for this fix?

Added now

PR #19612, opened 2025-01-07
feat: adds sortable schema property [DHIS2-18749]

jbee:
### Summary
Adds a `sortable` property to schema `Property` that indicates if a property can be used in the `order` parameter.

By default only simple non-collections properties are sortable. Exceptions can be made by annotating the property getter with `@Sortable(true)` or `@Sortable(false)`.

As far as I could tell we do not have any actual validation for the `order` parameter apart from simply ignoring collections and non-simple properties should they be present. Therefore that logic was used as the fall-back if no `@Sortable` annotation is present.

### Cleanup
IDE supported refactoring to 

* use `@Setter` and some `@Getter` instead of explicitly stating the method
* use `final`
* add nullable annotations to overridden method

Manual changes

* mark some computed properties as `access = READ_ONLY` for jackson

### Manual Testing
 check for example `/api/schemas/dataElement` and look at the `properties` array

Properties now have a `sortable` attribute, for example `name` looks like
```JSON
{
      "klass": "java.lang.String",
      "propertyType": "TEXT",
      "name": "name",
      "fieldName": "name",
      "persisted": true,
      "description": "The name of this Object. Required and unique.",
      "attribute": true,
      "simple": true,
      "collection": false,
      "ordered": false,
      "owner": true,
      "identifiableObject": false,
      "nameableObject": false,
      "embeddedObject": false,
      "analyticalObject": false,
      "readable": true,
      "writable": true,
      "unique": true,
      "required": true,
      "length": 230,
      "max": 230.0,
      "min": 1.0,
      "manyToMany": false,
      "oneToOne": false,
      "manyToOne": false,
      "oneToMany": false,
      "translatable": true,
      "translationKey": "NAME",
      "i18nTranslationKey": "name",
      "gistPreferences": {
        "included": "AUTO",
        "transformation": "AUTO"
      },
      "sortable": true,
      "propertyTransformer": false
    }
```

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/schema/Property.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/schema/Property.java
@@ -38,10 +38,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import lombok.Getter;
+import lombok.Setter;
 import org.hisp.dhis.common.DxfNamespaces;
 import org.hisp.dhis.common.EmbeddedObject;
 import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.common.NameableObject;
+import org.hisp.dhis.common.Sortable;
 import org.hisp.dhis.hibernate.HibernateProxyUtils;
 import org.springframework.core.Ordered;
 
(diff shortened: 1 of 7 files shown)


david-mackessy (approved):
can a test be added to check for the new field present?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/schema/Property.java
```diff
@@ -259,144 +263,83 @@ public PropertyType getPropertyType() {
...
     return collection;
   }
 
-  public void setCollection(boolean collection) {
-    this.collection = collection;
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
+  public boolean isSortable() {
+    Sortable sortable = getterMethod == null ? null : getterMethod.getAnnotation(Sortable.class);
+    return sortable != null ? sortable.value() : !isCollection() && isSimple();
```
vietnguyen, line 342: I think we need to check if the property is `persisted` , meaning it is mapped to a database column.
For example: a class may not have the property `shortName` but it is still inherit the property from base class.

jbee: Good point, but I will have to find a way to not exclude a computed property like `displayName` that way since it isn't persisted but still possible to sort by.

jbee:
> can a test be added to check for the new field present?

Added a test now

dhis-2/dhis-api/src/main/java/org/hisp/dhis/schema/Property.java
jbee: I use `persisted` now where the default is that it has to be persisted. To make properties like `displayName` still be sortable they must be annotated with `@Sortable(whenPersisted=false)`. This is not perfect but covers most of the cases.

vietnguyen (approved):

PR #19621, opened 2025-01-08
feat: include most recent job progress in post condition error message [DHIS2-18751]

jbee:
### Summary
While the plan was just to add a better post-condition logging I found that executing jobs manually was broken and so I did the post spring 6 cleanup in this context as well.

### Post-condition Logging
* added the last process, stage and item description to the exception message
* added a new `failWithPostCondition` flag to the `TEST` job so a post-condition failure could be simulated

### Scheduling with Spring 6
* remove the quasi duplicate code path added during spring 6 migration
* removed the test vs prod "helper" workaround (it was identical at this point)
* added a new service `JobExecutionService` for the execution of one-off jobs to have a layer for right TX handling
* use `getByUidNoAcl` whenever a job is loaded within the scheduling so make sure ACL isn't in the way
* initialize the `lastAlive` counter with zero so the first update happens immediately giving faster feedback for the observer that the job is actually running
* replaced `getSingleResult()` with either `getSingleResult(Query)` or `getResultStream().findFirst().orElse(null)`
* removed the `NotFoundException` from being thrown when calling `executeOnceNow` (one-off jobs)

### Automatic Testing
There are plenty of tests covering scheduling. Also once must check that the e2e tests run in about 10-15min not 90min. 

### Manual Testing
* Create a job of type `TEST`
* set the `failWithPostCondition` to `true`
* run the job manually or wait for its scheduled time to occur
* check the logs for the exception

Example:
```
java.util.concurrent.CancellationException: Non-null post-condition failed after: Test job
  => Stage 1
	at org.hisp.dhis.scheduling.RecordingJobProgress.cancellationException(RecordingJobProgress.java:217) ~[classes/:?]
	at org.hisp.dhis.scheduling.RecordingJobProgress.nonNullStagePostCondition(RecordingJobProgress.java:307) ~[classes/:?]
```


larshelge (approved):

david-mackessy (approved):
this should be very helpful 👍

PR #19630, opened 2025-01-09
test: Centralize creation of org units in test

larshelge:
This PR centralizes instantiation of org units through the `createOrganisationUnit` in `TestBase`. This makes the test more realistic (by setting UID, last updated, path, etc) and makes it easier to change later.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
@@ -807,6 +807,14 @@ public class OrganisationUnit extends BaseDimensionalItemObject
     return hierarchyLevel;
   }
 
+  /**
+   * Note that the {@code path} is mapped with the "property access" mode. This method is for unit
+   * testing purposes only.
+   */
+  public void updatePath() {
+    setPath(getPath());
+  }
+
   /** Do not set directly. */
   public void setHierarchyLevel(Integer hierarchyLevel) {
     this.hierarchyLevel = hierarchyLevel;
--- a/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/query/DefaultQueryServiceTest.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/query/DefaultQueryServiceTest.java
@@ -35,7 +35,6 @@ import static org.mockito.Mockito.when;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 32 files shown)


dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/query/DefaultQueryServiceTest.java
```diff
@@ -96,7 +95,7 @@ private List<OrganisationUnit> createOrgUnits(int size) {
 
     List<OrganisationUnit> result = new ArrayList<>();
     for (int i = 0; i < size; i++) {
-      result.add(createOrganisationUnit(RandomStringUtils.randomAlphabetic(1)));
+      result.add(createOrganisationUnit((char) (i + 65)));
```
jbee, line 98: You can write `i + 'A'` for more clarity

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
```diff
@@ -807,6 +807,14 @@ public Integer getHierarchyLevel() {
     return hierarchyLevel;
   }
 
+  /**
+   * Note that the {@code path} is mapped with the "property access" mode. This method is for unit
+   * testing purposes only.
+   */
+  public void updatePath() {
```
jbee, line 814: FYI: We recently had a discussion about computation of the `path` and outdated path in DB - I guess this is that same issue just in the case of test setup.

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
larshelge: Yes this is a part of a proposed fix, more info coming.

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/query/DefaultQueryServiceTest.java
larshelge: Thanks.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/organisationunit/OrganisationUnit.java
david-mackessy: Adding something only needed for tests into PRD code is not a good idea. We're opening up a door and inviting future issues.
Is there some way to get around this or prohibit using it in PRD code?

PR #19377, opened 2024-12-04
chore: get one trackedEntity via path and query param is identical DHIS2-18541

teleivo:
as already done for `/tracker/events` in https://github.com/dhis2/dhis2-core/pull/18943

PR #19633, opened 2025-01-10
chore: simplify TrackedEntityService.getTrackedEntity(UID, ...) logic DHIS2-18541

teleivo:
This change is necessary for https://github.com/dhis2/dhis2-core/pull/19377 and should make changes in behavior easier to see/discuss.

We have two methods for getting a single TE. They were implemented independently with convoluted logic. Let one reuse the other. Separate fetching, ACL, mapping to make the flow easier to follow. Move the TETAV logic in one place. This is the only way we can ensure the behavior is the same across different code paths.

Add an `assertContainsOnly` variant that lets us map both collection before making the actual assertion. This makes the assertion error easier to read.

```java
assertContainsOnly(
    Set.of(tetavA, tetavB),
    trackedEntity.getTrackedEntityAttributeValues(),
    TrackedEntityAttributeValue::getValue);
```

gives you

```
org.opentest4j.AssertionFailedError: Expected [C] NOT to be in [A, B, C] ==> 
Expected :true
Actual   :false
```

instead of a giant output with most fields being equal and or irrelevant.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
@@ -143,11 +143,11 @@ class TrackedEntityServiceTest extends PostgresIntegrationTestBase {
 
   private TrackedEntityAttribute teaC;
 
-  private TrackedEntityAttributeValue trackedEntityAttributeValueA;
+  private TrackedEntityAttributeValue tetavA;
 
-  private TrackedEntityAttributeValue trackedEntityAttributeValueB;
+  private TrackedEntityAttributeValue tetavB;
 
-  private TrackedEntityAttributeValue trackedEntityAttributeValueC;
+  private TrackedEntityAttributeValue pteavC;
 
   private TrackedEntityType trackedEntityTypeA;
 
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -236,13 +236,13 @@ void setUp() {
...
         Set.of(orgUnitA, orgUnitB, orgUnitC));
 
-    teaA = createTrackedEntityAttribute('A', ValueType.TEXT);
-    teaA.setUnique(true);
-    manager.save(teaA, false);
-    TrackedEntityAttribute teaB = createTrackedEntityAttribute('B', ValueType.TEXT);
-    manager.save(teaB, false);
-    teaC = createTrackedEntityAttribute('C', ValueType.TEXT);
-    manager.save(teaC, false);
+    tetaA = createTrackedEntityAttribute('A', ValueType.TEXT);
```
muilpp, line 239: Should these variables be called teaA, teaB and pteaC?
Their type is `TrackedEntityAttribute`, not `TrackedEntityTypeAttribute`.

muilpp (approved):

ameenhere (approved):

PR #19647, opened 2025-01-13
chore: align fetching a single vs multiple TEs more

teleivo:
continuing on https://github.com/dhis2/dhis2-core/pull/19633

The goal is to ultimately use the same code to fetch a single vs multiple TEs. This PR makes another step towards that goal by

* using the same methods to set the `RelationshipItem`s
* using the same method to set the `TrackedEntityAttributeValue`s. `/trackedEntities?program=uid`returned all TEAV so far while `/trackedEntities/uid?program=uid` only returned TET+program specific TEAV. That behavior was introduced only to `/trackedEntities/uid?program=uid` in https://github.com/dhis2/dhis2-core/pull/17572

Prefix methods that mutate with `set` and those that map with `map`.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -477,102 +410,128 @@ public List<TrackedEntity> getTrackedEntities(
...
       List<TrackedEntity> trackedEntities, TrackedEntityParams params, boolean includeDeleted)
       throws NotFoundException {
+    for (TrackedEntity trackedEntity : trackedEntities) {
+      setRelationshipItems(trackedEntity, trackedEntity, params, includeDeleted);
+    }
+  }
+
+  private void setRelationshipItems(
+      TrackedEntity targetTrackedEntity,
+      TrackedEntity sourceTrackedEntity,
```
enricocolasante, line 441: I am confused by this target and source `trackedEntity` here.
When `setRelationshipItems` is called by `getTrackedEntities` method, source and target is the same `trackedEntity` and the nested for loops seem not necessary.
When `setRelationshipItems` is called by `getTrackedEntity`, source is the `trackedEntity`to be mapped and the target is the `trackedEntity` coming from the DB. At this point source `enrollments` and `events` are not set yet, so no relationship item will be set for enrollments and events even if the params are set to include relationhips.

enricocolasante (approved):

muilpp (approved):

PR #19670, opened 2025-01-15
fix: Allow granting temporary access only when user in search scope [DHIS2-18784]

muilpp:
There are two conditions required to be granted temporary access to an enrollment:
    1. The program must be protected.
    2. The TE/program owner must be within the user's search scope.

Currently, it is possible to gain temporary access to an enrollment even if the user is not within the search scope. This PR addresses and fixes that issue. After the fix, if the org unit is not within the search scope, an exception will be thrown.
Similarly, if the program is not protected, it will also throw an exception. Previously, access was denied in such cases, but no clear feedback was provided, which was confusing.

Additionally, I am removing the `assignOwnership` method, as it was only used in tests and is no longer necessary.

Finally, when transferring ownerships, I am including the program parameter when searching for the TE. Since the program is available in the request, it makes more sense to check the access to that TE/program directly when transferring an enrollment to another org unit.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-18784

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
@@ -32,6 +32,7 @@ import static org.hisp.dhis.external.conf.ConfigurationKey.CHANGELOG_TRACKER;
 import java.util.Optional;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
+import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
 import org.hibernate.Hibernate;
 import org.hisp.dhis.cache.Cache;
@@ -149,39 +150,40 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
   @Override
   @Transactional
   public void grantTemporaryOwnership(
-      TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      @Nonnull TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
       throws ForbiddenException {
-    if (canSkipOwnershipCheck(user, program) || trackedEntity == null) {
-      return;
+    if (canSkipOwnershipCheck(user, program)) {
+      throw new ForbiddenException(
+          "Temporary ownership not created. Either current user is a superuser, program supplied is does not exist or program supplied is not a tracker program.");
     }
 
-    if (program.isProtected()) {
-      if (!isOwnerInUserSearchScope(user, trackedEntity, program)) {
-        throw new ForbiddenException(
-            "The owner of the entity-program combination is not in the user's search scope.");
-      }
-
-      if (config.isEnabled(CHANGELOG_TRACKER)) {
-        programTempOwnershipAuditService.addProgramTempOwnershipAudit(
-            new ProgramTempOwnershipAudit(program, trackedEntity, reason, user.getUsername()));
-      }
-
-      ProgramTempOwner programTempOwner =
-          new ProgramTempOwner(
-              program,
-              trackedEntity,
-              reason,
-              userService.getUser(user.getUid()),
-              TEMPORARY_OWNERSHIP_VALIDITY_IN_HOURS);
-      programTempOwnerService.addProgramTempOwner(programTempOwner);
-      tempOwnerCache.invalidate(
-          getTempOwnershipCacheKey(trackedEntity.getUid(), program.getUid(), user.getUid()));
-    } else {
+    if (!program.isProtected()) {
       throw new ForbiddenException(
           String.format(
               "Temporary ownership can only be granted to protected programs. %s access level is %s.",
               program.getUid(), program.getAccessLevel().name()));
     }
+
+    if (!isOwnerInUserSearchScope(user, trackedEntity, program)) {
+      throw new ForbiddenException(
+          "The owner of the entity-program combination is not in the user's search scope.");
+    }
+
+    if (config.isEnabled(CHANGELOG_TRACKER)) {
+      programTempOwnershipAuditService.addProgramTempOwnershipAudit(
+          new ProgramTempOwnershipAudit(program, trackedEntity, reason, user.getUsername()));
+    }
+
+    ProgramTempOwner programTempOwner =
+        new ProgramTempOwner(
+            program,
+            trackedEntity,
+            reason,
+            userService.getUser(user.getUid()),
+            TEMPORARY_OWNERSHIP_VALIDITY_IN_HOURS);
+    programTempOwnerService.addProgramTempOwner(programTempOwner);
+    tempOwnerCache.invalidate(
+        getTempOwnershipCacheKey(trackedEntity.getUid(), program.getUid(), user.getUid()));
   }
 
   @Override
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
@@ -158,7 +158,7 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
     if (program.isProtected()) {
       if (!isOwnerInUserSearchScope(user, trackedEntity, program)) {
         throw new ForbiddenException(
-            "The owner of the entity-program combination is not in the user's search scope");
+            "The owner of the entity-program combination is not in the user's search scope.");
       }
 
       if (config.isEnabled(CHANGELOG_TRACKER)) {
@@ -179,7 +179,7 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
     } else {
       throw new ForbiddenException(
           String.format(
-              "It's only possible to grant temporary ownership to protected programs, and %s access level is %s",
+              "Temporary ownership can only be granted to protected programs. %s access level is %s.",
               program.getUid(), program.getAccessLevel().name()));
     }
   }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
@@ -154,7 +154,7 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
       throws ForbiddenException {
     if (canSkipOwnershipCheck(user, program)) {
       throw new ForbiddenException(
-          "Temporary ownership not created. Either current user is a superuser, program supplied is does not exist or program supplied is not a tracker program.");
+          "Temporary ownership not created. Either current user is a superuser, program supplied does not exist or program supplied is not a tracker program.");
     }
 
     if (!program.isProtected()) {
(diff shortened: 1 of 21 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -147,63 +146,26 @@ public void transferOwnership(
...
+      TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      throws ForbiddenException {
     if (canSkipOwnershipCheck(user, program) || trackedEntity == null) {
       return;
     }
 
     if (program.isProtected()) {
+      if (!isOwnerInUserSearchScope(user, trackedEntity, program)) {
+        throw new ForbiddenException(
+            "The owner of the entity-program combination is not in the user's search scope");
```
muilpp: Better to say TE not found? It could be the TE is just registered and not enrolled, but more often than not, the TE will be enrolled

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -147,63 +146,26 @@ public void transferOwnership(
...
+      TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      throws ForbiddenException {
     if (canSkipOwnershipCheck(user, program) || trackedEntity == null) {
       return;
     }
 
     if (program.isProtected()) {
+      if (!isOwnerInUserSearchScope(user, trackedEntity, program)) {
+        throw new ForbiddenException(
+            "The owner of the entity-program combination is not in the user's search scope.");
```
muilpp, line 161: Better saying `TE not found`? It could be the TE is just registered and not enrolled, but more often than not, the TE will be enrolled

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -147,63 +146,26 @@ public void transferOwnership(
...
-    }
-  }
-
   @Override
   @Transactional
   public void grantTemporaryOwnership(
-      TrackedEntity trackedEntity, Program program, UserDetails user, String reason) {
+      TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      throws ForbiddenException {
     if (canSkipOwnershipCheck(user, program) || trackedEntity == null) {
```
enricocolasante, line 154: Not really part of your fix code but does it make sense to just say to the user `"Temporary Ownership granted"` when the program was not found and no ownership was really granted?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -147,63 +146,26 @@ public void transferOwnership(
...
-      throw new AccessDeniedException(
-          "User does not have access to assign ownership for the entity-program combination");
-    }
-  }
-
   @Override
   @Transactional
   public void grantTemporaryOwnership(
-      TrackedEntity trackedEntity, Program program, UserDetails user, String reason) {
+      TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
```
enricocolasante, line 152: `trackedEntity` could be annotated with @NotNull as the service in the controller makes sure the result is always returned and then we can remove `trackedEntity == null` from the if in next line.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/ownership/TrackerOwnershipController.java
```diff
@@ -97,15 +98,16 @@ public WebMessage updateTrackerProgramOwner(
             "trackedEntityInstance", trackedEntityInstance, "trackedEntity", trackedEntity);
 
     trackerOwnershipAccessManager.transferOwnership(
-        trackedEntityService.getTrackedEntity(trackedEntityUid),
+        trackedEntityService.getTrackedEntity(
+            trackedEntityUid, UID.of(program), TrackedEntityParams.TRUE),
```
teleivo, line 102: Do you need to get enrollments, ... as defined in `TrackedEntityParams.TRUE`? Double-check to only fetch what's really needed.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/ownership/TrackerOwnershipController.java
```diff
@@ -97,15 +98,16 @@ public WebMessage updateTrackerProgramOwner(
...
         programService.getProgram(program),
         organisationUnitService.getOrganisationUnit(ou));
     return ok("Ownership transferred");
   }
 
   @PostMapping(value = "/override", produces = APPLICATION_JSON_VALUE)
   @ResponseBody
-  public WebMessage overrideOwnershipAccess(
+  public WebMessage grantTemporaryAccess(
```
teleivo, line 110: Does this mean the path `/override` is also not accurate?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -149,72 +149,41 @@ public void transferOwnership(
...
-      boolean skipAccessValidation,
-      boolean overwriteIfExists) {
-    if (trackedEntity == null || program == null || organisationUnit == null) {
-      return;
+  public void grantTemporaryOwnership(
+      @Nonnull TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      throws ForbiddenException {
+    if (canSkipOwnershipCheck(user, program)) {
+      throw new ForbiddenException(
+          "Temporary ownership not created. Either current user is a superuser, program supplied is does not exist or program supplied is not a tracker program.");
```
enricocolasante, line 157: Here we can easily tell the user exactly what is the issue without the need of reusing `canSkipOwnershipCheck` method.

I think we have enough cases to create a validate method and to give useful error message to the client.

```suggestion
          "Temporary ownership not created. Either current user is a superuser, program supplied does not exist or program supplied is not a tracker program.");
```

enricocolasante (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/ownership/TrackerOwnershipController.java
ameenhere: From a strict RESTful naming convention, I guess having a verb is odd for the API path. APIs could have been `api/tracker/ownership/permanent` and `api/tracker/ownership/temporary`

But apart from that, `override` is not inaccurate.  The functionality for this API is called "Breaking the glass" which is pretty much overriding/bypassing existing access check mechanisms. Atleast for now, I guess we really don't have to change API paths for this. Unless there is very strong reasonings, which I don't see at this point.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -149,72 +149,41 @@ public void transferOwnership(
...
-      OrganisationUnit organisationUnit,
-      boolean skipAccessValidation,
-      boolean overwriteIfExists) {
-    if (trackedEntity == null || program == null || organisationUnit == null) {
-      return;
+  public void grantTemporaryOwnership(
+      @Nonnull TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
+      throws ForbiddenException {
+    if (canSkipOwnershipCheck(user, program)) {
+      throw new ForbiddenException(
```
ameenhere, line 156: Doesn't this break the API? 
I feel it may be intrusive to throw an exception for the scenario were the check itself could be skipped? Thoughts?

enricocolasante: The issue was that the check was used to skip the actual action that is granting temporary access.
In my opinion the API was just wrong because when the user was asking access for a not existent program, the API would return with a positive answer even though nothing happened.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/ownership/TrackerOwnershipController.java
muilpp: I'm fine with not renaming the path for now, but I'd stick to the new method name.
IMO, `override` means replacing the existing value with a new one, and that's not the case here, is it?
Here we simply grant temporary access to someone who's not in the CS, but everyone who is in the CS still can see this enrollment, so their access is not overridden.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
muilpp: @ameenhere what's your take on this one?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/ownership/TrackerOwnershipController.java
ameenhere: Replacing existing value with a new one is `overwrite`. 

`Override` in Security/access control only means, for example, to allow a user to access something which in normal scenario they are not allowed to access. This is exactly what we do when saying granting temporary access.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
ameenhere: No strong opinions really. I guess this is fine. 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
ameenhere: I agree with the program non-existent and non-tracker program part. 
For superuser, I am a bit concerned if it will break the UI workflow in some way. But maybe it should still not end up here anyway, since they are a superuser 🤔

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
muilpp: That was my thinking. In our current flow, a superuser will not end up here since they have access to every TE in the system, right?
It's my understanding that a superuser will always have capture scope access to all org units in the implementation, so the capture app would never send such a user to this endpoint.
If this assumption is incorrect, should we consider granting temporary access to the superuser as well?

ameenhere: I guess the assumption is correct. Let us see how this goes.

PR #19686, opened 2025-01-16
fix: FK constraint error when delete Program with MapView

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-18702

### Issue:
- Database throw FK constraint violation when deleting a Program which is referenced by MapView.

### Fix
- Remove the program reference from MapView.
- The MapView table also has ProgramStage foreign key, so it also needed to be removed.
- After deleting the Program, user might see an error when viewing the Map. If that happens, user can delete the Map in the Maps app.

### Test
- Added unit test.
- Manual test steps are described in [DHIS2-18702](https://dhis2.atlassian.net/browse/DHIS2-18702)

[DHIS2-18702]: https://dhis2.atlassian.net/browse/DHIS2-18702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/mapping/MapViewDeletionHandler.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/mapping/MapViewDeletionHandler.java
@@ -32,6 +32,7 @@ import static org.hisp.dhis.system.deletion.DeletionVeto.ACCEPT;
 
 import java.util.List;
 import org.hisp.dhis.common.GenericAnalyticalObjectDeletionHandler;
+import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.dataelement.DataElement;
 import org.hisp.dhis.dataset.DataSet;
 import org.hisp.dhis.expressiondimensionitem.ExpressionDimensionItem;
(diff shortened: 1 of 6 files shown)


david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/mapping/MapViewDeletionHandler.java
```diff
@@ -83,6 +85,24 @@ private void deleteLegendSet(LegendSet legendSet) {
...
+    for (MapView mapView : mapViews) {
+      mapView.setProgram(null);
+      if (mapView.getProgramStage() != null) {
+        program
+            .getProgramStages()
+            .forEach(
+                programStage -> {
+                  if (mapView.getProgramStage().getUid().equals(programStage.getUid())) {
+                    mapView.setProgramStage(null);
+                  }
```
jbee, line 99: This looks a bit like `forEach` abuse. If you want to set the program stage of the map-view `null` in case the program has a stage with the same ID as the map-view is using I think something like 
```java
if (mapView.getProgramStage() != null && program         .getProgramStages().contains(mapView.getProgramStage()) {
   mapView.setProgramStage(null);
}            
```
is more clear. Maybe you want to compare IDs to be more safe that contains can be written similarly using `program.getProgramStages().stream().map(IdentifiableObject::getUid).toList().containts...`

jbee (approved):

PR #19687, opened 2025-01-16
fix: Update data dimension items as part of cat opt combo merge [DHIS2-18321]

david-mackessy:
# Problem
Luckily, while Jason P was testing some merge scenarios, he stumbled across this error when trying to merge 2 COCs in the SL DB (`KPP63zJPkOu`, `TDb5JyDQqho`).
`Object could not be deleted because it is associated with another object: DataDimensionItem`

# Cause
After looking this, it seems that a `DataElementOperand` is also a `DataDimensionItem`. Although the `DataElementOperand` is updated as part of the merge, this change is not reflected in the `datadimensionitem` table even though it is the same `DataElementOperand` :( 

The `DataDimensionItem` type does not have a `CategoryOptionCombo` property which is how this was missed originally.

# Fix
The `datadimensionitem` table needs to be updated also as part of the merge.

# Testing
## Automated
E2E test updated so that data dimension items are persisted, which also need updating

## Manual
Using this payload failed before the fix, and passes now
`POST` `/api/categoryOptionCombos/merge`
```json
{
    "sources": [
        "KPP63zJPkOu"
    ],
    "target": "TDb5JyDQqho",
    "deleteSources": true,
    "dataMergeStrategy": "LAST_UPDATED"
}
```


vietnguyen (approved):

jbee (approved):
The `EmbeddedObject` did strike again.

PR #19689, opened 2025-01-16
chore: Use EventService in EnrollmentService [DHIS2-18791]

enricocolasante:
Use `https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java#L242` to get the events from the enrollment instead of getting them through the Hibernate proxy.

Getting the events we can safely catch the `BadRequestException` as we are building `EventOperationParams`.
We can also catch `ForbiddenException`. This is less intuitive but this exception is thrown only in the process of mapping the `EventOperationParams` to `EventQueryParams` and we are not using any of the parameters that could thrown this exception.

**Other changes**
- Assign `ACCESSIBLE` as default value to `orgUnitMode` in `EventOperationParams`.
- Remove `EnrollmentParams` from `getEnrollmentInRelationshipItem` signature. The shape of the enrollment to get for a relationship is always the same and it should be hardcoded in the method itself. We are also changing `EnrollmentParams` here to not retrieve the events as they are never part of the enrollment in the relationship https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipItemMapper.java#L72-L88

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -53,6 +53,9 @@ import org.hisp.dhis.tracker.acl.TrackerAccessManager;
 import org.hisp.dhis.tracker.acl.TrackerOwnershipManager;
 import org.hisp.dhis.tracker.export.Page;
 import org.hisp.dhis.tracker.export.PageParams;
+import org.hisp.dhis.tracker.export.event.EventOperationParams;
+import org.hisp.dhis.tracker.export.event.EventParams;
+import org.hisp.dhis.tracker.export.event.EventService;
 import org.hisp.dhis.user.UserDetails;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
@@ -62,6 +65,8 @@ import org.springframework.transaction.annotation.Transactional;
 @Service("org.hisp.dhis.tracker.export.enrollment.EnrollmentService")
 class DefaultEnrollmentService implements EnrollmentService {
   private final EnrollmentStore enrollmentStore;
+
+  private final EventService eventService;
 
   private final TrackerOwnershipManager trackerOwnershipAccessManager;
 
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -162,20 +168,29 @@ public RelationshipItem getEnrollmentInRelationshipItem(
       return null;
     }
 
-    relationshipItem.setEnrollment(getEnrollment(enrollment, params, includeDeleted, currentUser));
+    relationshipItem.setEnrollment(
+        getEnrollment(
+            enrollment,
+            EnrollmentParams.FALSE.withIncludeAttributes(true),
```
teleivo, line 174: I am not sure we always want the attributes when getting an enrollment as part of a relationship 🤔 you should be able to come up with a `fields=relationships[from[...` that excludes them.

I think this should be `params.withRels(false)` so it respects what the user wanted but prevents us from getting relationships again 😅

teleivo:
Does this change also apply to both `/enrollments/uid` and `/enrollments`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
enricocolasante: I agree.
But this is a wider fix, at the moment we are not passing the field parameters at all to the `getRelationshipItems` methods.
In this PR I just changed the hardcoded params to exclude the events that before were included.

enricocolasante:
> Does this change also apply to both `/enrollments/uid` and `/enrollments`?

Yes, underneath the both use the private `GetEnrollment` method that now is using `EventService`.
Anyway, in the next PR I am doing more clean up of the paths

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
teleivo: ok,as long as we do it before the release 😋

teleivo (approved):

ameenhere (approved):

PR #19702, opened 2025-01-17
chore: cleanup TE aggregate and audit DHIS2-18541

teleivo:
* remove unnecessary interfaces: we discussed to only uses interfaces if we have multiple implementations or want to isolate components in a test
* reduce visibility of aggregate classes so we can be sure later changes only have an effect on the aggregate package
* make the TE audit service API symmetric when it comes to creating an audit for one or many TEs
* ensure the `isAllowAuditLog` field is mapped in the aggregates TE query (`TrackedEntityRowCallbackHandler`). Note the join on the TET table was already present
* fix audit type in `DefaultTrackerObjectsDeletionService` which should be DELETE

## Next

* use the `EnrollmentService` to get enrollments if needed by `/trackedEntities/uid?fields=enrollments`
* use the `EnrollmentService` to get enrollments if needed by `/trackedEntities?fields=enrollments`
* merge code paths for single vs multiple entities


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/audit/TrackedEntityAuditService.java
```diff
@@ -40,11 +41,11 @@ public interface TrackedEntityAuditService {
 
   String ID = TrackedEntityAuditService.class.getName();
 
-  void addTrackedEntityAudit(
-      TrackedEntity trackedEntity, String username, AuditOperationType auditOperationType);
+  void addTrackedEntityAudit(AuditOperationType type, String username, TrackedEntity trackedEntity);
```
ameenhere, line 44: trackedEntity here could be `@Nonnull` as well?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/aggregates/AclStore.java
```diff
@@ -28,16 +28,101 @@
...
 
-  List<Long> getAccessiblePrograms(String userUID, List<String> userGroupUIDs);
+  private static final String USER_SQL_PARAM_NAME = "userId";
 
-  List<Long> getAccessibleProgramStages(String userUID, List<String> userGroupUIDs);
+  private static final String USER_GROUP_SQL_PARAM_NAME = "userGroupUIDs";
 
-  List<Long> getAccessibleRelationshipTypes(String userUID, List<String> userGroupUIDs);
+  private static final String PUBLIC_ACCESS_CONDITION =
+      "sharing->>'public' LIKE '__r%' OR sharing->>'public' IS NULL";
```
ameenhere, line 51: Does this mean the functionally it is the same if shating->>public begins with __r or null?

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/audit/TrackedEntityAuditService.java
teleivo: TE does not need to be non-null see implementation

PR #19727, opened 2025-01-20
fix: IN filter now works with org units data elements [DHIS2-18845]

gnespolino:
This pull request will fix a small bug with IN query operator, making it properly quotes values when needed (it was only quoting `TEXT` value types before).
Moreover, it will properly support Data Elements of type ORG_UNIT, by replacing the filter values with org unit UIDs when needed

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DimensionalObjectProvider.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DimensionalObjectProvider.java
@@ -424,6 +424,24 @@ public class DimensionalObjectProvider {
   }
 
   /**
+   * This method will return a list of {@link OrganisationUnit} UIDs based on the given items and
+   * user organisation units.
+   *
+   * @param items the list of items that might be included into the resulting organisation unit and
+   *     its keywords.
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/data/DimensionalObjectProvider.java
```diff
@@ -423,6 +423,24 @@ public BaseDimensionalObject getOrgUnitDimension(
...
+   *
+   * @param items the list of items that might be included into the resulting organisation unit and
+   *     its keywords.
+   * @param userOrgUnits the list of organisation units associated with the current user.
+   * @return a list of {@link OrganisationUnit} UIDs.
+   */
+  public List<String> getOrgUnitDimensionUid(
+      List<String> items, List<OrganisationUnit> userOrgUnits) {
+    return getOrgUnitDimensionItems(
+            items, userOrgUnits, IdScheme.UID, new ArrayList<>(), new ArrayList<>())
```
maikelarabori, line 438: Could we use List.of(), instead of new ArrayList<>()?

gnespolino: for some reason, those lists need to be mutable. So cannot use `List.of()`

PR #19729, opened 2025-01-20
feat: Add data element name as a fallback when sorting [DHIS2-18871]

muilpp:
Adding data element name as a fallback in case form name is not set when sorting data elements change logs

### DIFF
--- a/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
@@ -390,6 +390,7 @@
       },
       "name": "test-dataelement1",
       "shortName": "test-dataelement1",
+      "formName": "test-dataelement1",
       "user": {
         "id": "tTgjgobT1oS"
       },
--- a/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
+++ b/dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
@@ -390,7 +390,7 @@
       },
       "name": "test-dataelement1",
       "shortName": "test-dataelement1",
-      "formName": "test-dataelement1",
+      "formName": "test-dataelement9",
       "user": {
         "id": "tTgjgobT1oS"
       },
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
@@ -288,10 +288,10 @@ class OrderAndFilterEventChangeLogTest extends TrackerTest {
     assertEquals(
         List.of(
             "with-option-set",
+            "test-dataelement9",
+            "test-dataelement9",
+            "test-dataelement9",
             "test-dataelement6",
-            "test-dataelement1",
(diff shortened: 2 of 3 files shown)


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
```diff
@@ -268,6 +268,31 @@ void shouldSortChangeLogsWhenOrderingByChangeDescAndChangesOnlyToEventFields()
...
+
+    updateDataValues(event, "DATAEL00001", "value00002", "value00003");
+
+    List<EventChangeLog> changeLogs =
+        eventChangeLogService
+            .getEventChangeLog(UID.of("pTzf9KYMk72"), params, defaultPageParams)
+            .getItems();
+
+    assertNumberOfChanges(7, changeLogs);
+    assertAll(
```
enricocolasante, line 286: Maybe we could have an assertion just to check the order like mapping the changelogs to the order field (if possible) or to the `UID` and make a comment to show what is the value of the name.
Something like:
`expected = List.of("with-option-set", "test-dataleement6", "test-dataelement1", "test-dataelement1", "test-dataelement1", "scheduledAt", "occurredAt")`

This way would be easy to understand why the assertion makes sense.

enricocolasante (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
muilpp: is [this](https://github.com/dhis2/dhis2-core/pull/19729/files#diff-2389d877f149c40712b08260baa60ae93237355a62f970e917258ec0652b1b9dR288) what you meant?

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
```diff
@@ -390,6 +390,7 @@
       },
       "name": "test-dataelement1",
       "shortName": "test-dataelement1",
+      "formName": "test-dataelement1",
```
teleivo, line 393: Should the `formName` get a different value than the `name` to test some logic? Imagine you introduce a bug in your logic would the test still pass and thus hide that bug?

Always good to let your test fail by messing up your implementation afterwards (when not doing TDD 😂).

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
```diff
@@ -268,6 +268,35 @@ void shouldSortChangeLogsWhenOrderingByChangeDescAndChangesOnlyToEventFields()
...
+    Event event = getEvent("pTzf9KYMk72");
+
+    updateDataValues(event, "DATAEL00001", "value00002", "value00003");
+
+    List<String> changeLogs =
+        eventChangeLogService
+            .getEventChangeLog(UID.of("pTzf9KYMk72"), params, defaultPageParams)
+            .getItems()
+            .stream()
+            .map(this::getDisplayName)
```
teleivo, line 285: Maybe you cold replace the `getDisplayName` method with with `ObjectUtils.firstNonNull`

teleivo (approved):

PR #19743, opened 2025-01-22
refactor: DataElement merge, use SQL when merging DataValues and EventDataValues [DHIS2-14429]

david-mackessy:
# Summary
`DataElement` merge refactor changes, for significant performance gains:
- using SQL instead of Java when:
  - merging `DataValue`s
  - merging `EventDataValue`s
  - deleting `EventDataValue`s

- remove previously used store methods & other merge-related code, now redundant
- re-use the SQL merge script
  - the merging of `DataValue`s is practically identical for 3 use cases:
    - data element
    - category option combo
    - attribute option combo
  - use placeholders in the script to keep readability and replace the values needed for each use case

# Testing
## Automated
Some tests added and updated for merging of data values & event data values

## Performance

| Scenario | Execution time before change | Execution time after change | entities involved |
|----------|----------------|--------------| ----|
| Merge data values | 10s | 3s | 155k |
| Merge event data values | 1m 10s | 14s | 1.1 million (JSONB) |
| Delete event data values | 1m 22s | 8s | 1.1 million (JSONB) |

For extra confidence, given that the merging and deletion of data is now done via SQL, these manual checks were performed during testing to make sure unrelated entities (no source or target relation) were not affected:
| Scenario | No. of unrelated entities before merge | No. of unrelated entities after merge |
|----------|----------------|--------------|
| Merge data values | 4,701,315 | 4,701,315 |
| Merge event data values | 874,697 | 874,697 |
| Delete event data values | 874,697 | 874,697 |

Existing tests also test for unrelated entities still existing after merging.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/datavalue/hibernate/HibernateDataValueStore.java
```diff
@@ -1028,4 +836,111 @@ private void getDdvAddToBlockingQueue(
...
+          .replace("data_element", "source_dv.dataelementid")
+          .replace("cat_opt_combo", "source_dv.categoryoptioncomboid")
+          .replace("attr_opt_combo", "target_id");
+    }
+    // if SQL params haven't been replaced there's no point trying to execute a bad SQL query
+    throw new IllegalArgumentException(
+        "Error while trying to construct SQL for data value merge for " + mergeType);
+  }
+
+  enum DataValueMergeType {
```
jbee, line 941: I think you can make the type `private`

larshelge (approved):

PR #19756, opened 2025-01-23
chore: join code paths for serving one vs many TEs DHIS2-18541

teleivo:
* convert `ExportControllerPaginationTest` to postgres so we test against valid data with a more realistic user setup
* change expectations of 403 to 404 in `/tracker/trackedEntities/[uid]` tests as we are now reusing the code of `/tracker/trackedEntities` which only filters out TEs the user does not have access to


teleivo:
closing as @enricocolasante is taking over 😁

PR #19760, opened 2025-01-23
feat: Filter events by existing data element [DHIS2-15954]

muilpp:
### Part 1
Requesting events with a filter like `filter={DE_UID}` doesn't work as expected, events are returned regardless of the DEs they contain. This PR aims to fix this issue.

The DEs are stored in a map that uses the UID as the key and a `List<QueryFilter>` as the value. In this case, the value list is always empty. The issue is that this map is also used when sorting. Even though the DEs are present in the order list, we are still using the DE map to add fields to the select clause.

I am changing this in favor of the same approach we use when sorting TEAs, using the order list to add any DE to the select statement.

### Next
This fix will allow users to filter for events containing a value in a given DE.
The next PR will enable filtering for events not containing a value in a given DE.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -894,7 +894,7 @@ left join dataelement de on de.uid = eventdatavalue.dataelement_uid
             .append(", coc_agg.attributevalues as ")
             .append(COLUMN_EVENT_ATTRIBUTE_OPTION_COMBO_ATTRIBUTE_VALUES)
             .append(", coc_agg.co_values AS co_values, coc_agg.co_count AS option_size, ")
-            .append(addOrderFieldsToSelectClause(params.getOrder()));
+            .append(getOrderFieldsForSelectClause(params.getOrder()));
 
     return selectBuilder
         .append(
@@ -920,7 +920,7 @@ left join dataelement de on de.uid = eventdatavalue.dataelement_uid
         .toString();
   }
 
-  private String addOrderFieldsToSelectClause(List<Order> orders) {
+  private String getOrderFieldsForSelectClause(List<Order> orders) {
     StringBuilder selectBuilder = new StringBuilder();
 
     for (Order order : orders) {
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -1444,6 +1444,11 @@ left join dataelement de on de.uid = eventdatavalue.dataelement_uid
             }
           }
         }
+      } else {
+        eventDataValuesWhereSql.append(hlp.whereAnd());
+        eventDataValuesWhereSql.append(" (ev.eventdatavalues ?? '");
+        eventDataValuesWhereSql.append(item.getKey().getUid());
+        eventDataValuesWhereSql.append("')");
       }
     }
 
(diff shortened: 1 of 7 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventQueryParamsTest.java
```diff
@@ -70,13 +70,13 @@ void shouldKeepExistingAttributeFiltersWhenOrderingByAttribute() {
...
   @Test
-  void shouldAddDataElementToOrderAndDataElementsWhenOrderingByDataElement() {
+  void shouldAddDataElementToOrderButNotToDataElementsWhenOrderingByDataElement() {
     EventQueryParams params = new EventQueryParams();
 
     params.orderBy(de1, SortDirection.ASC);
 
     assertEquals(List.of(new Order(de1, SortDirection.ASC)), params.getOrder());
-    assertEquals(Map.of(de1, List.of()), params.getDataElements());
+    assertTrue(params.getDataElements().isEmpty());
```
teleivo, line 79: use `assertIsEmpty`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -928,6 +920,32 @@ private String getEventSelectQuery(
         .toString();
   }
 
+  private String addOrderFieldsToSelectClause(List<Order> orders) {
+    StringBuilder selectBuilder = new StringBuilder();
+
+    for (Order order : orders) {
+      if (order.getField() instanceof TrackedEntityAttribute tea) {
```
teleivo, line 927: https://github.com/dhis2/dhis2-core/blob/786a4dab2c7f377db0b6b204d4ab00143185e9ce/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventQueryParams.java#L108-L119

how are field names handled? 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -928,6 +920,32 @@ private String getEventSelectQuery(
         .toString();
   }
 
+  private String addOrderFieldsToSelectClause(List<Order> orders) {
```
teleivo, line 923:
```suggestion
  private String getOrderFieldsToSelectClause(List<Order> orders) {
```

Its a pure function `List<Order> -> String` that does not add anything. There might be a better name than my suggestion but `add` is misleading.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -590,6 +591,25 @@ void testExportEventsWhenFilteringByNumericDataElements()
     assertContainsOnly(List.of("D9PbzJY8bJM"), events);
   }
 
+  @Test
+  void shouldFilterByEventsWithGivenDataValuesWhenFilterContainsDataElementUIDsOnly()
+      throws ForbiddenException, BadRequestException {
+    EventOperationParams params =
+        EventOperationParams.builder()
+            .eventParams(EventParams.FALSE)
```
teleivo, line 599: I think we should make this the default in `EventOperationParams`, do you think so too @muilpp @enricocolasante ? So we fetch as little as possible by default.

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
enricocolasante: Agreed.
We have it for enrollments and tracked entities.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: We add all field names in the select clause [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java#L821)

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/event/EventQueryParamsTest.java
muilpp: this is a Map, not a Collection, that's why I didn't use it
I didn't deem it necessary to create a helper method for this one

PR #19762, opened 2025-01-23
feat: Add Prometheus metrics for data integrity checks

jason-p-pickering:


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/utils/PrometheusMetricsConstants.java
```diff
@@ -0,0 +1,17 @@
...
+ * Utility class that defines constants for Prometheus metrics types.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PrometheusMetricsConstants
+{
+    public static final String GAUGE = "gauge";
+    public static final String COUNTER = "counter";
+    public static final String SUMMARY = "summary";
+    public static final String HISTOGRAM = "histogram";
+    public static final String UNTYPED = "untyped";
```
jbee, line 16: As I explained in my other comment, I think these belong in string templates. So, one way to do that would be to have dedicated methods in `PrometheusTextBuilder`, like `gaugeLine` which would be similar to `typeLine` just that the type is already defined by the method called. Or just put them inline as an argument string to `helpLine`, like, `metrics.typeLine("dhis_data_integrity_check", "gauge" );`. Which as you see I would also do for the name. the goal with things like this is always to make the code read like the result. And the result is: `# TYPE dhis_data_integrity_check gauge`. So personally, I'd even go as far as naming the method `TYPE`, so
`metrics.TYPE("dhis_data_integrity_check", "gauge");` If we compare code and result

```
metrics.TYPE("dhis_data_integrity_check", "gauge");
#TYPE dhis_data_integrity_check gauge
```
To me that is a compelling reason. If it reads like what you want to create chances are it will just do that correctly. Everything that adds clutter and translation to it just taxes you brain into making mistakes. 
As I explained, in that regard I don't think that general coding rules and conversions should apply, like not naming a method `TYPE` because of camel-case is the norm.

jason-p-pickering: Thanks for the feedback @jbee . In the last commit, I have removed the constants class, and tried to simplify things a bit inside the class. For now, the only type of metrics we have is a "gauge", and considering how difficult it is for me to spell it, a constant helps in this case. :-D I did add some default methods in the PrometheusTextBuilder though, which makes by default, the metric type a gauge. I think this makes sense (at least for now) since all of the metrics we are exposing are this type. I also renamed the methods per your example.

jbee (approved):

jbee (approved):

PR #19763, opened 2025-01-23

cjmamo:
* Add headers and query params auth schemes to Route API to support Climate health project requirements
* Add basic integration test coverage for Route API
* Turn encryption/decryption of auth schemes into a strategy pattern to avoid IF conditions

### DIFF
--- a/dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/notification/NotifierTest.java
+++ b/dhis-2/dhis-support/dhis-support-system/src/test/java/org/hisp/dhis/system/notification/NotifierTest.java
@@ -194,14 +194,14 @@ class NotifierTest {
         .forEach(
             i -> {
               for (Notification notification :
-                  getNotifications(METADATA_IMPORT).get(jobConfig.getUid())) {
+                  notifier.getNotificationsByJobId(METADATA_IMPORT, jobConfig.getUid())) {
                 // Iterate over notifications when new notification are added
                 assertNotNull(notification.getUid());
               }
             });
     awaitTermination(e);
     awaitIdle();
-    assertEquals(101, getNotifications(METADATA_IMPORT).get(jobConfig.getUid()).size());
+    assertEquals(101, notifier.getNotificationsByJobId(METADATA_IMPORT, jobConfig.getUid()).size());
   }
 
   public void awaitTermination(ExecutorService threadPool) {


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatusCode;
+import org.springframework.http.ResponseEntity;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Tests the {@link DatastoreController} using (mocked) REST requests.
+ *
+ * @author Jan Bernitt
```
david-mackessy, line 64: Maybe change this :)

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+  @Test
+  void testAddRouteGivenApiHeadersAuthScheme() throws JsonProcessingException {
+    ApiHeadersAuthScheme apiHeadersAuthScheme = new ApiHeadersAuthScheme();
+    apiHeadersAuthScheme.setHeaders(Map.of("X-API-KEY", "foo"));
+
+    Route route = new Route();
+    route.setName("route-under-test");
+    route.setAuth(apiHeadersAuthScheme);
+    route.setUrl("https://postman-echo.com/post");
```
david-mackessy, line 158: I know the mock rest template is used to respond, but this url makes it look like the test might be calling an actual live endpoint. It makes the reader have to check the test setup to see that it is not. If the url doesn't matter, maybe make it obvious to the reader that it is not a real endpoint.

david-mackessy (approved):
Looks good to me.
- As a PR reviewer, it's always nice when the PR has a description body (what the new change is, why etc. and mentioning testing)
- I added Morten S as a reviewer as he works closely with the auth, in case he has any observations.

jbee:
@cjmamo is this ready for review? ping me or use "draft" mode as long as you still work on it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/AuthScheme.java
```diff
@@ -50,16 +51,23 @@
...
 @JsonSubTypes({
-  @JsonSubTypes.Type(value = HttpBasicAuth.class, name = "http-basic"),
-  @JsonSubTypes.Type(value = ApiTokenAuth.class, name = "api-token")
+  @JsonSubTypes.Type(value = HttpBasicAuthScheme.class, name = "http-basic"),
+  @JsonSubTypes.Type(value = ApiTokenAuthScheme.class, name = "api-token"),
+  @JsonSubTypes.Type(value = ApiHeadersAuthScheme.class, name = "api-headers"),
+  @JsonSubTypes.Type(value = ApiQueryParamsAuthScheme.class, name = "api-query-params")
 })
-public abstract class Auth implements Serializable {
+public abstract class AuthScheme implements Serializable {
```
jbee, line 59: I don't know much about this domain but using sub-classing seems overkill and unnecessarily complex to me. Generally with data I would recommend to stay clear of class hierarchies and instead try to find a common way of representing the information. For example, adding a `Map<String,String>` to keep the data and an enum type to indicate which of the auth types the data represents might work. Just a suggestion.

netroms (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -162,14 +162,11 @@ public Route getRouteWithDecryptedAuth(@Nonnull String id) {
       return null;
     }
 
-    try {
-      route = objectMapper.readValue(objectMapper.writeValueAsString(route), Route.class);
-    } catch (JsonProcessingException ex) {
-      log.error("Unable to create copy of route: '{}'", route.getUid());
-      return null;
-    }
+    route = Hibernate.unproxy(route, Route.class);
```
jbee, line 165: I guess this deserves a comment to explain why unproxying is a good idea here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -188,26 +185,25 @@ public Route getRouteWithDecryptedAuth(@Nonnull String id) {
...
       Route route, UserDetails userDetails, Optional<String> subPath, HttpServletRequest request)
       throws IOException, BadRequestException {
-    HttpHeaders headers = filterRequestHeaders(request);
 
+    HttpHeaders headers = filterRequestHeaders(request);
     route.getHeaders().forEach(headers::add);
-
     addForwardedUserHeader(userDetails, headers);
 
+    MultiValueMap<String, String> queryParameters = new LinkedMultiValueMap<>();
```
jbee, line 193: I think we want to avoid using any other collections and utils than JDK provided ones. The stream API and methods like `Map.computeIfAbsent` can be used to do the same. 
```java
map.computeIfAbsent("key", k -> new ArrayList<>()).addAll(...)
```

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/hibernate/jsonb/type/IgnoreJsonPropertyAccessJacksonAnnotationIntrospector.java
```diff
@@ -0,0 +1,47 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.hibernate.jsonb.type;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
+
+/** Deserialise write only properties since they may need to be read internally. */
+public class IgnoreJsonPropertyAccessJacksonAnnotationIntrospector
```
jbee, line 35: I assume this will then be applied to all JSON mapping done via jackson? 
In that case I don't think we want to do this. We rather invent a new annotation to mark just the fields you have an issue with to handle just those properties with a different policy.

jbee (commented):
Nice work on making the code more clear and getting rid of the ignore list map in the field filtering.

The issue that I think we have to look at is the change of jackson handling for write-only. Changing that in general is simply to big of a change and will likely not do what is intended in other existing usages of write-only.

jbee (changes requested):
Since you got 2 approval I better change to request changes since I think we cannot change jackson behaviour for write-only properties in general.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/ApiHeadersAuthScheme.java
```diff
@@ -0,0 +1,87 @@
...
+    return copy(encryptedHeaders);
+  }
+
+  @Override
+  public ApiHeadersAuthScheme decrypt(UnaryOperator<String> decryptFunc) {
+    Map<String, String> encryptedHeaders =
+        headers.entrySet().stream()
+            .map(e -> new AbstractMap.SimpleEntry<>(e.getKey(), decryptFunc.apply(e.getValue())))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    return copy(encryptedHeaders);
```
amcgee, line 73: Should this variable be called `decryptedHeaders` not `encryptedHeaders`?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/ApiQueryParamsAuthScheme.java
```diff
@@ -0,0 +1,89 @@
...
+    return copy(encryptedQueryParams);
+  }
+
+  @Override
+  public ApiQueryParamsAuthScheme decrypt(UnaryOperator<String> decryptFunc) {
+    Map<String, String> encryptedQueryParams =
+        queryParams.entrySet().stream()
+            .map(e -> new AbstractMap.SimpleEntry<>(e.getKey(), decryptFunc.apply(e.getValue())))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    return copy(encryptedQueryParams);
```
amcgee, line 75: Likewise, I think this should be `decryptedQueryParams`

dhis-2/dhis-support/dhis-support-hibernate/src/test/java/org/hisp/dhis/hibernate/jsonb/type/IgnoreJsonPropertyAccessJacksonAnnotationIntrospectorTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    ClassUnderTest classUnderTest = new ClassUnderTest();
+    classUnderTest.writeOnlyProperty = "Foo";
+    classUnderTest.readAndWriteProperty = "Bar";
+
+    String json = objectMapper.writeValueAsString(classUnderTest);
+
+    assertEquals("{\"writeOnlyProperty\":\"Foo\",\"readAndWriteProperty\":\"Bar\"}", json);
+    ClassUnderTest newClassUnderTest = objectMapper.readValue(json, ClassUnderTest.class);
+    assertEquals("Foo", newClassUnderTest.writeOnlyProperty);
```
amcgee, line 61: shouldn't this NOT contain the write-only property?

amcgee (approved):
LGTM, few minor comments.  Nice improvement!

jbee (approved):

PR #19770, opened 2025-01-24
fix: Return all requested fields in relationships and prevent cycles [DHIS2-18791]

enricocolasante:
When retrieving an enrollment for a relationship item the events were always excluded.
This PR changes the fields that need to be retrieved for enrollment inside a relationship and ensures that the enrollment and the underneath events are not showing their relationships to avoid any possible cycle.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -167,7 +167,14 @@ class DefaultEnrollmentService implements EnrollmentService {
   public RelationshipItem getEnrollmentInRelationshipItem(@Nonnull UID uid) {
     Enrollment enrollment;
     try {
-      enrollment = getEnrollment(uid);
+      enrollment =
+          getEnrollment(
+              uid,
+              EnrollmentParams.TRUE
+                  .withIncludeRelationships(false)
+                  .withEnrollmentEventsParams(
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -167,7 +167,14 @@ public Page<Enrollment> getEnrollments(
   public RelationshipItem getEnrollmentInRelationshipItem(@Nonnull UID uid) {
     Enrollment enrollment;
     try {
-      enrollment = getEnrollment(uid);
+      enrollment =
+          getEnrollment(
+              uid,
+              EnrollmentParams.TRUE
```
teleivo, line 173: Can this conflict with a `fields` param the users can send when for example requesting `/tracker/trackedEntities`? In my mind I would expect it to be something like `enrollmentParams.withIncludeRelationships(false)` passing in the params 🤔 not 100% sure though. With this it looks to me as if they would always get the fully expanded enrollment but might actually want something more minimal.

enricocolasante: That is not an issue for the user because the field filtering will happen in the controller with the right fields.
It is an optimization that we should add (I thought about the case but forgot to put it in the description).
I will add a TODO for [`DHIS2-18883`](https://dhis2.atlassian.net/browse/DHIS2-18883)

teleivo (approved):

muilpp (approved):

PR #19777, opened 2025-01-27

cjmamo:
Backport #19763


netroms (approved):

netroms (approved):

jbee (approved):
As far as I can see you didn't change jackson the same way you did in `master` so I think this one is good to go.

PR #19778, opened 2025-01-27
feat: Filter events by DE existence in event [DHIS2-15954]

muilpp:
### Part 2

Continuing the work started in [part 1](https://github.com/dhis2/dhis2-core/pull/19760), this PR introduces unary operators. These are operators that do not require a value in a filter. Initially, we'll support `null` and `!null`, which allow filtering events based on whether a specified data element has a value.

These operators can be combined with each other and with other operators, meaning the following filters are valid:
- `filter=UID:null:!null`
- `filter=UID:null:eq:value`

As explained [here](https://github.com/dhis2/dhis2-core/pull/19778#issuecomment-2652899234), this is platform's behavior, so we do the same.

`/programStageWorkingLists` can be used to store such operators in a working list, here's a sample payload:
```
{
    "name":"Test WL,
    "program":{"id":"IpHINAT79UW"},
    "programStage":{"id":"A03MvHHogjR"},
    "programStageQueryCriteria":{
        "dataFilters":[{"null":true,"dataItem":"UXz7xuGCEhU"}]
    }
}
```
In the payload, `dataFilters` is a JSON array, and the filter is stored in a JSONB field in the `programstageworkinglist` table. To address this, the null unary operator will be of type Boolean.

Also removing the join we do to `optionvalue` when fetching events, as it's purely to validate that, when the data element contains an option set, the value is present in the option value table too. This validation is done when importing, and since we are not validating the type of the data element either, I'm removing it.

### Next
Previously, the parsing logic for filters was shared between attributes and data elements. To accommodate the new behavior, I had to split it so that it applies only to data elements. In upcoming PR(s), I plan to extend this logic to attributes as well.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -51,6 +51,7 @@ public enum QueryOperator {
   IN("in", true),
   SW("sw"),
   EW("ew"),
+  EX("??"),
```
ameenhere, line 54: Instead of question marks, we can just use "ex" for this?

muilpp: The question marks are part of PostgreSQL syntax. They are used in the query to filter JSONB entries based on the presence of specific keys. I can't change them to something else.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -120,7 +122,8 @@ private FileResource getFileResourceMetadata(UID eventUid, UID dataElementUid)
               .orgUnitMode(OrganisationUnitSelectionMode.ACCESSIBLE)
               .events(Set.of(eventUid))
               .eventParams(EventParams.FALSE)
-              .dataElementFilters(Map.of(dataElementUid, List.of()))
+              .dataElementFilters(
```
teleivo, line 125: Does

`.dataElementFilters(Map.of(dataElementUid, List.of()))` still work but 

`Map.of(dataElementUid, List.of(new QueryFilter(QueryOperator.NNULL))))` is more clear in describing what we want here?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -51,6 +51,8 @@ public enum QueryOperator {
   IN("in", true),
   SW("sw"),
   EW("ew"),
+  NULL("??"),
```
teleivo, line 54:
```sql
eventdatavalues->'D7m8vpzxHDJ' is null
```

as well as `not null` works. I would favor using the `null` sql operator as it can be reused for non jsonb as well.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -65,6 +65,9 @@ public class EventDataFilter implements Serializable {
   /** Like */
   private String like;
 
+  /** Exists */
+  private String ex;
```
teleivo, line 69: Why are we using a different operator name in the working list? Am I right that the operator is `null/!null` in our exporter API?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1375,6 +1375,13 @@ private StringBuilder dataElementAndFiltersSql(
           .append(" as ")
           .append(deUid);
 
+      if (filterContainsOnlyUnaryOperator(filters)) {
```
teleivo, line 1378: Why do we need to treat only unary differently from cases below?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -592,23 +591,83 @@ void testExportEventsWhenFilteringByNumericDataElements()
...
+        EventOperationParams.builder()
+            .enrollments(UID.of("nxP7UnKhomJ", "TvctPPhpD8z"))
+            .programStage(programStage)
+            .eventParams(EventParams.FALSE)
+            .dataElementFilters(
+                Map.of(
+                    UID.of("DATAEL00002"),
+                    List.of(
+                        new QueryFilter(QueryOperator.NNULL),
+                        new QueryFilter(QueryOperator.NNULL))))
```
teleivo, line 663: Why would we allow/support the same unary operator for the same data element?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
```diff
@@ -353,6 +366,81 @@ void shouldParseFilterWhenFilterHasMultipleOperatorWithFinalColon() throws BadRe
...
+  @Test
+  void shouldParseDataElementFilterWhenSingleUnaryOperator() throws BadRequestException {
+    Map<UID, List<QueryFilter>> filters = parseDataElementFilters(DE_1_UID + ":!null");
+
+    assertEquals(Map.of(DE_1_UID, List.of(new QueryFilter(QueryOperator.NNULL))), filters);
+  }
+
+  @Test
+  void shouldParseDataElementFilterWhenMultipleUnaryOperatorsCombined() throws BadRequestException {
+    Map<UID, List<QueryFilter>> filters = parseDataElementFilters(DE_1_UID + ":!null:null");
```
teleivo, line 393: that should not be valid

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
```diff
@@ -353,6 +366,81 @@ void shouldParseFilterWhenFilterHasMultipleOperatorWithFinalColon() throws BadRe
...
+        Map.of(
+            DE_1_UID,
+            List.of(new QueryFilter(QueryOperator.NNULL), new QueryFilter(QueryOperator.NULL))),
+        filters);
+  }
+
+  @Test
+  void shouldParseDataElementFilterWhenUnaryAndBinaryOperatorsCombined()
+      throws BadRequestException {
+    Map<UID, List<QueryFilter>> filters = parseDataElementFilters(DE_1_UID + ":null:gt:10");
```
teleivo, line 405: this should also not be valid

teleivo (commented):
> Continuing the work started in [part 1](https://github.com/dhis2/dhis2-core/pull/19760), this PR adds a new `EX` operator, which filters events based on whether a specified data element has a value.

I am confused by you saying this adds operator `EX` as we discussed adding unary operators `null/!null`. Can you clarify that. Please add example requests/body to PR descriptions if you add query params or bodies. This is shorter and clearer than prose IMHO.

muilpp:
@teleivo I pushed a commit but I'm not done yet, just set it back to draft, but good to have some comments anyway

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -65,6 +65,12 @@ public class EventDataFilter implements Serializable {
   /** Like */
   private String like;
 
+  /** Null */
+  private String isNull;
```
ameenhere, line 69: to be consistent with the other attributes, simply having "null" as the property name would be better right?
Also for the opposite "notNull" maybe. 🤔

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
muilpp: yes, this is better, thanks

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
muilpp: as discussed [here](https://dhis2.slack.com/archives/C05S1NTHBJ9/p1739176522707899?thread_ts=1738232925.732909&cid=C05S1NTHBJ9), that's what plaftorm does and we are adopting the same approach

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
muilpp: see [here](https://dhis2.slack.com/archives/C05S1NTHBJ9/p1739176522707899?thread_ts=1738232925.732909&cid=C05S1NTHBJ9)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
muilpp: I agree, but "null" is a keyword in java, I could rename it to `nullValue` or `nullFilter`, or something else that might make more sense.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
muilpp: `.dataElementFilters(Map.of(dataElementUid, List.of()))` is something I introduced in part 1 of this PR as a temporary step.

In this PR, on [the web layer](https://github.com/dhis2/dhis2-core/pull/19778/files#diff-821fe85408531050058036e5a373fee939b3ca9f8ed35c8f09a7681f59a4c50cR406) I'm adding the `!null` operator in case a filter has a UID only, so we can't have an empty list as a value of this map on the service layer anymore.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: This one I doubted a little bit.
Do we care about the DE being an option set when the filter contains a unary operator?
If all we want to do is check whether something exists, shouldn't we just check the JSONB in events and forget about the `optionvalue` join?

teleivo:
> * `filter=UID:null:!null`

just for the record as its not mentioned: this is legal but obviously makes no sense. The same approach is used by platform. We only ensure the resulting SQL is valid. Validating the meaning is more involved as we would need to also deal with things like `:gt:10:lt:10`, ...

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
```diff
@@ -353,6 +366,88 @@ void shouldParseFilterWhenFilterHasMultipleOperatorWithFinalColon() throws BadRe
...
+
+  @Test
+  void
+      shouldFailParsingDataElementFiltersWhenUnaryAndBinaryOperatorsCombinedAndUnaryContainsValue() {
+    Exception exception =
+        assertThrows(
+            BadRequestException.class,
+            () -> parseDataElementFilters(DE_1_UID + ":gt:10:null:value"));
+    assertEquals(
+        "Operator in filter can't be used with a value: " + DE_1_UID + ":gt:10:null:value",
```
teleivo, line 438: Would be great if we can tell the user which operator is not allowed to have a value = unary. Same in above case.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidator.java
```diff
@@ -496,16 +596,33 @@ public static void validateUnsupportedParameter(
...
+      queryOperator = QueryOperator.fromString(operator);
     } catch (IllegalArgumentException exception) {
       throw new BadRequestException(INVALID_FILTER + filter);
     }
+
+    if (queryOperator == null) {
+      throw new BadRequestException(INVALID_FILTER + filter);
+    }
+
+    if (queryOperator.isUnary()) {
```
teleivo, line 614: Please tell the user which operator is used incorrectly. Always think about how you would fix an issue if you were the user given the info that you get here.

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/RequestParamsValidatorTest.java
```diff
@@ -353,6 +366,88 @@ void shouldParseFilterWhenFilterHasMultipleOperatorWithFinalColon() throws BadRe
...
+        filters);
+  }
+
+  @Test
+  void shouldFailParsingDataElementFiltersWhenUnaryOperatorContainsValue() {
+    Exception exception =
+        assertThrows(
+            BadRequestException.class, () -> parseDataElementFilters(DE_1_UID + ":!null:value"));
+    assertEquals(
+        "Query item or filter is invalid: " + DE_1_UID + ":!null:value", exception.getMessage());
```
teleivo, line 427: Can we be more specific instead of saying `item or filter` is invalid?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
```diff
@@ -592,23 +591,83 @@ void testExportEventsWhenFilteringByNumericDataElements()
...
-  void shouldFilterByEventsWithGivenDataValuesWhenFilterContainsDataElementUIDsOnly()
+  void shouldFilterByEventsContainingGivenDataValuesWhenFilteringByNonNullDataValues()
       throws ForbiddenException, BadRequestException {
     EventOperationParams params =
         EventOperationParams.builder()
+            .eventParams(EventParams.FALSE)
             .dataElementFilters(
                 Map.of(
                     UID.of("GieVkTxp4HH"),
-                    new ArrayList<>(),
```
teleivo, line 602: Did you decide not to improve the API of the EventOperationParams or will you do it later? Or is

`new ArrayList<>()` vs `List.of(new QueryFilter(QueryOperator.NNULL))`

not an issue anymore as we are dealing with it in the query params 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
teleivo: @enricocolasante I think I need your help here 😅 🤔 

@muilpp are you saying that this code is an optimization? So if all a user cares about is if an event has a data value for a DE than we skip the join to `optionvalue`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1412,7 +1419,10 @@ private StringBuilder dataElementAndFiltersSql(
           String bindParameter = "parameter_" + filterCount;
           int itemType = de.getValueType().isNumeric() ? Types.NUMERIC : Types.VARCHAR;
 
-          if (!de.hasOptionSet()) {
+          if (filter.getOperator().isUnary()) {
```
teleivo, line 1422: This feels all very convoluted 😅 Not due to your change but already! I wonder how we could simplify this. Should it not be possible to have one function with maybe a switch for the different filters returning the individual expressions of the where clause. Then one mapping all the filters into the where clause 🤷🏻 maybe something to think about for some other day

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: when filtering a DE that has an option set, a join like the one below is added to the query:
```
	inner join optionvalue as opt_1 on
		lower(opt_1.code) = lower(ev.eventdatavalues #>> '{AAaJGnWR5js, value}')
		and opt_1.optionsetid = :optset_1
		and lower(opt_1.name) = :parameter_2
```
`opt_1` is not used anywhere else, so it seems to me this is just to validate the value in the event DE exists in the corresponding `optionvalue` entry, which sounds like something we should do while importing, isn't it?

that said, what happens when a user deletes a certain option value? is that value deleted from events too? if it's not, then this join makes sense.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/EventExporterTest.java
muilpp: I still want to do it, but in a different PR

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -170,8 +170,10 @@ public String getNull() {
     return nullFilter;
   }
 
-  public void setNull(String isNull) {
-    this.nullFilter = isNull == null ? "" : isNull;
+  public void setNull(String nullFilter) {
```
teleivo, line 173: I thought this is going to be a boolean? 🤔

muilpp: yes but didn't we say we only want to store it when it's true?
if it was a boolean, it'd be false by default and stored in the database
if it's a String, we can just set it to null by default, so it'll be excluded by Jackson, unless we explicitly set it to "true"

teleivo: > yes but didn't we say we only want to store it when it's true?
I think so but why does that mean we accept/require `{"null": ""}`? I though we are going for the not perfect but slightly better `{"null": true}` if users want to filter for null.

@enricocolasante maybe I misremember 😅

enricocolasante: Didn't we agree that we accept any value there? `{"null": [ANYTHING_BUT_NULL]}`?
Then when saving we remove the value there and use `true`

ameenhere: Why can't we use Boolean (which is nullable). We use it for "followUp" as well.

ameenhere (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
ameenhere: Atleast the values would then be limited to only three valid values, true/false/null.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -65,6 +65,12 @@ public class EventDataFilter implements Serializable {
   /** Like */
   private String like;
 
+  /** Null */
+  private String nullFilter;
+
+  /** Not Null */
+  private String notNullFilter;
```
ameenhere, line 72: Maybe you already discussed this. But can't we simply reuse the nullFilter for both null and non null unary operator? FE anyway does a transformation.

ameenhere (changes requested):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
muilpp: If we leave it up to FE to figure it out, then yes, we can use just one field for both operators.

enricocolasante (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -158,6 +161,18 @@ public void setLike(String like) {
     this.like = like;
   }
 
+  @JsonProperty
+  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
+  public Boolean getNull() {
+    return isNull;
+  }
+
+  public void setNull(Boolean nullFilter) {
```
ameenhere, line 170: I see a mix of isNull and nullFilter here. Lets stick to nullFilter?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programstagefilter/EventDataFilter.java
```diff
@@ -158,6 +161,18 @@ public void setLike(String like) {
...
+    if (nullFilter != null) {
```
ameenhere, line 171: Do we really need this if construct?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1310,105 +1307,57 @@ private StringBuilder dataElementAndFiltersSql(
...
-          int itemType = de.getValueType().isNumeric() ? Types.NUMERIC : Types.VARCHAR;
+        final String queryCol =
+            de.getValueType().isNumeric()
+                ? castToNumber(dataValueValueSql)
+                : lower(dataValueValueSql);
 
-          if (!de.hasOptionSet()) {
-            eventDataValuesWhereSql.append(hlp.whereAnd());
+        String bindParameter = "parameter_" + filterCount;
+        int itemType = de.getValueType().isNumeric() ? Types.NUMERIC : Types.VARCHAR;
```
ameenhere, line 1319: This can be done above/outside the for loop right? Also suggesting itemValueType as the variable name?

ameenhere (approved):

PR #19780, opened 2025-01-27
fix: make min and max password length changeable

netroms:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/PasswordValidationRule.java
```diff
@@ -27,10 +27,19 @@
...
+import java.util.regex.Pattern;
+
 /** Created by zubair on 08.03.17. */
 @FunctionalInterface
 public interface PasswordValidationRule {
 
+  String DEFAULT_PASSWORD_VALIDATION_PATTERN =
+      "^(?=.*[A-Z])(?=.*[a-z])(?=.*\\d)(?=.*[\\W_])[A-Za-z\\d\\W_]{%d,%d}$";
+
+  Pattern META_DEFAULT_PASSWORD_VALIDATION_PATTERN =
```
david-mackessy, line 39: is this the same regex as the one above but formatted differently for some reason? Hard to follow from the name alone.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/startup/SettingUpgrader.java
```diff
@@ -39,9 +42,22 @@ public class SettingUpgrader extends AbstractStartupRoutine {
...
     if ("dhis-web-dashboard-integration".equals(startModule)) {
       settingsService.put("startModule", "dhis-web-dashboard");
     }
+
+    String passwordValidationPattern =
+        settingsService.getCurrentSettings().getPasswordValidationPattern();
+    // 1st startup DEFAULT_PASSWORD_VALIDATION_PATTERN is equal to the pattern.
+    // 2nd and subsequent startups META_DEFAULT_PASSWORD_VALIDATION_PATTERN is equal to the pattern,
+    // unless the pattern is set to a custom value.
+    if (passwordValidationPattern.equalsIgnoreCase(DEFAULT_PASSWORD_VALIDATION_PATTERN)
```
david-mackessy, line 54: do we change the default password validation somewhere else in the system after the 1st start-up? Do we make it stricter or use the META regex? (what's the reason for that?)
The if condition checks if either are are in use but always returns the DEFAULT. Are we sure that the DEFAULT & META are essentially the same?

vietnguyen (approved):

PR #19783, opened 2025-01-27
feat: include Uzbek variants in default DB locales

Philip-Larsen-Donnelly:
This adds Uzbek variants to the default database locales. 

The previous code parsed the default strings only one way; which could set the variants, but resulted in locales in the database that couldn't be handled by the core upon retrieval.

The changes here now parse the language, country, and variant parts separately, which results in a locale in the database that is supported without changes elsewhere.

Note that these changes do not affect any of the existing defaults, which remain the same, but only add the two new Uzbek variants in the supported format.
Note also that these defaults only apply to a blank database.

To Test:
1. Run against a blank database.
2. Check the list of default DB locales. They should include the two Uzbek variants (cyrillic and latin).
3. Add a metadata object (like, at least an org unit).
4. Add an Uzbek translation to the object

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/startup/I18nLocalePopulator.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/startup/I18nLocalePopulator.java
@@ -55,7 +55,8 @@ public class I18nLocalePopulator extends TransactionContextStartupRoutine {
   private static final Set<String> DEFAULT_LOCALES =
       Set.of(
           "af", "ar", "bi", "am", "de", "dz", "en", "es", "fa", "fr", "gu", "hi", "id", "it", "km",
-          "lo", "my", "ne", "nl", "no", "ps", "pt", "ru", "rw", "sw", "tg", "vi", "zh");
+          "lo", "my", "ne", "nl", "no", "ps", "pt", "ru", "rw", "sw", "tg", "uz__latn", "uz__cyrl", "vi", "zh"
+          );
 
   @Override
   public void executeInTransaction() {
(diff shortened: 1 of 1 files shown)


jbee (commented):
How do we know the rest of the system can actually handle the locales. There are plenty of `Locale` from `String` places in the code. Why would they now handle these correct when you had to introduce `parseLocaleString` here to make it work? I would think that if this hasn't produced an error yet for you, you just haven't tried the function that does. 

In any case we should add at least one test that goes through the scenario you describe in the issue description.

jbee:
@Philip-Larsen-Donnelly The way I see it is, either we do not need the custom parse function you added because the parsing for locales we have elsewhere can handle it, or, if it cannot handle it, adding these to the DB as you say is asking for errors later on when they do get parsed by the locale parsing that is in place elsewhere. If these have been inserted into DB manually without causing trouble then I suppose you don't need the custom parsing because apparently the locale parsing we have can handle it, or, as I say they just never invoked the operation what would fail parsing them. I would like to to know which one it is before we add this.

jason-p-pickering:
I wanted to see if we could get moving on this again as it came up in another conversation with someone, but just wanted to check here. Looking at the spec for [Locale](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html), I think we should be using "uz_Cyrl" or "uz_Latn" to specify either Uzbek written in Cyrillic or Latin characters. When making this change, they do appear to work out of the box

image


I think the potential problem with our internal Local object is that we ask for Language + Country but not Language + Script + Country. That might be another problem though....

Thoughts @jbee @Philip-Larsen-Donnelly ?

jason-p-pickering:
See https://github.com/dhis2/dhis2-core/pull/21417

PR #19789, opened 2025-01-28
chore:Move relationship handling into RelationshipService [DHIS2-18883]

enricocolasante:
Move relationship handling into `RelationshipService` (Remove all `getRelationshiItems` methods from entity service and centralize them in `RelationshipService`).
Now, all entity services depend on `RelationshipService` to retrieve relationships, and `RelationshipService` itself does not depend on any other entity service.

***Test***
In `TrackedEntitiesExportControllerTest` we are now using `TrackerObjects` to make assertions instead of Hibernate Objects. This way we do not have any issues with the session and we can never experience `LazyLoadInitialization` issues (and we also removed unnecessary DB calls to retrieve Hibernate Objects).

***Next steps***
- Continue to migrate and fix commented tests in `RelationshipsExportControllerTest`.
- Remove `RelationshipStore` interface.
- Check how we should deal with `deleted` entities in `RelationshipService`.
- Check if `trackerDataView` is working.

***Open Questions***
- When `assignedUser` is defined in the payload and the `programStage` has `enableUserAssignment=false` a warning is returned and the `assignedUser` is not persisted. Is this the expected behavior?

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
@@ -227,8 +227,8 @@ class DefaultEventService implements EventService {
       throws BadRequestException, ForbiddenException {
     EventQueryParams queryParams = paramsMapper.map(operationParams, getCurrentUserDetails());
     List<Event> events = eventStore.getEvents(queryParams);
-    for (Event event : events) {
-      if (operationParams.getEventParams().isIncludeRelationships()) {
+    if (operationParams.getEventParams().isIncludeRelationships()) {
+      for (Event event : events) {
         event.setRelationshipItems(
             relationshipService.getRelationshipItems(TrackerType.EVENT, UID.of(event)));
       }
@@ -243,8 +243,8 @@ class DefaultEventService implements EventService {
       throws BadRequestException, ForbiddenException {
     EventQueryParams queryParams = paramsMapper.map(operationParams, getCurrentUserDetails());
     Page<Event> events = eventStore.getEvents(queryParams, pageParams);
-    for (Event event : events.getItems()) {
-      if (operationParams.getEventParams().isIncludeRelationships()) {
+    if (operationParams.getEventParams().isIncludeRelationships()) {
+      for (Event event : events.getItems()) {
         event.setRelationshipItems(
             relationshipService.getRelationshipItems(TrackerType.EVENT, UID.of(event)));
       }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
@@ -218,11 +218,6 @@ class DefaultEventService implements EventService {
     }
     event.setEventDataValues(dataValues);
 
-    if (eventParams.isIncludeRelationships()) {
-      event.setRelationshipItems(
-          relationshipService.getRelationshipItems(TrackerType.EVENT, eventUid));
-    }
-
     return event;
   }
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
@@ -51,16 +51,15 @@ import org.hisp.dhis.fileresource.FileResource;
 import org.hisp.dhis.fileresource.FileResourceService;
 import org.hisp.dhis.fileresource.ImageFileDimension;
 import org.hisp.dhis.program.Event;
-import org.hisp.dhis.relationship.Relationship;
-import org.hisp.dhis.relationship.RelationshipItem;
 import org.hisp.dhis.tracker.TrackerIdScheme;
 import org.hisp.dhis.tracker.TrackerIdSchemeParam;
 import org.hisp.dhis.tracker.TrackerIdSchemeParams;
+import org.hisp.dhis.tracker.TrackerType;
 import org.hisp.dhis.tracker.acl.TrackerAccessManager;
 import org.hisp.dhis.tracker.export.FileResourceStream;
 import org.hisp.dhis.tracker.export.Page;
 import org.hisp.dhis.tracker.export.PageParams;
-import org.hisp.dhis.user.UserDetails;
+import org.hisp.dhis.tracker.export.relationship.RelationshipService;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -84,6 +83,8 @@ class DefaultEventService implements EventService {
   private final FileResourceService fileResourceService;
 
   private final EventOperationParamsMapper paramsMapper;
+
+  private final RelationshipService relationshipService;
 
   @Override
   public FileResourceStream getFileResource(@Nonnull UID event, @Nonnull UID dataElement)
@@ -161,25 +162,15 @@ class DefaultEventService implements EventService {
 
   @Override
   public Event getEvent(@Nonnull UID event) throws ForbiddenException, NotFoundException {
-    return getEvent(
-        event, TrackerIdSchemeParams.builder().build(), EventParams.FALSE, getCurrentUserDetails());
+    return getEvent(event, TrackerIdSchemeParams.builder().build(), EventParams.FALSE);
   }
 
   @Override
   public Event getEvent(
-      @Nonnull UID event,
+      @Nonnull UID eventUid,
       @Nonnull TrackerIdSchemeParams idSchemeParams,
       @Nonnull EventParams eventParams)
       throws ForbiddenException, NotFoundException {
-    return getEvent(event, idSchemeParams, eventParams, getCurrentUserDetails());
-  }
-
-  private Event getEvent(
-      @Nonnull UID eventUid,
-      @Nonnull TrackerIdSchemeParams idSchemeParams,
-      @Nonnull EventParams eventParams,
-      @Nonnull UserDetails user)
-      throws NotFoundException, ForbiddenException {
     Page<Event> events;
     try {
       EventOperationParams operationParams =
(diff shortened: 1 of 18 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -228,16 +219,8 @@ private Event getEvent(
     event.setEventDataValues(dataValues);
 
     if (eventParams.isIncludeRelationships()) {
```
teleivo, line 221: Can we move this if into the `getEvents(operationParams, new PageParams(1, 1, false))`? I think we can have one `get` method in every service that does interface with stores. The other `get` variants job is to create the `EventOperationParams` and call the one method.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -42,23 +43,47 @@
...
+          case TRACKED_ENTITY -> relationshipStore.getRelationshipItemsByTrackedEntity(uid);
+          case ENROLLMENT -> relationshipStore.getRelationshipItemsByEnrollment(uid);
+          case EVENT -> relationshipStore.getRelationshipItemsByEvent(uid);
+          case RELATIONSHIP -> throw new IllegalArgumentException("Unsupported type");
+        };
+    return relationshipItems.stream()
+        .filter(
+            ri ->
+                trackerAccessManager
+                    .canRead(CurrentUserUtil.getCurrentUserDetails(), ri.getRelationship())
```
teleivo, line 80: Is it still an open question if we should show a relationship with only one of the items in case the user does not have access to the other?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -197,6 +197,43 @@ public List<Relationship> getUidsByRelationshipKeys(List<String> relationshipKey
...
+  @Override
+  public List<RelationshipItem> getRelationshipItemsByTrackedEntity(UID trackedEntity) {
+    @Language("hql")
+    String hql =
+        """
+                from RelationshipItem ri
+                where ri.trackedEntity.uid = :trackedEntity
+                """;
+    return getQuery(hql, RelationshipItem.class)
+        .setParameter(TRACKED_ENTITY, trackedEntity.getValue())
```
teleivo, line 209: I have a dejavu 😅 I would use literals if we use templates that define the parameter without a constant as scanning `:trackedEntity` with `"trackedEntity" is easier than navigating to the constant and back. ok maybe an ide lets you hover but who wants to touch a 🐭 😂 .

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipService.java
```diff
@@ -35,11 +35,16 @@
...
 import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.relationship.Relationship;
+import org.hisp.dhis.relationship.RelationshipItem;
+import org.hisp.dhis.tracker.TrackerType;
 import org.hisp.dhis.tracker.export.Page;
 import org.hisp.dhis.tracker.export.PageParams;
 
 public interface RelationshipService {
 
+  /** Get all relationships matching given params. */
```
teleivo, line 45:
```suggestion
  /** Get all relationship items matching given params. */
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipStore.java
```diff
@@ -35,13 +35,20 @@
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.Event;
 import org.hisp.dhis.relationship.Relationship;
+import org.hisp.dhis.relationship.RelationshipItem;
 import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.tracker.export.Page;
 import org.hisp.dhis.tracker.export.PageParams;
 
 public interface RelationshipStore extends IdentifiableObjectStore<Relationship> {
```
teleivo, line 43: Can we get rid of that interface?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
```diff
@@ -478,152 +479,96 @@ void getRelationshipsByTrackedEntity() {
...
+  }
+
+  @Test
+  void shouldNotGetRelationshipsByTrackedEntityWhenRelationshipIsDeleted() {
+    assertNoRelationships(
+        GET("/tracker/relationships?trackedEntity={te}", "guVNoAerxWo").content(HttpStatus.OK));
+  }
+
+  @Test
+  void shouldNotRelationshipsByTrackedEntityWhenRelationshipIsDeleted() {
```
teleivo, line 499:
```suggestion
  void shouldGetRelationshipsByTrackedEntityWhenRelationshipIsDeleted() {
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
enricocolasante: Too many rebases and cherry-picks on my side :grimacing:

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
enricocolasante: Yes, this if is not necessary here at all. We are already mapping the relationships in the `getEvents` method

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipStore.java
enricocolasante: Not yet.
I will add another bullet point in `Next Steps`.
Nothing big but we are using the store directly from the deletion service and a supplier.
Will remove in another PR.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
enricocolasante: It is temporary closed :grimacing: 
It is something that in the future could make sense. For now we agreed that if the user has no access to one of the items then he has no access to the relationship

enricocolasante:
> What are the downsides of using `TrackerObjects`? I assume we cannot use them assert things like `createdAt` which are generated server-side and thus not in the payload.

We should have a chat about that.
I am not fully convinced that using `TrackerObjects` is that brilliant idea but it is simplifying a lot the setup of the test.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
ameenhere: I think so. But we should get back to this in the ACL big picture. 
Since we do not have anyway of showing "just a relationship, without what it relates to" in the Capture App, we can keep it this way. Not show the the relationship if no access to relationshipitem.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -42,23 +43,47 @@
...
+                    .isEmpty())
+        .filter(ri -> !ri.getRelationship().isDeleted())
```
ameenhere, line 82: Can the deleted check be the first filter? To avoid acl check filter on deleted relationships

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -248,7 +226,14 @@ private Event getEvent(
   public List<Event> getEvents(@Nonnull EventOperationParams operationParams)
       throws BadRequestException, ForbiddenException {
     EventQueryParams queryParams = paramsMapper.map(operationParams, getCurrentUserDetails());
-    return eventStore.getEvents(queryParams);
+    List<Event> events = eventStore.getEvents(queryParams);
+    for (Event event : events) {
+      if (operationParams.getEventParams().isIncludeRelationships()) {
```
ameenhere, line 231: cant this if construct be outside the for loop (before it) ?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/DefaultEventService.java
```diff
@@ -257,27 +242,14 @@ public Page<Event> getEvents(
...
-          getEvent(
-              uid,
-              TrackerIdSchemeParams.builder().build(),
-              EventParams.TRUE.withIncludeRelationships(false));
-    } catch (NotFoundException | ForbiddenException e) {
-      // events are not shown in relationships if the user has no access to them
-      return null;
+    Page<Event> events = eventStore.getEvents(queryParams, pageParams);
+    for (Event event : events.getItems()) {
+      if (operationParams.getEventParams().isIncludeRelationships()) {
```
ameenhere, line 247: if construct right before for loop?

ameenhere (approved):

PR #19791, opened 2025-01-28
[WAIT] fix: Issue with missing headers when using Doris analytics database

luciano-fiandesio:
## Summary  
This PR resolves an issue with non-aggregated analytics SQL queries in Doris/Clickhouse, specifically concerning their inability to process Postgis-related columns like latitude and longitude.

## Changes  
- Introduced the concept of a "virtual grid header" to address the mismatch.
  - A virtual `GridHeader` serves as a placeholder without corresponding to any column in the SQL result set.
  - It ensures the preservation of the original `GridHeaders` by providing default (empty) values in the grid when no underlying data is available.


larshelge:
I am wondering if it would be better to handle this client side, and have the client inspect the headers and show/hide the long/lat fields based on the presence in the headers.

PR #19792, opened 2025-01-28
Make `concat` function in Doris to work like Postgres

luciano-fiandesio:
## PR Description

This PR resolves compatibility issues with the `concat` function in Apache Doris, which differs from PostgreSQL in handling `NULL` values. In Doris (and MySQL), if any argument in `CONCAT()` is `NULL`, the entire result is `NULL`. 

To address this, dynamic column expressions are now wrapped with `TRIM(NULLIF(...))`, ensuring identical behavior across both databases.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/AbstractSqlBuilder.java
+++ b/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/AbstractSqlBuilder.java
@@ -303,4 +303,17 @@ public abstract class AbstractSqlBuilder implements SqlBuilder {
   protected <T> String toCommaSeparated(Collection<T> collection, Function<T, String> mapper) {
     return collection.stream().map(mapper).collect(Collectors.joining(","));
   }
+
+  /**
+   * Checks if the given input is quoted.
+   *
+   * @param input the input string.
+   * @return true if the input is quoted, false otherwise.
+   */
+  protected static boolean isQuoted(String input) {
+    return input != null
+        && input.length() >= 2
+        && input.startsWith(SINGLE_QUOTE)
+        && input.endsWith(SINGLE_QUOTE);
+  }
 }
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/AbstractSqlBuilder.java
```diff
@@ -303,4 +303,17 @@ protected String notSupported() {
...
     return collection.stream().map(mapper).collect(Collectors.joining(","));
   }
+
+  /**
+   * Checks if the given input is quoted.
+   *
+   * @param input the input string.
+   * @return true if the input is quoted, false otherwise.
+   */
+  protected static boolean isQuoted(String input) {
```
maikelarabori, line 313: Probably this method can be moved to org.hisp.dhis.commons.util.TextUtils.java

Or the logic could be extracted to that class (TextUtils) and invoked from this method.

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisSqlBuilder.java
```diff
@@ -470,6 +481,28 @@ public String createCatalog(String connectionUrl, String username, String passwo
             "driver_filename", driverFilename));
   }
 
+  /**
+   * Ensures `TRIM(NULLIF(...))` regardless of incoming column formatting
+   *
+   * @param column the column to be wrapped
+   * @return the wrapped column
+   */
+  private String wrapTrimNullIf(String column) {
```
maikelarabori, line 490:
```suggestion
  private String wrapTrimNullIf(@Nonnull String column) {
```

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisSqlBuilder.java
```diff
@@ -214,6 +216,15 @@ public String dateTrunc(String text, String timestamp) {
     return String.format("date_trunc(%s, %s)", timestamp, singleQuote(text));
   }
 
+  @Override
+  public String concat(String... columns) {
```
maikelarabori, line 220:
```suggestion
  public String concat(@Nonnull String... columns) {
```

maikelarabori: From `javax.annotation.Nonnull`

maikelarabori (approved):
Looks good @luciano-fiandesio, just a few minor suggestions.
Thx!

larshelge (approved):

PR #19801, opened 2025-01-29
chore: remove J2ME related code [DHIS2-18213]

teleivo:
> J2ME mobile application is now discontinued and will no longer work
[Release notes 2.35](https://docs.dhis2.org/en/implement/software-release-information/dhis2-core-releases/dhis-core-version-235/upgrade-notes.html#functionality) 

This PR finally removes all the backend code related to J2ME.

## Enum storage

The `ParserType` is stored using the enums ordinal numbers in the DB. This means we cannot simply remove the `J2ME_PARSER` as this changes the order/integer value of all the parser types following the `J2ME_PARSER`.

I chose to migrate the `parsertype` column from `integer` to `varchar`. This makes the parser type easier to maintain. We don't expect many `SMSCommands` so the larger storage and slower access times should not matter. The complexity of approaches like http://www.gabiaxel.com/2011/01/better-enum-mapping-with-hibernate.html using a stable and more efficient identifier are not justified.

## Deletion of `J2ME_PARSER` `SMSCommands`

The table layout (also discussed [here](https://dhis2.slack.com/archives/G4513PWGY/p1738255118398889)) does not allow using `on delete cascade`.

smstables dot

The `SMSCommand` has codes which are stored in the smscodes table. The junction table and references indicate many-to-many relationship between `smscommand` and `smscode`. I think in reality it's more a one (`SMSCommand`) to many (`SMSCode`) relationship. I say that due the pk and unique constraint of the `smscommandcodes` and the code. An `SMSCode` can only ever be used/related to one `SMSCommand`.

```
~# \d smscommandcodes
          Table "public.smscommandcodes"
 Column │  Type   │ Collation │ Nullable │ Default
════════╪═════════╪═══════════╪══════════╪═════════
 id     │ bigint  │           │ not null │
 codeid │ integer │           │ not null │
Indexes:
    "smscommandcodes_pkey" PRIMARY KEY, btree (id, codeid)
    "smscommandcodes_codeid_key" UNIQUE CONSTRAINT, btree (codeid)
Foreign-key constraints:
    "fk22ab87941234e10f" FOREIGN KEY (id) REFERENCES smscommands(smscommandid) ON DELETE CASCADE
    "fk22ab8794f99215ac" FOREIGN KEY (codeid) REFERENCES smscodes(smscodeid) ON DELETE CASCADE
```

`on delete cascade` will leave orphaned `smscodes` and `smsspecialcharacter` as it only deletes rows from the junction tables. This is why I have to "manually" delete rows from tables in topological order and capture and delete the `smscodes` and `smsspecialcharacter`.


larshelge (approved):

jason-p-pickering:
@teleivo Just in case you need to go further back in time to be 100% sure about the parser type, the historical repo (prior to the initial commit on GitHub) is available here https://github.com/jason-p-pickering/dhis2-history

teleivo:
> @teleivo Just in case you need to go further back in time to be 100% sure about the parser type, the historical repo (prior to the initial commit on GitHub) is available here https://github.com/jason-p-pickering/dhis2-history

thank you @jason-p-pickering !

there is a diff between the last commit in the history you shared

https://github.com/jason-p-pickering/dhis2-history/blob/7b93a56a14f438a3788abdc76e7e78b1b99c242a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/sms/parse/ParserType.java#L31-L40

and the initial version in this repo

https://github.com/dhis2/dhis2-core/blob/29c829d80835eee905090137747915b7b5b60b95/dhis-2/dhis-api/src/main/java/org/hisp/dhis/sms/parse/ParserType.java

```diff
diff -C 2 parser-type-9d2e684 parser-type-29c829d80
*** parser-type-9d2e684 2025-01-31 08:44:41.492082626 +0100
--- parser-type-29c829d80       2025-01-31 08:43:34.792078557 +0100
***************
*** 3,7 ****
  ALERT_PARSER
  UNREGISTERED_PARSER
- ANONYMOUS_PROGRAM_PARSER
  TRACKED_ENTITY_REGISTRATION_PARSER
  PROGRAM_STAGE_DATAENTRY_PARSER
--- 3,7 ----
  ALERT_PARSER
  UNREGISTERED_PARSER
  TRACKED_ENTITY_REGISTRATION_PARSER
  PROGRAM_STAGE_DATAENTRY_PARSER
+ EVENT_REGISTRATION_PARSER
```

but I don't think that this should influence the migration I am going to add in this PR. I mean these changes are so old so this change in ordinal values must already have been dealt with.

enricocolasante (approved):

PR #19804, opened 2025-01-29

cjmamo:


jbee (approved):

david-mackessy:
> @jbee do you know why this test keeps failing?
@cjmamo  It says why it's failing in the error message. The expected error message is different than the actual error message.
Possibly due to a library/JDK change that makes the web call during the test.
I'd suggest that you make the assert more lenient so it will pass a string assertion.
The test is also not in current master, so maybe it has been removed completely, you can check into that.

PR #19806, opened 2025-01-29
fix: Adjust data set expiry day comparison

jason-p-pickering:
It was noted that this comparison was not correct. Expiry days of 1.5 should mean that users have until 1.5 days to submit data after the data entry period has actually started. We were using the `endDate` of the period, whose which is always something like "Jan 31 2025 00:00:00". Since data entry cannot start until the end of this day (midnight the next day), we add a day to the calculation to account for this. 
Adjusted some tests to make it more clear and readable what is actually supposed to occur.


jbee (approved):

david-mackessy (approved):
This approval is purely trust based :) 
I don't know this part of the code well enough to be able to know what's going on with these dates.

PR #19808, opened 2025-01-29
fix: Add format Datetime method for Doris [DHIS2-16705]

luciano-fiandesio:
## Summary  
This PR addresses inconsistencies in datetime rendering between MySql/Apache Doris and Postgres JDBC drivers, ensuring uniform output across analytics databases.

## Changes  
- Standardized the conversion and rendering of `datetime` columns when fetched via the JDBC driver.
- Aligned the format for datetime representation with `2021-11-13 00:00:00.000` across different databases.

### DIFF
--- a/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisAnalyticsSqlBuilder.java
+++ b/dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisAnalyticsSqlBuilder.java
@@ -27,9 +27,23 @@
  */
 package org.hisp.dhis.db.sql;
 
+import java.time.LocalDateTime;
+import org.apache.commons.lang3.StringUtils;
+
 public class DorisAnalyticsSqlBuilder implements AnalyticsSqlBuilder {
   @Override
   public String getEventDataValues() {
     return "ev.eventdatavalues";
+  }
+
+  @Override
+  public String renderTimestamp(String timestampAsString) {
+    if (StringUtils.isEmpty(timestampAsString)) return null;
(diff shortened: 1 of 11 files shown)


dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisAnalyticsSqlBuilder.java
```diff
@@ -27,9 +27,23 @@
...
+
 public class DorisAnalyticsSqlBuilder implements AnalyticsSqlBuilder {
   @Override
   public String getEventDataValues() {
     return "ev.eventdatavalues";
   }
+
+  @Override
+  public String renderTimestamp(String timestampAsString) {
+    if (StringUtils.isEmpty(timestampAsString)) return null;
```
maikelarabori, line 41: I think it's safer to use StringUtils.isBlank in this case...

dhis-2/dhis-support/dhis-support-sql/src/main/java/org/hisp/dhis/db/sql/DorisAnalyticsSqlBuilder.java
```diff
@@ -27,9 +27,23 @@
...
+import org.apache.commons.lang3.StringUtils;
+
 public class DorisAnalyticsSqlBuilder implements AnalyticsSqlBuilder {
   @Override
   public String getEventDataValues() {
     return "ev.eventdatavalues";
   }
+
+  @Override
+  public String renderTimestamp(String timestampAsString) {
```
maikelarabori, line 40: Should we add a unit test for this method?

gnespolino (approved):

larshelge (approved):

PR #19829, opened 2025-01-31
feat: Add integrity check for datasets with same name or shortname

jason-p-pickering:
Adds an integrity check for datasets with the same name or short name. The shortname was made unique and not null [here](https://github.com/dhis2/dhis2-core/commit/ecf7d9990ec39aaeded74cd287633022c7769493) in 2023, but otherwise has been both nullable and non-unique since around 2013. I could not find a corresponding flyway migration script to enforce the uniqueness constraint @jbee , but maybe its there? Anyway, we could have situations where both the name and/or shortname are non-unique. The current thinking is that we should not change the name to being unique for the time being, but at least warn people that this may not be best practice.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks.yaml
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks.yaml
@@ -21,6 +21,7 @@ checks:
   - datasets/datasets_empty.yaml
   - datasets/datasets_not_assigned_to_org_units.yaml
   - datasets/datasets_custom_data_entry_forms_empty.yaml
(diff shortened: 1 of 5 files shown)


jbee (approved):

jbee:
> I could not find a corresponding flyway migration script to enforce the uniqueness constraint 

@jason-p-pickering I think you were looking for this
https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/v40/V2_40_16__Unique_non_null_short_names.java#L59

When it comes to making columns unique those flyway are always java scripts since we need an algorithm to handle collisions

david-mackessy (approved):

PR #19830, opened 2025-01-31
fix: Programinstance permanent deletion by first deleting sms entries

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-18693


zubaira:
these tables are not currently in use. So closing this PR.

PR #19831, opened 2025-01-31
fix: when changing email with email based 2FA enabled, fail fast

netroms:
## Summary
Throw an exception if a user with email-based 2FA enabled attempts to change their email address.
The user must be forced to disable their 2FA, before they can change their email address.
If this is not enforced, the consequence will be that the user no longer has a verified email, since the 2FA code only will be sent to a verified email address, the user will not be able to log in.

### Automatic tests
* TwoFactorControllerTest

### Manual test steps:
1. Enable email 2FA on a user
2. Try to change the user's email via /me endpoint
3. Observe you get an error
4. Try to change the user's email via /user endpoint, as admin
5. Observe you get an error


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/spring2fa/TwoFactorAuthenticationProvider.java
```diff
@@ -136,15 +136,14 @@ public Authentication authenticate(Authentication auth) throws AuthenticationExc
...
-        isTypeEnabled =
-            configurationProvider.getProperty(ConfigurationKey.EMAIL_2FA_ENABLED).equals("on");
+        isTypeEnabled = configurationProvider.isEnabled(ConfigurationKey.EMAIL_2FA_ENABLED);
       } else if (type == TwoFactorType.TOTP_ENABLED) {
-        isTypeEnabled =
-            configurationProvider.getProperty(ConfigurationKey.TOTP_2FA_ENABLED).equals("on");
+        isTypeEnabled = configurationProvider.isEnabled(ConfigurationKey.TOTP_2FA_ENABLED);
       }
 
       // Only validate 2FA code if the type is enabled
```
david-mackessy, line 149: I'm not familiar with this code, but it looks like there are many user-specific 2FA checks before actually checking if the system has 2FA enabled. If that's correct, would it be better if the system 2FA check was done first, to eliminate any unnecessary user 2FA checks?

Also, if 2FA is not enabled, why are we even in this class? (`TwoFactorAuthenticationProvider`).
I'd expect to know that 2FA is system-enabled before trying to authenticate with `TwoFactorAuthenticationProvider`.

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/security/spring2fa/TwoFactorAuthenticationProvider.java
netroms: Yeah, the last question "why are we in this class?" is a good one. This is because for legacy reasons they called the default authentication provider for TwoFactorAuthenticationProvider, so this is used for all authtentication, with 2FA and without 2FA.

netroms: I agree, it is probably ways to optimize the checks in the authenticate() method.

netroms: I've refactored the authenticate method as you suggested.

jbee (approved):

PR #19832, opened 2025-01-31
chore: Remove entities service dependency from relationship service [DHIS2-18883]

enricocolasante:
Our goal is to have every entity service depending only on its child services and always consider relationships as leaves in the dependency tree.
First step is to remove the dependency on `TrackedEntityService`, `EnrollmentService` and `EventService` that was present in `RelationshipOperationParamsMapper` and create methods in `RelationshipStore` to retrieve a single entity of any type using Hibernate.

***Next steps***
- Migrate and fix commented tests in `RelationshipsExportControllerTest`
- Move relationship handling into `RelationshipService` (Remove all `getRelationshiItems` methods from entity service and centralize them in `RelationshipService`)
- Pass field params to `RelationshipService` in order to only retrieve from the DB what is needed
- Check how we should deal with `deleted` entities in `RelationshipService`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
@@ -38,6 +38,7 @@ import jakarta.persistence.criteria.Subquery;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.LongSupplier;
@@ -47,6 +48,7 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.hisp.dhis.common.IdentifiableObject;
 import org.hisp.dhis.common.SoftDeletableObject;
 import org.hisp.dhis.common.SortDirection;
+import org.hisp.dhis.common.UID;
 import org.hisp.dhis.common.hibernate.SoftDeleteHibernateObjectStore;
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.Event;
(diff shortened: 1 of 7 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -88,6 +90,49 @@ public HibernateRelationshipStore(
     super(entityManager, jdbcTemplate, publisher, Relationship.class, aclService, true);
   }
 
+  @Override
```
teleivo, line 93: please remind me again 😅 what is the reason this store implements an interface?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -88,6 +90,49 @@ public HibernateRelationshipStore(
...
+  public Optional<TrackedEntity> findTrackedEntity(UID trackedEntity) {
+    @Language("hql")
+    String hql =
+        """
+                from TrackedEntity te
+                where te.uid = :trackedEntity
+                """;
+    List<TrackedEntity> trackedEntities =
+        getQuery(hql, TrackedEntity.class)
+            .setParameter(TRACKED_ENTITY, trackedEntity.getValue())
```
teleivo, line 103:
```suggestion
            .setParameter("trackedEntity", trackedEntity.getValue())
```

it's easier to read IMHO as the query itself has the parameter literal in the query string

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
enricocolasante: At the moment is still used in services/stores of other entities.
Maybe in next PRs we will remove the interface as it will only be used internally by relationships.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
enricocolasante: I completely agree.
Reversing it.
I automatically fixed the code smells here but it doesn't make sense in this case.

muilpp (approved):

PR #16782, opened 2024-03-14
docs: Add details to TrackedEntitiesExportController spec

Bekkalizer:

PR #19848, opened 2025-02-04
docs: clarify program is a tracker program in OpenAPI spec

teleivo:
TE part is taken from https://github.com/dhis2/dhis2-core/pull/16782 + clarifying its a tracker program

### DIFF
--- a/dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
+++ b/dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
@@ -18,7 +18,7 @@ Get a tracked entity with given UID.
 
 Get tracked entity with tracked entity attribute and enrollment data from the specified tracker
 program. The ownership of the given `program` will be used to determine access to the tracked
-entities.
+entity.
 
 ### `getTrackedEntityByUid.parameter.fields`
 
@@ -92,9 +92,9 @@ Get tracked entities using given organisation unit mode.
 
 ### `*.parameter.TrackedEntityRequestParams.program`
 
-Get tracked entity with tracked entity attribute and enrollment data from the specified tracker
+Get tracked entities with tracked entity attribute and enrollment data from the specified tracker
 program. The ownership of the given `program` will be used to determine access to the tracked
-entities.
+entities. Only tracked entities with an enrollment into the `program` will be returned.
 
 ### `*.parameter.TrackedEntityRequestParams.enrollmentStatus`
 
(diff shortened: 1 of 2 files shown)


enricocolasante (approved):

dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
```diff
@@ -78,7 +94,19 @@ Get tracked entities owned by given `orgUnit`.
...
 ### `*.parameter.TrackedEntityRequestParams.orgUnitMode`
 
-Get tracked entities using given organisation unit mode.
+Get tracked entities and enrollments using given `orgUnitMode` and `program` parameters. If a
+`program` is provided, the ownership is determied with this program. When no program is provided,
+the registration organisation unit for the tracked entity would be used to determine ownership.
+
+- When `orgUnitMode=SELECTED`, `orgUnitMode=CHILDREN` or `orgUnitMode=DESCENDANTS`, the `orgUnit`
+  parameter is required to specify which tracked entities or enrollments to return.
+- When `orgUnitMode=ALL` tracked entities or enrollments will be downloaded irrespective of the
```
teleivo, line 103: @muilpp reading https://dhis2.atlassian.net/browse/DHIS2-16624 the mode `ALL` can also be used by users with authority `ALL`. This also seems to be true if I follow this valdation 

https://github.com/dhis2/dhis2-core/blob/aa995696e7e75aacf9a08773e33fd8ce39ec3291/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java#L89-L94 

https://github.com/dhis2/dhis2-core/blob/80e3d059be24212c82d0842f5f9ff31be55a57b5/dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetailsImpl.java#L114-L115

So we should add this to the docs here right? Or is it assumed that users with authority `ALL` can do all the things anyway so we do not mention it every time 🤷🏻

larshelge (approved):

dhis-2/dhis-web-api/src/main/resources/openapi/TrackedEntitiesExportController.md
muilpp: It's my understanding that only a system administrator will have the authority `ALL`, and this user bypasses most of the validations anyway, that's why I didn't add it to the documentation.

PR #19849, opened 2025-02-04
fix: make totp url ascii only

netroms:
## Problem
Use of non ASCII characters in the TOTP URI fails.
This makes in impossible to use 2FA with instances that has non ASCII characters in the application title. 

### Cause
The QR code encoder does not support non ASCII characters. 

### Fix
Remove non ASCII characters from the URI.


larshelge:
Something seems off with this branch @netroms , has 965 files in the diff.

david-mackessy (approved):

PR #19858, opened 2025-02-05
feat: Sort TE change logs by attribute name [DHIS2-18971]

muilpp:
Three things in this PR:
- When sorting TE change logs by attribute, we currently sort based on the attribute `UID`, but we want to sort by the attribute name instead. As we do for event change logs, we first get the attribute form name, and if it's not set, we fall back to the attribute name.

- Additionally, I fixed a bug in event change log sorting. Our collation is case-sensitive, and I hadn’t taken that into account when I implemented the sorting for data elements.

- Explicitly adding the property `foreign-key="none"` to the Hibernate change log mappings to skip the creation of a foreign key to the user table, which is causing issues in Play Dev. It has worked for some months, and I can't reproduce it locally, but this, along with [Ivo's fix](https://github.com/dhis2/dhis2-core/pull/19856), should be enough to fix the issue.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
@@ -36,7 +36,7 @@
 
     <property name="createdByUsername" column="createdby" type="string"/>
 
-    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" foreign-key="none"  property-ref="username">
+    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" foreign-key="none" property-ref="username">
       <column name="createdby" />
     </many-to-one>
   </class>
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
@@ -36,7 +36,7 @@
 
     <property name="createdByUsername" column="createdby" type="string"/>
 
-    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" property-ref="username">
+    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" foreign-key="none"  property-ref="username">
       <column name="createdby" />
     </many-to-one>
   </class>
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
@@ -36,7 +36,7 @@
 
     <property name="createdByUsername" column="createdby" type="string"/>
 
-    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" foreign-key="none" property-ref="username">
+    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" property-ref="username">
       <column name="createdby" />
     </many-to-one>
   </class>
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
@@ -58,9 +58,9 @@ public class HibernateEventChangeLogStore {
   private static final String COLUMN_CHANGELOG_DATA_ELEMENT = "d.uid";
   private static final String COLUMN_CHANGELOG_FIELD = "ecl.eventField";
   private static final String ORDER_CHANGE_EXPRESSION =
-      "CONCAT(COALESCE(d.formName, ''), COALESCE(d.name, ''), COALESCE("
+      "CONCAT(COALESCE(LOWER(d.formName), ''), COALESCE(LOWER(d.name), ''), COALESCE(LOWER("
           + COLUMN_CHANGELOG_FIELD
-          + ", ''))";
+          + "), ''))";
   private static final String DEFAULT_ORDER =
       COLUMN_CHANGELOG_CREATED + " " + SortDirection.DESC.getValue();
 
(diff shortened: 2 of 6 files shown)


enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -58,9 +58,9 @@ public class HibernateEventChangeLogStore {
   private static final String COLUMN_CHANGELOG_DATA_ELEMENT = "d.uid";
   private static final String COLUMN_CHANGELOG_FIELD = "ecl.eventField";
   private static final String ORDER_CHANGE_EXPRESSION =
-      "CONCAT(COALESCE(d.formName, ''), COALESCE(d.name, ''), COALESCE("
+      "CONCAT(COALESCE(LOWER(d.formName), ''), COALESCE(LOWER(d.name), ''), COALESCE(LOWER("
```
teleivo, line 61: > Additionally, I fixed a bug in event change log sorting. Our collation is case-sensitive, and I hadn’t taken that into account when I implemented the sorting for data elements.

Can you please elaborate? How is it supposed to behave vs how was it behaving?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityChangeLogStore.java
```diff
@@ -66,12 +68,12 @@ public class HibernateTrackedEntityChangeLogStore {
...
           entry("createdAt", COLUMN_CHANGELOG_CREATED),
           entry("username", COLUMN_CHANGELOG_USER),
-          entry("attribute", COLUMN_CHANGELOG_DATA_ELEMENT));
+          entry("attribute", ORDER_ATTRIBUTE_EXPRESSION));
 
   private static final Map<Pair<String, Class<?>>, String> FILTERABLE_FIELDS =
       Map.ofEntries(
           entry(Pair.of("username", String.class), COLUMN_CHANGELOG_USER),
-          entry(Pair.of("attribute", UID.class), COLUMN_CHANGELOG_DATA_ELEMENT));
+          entry(Pair.of("attribute", UID.class), COLUMN_CHANGELOG_DATA_ELEMENT_UID));
```
teleivo, line 76: Is data element a copy paste remnant? I would expect this to be attribute uid 🤔

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
```diff
@@ -530,7 +530,7 @@
       "lastUpdated": "2015-03-31T11:22:51.642",
       "name": "Height in cm",
       "shortName": "Height in cm",
-      "formName": "Height in cm",
+      "formName": "height in cm",
```
teleivo, line 533: Why is that necessary? I would expect real data to read `"Height in cm"` in the UI

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
muilpp: The DB collation is case sensitive, so the issue was that a list like `Clinic 1, clinic 2, Clinic 3`, would be sorted like this: `Clinic 1, Clinic 3, clinic 2`.

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/simple_metadata.json
muilpp: I'm not testing how it's shown in the UI, I'm just testing that even if the form name starts with a lowercase character, it's sorted correctly when I get the change logs.

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
```diff
@@ -36,7 +36,7 @@
 
     <property name="createdByUsername" column="createdby" type="string"/>
 
-    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" property-ref="username">
+    <many-to-one name="createdBy" class="org.hisp.dhis.user.User" insert="false" update="false" foreign-key="none" property-ref="username">
```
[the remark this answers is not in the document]
muilpp: That makes sense, but then let's remove the others like [this one](https://github.com/dhis2/dhis2-core/pull/19858/files#diff-19b1fa3684c4351530da91648da0b26a0f8387bec4d7b57cb26fbfb37d6ab3b3R19) too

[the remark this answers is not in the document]
teleivo: I don't know what you refer to as your link points to this line 😅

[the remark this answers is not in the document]
teleivo: and keep in mind that Viet is working on removing the hbm.xml files

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
teleivo: Maybe worth discussing https://www.postgresql.org/docs/current/collation.html in the backend meeting. Is that always expected? if so why do we not configure the DB instead. Or is this something implementation admins should configure

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/EventChangeLog.hbm.xml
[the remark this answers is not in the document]
muilpp: I meant every other place where we mention constraints, like line 19 in this file:
```
    <many-to-one name="dataElement" class="org.hisp.dhis.dataelement.DataElement" column="dataelementid"
foreign-key="fk_eventchangelog_dataelementid" />
```
Fine, but if we decide to keep it, we can achieve the same result with annotations

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
muilpp: Yes, this I'll bring up next week.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/OrderAndFilterTrackedEntityChangeLogTest.java
```diff
@@ -163,18 +163,24 @@ void shouldSortChangeLogsWhenOrderingByAttributeAsc()
...
+            .toList();
 
-    assertAll(
-        () ->
-            assertUpdate(trackedEntityAttribute, "88", updatedValue, changeLogs.getItems().get(0)),
-        () -> assertCreate(trackedEntityAttribute, "88", changeLogs.getItems().get(1)),
-        () -> assertCreate("toUpdate000", "summer day", changeLogs.getItems().get(2)));
+    assertEquals(
+        List.of(
+            "numeric-attribute",
```
teleivo, line 178: I wonder if there is a way to more clearly tie what you expect to the setup. Before it was easier for me to see that you expect update, create, ... via the assertions

teleivo (approved):

PR #19868, opened 2025-02-06
feat: Event scheduling using program rule [DHIS2-16911]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-16911


dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionType.java
```diff
@@ -88,7 +88,8 @@ public enum ProgramRuleActionType {
           ERRORONCOMPLETE,
           WARNINGONCOMPLETE,
           ASSIGN,
-          SETMANDATORYFIELD);
+          SETMANDATORYFIELD,
+          CREATEEVENT);
```
ameenhere, line 94: Are we supporting creating an event (or only scheduling an event?)

zubaira: only scheduling. But scheduling is actually creation with `eventStatus=SCHEDULE`

zubaira: Moreover, this action type already existed, but it had not been implemented until now.

ameenhere: I kind of want it to be more explicit and name the rule action SCHEDULEEVENT, but will see what others think.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+    return query.getResultList();
+  }
+
+  private List<String> getProgramStageIds(TrackerPreheat preheat) {
+    List<Program> programs = preheat.getAll(Program.class);
+    if (programs.isEmpty()) {
+      return Collections.emptyList();
+    }
+
+    Set<String> programUids = programs.stream().map(Program::getUid).collect(Collectors.toSet());
```
enricocolasante: Why do you need programs here?
You can just remove the join with programs and the check of uid in programs and everything will work the same.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
```diff
@@ -0,0 +1,106 @@
...
+      return;
+    }
+
+    getProgramStages(programStageUids).stream()
+        .map(ProgramStageMapper.INSTANCE::map)
+        .filter(Objects::nonNull)
+        .forEach(preheat::put);
+  }
+
+  private List<ProgramStage> getProgramStages(List<String> programStageUids) {
```
enricocolasante: You can get program stages directly from the other query, you don't need to get the uids and then query again to get the objects.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/enrollment/CreateEventExecutor.java
```diff
@@ -0,0 +1,87 @@
...
+  public Optional<ProgramRuleIssue> executeRuleAction(TrackerBundle bundle, Enrollment enrollment) {
+    if (!DateUtils.dateIsValid(scheduledAt)) {
+      return Optional.of(ProgramRuleIssue.warning(programRule, ValidationCode.E1319, scheduledAt));
+    }
+
+    LocalDate localDate = LocalDate.parse(scheduledAt);
+    TrackerEvent scheduledEvent = new TrackerEvent();
+    scheduledEvent.setEvent(UID.generate());
+    scheduledEvent.setEnrollment(enrollment.getEnrollment());
+    scheduledEvent.setProgramStage(MetadataIdentifier.ofUid(programStage.getValue()));
```
enricocolasante:
```suggestion
    scheduledEvent.setProgramStage(MetadataIdentifier.ofUid(programStage));
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
enricocolasante, line 1: This supplier is getting all program stages without checking if the user has access to them or not.
We should use `AbstractSchemaStrategy` to correctly get metadata, to do so, we should collect program stage uids in `TrackerIdentifierCollector` similar to how it is done for other metadata in `collectProgramRulesFields`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -124,6 +121,12 @@ public ImportReport importTracker(
       validationReport = ValidationReport.merge(validationResult, result);
     }
 
+    jobProgress.startingStage("Calculating Payload Size");
+    final ValidationReport reportForSize = validationReport;
```
enricocolasante, line 126: What does this assignment do?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -155,12 +158,21 @@ private boolean exitOnError(ValidationReport validationReport, TrackerImportPara
     return validationReport.hasErrors() && params.getAtomicMode() == AtomicMode.ALL;
   }
 
-  private Map<TrackerType, Integer> calculatePayloadSize(TrackerBundle bundle) {
-    return Map.of(
-        TrackerType.TRACKED_ENTITY, bundle.getTrackedEntities().size(),
-        TrackerType.ENROLLMENT, bundle.getEnrollments().size(),
-        TrackerType.EVENT, bundle.getEvents().size(),
-        TrackerType.RELATIONSHIP, bundle.getRelationships().size());
+  private Map<TrackerType, Integer> calculatePayloadSize(
```
enricocolasante, line 169: I cannot understand what is happening here.
Can we make the code more intuitive or add some docs?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/Each.java
```diff
@@ -77,7 +78,8 @@ public static <T, R> Each<T, R> each(
 
   @Override
   public void validate(Reporter reporter, TrackerBundle bundle, T input) {
-    for (R in : map.apply(input)) {
+    Collection<R> items = new ArrayList<>(map.apply(input)); // defensive copy
+    for (R in : items) {
```
enricocolasante: Why do we need that?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/event/CreateEventExecutor.java
```diff
@@ -0,0 +1,87 @@
...
+    scheduledEvent.setProgramStage(MetadataIdentifier.ofUid(programStage.getValue()));
+    scheduledEvent.setAttributeOptionCombo(event.getAttributeOptionCombo());
+    scheduledEvent.setProgram(event.getProgram());
+    scheduledEvent.setOrgUnit(event.getOrgUnit());
+    scheduledEvent.setOccurredAt(null);
+    scheduledEvent.setScheduledAt(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant());
+    scheduledEvent.setStatus(EventStatus.SCHEDULE);
+    List<TrackerEvent> trackerEvents = new ArrayList<>(bundle.getTrackerEvents());
+    trackerEvents.add(scheduledEvent);
+    bundle.setTrackerEvents(Collections.unmodifiableList(trackerEvents));
```
enricocolasante: This event is not passing through validation, for example what if program stage is not repeatable and there is already an event in the enrollment?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionType.java
enricocolasante: Maintenance App is not supporting CREATEEVENT anyway, I would prefer to have a new SCHEDULEEVENT action as well.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/event/CreateEventExecutor.java
zubaira: Firstly, the event is not being created by a user—it is generated by the system. So, do we still need to apply validation in this case?

Regarding existing events for a non-repeatable programStage, I’ve noted in the PR description that the next step will be to implement logic to prevent duplicate event creation.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
zubaira: In a system with multiple programs, removing this filter would cause the query to return ProgramStages from all programs, not just those relevant to the current context. For example, if TrackerPreheat is initialized with only a single program, this query could still fetch stages from other unrelated programs. 

let me know if you'd still prefer to remove the join with Program.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
zubaira: resolved

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionType.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
zubaira: Because validationReport is neither final nor effectively final, it can't be used directly within a lambda expression

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/Each.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
zubaira: added docs

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageSupplier.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/event/CreateEventExecutor.java
enricocolasante: We need to review all the validators and check if there is something that could be invalid even if the system is creating the event.
For example, are we sure that the user has write access to the program stage?
And, how do we know that the event is a tracker event? Could it not be a single event?

zubaira: Theoretically, yes — it could be a single event (non-tracker). However, from a practical standpoint, this feature doesn’t apply meaningfully to single events. Since single events don’t involve enrollments and can’t have more than one occurrence, event scheduling isn’t relevant in that context.

Regarding validation, there are three possible approaches moving forward:

1. Re-run the validation phase after rule-engine validation
2. Adjust the execution order so that the validation phase runs after rule-engine validation.
3. Manually check the user's access to metadata in `ScheduleEventExecutor`

We can discuss which option aligns better with the overall flow and performance considerations.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleActionStore.java
```diff
@@ -103,4 +103,16 @@ public List<ProgramRuleAction> getByDataElement(Collection<DataElement> dataElem
         .setParameter("dataElements", dataElements)
         .list();
   }
+
+  @Override
+  public List<String> getProgramStagesPresentInProRuleActions(ProgramRuleActionType type) {
```
enricocolasante:
```suggestion
  public List<String> getProgramStagesPresentInProgramRuleActions(ProgramRuleActionType type) {
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/engine/DefaultProgramRuleEntityMapperService.java
```diff
@@ -177,6 +178,11 @@ private RuleAction toRuleAction(@Nonnull ProgramRuleAction pra) {
                   getAssignedParameter(pra),
                   ATTRIBUTE_TYPE,
                   getAttributeType(pra).name()));
+      case SCHEDULEEVENT ->
+          new RuleAction(
+              pra.getData(),
+              SCHEDULEEVENT.name(),
+              createValues(CONTENT, pra.getContent(), FIELD, pra.getProgramStage().getUid()));
```
enricocolasante: We do not need and use `CONTENT` here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/config/ServiceConfig.java
```diff
@@ -106,7 +106,7 @@ public Map<ProgramRuleActionType, ProgramRuleActionValidator> programRuleActionV
             ProgramRuleActionType.HIDEFIELD,
             getProgramRuleActionValidatorByClass(BaseProgramRuleActionValidator.class))
         .put(
-            ProgramRuleActionType.CREATEEVENT,
+            ProgramRuleActionType.SCHEDULEEVENT,
```
enricocolasante, line 110: Can we remove `CREATEEVENT` action?
Are the validation for `CREATEEVENT` enough for our use case?
In particular, do we validate that the program stage is part of the program? We should validate that, right?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleActionStore.java
```diff
@@ -103,4 +103,16 @@ public List<ProgramRuleAction> getByDataElement(Collection<DataElement> dataElem
...
+    String sql =
+        """
+            select distinct ps.uid
+            from ProgramRuleAction pra
+            join pra.programStage ps
+            where pra.programRuleActionType = :actionType and pra.programStage is not null
```
enricocolasante: This join should translate to an inner join, we do not need `and pra.programStage is not null`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerIdentifierCollector.java
```diff
@@ -98,6 +101,14 @@ private void collectProgramRulesFields(Map<Class<?>, Set<String>> map) {
         .forEach(attribute -> addIdentifier(map, TrackedEntityAttribute.class, attribute));
   }
 
+  private void collectProgramStagesPresentInRuleActions(Map<Class<?>, Set<String>> map) {
+    // Adds to the preheat cache all programStages that may be scheduled as a result of
+    // SCHEDULEEVENT rule action
+    programRuleActionService
```
enricocolasante, line 102: Would it makes sense to have this code in `collectProgramRulesFields` method and maybe have `getProgramStagesPresentInProRuleActions` implemented in `ProgramRuleService`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/event/CreateEventExecutor.java
enricocolasante: We need to tackle this 2 points:
1. If CREATEEVENT doesn't make sense for single events, we need to block it in that case. We could:
  - Validate the action in the metadata importer to not be created for programs without registration (best way to go)
  - We can validate in the executor and send a warning and effectively ignore the action.
2. We need validation around the program stage for the SCHEDULEEVENT action:
 - Do user have write access to program stage? (must be done in tracker importer, executor or validation phase)
 - Is program stage part of program? (could be done in metadata importer or tracker importer)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleActionStore.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerIdentifierCollector.java
zubaira: resolved. 
Added 3 methods in `programRuleActionService` as we are fetching data from `programruleaction` table
```

- getProgramStagesPresentInProgramRuleActions
- getDataElementsPresentInProgramRuleActions
- getTrackedEntityAttributesPresentInProgramRuleActions

```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/hibernate/HibernateProgramRuleActionStore.java
zubaira: resolved

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/action/validation/ScheduleEventProgramRuleActionValidator.java
```diff
@@ -0,0 +1,101 @@
...
+    if (programStage == null && programRuleAction.hasProgramStage()) {
+      programStage =
+          validationContext
+              .getProgramRuleActionValidationService()
+              .getProgramStageService()
+              .getProgramStage(programRuleAction.getProgramStage().getUid());
+    }
+
+    if (programStage == null) {
+      log.debug("ProgramStage is null in ruleAction for rule: {}", rule.getName());
```
enricocolasante: Why do you need to log this failing validations in the logs?
They are coming from the user payload, I am not sure I get the value to have them stored in the logs.
If it is failing, you could easily see it from the result of the request.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/DefaultProgramRuleActionService.java
```diff
@@ -96,4 +96,25 @@ public List<ProgramRuleAction> getProgramRuleActionsWithNoStageId() {
     return programRuleActionStore.getMalFormedRuleActionsByType(
         ProgramRuleActionType.HIDEPROGRAMSTAGE);
   }
+
+  @Override
+  @Transactional(readOnly = true)
+  public List<String> getProgramStagesPresentInProgramRuleActions() {
```
enricocolasante: This method is specific for `SCHEDULEDEVENT` action type, could we put it in the name so we also remove the need for a comment in the dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/TrackerIdentifierCollector.java?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
```diff
@@ -0,0 +1,104 @@
...
+      MetadataIdentifier attributeOptionCombo,
+      MetadataIdentifier program,
+      MetadataIdentifier orgUnit) {
+
+    final String scheduledAt = validationEffect.data();
+    final UID programRule = validationEffect.rule();
+    final UID programStage = validationEffect.field();
+
+    if (!DateUtils.dateIsValid(scheduledAt)) {
+      return Optional.of(ProgramRuleIssue.warning(programRule, ValidationCode.E1319, scheduledAt));
```
enricocolasante, line 68: Should we validate this when creating the rule action, like the other validations around program stage?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (!canWriteToProgramStage) {
+      return Optional.of(
+          ProgramRuleIssue.warning(
+              programRule,
+              ValidationCode.E1095,
```
enricocolasante: I am not sure we are giving enough context to the user to understand what is going on.
In the report, the user will receive a warning on an event that is not present in the payload and we are not saying anywhere that it is coming from a program rule.
I would create a new `ValidationCode` explaining exactly what is going on.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/RuleEngineTests.java
```diff
@@ -171,6 +170,78 @@ public void shouldShowErrorOnCompleteInTrackerEvents() throws Exception {
         .body("message", hasItem(stringContainsInOrder("ERROR ON COMPLETE ")));
   }
 
+  @Test
```
enricocolasante: Can we add a test to simulate when the user do not have access to the program stage present in the program rule action?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
zubaira: We're not certain whether the user will provide a plain date, a `d2`: function, or a variable (e.g., `V{current_date}`).
So validation can only occur after the expression is fully evaluated and transformed into an actual date.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
zubaira: added new `ValidationCode`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionType.java
```diff
@@ -54,7 +54,7 @@ public enum ProgramRuleActionType {
   WARNINGONCOMPLETE("warningoncomplete"),
   SHOWERROR("showerror"),
   ERRORONCOMPLETE("erroroncomplete"),
-  CREATEEVENT("createevent"),
```
ameenhere, line 57: Remind me, what are the consequences for this renaming an existing action type? Will there be instances out there that has the create event action type somewhere in their DB which would start to fail? 

I know we never supported it yet. But just thinking, if some values exist in the DB?

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/RuleEngineTests.java
zubaira: fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionType.java
zubaira: Yes, it's possible since we're not restricting the creation of CREATEEVENT via APIs (only UI does not support it). Added CREATEEVENT again

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionService.java
```diff
@@ -30,6 +30,7 @@
 package org.hisp.dhis.programrule;
 
 import java.util.List;
+import org.springframework.transaction.annotation.Transactional;
```
enricocolasante: Transactional shouldn't be declared in the interface, only in the implementation.
We can remove it here

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/RuleEngineTests.java
```diff
@@ -210,7 +209,6 @@ public void shouldAssignValue() {
   }
 
   @Test
-  @Disabled("TODO: fix this test 12098")
```
enricocolasante: Why can we enable this test now?

enricocolasante (commented):
For me now the PR is in good shape but we should have a clear plan on how to avoid duplication of scheduled events.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/RuleEngineTests.java
zubaira: Since [this](https://dhis2.atlassian.net/browse/DHIS2-12098) is now fixed, I tried enabling it and it worked — so why not keep it enabled?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleActionService.java
zubaira: Totally agree. This was a copy-paste oversight, as it was originally present in the previous interface [here](https://github.com/dhis2/dhis2-core/blob/d296cd979dbaf47c567279997ed4d52a26794c9a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleService.java#L71). I'll remove it.

zubaira:
The changes in this PR were covered in [another](https://github.com/dhis2/dhis2-core/pull/21433) separate PR. So closing this one.

PR #19872, opened 2025-02-06
DHIS2-18568: allow Route response timeout to be customised

cjmamo:
* Replaced blocking route HTTP client with non-blocking client to remove connection pool
* Allowing route response timeout to be customised
* Stream request and response bodies
* Support binary payloads


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -85,7 +86,9 @@ public class RouteService {
   @Qualifier(AES_128_STRING_ENCRYPTOR)
   private final PBEStringCleanablePasswordEncryptor encryptor;
 
-  @Autowired @Getter @Setter private RestTemplate restTemplate;
+  @Autowired @Getter @Setter private ClientHttpConnector clientHttpConnector;
+
+  @Autowired @Getter @Setter private ObjectMapper objectMapper;
```
david-mackessy, line 91: I don't think `@Autowired` is needed when using `@RequiredArgsConstructor`. Just make them `private final`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
```diff
@@ -67,6 +67,9 @@ public class Route extends BaseIdentifiableObject implements MetadataObject {
   /** Optional. Required authorities for invoking the route. */
   @JsonProperty private List<String> authorities = new ArrayList<>();
 
+  @JsonProperty(defaultValue = "10")
+  private Integer responseTimeout;
```
david-mackessy, line 71: Is there a specific reason for using `Integer` v `int`?
It looks like we have a DB constraint to always check the value is valid. So we could avoid null checks.

amcgee (approved):
LGTM!  Would be great to get one more review from the backend team

dhis-2/dhis-api/src/main/java/org/hisp/dhis/route/Route.java
```diff
@@ -48,6 +48,7 @@
 @EqualsAndHashCode(callSuper = true)
 @Accessors(chain = true)
 public class Route extends BaseIdentifiableObject implements MetadataObject {
+  public static final String DEFAULT_RESPONSE_TIMEOUT_SECONDS = "5";
```
jbee, line 51: FYI: You can declare the constant as an `int` and use `""+DEFAULT_RESPONSE_TIMEOUT_SECONDS` in the annotation to turn it into a string.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -224,24 +202,58 @@ public ResponseEntity<String> execute(
...
+            .toEntity(String.class)
+            .onErrorReturn(
+                throwable -> throwable.getCause() instanceof ReadTimeoutException,
+                new ResponseEntity<>(HttpStatus.GATEWAY_TIMEOUT))
+            .onErrorResume(
+                throwable -> throwable.getCause() instanceof DataBufferLimitException,
+                throwable -> {
+                  JsonObject message =
+                      Json.object(
+                          jsonObjectBuilder ->
```
jbee, line 250: I'd name this just `obj` or `object` because in the scope where the variable is valid it reads much better to pretend this is already an object. The "building" will happen implicitly at the end of the scope so it doesn't add anything to make the reader aware that the interface you are using is in fact a builder.

jbee (approved):

vietnguyen (approved):

PR #19873, opened 2025-02-06
feat: add hibernate.show_sql to dhis.conf

vietnguyen:


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -135,6 +135,12 @@ public enum ConfigurationKey {
   /** Sets 'hibernate.cache.use_query_cache'. (default: true) */
   USE_QUERY_CACHE("hibernate.cache.use_query_cache", "true", false),
 
+  /**
+   * Show SQL statements generated by hibernate in the log. Can be 'true', 'false'. (default: false)
+   * This will also enable hibernate.format_sql and hibernate.highlight_sql
+   */
+  HIBERNATE_SHOW_SQL("hibernate.show_sql", "false", false),
```
teleivo, line 142: How is this different from changing the log4j2 config to enable logging 🤔

teleivo: https://github.com/dhis2/dhis2-core/blob/c27640eae8b7ecdf6f2591515a20033b0f2c607b/dhis-2/dhis-test-integration/src/test/resources/log4j2-test.xml#L60-L79

vietnguyen: @teleivo  Sometimes changing the log4j config doesn't work for me. Have you ever experienced that ?

teleivo: Mhm, I mean our log setup is pretty complex 😅 You need to configure the correct file depending on what/how you execute code. Maybe share your use case and we'll figure it out together. I just wonder if adding another way is not more confusing/complex 🤔 also if you want to see jdbc logs you'll need to go to log4j2.

vietnguyen: My use case is just to know what queries hibernate are generating in test and also production code.

I think using the log4j config is more complicated, and I also have to remember revert the changed before merging.

Using dhis.conf is straight forward, and i just need to change my local file. Maybe it's easier for debugging deployed instance, you don't need to look for the log4j file in the built folder. 

I don't think there is a confusion regarding adding two ways to enable hibernate debug. It just helps to make the debugging job easier.

teleivo: > Using dhis.conf is straight forward, and i just need to change my local file. Maybe it's easier for debugging deployed instance, you don't need to look for the log4j file in the built folder.

I think that is due to a lack of understanding logs 😅 Users can provide their config. They do not need to touch the config we ship by default.

> I don't think there is a confusion regarding adding two ways to enable hibernate debug. It just helps to make the debugging job easier.

I disagree but maybe others agree. With this you now have 2 options to enable hibernate logging. If you want to enable jdbc logging you are then left wondering why there is no `dhis.conf` option for that and have to go to log4j2 anyway.

vietnguyen:
Close this for now. If there would be more requests from others then we could reopen.

PR #19880, opened 2025-02-07
DHIS2-18568: hard-code response timeout to 30 seconds

cjmamo:


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
```diff
@@ -223,4 +256,44 @@ void testGetRouteGivenApiQueryParamsAuthScheme() throws JsonProcessingException
...
+  @Test
+  void testRunRouteWhenResponseDurationExceedsRouteResponseTimeout()
+      throws JsonProcessingException {
+    mockServerClient
+        .when(request().withPath("/"))
+        .respond(org.mockserver.model.HttpResponse.response("{}").withDelay(TimeUnit.SECONDS, 31));
+
+    Map<String, Object> route = new HashMap<>();
+    route.put("name", "route-under-test");
+    route.put("url", "http://localhost:" + mockServerContainer.getFirstMappedPort());
```
jbee, line 269: You can do this with `json-tree`, have a look at `Json.object`

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
```diff
@@ -59,13 +64,41 @@
...
 
 @Transactional
 class RouteControllerTest extends DhisControllerIntegrationTest {
 
+  private static GenericContainer<?> mockServerContainer;
+
   @Autowired private RouteService service;
 
   @Autowired private ObjectMapper jsonMapper;
+  private MockServerClient mockServerClient;
```
jbee, line 78: I assume you are mocking a service that the routing somehow wants to talk with but that doesn't get clear in any why from the code. Maybe find some names that don't tell me what it is on a technical level (I can see that in the class name) but what the instance is doing, what role it has.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
jbee: It saves you from having to use jackson.

PR #19893, opened 2025-02-10
fix: Return index.html when object storage used with an empty page name

david-mackessy:
# Issue
[Reports on Slack](https://dhis2.slack.com/archives/C045593A2CE/p1738321271354609) advised that when clicking the `edit profile` link, an empty page would load with a 500 response
That link triggers a call to `GET` `/dhis-web-user-profile/#/profile`
No bug raised as yet.

# Cause
When DHIS2 is setup using object storage (MinIO in Play), a call for an App resource directory (e.g. `/dhis-web-user-profile`) would fail as this results in an invalid call for an underlying file (key) resource in the object storage.

When using local file storage, the code path does resolve a directory call with an appended `index.html` suffix. When using object storage, the path to resolve a directory with `index.html` is never reached (uri is never null)

# Fix
After the requirements for how/what the server should be able to resolve were made clear, more local testing was carried out.
The testing and fix uses the following expectations of the server:
| Resource | Should resolve to |
| ------------- | ------------- |
| '' (empty string)  | `index.html`  |
| `index.html`  | `index.html`  |
| `/index.html`  | `/index.html`  |
| `subDir`  | `subDir/index.html`  |
| `subDir/`  | `subDir/index.html`  |
| `subDir/index.html`  | `subDir/index.html`  |

This should work for resource types:
1. file names
2. directories (with or without trailing `/`)
3. other sub-directories (with or without trailing `/`)

As well as the bug reported for requests to directory with no trailing `/`, another bug was found and fixed, when using object storage. The server was returning empty 200 responses for requests to directories with a trailing `/` (never resolving to a file resource).

# Testing
## Automated
Unit tests added for StringUtils
Integration tests added when using MinIO (testing multiple resource variations)

## Manual
MinIO: Test the reported issue and the issue seems to be fixed. Also tested navigating around different apps, installing & opening apps.
File storage: Check that the reported issue is not present when using file storage, also navigating around different apps, installing & opening

# Info
Although the fix does look straightforward, test coverage doesn't seem great in this area, so manual testing looks like the best way to catch errors with this fix. We also use MinIO as the default in Play, which is the opposite of what the DHIS2 default is (file storage). This will be raised with Devops & QA.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -351,11 +359,7 @@ public Resource getAppResource(App app, String pageName) throws IOException {
         return null;
       }
 
-      if (res.isDirectory()) {
```
david-mackessy, line 354: This logic seems like it can be removed now, because a blank page name is not possible anymore at this point.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -328,14 +329,21 @@ public void deleteApp(App app) {
...
     if (app == null || !app.getAppStorageSource().equals(AppStorageSource.JCLOUDS)) {
       log.warn(
           "Can't look up resource {}. The specified app was not found in JClouds storage.",
           pageName);
       return null;
     }
 
+    // assigning index.html if blank as it looks like we always try to return the index.html
+    // for all types of storage (file, object)
+    if (pageName.isBlank()) {
```
amcgee, line 342: @david-mackessy I'm not sure this works for requests to app sub-directories?  Any bare directory request should respond with index.html if found within that dir

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -365,6 +373,45 @@ public Resource getAppResource(App app, String pageName) throws IOException {
...
+    }
+
+    // this condition is treated as a directory (any level) resource, ending in '/'
+    if (resource.endsWith("/")) {
+      log.debug("Resource ends with '/', appending 'index.html' to {}", resource);
+      return resource + "index.html";
+    }
+
+    // this condition is treated as a directory resource not ending in '/'
+    if (!resource.contains(".")) { // not a file extension
```
amcgee, line 405: @david-mackessy I'm not sure this is correct... `.` is a valid character in a directory name, so this might incorrectly fail to append `index.html`

Technically the best way to handle this is to check the underlying object store with the following sequence:

1. [x] If the requested path ends with `/`, respond with contents of `./index.html` (if it exists, otherwise 404)
2. [x] Try to fetch the underlying resource for the requested path (no trailing slash) - if it exists as a file, return it.
3. [ ] If it exists as a directory (local storage only), *return a 302 redirect to the request path with a `/` appended*
4. [ ] If the requested path doesn't exist, try appending a `/` and see if it exists as a directory (directory object type in s3) then *return a 302 redirect to the request path with a `/` appended*

Sorry if this is a little complicated / unclear - feel free to let me know if we should pair on it in the next day or two.

See some references to trailing slash redirect pattern:
https://httpd.apache.org/docs/2.4/en/mod/mod_dir.html#directoryslash
https://rankmath.com/kb/how-to-redirect-to-urls-with-trailing-slash/
https://docs.netlify.com/routing/redirects/redirect-options/#trailing-slash

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
amcgee: Thanks @david-mackessy !  Looking generally quite nice.  I think we should clean up a bit how raw directory requests (without a trailing slash) are detected and handled, see my other comment.  I realize this was a bit ambiguous / inconsistent before 42 as well.

david-mackessy: Thanks @amcgee , it's good that there does seem to be some expected behaviour from the server regarding how app resources should be handled (this is not well known or documented AFAIK). If these (pretty important) details could be centralized for easy access by all, and then added to a Jira issue, it will make things so much easier. I'll close this PR and re-open a new one once there's a Jira issue to track the work.

PR #19906, opened 2025-02-11
feat: Support for Options in /dataItems API [DHIS2-18956]

maikelarabori:
The endpoint needs to support new objects as part of the support for Option in analytics.

New objects introduced:
PROGRAM_DATA_ELEMENT_OPTION (PDEO)
PROGRAM_ATTRIBUTE_OPTION (PAO)

Those objects will only be returned on /dataItems when "dimensionItemType" and "programDataElementId"/"programAttributeId" filters are specified:

`dimensionItemType:in:[PROGRAM_DATA_ELEMENT_OPTION,PROGRAM_ATTRIBUTE_OPTION]`
`programDataElementId:eq:eBAyeGv0exc.fWIAEtYVEGk`

When these filters are not specified, we don't want PDEO and PAO to be returned in a global search, as there could be many of them.

PDEO and PAO objects will have a property that holds the `id` of the associated program data element or program attribute. Example:
```
{
	"valueType": "TEXT",
	"simplifiedValueType": "TEXT",
	"displayName": "Discharged (Mode of Discharde, Inpatient morbidity and mortality)",
	"name": "Discharged (Mode of Discharde, Inpatient morbidity and mortality)",
	"programDataElementId": "eBAyeGv0exc.fWIAEtYVEGk",
	"id": "eBAyeGv0exc.fWIAEtYVEGk.yeod5tOXpkP",
	"shortName": "Discharged (Mode of Discharde, Inpatient morbidity and mortality)",
	"displayShortName": "Discharged (Mode of Discharde, Inpatient morbidity and mortality)",
	"dimensionItemType": "PROGRAM_DATA_ELEMENT_OPTION"
}
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataitem/query/ProgramDataElementOptionQuery.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataitem/query/ProgramDataElementOptionQuery.java
@@ -97,11 +97,13 @@ public class ProgramDataElementOptionQuery implements DataItemQuery {
       "program.uid, dataelement.uid, optionset.uid, optionvalue.uid, optionvalue.code, optionvalue.name";
 
   private static final String JOINS =
-      "   join programstagedataelement on programstagedataelement.dataelementid = dataelement.dataelementid"
-          + " join programstage on programstage.programstageid = programstagedataelement.programstageid"
-          + " join program on programstage.programid = program.programid"
-          + " join optionset on dataelement.optionsetid = optionset.optionsetid"
-          + " join optionvalue on optionvalue.optionsetid = optionset.optionsetid";
+      """
+        join programstagedataelement on programstagedataelement.dataelementid = dataelement.dataelementid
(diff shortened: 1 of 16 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataitem/query/ProgramDataElementOptionQuery.java
```diff
@@ -0,0 +1,225 @@
...
+              Pair.of("optionvalue_name", "optionvalue.name"),
+              Pair.of("optionvalue_code", "optionvalue.code"))
+          .stream()
+          .map(pair -> pair.getRight() + " as " + pair.getLeft())
+          .collect(joining(", "));
+
+  private static final String COMMON_UIDS =
+      "program.uid, dataelement.uid, optionset.uid, optionvalue.uid, optionvalue.code, optionvalue.name";
+
+  private static final String JOINS =
```
larshelge, line 99: Could we use text blocks here with """ and \ at the end of lines?

PR #19911, opened 2025-02-12
fix: Prevent duplicate descriptions in data integrity checks

jason-p-pickering:
It was possible to have two separate integrity checks with the same description. We already checked for duplicate codes and names. This PR fixes a defect and adds a unit test to ensure we do not duplicate the description.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReaderTest.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReaderTest.java
@@ -76,6 +76,11 @@ class DataIntegrityYamlReaderTest {
     List<String> codeList = checks.stream().map(DataIntegrityCheck::getCode).sorted().toList();
     assertEquals(codeList.size(), Set.copyOf(codeList).size());
 
+    // Assert that all the descriptions are unique.
+    List<String> nameList =
+        checks.stream().map(DataIntegrityCheck::getDescription).sorted().toList();
+    assertEquals(nameList.size(), Set.copyOf(nameList).size());
+
     // Assert that codes consist of upper case letter and numbers only
     String regEx = "^[A-Z0-9]+$";
     Predicate<String> IS_NOT_CAPS = Pattern.compile(regEx).asPredicate().negate();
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-administration/src/test/java/org/hisp/dhis/dataintegrity/DataIntegrityYamlReaderTest.java
```diff
@@ -76,6 +76,11 @@ void testReadDataIntegrityYaml() {
     List<String> codeList = checks.stream().map(DataIntegrityCheck::getCode).sorted().toList();
     assertEquals(codeList.size(), Set.copyOf(codeList).size());
 
+    // Assert that all the descriptions are unique.
+    List<String> nameList =
+        checks.stream().map(DataIntegrityCheck::getDescription).sorted().toList();
+    assertEquals(nameList.size(), Set.copyOf(nameList).size());
```
david-mackessy, line 82: This will ensure any bundled descriptions are unique. What about custom checks? Is the intention to keep them unique for all? (bundled + custom). If not, why not?

jason-p-pickering: Good point. Maybe for custom checks, we should just not load them if the codes, names or descriptions are duplicated? Maybe around here someplace?

```
try {
        String dhis2Home = locationManager.getExternalDirectoryPath();
        loadChecks(
            FILE_SYSTEM,
            dhis2Home + "/custom-data-integrity-checks.yaml",
            dhis2Home + "/custom-data-integrity-checks");
```

david-mackessy: We already do something similar for `name` & `code`, [code here](https://github.com/dhis2/dhis2-core/blob/0629c05edc20648a7f23c0366fa6e8b891e3c08c/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java#L854).

See the [docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-241/data-validation.html?h=custom+data+integrity+2.41#name-and-code-constraints).

We could just extend it for description as well.
We do log a warning if a custom check is not loaded, so that could be reused for description too.
If we do add this check, it could impact users who have custom checks which violate this new condition. At least it would be logged, but it could be easily missed sneaky.
Docs should be updated if we add this.

david-mackessy: yep, that's fine. As discussed, checking for duplicate descriptions may be overkill (we check code and name).

david-mackessy (approved):

jbee (approved):

PR #19912, opened 2025-02-12
fix: settings - empty value is always valid (reset to default) [DHIS2-18962]

jbee:
Settings validation was missing to short-circuit when the new value was empty, which means it is unset and falls back to the default.


david-mackessy (commented):
Can a test be added?
Methods with many different conditional paths should really have tests.

larshelge (approved):

david-mackessy (approved):

PR #19919, opened 2025-02-13
chore: Propagate includeDeleted through relationships [DHIS2-18883]

enricocolasante:
Propagate `includeDeleted` params from any entity service to `RelationshipService` to include soft-deleted `relationshipItems` when `includeDeleted` is `true`.

***Next steps***
- Fix `isDeleted` check in `OperationsParamsValidator`

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.webapi.controller.tracker.export.enrollment;
 
+import static org.hisp.dhis.test.utils.Assertions.assertIsEmpty;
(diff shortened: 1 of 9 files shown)


ameenhere (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
```diff
@@ -203,6 +204,94 @@ void shouldGetEnrollmentsWithAttributes(
...
+    manager.delete(relationship);
+
+    assertNotNull(
+        relationship.getTo().getEnrollment(),
+        "test expects relationship to have a 'to' enrollment");
+    Enrollment enrollment = relationship.getTo().getEnrollment();
+
+    JsonList<JsonRelationship> jsonRelationships =
+        GET(
+                "/tracker/enrollments?enrollments={id}&fields=*&includeDeleted=false",
```
teleivo, line 219: Our tests should probably also only ask for what they need in the `fields` to make them as slow as necessary 😅

teleivo (approved):

PR #19921, opened 2025-02-13
chore: Remove preprocess that filters out deleted entitites in tracker stores [DHIS2-18883]

enricocolasante:
Remove `preProcessPredicates` and `postProcessObject` methods from entity stores. Now, when getting an entity using any method from the store, we need to check if the entity is deleted or not.

We still need to have the `isDeleted` check in a few places in our services but we want to remove them, we should check for it only in the stores:
- `TrackedEntityService`: it will be removed when completing https://dhis2.atlassian.net/browse/DHIS2-18541. (One of next PRs)
- `RelationshipService`: relationships do not work as the other entities so there is some thinking to do here (we don't have a clear single vs multiple methods as the other entities have). It may be removed in one of next PRs
- `OperationsParamsValidator`: we need to retrieve the parent entity (trackedEntity for enrollment and enrollment for event) to check for existence. We need to fix the logic to consider parameter `includeDeleted` when checking for existence. This is needed when a parent service is calling a child service (`TrackedEntityService` calling `EnrollmentService`) and the parent entity is soft-deleted. More details will come in the PR working on this.

***Next steps***
- Fix `isDeleted` check in `OperationsParamsValidator`
- Check how we should deal with `deleted` entities in `RelationshipService`.


teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/PageParams.java
```diff
@@ -43,6 +43,10 @@ public class PageParams {
   private static final int DEFAULT_PAGE = 1;
   private static final int DEFAULT_PAGE_SIZE = 50;
 
+  public static PageParams single() {
```
ameenhere, line 46: Just throwing this out there. I have see  "first()" used in libs, that could be a potential name here as well. No strong opinions

ameenhere (approved):

PR #19930, opened 2025-02-14
fix: cannot get TE if there are no tracker programs DHIS2-18991

teleivo:
cherry-pick of https://github.com/dhis2/dhis2-core/pull/19918

https://dhis2.atlassian.net/browse/DHIS2-18991 is the same bug as https://dhis2.atlassian.net/browse/DHIS2-17959 except it manifests only in tracker exporter as tracker is not part of metadata sync in 42.

So a user requesting TEs

* with no access to any tracker program
* or on a system without tracker programs

would get a 500 due to invalid SQL.


enricocolasante (approved):

muilpp (approved):

radnov (approved):
Docker Compose changes look good to me. 👍

PR #19937, opened 2025-02-16
fix: make sortOrder 1-indexed

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-18473

### Issue
- User reported that the integrity check failed because OptionSet.options sortOrder start from 0.

### Fix
- Update code to make sortOrder to start from 1

### Test
- Adjust existing tests to accommodate new behavior.


jason-p-pickering:
@vietnguyen please hold on merging this. I spoke with Calle at length about this, and at the time, he felt that 1-indexed made more sense (which I agree with). Having said that, there are many systems out there which are 0-indexed (as this was the previously default). Nowadays, I do not think it matters in principle if option sets start with 0 or 1, both work. Calle felt strongly they should start with 1, as this made more sense intuitively.  If the integrity checks are flagging incorrectly option sets which start with 0, then we need to fix the integrity check. Its not clear from the PR how or if we intend to "upgrade" old option sets which do start with zero. Thoughts?

vietnguyen:
> @vietnguyen please hold on merging this. I spoke with Calle at length about this, and at the time, he felt that 1-indexed made more sense (which I agree with). Having said that, there are many systems out there which are 0-indexed (as this was the previously default). Nowadays, I do not think it matters in principle if option sets start with 0 or 1, both work. Calle felt strongly they should start with 1, as this made more sense intuitively.  If the integrity checks are flagging incorrectly option sets which start with 0, then we need to fix the integrity check. Its not clear from the PR how or if we intend to "upgrade" old option sets which do start with zero. Thoughts? 

@jason-p-pickering 
Would be great if you could change the integrity check to accept 0-indexed.

I think there will be no issue for the api consumer. Sorting function won't throw error if index start from 1.

This fix only affects the update action. So if no update request to OptionSet then the sort order remain the same.

Let me know if you decide to update the integrity check so I can close the PR and jira ticket.

jason-p-pickering:
@vietnguyen I have adjusted the data integrity check now so that option sets with sort orders of either 0 or 1 will be accepted. Given that, I am not sure we really need this PR, but I am not totally opposed to it, as I think there is something to be said to keep things "tidy" by having the option set sort orders start at 1. @jbee might have thoughts here as well.

jason-p-pickering (approved):

jbee (approved):
To me this is a technical solution and even if the value is available in the API we never made any commitment to either starting value (I hope). I think this should be one of those things we can change whenever we want to whatever we want but product might have a different opinion on this so I suggest to consult with David K.

PR #19949, opened 2025-02-17
fix: use error only notifier logging for frequent jobs [DHIS2-17998]

jbee:
### Summary
I noticed that with the changes made to the notifier where all jobs do forward at least the first and the last message (procress start/end) to the notifier jobs that do run frequently (like the housekeeping job) basically spam the log with 1 line per run through the notifier logging.  

To prevent it the notifier log level is set to ERROR if a job should use non-verbose logging. This means for the first and last message where the notification is still always send the message is blanked which also prevents logging.

### Automatic Testing
There is no good way of testing it and I don't think it is worth making a complicated setup to verify.

### Manual Testing
* start the server
* check the logs for not showing a line from the `DefaultNotifier` once every 30 sec

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobSchedulerLoopService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobSchedulerLoopService.java
@@ -260,19 +260,20 @@ public class DefaultJobSchedulerLoopService implements JobSchedulerLoopService {
 
   private JobProgress startRecording(@Nonnull JobConfiguration job, @Nonnull Runnable observer) {
     SystemSettings settings = settingsProvider.getCurrentSettings();
-    boolean logInfoOnDebug = isLogInfoOnDebug(job, settings);
+    boolean nonVerboseLogging = isNonVerboseLogging(job, settings);
     NotificationLevel level =
-        !logInfoOnDebug && job.getJobType().isUsingNotifications()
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobSchedulerLoopService.java
```diff
@@ -260,22 +260,26 @@ private static void logError(String message, Exception ex) {
 
   private JobProgress startRecording(@Nonnull JobConfiguration job, @Nonnull Runnable observer) {
     SystemSettings settings = settingsProvider.getCurrentSettings();
+    boolean nonVerboseLogging = isNonVerboseLogging(job, settings);
     NotificationLevel level =
-        job.getJobType().isUsingNotifications()
+        job.getJobType().isUsingNotifications() && !nonVerboseLogging
```
david-mackessy, line 265: Double negative here. Could `!nonVerboseLogging` just be `verboseLogging`? (If I'm understanding this correctly)

netroms (approved):

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/scheduling/DefaultJobSchedulerLoopService.java
jbee: I thought about that too but I didn't want to inverse the logic in the function and I did not find another why to phrase it so in the end I thought the double negation was the lesser problem.

PR #19951, opened 2025-02-17
fix: Propagate includeDeleted from TEs to Enrollments services [DHIS2-18883]

enricocolasante:
Propagate `includeDeleted` from `TrackedEntityService` to `EnrollmentService` and fix `validateTrackedEntity` method: this was a bug only and it was only present in master. Getting `/tracker/trackedEntities?trackedEntity={te_uid}&includeDeleted=true&fields=* when TE is deleted  return a 500 because it was trying to get enrollments filtering by a soft-deleted `trackedEntity` that was throwing the error [here](https://github.com/dhis2/dhis2-core/blob/68ffdfbcb8bc9ac8013c620be76c045c54ae2b64/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/aggregates/EnrollmentAggregate.java#L77).

Now the behavior of `validateTrackedEntity` changed and a deleted `trackedEntity` is a valid parameter if the `includeDeleted` parameter is `true`.

Additionally, partial ACL check in `validateTrackedEntity` method was removed as any child entity (eg. event or enrollment) will check the ACL of the parent when checking its own ACL.

Enrollment [`canRead`](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java#L252-L256)
Event [`canRead`](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java#L406-L410) 

The same propagation for `includeDeleted` was added in `RelationshipService`, so a deleted entity is a valid parameter if the `includeDeleted` parameter is `true`.

***Next steps***

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
@@ -58,9 +58,10 @@ class RelationshipOperationParamsMapper {
 
     IdentifiableObject entity =
         switch (params.getType()) {
-          case TRACKED_ENTITY -> getTrackedEntity(params.getIdentifier());
-          case ENROLLMENT -> getEnrollment(params.getIdentifier());
-          case EVENT -> getEvent(params.getIdentifier());
+          case TRACKED_ENTITY ->
+              getTrackedEntity(params.getIdentifier(), params.isIncludeDeleted());
+          case ENROLLMENT -> getEnrollment(params.getIdentifier(), params.isIncludeDeleted());
+          case EVENT -> getEvent(params.getIdentifier(), params.isIncludeDeleted());
           case RELATIONSHIP -> throw new IllegalArgumentException("Unsupported type");
         };
 
@@ -71,11 +72,12 @@ class RelationshipOperationParamsMapper {
         .build();
   }
 
-  private TrackedEntity getTrackedEntity(UID trackedEntityUid)
+  private TrackedEntity getTrackedEntity(UID trackedEntityUid, boolean includeDeleted)
       throws NotFoundException, ForbiddenException {
     TrackedEntity trackedEntity =
         relationshipStore
             .findTrackedEntity(trackedEntityUid)
+            .filter(te -> (includeDeleted || !te.isDeleted()))
             .orElseThrow(() -> new NotFoundException(TrackedEntity.class, trackedEntityUid));
     if (!trackerAccessManager
         .canRead(CurrentUserUtil.getCurrentUserDetails(), trackedEntity)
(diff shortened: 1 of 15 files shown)


teleivo:
> Additionally, partial ACL check in `validateTrackedEntity` method was removed as any child entity (eg. event or enrollment) will check the ACL of the parent when checking its own ACL.

Where do they validate the ACL of the parent if you remove that check from the shared `OperationsParamsValidator`?

This

https://github.com/dhis2/dhis2-core/blob/5d640681a1205de84f06bfe72eb6d730cf5c63a8/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentOperationParamsMapper.java#L69-L70

now does not check data read on the TET, unless the TET parameter is set 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -71,11 +72,12 @@ public RelationshipQueryParams map(@Nonnull RelationshipOperationParams params)
         .build();
   }
 
-  private TrackedEntity getTrackedEntity(UID trackedEntityUid)
+  private TrackedEntity getTrackedEntity(UID trackedEntityUid, boolean includeDeleted)
       throws NotFoundException, ForbiddenException {
     TrackedEntity trackedEntity =
         relationshipStore
             .findTrackedEntity(trackedEntityUid)
+            .filter(te -> (includeDeleted || !te.isDeleted()))
```
teleivo, line 80: Should we move this into the rel store?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerTest.java
```diff
@@ -469,8 +470,9 @@ void getTrackedEntityByIdyWithFieldsRelationshipsNoAccessToTrackedEntityType() {
     relationship(from, to);
     this.switchContextToUser(user);
 
-    GET("/tracker/trackedEntities/{id}?fields=relationships", from.getUid())
```
teleivo, line 472: Does this style not work? or does the `assertEquals()` version give better error messages

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerTest.java
```diff
@@ -613,6 +613,26 @@ void getTrackedEntityReturnsCsvGZipFormat() {
...
+        GET(
+                "/tracker/trackedEntities?trackedEntities={id}&fields=enrollments&includeDeleted=true",
+                "woitxQbWYNq")
+            .content(HttpStatus.OK)
+            .getList("trackedEntities", JsonTrackedEntity.class);
+
+    JsonList<JsonEnrollment> enrollments = json.get(0).getEnrollments();
+    enrollments.forEach(
+        en -> {
+          assertEquals(EnrollmentStatus.CANCELLED.name(), en.getStatus());
```
teleivo, line 628: do we set the status to cancelled on delete?

teleivo (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerTest.java
enricocolasante: Yes :man_shrugging:

muilpp (approved):

PR #19953, opened 2025-02-17
feat(DHIS2-18994): support OAuth2 client credentials flow for Route API

cjmamo:
* Support OAuth2 client credentials flow for Route API
* Solve nasty security regression which dirties the route entity during decryption causing the clear text secret to be inadvertently persisted. Will attempt to implement a safeguard in a separate PR so that this doesn't happen again.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/OAuth2ClientCredentialsAuthScheme.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/auth/OAuth2ClientCredentialsAuthScheme.java
@@ -63,10 +63,12 @@ public class OAuth2ClientCredentialsAuthScheme implements AuthScheme {
 
   public static final Authentication ANONYMOUS_AUTHENTICATION =
       new AbstractAuthenticationToken(null) {
+        @Override
         public Object getCredentials() {
           return "";
         }
 
+        @Override
         public Object getPrincipal() {
           return "anonymous";
         }


dhis-2/dhis-api/src/test/java/org/hisp/dhis/eventhook/targets/auth/ApiTokenAuthSchemeTest.java
```diff
@@ -47,7 +47,7 @@ void testAuthorizationHeaderSet() {
         new ApiTokenAuthScheme().setToken("90619873-3287-4296-8C22-9E1D49C0201F");
 
     MultiValueMap<String, String> headers = new LinkedMultiValueMap<>();
-    auth.apply(headers, null);
+    auth.apply(null, headers, null);
```
jbee, line 50: IDK what is going on here but a generic function being called with 2 `null` values deserves a comment to explain why that is intentional.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -171,7 +170,7 @@ public Route getRouteWithDecryptedAuth(@Nonnull String id) {
    */
   public ResponseEntity<StreamingResponseBody> execute(
       Route route, UserDetails userDetails, Optional<String> subPath, HttpServletRequest request)
-      throws IOException, BadRequestException {
+      throws Exception {
```
jbee, line 173: A service throwing a non-specific exception feels a bit off. Can't we narrow it down to a set of specific ones?
I agree that we don't want a service to throw an `IOException` either but something more semantic that translates to a HTTP response code. But I assume this might be a case where this would require creating more exceptions which I also don't think we should do. Just want to double-check to have you consider this once more.

dhis-2/dhis-services/dhis-service-event-hook/src/main/java/org/hisp/dhis/eventhook/handlers/WebhookHandler.java
```diff
@@ -74,7 +78,11 @@ public void run(EventHook eventHook, Event event, String payload) {
 
     MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
     if (webhookTarget.getAuth() != null) {
-      webhookTarget.getAuth().apply(httpHeaders, queryParams);
+      try {
+        webhookTarget.getAuth().apply(applicationContext, httpHeaders, queryParams);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
jbee, line 84: I think this is the async exception we talked about. Maybe we should leave a comment stating that this is async and therefore the type of exception this is re-thrown as does not matter much which is why it just converts it to a runtime exception?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/RouteController.java
```diff
@@ -80,7 +84,7 @@ public ResponseEntity<StreamingResponseBody> run(
       @PathVariable("id") String id,
       @CurrentUser UserDetails currentUser,
       HttpServletRequest request)
-      throws IOException, ForbiddenException, NotFoundException, BadRequestException {
+      throws Exception {
```
jbee, line 87: Please avoid this in controllers. If we can list specific exceptions we have to so OpenAPI docs "sees" them.

jbee (approved):
I can't really judge the details of the security related code but the rest looks fine. The one comment that is important to me is about the `throws Exception` in the controller. If possible we want to list concrete exceptions as that is how OpenAPI knows what error responses are possible.

dhis-2/dhis-api/src/test/java/org/hisp/dhis/eventhook/targets/auth/ApiTokenAuthSchemeTest.java
jbee: `null` is fine if it works, I just would leave a comment explaining why null is the value to use and what it means.

amcgee (approved):
LGTM - leaving the comments we discussed

PR #19960, opened 2025-02-18
task: Validate TE ACL on Deduplication controller [DHIS2-19030]

muilpp:
While addressing technical debt by moving all tracker-related logic into the tracker module, we temporarily used the manager to fetch tracked entities.
One of the places where we did this was in Deduplication, which can now use the `TrackedEntityService` in the tracker module and apply ACL like every other part of the system that fetches tracked entities. 

While at it, I'm removing the `DeduplicationMvcTest`, as it contains duplicated tests. The ones that are not duplicated I'm moving to the `DeduplicationControllerTest`.


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -66,38 +70,37 @@ class DeduplicationControllerTest extends H2ControllerIntegrationTestBase {
...
   private TrackedEntity origin;
-  private TrackedEntity duplicate1;
+  private TrackedEntity duplicate;
   private PotentialDuplicate potentialDuplicate1;
   private PotentialDuplicate potentialDuplicate2;
+  private User user;
 
   @BeforeEach
   void setUp() {
+    injectAdminIntoSecurityContext();
```
teleivo, line 80: Are you doing that to undo switching to another user? Asking because that should already be done for you in https://github.com/dhis2/dhis2-core/blob/5f32ce23c66a2b75018a2cba07529f58e998710e/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/test/webapi/ControllerIntegrationTestBase.java#L118-L119 

There are so many different `inject/switch/yyy/user` methods that we don't really know which one we should be using but it would be great to figure it out and settle on one 😅

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -66,38 +70,37 @@ class DeduplicationControllerTest extends H2ControllerIntegrationTestBase {
...
-    dbmsManager.save(duplicate1);
-    dbmsManager.save(duplicate2);
-
-    potentialDuplicate1 = new PotentialDuplicate(UID.of(origin), UID.of(duplicate1));
+    potentialDuplicate1 = new PotentialDuplicate(UID.of(origin), UID.of(duplicate));
     save(potentialDuplicate1);
     potentialDuplicate2 = new PotentialDuplicate(UID.of(origin), UID.of(duplicate2));
     save(potentialDuplicate2);
+
+    user = createAndAddUser("username");
```
teleivo, line 95:
```suggestion
    user = createAndAddUser("marceline");
```

Using the field name as the value is confusing IMHO

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -177,25 +180,69 @@ void shouldGetNonPaginatedItemsWithPagingSetToFalse() {
   }
 
   @Test
-  void shouldThrowPostPotentialDuplicateWhenMissingDuplicateTeiInPayload() throws Exception {
+  void shouldFailWhenSkipPagingAndPagingAreFalse() {
```
teleivo, line 183: shouldFailWhenSkipPagingAndPagingAreFalse and shouldFailWhenSkipPagingAndPagingAreTrue were just deleted. Maybe you need to rebase first

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -177,25 +180,69 @@ void shouldGetNonPaginatedItemsWithPagingSetToFalse() {
...
 
   @Test
-  void shouldThrowPostPotentialDuplicateWhenMissingOriginTeiInPayload() throws Exception {
-    PotentialDuplicate potentialDuplicate = new PotentialDuplicate(null, UID.of(duplicate1));
+  void shouldThrowNotFoundExceptionWhenTeInPayloadDoesNotExist() throws Exception {
+    String nonExistentUID = CodeGenerator.generateUid();
+    PotentialDuplicate potentialDuplicate =
+        new PotentialDuplicate(UID.of(origin), UID.of(nonExistentUID));
+    assertStatus(
+        HttpStatus.NOT_FOUND, POST(ENDPOINT, objectMapper.writeValueAsString(potentialDuplicate)));
```
teleivo, line 218: Is there a pro of using this style vs the `POST().content(HttpStatus.OK);` that is also in our code? I find the latter more readable but if the one you are using has an advantage I can be convinced 😁

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
```diff
@@ -229,9 +232,9 @@ private void validatePotentialDuplicate(PotentialDuplicate potentialDuplicate)
 
     checkValidTrackedEntity(potentialDuplicate.getDuplicate(), "duplicate");
 
-    canReadTrackedEntity(getTrackedEntity(potentialDuplicate.getOriginal()));
+    canReadTrackedEntity(trackedEntityService.getTrackedEntity(potentialDuplicate.getOriginal()));
```
teleivo, line 235: is the `canRead` check not already done by the service?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
muilpp: good to know thanks!

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
muilpp: ok, how about "regular-user"?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
muilpp: not that I know, just used it because every other tests uses that style too, but now that you bring it up, I prefer your proposal too

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
enricocolasante: This is true but the thrown exception will change once https://dhis2.atlassian.net/browse/DHIS2-18541 is completed.
When we are serving multiple and single endpoints with the same code we lose the ability to differentiate between 404 and 403.

ameenhere (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
muilpp: Is that just something to keep in mind, or are you saying we should continue returning 403 in this case?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/deduplication/DeduplicationControllerTest.java
```diff
@@ -177,73 +176,137 @@ void shouldGetNonPaginatedItemsWithPagingSetToFalse() {
...
-        HttpStatus.BAD_REQUEST,
-        POST(ENDPOINT, objectMapper.writeValueAsString(potentialDuplicate)));
+
+    POST(ENDPOINT, objectMapper.writeValueAsString(potentialDuplicate))
+        .content(HttpStatus.BAD_REQUEST);
+  }
+
+  @Test
+  void shouldThrowNotFoundExceptionWhenTeInPayloadDoesNotExist() throws Exception {
+    String nonExistentUID = CodeGenerator.generateUid();
```
teleivo, line 188:
```suggestion
    String nonExistentUID = UID.generate();
```

makes the same call under the hood but is nicer to read 😁

teleivo (commented):
I'll defer to @enricocolasante for the ✅ due to the unresolved ACL comment

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
enricocolasante: Just to keep in mind.
And maybe worth a discussion on the backend because we are declaring `ForbiddenException` in methods that cannot really throw such exception, and it is confusing.
(For example [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java#L90-L107))

enricocolasante (approved):

PR #19973, opened 2025-02-18
fix: Remove handle deleted in RelationshipService [DHIS2-18883]

enricocolasante:
Move `isDeleted` check from `RelationshipService` to `HibernateRelationshiStore`.
In the process, I refactored the way we build `RelationshipOperationParams` in order to make them safe and not possible to create an invalid object. I delomboked the builder for `RelationshipOperationParams` to do exactly what I need it to do but I am open to any improvement here.
Now all the paths to get relationships are using `RelationshipOperationParams`, all the different paths by entity are removed and the different branches of the query are handled in the store.
`RelationshipQueryParams` is not safe but it is internal and it should only be created by `RelationshipOperationParamsMapper`.

***Next steps***


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -91,86 +91,67 @@ public HibernateRelationshipStore(
...
-      final RelationshipQueryParams queryParams,
-      @Nonnull PageParams pageParams) {
-
-    return getPage(
-        pageParams,
-        relationshipsList(trackedEntity, queryParams, pageParams),
-        () -> countRelationships(trackedEntity, queryParams));
-  }
+  public Page<Relationship> getRelationships(
+      final RelationshipQueryParams queryParams, @Nonnull PageParams pageParams) {
```
teleivo, line 149:
```suggestion
    @Nonnull final RelationshipQueryParams queryParams, @Nonnull PageParams pageParams) {
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParams.java
```diff
@@ -42,51 +43,97 @@
 import org.hisp.dhis.tracker.export.Order;
 
 @Getter
-@Builder(toBuilder = true)
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class RelationshipOperationParams {
   private TrackerType type;
```
teleivo, line 48: Can these fields now all be final 🤔 as the builder sets its own fields and on build passes them on to this classes constructor

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -54,9 +56,21 @@ class RelationshipOperationParamsMapper {
...
   public RelationshipQueryParams map(@Nonnull RelationshipOperationParams params)
-      throws NotFoundException, ForbiddenException {
+      throws NotFoundException, ForbiddenException, BadRequestException {
+
+    if (params.getType() == null) {
+      if (CollectionUtils.isEmpty(params.getRelationships())) {
+        throw new BadRequestException(
+            "Either one between Relationships or TrackerType must be provided");
+      }
+      return RelationshipQueryParams.builder()
```
teleivo, line 66: if `RelationshipQueryParams` is internal and all we do is a call like this we don't need a builder but a plain old constructor

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -54,9 +56,21 @@ class RelationshipOperationParamsMapper {
 
   @Transactional(readOnly = true)
   public RelationshipQueryParams map(@Nonnull RelationshipOperationParams params)
-      throws NotFoundException, ForbiddenException {
+      throws NotFoundException, ForbiddenException, BadRequestException {
+
+    if (params.getType() == null) {
+      if (CollectionUtils.isEmpty(params.getRelationships())) {
+        throw new BadRequestException(
+            "Either one between Relationships or TrackerType must be provided");
```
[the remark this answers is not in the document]
enricocolasante: That is actually an impossible state because of the way we build `RelationshipOperationParams`.
I changed it to an `IllegalArgumentException`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -76,8 +90,7 @@ private TrackedEntity getTrackedEntity(UID trackedEntityUid, boolean includeDele
       throws NotFoundException, ForbiddenException {
     TrackedEntity trackedEntity =
         relationshipStore
-            .findTrackedEntity(trackedEntityUid)
```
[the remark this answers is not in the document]
enricocolasante: I agree the structure is weird.
We can discuss it and it could be part of another PR.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -116,22 +116,12 @@ public List<Enrollment> getEnrollments(@Nonnull Set<UID> uids) throws ForbiddenE
...
     } catch (BadRequestException e) {
       throw new IllegalArgumentException(
           "this must be a bug in how the EnrollmentOperationParams are built");
     }
-
-    return getEnrollments(
-        new ArrayList<>(enrollmentStore.getEnrollments(queryParams)),
-        EnrollmentParams.FALSE,
-        false,
-        queryParams.getOrganisationUnitMode());
```
muilpp, line 134: why do we not need the org unit mode anymore?

enricocolasante: We still need it.
This change do not remove it, it just creates the `EnrollmentOperationParams` and call the `getEnrollments` method instead of the manually creating `EnrollmentQueryParams` and calling the store.
If you go inside `getEnrollments` method you will see that this code is still there.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
muilpp: I just meant it in this particular case, but I guess it's why there's no org unit param from the user so we end up using the default one, right?

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
enricocolasante: The org unit mode is needed to build the `EnrollmentQueryParams` and if it wasn't previously set, then the default one is used.
I just wanted to be clear because of your `anymore` on the first question, nothing changed with this refactor.

PR #19982, opened 2025-02-19
fix: TransientObjectException on LegendSet during import EventVisualization [DHIS2-17445-40]

vietnguyen:
backport https://github.com/dhis2/dhis2-core/pull/19938


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AnalyticalObject.java
```diff
@@ -83,4 +85,9 @@ public interface AnalyticalObject
   boolean hasUserOrgUnit();
 
   void clearTransientState();
+
+  @Nullable
+  default LegendDefinitions getLegendDefinitions() {
+    return null;
+  }
```
vietnguyen, line 92: This default method is added in order to fix the issue but not breaking legacy object model

netroms (approved):

jbee (approved):

PR #19988, opened 2025-02-19
ci: upload IM db from stable pipeline if one is missing [skip ci]

radnov:
With this PR the stable pipeline will upload a DB to the IM for the new released version, if one is missing.
The `stable.json` update PR will also be merged directly from the pipeline.

Related to https://github.com/dhis2/jenkins-pipeline-library/pull/24

Tested with the following build - https://ci.dhis2.org/view/dhis2-core/job/dhis2-core-stable/view/tags/job/2.55.5/7/

### DIFF
--- a/jenkinsfiles/stable
+++ b/jenkinsfiles/stable
@@ -204,6 +204,7 @@ pipeline {
                             sh 'git push https://$GITHUB_TOKEN@github.com/$DHIS2_RELEASES_REPO'
 
                             sh "gh pr create --head $releasesBranch --fill-first --reviewer Philip-Larsen-Donnelly,dhis2/devops"
+                            sh "gh pr merge $releasesBranch"
                         }
                     }
                 }
--- a/jenkinsfiles/stable
+++ b/jenkinsfiles/stable
@@ -252,14 +252,31 @@ pipeline {
                             }
 
                             dir('scripts/databases') {
-                                env.DATABASE_ID = sh(
-                                        returnStdout: true,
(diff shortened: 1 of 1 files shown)


jenkinsfiles/stable
```diff
@@ -252,14 +253,31 @@ pipeline {
...
+
+                                        shortVersion = version.split('\\.').take(2).join('.')
+                                        env.DATABASE_ID = im.findDatabaseId(DATABASE_GROUP_NAME, shortVersion)
 
-                            sh '[ -n "$DATABASE_ID" ]'
-                            echo "Database: ${env.DATABASE_ID}"
+                                        if (!env.DATABASE_ID) {
+                                            echo "Couldn't find database for $shortVersion"
+
+                                            env.DATABASE_ID = im.uploadNewDatabase(DATABASE_GROUP_NAME, shortVersion)
```
Philip-Larsen-Donnelly, line 273: should we catch any errors here? - (just to handle them gracefully)

_not sure it's necessary, as it is quite unlikely to reach here_

radnov: If we can't find a database even by the "short version", like 2.41, then the pipeline should fail I think.

Philip-Larsen-Donnelly: Yes, agreed. 

Will that just "happen" with the current implementation, or should we throw an error here and exit?

radnov: Yeah, the uploadNewDatabase function will fail if there's no database to download for the given version.

PR #20020, opened 2025-02-21
feat: PI Disaggregation objects [DHIS2-18745]

jimgrace:
See [DHIS2-18745](https://dhis2.atlassian.net/issues/DHIS2-18745). Adds the backend objects needed for Program Indicator disaggregtion

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/InconsistentMetadataException.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/InconsistentMetadataException.java
@@ -27,61 +27,21 @@
  */
 package org.hisp.dhis.common;
 
-import java.util.Arrays;
-import java.util.List;
-import lombok.Getter;
 import org.hisp.dhis.feedback.ErrorCode;
-import org.hisp.dhis.feedback.ErrorMessage;
 
 /**
- * Runtime exception which references an {@link ErrorCode}.
+ * Exception for inconsistent metadata, including metadata relationships.
  *
- * @author Lars Helge Overland
+ * @author Jim Grace
  */
-@Getter
-public class ErrorCodeException extends RuntimeException {
-  private final ErrorCode errorCode;
-
-  private final List<Object> args;
-
-  /** Deprecated, use {@link ErrorCode} or {@ErrorMessage} constructors instead. */
-  public ErrorCodeException(String message) {
-    super(message);
-    this.errorCode = null;
-    this.args = null;
-  }
-
+public class InconsistentMetadataException extends ErrorCodeException {
   /**
-   * Constructor.
-   *
-   * @param errorCode the {@link ErrorCode}.
-   */
-  public ErrorCodeException(ErrorCode errorCode) {
-    super(errorCode.getMessage());
-    this.errorCode = errorCode;
-    this.args = null;
-  }
-
-  /**
-   * Constructor.
-   *
-   * @param errorMessage the {@link ErrorMessage}.
-   */
-  public ErrorCodeException(ErrorMessage errorMessage) {
-    super(errorMessage.getMessage());
-    this.errorCode = errorMessage.getErrorCode();
-    this.args = null;
-  }
-
-  /**
-   * Constructor.
+   * Constructor. Sets the message based on the error code and arguments.
    *
    * @param errorCode the {@link ErrorCode}.
    * @param args the message format arguments.
    */
-  public ErrorCodeException(ErrorCode errorCode, Object... args) {
-    super(new ErrorMessage(errorCode, args).getMessage());
-    this.errorCode = errorCode;
-    this.args = Arrays.asList(args);
+  public InconsistentMetadataException(ErrorCode errorCode, Object... args) {
+    super(errorCode, args);
   }
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
@@ -28,6 +28,7 @@
 package org.hisp.dhis.program;
 
 import static java.util.stream.Collectors.toSet;
+import static org.hisp.dhis.common.DxfNamespaces.DXF_2_0;
 import static org.hisp.dhis.util.ObjectUtils.copyOf;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -188,6 +189,9 @@ public class Program extends BaseNameableObject implements VersionedObject, Meta
 
   /** Property indicating level of access */
   private AccessLevel accessLevel = AccessLevel.OPEN;
+
+  /** Library of Category Mappings available to this program's program indicators */
+  private Set<ProgramCategoryMapping> categoryMappings = new HashSet<>();
 
   // -------------------------------------------------------------------------
   // Constructors
@@ -935,6 +939,21 @@ public class Program extends BaseNameableObject implements VersionedObject, Meta
     this.accessLevel = accessLevel;
   }
 
+  @JsonProperty("categoryMappings")
+  @JacksonXmlElementWrapper(localName = "categoryMappings", namespace = DXF_2_0)
+  @JacksonXmlProperty(localName = "categoryMappings", namespace = DXF_2_0)
+  public Set<ProgramCategoryMapping> getCategoryMappings() {
+    return categoryMappings;
+  }
+
+  public void setCategoryMappings(Set<ProgramCategoryMapping> categoryMappings) {
+    this.categoryMappings = categoryMappings;
+  }
+
+  // -------------------------------------------------------------------------
+  // Copy logic
+  // -------------------------------------------------------------------------
+
   public static Program shallowCopy(Program original, Map<String, String> options) {
     Program copy = new Program();
     copy.setAutoFields();
@@ -948,6 +967,7 @@ public class Program extends BaseNameableObject implements VersionedObject, Meta
     copy.setAccessLevel(original.getAccessLevel());
     copy.setProgramAttributes(new ArrayList<>());
     copy.setCategoryCombo(original.getCategoryCombo());
+    copy.setCategoryMappings(copyOf(original.getCategoryMappings()));
     copy.setCompleteEventsExpiryDays(original.getCompleteEventsExpiryDays());
     copy.setDataEntryForm(original.getDataEntryForm());
     copy.setDescription(original.getDescription());
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2004-2022, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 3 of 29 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
```diff
@@ -935,6 +939,21 @@ public void setAccessLevel(AccessLevel accessLevel) {
     this.accessLevel = accessLevel;
   }
 
+  @JsonProperty("categoryMappings")
+  @JacksonXmlElementWrapper(localName = "categoryMappings", namespace = DXF_2_0)
+  @JacksonXmlProperty(localName = "categoryMappings", namespace = DXF_2_0)
```
david-mackessy, line 944: Are all of the xml annotations needed anymore? We dropped xml support many versions ago AFAIK.

david-mackessy (approved):
This feature is above my head but otherwise everything looks ok to me.
Left a comment about xml annotations as I'm pretty sure we don't support that anymore, unless there's some other reason for using them?

From a testing perspective, the ObjectBundleHook looks like it could use some tests if possible. Is the intention that this feature is to be used in analytics? If so, is there any way to have some tests to confirm that the feature is working as expected? It looks like there are tests for most of the parts, but not the actual end goal of the feature (if I'm understanding correctly).

jimgrace:
@david-mackessy Thanks!
- I've answered you comment about the XML annotations.
- I've now added tests to `ProgramObjectBundleHookTest` for all the functionality I added to `ProgramObjectBundleHook`. I had been meaning to do this but then forgot. Good catch, thank you!!!
- You're right, the actual use of this metadata in analytics is not yet written. The idea was to first implement the new metadata for metadata import/export so frontend code could be written at the same time as I'm writing the backend analytics code.
- `ProgramCategoryMappingResolver` is intended to do double duty, both at metadata import time and at analytics run time.  During metadata import it is called by `ProgramObjectBundleHook`, which also does some additional validation not needed at runtime. At runtime it will be called by the analytics code I'm writing to resolve the UID references for `Category` and `CategoryOption` objects, and to make sure those objects still exist, before proceeding.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/InconsistentMetadataException.java
```diff
@@ -0,0 +1,47 @@
...
+package org.hisp.dhis.common;
+
+import org.hisp.dhis.feedback.ErrorCode;
+
+/**
+ * Exception for inconsistent metadata, including metadata relationships.
+ *
+ * @author Jim Grace
+ */
+public class InconsistentMetadataException extends ErrorCodeException {
```
jbee, line 37: Generally we don't want to create more exception types. Would it work to use `ErrorCodeException` directly? Or even better one of the exceptions from the `feedback` package, which translate to the HTTP error code you want to send back, in case the problem occurs that you created this exception for. They all can be used with the new error codes you created. Using new error codes with existing exceptions is the preferred way.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
jbee: You are right that we do read them for schema but if there are JSON annotations they should not be needed. We just never removed most of them but we should. Since that is the case you can leave them and we get them when we do a big cleanup of XML annotations.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /** The category for this category mapping (from the category option UID) */
+  @Setter @Getter @EqualsAndHashCode.Include private transient CategoryOption option;
+
+  /** The filter for this category option */
+  private String filter;
+
+  // -------------------------------------------------------------------------
+  // Getters and setters
+  // -------------------------------------------------------------------------
```
jbee, line 69: I am not sure if you copied this or if you are the origin of this type of divider comments but I try to delete them when I find them so please don't create new work for me :)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
```diff
@@ -0,0 +1,90 @@
...
+@NoArgsConstructor
+@EqualsAndHashCode
+@JacksonXmlRootElement(localName = "optionMapping", namespace = DxfNamespaces.DXF_2_0)
+public class ProgramCategoryOptionMapping implements EmbeddedObject, Serializable {
+
+  /** The category option UID to be mapped */
+  private String optionId;
+
+  /** The category for this category mapping (from the category option UID) */
+  @Setter @Getter @EqualsAndHashCode.Include private transient CategoryOption option;
```
jbee, line 62: The `@EqualsAndHashCode.Include` should be on the `optionId` - way less risk of odd behaviour IMHO

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramIndicatorService.java
```diff
@@ -127,6 +127,15 @@ public interface ProgramIndicatorService {
...
   String getFilterDescription(String expression);
 
+  /**
+   * Gets a program indicator expression filter (must evaluate to a Boolean) without declaring a
+   * transaction.
+   *
+   * @param expression A program indicator expression or filter string
+   * @return The description
+   */
+  String getFilterDescriptionWithoutTransaction(String expression);
```
jbee, line 137: That name is clear but I'd rather know why that matters or how it can do this if otherwise a TX would be needed. What is this skipping to make that happen? What is the difference to `getFilterDescription` in the result? That should be in the name and javadoc. The fact this does not need a TX can be communicated in the implementation by annotating it `@NonTransactional` - If you want you can even add that annotation in the interface if you think this is important for the caller to know.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
```diff
@@ -202,6 +202,11 @@ public String getExpressionDescription(String expression) {
   @Override
   @Transactional(readOnly = true)
   public String getFilterDescription(String expression) {
+    return getFilterDescriptionWithoutTransaction(expression);
+  }
+
+  @Override
+  public String getFilterDescriptionWithoutTransaction(String expression) {
```
jbee, line 209: The implementation makes me think that this does require a TX but what you want is to not open a new one in an already opened one? Or why would `getFilterDescription` open one if it is not needed?

jbee (commented):
I am a bit confused about mapping being tables since we said to use JSON but I put that down to you trying to finish this given the circumstances at PEPFAR which is fine by me. 

A few comments, nothing major, but little things to take care of.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/InconsistentMetadataException.java
jimgrace: Aha, thanks! `...feedback.ConflictException` does exactly what I need; I've switched to that.

I did look for exceptions to use before defining my own, but I didn't notice `ConflictException`. I needed an exception that stored the error code and I saw that `ErrorCodeException` did this. It was used only through subclasses and not directly so I thought it best to follow that pattern. All the subclasses were used for particular types of errors that were unlike what I had, so I created my own subclass. As always, I tried to be consistent with existing coding patterns. I even enhanced `ErrorCodeException` to return the error message args which I needed for the import report.

I didn't realize that there were other exceptions that also stored the error code -- and even the args. I've reverted my additions to `ErrorCodeException` and switched to using `ConflictException`!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
jimgrace: Ha! These divider comments already permeated the codebase in 2013 when I joined. :) At first I thought they were a bit clumsy but I used them because that was the DHIS2 coding style. I grew to appreciate them because it made it easy to skim the file and find the logic amidst the boilerplate constructors, setters & getters, equals and hashCode, etc. I had not noticed that you are intentionally deleting them.

I suppose that the more we use lombok for our boilerplate code the less important these dividers become. I still think they could be useful in separating the public methods from the "Supportive" methods.

But that's your call; I've removed all the new ones I had added.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramCategoryOptionMapping.java
jimgrace: OK, I've now removed those annotations on the `transient` objects in `ProgramCategoryOptionMapping` and also in `ProgramCategoryMapping`. You were right to question them and, had I left them in, you would not have been the last to wonder about them.

I had put them in for some of tests in `ProgramCategoryMappingResolverTest`. To test that objects got resolved from the UIDs, I created "resolved" mappings with the objects included. Then I could simply compare the result `Set<ProgramCategoryMapping>` with an expected `Set<ProgramCategoryMapping>` containing the resolved objects. (Each `ProgramCategoryMapping`, in turn, contains a resolved `Set<ProgramCategoryOptionMapping>`.) So it made the test a one-liner.

Since the fields of these objects are not persisted as columns in a database table -- and the objects are only resolved in copies of the objects that are used temporarily in memory -- I wouldn't expect the usual odd behaviour of including these objects in equals and hashCode. However, I realize that it doesn't look good and it goes against the usual way things are done. Worse yet, if someone removed these annotations the tests would still succeed but they wouldn't be testing that the objects are resolved from the UIDs.

So I've now removed these annotations. I've refactored the resolver tests to (A) check that all the non-transient fields are still present, and (B) explicitly test for the presence of teach transient object (by comparing the UID field with the UID found inside the the resolved test object).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ProgramIndicatorService.java
jimgrace: I've removed this method because it turns out I didn't need it after all. I had it there because I was originally testing each mapping filter expression on metadata import. `getFilterDescription` was annotated as a transaction with `(readOnly = true)`, and it was being called from `ProgramObjectBundleHook` which eventually writes to the database, causing a conflict. Having a non-transaction way of validating the expression resolved this error.

I debated back-and-forth whether I should try to validate the mapping expressions while importing the metadata. The reason against this is that we don't do it with any other expressions and it adds time to the import. The reason in favor is that we have a lot of expressions in a single metadata object and it could be inconvenient if any of them are invalid. (some installations have thousands of attribute option combos; each would need its own filter expression, presumably generated by a script that generates the metadata.)

At one point I had coded it to do the expression checking from the bundle hook. Eventually I decided we shouldn't validate the expressions as we were importing them, so I removed this code. I just forgot to remove `getFilterDescriptionWithoutTransaction`.

The deciding factor was that the expression may depend on other metadata in the same import that is not yet persisted. For example, the filter expression (inside the `Program` object) may contain `#{programStageUid.dataElementUid}` and it may be part of a metadata package that contains both `Program` and `ProgramStage` objects, including the `ProgramStage` object referenced in the expression.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/DefaultProgramIndicatorService.java
jimgrace: Removed. (See previous.)

jimgrace:
@jbee 
> I am a bit confused about mapping being tables...

Thanks. `ProgramCategoryMapping` and `ProgramCategoryOptionMapping` are WebAPI obejcts but they are stored in jsonb fields as we agreed. There are no new tables. See: [V2_42_37__Program_indicator_disaggregation.sql](dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_37__Program_indicator_disaggregation.sql)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Program.java
jimgrace: I looked further into this and I see that @JsonProperty can apply to fields that have lombok setters and getters. Wonderful!

So the entire ProgramCategoryMapping class should reduce to:

```
@Data
@Builder(toBuilder = true)
@JsonRootName(value = "categoryMapping", namespace = DxfNamespaces.DXF_2_0)
public class ProgramCategoryMapping implements EmbeddedObject, Serializable {

  /** The UID for this category mapping */
  @JsonProperty private String id = "";

  /** The category UID for this category mapping */
  @JsonProperty private String categoryId = "";

  /** The category for this category mapping (from the category UID) */
  @Setter @Getter private transient Category category;

  /** The category mapping name, must be unique within a program and category */
  @JsonProperty private String mappingName = "";

  /** The category option mappings for this program, category, and name */
  @JsonProperty private Set<ProgramCategoryOptionMapping> optionMappings = new HashSet<>();
}
```
However when I tried this on my system, `ProgramStoreTest.testGetAndDeleteProgramWithCategoryMappings` failed with: `Hibernate Could not deep copy JSONB object`. So I've left it as it is for now.

I look forward to when we can get this all simplified and working!

jbee (approved):

PR #20030, opened 2025-02-21
feat: Add unary operators to tracked entity filters [DHIS2-15954]

muilpp:
This PR introduces unary operators for tracked entity filters.

To maintain consistency with data element filtering, the filter parameter supports any combination of operators as well, meaning expressions like `filter=null:eq:2` are allowed.

All entries in `trackedentityattributevalue` have a defined value, as validation prevents `NULL` values before insertion.

To reflect this, a `LEFT JOIN` is now used for `trackedentityattributevalue`, along with a new condition in the `WHERE` clause that filters based on the provided operator and its value, if applicable.

Additionally, the methods `parseAttributeFilters` and `parseDataElementFilters` have been merged into `parseFilters`. This was originally a temporary measure when implementing data element filtering, but it makes sense to consolidate them for consistency.

**Performance Analysis**
I evaluated two scenarios to assess performance impact:

- Filtering by first name, a TEA present for almost every TE, meaning many entries exist in `trackedentityattributevalue`.
- Filtering by a newly introduced TEA, used by only one TE, resulting in a single entry in `trackedentityattributevalue`.

In both cases, LEFT JOIN and INNER JOIN perform similarly. By the time the LEFT JOIN is executed, previous JOIN conditions (on program, tracked entity type, program owner, and organization unit) have already eliminated most irrelevant data. At this stage, only 15 results remain. Filtering by first name returns all 15, while filtering by the new TEA returns just one.

**Performance comparisons**
- LEFT JOIN on first name: [Plan](https://explain.dalibo.com/plan/b163bc1hg7a557d0)
- INNER JOIN on first name: [Plan](https://explain.dalibo.com/plan/48eb6gcb9cbd8a06)
- LEFT JOIN on new attribute: [Plan](https://explain.dalibo.com/plan/g38b1fgdbeggg7ca)
- INNER JOIN on new attribute: [Plan](https://explain.dalibo.com/plan/b3146fb4dcg76cfa)

**Next**
Apply the same logic when filtering by attributes on events


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -556,8 +557,12 @@ private String joinAttributeValue(TrackedEntityQueryParams params) {
       String teav = "lower(" + col + ".value)";
       String ted = col + ".trackedentityid";
 
+      if (filters.getValue().stream().anyMatch(qf -> qf.getOperator().isNull())) {
```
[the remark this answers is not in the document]
muilpp: That's something I considered, but the `inner join` filters out unwanted rows directly, whereas the `left join` does not, leaving them to be filtered later in the `where` clause. That's why I prefer not to create a `left join` unless it's necessary.

Anyway, I removed the `left join` when the operator is `null`. In this case, no join is performed, leaving it to the `where` clause to filter entries for attributes that don't exist. Do you also find the flow easier to follow this way?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -556,6 +557,10 @@ private String joinAttributeValue(TrackedEntityQueryParams params) {
       String teav = "lower(" + col + ".value)";
       String ted = col + ".trackedentityid";
 
+      if (filters.getValue().stream().allMatch(qf -> qf.getOperator().isNull())) {
```
enricocolasante, line 560: I am confused by that `if`
You are in a loop, what if there are filters on 2 attributes and the first one has only `null` filter?
Would we skip the required `inner join` for the second attribute?

muilpp: Fixed, the result is the same, that's why no test failed.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1701,6 +1638,71 @@ void shouldReturnTrackedEntityRelationshipsWithTe2Event()
         () -> assertEquals(eventA.getUid(), actual.getTo().getEvent().getUid()));
   }
 
+  @Test
+  void shouldReturnEmptyCollectionGivenSingleQuoteInAttributeSearchInput()
+      throws ForbiddenException, NotFoundException, BadRequestException {
+    TrackedEntityOperationParams operationParams =
+        TrackedEntityOperationParams.builder()
```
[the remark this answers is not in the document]
enricocolasante: That's the point, I can see the differences in the filters but then I cannot understand the differences in the orgUnitModes and makes me wonder if they have something to do with the test itself.
Having the same params, except the filters, in all the tests, would make it easier to spot the relevant differences.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -386,17 +386,18 @@ private String getFromSubQuery(
...
 
             // LEFT JOIN attributes we need to sort on.
             .append(getFromSubQueryJoinOrderByAttributes(params))
+            .append(getLeftJoinFromFilterConditions(params))
 
             // WHERE
             .append(getFromSubQueryTrackedEntityConditions(whereAnd, params))
-            .append(getFromSubQueryEnrollmentConditions(whereAnd, params));
+            .append(getFromSubQueryEnrollmentConditions(whereAnd, params))
+            .append(getWhereClauseFromFilterConditions(whereAnd, params));
```
enricocolasante, line 400: If it is possible, it would make sense `getWhereClauseFromFilterConditions` to be first in the `where` section to be as close as possible to the related left join

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityQueryParams.java
```diff
@@ -520,7 +521,7 @@ public TrackedEntityQueryParams filterBy(TrackedEntityAttribute tea, QueryFilter
    * tea}.
    */
   public TrackedEntityQueryParams filterBy(TrackedEntityAttribute tea) {
-    this.filters.putIfAbsent(tea, new ArrayList<>());
+    this.filters.putIfAbsent(tea, List.of(new QueryFilter(QueryOperator.NNULL)));
```
ameenhere, line 524: Just to understand, what does this mean?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -1701,16 +1639,82 @@ void shouldReturnTrackedEntityRelationshipsWithTe2Event()
...
+
+  @Test
+  void shouldReturnTrackedEntitiesIfTheyHaveGivenAttributeFilteredUsingOnlyUID()
+      throws ForbiddenException, NotFoundException, BadRequestException {
+    TrackedEntityOperationParams operationParams =
+        TrackedEntityOperationParams.builder()
+            .organisationUnits(orgUnitA, orgUnitB, orgUnitChildA)
+            .orgUnitMode(DESCENDANTS)
+            .trackedEntityType(trackedEntityTypeA)
+            .filters(Map.of(UID.of(teaA), List.of()))
```
teleivo, line 1698: I thought you were going to make this API safe 🤔 like `filterBy(uid)` which then does the correct thing under the hood. and `filterBy(uid, filter1, filter2, ...)`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -544,61 +543,34 @@ private String getFromSubQueryTrackedEntityConditions(
...
           .append(" LEFT JOIN trackedentityattributevalue AS ")
-          .append(quote(orderAttribute.getUid()))
+          .append(col)
           .append(" ON ")
-          .append(quote(orderAttribute.getUid()))
+          .append(col)
           .append(".trackedentityid = TE.trackedentityid ")
-          .append("AND ")
-          .append(quote(orderAttribute.getUid()))
+          .append(" AND ")
```
teleivo, line 566: merge the `" AND "` with the above append

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParams.java
```diff
@@ -171,10 +168,14 @@ public class TrackedEntityOperationParams {
    */
   private List<Order> order;
 
+  private final Map<UID, List<QueryFilter>> filters;
```
teleivo, line 171: Get the javadoc back

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParams.java
```diff
@@ -249,10 +250,24 @@ public TrackedEntityOperationParamsBuilder organisationUnits(
...
+    // does not support. The repeated filters field and private filters method prevent access to
+    // order
+    // via the builder.
+    // Order should be added via the filterBy builder methods.
+    private TrackedEntityOperationParamsBuilder filters(Map<UID, List<QueryFilter>> filters) {
+      return this;
+    }
+
+    public TrackedEntityOperationParamsBuilder filterBy(
+        UID attribute, List<QueryFilter> queryFilters) {
```
teleivo, line 263: Add `@Nonnull` where applicable

teleivo (approved):

PR #20031, opened 2025-02-21
feat: add prev/nextPage to /tracker/trackedEntities DHIS2-19021

teleivo:
following


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -288,9 +288,9 @@ private void validateGlobalSearchParameters(TrackedEntityQueryParams params, Use
         validateSearchableAttributes(params, searchableAttributeIds);
       }
 
-      int maxTeiLimit = getMaxTeiLimit(params);
-      checkIfMaxTeiLimitIsReached(params, maxTeiLimit);
-      params.setMaxTeLimit(maxTeiLimit);
+      int maxTeLimit = getMaxTeLimit(params);
+      checkIfMaxTeLimitIsReached(params, maxTeLimit);
+      params.setMaxTeLimit(maxTeLimit);
```
ameenhere, line 293: If you move this like one up, then you don't need to send the `maxTeLimit` as a parameter to the `checkIfMaxTeLimitIsReached` function anymore, as it is already contained in the `params` parameter.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -413,11 +413,11 @@ private boolean isProgramMinAttributesViolated(TrackedEntityQueryParams params)
                 < params.getEnrolledInTrackerProgram().getMinAttributesRequiredToSearch());
   }
 
-  private void checkIfMaxTeiLimitIsReached(TrackedEntityQueryParams params, int maxTeiLimit) {
-    if (maxTeiLimit > 0) {
+  private void checkIfMaxTeLimitIsReached(TrackedEntityQueryParams params, int maxTeLimit) {
+    if (maxTeLimit > 0) {
```
ameenhere, line 417: NitPicky, but if the `maxTeLimit` is not sent as an additional parameter, and if reusing the one inside `params` , a method like `hasMaxTeLimit` inside `TrackedEntityQueryParams` could be used here instead, that does the same check.

teleivo: I should not have renamed the existing code 😂

enricocolasante (approved):

enricocolasante (approved):

ameenhere (approved):

PR #20034, opened 2025-02-21
DHIS-19010 Append geo suffix to OU Data Element fields

luciano-fiandesio:
## Summary

Fix a [bug](https://dhis2.atlassian.net/browse/DHIS2-19010) that caused an Analytics Event API request to fail if the selected Data Element for the Program is type `OU`.

## Issue

When composing the Select statement for the query against `analytics_event_*` table, the wrong column is selected if the Data Element is of type OU.

```
-- Assuming n1rtSHYf6O6 is a OU DE

select .. from ..

ST_AsGeoJSON(coalesce(ax."n1rtSHYf6O6"),6) as geometry,
```

The column `n1rtSHYf6O6` on the `analytics_event_*` table **is not** a geometry column, therefore the SQL statement fails.

## Resolution

When collecting the coordinate fields (`DefaultEventDataQueryService.getCoordinateFields`), append the `_geom` suffix to the column name, so that it can be correctly passed to the `ST_AsGeoJSON` function.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventCoordinateService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventCoordinateService.java
@@ -133,7 +133,7 @@ public class DefaultEventCoordinateService implements EventCoordinateService {
     if (ValueType.COORDINATE != valueType && ValueType.ORGANISATION_UNIT != valueType) {
       throwIllegalQueryEx(errorCode, field);
     }
-    // append the "_geom" suffix to the field
+    // Append the "_geom" suffix to the field
     // so that the correct geometry column
     // is selected
     return field + OU_GEOMETRY_COL_SUFFIX;
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventCoordinateService.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventCoordinateService.java
@@ -27,6 +27,7 @@
  */
 package org.hisp.dhis.analytics.event.data;
 
+import static org.hisp.dhis.analytics.table.AbstractEventJdbcTableManager.OU_GEOMETRY_COL_SUFFIX;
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/DefaultEventCoordinateService.java
```diff
@@ -132,8 +133,10 @@ public String validateCoordinateField(ValueType valueType, String field, ErrorCo
     if (ValueType.COORDINATE != valueType && ValueType.ORGANISATION_UNIT != valueType) {
       throwIllegalQueryEx(errorCode, field);
     }
-
-    return field;
+    // append the "_geom" suffix to the field
```
maikelarabori, line 136:
```suggestion
    // Append the "_geom" suffix to the field
```

maikelarabori (approved):

larshelge (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/analytics/event/data/EventDataQueryServiceTest.java
```diff
@@ -513,7 +513,7 @@ void testGetCoordinateField() {
         dataQueryService.getCoordinateFields(
             toRequest(prA.getUid(), null, "eventgeometry", false)));
     assertEquals(
-        List.of(deC.getUid()),
+        List.of(deC.getUid() + "_geom"),
```
larshelge, line 516: Could we use the `OU_GEOMETRY_COL_SUFFIX` constant to centralize?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/analytics/event/data/EventDataQueryServiceTest.java
```diff
@@ -548,7 +548,7 @@ void testGetBackwardCompatibleCoordinateField() {
         dataQueryService.getCoordinateFields(
             toRequest(prA.getUid(), null, OLD_COL_NAME_TRACKED_ENTITY_GEOMETRY, false)));
     assertEquals(
-        List.of(deC.getUid()),
+        List.of(deC.getUid() + "_geom"),
```
larshelge, line 551: Could we use the `OU_GEOMETRY_COL_SUFFIX` constant to centralize?

PR #20066, opened 2025-02-25
chore: feat: add prev/nextPage to /potentialDuplicates DHIS2-19021

teleivo:
* add prev/nextPage to /potentialDuplicates
* align service with our exporter services
  * remove `getById`
  * name `get` without `byUid`
  * throw `NotFoundException` in `get`
  * separate `List` and `Page` methods

## Tests

Extract a `TestSetup` class from [TrackerTest](https://github.com/dhis2/dhis2-core/blob/9c389f16bba4544916600cdd091c5e806edbc90e/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/TrackerTest.java#L57-L65). `TrackerTest` is not composable. The exporter can be tested using test lifecycle class as most tests can share the same setup and are read only. Code like the deduplication service does read and write. This means we need full control of the test annotations. This is inline [with the work](https://github.com/dhis2/dhis2-core/pull/18270) we've done on removing the number of test base classes so annotations can live in the test classes themselves.

Duplicate `TestSetup` and `Assertions` classes so we can use them in our `dhis-test-integration` and `dhis-test-web-api` tests. There is no better way right now to share such code. We could only put it into `main` but we do not want to include test code in production code. The other option of adding it into `dhis-support-test` introduces cycles. We should revisit our test module arrangement at some point.

We can share JSON though via `dhis-support-test` (see `tracker/simple_metadata.json` and `tracker/deduplication/potential_duplicates.json`).

* migrate deduplication controller test to postgres like we did for other tests
* move test into correct package
* use the tracker objects like we now do in exporter controller tests
  * add methods to `TrackerObjects` to find entities. We are already have them in [TrackerBundle](https://github.com/dhis2/dhis2-core/blob/13fd8bdc9f2f9b8a1a4672bc68af9561d40918f4/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java#L142). We might be able to remove them when we move to using the importer using the actual importer JSON payload.
* add potential duplicate specific TEs in `tracker/deduplication/potential_duplicates.json`. Note that the UIDs of the duplicate TEs match their `tracker/event_and_enrollment.json` counterpart expect for the `D` prefix.
* align service test with our other tests in terms of naming and grouping of arrange, act and assert

## Next

* simplify `view/Page` and integrate types well `PageRequestParams` -> `tracker/PageParams` -> `tracker/Page` -> `view/Page`
* handle requests to endpoints that do not support `totalPages`?
* reuse `TestSetup` to setup metadata and data in tracker integration tests
* reuse `TestSetup` to setup metadata and data in tracker web api tests
* maybe use TE service inside of the deduplication service. I'll do that if there is time. Right now it's odd that the service accepts TEs that might not even exist. That logic is in the controller. ACL is definitely also not done with the same level of detail as in the exporter service 😓


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DefaultDeduplicationService.java
```diff
@@ -127,17 +130,24 @@ public void manualMerge(DeduplicationMergeParams deduplicationMergeParams)
...
-  private String getAutoMergeConflictErrors(TrackedEntity original, TrackedEntity duplicate) {
+  private void validateCanBeAutoMerged(DeduplicationMergeParams params)
+      throws PotentialDuplicateConflictException {
+    TrackedEntity original = params.getOriginal();
+    TrackedEntity duplicate = params.getDuplicate();
+
     if (!original.getTrackedEntityType().equals(duplicate.getTrackedEntityType())) {
-      return "Entities have different Tracked Entity Types.";
+      throw new PotentialDuplicateConflictException(
+          "PotentialDuplicate cannot be merged automatically: Entities have different Tracked Entity Types.");
```
muilpp, line 140: can we have the first part of this sentence stored in a String so we don't repeat it 4 times?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
```diff
@@ -99,22 +99,19 @@ private String buildCountEnrollmentHql(EnrollmentQueryParams params) {
...
     Query<Enrollment> query = getQuery(hql);
-
     return query.list();
   }
 
   public Page<Enrollment> getEnrollments(EnrollmentQueryParams params, PageParams pageParams) {
     String hql = buildEnrollmentHql(params).getFullQuery();
-
     Query<Enrollment> query = getQuery(hql);
+
```
muilpp, line 109: why this line break here?

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/deduplication/potential_duplicates.json
```diff
@@ -0,0 +1,159 @@
+{
```
muilpp, line 1: it's definitely easier to understand the potential duplicate tests if we use this json, but didn't we say we want to have one json for metadata and one for data only?

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
teleivo: I compared the non-paginated with the paginated side by side and did not see the similarity at first. Thats why I grouped the statements.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/DefaultDeduplicationService.java
teleivo: we can

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/deduplication/potential_duplicates.json
teleivo: I am not sure that every tests requirements should affect all other tests 😅 All the exporter tests we have of which there are many do not need these TEs.

ameenhere (approved):

PR #20074, opened 2025-02-25
fix: use url-safe key when rendering apps as web modules, correct app override filter order [DHIS2-19118]

amcgee:
Fixes [DHIS2-19118](https://dhis2.atlassian.net/browse/DHIS2-19118)

This contains three changes which are required for the Global Shell:

1. Ensure that the Global Shell filter runs BEFORE the App Override filter, otherwise overridden apps will not correctly redirect to the global shell.
2. When rendering installed apps as WebModules (in the `/api/apps/menu` endpoint) use url-safe app key and correctly-formatted basepath (see below)
3. Don't exclude User Profile app from `/api/apps/menu` response

---

Prior to this change, custom apps were reandered as WebModules in the `/api/apps/menu` API (introduced with the removal or struts) with unsanitized name fields.  This causes issues when app names have special characters or spaces.  Additionally, the "namespace" property renders the relative base path for bundled apps but incorrectly was set to the name for installed apps.

I added also a `basePath` property to the `App` class which exposes the relative base path also under the `/api/apps` endpoint.

Before this change (including also a bundled app for reference):
```
{
  "name": "dhis-web-event-reports",
  "namespace": "/dhis-web-event-reports",
  "defaultAction": "../dhis-web-event-reports/index.html",
  "displayName": "Event Reports",
  "icon": "../icons/dhis-web-event-reports.png"
},
{
  "name": "android-settings-app",
  "namespace": "android-settings-app",
  "defaultAction": "http://localhost:8080/api/apps/android-settings-app/index.html",
  "displayName": "Android Settings",
  "icon": "http://localhost:8080/api/apps/android-settings-app/dhis2-app-icon.png",
  "description": "Configure synchronization parameters for the DHIS2 Android App, customize appear"
},
{
  "name": "Custom Js Css",
  "namespace": "Custom Js Css",
  "defaultAction": "http://localhost:8080/api/apps/Custom-Js-Css/index.html",
  "displayName": "Custom Js Css",
  "icon": "http://localhost:8080/api/apps/Custom-Js-Css/img/Logo_48.png",
  "description": "Custom Js Css"
}
```

After this change:
```
{
  "name": "dhis-web-event-reports",
  "namespace": "/dhis-web-event-reports",
  "defaultAction": "../dhis-web-event-reports/index.html",
  "displayName": "Event Reports",
  "icon": "../icons/dhis-web-event-reports.png"
},
{
  "name": "android-settings-app",
  "namespace": "/api/apps/android-settings-app",
  "defaultAction": "http://localhost:8080/api/apps/android-settings-app/index.html",
  "displayName": "Android Settings",
  "icon": "http://localhost:8080/api/apps/android-settings-app/dhis2-app-icon.png",
  "description": "Configure synchronization parameters for the DHIS2 Android App, customize appear"
},
{
  "name": "Custom-Js-Css",
  "namespace": "/api/apps/Custom-Js-Css",
  "defaultAction": "http://localhost:8080/api/apps/Custom-Js-Css/index.html",
  "displayName": "Custom Js Css",
  "icon": "http://localhost:8080/api/apps/Custom-Js-Css/img/Logo_48.png",
  "description": "Custom Js Css"
}
```

Note that the relative paths for bundled apps and the lack of case standardization are a bit odd and should probably be improved, but I left those as they were previously for now.

[DHIS2-19118]: https://dhis2.atlassian.net/browse/DHIS2-19118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-api/src/test/java/org/hisp/dhis/appmanager/AppTest.java
+++ b/dhis-2/dhis-api/src/test/java/org/hisp/dhis/appmanager/AppTest.java
@@ -87,7 +87,7 @@ class AppTest {
 
   @Test
   void testGetBaseUrl() {
-    assertEquals("http://localhost/api/apps/Test-App", app.getBaseUrl());
+    assertEquals("https://example.com/api/apps/Test-App", app.getBaseUrl());
   }
 
   // TODO: Complete test for skipped optional properties


jbee (approved):

david-mackessy (approved):

david-mackessy (approved):

amcgee:
> tested that it works on the global shell demo instance - The only (minor) thing I can notice that a shortName with non-latin characters would end up being empty with the current logic, I don't think we should support non-latin characters necessarily but just guard against the baseUrl being empty as it could mess up some of the other logic

Good catch @kabaros - we should probably block that at the app install stage..?  Not much we can do if an app has already been installed with a short name that normalizes to an empty string

PR #20075, opened 2025-02-25
chore: metadata query cleanup [DHIS2-18041]

jbee:
* replace `Schema` with `Class<T>` in `Query` (`Schema` is a service "internal" solution)
* add generic to `Query` and propagate it in all APIs and usages
* rename `Query.from` to `Query.of` (since it no longer is created from a `Schema` but for/of a object type `Class`)
* move the `Query.setDefaultOrders()` addition of actual `Order`s into the service (since they need `Schema` information)
* added `Query.shortNamePersisted` to remember if the schema has a persisted short name


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/query/InMemoryQueryEngine.java
```diff
@@ -117,62 +112,77 @@ private <T extends IdentifiableObject> List<T> runSorter(Query query, List<T> ob
...
 
-  private <T> boolean matches(Query query, T object, Filter filter) {
+  /**
+   * The main purpose of matchers is to "cache" the knowledge of how to extract a value for a filter
+   * test from an object of the listed object type.
+   *
+   * @param filter the filter that is applied
+   * @param value a function to extract the filter property value from an object
+   */
+  private record Matcher(Filter filter, Function<IdentifiableObject, Object> value) {}
```
jbee, line 134: FYI: ATM the `Matcher` indirection adds no benefit. I started this and then found that the refactoring the `getValue` needs will be quite extensive so I rather do it in another PR. But this is a preparation for it.

david-mackessy (approved):

PR #20084, opened 2025-02-26
chore: add OpenAPI spec and ignore totalPages DHIS2-19021

teleivo:
* implement common tracker paging interface `PageRequestParams` in changeLog parameters
  * this ensures they work like other tracker endpoints
  * lets us reuse the type
* ignore request parameter `totalPages` in endpoints that do not support it and use `PageRequestParams`. Also exclude it from the OpenAPI spec.
  * `/potentialDuplicates`
  * `/tracker/trackedEntities/{id}/changeLogs`
  * `/tracker/events/{id}/changeLogs`
* ignore request parameter `paging` in changeLog endpoints as they are always paginated
* `throw new UnsupportedOperationException("pageTotal is not supported")` in stores of endpoints that don't support it. This is for the service layer where `PageParams.isPageTotal()` could be true if a dev does not remember its not supported.
* `/potentialDuplicates?fields=` change code to match exporters and add test
* add OpenAPI spec directly in request parameters so we remember to add them/keep them up to date

## Next

* simplify `view/Page` and integrate types well `PageRequestParams` -> `tracker/PageParams` -> `tracker/Page` -> `view/Page`
* reuse `TestSetup` to setup metadata and data in tracker integration tests
* reuse `TestSetup` to setup metadata and data in tracker web api tests
* maybe use TE service inside of the deduplication service. I'll do that if there is time. Right now it's odd that the service accepts TEs that might not even exist. That logic is in the controller. ACL is definitely also not done with the same level of detail as in the exporter service 😓

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/HibernatePotentialDuplicateStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/deduplication/HibernatePotentialDuplicateStore.java
@@ -107,6 +107,10 @@ class HibernatePotentialDuplicateStore
 
   public Page<PotentialDuplicate> getPotentialDuplicates(
       PotentialDuplicateCriteria criteria, PageParams pageParams) {
+    if (pageParams.isPageTotal()) {
+      throw new UnsupportedOperationException("pageTotal is not supported");
+    }
+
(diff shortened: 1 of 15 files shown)


teleivo:
the code scan issue will be fixed in in https://dhis2.atlassian.net/browse/DHIS2-19039

ameenhere (approved):

enricocolasante (approved):

PR #20089, opened 2025-02-26
fix: Use correct get single TE for ownhership feature [DHIS2-18541]

enricocolasante:

In this PR we are starting to change tracker ownership service and controller to use the new and correct `getNewTrackedEntity` method.

***Changes in this PR***

- `getNewTrackedEntity` method is returning a detached entity (that has no `id` set). This is in order to avoid unwanted changes on this object and by consequence unwanted field changes in the DB. Services that expect a managed entity are retrieving one using the `IdentifiableObjectManager` class using `TrackedEntity` uid.

- `TrackedEntityOperationParamsMapper` was throwing a `BadRequestException` when the user had no access to any tracked entity type, it was change to a more suitable `ForbiddenException`

***Open question***
Can we transfer an ownership if there is no ownership present in the system?
When we transfer we are trying to get the tracked entity in the context of a program and if there is no ownership (in other words if the tracked entity is not enrolled or it wasn't previously enrolled) the tracked entity is not found.
Does it make sense or we should search the tracked entity outside the context of any program?

***Bug alert***
There are 2 integration tests failing because the user has no ownership access on the tracked entity, the program is protected and the user has the program organisation unit in its search scope but the user can retrieve the tracked entity.

***Challenges in tests***

***Next Steps***
- Move new methods for changelog feature

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
@@ -130,17 +130,20 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
       TrackedEntityProgramOwner teProgramOwner =
           trackedEntityProgramOwnerService.getTrackedEntityProgramOwner(trackedEntity, program);
 
+      // TODO(tracker) jdbc-hibernate: check the impact on performance
+      TrackedEntity hibernateTrackedEntity =
+          manager.get(TrackedEntity.class, trackedEntity.getUid());
       if (teProgramOwner != null && !teProgramOwner.getOrganisationUnit().equals(orgUnit)) {
         ProgramOwnershipHistory programOwnershipHistory =
             new ProgramOwnershipHistory(
                 program,
-                trackedEntity,
+                hibernateTrackedEntity,
                 teProgramOwner.getOrganisationUnit(),
                 teProgramOwner.getLastUpdated(),
                 teProgramOwner.getCreatedBy());
         programOwnershipHistoryService.addProgramOwnershipHistory(programOwnershipHistory);
         trackedEntityProgramOwnerService.updateTrackedEntityProgramOwner(
-            trackedEntity, program, orgUnit);
+            hibernateTrackedEntity, program, orgUnit);
       }
 
       ownerCache.invalidate(getOwnershipCacheKey(trackedEntity::getId, program));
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -227,6 +259,29 @@ public TrackedEntity getTrackedEntity(
     return getTrackedEntity(trackedEntityUid, program, params, getCurrentUserDetails());
   }
 
+  private TrackedEntity getNewTrackedEntity(UID uid, Program program, TrackedEntityParams params)
```
ameenhere, line 262: Is this really returning a detached TrackedEntity? If so, I feel we should rename the "getNewTrackedEntity" into something else, maybe even "getDetachedTrackedEntity"... or something more clearer.

teleivo: `getNewTrackedEntity` is a temporary name. All tracker exporters should return detached entities. `getNewTrackedEntity` is slowly replacing `getTrackedEntity` at which point it will itself be called `getTrackedEntity`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -154,15 +158,17 @@ public void grantTemporaryOwnership(
       throws ForbiddenException {
     validateGrantTemporaryOwnershipInputs(trackedEntity, program, user);
 
+    TrackedEntity hibernateTrackedEntity = manager.get(TrackedEntity.class, trackedEntity.getUid());
```
teleivo, line 161: Please add a `TODO(tracker) jdbc-hibernate: ....` so we mark all the places like this to revisit them. `jdbc-hibernate` is just an idea of a common prefix so we can search for `TODO(tracker) jdbc-hibernate`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityService.java
```diff
@@ -61,6 +70,17 @@ FileResourceStream getFileResourceImage(
   TrackedEntity getTrackedEntity(@Nonnull UID uid)
```
teleivo, line 70: add deprecated for removal so none of us uses it in new code

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapperTest.java
```diff
@@ -567,10 +567,10 @@ void shouldFailWhenUserHasNoAccessToAnyTrackedEntityType() {
     TrackedEntityOperationParams operationParams =
         TrackedEntityOperationParams.builder().orgUnitMode(ACCESSIBLE).build();
 
-    Exception badRequestException =
+    Exception forbiddenException =
```
teleivo, line 570:
```suggestion
    Exception exception =
```

for new tests at least. The type is already in the `assertThrows` so there is less to change in a case like this.

teleivo (approved):

ameenhere (approved):

PR #20090, opened 2025-02-26
fix: SQL condition for program indicator with program stage boundaries (master)

tokland:
[JIRA Issue: 18928](https://dhis2.atlassian.net/browse/DHIS2-18928)

Note: The primary goal of this PR was to fix the program stage condition. However, during testing, we also discovered that the generated SQL was invalid when boundaries contained multiple program stages, as the 'AND' operator was missing. The PR includes also the fix for this.


maikelarabori (approved):
It would be nice to have a unit test in `TimeFieldSqlRendererTest`, but I can add one once this is merged.

PR #20106, opened 2025-02-27
refactor: Update COC generation

david-mackessy:


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityCategoryOptionComboWrongCardinalityControllerTest.java
```diff
@@ -47,26 +47,6 @@ class DataIntegrityCategoryOptionComboWrongCardinalityControllerTest
 
   private String categoryColor;
 
-  private String testCatCombo;
-
-  @Test
```
david-mackessy, line 52: this test doesn't work anymore because the auto generation of COCs has the correct cardinality after removing the old COC (with 2 options) and adding the new COC (with 1 option)

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityCategoryOptionCombosNoNamesControllerTest.java
```diff
@@ -51,36 +47,6 @@ class DataIntegrityCategoryOptionCombosNoNames extends AbstractDataIntegrityInte
 
   private String categoryOptionRed;
 
-  @Test
```
david-mackessy, line 54: this test doesn't work anymore after the change. It may have been a test/code quirk that allowed updating a COC to have an empty name. I tried this through the API and wasn't able to do it (locally with new change and in IM dev).
It doesn't seem possible to update a COC to have an empty name.

jbee (approved):

larshelge (approved):

PR #20109, opened 2025-02-27
fix: Use correct get single TE for changelog feature [DHIS2-18541]

enricocolasante:

In this PR we are starting to change tracker ownership service and controller to use the new and correct `getNewTrackedEntity` method.

***Changes in this PR***

- Use `getNewTrackedEntity` method in `DefaultTrackedEntityChangeLogService`.

- Map `trackedEntityType` id in order to be used in other services that need managed metadata to work properly. Should we formalize that tracker data are always detached and metadata are always managed, even if they are coming from a tracker service?

***Challenges in tests***

***Next Steps***
- Move new methods for changelog feature

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
@@ -37,7 +37,6 @@ import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.tuple.Pair;
 import org.hisp.dhis.changelog.ChangeLogType;
 import org.hisp.dhis.common.UID;
-import org.hisp.dhis.feedback.BadRequestException;
 import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.feedback.NotFoundException;
 import org.hisp.dhis.program.Program;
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -103,19 +102,17 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
...
+            trackedEntityUid, programUid, TrackedEntityParams.FALSE.withIncludeAttributes(true));
 
     UserDetails currentUser = CurrentUserUtil.getCurrentUserDetails();
     Set<UID> trackedEntityAttributes = Collections.emptySet();
     if (programUid != null) {
       Program program = validateProgram(programUid.getValue());
       validateOwnership(currentUser, trackedEntity, program);
     } else {
-      validateTrackedEntity(currentUser, trackedEntity);
       trackedEntityAttributes = validateTrackedEntityAttributes(trackedEntity);
```
teleivo, line 116: is that still needed as ACL on TEA might not otherwise be done?

teleivo (approved):

teleivo:
> * Map `trackedEntityType` id in order to be used in other services that need managed metadata to work properly. Should we formalize that tracker data are always detached and metadata are always managed, even if they are coming from a tracker service?

maybe

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
enricocolasante: Maybe @muilpp can help us understand this better.
This validation is just checking if there are teas, and, if not, it is returning a tracked entity not found.

enricocolasante: @muilpp I was asking about `validateTrackedEntityAttributes`.
The other validation is not needed anymore as it is now present in `TrackerEntityService`

PR #20110, opened 2025-02-27
fix: do not pass null attributes to rule-engine[DHIS2-18621]

zubaira:
PR fix the handling of null attribute while building up rule-engine context.


enricocolasante (commented):
It would be good to have a test for this issue that is failing before your fix and then it will pass when you fix is applied.
TDD approach is very beneficial in cases like this, it allows us to spot the issue and to make sure that the fix is exactly what we need.

enricocolasante (approved):

ameenhere (approved):

PR #20111, opened 2025-02-27
chore: use JSON diff to assert pager

jbee:
An example of how `JsonValue.diff` can be used to write asserts that
* do structural matching
* are more readable
* fail with helpful details

Mainly this allows to ignore details like order of properties or presence of other properties the test does not care about while still allowing comparing the result to a JSON value.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/test/webapi/AssertionsTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/test/webapi/AssertionsTest.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2004-2025, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ * Neither the name of the HISP project nor the names of its contributors may
+ * be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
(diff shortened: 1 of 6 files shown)


jbee:
Needs to wait unti https://github.com/dhis2/json-tree/pull/74 is released

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/test/webapi/AssertionsTest.java
```diff
@@ -0,0 +1,60 @@
...
+            () ->
+                assertNoDiff(
+                    """
+                { "page": 1, "pageSize": 50 }""",
+                    """
+                { "paga": 1, "pageSize": 50 }"""));
+    assertEquals(
+        """
+        JSON has 2 structural differences:
+          ++ $.paga: ? <> 1
```
teleivo, line 54: What does `? <> 1` and its reverse below mean?

jbee: `?` is "not existing"/"nothing", `<>` is "differs from"/"not equal to" (`!=` was already taken in the context)

teleivo: and why is `nothing is different from 1` showing up on an unexpected key?

and the reverse is then `1 is different from nothing` 😅

david-mackessy (approved):
Looks nice for small payloads, not sure if I'd prefer it for larger payloads/checks.
I usually check for the presence/absence of things, is that possible?

For the example in this PR, I think the error message just saying:
`expected: <{ "page": 1, "pageSize": 50 }> but was: <{ "paga": 1, "pageSize": 50 }>` is good enough. It gets a little confusing having to think and learn about these new symbols :)

teleivo:
> I usually check for the presence/absence of things (as opposed to a straight diff), is that possible?

You can do that with other json-tree APIs 👌🏻 

I often find that tedious. The tracker controller test assertions are an example 😅 Being able to extract part of the JSON via json-tree and then asserting that using this diff makes them easier to read.

I do agree that I don't yet get the new symbols 😅

vietnguyen (approved):

PR #20121, opened 2025-02-28
fix: Use correct get single TE for sms feature [DHIS2-18541]

enricocolasante:

***Changes in this PR***

- Use `getNewTrackedEntity` method in `EnrollmentSMSListener`.
- Use `getNewTrackedEntity` method in `TrackerAccessManagerTest` and `MaintenanceServiceTest`.

***Challenges in tests***

***Next Steps***

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
@@ -33,7 +33,6 @@ import static org.hisp.dhis.user.CurrentUserUtil.getCurrentUsername;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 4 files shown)


muilpp (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/maintenance/MaintenanceServiceTest.java
```diff
@@ -451,10 +451,11 @@ void testDeleteSoftDeletedEnrollmentLinkedToARelationshipItem()
 
   @Test
   @Disabled("until we can inject dhis.conf property overrides")
-  void testAuditEntryForDeletionOfSoftDeletedTrackedEntity()
-      throws ForbiddenException, NotFoundException, BadRequestException {
+  void testAuditEntryForDeletionOfSoftDeletedTrackedEntity() {
     manager.delete(trackedEntityWithAssociations);
-    assertNull(trackedEntityService.getTrackedEntity(UID.of(trackedEntityWithAssociations)));
+    assertThrows(
+        NotFoundException.class,
```
teleivo, line 457: If this is a common case we should add the `Optional<T> find(UID)` method to our service

teleivo (approved):

PR #20124, opened 2025-02-28
fix: Use correct get single TE for deduplication feature [DHIS2-18541]

enricocolasante:

***Changes in this PR***

- Use `getNewTrackedEntity` method in `DeduplicationController`. We are using the `TrackedEntityService` only to validate the tracked entity, then we use the manager to get a managed object that is needed for the deduplication operations.
- Remove now unused `getTrackedEntity(UID)` method.

***Challenges in tests***

***Next Steps***


dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/deduplication/merge/PotentialDuplicatesMergeTests.java
```diff
@@ -108,7 +109,9 @@ public void shouldCheckReferences(String property) {
   @Test
   public void shouldNotMergeDifferentTypeTrackedEntities() {
     String trackedEntityType = new TrackedEntityTypeActions().create();
-
+    // A tracked entity is accessible only if it is enrollable somewhere
+    // so we need at least one program to make such teB accessible.
```
teleivo, line 113:
```suggestion
    // so we need at least one program to make such te accessible.
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicateRemoveTrackedEntityTest.java
```diff
@@ -199,28 +200,23 @@ void shouldDeleteEnrollments() throws ForbiddenException, NotFoundException, Bad
...
     manager.update(control1);
     manager.update(control2);
-    assertNotNull(trackedEntityService.getTrackedEntity(UID.of(original)));
-    assertNotNull(trackedEntityService.getTrackedEntity(UID.of(duplicate)));
-    assertNotNull(trackedEntityService.getTrackedEntity(UID.of(control1)));
-    assertNotNull(trackedEntityService.getTrackedEntity(UID.of(control2)));
+    assertNotNull(trackedEntityService.getNewTrackedEntity(UID.of(original)));
+    assertNotNull(trackedEntityService.getNewTrackedEntity(UID.of(duplicate)));
+    assertNotNull(trackedEntityService.getNewTrackedEntity(UID.of(control1)));
+    assertNotNull(trackedEntityService.getNewTrackedEntity(UID.of(control2)));
```
teleivo, line 206: getNewTrackedEntity does not return null right? it throws an exception

we should also annotate the method with `@Nonnull` to indicate that

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/deduplication/PotentialDuplicateRemoveTrackedEntityTest.java
enricocolasante: It is already annotated with @Nonnull.
We could change the assertion to `assertDoesNotThrow` but I will postpone this when I create the `findTrackedEntity` method, then the assertion will be cleaner

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/deduplication/DeduplicationController.java
```diff
@@ -149,30 +152,33 @@ public void mergePotentialDuplicate(
...
+    TrackedEntity hibernateDuplicate =
+        manager.get(TrackedEntity.class, potentialDuplicate.getDuplicate());
+
     DeduplicationMergeParams params =
         DeduplicationMergeParams.builder()
             .potentialDuplicate(potentialDuplicate)
             .mergeObject(mergeObject)
-            .original(original)
-            .duplicate(duplicate)
+            .original(hibernateOriginal)
```
muilpp, line 180: is this a temporary fix only?

enricocolasante: Yes, in this case we should completely refactor deduplication service and use the importer instead of custom stores. And then, we wouldn't need to get the tracked entities using the manager.

PR #20128, opened 2025-02-28
feat: Filter by unary operators on attributes in events [DHIS2-15954]

muilpp:
The last PR, that introduces unary operators for attribute filters in the event endpoint.
As in the tracked entities endpoint, I'm performing a left join to `trackedentityattributevalue`, which is used for both ordering and filtering. The `where` clause is then built using the user provided attribute filters.

Following the approach in `TrackedEntityQueryParams`, I refactored `EventOperationParamsMapper` to expose `filterByDataElement` and `filterByAttribute` in the service API, so users no longer have to deal with a Map structure.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -28,11 +28,9 @@
 package org.hisp.dhis.tracker.export.event;
 
 import static java.util.Map.entry;
-import static org.hisp.dhis.common.ValueType.NUMERIC_TYPES;
 import static org.hisp.dhis.system.util.SqlUtils.castToNumber;
 import static org.hisp.dhis.system.util.SqlUtils.lower;
 import static org.hisp.dhis.system.util.SqlUtils.quote;
-import static org.hisp.dhis.system.util.SqlUtils.singleQuote;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -82,7 +80,6 @@ import org.hisp.dhis.program.ProgramStage;
 import org.hisp.dhis.program.ProgramType;
 import org.hisp.dhis.query.JpaQueryUtils;
 import org.hisp.dhis.security.acl.AclService;
-import org.hisp.dhis.system.util.SqlUtils;
 import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
 import org.hisp.dhis.tracker.Page;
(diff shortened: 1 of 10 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -603,141 +600,113 @@ lateral jsonb_each(
...
 
     if (filters.isEmpty()) {
       return "";
     }
 
-    StringBuilder query = new StringBuilder(AND);
-    // In SQL the order of expressions linked by AND is not
-    // guaranteed.
-    // So when casting to number we need to be sure that the value
-    // to cast is really a number.
```
teleivo, line 665: Since you removed the comment I assume this not apply anymore. Why? Please explain that change in the SQL. Seems like you removed the `case`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -603,141 +600,113 @@ lateral jsonb_each(
...
+    StringBuilder query = new StringBuilder();
     List<String> filterStrings = new ArrayList<>();
 
     for (int i = 0; i < filters.size(); i++) {
       QueryFilter filter = filters.get(i);
-      StringBuilder filterString = new StringBuilder();
       final String queryCol =
           isNumericTea ? castToNumber(teaValueCol + ".value") : lower(teaValueCol + ".value");
       int itemType = isNumericTea ? Types.NUMERIC : Types.VARCHAR;
       String parameterKey = "attributeFilter_%s_%d".formatted(teaValueCol.replace("\"", ""), i);
```
teleivo, line 649: This looks odd. We pass the quoted UID `String teaValueCol = quote(teaUid);` in to then unquote it. We should be able to pass in the UID and quote it only when/where necessary, right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -603,141 +600,113 @@ lateral jsonb_each(
...
+    for (TrackedEntityAttribute attribute : params.leftJoinAttributes()) {
+      attributes
           .append(" left join trackedentityattributevalue as ")
-          .append(quote(orderAttribute.getUid()))
+          .append(quote(attribute.getUid()))
           .append(" on ")
-          .append(quote(orderAttribute.getUid()))
+          .append(quote(attribute.getUid()))
           .append(".trackedentityid = TE.trackedentityid ")
           .append("and ")
```
teleivo, line 702: It might be a drop in the ocean but I don't get why we make calls to append for strings we should concatenate right away like `".trackedentityid = TE.trackedentityid and "`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -911,15 +880,17 @@ private StringBuilder getFromWhereClause(
...
-    fromBuilder.append(getFromSubQueryJoinOrderByAttributes(params));
+    // LEFT JOIN attributes we need to sort on and/or filter by.
+    fromBuilder.append(getLeftJoinFromAttributes(params));
 
     fromBuilder.append(getCategoryOptionComboQuery(user));
 
-    fromBuilder.append(dataElementAndFiltersSql);
+    fromBuilder.append(dataElementFiltersSql);
+
+    if (!params.getAttributes().isEmpty()) {
```
teleivo, line 890: Do we need this conditional? `getLeftJoinFromAttributes` and all the other methods before it do not need it. I would hope the `getWhereClauseFromAttributeFilterConditions` can already deal with an empty `params.getAttributes()`.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -603,141 +600,113 @@ lateral jsonb_each(
...
-          .append(" = TE.trackedentityid ")
-          .append(" inner join trackedentityattribute ")
-          .append(teaCol)
-          .append(" on ")
-          .append(teaValueCol + ".trackedentityattributeid")
-          .append(EQUALS)
-          .append(teaCol + ".trackedentityattributeid")
+      fromBuilder
+          .append(hlp.whereAnd())
+          .append(" TE.trackedentityid is not null ") // filters out results from event programs
```
enricocolasante, line 618: Is this filter automatically filtering out all the event programs if there is one attribute filter present in the request?
I mean, if we have a request asking for a null attribute, are we only going to return tracker events and filtering out all the program events?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: Someone correct me if I'm missing something, but we are already checking if the TEA type is numeric `[here](https://github.com/dhis2/dhis2-core/pull/20128/files#diff-68b5ae1d0a8364a34a53140313e32a1ecc8eb172c131a37abe2d136fe1244191R625)`

This flag is then used [here](https://github.com/dhis2/dhis2-core/pull/20128/files#diff-68b5ae1d0a8364a34a53140313e32a1ecc8eb172c131a37abe2d136fe1244191R646) and [here](https://github.com/dhis2/dhis2-core/pull/20128/files#diff-68b5ae1d0a8364a34a53140313e32a1ecc8eb172c131a37abe2d136fe1244191R659) to build the query.

That's why I don't think this `case` is needed anymore.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: I'm working under the assumption that:

> An event program doesn’t have attributes, only data elements.

I arrived at this conclusion because the UI doesn’t provide an option to assign attributes to an event program, and when I query the database, event programs have no entries in `program_attributes` or `trackedentityattribute`.

Given this, I need to exclude event programs from the results when filtering by the `null` operator, as attributes are always null for event programs.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
teleivo: and there is a test for this?

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
muilpp: yes, we have tests that cover the TEA being numeric and text

PR #20129, opened 2025-02-28
Enable async support in filter to allow async request processing when running routes

cjmamo:
Enable async support in filter to allow async request processing when running routes. Without this change, Spring complains as follows when running routes in Tomcat:

>Async support must be enabled on a servlet and for all filters involved in async request processing. This is done in Java code using the Servlet API or by adding \"<async-supported>true</async-supported>\" to servlet and filter declarations in web.xml.


kabaros (approved):
tested that it works - not sure if you want to add e2e tests as part of this PR or separately?

jbee (approved):

PR #20148, opened 2025-03-03
feat: Save unary operator on tracked entity filter [DHIS2-15954]

muilpp:
As with program stage working lists, I'm adding the null operator in tracked entity filters so it can be persisted to the database.

To validate the logic, I've created a new controller test, `TrackedEntityFilterControllerTest`, which replaces the `TrackedEntityFilterTest` end-to-end test.

Unrelated to this ticket, but while I'm at it, I'm using specific arguments in the change log test to create the page parameters, making the object creation explicit.


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/trackedentityfilter/JsonEntityQueryCriteria.java
```diff
@@ -0,0 +1,43 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.webapi.controller.trackedentityfilter;
+
+import org.hisp.dhis.jsontree.JsonArray;
+import org.hisp.dhis.jsontree.JsonObject;
+
+/** Representation of entity query criteria. */
+public interface JsonEntityQueryCriteria extends JsonObject {
```
ameenhere, line 34: Why was this created?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/trackedentityfilter/JsonTrackedEntityFilter.java
```diff
@@ -0,0 +1,46 @@
...
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.webapi.controller.trackedentityfilter;
+
+import org.hisp.dhis.jsontree.JsonObject;
+
+/** Representation of a JSON object with entity query criteria. */
+public interface JsonTrackedEntityFilter extends JsonObject {
```
ameenhere, line 33: Why was this created?

ameenhere (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/trackedentityfilter/TrackedEntityFilterControllerTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+    assertFalse(trackedEntityFilter.isEmpty());
+    assertEquals(programId, trackedEntityFilter.getProgram());
+    JsonObject attributeObject =
+        trackedEntityFilter.getEntityQueryCriteria().getAttributeValueFilters().getObject(0);
+    assertEquals(trackedEntityAttributeId, attributeObject.getString("attribute").string());
+    assertEquals(Boolean.TRUE, attributeObject.getBoolean("null").booleanValue());
+  }
+
+  @Test
```
enricocolasante, line 205: Would it make sense to add a test case when the value for the null operator is `null`?
The creation should fail, right?

enricocolasante (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/trackedentityfilter/TrackedEntityFilterControllerTest.java
muilpp: if the value is `null` it doesn't fail, and, regardless of the operator, it will only store the attribute.
as an example:

`"attributeValueFilters": [{"eq": null, "attribute": "ruQQnf6rswq"}]`
`"attributeValueFilters": [{"null": null, "attribute": "ruQQnf6rswq"}]`

both store only `{"attributeValueFilters": [{"attribute": "ruQQnf6rswq"}]}`

PR #20158, opened 2025-03-04
chore: Create findEnrollment method [DHIS2-18541]

enricocolasante:

***Changes in this PR***

- Create `findEnrollment` method and use it when needed.
- Catch `ForbiddenException` in `getEnrollment` as it can only be thrown when mapping `EnrollmentOperationParams` so if falls into the case here 
`throw new IllegalArgumentException(
          "this must be a bug in how the EnrollmentOperationParams are built");`

***Challenges in tests***

***Next Steps***
- Fix exception thrown from `TrackedEntityAttributeValueService`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -33,6 +33,7 @@ import static org.hisp.dhis.user.CurrentUserUtil.getCurrentUserDetails;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
```diff
@@ -39,12 +40,40 @@
...
+   * Finds the enrollment that matches the given {@code UID} based on the privileges of the
+   * currently authenticated user. Returns an {@link Optional} indicating whether the enrollment was
+   * found.
+   *
+   * @return an {@link Optional} containing the enrollment if found, or an empty {@link Optional} if
+   *     not
+   */
   @Nonnull
-  Enrollment getEnrollment(UID uid) throws ForbiddenException, NotFoundException;
+  Optional<Enrollment> findEnrollment(@Nonnull UID uid);
```
muilpp, line 53: I see it's only used in tests. Should it be moved to a test helper class instead?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -97,7 +108,7 @@ public Enrollment getEnrollment(@Nonnull UID uid, @Nonnull EnrollmentParams para
               .enrollmentParams(params)
               .build();
       enrollments = getEnrollments(operationParams, PageParams.single());
-    } catch (BadRequestException e) {
+    } catch (BadRequestException | ForbiddenException e) {
```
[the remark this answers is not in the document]
enricocolasante: Because `ForbiddenException ` can only be thrown when mapping EnrollmentOperationParams so if falls into the case here
`throw new IllegalArgumentException( "this must be a bug in how the EnrollmentOperationParams are built");`

The `EnrollmentOperationParams` are not supposed to throw a `ForbiddenException`, if that happens it is a bug.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentService.java
enricocolasante: I think we agreed that this `get` and `find` are the 2 different ways to retrieve entities, even if it is not (yet) used it can make sense that our services are exposing such a method.

muilpp (approved):

teleivo (approved):

PR #20160, opened 2025-03-04
fix: Add tests to validate expected logic [DHIS2-19050]

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapperTest.java
```diff
@@ -521,36 +524,6 @@ void shouldFailWhenGlobalSearchAndNoAttributeSpecified()
         exception.getMessage());
   }
 
-  @Test
-  void shouldFailWhenGlobalSearchAndMaxTeLimitReached()
```
ameenhere, line 525: Why was this removed? To my naked eye, this "kinda" looks correct or am i missing something?

muilpp: because I'm testing everything related to the limits (program, TET and system) in `TrackedEntityMaxLimitServiceTest`
This particular test is now [here](https://github.com/dhis2/dhis2-core/pull/20160/files#diff-7ddf1e75d016bdef7a4b29ca6b8ac85c6e9f749aba027d335a24b5261db3aa6aR105) and [here](https://github.com/dhis2/dhis2-core/pull/20160/files#diff-7ddf1e75d016bdef7a4b29ca6b8ac85c6e9f749aba027d335a24b5261db3aa6aR127)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,6 +287,20 @@ private void validateTrackedEntityAttributeFilters(
...
+  private void validatePagination(PageParams pageParams) throws BadRequestException {
+    if (pageParams == null) {
+      return;
+    }
+
+    int maxLimit = systemSettingsService.getCurrentSettings().getTrackedEntityMaxLimit();
+    if (pageParams.getPageSize() > maxLimit) {
+      throw new BadRequestException(
+          String.format(
+              "Invalid page size: %d. It must not exceed the system limit of %d.",
```
teleivo, line 299:
```suggestion
              "Invalid page size: %d. It must not exceed the system limit of 'fill in the config name' which is currently set to %d.",
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,6 +287,20 @@ private void validateTrackedEntityAttributeFilters(
     }
   }
 
+  private void validatePagination(PageParams pageParams) throws BadRequestException {
+    if (pageParams == null) {
+      return;
+    }
+
```
teleivo, line 294: https://github.com/dhis2/dhis2-core/blob/44fa24febae14f6a20a5d7fcb49e3737d7586e56/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java#L574-L586 should actually be done by the `PageParams` constructor 😅 it was ok before when our services were only used by our controllers.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -114,6 +125,8 @@ public TrackedEntityQueryParams map(
     validateTrackedEntityAttributeFilters(
         program, requestedTrackedEntityType, operationParams, orgUnits, params);
 
+    validatePagination(pageParams);
```
teleivo, line 128: Maybe do this early as this seems like one of the cheaper checks we do. So we return early in the scenario and don't fetch all these other things.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -318,9 +303,7 @@ private String getCountQueryWithMaxTrackedEntityLimit(TrackedEntityQueryParams p
         + getQuerySelect(params)
         + "FROM "
         + getFromSubQuery(params, true, null)
-        + (params.getEnrolledInTrackerProgram().getMaxTeiCountToReturn() > 0
-            ? getLimitClause(params.getEnrolledInTrackerProgram().getMaxTeiCountToReturn() + 1)
-            : "")
+        + (params.getMaxTeLimit() > 0 ? getLimitClause(params.getMaxTeLimit() + 1) : "")
```
ameenhere, line 306: I think we can optimize this further. This count method would/should only be called if there is maxTeLimit >0 . So either we start this method with something like 

```java
if (!params.hasMaxTeLimit())
return 1;
``` 
We do not call this method if maxTeLimit is 0, but this is just to prevent another caller from calling this method maybe. 

Once that is done, then we can safely guarantee the existence of the limit in the params and just use that. And avoid this ternary expression.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -1059,28 +1021,10 @@ private String getFromSubQueryLimitAndOffset(
...
-          .toString();
-    } else {
-      return limitOffset
-          .append(LIMIT)
-          .append(SPACE)
-          .append(limit + 1) // We add +1, since we use this limit to
-          // restrict a user to search to wide.
-          .append(SPACE)
-          .toString();
+    } else if (systemTeLimit != 0) {
```
ameenhere, line 1024: systemTeLimit > 0

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
teleivo: @ameenhere 

This count is only used if `totalPages=true` to show the count in the pager. Why do we need the +1 in that case? I also don't get why we use `params.getMaxTeLimit()` here 🤔 

If I get this: `params.getMaxTeLimit()` is set if there is a TET/Program limit. The check that the users search is within this limit is already done in the service.

Should the `systemTeLimit` not also be considered here? We did validate that the `pageSize` <= `systemTeLimit` in the service but the count can exceed that

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityMaxLimitServiceTest.java
```diff
@@ -0,0 +1,287 @@
...
+    updateTrackedEntityTypeMaxLimit(trackedEntityType.getUid(), 0);
+    updateProgramMaxLimit(program.getUid(), 0);
+    updateSystemSettingLimit("0");
+  }
+
+  @Test
+  void shouldReturnEntitiesWhenGlobalSearchAndProgramMaxLimitNotReached()
+      throws ForbiddenException, BadRequestException, NotFoundException {
+    updateProgramMaxLimit(program.getUid(), 5);
+    injectSecurityContextUser(userService.getUser("FIgVWzUCkpw"));
```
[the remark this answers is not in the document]
muilpp: not sure there's a way, as this user doesn't have authorization to update the limits, which is done at the beginning of every test... but I still need it to be the one to execute the request

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
muilpp: This is the count used by the service to verify we are not over the limit of the program or TET when searching outside the capture scope.

In the service, there's the validation `maxTeLimit>0`, and since I think of the store as a class that only the tracker team uses, I didn't think about adding the same guard clause there, but I'm open to it if you feel it's beneficial.

@teleivo can it be you are referring to `getCountQuery(TrackedEntityQueryParams params)`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
teleivo: I mean we can move the logic of `validatePaginationBounds` into the `PageParams` constructor. Before `PageParams` were validated in the web and only created by the web layer. Now that our service API is used by different modules we need to make sure the `PageParams` that is passed in is well formed. So `PageParams` should have the `validatePaginationBounds` validation. I am not talking about the system limits here 👌🏻 .

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityMaxLimitServiceTest.java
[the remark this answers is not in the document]
teleivo: could the update method switch user set the limit and switch back to the previous one?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
teleivo: > @teleivo can it be you are referring to getCountQuery(TrackedEntityQueryParams params)?

you are right @muilpp , my bad 😅 

You are also right that the store is only used by our service. So we don't need to redo any validation as long as we ensure the inputs to the store are safe to use from all code paths 👌🏻

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityMaxLimitServiceTest.java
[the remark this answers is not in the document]
muilpp: it could, but then we'd need to do the same when updating TET and system limit too
and in the tests where we update multiple limits, we'd be switching the user back and forth unnecessarily

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/PageParams.java
```diff
@@ -55,13 +56,21 @@ public class PageParams {
...
+  private PageParams(Integer page, Integer pageSize, boolean pageTotal) {
     this.page = Objects.requireNonNullElse(page, DEFAULT_PAGE);
     this.pageSize = Objects.requireNonNullElse(pageSize, DEFAULT_PAGE_SIZE);
     this.pageTotal = pageTotal;
   }
 
-  public static PageParams of(Integer page, Integer pageSize, boolean pageTotal) {
+  public static PageParams of(Integer page, Integer pageSize, boolean pageTotal)
+      throws BadRequestException {
+    if (page != null && page < 1) {
```
teleivo, line 67: move that validation into the constructor

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/view/PageTest.java
```diff
@@ -122,6 +124,21 @@ void shouldSetNextPage() {
         "fields=displayName");
   }
 
+  @Test
+  void shouldFailWhenPageIsNotNullAndSmallerThanOne() {
```
teleivo, line 128: these tests do not belong in the `view/PageTest` as you are testing `tracker/PageParams`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/PageParams.java
muilpp: isn't the point of using a factory to run the validations within it?
if anything, I think we should move the validations in the constructor here. Would you agree?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
teleivo: Please look at the flow (maybe debug). The params limit is set after this method is even reached AFAIK. So we would never actually add a `limit` to the SQL query.

ameenhere: Yes the maxTeLimit obtained (from Program/TET) is not set into the params early enough.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,22 +287,35 @@ private void validateTrackedEntityAttributeFilters(
...
 
-  private void validateGlobalSearchParameters(TrackedEntityQueryParams params, UserDetails user)
-      throws IllegalQueryException {
-    if (!isLocalSearch(params, user)) {
+  private void validatePagination(PageParams pageParams) throws BadRequestException {
+    if (pageParams == null) {
+      return;
+    }
 
+    int maxLimit = systemSettingsService.getCurrentSettings().getTrackedEntityMaxLimit();
```
ameenhere, line 295: Lets change this variable name similar to the other place, `systemMaxPageSize`  here?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -359,31 +358,35 @@ private void validateSearchableAttributes(
...
+    }
+  }
 
-      if (!params.hasTrackedEntities() && isProgramMinAttributesViolated(params)) {
-        throw new IllegalQueryException(
-            "At least "
-                + params.getEnrolledInTrackerProgram().getMinAttributesRequiredToSearch()
-                + " attributes should be mentioned in the search criteria.");
-      }
+  private int getMaxTeLimit(TrackedEntityQueryParams params) {
```
ameenhere, line 382: I am thinking why can't this method be moved inside the params? As a convenience method that gets max te limit based on the params?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -309,13 +309,12 @@ private void validateSearchOutsideCaptureScopeParameters(
...
-      if (maxTeLimit > 0
-          && trackedEntityStore.getTrackedEntityCountWithMaxTrackedEntityLimit(params)
-              > maxTeLimit) {
-        throw new IllegalQueryException("maxteicountreached");
+      if (!params.hasTrackedEntities()) {
+        validateMinAttributesToSearch(params);
       }
+
+      params.setMaxTeLimit(getMaxTeLimit(params));
+      validateMaxTrackedEntitiesInResponse(params);
```
ameenhere, line 317: The method name is not clear enough in my opinion. 

Some suggestions:
validateTrackedEntitySearchResultsWithinLimit
validateMaxTrackedEntitySearchResults
validateSearchResultsWithinMaxTeLimit
validateTotalResultsWithinMaxTeLimit

or some combination of these words. 

"InResponse" sounds misleading since it could indicate the paginated response.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -1012,43 +995,22 @@ private String getQueryOrderBy(TrackedEntityQueryParams params, boolean innerOrd
...
-   * @return a SQL LIMIT and OFFSET clause, or empty string if no LIMIT can be deducted.
+   * @return a SQL LIMIT and OFFSET clause, or empty string if no LIMIT can be determined.
    */
-  private String getFromSubQueryLimitAndOffset(
-      TrackedEntityQueryParams params, PageParams pageParams) {
+  private String getFromSubQueryLimitAndOffset(PageParams pageParams) {
     StringBuilder limitOffset = new StringBuilder();
-    int limit = params.getMaxTeLimit();
-    int teQueryLimit = settingsProvider.getCurrentSettings().getTrackedEntityMaxLimit();
+    int systemMaxPageSize = settingsProvider.getCurrentSettings().getTrackedEntityMaxLimit();
```
teleivo, line 1011:
```suggestion
    int systemMaxLimit = settingsProvider.getCurrentSettings().getTrackedEntityMaxLimit();
```

it's not a page size/not related to pagination

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,22 +287,34 @@ private void validateTrackedEntityAttributeFilters(
...
 
-  private void validateGlobalSearchParameters(TrackedEntityQueryParams params, UserDetails user)
-      throws IllegalQueryException {
-    if (!isLocalSearch(params, user)) {
+  private void validatePagination(PageParams pageParams) throws BadRequestException {
+    if (pageParams == null) {
+      return;
+    }
 
+    int systemMaxPageSize = systemSettingsService.getCurrentSettings().getTrackedEntityMaxLimit();
```
teleivo, line 295:
```suggestion
    int systemMaxLimit = systemSettingsService.getCurrentSettings().getTrackedEntityMaxLimit();
```

it's not a page size

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,22 +287,34 @@ private void validateTrackedEntityAttributeFilters(
...
+      throw new BadRequestException(
+          String.format(
+              "Invalid page size: %d. It must not exceed the system limit of KeyTrackedEntityMaxLimit %d.",
+              pageParams.getPageSize(), systemMaxPageSize));
+    }
+  }
+
+  private void validateSearchOutsideCaptureScopeParameters(
+      TrackedEntityQueryParams params, UserDetails user) throws IllegalQueryException {
+    if (!isSearchInCaptureScope(params, user)) {
```
teleivo, line 306: Please look for guard clauses like this in general where inverting it and returning early reduces nesting. This is a general thing we can apply to all our code to make it easier to read.

So you can invert it and write it like

```java
if (isSearchInCaptureScope(params, user)) {
  return;
}
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -274,22 +287,34 @@ private void validateTrackedEntityAttributeFilters(
...
         validateSearchableAttributes(params, searchableAttributeIds);
       }
 
-      int maxTeLimit = getMaxTeLimit(params);
-      if (maxTeLimit > 0
-          && trackedEntityStore.getTrackedEntityCountWithMaxTrackedEntityLimit(params)
-              > maxTeLimit) {
-        throw new IllegalQueryException("maxteicountreached");
+      if (!params.hasTrackedEntities()) {
+        validateMinAttributesToSearch(params);
```
teleivo, line 313: if you then apply the above suggestion you can move these guards inside the targeted validation functions

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
teleivo: I think this shows the awkwardness of this code. Even if we do what you suggest we still need to ensure that `params.getTrackedEntityType()` and `params.getEnrolledInTrackerProgram()` are set before calling `trackedEntityStore.getTrackedEntityCountWithMaxLimit`. This temporal coupling is error prone. The issue I flagged last of reading a field before it was written is exactly that.

We should not have to mutate `TrackedEntityQueryParams` after it has been created. Is there not a way to move the validation of `trackedEntityStore.getTrackedEntityCountWithMaxLimit` into the store? Thereby making `trackedEntityStore.getTrackedEntityCountWithMaxLimit` private and removing the temporal coupling. We then create `TrackedEntityQueryParams` with `params.getTrackedEntityType()` and `params.getEnrolledInTrackerProgram()` and pass the final `TrackedEntityQueryParams` to the store.

muilpp: I could add a window function to perform the count in the store and run it along with the current query. This way, we'd get the same results as we do now, plus the total count, all in one. Wrapping it in a `CASE` clause would ensure the count is only executed when the search is outside the capture scope.

However, I'm not sure how this change would affect performance. Besides, didn't we agree to leave this validation in the store for now? This isn't really an effective way to prevent users from snooping around, and we plan to implement a better solution in v43.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -1021,15 +1037,26 @@ private String getFromSubQueryLimitAndOffset(PageParams pageParams) {
...
-          .append(systemMaxPageSize)
+          .append(systemMaxLimit)
           .append(SPACE)
           .toString();
     }
 
     return limitOffset.toString();
   }
+
+  /** Returns the maximum TE retrieval limit. 0 no limit. */
```
teleivo, line 1052: Maybe add that "the assumption is that only one of TET/Program will ever be set" so we don't forget 😅

teleivo (approved):

PR #20171, opened 2025-03-05
feat: bundle global app shell

amcgee:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -405,7 +414,13 @@ public void reloadApps() {
         .filter(app -> !exists(app.getKey()))
         .forEach(this::installApp);
 
-    jCloudsAppStorageService.discoverInstalledApps().values().forEach(this::installApp);
+    jCloudsAppStorageService.discoverInstalledApps().values().stream()
+        .filter(app -> !exists(app.getKey()))
+        .forEach(this::installApp);
+
+    bundledAppStorageService.discoverInstalledApps().values().stream()
```
david-mackessy, line 421: So this is where a bundled app would not get 'loaded' if it has an installed version, right? If so, it would be nice to distinguish this important info somehow (comment maybe? //only load bundled app if no installed version present )

david-mackessy (commented):
In principal looks good to me 👍 (acknowledging that you already acknowledged the commented-out code and todos)

I like:
- the `BundledAppStorageService` as this demystifies where we load static resources for bundled apps from and brings it front and center (big plus)
- simplifying `AppOverrideFilter` as it seemed to be doing too much

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
amcgee: Yeah it's the next line (`.filter(app -> !exists(app.getKey()))`)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -405,7 +414,13 @@ public void reloadApps() {
         .filter(app -> !exists(app.getKey()))
         .forEach(this::installApp);
 
-    jCloudsAppStorageService.discoverInstalledApps().values().forEach(this::installApp);
+    jCloudsAppStorageService.discoverInstalledApps().values().stream()
```
kabaros, line 417: +1 for David's point - a comment here would be great (especially that order matters, I take it - this works because we start with jCloudApps before checking bundleAppStorage from what I can see)

in all cases, it took me a bit of time to figure it out

kabaros (approved):
lgtm - left some minor comments/questions. I tested apps (bundled, overridden, installed) and they work as expected, including shortcuts.

I didn't test Line-Listing app (I am not sure about the bit around HTML templates in general).

One last thing is that I think we should mark the change to /apps returning bundled apps as breaking (I am not sure if you needed to change anything in app-managemt app now to accommodate for bundled apps being returned).

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -99,22 +105,24 @@ public DefaultAppManager(
           AppStorageService localAppStorageService,
       @Qualifier("org.hisp.dhis.appmanager.JCloudsAppStorageService")
           AppStorageService jCloudsAppStorageService,
+      @Qualifier("org.hisp.dhis.appmanager.BundledAppStorageService")
```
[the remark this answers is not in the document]
amcgee: Yes they can I think, with the exception of appCache which needs to be explicitly initialized.  That would be a lot nicer but I didn't get to it yet.

david-mackessy (approved):

PR #20178, opened 2025-03-05
fix: Org Unit merge assigns merged units to correct users [DHIS2-19020]

david-mackessy:
# Issue
When merging `OrganisationUnit`s, `User`s (potentially all in the system) are being assigned access to the merge `User`'s `OrganisationUnit`s (not the intended merge `OrganisationUnit`s).

# Cause
After creating the `UserQueryParams` with the `OrganisationUnit` sources, those `OrganisationUnit`s were being overridden to use the `User`'s `OrganisationUnit`s carrying out the merge.

In cases where the `User` either had no `organisationUnits`, `dataViewOrganisationUnits`, or `teiSearchOrganisationUnits`, then all the `User`s in the DB would be returned from the query, resulting in all `User`s getting the merge `OrganisationUnit`s, whether they had access to any of the source `OrganisationUnit`s or not.

# Fix
This issue was practically identical to #19239 and the same fix has been re-used.
- Use a dedicated query to fetch the correct `User`s to be actioned with the correct `OrganisationUnit`s

# Testing
## Automated
New integration test added checking that different users all get the appropriate merged org unit access

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/orgunit/handler/MetadataOrgUnitMergeHandler.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/orgunit/handler/MetadataOrgUnitMergeHandler.java
@@ -33,7 +33,6 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import org.hisp.dhis.category.CategoryOption;
-import org.hisp.dhis.common.BaseIdentifiableObject;
 import org.hisp.dhis.common.UID;
 import org.hisp.dhis.configuration.Configuration;
 import org.hisp.dhis.configuration.ConfigurationService;


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/orgunit/handler/MetadataOrgUnitMergeHandler.java
```diff
@@ -125,10 +127,12 @@ public void mergeOrganisationUnits(OrgUnitMergeRequest request) {
...
-        userService.getUsers(
-            new UserQueryParams()
-                .setCanSeeOwnRoles(true)
-                .setOrganisationUnits(request.getSources()));
+        userService.getUsersWithOrgUnits(
+            UserOrgUnitProperty.ORG_UNITS,
+            UID.of(
+                request.getSources().stream()
+                    .map(BaseIdentifiableObject::getUid)
+                    .collect(Collectors.toSet())));
```
jbee, line 135: I think there is a missing `of` method hiding here for `Set<UID> of(Collection<? extends UidObject> c)` and/or `Set<UID> of(Stream<? extends UidObject> c)`

jbee (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/orgunit/handler/MetadataOrgUnitMergeHandler.java
david-mackessy: Yeah, the collection one is hard to fit in as the compiler complains of type erasure with other signatures. The stream one fits in though 👍

larshelge (approved):

PR #20180, opened 2025-03-05
feat(apps): expose shortcuts for apps / WebModules [do not merge]

kabaros:
This exposes shortcuts and versions through the `/apps/menu` (and `/dhis-web-commons/menu/getModules.action`) endpoints.

There are four cases that needs to be considered for testing:
- Bundled apps
- Bundled apps that are updated to a newer version
- Installed apps from AppHub
- Manually installed apps

In order to make it possible to test this:
- Updated the AppHub url to staging - in order to publish and test fake apps with a shortcuts section
- Pointed the bundled apps to a version that includes the shortcuts
 

## ToDo:

- [ ]  (Try to) move the logic for reading the webmanifest to apps discovery so that installed and bundled apps are treated the same way
- [ ] **Revert the updated `apps-.bundled.json`, and the `ConfigurationKey` file**
- [ ]  ? Update the `build` command in app-platform to copy the shortcuts to manifest.webapp


amcgee:
Taking a look at this - @kabaros can you add a Jira reference so it gets linked in the changelog?

amcgee (approved):
Code LGTM!

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppShortcut.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
+import java.io.Serializable;
+import org.hisp.dhis.common.DxfNamespaces;
+
+@JacksonXmlRootElement(localName = "shortcut", namespace = DxfNamespaces.DXF_2_0)
+public class AppShortcut implements Serializable {
+  /** Determines if a de-serialized file is compatible with this class. */
+  private static final long serialVersionUID = -8865601558938806456L;
+
+  /** Required. */
```
david-mackessy, line 41: If it is required, should there be code enforcing that somewhere? It looks like there's an empty constructor below also, so it doesn't seem to be required. Maybe remove or update the comment so it's not contradicting the code.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppShortcut.java
```diff
@@ -0,0 +1,65 @@
...
+  @JsonProperty
+  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
+  private String name;
+
+  /** Required. */
```
david-mackessy, line 46: same here

david-mackessy (approved):
looks good, just left a comment about a comment :)

PR #20186, opened 2025-03-06
chore: OpenAPI HTML - more toned colors, higher contrast

jbee:
The initial color scheme and coloring was mostly inspired by the swagger rendering.
But it seemed very overloaded with color. Recently I saw a IDE that gave me an idea of how to keep the color coding but tone it way down to read more like a documentation. 

The result looks like this:
DHIS2-API-2-42-03-06-2025_03_48_PM
DHIS2-API-2-42-03-06-2025_03_48_PM (1)

Since the new coloring does not use transparency on large areas I also feel the usability is better when viewing the entire API.


netroms (approved):

david-mackessy (approved):
looks nice

PR #20196, opened 2025-03-07
fix: Return program attributes in change log [DHIS2-19141]

muilpp:
After the recent refactor of the `trackedEntityService`, we were no longer fetching program attributes in the tracked entity change logs, even if a program was specified.
This issue was not detected earlier because the test that validates this behavior used a tracked entity attribute that, although defined as a program attribute, was also defined as a tracked entity type attribute. As a result, it was still present in the output.

I'm also removing `getAllUserReadableTrackedEntityAttributes` because it was not used anywhere.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
@@ -27,7 +27,10 @@
  */
 package org.hisp.dhis.tracker.export.trackedentity;
 
+import static java.util.Collections.emptyList;
+
 import java.util.Date;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -97,10 +104,14 @@ public Page<TrackedEntityChangeLog> getTrackedEntityChangeLog(
     TrackedEntity trackedEntity =
         trackedEntityService.getTrackedEntity(
             trackedEntityUid, programUid, TrackedEntityParams.FALSE.withIncludeAttributes(true));
+    Program program =
+        (programUid != null) ? programService.getProgram(programUid.getValue()) : null;
 
     Set<UID> trackedEntityAttributes =
         trackedEntityAttributeService
-            .getTrackedEntityTypeAttributes(trackedEntity.getTrackedEntityType())
+            .getAllUserReadableTrackedEntityAttributes(
```
enricocolasante, line 112: `trackedEntityService.getTrackedEntity(
            trackedEntityUid, programUid, TrackedEntityParams.FALSE.withIncludeAttributes(true));`
is getting the tracked entity and all the tracked entity type attributes and program attributes, should we simply get the attributes with `trackedEntity.trackedEntityAttributeValues()`?

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityChangeLogService.java
```diff
@@ -52,6 +53,8 @@ public class DefaultTrackedEntityChangeLogService implements TrackedEntityChange
 
   private final TrackedEntityService trackedEntityService;
 
+  private final ProgramService programService;
```
teleivo, line 56: I think you added an unused dependency here @muilpp

PR #20206, opened 2025-03-10
fix: OpenAPI small issue fixes

jbee:
Fixes a couple of issues I noticed when using OpenAPI HTML.

Issues

- [x] `/api/{entity-type}/openapi.html` did not limit scope to a single entity type
- [x] open object type in new tab link broken on play as context path wasn't considered
- [x] `@OpenApi.Property` was ignored when `@JsonIgnore` is present

Improvement

- [x] operations by default list the success response code only
- [x] operations link to GH controller file it is declared in (if not merged)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/openapi/Property.java
```diff
@@ -142,16 +142,16 @@ private static List<Property> propertiesIn(Class<?> object) {
...
+        .filter(Property::isAccessor)
         .filter(method -> !ignoredFields.contains(getJavaPropertyName(method)))
         .forEach(
             method ->
                 addMethod.accept(
                     method, fieldsByJavaPropertyName.get(getJavaPropertyName(method))));
     if (properties.isEmpty() || includeByDefault) {
       methodsIn(object)
-          .filter(Property::isGetter)
+          .filter(Property::isAccessor)
```
jbee, line 154: @cjmamo Please have a look at this change again in the context of your use case. 

I checked back with the changes you had made a while back to the properties selection logic. Back then you did also consider both getters and setters and I think I changed that later to just getters. As far as I can recall I did that thinking setters were unnecessary. But that is not the case. Sometimes we use setters to allow the property input to be another type or to be given using an alias and similar.

larshelge (approved):

david-mackessy (approved):

PR #20212, opened 2025-03-10
fix: Unidirectional relationships only consider from end when exporting [DHIS2-19169]

enricocolasante:
***How relationships should work***
Relationships are entities that connect 2 tracker objects (tracked entity, enrollment and event) and they can be of 2 types: unidirectional or bidirectional.

Screenshot 2025-03-11 at 12 36 56

A unidirectional relationship from A to C connects A to C, but does not connect C to A.
A bidirectional relationship from A to B connects A to B and connects B to A.

`/tracker/trackedEntities/{TrackedEntityA}?fields=relationships` returns `{relationshipA, relationshipB}`
`/tracker/trackedEntities/{TrackedEntityB}?fields=relationships` returns `{relationshipA}`
`/tracker/trackedEntities/{TrackedEntityC}?fields=relationships` returns `{}`

What does it mean "connects"?
It means that retrieving entity A, we can retrieve the relationships for A (`{relationshipA, relationshipB}`) and retrieve the other ends of the relationships (`{B,C}`).
If a relationship is unidirectional only "works" in one direction, so in the case of a unidirectional relationship from A to C, when retrieving C, we cannot retrieve such relationship (ie. `/tracker/trackedEntities/{TEc}?fields=relationships` should not return `relationshipB`).

***Status before the fix***
We were treating all relationships as bidirectional, any end of the relationship could retrieve the relationship.

***Status after the fix***
Tracker objects could retrieve only "outgoing" relationships (where the relationship is bidirectional or the object  is in the "from" end and the relationship is unidirectional).

Is "outgoing" a good and clear enough term to be used in the docs?

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
@@ -78,6 +78,10 @@ public class DefaultRelationshipService implements RelationshipService {
     return relationshipItems.stream()
         .filter(
             ri ->
+                ri.getRelationship().getFrom().equals(ri)
+                    || ri.getRelationship().getRelationshipType().isBidirectional())
+        .filter(
+            ri ->
                 trackerAccessManager
                     .canRead(CurrentUserUtil.getCurrentUserDetails(), ri.getRelationship())
                     .isEmpty())
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
@@ -317,7 +317,8 @@ class HibernateRelationshipStore extends SoftDeleteHibernateObjectStore<Relation
 
     toSubQuery.where(
         builder.equal(root.get("to"), toRoot.get("id")),
-        builder.equal(toRoot.get(relationshipEntityType), entity.getId()));
+        builder.equal(toRoot.get(relationshipEntityType), entity.getId()),
+        builder.equal(root.get("relationshipType").get("bidirectional"), true));
 
     toSubQuery.select(toRoot.get("id"));
 
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
@@ -317,8 +317,7 @@ class HibernateRelationshipStore extends SoftDeleteHibernateObjectStore<Relation
 
     toSubQuery.where(
         builder.equal(root.get("to"), toRoot.get("id")),
-        builder.equal(toRoot.get(relationshipEntityType), entity.getId()),
-        builder.equal(root.get("relationshipType").get("bidirectional"), true));
+        builder.equal(toRoot.get(relationshipEntityType), entity.getId()));
 
     toSubQuery.select(toRoot.get("id"));
 
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/deduplication/merge/PotentialDuplicatesRelationshipTests.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/deduplication/merge/PotentialDuplicatesRelationshipTests.java
@@ -82,8 +82,15 @@ public class PotentialDuplicatesRelationshipTests extends PotentialDuplicatesApi
         .getTrackedEntity(teA + "?fields=*")
         .validate()
         .statusCode(200)
-        .body("relationships", hasSize(2))
-        .body("relationships.relationship", hasItems(relationship3, relationship2));
+        .body("relationships", hasSize(1))
+        .body("relationships.relationship", hasItems(relationship2));
+
+    trackerImportExportActions
+        .getTrackedEntity(teC + "?fields=*")
+        .validate()
+        .statusCode(200)
+        .body("relationships", hasSize(1))
+        .body("relationships.relationship", hasItems(relationship3));
   }
 
   @Test
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
@@ -265,7 +265,7 @@ class EnrollmentsExportControllerTest extends PostgresControllerIntegrationTestB
 
   @Test
   void
-      shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnToEndOfAUnidirectionalRelationship() {
+      shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnTheToSideOfAUnidirectionalRelationship() {
     Relationship relationship = get(Relationship.class, "p53a6314631");
 
     assertNotNull(
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
@@ -55,6 +55,7 @@ import org.hisp.dhis.jsontree.JsonList;
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.Event;
 import org.hisp.dhis.relationship.Relationship;
+import org.hisp.dhis.relationship.RelationshipType;
 import org.hisp.dhis.test.webapi.PostgresControllerIntegrationTestBase;
 import org.hisp.dhis.trackedentity.TrackedEntity;
 import org.hisp.dhis.trackedentity.TrackedEntityAttribute;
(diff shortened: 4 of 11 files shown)


ameenhere (approved):

muilpp:
do we need to use any term at all?
I'm asking because with the tracked entity limits we used the terms "local search" and "global search". They are just another way of saying "search in/out of the capture scope", which is something that everyone in the DHIS2 community understands, maybe here we have a similar issue.

could we just add this to the docs?
_Tracker objects can retrieve bidirectional relationships, or unidirectional relationships if the object is in the "from" end of that relationship._

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
```diff
@@ -262,6 +263,30 @@ void shouldGetEnrollmentWithRelationshipsWhenRelationshipIsDeletedAndIncludeDele
         () -> assertHasMember(jsonRelationship, "bidirectional"));
   }
 
+  @Test
+  void
+      shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnToEndOfAUnidirectionalRelationship() {
```
muilpp, line 268: I'm not sure I'm reading this right, did you mean to say `WhenEnrollmentIsOn**The**End`?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
```diff
@@ -478,6 +507,30 @@ void shouldGetRelationshipsByTrackedEntity() {
...
   }
 
+  @Test
+  void
+      shouldGetNoRelationshipsByTrackedEntityWhenRelationshipTypeIsUnidirectionalAndTrackedEntityIsOnToEnd() {
+    RelationshipType relationshipType = manager.get(RelationshipType.class, "m1575931405");
+    relationshipType.setBidirectional(false);
+    manager.update(relationshipType);
+
+    assertEquals(
```
teleivo, line 517: Can you use the same setup like in

https://github.com/dhis2/dhis2-core/pull/20212/files#diff-ac504135039dad6a3a06c6969c94ff471fa3a1c9f1bd10f3441d0e48e2f872d4R354-R356

? Looks shorter. Or use the same setup for all the 3 tests you've added. The above 2 use yet another style to get the entity

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/relationship/RelationshipsExportControllerTest.java
```diff
@@ -342,6 +356,21 @@ void shouldGetRelationshipsByEnrollment() {
...
   }
 
+  @Test
+  void
+      shouldGetNoRelationshipsByEnrollmentWhenRelationshipTypeIsUnidirectionalAndEnrollmentIsOnToEnd() {
+    RelationshipType relationshipType =
+        manager.get(RelationshipType.class, relationship2.getRelationshipType().getIdentifier());
+    relationshipType.setBidirectional(false);
+    manager.update(relationshipType);
+    JsonList<JsonRelationship> jsonRelationships =
```
teleivo, line 366: newline to separate setup from test subject, same above

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -312,6 +312,29 @@ void shouldGetEventWithRelationshipsWhenRelationshipIsDeletedAndIncludeDeletedIs
...
+
+    RelationshipType relationshipType =
+        manager.get(RelationshipType.class, relationship.getRelationshipType().getUid());
+    relationshipType.setBidirectional(false);
+    manager.update(relationshipType);
+
+    switchContextToUser(user);
+
+    JsonList<JsonRelationship> relationships =
+        GET("/tracker/events/?events={id}&fields=relationships&includeDeleted=true", to.getUid())
```
teleivo, line 329: Why `includeDeleted=true` is the event or relationship deleted?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -312,6 +312,29 @@ void shouldGetEventWithRelationshipsWhenRelationshipIsDeletedAndIncludeDeletedIs
     assertHasMember(jsonRelationship, "bidirectional");
   }
 
+  @Test
+  void shouldGetEventWithNoRelationshipsWhenEventIsOnToEndOfAUnidirectionalRelationship() {
+    TrackedEntity from = trackedEntity();
```
teleivo, line 317: Can you use the same setup as in dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java where you get a "real" relationship from the base_data?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
teleivo, line 268:
```suggestion
      shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnTheEndOfAUnidirectionalRelationship() {
```

This applies to more tests than this

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/deduplication/merge/PotentialDuplicatesRelationshipTests.java
```diff
@@ -82,8 +82,15 @@ public void shouldAutoMergeRelationshipsWithNonSuperUser() {
         .getTrackedEntity(teA + "?fields=*")
         .validate()
         .statusCode(200)
-        .body("relationships", hasSize(2))
-        .body("relationships.relationship", hasItems(relationship3, relationship2));
```
teleivo, line 86:
```java
    String relationship2 =
        createUniDirectionalRelationship(teA, teC).extractImportedRelationships().get(0);
    String relationship3 =
        createUniDirectionalRelationship(teC, teB).extractImportedRelationships().get(0);
```

Why was rel3 in here 🤔 ? `teA` is not even involved in rel3

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -317,7 +317,8 @@ private <T extends SoftDeletableObject> Predicate[] whereConditionPredicates(
 
     toSubQuery.where(
         builder.equal(root.get("to"), toRoot.get("id")),
-        builder.equal(toRoot.get(relationshipEntityType), entity.getId()));
+        builder.equal(toRoot.get(relationshipEntityType), entity.getId()),
+        builder.equal(root.get("relationshipType").get("bidirectional"), true));
```
teleivo, line 321: why is this needed?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
```diff
@@ -76,6 +76,10 @@ public Set<RelationshipItem> findRelationshipItems(
           case RELATIONSHIP -> throw new IllegalArgumentException("Unsupported type");
         };
     return relationshipItems.stream()
+        .filter(
+            ri ->
+                ri.getRelationship().getFrom().equals(ri)
```
teleivo, line 81: I got bitten by using `.equals` due to our equals implementations in the past 😅 just double-checking that this is safe.

teleivo (commented):
You were asking for a definition, maybe this one is understandable:

The relationships shown for a tracker object only include relationships that originate in that entity. So the entity is either in a bidirectional relationship or the from entity in a unidirectional relationship.

So using the word "origin", or source/target could also work. https://www.thesaurus.com/browse/origin synonyms/antonyms often help me. Or CS related resources for this as its graph theory.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
enricocolasante: Because we need to implement the `rule` also in the store.
In `tracker/relationships?trackedEntity={TEa}` we need to return only the relationships originated from `TEa`, so we need to include in the query the relationships where `TEa` is on the `to` side only if the relationship is bidirectional.

In a perfect world I would like to have this rule written once in the code and used by the different place that need to apply it, but I think we will not be able to get away from the duplication at least in the service and in the store.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
enricocolasante: Not really, this test class is not using the base_data, so I would need to refactor the entire class.
I am not doing it in this PR.
Still hope to do it soon enough though

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportControllerTest.java
enricocolasante: That was confusing.
I renamed it to `shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnTheToSideOfAUnidirectionalRelationship`

enricocolasante: That was confusing.
I renamed it to `shouldGetEnrollmentWithNoRelationshipsWhenEnrollmentIsOnTheToSideOfAUnidirectionalRelationship`

muilpp (approved):

enricocolasante:
> do we need to use any term at all? I'm asking because with the tracked entity limits we used the terms "local search" and "global search". They are just another way of saying "search in/out of the capture scope", which is something that everyone in the DHIS2 community understands, maybe here we have a similar issue.
> 
> could we just add this to the docs? _Tracker objects can retrieve bidirectional relationships, or unidirectional relationships if the object is in the "from" end of that relationship._

Naming things is hard and if the name doesn't become part of a common language then it makes things even more difficult to understand.
But in this case, I found myself to have to repeat this rule explanation in several and different context: 
- Metadata import (When creating a RelationshipType)
- Tracker import (When updating `updatedAt` and decide which TEs to update and which not)
- TrackedEntity, Enrollment and Event exporters (when `fields=relationships` is passed and we need to decide which relationships to show)
- Relationship exporter (when deciding which relationships to show based on the entity passed in the parameters)

I think here it would be beneficial to have a shared concept that explains how relationships and the other entity connect to each other.

enricocolasante:
> You were asking for a definition, maybe this one is understandable:
> 
> The relationships shown for a tracker object only include relationships that originate in that entity. So the entity is either in a bidirectional relationship or the from entity in a unidirectional relationship.
> 
> So using the word "origin", or source/target could also work. https://www.thesaurus.com/browse/origin synonyms/antonyms often help me. Or CS related resources for this as its graph theory.

I like origin/originate.
I will update the docs and we can refine the concept there.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/DefaultRelationshipService.java
teleivo: it was more related to our many mappings and that the `equals` implementations surprised me in the past 😅 we should be ok here since relationshipItems come from our private store

PR #20213, opened 2025-03-10
fix: correct redirects for global shell

amcgee:
This updates the redirects sent in the GlobalShellFilter.  In particular:

1. Only redirect from legacy apps to the global shell when receiving a top-level navigation request (header `Sec-Fetch-Mode=navigate`) https://dhis2.atlassian.net/browse/DHIS2-18454
2. Redirect from global app shell paths `/apps/<appName>` back to the original path (i.e. `/dhis-web-<appName>` or `/api/apps/<appName>`), in case someone has started with Global Shell enabled and later disabled it. [DHIS2-19166](https://dhis2.atlassian.net/browse/DHIS2-19166)

Will follow up with integration tests for all global shell cases in separate PR, since that will be quite helpful here but the infrastructure doesn't exist yet

[DHIS2-19166]: https://dhis2.atlassian.net/browse/DHIS2-19166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/GlobalShellFilter.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/GlobalShellFilter.java
@@ -40,6 +40,7 @@ import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.hisp.dhis.appmanager.App;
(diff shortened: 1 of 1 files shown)


KaiVandivier (approved):
Approving now since it's already pretty good 🙂

larshelge (approved):

amcgee:
@KaiVandivier thanks for the thorough testing - since this has now been merged I'll create a separate PR for the case of the missing trailing slash

PR #20227, opened 2025-03-12
feat: Analyse `server.base.url` usage [DHIS2-19183]

david-mackessy:
# Summary
The `dhis.conf` config `server.base.url` was brought up for discussion as a result of [this original issue](https://dhis2.atlassian.net/browse/DHIS2-17980).

Early discussions played with the idea of making this config mandatory as it is used in many features (but not essential for server startup). This idea was dismissed as it was seen as too disruptive and intrusive for current implementations (whether they use the config property or not).

We then discussed raising an exception when this value is not present, but again, this was seen as too disruptive and intrusive.

We settled on the solution of analysing all callers of `DefaultDhisConfigurationProvider#getServerBaseUrl` and other places where `SystemInfo#instanceBaseUrl` is used. Creating tickets where `null`s are not handled (so teams can decide what action they want to take in those scenarios) and annotating the config with `@CheckForNull` so any new callers should follow that advice.

# Analysis
Callers of `DefaultDhisConfigurationProvider#getServerBaseUrl`
| Class | method | null handled | action |
|--|--|--|--|
| DefaultInterpretationService | getInterpretationLink | no | [issue created](https://dhis2.atlassian.net/browse/DHIS2-19214) |
| DefaultPushAnalysisService | generateHtmlReport | no | [issue created](https://dhis2.atlassian.net/browse/DHIS2-19214) |
| DefaultPushAnalysisService | getItemLink | no | [issue created](https://dhis2.atlassian.net/browse/DHIS2-19214) |
| DefaultPushAnalysisService | uploadImage | no | [issue created](https://dhis2.atlassian.net/browse/DHIS2-19214) |
| DefaultMessageService | asyncSendSystemErrorNotification | yes | informational purposes, no URL formed - leave as is |
| DefaultMessageService | getMessageFooter | yes | n/a |
| EmailMessageSender | sendMessage(line 97) | yes | n/a |
| EmailMessageSender | sendMessage(line 190) | yes | n/a |
| DefaultSystemService | getSystemInfo | yes/no | see next table for users of SystemInfo |
| DefaultPushAnalysisService | runPushAnalysis | yes | n/a |
| AnalyticsController | getGrid | yes | n/a |
| UserAccountController | forgotPassword | yes | n/a |

Users of the `SystemInfo` model are faced with the issue of a potential `null` `instanceBaseUrl`. Looking into its usage throughout the system, it was observed that most callers only actually use a few fields from `SystemInfo` and could have specific data classes for their specific use cases. Below is the info surrounding `SystemInfo` usage:

| Class | method | null handled | action |
|--|--|--|--|
| DefaultMonitoringService | pushMonitoringInfo | no | informational purposes, no URL formed - leave as is |
| SystemController | getSystemInfo | yes | if null here, it's left out of JSON response - expected |
| DefaultMetadataExportService | getMetadataAsObjectNode | field not used | use new record |
| DefaultMetadataExportService | getMetadataAsObjectNodeStream | field not used | use new record |
| DefaultMetadataExportService | getMetadataWithDependenciesAsNodeStream | field not used | use new record |
| MetadataSyncDelegate | shouldStopSync | field not used | use new record |
| DefaultDataStatisticsService | getDhis2Info | field not used | use new record |
| LoginConfigController | getConfig | field not used | use new record |
| AppCacheFilter | doFilter | field not used | use new record |

# Outcome
- no impact on current implementations
- `@CheckForNull` added to hopefully help prevent future issues (users of this API should be handling null cases)
- Users of `SystemInfo` are a little cleaner and have a smaller footprint (objects, service calls etc.)
- [issue created](https://dhis2.atlassian.net/browse/DHIS2-19214) for Analytics team to look into, where a valid URL does seem to be required
- 2 places where null is not checked in platform code, but a valid URL is not required, it looks like it's purely informational
- system id is retrieved from config at startup, this has not been changed


dhis-2/dhis-api/src/main/java/org/hisp/dhis/system/SystemInfo.java
```diff
@@ -136,4 +136,25 @@ public SystemInfo withoutSensitiveInfo() {
...
+   * @param buildTime
+   */
+  public record SystemVersionBuildTime(SystemIdVersionDate systemIdVersionDate, Date buildTime) {}
+
+  /**
+   * @param revision
+   * @param version
+   * @param calendar
+   */
+  public record SystemVersionCalendar(String revision, String version, String calendar) {}
```
david-mackessy, line 159: Choosing names for these was tricky, any other suggestions welcome. Didn't want them to be too verbose, while still giving an indication what they are.

netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/system/SystemInfo.java
```diff
@@ -136,4 +136,25 @@ public SystemInfo withoutSensitiveInfo() {
         .databaseInfo(databaseInfo.withoutSensitiveInfo())
         .build();
   }
+
+  /**
+   * @param id
```
larshelge, line 141: Add description of each param to make it valid Javadoc.

larshelge (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/system/SystemInfo.java
david-mackessy: I'll flatten them. I originally used the use case for the name but changed my mind when one used another one as a field, but flattening makes that a non issue now. That's better.

PR #20246, opened 2025-03-13
feat(apps): support app manifest translations [DHIS2-19189]

kabaros:
This PR makes app manifest translations possible. 

- The FE PR (https://github.com/dhis2/app-platform/pull/913)  pulls strings out of the d2.config file and makes them available for translation in Transifex - these translations are generated and saved in `manifest.webapp.translations.json`
   - followup for updated format of file: https://github.com/dhis2/app-platform/pull/926
- Old mainteance app PR: https://github.com/dhis2/maintenance-app/pull/3073
- 
This PR reads the json file generated by the build script of the app, and attaches the translations available in it to the App object. This process happens once at the app discovery stage (or when an app is installed or deleted, which retriggers the discovery). The translations are then saved with the App in the App cache used to serve the applications.

When the user makes a call to /api/menu, then the user's locale is used to find matching translation from the cache and serve it as a displayName property to the user.

This can be tested in this temporary PR: https://github.com/dhis2/dhis2-core/pull/20321

[manifest_translations.webm](https://github.com/user-attachments/assets/4277f317-ed64-4de0-af97-50885a834f83)

Example of response:

image

implements https://dhis2.atlassian.net/browse/DHIS2-19189


amcgee:
Nice work @kabaros !  There might be some overlap with #20171 which sets up loading bundled apps directly into the app cache.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppShortcut.java
```diff
@@ -33,33 +33,42 @@
...
+import lombok.Getter;
+import lombok.Setter;
 import org.hisp.dhis.common.DxfNamespaces;
 
 @JacksonXmlRootElement(localName = "shortcut", namespace = DxfNamespaces.DXF_2_0)
 public class AppShortcut implements Serializable {
   /** Determines if a de-serialized file is compatible with this class. */
   private static final long serialVersionUID = -8865601558938806456L;
 
+  /** Required. */
```
david-mackessy, line 45: This comment is misleading. Either update the comment, remove it or add constraints to the class so that this field is actually required.
There are 2 constructors, and neither require this field.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppShortcut.java
```diff
@@ -33,33 +33,42 @@
...
+  @Getter
   @JsonProperty
   @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
   private String name;
 
+  /** Required. */
```
david-mackessy, line 51: same here

dhis-2/dhis-api/src/test/java/org/hisp/dhis/appmanager/AppTest.java
```diff
@@ -161,4 +171,136 @@ void testGetPluginLaunchUrl() {
         "https://example.com/api/apps/Test-App/index.html", appWithoutPlugin.getLaunchUrl());
     assertNull(appWithoutPlugin.getPluginLaunchUrl());
   }
+
+  List<AppManifestTranslation> getTranslation(String translationJSON) throws IOException {
+    ObjectMapper objectMapper = new ObjectMapper();
+    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
david-mackessy, line 177: This is a different mapper than the one used in PRD code, it could easily become out of sync and not catch differences if the PRD one ever changes.
Ideally they'd use the same one. Whether this behaviour is covered in a different test or not, I'm not sure.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AppControllerTest.java
```diff
@@ -179,21 +187,97 @@ void testInstalledAppReturnsShortcuts() throws IOException {
...
+  void testInstalledAppReturnsTranslatedShortcuts_WithLanguageFallback() throws IOException {
+    POST("/userSettings/keyUiLocale/?userId=" + ADMIN_USER_UID + "&value=ar_IQ");
+
+    appManager.installApp(
+        new ClassPathResource("app/test-app-with-translated-manifest.zip").getFile(),
+        "test-app-with-translated-manifest.zip");
+
+    HttpResponse response = GET("/apps/menu");
+    assertEquals(HttpStatus.OK, response.status());
+    ObjectMapper mapper = new ObjectMapper();
```
david-mackessy, line 269: and here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -308,6 +309,24 @@ public App installApp(File file, String filename, Cache<App> appCache) {
     return app;
   }
 
+  private static void extractManifestTranslations(ZipFile zip, String prefix, App app) {
+    try {
+      ZipEntry translationFiles = zip.getEntry(prefix + MANIFEST_TRANSLATION_FILENAME);
+      ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
kabaros: @david-mackessy is [this pattern](https://github.com/dhis2/dhis2-core/pull/20246/commits/2d877dac348b14ebd965d667860915dd56e032b4) for reusing the mapper OK?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppShortcut.java
kabaros: removed

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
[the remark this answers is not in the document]
david-mackessy: Yep, that looks good 👍. Didn't know there were so many other instances either :) 
Maybe have it in one of the services as opposed to the App though.

kabaros:
@amcgee @KaiVandivier updated this PR as discussed:
1. Updated the format of the translations JSON to have a shortcuts property, and fields for name and description, as opposed to everything being under one map before - the translations file looks better, but this complicated the logic slightly mainly because before I could rely on the weakly typed hash map of properties, to check if a trasnlation exists or not (and fallback accordingly) but now we have instance properties for name and display. 
2. Gave precedence to manifest translations over bundled translations in the core, with the idea of them becoming deprecated eventually
3. /apps returns localised apps  
4. added `displayDescription`, similar to `displayName` in line with other places in the code base

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
```diff
@@ -520,4 +575,98 @@ public Set<String> getAdditionalAuthorities() {
...
+    if (hasManifestTranslations(translationsToUse)) {
+      App localisedApp = SerializationUtils.clone(this);
+      translateShortcuts(localisedApp, translationsToUse);
+
+      translateAppInfo(localisedApp, translationsToUse);
+
+      localisedApp.setIsLocalised(true);
+      return localisedApp;
+    }
+    return this;
```
amcgee: (minor) I wonder if it would be better to return a clone even in this case?  That way this `localise` function ALWAYS returns a clone of the original object, instead of possible returning a mutable reference to the original...

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
```diff
@@ -520,4 +575,98 @@ public Set<String> getAdditionalAuthorities() {
...
+    for (AppManifestTranslation t : translationsToUse) {
+      if (!StringUtils.isEmpty(t.getTitle())) {
+        nameToUse = t.getTitle();
+      }
+      if (!StringUtils.isEmpty(t.getDescription())) {
+        descriptionToUse = t.getDescription();
+      }
+    }
+    localisedApp.setDisplayName(nameToUse);
+    localisedApp.setDisplayDescription(descriptionToUse);
```
amcgee: It seems like maybe this logic could be a little simpler...?  Since the fallback is already implemented in `getDisplayName` and `getDisplayDescription`

```suggestion
    for (AppManifestTranslation t : translationsToUse) {
      if (!StringUtils.isEmpty(t.getTitle())) {
        localisedApp.setDisplayName( t.getTitle() );
      }
      if (!StringUtils.isEmpty(t.getDescription())) {
        localisedApp.setDisplayDescription( t.getDescription() );
      }
    }
```

amcgee: Also... won't this "fall back" from the more specific locale to the less specific language...?  Seems that logic is a bit hard to follow in the code here since we're constructing a list with implicit ordering.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
```diff
@@ -520,4 +575,98 @@ public Set<String> getAdditionalAuthorities() {
...
+    String script = locale.getScript();
+
+    Optional<AppManifestTranslation> matchingLocale = getMatchingLocale(language, country, script);
+    Optional<AppManifestTranslation> matchingLanguage = getMatchingLanguage(language);
+
+    var list = new ArrayList<AppManifestTranslation>();
+    matchingLocale.ifPresent(list::add);
+    matchingLanguage.ifPresent(list::add);
+
+    return list;
```
amcgee: Maybe this could explicitly "merge" the locale and language translations to return just a single `AppManifestTranslation` instead of a list?

Potential refactor for the future... not necessary immediately

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/BundledAppStorageService.java
```diff
@@ -104,17 +103,44 @@ public Map<String, App> discoverInstalledApps() {
     return apps;
   }
 
+  private void checkManifestTranslations(App app) {
+    try {
+      // Read translations for possible manifest translations
+      String resourceName = app.getFolderName() + "/manifest.webapp.translations.json";
```
amcgee: Should this use the static constant defined in AppStorageService ..?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/BundledAppStorageService.java
```diff
@@ -92,6 +90,7 @@ public Map<String, App> discoverInstalledApps() {
           app.setAppStorageSource(AppStorageSource.BUNDLED);
           app.setFolderName(path.replaceAll("/manifest.webapp$", ""));
 
+          checkManifestTranslations(app);
```
amcgee: Why do we have to do this here but not for other app storage types..?  (Probably I'm just missing something)

kabaros: you're right .. we should do it in other storage types. I will do it in a followup PR though if you don't mind..

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
kabaros: I've refactored the way we do this fallback logic, implementing merge like you suggested - maybe it can still be improved so that the merge is done on discovery to fill empty values but at least, this makes the logic a bit easier to follow https://github.com/dhis2/dhis2-core/pull/20246/commits/c530de96ae7e1735c2092ba6c00725db7750cff0

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
kabaros: fixed in the new logic .. and added a test

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
amcgee: Nice!  Should `merge` be non-destructive?  As currently written it looks like it will modify *both* source and target object... maybe that's fine, but could lead to some odd bugs in the future

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java
amcgee: Nice!  See other comment about `merge` logic but this is much easier to follow for me 🎉

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/BundledAppStorageService.java
amcgee: Sounds good.  I think `check` might be the wrong verb here, since it actually sets translations on the App object... so this means that as written translations only work for bundled apps...?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/BundledAppStorageService.java
```diff
@@ -104,17 +96,45 @@ public Map<String, App> discoverInstalledApps() {
...
+      return App.MAPPER.readValue(inputStream, App.class);
     } catch (IOException e) {
       log.error(e.getLocalizedMessage(), e);
     }
     return null;
   }
 
+  private List<AppManifestTranslation> readAppManifestTranslation(Resource resource) {
+    try {
+      InputStream inputStream = resource.getInputStream();
```
david-mackessy: Is this `inputStream` handled correctly & closed? (success & failure paths)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -109,7 +105,7 @@ private void discoverInstalledApps(Consumer<App> handler) {
 
       try {
         InputStream inputStream = manifest.getPayload().openStream();
-        App app = mapper.readValue(inputStream, App.class);
+        App app = App.MAPPER.readValue(inputStream, App.class);
```
david-mackessy, line 107: same `inputStream` comment here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/BundledAppStorageService.java
```diff
@@ -104,17 +96,45 @@ public Map<String, App> discoverInstalledApps() {
...
+    }
+  }
+
   private App readAppManifest(Resource resource) {
     try {
       InputStream inputStream = resource.getInputStream();
-      objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-      return objectMapper.readValue(inputStream, App.class);
+
+      return App.MAPPER.readValue(inputStream, App.class);
```
david-mackessy, line 117: same `inputStream` comment here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -228,11 +224,12 @@ public App installApp(File file, String filename, Cache<App> appCache) {
       InputStream inputStream = zip.getInputStream(entry);
 
       app = jsonMapper.readValue(inputStream, App.class);
```
david-mackessy: same `inputStream` comment here

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -308,6 +305,23 @@ public App installApp(File file, String filename, Cache<App> appCache) {
...
   }
 
+  private static void extractManifestTranslations(ZipFile zip, String prefix, App app) {
+    try {
+      ZipEntry translationFiles = zip.getEntry(prefix + MANIFEST_TRANSLATION_FILENAME);
+
+      List<AppManifestTranslation> appManifestTranslations =
+          App.MAPPER
+              .readerForListOf(AppManifestTranslation.class)
+              .readValue(zip.getInputStream(translationFiles));
```
david-mackessy: same `inputStream` comment here

PR #20249, opened 2025-03-14
fix: Add TEA change logs only when config enabled [DHIS2-19200]

muilpp:
In this PR:
- We now log changes to TEAs only when the `CHANGELOG_TRACKER` config is enabled. Previously, we considered whether the TET allowed auditing, but that should no longer affect the change log feature.

- While at it, I'm fixing a bug related to deleting an attribute value. In a [PR last week](https://github.com/dhis2/dhis2-core/pull/20196/commits/655350366b97705f10ea789be6d0401d35ffcda5#r1985364609), I retrieved tracked entity attributes from the tracked entity attribute values. This worked as long as the value wasn't deleted, otherwise, it wouldn't be returned. That's why I now fetch the attributes from the service.
    
- I'm removing `single_te.json`, as I need multiple TEs to test this issue. Instead, I'm using `base_data.json` for change logs as well. There, I'm creating a new TE of a TET that is configured to not allow audit logs.

Next
- Apply the same logic to event change logs
- Use the audit log configuration for auditing and breaking the glass


ameenhere (approved):

enricocolasante:
> * While at it, I'm fixing a bug related to deleting an attribute value. In a [PR last week](https://github.com/dhis2/dhis2-core/pull/20196/commits/655350366b97705f10ea789be6d0401d35ffcda5#r1985364609), I retrieved tracked entity attributes from the tracked entity attribute values. This worked as long as the value wasn't deleted, otherwise, it wouldn't be returned. That's why I now fetch the attributes from the service.

Did you add a test for this issue?

muilpp:
> > * While at it, I'm fixing a bug related to deleting an attribute value. In a [PR last week](https://github.com/dhis2/dhis2-core/pull/20196/commits/655350366b97705f10ea789be6d0401d35ffcda5#r1985364609), I retrieved tracked entity attributes from the tracked entity attribute values. This worked as long as the value wasn't deleted, otherwise, it wouldn't be returned. That's why I now fetch the attributes from the service.
> 
> Did you add a test for this issue?

`shouldLogChangesWhenTrackedEntityTypeAuditConfigDisabled` implicitly validates this behavior. 
The last interaction is a deletion, and we retrieve the values when requested. Would you prefer a more explicit test?

enricocolasante (approved):

PR #20199, opened 2025-03-07
feat: add code and shortName to trackedEntityType [DHIS2-19067]

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-19067
### Issue
- This issue is about the data model. The TrackedEntityType entity is already a NameableObject in the backend model. The `shortName` and `code` should be added and persisted.

### Fix
- Add `shortName` property. The `code` property has already been added so it just need added on UI.

### Test
- Fixed all the tests which import TrackedEntityType without `shortName`

PR #20275, opened 2025-03-17
fix: make trackedEntityType.shortname unique

vietnguyen:
Continue on [this PR](https://github.com/dhis2/dhis2-core/pull/20199), `TrackedEntityType.shortname` should be unique

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_42__make_shortname_in_trackedentitytype_unique.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_42__make_shortname_in_trackedentitytype_unique.sql
@@ -1,0 +1,2 @@
+alter table trackedentitytype drop constraint if exists uk_trackedentitytype_shortname;
+alter table trackedentitytype  add constraint uk_trackedentitytype_shortname unique (shortname);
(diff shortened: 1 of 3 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_42__make_shortname_in_trackedentitytype_unique.sql
```diff
@@ -0,0 +1,2 @@
+alter table trackedentitytype drop constraint if exists trackedentitytype_shortname_unique;
```
ameenhere, line 1: Maybe this is not a written convention yet, but we have tried to follow 

uk_tablename_columnname format in the past. So this key name should be

`uk_trackedentitytype_shortname`

larshelge (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.42/V2_42_42__make_shortname_in_trackedentitytype_unique.sql
vietnguyen: thanks, updated now.

stian-sandvold (approved):

PR #20278, opened 2025-03-17
feat: support route URL allow list [DHIS2-18861]

cjmamo:
* Support route URL allow list with `route.remote_servers_allowed` in dhis.conf

* Deprecate `system.remote_servers_allowed` dhis.conf setting and replace it with `metadata.sync.remote_servers_allowed`

### DIFF
--- a/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
+++ b/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
@@ -721,7 +721,7 @@ public enum ConfigurationKey {
    * Servers should be in a comma-separated style and always end with '/' for security reasons <br>
    * e.g. route.remote_servers_allowed = https://server1.com/,https://server2.com/
    */
-  ROUTE_REMOTE_SERVERS_ALLOWED("route.remote_servers_allowed", "https://*", false);
+  ROUTE_REMOTE_SERVERS_ALLOWED("route.remote_servers_allowed", "https://*", false),
 
   /** Enable OAuth2 authentication server. (default: off) */
   OAUTH2_SERVER_ENABLED("oauth2.server.enabled", Constants.OFF, false),


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -480,11 +488,9 @@ public enum ConfigurationKey {
   SERVER_BASE_URL("server.base.url", "", false),
 
   /**
-   * Remote servers allowed to call. <br>
-   * Default is empty. <br>
-   * Servers should be in a comma-separated style and always end with '/' for security reasons <br>
-   * e.g. system.remote_servers_allowed = https://server1.com/,https://server2.com/
+   * @deprecated use META_SYNC_SERVERS_ALLOWED instead
```
david-mackessy, line 491: should it say `META_DATA_SYNC_SERVERS_ALLOWED`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -154,6 +188,43 @@ public Route getRoute(@Nonnull String id) {
     return route;
   }
 
+  protected boolean isRouteUrlAllowed(Route route) {
+    if (route.getUrl().toLowerCase().startsWith("http:")
```
david-mackessy, line 192: Does this mean that if `allowedRouteRegexRemoteServers` is empty, any `https:` route is allowed?
A javadoc would help readers of this code understand what we intend to allow here.

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/DefaultDhisConfigurationProvider.java
```diff
@@ -221,23 +221,27 @@ public boolean isClusterEnabled() {
 
   @Override
   public String getServerBaseUrl() {
-    return StringUtils.trimToNull(
-        properties.getProperty(ConfigurationKey.SERVER_BASE_URL.getKey()));
+    return properties.getProperty(ConfigurationKey.SERVER_BASE_URL.getKey(), "").strip();
```
david-mackessy, line 224: Is there a specific reason to change this return value? (null if empty -> empty string if empty)
This value is used in numerous places around the system where a valid URL is attempted to be formed using this value.
I'm adding a `@CheckForNull` to this method in a current PR I'm working on.
If you are using this in the route service somewhere, checking for null & deciding what should happen if null is probably best.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
amcgee: I agree this is a bit confusing to read in the code as well, might be good to refactor it a bit for readability and/or add a comment as @david-mackessy suggested 👍

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -126,12 +136,43 @@ public class RouteService {
...
+  protected void validateHost(String host) {
+    if (!(host.startsWith("http://") || host.startsWith("https://"))) {
+      throw new IllegalStateException(
+          "Allowed route URL scheme must be either http or https: " + host);
+    }
+    String urlWithoutPortNo =
+        host.substring(0, host.lastIndexOf(":") > 5 ? host.lastIndexOf(":") : host.length());
+    URL url;
+    try {
+      url = new URL(urlWithoutPortNo);
```
amcgee, line 167: It looks like this works also with wildcard routes like `https://*.dhis2.org`, which is a little surprising to me but if that's the case then should be ok!

amcgee (approved):

david-mackessy (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -126,12 +136,43 @@ public class RouteService {
...
+        allowedRouteRegexRemoteServers.add(TextUtils.createRegexFromGlob(host));
+      }
+    }
+
     webClient = WebClient.builder().clientConnector(clientHttpConnector).build();
     dataBufferFactory = new DefaultDataBufferFactory();
   }
 
+  protected void validateHost(String host) {
+    if (!(host.startsWith("http://") || host.startsWith("https://"))) {
```
larshelge, line 159: A regex pattern match would be nice, but this works too.

jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -126,12 +136,43 @@ public class RouteService {
...
+    }
+    String urlWithoutPortNo =
+        host.substring(0, host.lastIndexOf(":") > 5 ? host.lastIndexOf(":") : host.length());
+    URL url;
+    try {
+      url = new URL(urlWithoutPortNo);
+    } catch (MalformedURLException e) {
+      throw new IllegalStateException(e);
+    }
+    if (!url.getPath().isEmpty()) {
```
larshelge, line 171: Use StringUtils.isNotEmpty to avoid nullpointer?

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/route/RouteServiceTest.java
```diff
@@ -0,0 +1,195 @@
...
+  //    Properties properties = new Properties();
+  //    properties.setProperty(
+  //            ConfigurationKey.ROUTE_REMOTE_SERVERS_ALLOWED.getKey(), protocolUnderTest + "://*");
+  //    DhisConfigurationProvider dhisConfigurationProvider =
+  //            new TestDhisConfigurationProvider(properties);
+  //
+  //    RouteService routeService = new RouteService(null, null, dhisConfigurationProvider, null,
+  // null);
+  //    assertThrows(IllegalStateException.class, routeService::postConstruct);
+  //  }
```
larshelge, line 62: Remove commented code.

PR #20282, opened 2025-03-17
feat: [DHIS2-17541] Add ability to use read replica for tracker read queries

ameenhere:
In this PR we add the ability to configure read only replica details to be used when fetching Events or Tracked entities.

To make this change least disruptive/safe, the feature itself is controlled by configuration in dhis.conf.

1. To enable tracker stores to connect with read replica, a new config named `tracker.read_replica.enabled` has to be set to `on` or `true` (like other switches in the config).
2. If enabled, the server expects the config `tracker.read_replica.connection.url` to be provided and pointing to the jdbc url of the read repica. If this is missing in the config (and if 1. is enabled) then we fail fast at startup with a proper error message.
3. The  read only data source will re-use the same username, password as db user accounts are also replicated in read replica (useful for hot standby). The connection pool configs of the main database will be re-used for the read only connection pool.
4. Currently read replica will only be used when fetching Events (in JdbcEventStore)
          
    
If 1. is OFF then the actualDataSource will be used instead of the readOnlyDataSource


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -224,13 +226,34 @@ class JdbcEventStore {
...
   private final NamedParameterJdbcTemplate jdbcTemplate;
 
-  @Qualifier("dataValueJsonMapper")
   private final ObjectMapper jsonMapper;
 
   private final UserService userService;
 
   private final IdentifiableObjectManager manager;
 
+  @Autowired
```
enricocolasante, line 235: Should we remove lombok annotation to create a constructor?

enricocolasante (commented):
Is there a way to create a test in `dhis-test-web-api` module and use a configuration provider with the replica fields? We can use the same as the DB that we are using to test, the important part is to validate the whole logic.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
teleivo: We only use `@Autowired` in tests. Prod code uses constructors which are then used by Springs DI

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
```diff
@@ -65,28 +65,76 @@ public class DataSourceConfig {
...
+    try {
+      return DatabasePoolUtils.createDbPool(poolConfig);
+    } catch (SQLException | PropertyVetoException e) {
+      String message =
+          String.format(
+              "Connection test failed for tracker read replica database pool, jdbcUrl: '%s', user: '%s'",
+              jdbcUrl, username);
+
+      log.error(message);
+      log.error(DebugUtils.getStackTrace(e));
```
teleivo, line 134: Some general things on logging

1. We should handle an error once. That means either log and move on or throw.
2. The logger can log exceptions so do not log twice but use the appropriate method.
3. Logs should be actionable. Think of the admin that will then have to deal with multiple log entries that are all pointing to the same problem. Is that helpful?

For this particular case: I assume you want to prevent DHIS2 from starting and give a meaningful message so an admin can fix the config right? Then rethrow with a refined message.

Right now this leads to at least 2 log entries for your two `log.error` calls + likely another one higher up the stack for the exception you rethrow.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
```diff
@@ -65,28 +65,76 @@ public class DataSourceConfig {
...
+  }
 
-    return jdbcTemplate;
+  @Bean("readReplicaDataSource")
+  public DataSource readReplicaDataSource(
+      DhisConfigurationProvider config,
+      @Qualifier("actualDataSource") DataSource actualDataSource) {
+
+    if (!config.isTrackerReadReplicaEnabled()) {
+      log.info("Tracker Read replica is disabled");
```
teleivo, line 102: This to me is a debug log if any. What value does this provide to whom? We know its enabled based on the `dhis.conf`.

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -710,6 +713,9 @@ public enum ConfigurationKey {
   /** Enable TOTP-based 2FA authentication. (default: true) */
   TOTP_2FA_ENABLED("login.security.totp_2fa.enabled", Constants.ON, false),
 
+  /** Use read replica database for tracker queries. (default: false) */
+  TRACKER_READ_REPLICA_ENABLED("tracker.read_replica.enabled", Constants.OFF, false),
```
teleivo, line 717: What is the reason for two config keys? With two you end up having to deal with the case that its enabled but there is no url set. Users can toggle the usage of this feature by commenting the `tracker.read_replica.connection.url`. Or do you foresee some more configs related to this?

If we keep it group the configs. Why is this one down here.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -224,13 +226,34 @@ class JdbcEventStore {
...
+      NamedParameterJdbcTemplate namedParameterJdbcTemplate,
+      @Qualifier("readOnlyNamedParameterJdbcTemplate")
+          NamedParameterJdbcTemplate readOnlyNamedParameterJdbcTemplate,
+      @Qualifier("dataValueJsonMapper") ObjectMapper jsonMapper,
+      UserService userService,
+      IdentifiableObjectManager manager,
+      DhisConfigurationProvider configurationProvider) {
+
+    if (configurationProvider.isTrackerReadReplicaEnabled()) {
+      log.info("Tracker Event read queries directed to read replica database");
```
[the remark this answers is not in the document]
ameenhere: I initially had implemented a more granular configuration to redirect read queries to events, tracked entities (and for future enrollments and relationships) separately. Meaning implementations could toggle one or more of them as required.  Had to remove trackedEntity one as it requires more refactoring that is not possible to finish today. 

This log was meant to verify that the server has started up and configured the context (in this case,  this JdbcEventStore) as expected.
We have similar ones when for various other setup, where the startup logs indicate how the server actually started based on the config it read. 

For me, it provided value when I read the startup logs to verify the server started up with a Spring context that I expect.

dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
ameenhere: Good point. Yes initially there were more granular configuration related to this (mentioned in previous comment above) . But removed the others due to lack of time. But in general I still want to keep the main "enabling" config as the main toggler of the feature. I feel this makes it very explicit and the intentions are clear by a system admin. 

I'll group it together and I also foresee more configs coming for this (if we decide to proceed this route)

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
ameenhere: I really like logging these kind of Spring context setup it did based on the dhis config "it read".  dhis.conf could be read from multiple places and sometimes admins accidentally update a dhis.conf in the wrong place (or have spelling mistakes that they did not realize) and very late find out their mistake (sometimes too late). Since this is only logged on startup and it is useful to verify the server started up with a Spring context that  the admin expect.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
```diff
@@ -65,28 +65,76 @@ public class DataSourceConfig {
...
+      return actualDataSource;
+    }
+    String jdbcUrl = config.getProperty(ConfigurationKey.TRACKER_READ_REPLICA_CONNECTION_URL);
+    String username = config.getProperty(ConfigurationKey.CONNECTION_USERNAME);
+
+    if (!StringUtils.hasText(jdbcUrl)) {
+      String errorMessage =
+          "Tracker Read replica is enabled but read replica connection url not specified in dhis.conf.";
+      log.error(errorMessage);
```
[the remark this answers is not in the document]
ameenhere: Yep! I'll remove this. Copy Pasta mistake from something similar in the method below (above)

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/config/DataSourceConfig.java
teleivo: > dhis.conf could be read from multiple places and sometimes admins accidentally update a dhis.conf in the wrong place (or have spelling mistakes that they did not realize) and very late find out their mistake (sometimes too late).

We don't support merging configs, do we?
Logging which config is applied would be enough to satisfy this and reduce the amount of logs/code we have.

> (or have spelling mistakes that they did not realize) and very late find out their mistake (sometimes too late)
Mhm, I remember issues with the cloud store setup. There we assumed/defaulted to horrific defaults 🤦🏻 We changed it to fail startup with a message that tells users how to fix the issue. That to me is often the best action to take instead of logs.

Spelling mistakes for config keys is something I think should also be fixed in one go. Validate and prevent DHIS2 from starting with a proper error message.

ameenhere: >  We don't support merging configs, do we? 

Nope we don't. But we support several places in which config can be picked up from right?. My point was, an admin might mistakenly update a file named "dhis.conf" somewhere which may not be the one the server is reading. To get the confidence that the changes I made has been picked up correctly, reading startup logs would be good. 

Sometimes, there is a spelling mistake for a config (and since we have defaults), we would just use defaults and carry on, but the admin expects their value to be taken. Some scenarios could not be failed fast since defaults may be sensible enough. 

Anyway I have removed these logs and left the one log in DataSourceConfig that logs is Tracker Replica is enabled/disabled. 

Also this PR will not be merged today and will continue to work on this a bit here and there.

PR #20298, opened 2025-03-18
chore: Update license header to exactly match BSD-3 

larshelge:
This PR updates the source code license header to exactly match the text in BSD-3.

https://opensource.org/license/bsd-3-clause


david-mackessy (approved):
the E2E module is not part of the base pom so the spotless format will have to be run in that directory also.

larshelge:
Good point @david-mackessy .

PR #20300, opened 2025-03-18
fix: Use different TE validation flow when breaking the glass [DHIS2-18877]

muilpp:
### Issue
When breaking the glass, we can't use the regular flow and call `trackedEntityService.getTrackedEntity()` , because it validates whether the user has access to the owner of the TE/program couple.
That means, if the program is protected, it will check the user has capture scope access to that owner, and it in this case it won't, as that's the whole point of breaking the glass.

### Fix
For this reason, we will get the TE from the manager and run a separate set of validations just for the case when the user breaks the glass. Some of them were already in place, the ones I added are:
- TE not null
- Data access to program
- Data access to TET
- program TET is equal to TE TET

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
@@ -226,7 +226,7 @@ public class DefaultTrackerOwnershipManager implements TrackerOwnershipManager {
         program.getTrackedEntityType().getUid(), trackedEntity.getTrackedEntityType().getUid())) {
       throw new ForbiddenException(
           String.format(
-              "Temporary ownership not created. The the tracked entity type of the program %s differs from that of the tracked entity %s.",
+              "Temporary ownership not created. The tracked entity type of the program %s differs from that of the tracked entity %s.",
               program.getTrackedEntityType().getUid(),
               trackedEntity.getTrackedEntityType().getUid()));
     }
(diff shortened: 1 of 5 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -194,13 +201,34 @@ private void validateGrantTemporaryOwnershipInputs(
...
+      throw new ForbiddenException(
+          "Temporary ownership not created. User has no data read access to tracked entity type: "
+              + trackedEntity.getTrackedEntityType().getUid());
+    }
+
+    if (!Objects.equals(
+        program.getTrackedEntityType().getUid(), trackedEntity.getTrackedEntityType().getUid())) {
+      throw new ForbiddenException(
+          String.format(
+              "Temporary ownership not created. The the tracked entity type of the program %s differs from that of the tracked entity %s.",
```
enricocolasante, line 229:
```suggestion
              "Temporary ownership not created. The tracked entity type of the program %s differs from that of the tracked entity %s.",
```

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/ownership/TrackerOwnershipManagerTest.java
```diff
@@ -421,6 +424,78 @@ void shouldFailWhenGrantingTemporaryAccessIfProgramIsNotTrackerProgram() {
...
+
+    Exception exception =
+        Assertions.assertThrows(
+            ForbiddenException.class,
+            () ->
+                trackerOwnershipAccessManager.grantTemporaryOwnership(
+                    trackedEntityA1, programA, userDetailsA, "test temporary ownership"));
+
+    assertStartsWith(
+        "Temporary ownership not created. The the tracked entity type of the program",
```
enricocolasante, line 495:
```suggestion
        "Temporary ownership not created. The tracked entity type of the program",
```

enricocolasante (approved):

PR #20304, opened 2025-03-18
fix: correct context path handling for global shell

amcgee:
Fix an issue when deployed to non-root context paths.

If you have previously experienced an issue with your instance you may need to clear the cache storage in your browser (Dev Tools -> Application tab -> Cache Storage -> "Delete" each one) and unregister all service workers.

For testing incognito mode also works


Philip-Larsen-Donnelly (approved):
Approved based on functional preview.

kabaros (approved):

PR #20322, opened 2025-03-19
fix: Fix missing keys in parentGraph of EventVisualization [DHIS-19249]

luciano-fiandesio:
## Summary

Fixes an issue with the `parentGraph` population logic: in certain cases the value of a key in the map would be over-ridden with empty.


HendrikThePendric:
I will refrain from commenting on the code, but I have created an instance from this branch that shows the problem is gone: 
1. Visit this saved visualization: https://dev.im.dhis2.org/dhis2-19249-parent-graph-fix/apps/line-listing#/hoDzRijsi91
2. Click the `E2E - Organisation unit` chip
3. You now see a modal open where all 4 (nested) org-units are selected correctly

larshelge (approved):

maikelarabori (approved):

vietnguyen (approved):

PR #20324, opened 2025-03-19
fix: duplicate authorities in the authority endpoint

netroms:
Fix duplicate authorities in API response (DHIS2-18741)
Issue
The /api/authorities endpoint was returning duplicate authorities in the response. Specifically, the authority F_MOBILE_SENDSMS appeared twice with the same ID and name. This was caused by the getAuthorities() method in the AuthoritiesController using an ArrayList to collect authorities from multiple sources without deduplication.
Fix
Modified the getAuthorities() method in AuthoritiesController.java to use a HashSet instead of an ArrayList when collecting authorities from different sources. This ensures that duplicate authorities are automatically removed due to the set's unique element constraint.
Apply to d17f54d9f26f...
authorities
Test
Added a new test method testNoDuplicateAuthorities() to AuthoritiesControllerTest.java that verifies:
No duplicate authorities exist in the response by comparing the size of the original list with a deduplicated set
The F_MOBILE_SENDSMS authority that was previously duplicated now appears exactly once
Result
The authorities endpoint now returns a clean list of unique authorities, maintaining the original functionality while removing duplicates. All tests pass successfully.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AuthoritiesControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AuthoritiesControllerTest.java
@@ -29,9 +29,12 @@
  */
 package org.hisp.dhis.webapi.controller;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


jbee (approved):

david-mackessy (approved):
nice one

PR #20325, opened 2025-03-19
fix: correctly set contentLength for remote resources, whitelist access to global shell and user profile apps [DHIS2-19244]

amcgee:
This solves a few problems related to serving either (1) apps for which the user does not have view authority, or (2) installed apps when using remote object storage backends.

The issue was three-fold:
1. First, when using an object storage backend like MinIO or S3 for installed apps the Resource returned by AppManager is remote (a Url) so it was always setting the contentLength header to 0.  Instead, we now use a Head request to the storage backend to retrieve the real content length before setting the header.
2. Second, when a user didn't have access to an app they would "fall through" the GlobalShellFilter which uses a request dispatcher to forward to the AppController.  When this happens, we now return a 404.  This wasn't caught earlier because it can be masked if there is an existing cached app from a different user.
3. Users didn't by default have access to the global shell - they needed a specific authority.  Since the global shell is intended to be global, I've set an allow list of apps which do not require any special authority to view.  These are Login, Global Shell, and User Profile.  I've also excluded Login and Global Shell from the app menu.  Now, a user with no authorities whatsoever will still see the global shell and will see the User Profile app in the menu, but won't be able to navigate to any other apps (either in the shell or manually - the requests will return 404)

A few sideffects of this fix:
1. The login app is no longer shown in the apps menu even to users with the ALL authority.  This was always a bit confusing to people, and testing of login styling changes can always be done in an incognito window or simply by navigating to `dhis-web-login` while logged in.
2. The hard-coded authorization checks for bundled apps in [DhisWebApiWebSecurityConfig.java](https://github.com/dhis2/dhis2-core/pull/20325/files#diff-5da926f04078b11a15144b0faf8302b4a12b846e1e062a6e80c2d47387b70506) have been removed because this is now enforced in the same way as for any other app in the AppController / App Manager.  This also means that instead of getting a 403 top-level browser error when trying to navigate to an app they can't access, the user will instead be redirected to the global shell which tells them that the app cannot be found.  There is a slight change in behavior here because we return 404 (not found) instead of 403 (forbidden) when requesting resources from these apps.  This is a slight security enhancement because it prevents unauthorized users from snooping to determine which apps are installed in the system, but we can change it to 403 if we want the old behavior back.

Bonus:
* I also added translation keys for the newly-bundled Line Listing app's authority


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
```diff
@@ -164,66 +163,85 @@ public void reloadApps() {
...
+      log.info("App not found: {}", appName);
       throw new WebMessageException(notFound("App '" + appName + "' not found."));
     }
 
     if (!appManager.isAccessible(application)) {
-      throw new ForbiddenException("You don't have access to application " + appName + ".");
+      log.info("User does not have access to app: {}", appName);
+      throw new WebMessageException(
+          forbidden("User does not have access to app '" + appName + "'."));
     }
```
amcgee, line 180: getApp above already checks `isAccessible`, so this code path will technically always be skipped and a 403 never returned.  We get a 404 if the user doesn't have access to the app.  I left this here for now in case we want to move it up and go back to returning a 403.

Note that this was an existing bug for a long time, so for *installed apps* we have never returned 403.  Bundled apps were handled with hard-coded authority checks based on their path, so they did return top-level 403 errors.  Since we've harmonized how bundled and installed apps are served we need to pick one behavior - I've kept the 404 response for now to avoid disclosing the presence of apps to unauthorized users.

ameenhere (approved):

david-mackessy (approved):
Content length of 0 for object storage sounds like a regression from recent changes

PR #20342, opened 2025-03-20
fix: correctly set caching etag and lastModified headers when serving app resources [DHIS2-19245]

amcgee:
This solves an issue with caching headers used to determine if a 304 Not Modified response should be returned by the server (or by an intermediate cache)

Prior to this change, only the lastModified header was set.  For bundled apps the lastModified date of the file was indeterminate so this was always set to Jan 1, 1970.  

This was the behavior before 2.42 as well for installed apps.  Bundled apps were served natively by spring and correctly set some of these headers.

Before this change (missing headers - left: 2.41, right: before)

Screenshot 2025-03-18 at 22 29 53
Screenshot 2025-03-18 at 22 29 02
Screenshot 2025-03-18 at 22 15 09
Screenshot 2025-03-18 at 22 13 31

After this change (left: 2.41, right: after)
Screenshot 2025-03-20 at 12 19 58

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
@@ -37,19 +37,15 @@ import static org.hisp.dhis.dxf2.webmessage.WebMessageUtils.notFound;
 import static org.hisp.dhis.security.Authorities.M_DHIS_WEB_APP_MANAGEMENT;
 
 import com.google.common.collect.Lists;
-
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
-
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
-
 import lombok.extern.slf4j.Slf4j;
-
 import org.hisp.dhis.appmanager.App;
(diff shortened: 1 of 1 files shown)


ameenhere (approved):

KaiVandivier (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/AppController.java
```diff
@@ -168,57 +170,77 @@ public void renderApp(
...
+    // need to include the last modified timestamp
+    // Similarly, with classPath resources the lastModified timestamp may be missing or not
+    // reliable, so we need to include the version number
+    // See also AppHtmlNoCacheFilter for cache control headers set on index.html responses
+
+    long lastModified = resourceResult.resource().lastModified();
+    String etagSource = String.format("%s-%s", app.getVersion(), String.valueOf(lastModified));
+    String etag = HashUtils.hashMD5(etagSource.getBytes());
+
+    log.info("Generated etag {} from source {}, lastModified {}", etag, etagSource);
```
david-mackessy, line 241: If these are infrequent code paths, then log level `info` seems ok, but if this might spam the logs maybe `debug` would be better. Same for 2 lines down.

amcgee: Ah yes should be `debug` good spot @david-mackessy

amcgee:
Opened #20344, we shouldn't spam info logs on every http request

PR #20357, opened 2025-03-21
fix: Fetch org unit path when getting TE [DHIS2-19282]

muilpp:
Get the TE org unit path from the DB, so we can use it when checking the access to a TE/program owner.
Currently we don't set the path value, so access to a TE is always rejected.
Note that this only happens when there's no owner to a TE/program pair and we default to the registering org unit.

### DIFF
--- a/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
+++ b/dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
@@ -602,6 +602,34 @@ class TrackedEntityServiceTest extends PostgresIntegrationTestBase {
   }
 
   @Test
+  void shouldReturnTrackedEntitiesWithCorrectOrgUnitPropertiesMappedWhenTrackedEntityAccessible()
+      throws ForbiddenException, NotFoundException, BadRequestException {
+    TrackedEntityOperationParams operationParams =
+        TrackedEntityOperationParams.builder()
+            .organisationUnits(orgUnitChildA)
+            .orgUnitMode(SELECTED)
+            .trackedEntityType(trackedEntityTypeA)
+            .build();
+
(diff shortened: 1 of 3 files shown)


enricocolasante (approved):

ameenhere:
Are we sure that, if a programOwner is present, the path of that ownerOU is populated correctly?

ameenhere (approved):

muilpp:
> Are we sure that, if a programOwner is present, the path of that ownerOU is populated correctly?

Yes, when getting the owner we use Hibernate, and it populates the org unit path correctly.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityServiceTest.java
```diff
@@ -601,6 +601,34 @@ void shouldReturnTrackedEntitiesGivenUserHasDataReadAccessToTrackedEntityType()
         trackedEntities);
   }
 
+  @Test
+  void shouldReturnTrackedEntitiesWithCorrectOrgUnitPropertiesMappedWhenTrackedEntityAccessible()
+      throws ForbiddenException, NotFoundException, BadRequestException {
+    TrackedEntityOperationParams operationParams =
```
teleivo, line 607: Can we test

> Currently we don't set the path value, so access to a TE is always rejected.

AFAIK this test only tests the mapping which is not all that useful IMHO. This test also runs as the import user which I am not sure is a super user 🤔

muilpp: The user running this test is not a super user.

Here I'm testing the TE that's returned from the API call, it's not just a mapping, this is the data we store in the database.
If the TE is returned, it already validates it's no longer rejected.
On top of that, I'm making sure the org unit path is set, which was the reason why it was rejected before.

PR #20360, opened 2025-03-21
fix: Remove unnecessary TET and program validations [DHIS2-19286]

muilpp:
1) There were two validations in `canReadProgramAndTrackedEntityType`
    - Data access to the program
    - Data access to the TE TET

    These validations are run only when a program is specified in the request. In this case though, they are already validated [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java#L125). This method doesn't explicitly validate the TE TET, but it validates the program TET, which is effectively the same, as a TE can't be enrolled in a program with a different TET.

2) We were missing a call to `trackerAccessManager.canRead` in `DefaultTrackedEntityService#findTrackedEntities(TrackedEntityOperationParams)`, as `DefaultTrackedEntityService#findTrackedEntities(TrackedEntityOperationParams,PageParams)` already does. However, this validation could be refactored and run before we reach the database, at least the TET part. I'll create a ticket to tackle that. This would help when moving the ownership validation into the query as well.

3) I've created a new method in `TrackerAccessManager`, that allows us to skip the ownership validations at will, by using a flag. We need it, because when the org unit mode is `ALL`, we don't have to validate the user ownership, but we do need to validate the data access to the TET and program. This should be a temporary solution, as we need to push this into the query in order not to break pagination.

4) I'm removing `initializeTrackedEntityOrgUnitParents`. We used it to set the path of the owner org unit based on its parents. We no longer need it, because after this fix: https://github.com/dhis2/dhis2-core/pull/20357, now we fetch the path from the database directly.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -77,16 +77,35 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
    */
   @Override
   public List<String> canRead(@Nonnull UserDetails user, TrackedEntity trackedEntity) {
-    // always allow if user == null (internal process) or user is superuser
     if (user.isSuper() || trackedEntity == null) {
```
teleivo, line 80: This block is not needed as its part of the `canRead` you created and call later

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -77,16 +77,35 @@ public class DefaultTrackerAccessManager implements TrackerAccessManager {
...
+  /**
+   * Check the data read permissions and ownership of a tracked entity given the programs for which
+   * the user has metadata access to. Ownership validations will be skipped if the flag is set to
+   * true.
+   *
+   * @return No errors if a user has access to at least one program
+   */
+  @Override
+  public List<String> canRead(
+      @Nonnull UserDetails user, TrackedEntity trackedEntity, boolean skipOwnershipCheck) {
```
teleivo, line 96: These `skipOwnershipCheck` are a bit scary to me 😅 I would prefer if `canRead` contains the logic of "user x is an admin so I don't need to check ownership, ..."

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -108,26 +127,22 @@ private List<String> canRead(
                     p.isRegistration()
                         && Objects.equals(
                             p.getTrackedEntityType().getUid(),
-                            trackedEntity.getTrackedEntityType().getUid()))
+                            trackedEntity.getTrackedEntityType().getUid())
+                        && aclService.canDataRead(user, p))
```
teleivo, line 131: Will `programService.getAllPrograms()` return all programs the user has `metadata` read access to?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -204,36 +219,12 @@ private boolean canWrite(UserDetails user, TrackedEntity trackedEntity, Program
...
-    TrackedEntityType trackedEntityType = trackedEntity.getTrackedEntityType();
-
-    if (!aclService.canDataRead(user, trackedEntityType)) {
-      errors.add(
-          "User has no data read access to tracked entity type: " + trackedEntityType.getUid());
+      return List.of(OWNERSHIP_ACCESS_DENIED);
     }
 
-    return errors;
+    return emptyList();
```
teleivo, line 227:
```suggestion
    return List.of();
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/TrackerAccessManager.java
```diff
@@ -47,6 +47,8 @@
 public interface TrackerAccessManager {
   List<String> canRead(UserDetails user, TrackedEntity trackedEntity);
 
+  List<String> canRead(UserDetails user, TrackedEntity trackedEntity, boolean skipOwnershipCheck);
```
teleivo, line 50: When would I use this one? You said some method was temporary, is it this one? if so mark it deprecated for removal (add the version or issue in which it will be removed)

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -284,13 +284,15 @@ public Set<String> getOrderableFields() {
 
   private Predicate<TrackedEntity> getFilter(
       UserDetails user, TrackedEntityQueryParams queryParams) {
+    boolean skipOwnershipCheck = queryParams.getOrgUnitMode() == ALL;
```
teleivo, line 287: Users need a special authority to use the mode `ALL` right?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -274,7 +272,9 @@ private List<TrackedEntity> findTrackedEntities(
       }
     }
     trackedEntityAuditService.addTrackedEntityAudit(SEARCH, user.getUsername(), trackedEntities);
-    return trackedEntities;
+
+    // TODO(tracker): Push this filter into the store because it is breaking pagination
+    return trackedEntities.stream().filter(getFilter(user, queryParams)).toList();
```
teleivo, line 277: Can you find a more descriptive name for what we filter here than `getFilter` 😅

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
muilpp: I know, but I'd still leave it here.
The other method is just a temporary fix, and if remove this guard clause, it will be easy to forget to add it back when I remove the temporary method.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
muilpp: yes

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
muilpp: yes, but that's already validated in `OperationsParamsValidator`, before we hit the database

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
muilpp: I don't like it either, but the logic to decide if we can skip the check or not, is to validate the org unit mode is `ALL`. That's fine when using the exporter.
However, the importer doesn't use org unit modes, that's why I went for this solution.
Again, this is temporary and as soon as I'm done with the tickets that need to be merged before the release, I will remove this validation and add it into the query.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -237,7 +237,14 @@ private void validateGrantTemporaryOwnershipInputs(
...
   @Transactional(readOnly = true)
   public boolean hasAccess(UserDetails user, TrackedEntity trackedEntity, Program program) {
-    if (canSkipOwnershipCheck(user, program) || trackedEntity == null) {
+    return hasAccess(user, trackedEntity, program, false);
+  }
+
+  @Override
+  @Transactional(readOnly = true)
+  public boolean hasAccess(
+      UserDetails user, TrackedEntity trackedEntity, Program program, boolean skipOwnershipCheck) {
```
enricocolasante, line 246: I would still remove this `skipOwnershipCheck` and have the caller decide if it needs to call this method or not based on the flag.

enricocolasante (approved):

ameenhere (approved):
> I'm removing initializeTrackedEntityOrgUnitParents. We used it to set the path of the owner org unit based on its parents. We no longer need it, because after this fix: https://github.com/dhis2/dhis2-core/pull/20357, now we fetch the path from the database directly.

Just make sure we do not have "other code flows" that might depend on this method.. (For example during Import).

muilpp:
> > I'm removing initializeTrackedEntityOrgUnitParents. We used it to set the path of the owner org unit based on its parents. We no longer need it, because after this fix: #20357, now we fetch the path from the database directly.
> 
> Just make sure we do not have "other code flows" that might depend on this method.. (For example during Import).

I've check this and the importer, exporter (for both TE and relationships) and audit controller have the owner org unit path set after the fix, they don't require that method anymore.

PR #20375, opened 2025-03-24
fix: Remove program rules cache [DHIS2-18848]

enricocolasante:
Program rules used by tracker importer are always retrieved by program.
We had a cache in place to check if a program had no program rules defined. If there were program rules, we still needed to retrieve them.
The issue was that this cache was not invalidated correctly, hence when creating the first program rule for a program, the user needed to manually clean the cache in order to have tracker importer using the created program rule.

***This PR***
We remove the cache and add an index on `programid` column on `programrule` and `programrulevariable` tables.
Those tables are always accessed by filtering by one program when requesting data from tracker importer.
Now we do not have the cache anymore, when a program has no program rules we are performing a call to the DB, but the query is using the index to check the existence of any program rule.

While fixing the cache issue, I simplified the code, removing all the unnecessary filtering on program stage that is already done in `rule-engine` library itself.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/engine/DefaultProgramRuleEngine.java
```diff
@@ -148,12 +122,44 @@ public RuleValidationResult getDataExpressionDescription(String dataExpression,
...
+  }
+
+  @Override
+  public RuleEngineEffects evaluateProgramEvents(
+      @Nonnull List<RuleEvent> events, @Nonnull Program program, @Nonnull UserDetails user) {
+    return evaluateEnrollmentAndEvents(null, events, program, user);
+  }
+
+  private RuleEngineEffects evaluateEnrollmentAndEvents(
+      RuleEnrollment enrollment,
```
teleivo, line 141: `@CheckForNull` is the counterpart to `@Nonnull`. I don't think we need to add these to all methods public ones should be enough 🤔 the analyzer should then be able to analyze the different code paths of private methods as well.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_2__create_index_program_rule.sql
```diff
@@ -0,0 +1,5 @@
+drop index if exists "program_rule_program";
+drop index if exists "program_rule_variable_program";
+
+CREATE INDEX program_rule_program ON programrule USING btree (programid);
```
teleivo, line 4: Use one casing for the SQL, I would stick to lowercase.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/engine/DefaultProgramRuleEngine.java
enricocolasante: To be honest, here I thought about it, but I was not sure whether to add it.
Maybe it is just the naming that is confusing me, but I am not going to check the null in the method, I am just passing the value through, no matter if it is null or not. Does it still make sense to annotate with @CheckForNull?
I mean, if the annotation was called @Nullable, I wouldn't have any doubt.
I am going to add it anyway, the analyzer is going to be happy and this is what matters.

PR #20390, opened 2025-03-25
Fix hibernate caching for option sets

jason-p-pickering:


jbee:
FYI: This is an ongoing experiment.

PR #20391, opened 2025-03-25
refactor: Rename methods for stats and type report

david-mackessy:
Rename the increment and decrement methods in `Stats` & `TypeReport` for:
- created
- updated
- deleted
- ignored

The increment method by amount _n_ is required for collection sizes, so all have been kept looking consistent.


jbee (approved):
:+1:

netroms (approved):

PR #20396, opened 2025-03-25
perf: use SQL based option is valid checks [DHIS2-18135]

jbee:


netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionService.java
```diff
@@ -62,7 +65,13 @@ public interface OptionService {
...
   List<OptionSet> getAllOptionSets();
 
-  List<Option> getOptions(String optionSetId, String name, Integer max);
+  List<Option> findOptionsByNamePattern(
+      @Nonnull String optionSet, @CheckForNull String infix, @CheckForNull Integer maxResults);
+
+  boolean existsAllOptions(@Nonnull String optionSet, @Nonnull Collection<String> codes);
+
+  @CheckForNull
+  Option getOptionByCode(@Nonnull String optionSet, @Nonnull String code);
```
enricocolasante, line 74: I think we agreed somewhere to have `find` methods returning an `Optional` or `get` method throwing an exception.
Would it makes sense here to change it to `find` and return an `Optional`?

enricocolasante (approved):

PR #20399, opened 2025-03-25
feat: docker doris

tonsV2:
### Launch

DHIS2 and Doris can be launched using the below command

    DHIS2_IMAGE=dhis2/core-dev:42.0.0-rc DORIS_VERSION=3.0.4 docker compose -f docker-compose.yml -f docker-compose.doris.yml up

### Issues
* FE_SERVERS doesn't accept anything but an ip when referencing the server. When using Docker Compose it's ideally a string which is used to reference services. To workaround this issue I define a IPAM network and assign all services a static ip
* ~~For the sake of initializing Doris the service named doris-init is implemented. Its sole purpose is to create a database called dhis2 on Doris... But it doesn't work! I'm not sure why.~~


README.md
```diff
@@ -1,156 +1,171 @@
...
+docker compose up
+```
+
+### Launch with Doris
+
+```sh
+DHIS2_IMAGE=dhis2/core-dev:latest DORIS_VERSION=3.0.4 docker compose -f docker-compose.yml -f docker-compose.doris.yml up
+```
+
+#### Generate tables (using default DHIS2 password)
```
Philip-Larsen-Donnelly, line 85: Not sure this is necessary, as this is not specific to Doris

tonsV2: I'll remove it

tonsV2:
> What about putting the docker-compose.doris.yml file somewhere other than root (i.e. establish a good practice for these merge files)?

I were considering placing it in /docker but I didn't do it for two reasons
1. Since we only have one it's not polluting the root folder
2. Having it next to the main docker compose file is also a way of advertising Doris as a feature we support

Philip-Larsen-Donnelly (approved):

docker-compose.doris.yml
```diff
@@ -0,0 +1,75 @@
...
+      doris-fe:
+        condition: service_healthy
+    volumes:
+      - ./docker/doris/postgresql.jar:/opt/apache-doris/be/jdbc_drivers/postgresql.jar:ro
+      - ./docker/doris/postgresql-42.7.5.jar:/opt/apache-doris/be/jdbc_drivers/postgresql-42.7.5.jar:ro
+      - ./docker/doris/create-doris-database.sql:/docker-entrypoint-initdb.d/create-doris-database.sql:ro
+      - doris-be-storage:/opt/apache-doris/be/storage
+      - doris-be-log:/opt/apache-doris/be/log
+    healthcheck:
+      test: ["CMD", "curl", "-fsS", "http://localhost:8040/api/health"]
```
radnov, line 60: Might as well use `|| exit 1` for this healtcheck.
```suggestion
      test: ["CMD-SHELL", "curl -fsS http://localhost:8040/api/health || exit 1"]
```

docker-compose.doris.yml
```diff
@@ -0,0 +1,75 @@
...
+        ipv4_address: 192.168.100.10
+    ports:
+      - "9030:9030"
+      - "8030:8030"
+    environment:
+      - FE_SERVERS=fe1:192.168.100.10:9010
+      - FE_ID=1
+    volumes:
+      - ./docker/doris/postgresql.jar:/opt/apache-doris/fe/jdbc_drivers/postgresql.jar:ro
+      - ./docker/doris/postgresql-42.7.5.jar:/opt/apache-doris/fe/jdbc_drivers/postgresql-42.7.5.jar:ro
```
radnov, line 30: Do both files need to be mounted, given that one is symlink? 🤔

radnov (approved):
LGTM, just two comments 👍

docker-compose.doris.yml
tonsV2: Wouldn't the file be missing if we just mount the symlink? And vice versa?

tonsV2: But maybe a better solution would be to mount the file without the version :thinking:

PR #20423, opened 2025-03-26
fix: correctly fall back to installed app display name if no bundled translation found (42.0)

amcgee:
Backport of #20394 
DHIS2-19319 approved by RCB

(cherry picked from commit d14db5e2c6aaa8209136c0cfb6521007bd9af57e)


kabaros (approved):

amcgee:
> Would it be hard to add a test for this?
@cjmamo probably not, and it would be great to have one.  Do you have some time to look into this today?  If not, we could merge this as-is and add the test on master

amcgee:
@cjmamo let's add tests for this on master, we need to get it into patch/2.42.0 today

PR #20437, opened 2025-03-27
fix: validation of filter parser DHIS2-19331

teleivo:
This change is to the parsing of the tracker `filter` request parameter `filter=qrur9Dvnyt5:gt:10:lt:20:!null` used on `/tracker/trackedEntities?filter=` and `/tracker/events?filter=...&filterAttributes=...`.

The previous implementation relied on a mix of string replaces (due to escaped separator characters like `/:`, `/,`), regular expression matching and string splitting into individual filters and segments (uid, operator and value). The previous implementation also used multiple passes over the input string. The addition of unary operators `null`, `!null` made the implementation more complex. I also found a bug in the validation of unary operators having a value. I believe that bug is due to the complexity of the implementation.

This implementation is iterating over the input once*. We keep track of 3 segments

```
qrur9Dvnyt5:gt:10
^^^^^^^^^^^ ^^ ^^
     uid    operator
                operatorOrValue
```

You can think of the parsing function as a state machine with the above as the 3 states. Unescaped `:` separate segments and unescaped `,` separate filters.

DHIS2-19331 dot

We collect a filter once we have reached a `,`, the end of input or 3 segments and a `:`.

Most of the complexity we have is due to users being able to mix unary/binary operators. Additionally, due to us wanting to provide error messages that help user to fix their filters by themselves.

(*it's almost one pass as I kept the `removeEscapeCharacters` method which iterates over the value to unescape the `/`, `:`, `,`. I don't think we need to optimize this.)

## Behavior change

We had a limit of 2 operators per UID in a single request parameter 😅 

This meant

`filter=foo:sw:3:ew:5:neq:345`

was not allowed, while splitting it across multiple parameters was

`filter=foo:sw:3:ew:5&filter=foo:neq:345`

Since both mean the same thing we now consistently apply the limit to operators per UID. We also increased the limit to 3 so users can maybe combine operators like `sw`, `ew` and `neq`. This limit can be increased in the future or dropped. We don't really know where it came from. Maybe it was a limitation in the implementation.

## Bug

```http
### binary operator with no value following a unary operator does not fail but should
  program=VBqh0ynB2wv&
  fields=dataValues[dataElement,value]&
  filter=qrur9Dvnyt5:!null:gt
```

Now fails with

```json
{
  "httpStatus": "Bad Request",
  "httpStatusCode": 400,
  "status": "ERROR",
  "message": "filter=qrur9Dvnyt5:!null:gt is invalid. Binary operator 'gt' must have a value.",
  "errorCode": "E1003"
}
```

```http
### previously lead to 400: UID must be an alphanumeric string of 11 characters starting with a letter, but was:
# we do accept empty commas, so we should also let this one go
  program=VBqh0ynB2wv&
  fields=dataValues[dataElement,value]&
  filter=F3ogKBuviRA,,qrur9Dvnyt5,
```

This will now work and the empty `,,` is ignored.

## Example error messages

All error messages are now prefixed with `<paramName>=<value> is invalid`. The new messages below are shown without the prefix.

* `nouid:eq:2` which contains an invalid UID
  * before: failed with `IllegalArgumentException` the `UID must be an alphanumeric string of 11...` message
  * now: fails with `BadRequestException` and the `UID must be an alphanumeric string of 11...` message
* `TvjwTPToKHO:lt::gt:10` which contains multiple binary operators one of which does not have a value
  * before: `Operator in filter must be be used with a value: TvjwTPToKHO:lt::gt:10`
  * now: `Binary operator 'lt' must have a value` 
* `TvjwTPToKHO::null,cy2oRh2sNr6:!null`
  * before: `Query item or filter is invalid: qrur9Dvnyt5::gt:10`
  * now: `UID 'TvjwTPToKHO' is missing an operator (or has too many ':').`
* `filter=qrur9Dvnyt5:nll`
  * before: `Query item or filter is invalid: qrur9Dvnyt5:nll`
  * now: `'nll' is not a valid operator`

We are now pointing the user to which part of the filter is wrong.

## Tests

Added more tests as I was trying to understand the old implementation. One of which shows we do support filter values with unicode code points beyond the ASCII range.

## Previous changes to `filter` parsing

These are some important changes we made in the past

* https://github.com/dhis2/dhis2-core/pull/13981
* https://github.com/dhis2/dhis2-core/pull/19778

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
@@ -63,6 +63,7 @@ import org.hisp.dhis.commons.util.TextUtils;
 import org.hisp.dhis.feedback.BadRequestException;
 import org.hisp.dhis.util.ObjectUtils;
 import org.hisp.dhis.webapi.controller.event.webrequest.OrderCriteria;
+import org.hisp.dhis.webapi.controller.tracker.export.FilterParser;
 
 /**
  * RequestParamUtils are functions used to parse and transform tracker request parameters. This
@@ -283,7 +284,7 @@ public class RequestParamsValidator {
   /**
    * Validate the {@code filter} request parameter in change log tracker exporters. Allowed filter
    * values are {@code supportedFieldNames}. Only one field name at a time can be specified. If the
-   * endpoint supports UIDs use {@link #parseFilters(String)}.
+   * endpoint supports UIDs use {@link FilterParser#parseFilters(String)}.
    */
   public static void validateFilter(String filter, Set<Pair<String, Class<?>>> supportedFields)
       throws BadRequestException {
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
@@ -29,7 +29,6 @@
  */
 package org.hisp.dhis.webapi.controller.tracker;
 
-import static org.hisp.dhis.common.DimensionalObject.DIMENSION_NAME_SEP;
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ACCESSIBLE;
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.ALL;
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.CAPTURE;
@@ -38,25 +37,16 @@ import static org.hisp.dhis.common.OrganisationUnitSelectionMode.DESCENDANTS;
 import static org.hisp.dhis.common.OrganisationUnitSelectionMode.SELECTED;
 
 import jakarta.servlet.http.HttpServletRequest;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.hisp.dhis.common.CodeGenerator;
 import org.hisp.dhis.common.OrganisationUnitSelectionMode;
-import org.hisp.dhis.common.QueryFilter;
-import org.hisp.dhis.common.QueryOperator;
 import org.hisp.dhis.common.UID;
 import org.hisp.dhis.common.collection.CollectionUtils;
 import org.hisp.dhis.commons.util.TextUtils;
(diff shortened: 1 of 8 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
```diff
@@ -283,7 +284,7 @@ public static void validateOrderParams(List<OrderCriteria> order, Set<String> su
   /**
    * Validate the {@code filter} request parameter in change log tracker exporters. Allowed filter
    * values are {@code supportedFieldNames}. Only one field name at a time can be specified. If the
-   * endpoint supports UIDs use {@link #parseFilters(String)}.
+   * endpoint supports UIDs use {@link FilterParser#parseFilters(String)}.
```
enricocolasante, line 287: `parseFilters(String)` method in this class can now be removed, right?

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/FilterParserTest.java
```diff
@@ -0,0 +1,369 @@
...
+import org.hisp.dhis.common.QueryFilter;
+import org.hisp.dhis.common.QueryOperator;
+import org.hisp.dhis.common.UID;
+import org.hisp.dhis.feedback.BadRequestException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/** Tests {@link FilterParser}. */
+class FilterParserTest {
```
enricocolasante, line 50: I think there is a missing arrow in the state machine (and a missing test) from operator to UID when there is a `,` (`TvjwTPToKHO:null,cy2oRh2sNr6:like:foo` is valid)

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/FilterParser.java
```diff
@@ -0,0 +1,273 @@
...
+
+    int start = 0;
+    int end = 0;
+    UID uid = null;
+    String operator = null;
+    String valueOrOperator = null;
+    int length = input.length();
+    while (end < length) {
+      int curChar = input.codePointAt(end);
+      int charCount = Character.charCount(curChar);
```
enricocolasante, line 113: I would add a comment explaining that `charCount` is needed to correctly deal with UNICODE chars.
Is it what is doing, right?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/FilterParser.java
```diff
@@ -0,0 +1,273 @@
...
+
+      // skip escaped slash, colon and comma which allow users to pass them as part of values
+      if (curChar == ESCAPE && end + 1 < length) {
+        int nextChar = input.codePointAt(end + 1);
+        if (nextChar == ESCAPE || nextChar == COLON || nextChar == COMMA) {
+          end += charCount + Character.charCount(nextChar);
```
enricocolasante, line 119:
```suggestion
          end += charCount + 1;
```

Would it make sense to make it explicit that a separator or escape char counts as 1?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestParamsValidator.java
teleivo: of course 😂 🤦🏻

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/FilterParser.java
teleivo: by comment, constant or `Character.charCount(COMMA)` 😅

dhis-2/dhis-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/FilterParserTest.java
teleivo: Added test to `shouldParseFiltersWithCombinedUnaryAndBinaryOperators`

enricocolasante (approved):

muilpp (approved):

PR #20438, opened 2025-03-27
fix(merge-orgunits): remove two sources restriction

Birkbjo:
Fixes https://dhis2.atlassian.net/browse/DHIS2-19251


david-mackessy (approved):
This _might_ be all that's needed. The only way to know is to test it.
This comes with a big caveat: because this feature looks like it was originally designed to use an empty/placeholder target OrgUnit, it's not obvious whether all target metadata/data is kept as you might expect.
[Docs](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_organisation_unit_merge) even say `"at a minimum ensure that no aggregate data exists for the target org unit"`

jbee (approved):

jason-p-pickering:
@david-mackessy your comment has me completely stumped then. My thinking on this from the very start was that this feature was meant to essentially deduplicate a facility, not merge two into an empty one. This just does not fit with the actual normal use case where you have a duplicate facility and need to merge that duplicate into the "real" one. If you are saying that we cannot do this, then we need to probably take a step back and consider what merging an orgunit actually means.

david-mackessy:
I hear you @jason-p-pickering . [This feature](https://dhis2.atlassian.net/browse/DHIS2-11281) is ~3 years old and was the first of the metadata merge features. It's the only one I don't know intimately. [This looks like the list ](https://github.com/dhis2/dhis2-core/blob/b6e65e9136f42f16270fc8dc4bad6a1576b82a59/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/orgunit/DefaultOrgUnitMergeService.java#L120)of metadata/data that would need to be tested to make sure you get expected results. If there are any surprises or changes required, I can help out.

jason-p-pickering:
I think we need @larshelge to comment here, as per the original commit for this feature, it is stated "This issue is about introducing support for merging organisation units with related metadata, aggregate data and tracker data.A typical scenario is two or more administrative districts which have merged into one district."

I would say that this is **not** the typical scenario, rather "A typical scenario is when a duplicate facility has been created and needs to be merged into the real facility". 

Merging districts can happen (splits are more frequent), and in this case, I agree that you might take two districts and merge them into a single new one, but the case of needing to merge duplicate facilities (at the leaf node) occurs far more frequently. The code looks OK to me as do the tests, but it is probably prudent for more eyes on it to be sure we are not missing something.

larshelge (approved):

PR #20453, opened 2025-03-28
feat: Implement integrity check for default category option sharing

jason-p-pickering:
The default category option should be shared publicly. If it isn't this can cause issues for general data entry and visibility. This is especially important in current versions of DHIS2, as it is only possible to modify the sharing via the database (the API blocks any modification of the sharing settings, even if they are wrong).


jbee (approved):
:+1:

david-mackessy (approved):

PR #20450, opened 2025-03-28
chore: update google formatter version

teleivo:
to benefit from improved textblock handling

see
https://github.com/google/google-java-format/releases/tag/v1.26.0

```java
  private static final String yes =
        """
hello
""";
```

turns to

```java
  private static final String yes =
"""
hello
""";
```

while this is left alone

```java
  private static final String no =
      """
        hello
        """;
```

PR #20459, opened 2025-03-28
chore: update google formatter version [2.40]

teleivo:
backport of https://github.com/dhis2/dhis2-core/pull/20450

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ dhis-2/projectFilesBackup
 coverage
 node_modules
 **/rebel.xml
+.vscode
+.cursor
 # ignore generated artifact directory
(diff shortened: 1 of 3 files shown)


teleivo:
We won't update the formatter on 2.40 since:

Updating it for 2.40 failed on github as we use JDK 11 (and the formtter update needs a higher one). 2.40 requires a minimum of JDK 11 which is why we use JDK 11 on GH/Jenkins.
https://docs.dhis2.org/en/manage/getting-started/manual-install-on-ubuntu.html?h=jdk#install_java_installation

Textblocks for which this formatter update has some improvements came in JDK 15 so we will need to fiddle with backports anyway.

PR #20465, opened 2025-03-30
fix: use DATETIME(3) for Doris analytics table columns to preserve milliseconds [DHIS2-19531]

luciano-fiandesio:
## Summary

This PR updates the Doris SQL dialect to define `DATETIME` columns with a precision of 3 decimal places (i.e., `DATETIME(3)`) instead of the default `DATETIME(0)`.
Doris stores `DATETIME` values with no fractional seconds by default (`DATETIME` is treated as `DATETIME(0)`), causing the loss of millisecond information even when present in the source data.

By explicitly setting the precision to 3:

- Milliseconds are correctly stored and retrieved via JDBC.

## Details

- Modified `DorisSqlBuilder.dataTypeTimestamp()` and `dataTypeTimestampTz()` to return `datetime(3)` instead of `datetime`.
- This change ensures that when analytics tables are generated, timestamp columns preserve millisecond precision.


larshelge (approved):

larshelge:
Nice fix.

PR #20491, opened 2025-03-31
fix: correctly return apps in the order set in menu management [DHIS2-19124]

amcgee:
Prior to Struts removal the `getModules.action` endpoint returned apps in the order set in the Menu Management app (see https://github.com/dhis2/dhis2-core/blob/2.41/dhis-2/dhis-web/dhis-web-commons/src/main/java/org/hisp/dhis/webportal/menu/action/GetModulesAction.java#L61-L74)

This ordering wasn't supported in the `/api/apps/menu` endpoint, which is now serving the `getModules.action` path.

Ordering is preserved in the User object in the database, however if someone tries to re-order their favorite apps without this fix in place, then new order will be preserved (just not shown in the menu until this fix in merged)

Candidate for 2.42.0 but not yet approved by RCB

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
@@ -79,6 +79,8 @@ import org.hisp.dhis.jsontree.JsonString;
 import org.hisp.dhis.query.QueryParserException;
 import org.hisp.dhis.security.Authorities;
 import org.hisp.dhis.user.CurrentUserUtil;
+import org.hisp.dhis.user.User;
+import org.hisp.dhis.user.UserService;
 import org.hisp.dhis.util.AppHtmlTemplate;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):
a test would be nice to catch if it changes/breaks again in the future

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -242,7 +245,27 @@ public App getApp(String appName) {
...
-    return getAccessibleAppMenu(contextPath);
+    List<WebModule> modules = getAccessibleAppMenu(contextPath);
+
+    // Apply user-defined favorite apps order
+    User currentUser = userService.getUserByUsername(CurrentUserUtil.getCurrentUsername());
+
+    if (currentUser != null && currentUser.getApps() != null && !currentUser.getApps().isEmpty()) {
+      final List<String> userApps = new ArrayList<>(currentUser.getApps());
+
+      modules.sort(
```
jbee, line 256: I think what you want is `Comparator.comparing(User:getName).nullsLast()` or something like that, can't remember the exact names of how you control nulls but there is something for it.

amcgee: I think this is a little different because it's looking up the name in a different ordered list and placing any not-found at the end.  This is exactly copied from the previous implementation in GetModulesAction (now removed), definitely could be improved in the future but that has been working for a long time so I think the lowest-risk thing to do is just keep it as-is for now and refactor in the future.

PR #20513, opened 2025-04-02
feat: Add enrollment count metrics

jason-p-pickering:
Small PR which adds enrollments to the data summary API JSON and metrics output.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/datasummary/DataSummary.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/datasummary/DataSummary.java
@@ -60,5 +60,7 @@ public class DataSummary {
 
   @JsonProperty private Map<Integer, Long> eventCount = new HashMap<>();
 
+  @JsonProperty private Map<Integer, Long> enrollmentCount = new HashMap<>();
+
   @JsonProperty private Dhis2Info system;
 }
--- a/dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datastatistics/DefaultDataStatisticsService.java
+++ b/dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datastatistics/DefaultDataStatisticsService.java
@@ -245,6 +245,29 @@ public class DefaultDataStatisticsService implements DataStatisticsService {
         30, (long) idObjectManager.getCountByLastUpdated(Event.class, todayMinusDays(30)));
     statistics.setEventCount(eventCount);
 
+    Map<Integer, Long> enrollmentCount = new HashMap<>();
+    enrollmentCount.put(
+        0,
+        (long)
+            idObjectManager.getCountByLastUpdated(
+                org.hisp.dhis.program.Enrollment.class, todayMinusDays(0)));
+    enrollmentCount.put(
+        1,
+        (long)
+            idObjectManager.getCountByLastUpdated(
+                org.hisp.dhis.program.Enrollment.class, todayMinusDays(1)));
+    enrollmentCount.put(
+        7,
+        (long)
+            idObjectManager.getCountByLastUpdated(
+                org.hisp.dhis.program.Enrollment.class, todayMinusDays(7)));
+    enrollmentCount.put(
+        30,
+        (long)
+            idObjectManager.getCountByLastUpdated(
+                org.hisp.dhis.program.Enrollment.class, todayMinusDays(30)));
+    statistics.setEnrollmentCount(enrollmentCount);
+
     statistics.setSystem(getDhis2Info());
 
     return statistics;
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DataSummaryControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/DataSummaryControllerTest.java
@@ -70,6 +70,9 @@ class DataSummaryControllerTest extends PostgresControllerIntegrationTestBase {
         content.lines().anyMatch(line -> line.startsWith("data_summary_event_count")),
         "Event count metric is missing");
     assertTrue(
+        content.lines().anyMatch(line -> line.startsWith("data_summary_enrollment_count")),
+        "Enrollment count metric is missing");
+    assertTrue(
         content.contains("# HELP data_summary_build_info"), "Build info help text is missing");
     // data_summary_build_info should end with an integer representing the build time in seconds
     // since epoch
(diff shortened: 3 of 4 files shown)


jason-p-pickering:
@david-mackessy thank you for your thorough review and clarification on how the backend team would like to approve external PRs. Going forward, I will be sure to adhere to your instructions. 

In terms of performance implications, here are the results from my profiling of the current master branch (a73de7a) and this feature branch. 

For the master branch: 

Selection_767

For the feature branch: 

Selection_766

Thus, the timings of the two versions seem to be very similar to each other. 

Additionally, these types of metrics only need to be collected rather infrequently, perhaps at most a few times a day. Considering that the timings for the methods are under half a second and similar, I do not think there is any significant performance hit by this feature. If additional information is required, I can try and provide it.

david-mackessy (approved):
Thanks @jason-p-pickering , that's fantastic info 👍
The reason I asked about it was because I don't know how the metrics work and I find that seldom, if ever, reviewers question what kind of testing was done and code just gets pushed through (not knowing if the change has had a positive or a negative effect)

jason-p-pickering:
Under the hood, the method in question here uses ultimately 

```
  @Override
  public <T extends IdentifiableObject> int getCountByLastUpdated(
      @Nonnull Class<T> type, @Nonnull Date lastUpdated) {
    IdentifiableObjectStore<T> store = getIdentifiableObjectStore(type);

    if (store != null) {
      return store.getCountGeLastUpdated(lastUpdated);
    }

    return 0;
  }
```
This could probably be sped up even further with better indexing on the `enrollment` table, since the `lastupdated` field is not actually indexed

```
Indexes:
    "programinstance_pkey" PRIMARY KEY, btree (enrollmentid)
    "in_programinstance_deleted" btree (deleted)
    "in_programinstance_programid" btree (programid)
    "in_programinstance_trackedentityinstanceid" btree (trackedentityid)
    "index_programinstance" btree (enrollmentid)
    "programinstance_uid_key" UNIQUE CONSTRAINT, btree (uid)

```

This process could likely be optimized a bit if an additional index was added. 

I am not certain what the exact query is, but with no index on the table, the query planner is going to have to do a sequential scan. 

```
events=# EXPLAIN ANALYZE SELECT COUNT(*) FROM enrollment where age(now(),lastupdated) <='30 days'::interval;
                                                    QUERY PLAN                                                    
------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1409.56..1409.57 rows=1 width=8) (actual time=9.674..9.675 rows=1 loops=1)
   ->  Seq Scan on enrollment  (cost=0.00..1392.46 rows=6841 width=0) (actual time=0.061..9.641 rows=874 loops=1)
         Filter: (age(now(), (lastupdated)::timestamp with time zone) <= '30 days'::interval)
         Rows Removed by Filter: 19650
 Planning Time: 0.058 ms
 Execution Time: 9.695 ms
(6 rows)
```

With the index though in place, things likely would be much faster (at least for this operation). These optimziations might be better left to individual implementers to decide to put in place, unless there is a general agreement that this might be warranted. 

```
events=# CREATE INDEX idx_enrollment_lastupdated on enrollment(lastupdated);
CREATE INDEX
events=# EXPLAIN ANALYZE SELECT COUNT(*) FROM enrollment where age(now(),lastupdated) <='30 days'::interval;
                                                                       QUERY PLAN                                                                        
---------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=751.52..751.53 rows=1 width=8) (actual time=4.061..4.062 rows=1 loops=1)
   ->  Index Only Scan using idx_enrollment_lastupdated on enrollment  (cost=0.29..734.42 rows=6841 width=0) (actual time=3.837..4.032 rows=874 loops=1)
         Filter: (age(now(), (lastupdated)::timestamp with time zone) <= '30 days'::interval)
         Rows Removed by Filter: 19650
         Heap Fetches: 103
 Planning Time: 0.185 ms
 Execution Time: 4.077 ms
(7 rows)
```


This is basically the same performance problem I describe in this ticket https://dhis2.atlassian.net/browse/DHIS2-19112, which makes use of the same method.  For enrollments (at least on this DB with 62k) its not an issue, but I suspect for larger tracker implementations it will be an issue.

ameenhere (approved):
Approving this, but these type of metrics in prometheus is an anti-pattern. Will have to take this up in another discussion. 

Basically these kind of database queries feeding the metrics is not ideal. In a normal monitoring setup that consumes these `/metrics` endpoint (like Grafana or any other monitoring tool), there will be a scraper that hits this endpoint regularly, every 10 seconds for example. This simply adds unnecessary load. Granted that "nobody" may have such monitoring setup currently. But still. 

Even keeping the scraping factor aside, I'd still prefer these statistics to be of lower overhead somehow. Not sure if we cache sufficiently here?  Or use the internal pg_stats table for approximate counts? In general, the current approach for getting data statistics makes me uncomfortable.

ameenhere:
@jason-p-pickering I agree with a need to see the summary/statistics of things once a day (or at some lower frequency). My only concern is attaching (or mixing) these metrics into api/dataSummary/metrics endpoint which "looks" like is designed to be consumed by a prometheus scraper for example (or any other "live" monitoring tool). The controller is named "getPrometheusMetrics", the metrics object is Prometheus style and that is an additional place where we [attach](https://github.com/dhis2/dhis2-core/pull/20513/files#diff-4f2861012b6b96ea0d11d913e2e3008217d9d99d788a2a8f57bf361d01c79437R139) all these heavy metrics. 

Maybe it is not intended to be consumed like that. Maybe the naming is just misleading and prompted me to give this feedback.  As long as implementations are clear this endpoint should not be used that way, then we are good.

jason-p-pickering:
@ameenhere I am not sure you saw the results of my profile and SQL analysis higher up in this ticket?  The main problem with any query to identifiable objects is the lack of an index on the `lastUpdated` field. 

It is worth pointing out there that clients such as as the capture app make "scrapes" to their endpoints which are many many orders of magnitude more expensive, often many times a second driven by users. There are quite a few tickets on this issue already.  This is a metrics endpoint intended for administrators, who should understand that it is not really useful to scrape real-time metrics on enrollment counts since we are only offering metrics on a daily basis anyway. If people need higher resolution, they can get it from reverse proxy logs or the normal metrics endpoint.  Considering that this particular query (on a database with out an index on the proper column) only takes a few seconds and should not be accessed that frequently, I am not sure it is really an issue at all (considering other performance issues which we have). 

I am not disagreeing with you that we should look into better ways of collecting this data, but there are much bigger performance issues than this particular one, which is something only of interest to admins and should be accessed extremely infrequently.

ameenhere:
@jason-p-pickering 

> I am not sure you saw the results of my profile and SQL analysis higher up in this ticket?
Yes, I did! I also approved this PR 😄. I just felt that in general the approach of tying db queries into prometheus style metrics endpoint may cause issues if the endpoint is consumed regularly. But in this specific case it is OK since it is not used for live monitoring and only used by admins infrequently.

Just to be clear, I am not implying "this PR" is problematic or causing me concern. But simply pointing out that at some point we should look at better ways to fulfill the requirements of DataSummaryController endpoints (especially the metrics endpoint). 

And regarding Capture / Tracker performance issues, we are actively looking into them. 👍  We will follow up with you (and other implementation experts) to make sure we are on the right track.

PR #20551, opened 2025-04-07
fix: restore url navigation using address bar in analytics apps [DHIS2-19275]

jenniferarnesen:


janhenrikoverland (approved):
Verified on d2-ci

edoardo (approved):

PR #20553, opened 2025-04-07
fix: don't redirect to global shell if request comes from an iframe

amcgee:


KaiVandivier (approved):
Tested in PR deploy; also in this instance: https://dev.im.dhis2.org/global-shell-redirect-changes

PR #20555, opened 2025-04-07
feat: Allow for maxResults in validation rule and outlier analysis

jason-p-pickering:
Allow for an optional parameter `maxResults` in two data quality analysis endpoints `/api/dataAnalysis/validationRules`  and `api/outlierDetection` to override the default of hard-coded limit of 500 results per query. The current outlier API allows for specification of the maximum number of results, but it is capped at a maximum of 500. 

Example payload for the SL database

```
{"startDate":"2025-01-01",
"endDate":"2025-04-07",
"ou":"ImspTQPwCqd",
"notification":false,
"persist":false, 
"maxResults": 5000}
```

When performing bulk analysis (perhaps at the national or provincial level) having more than 500 results is very common. While we may not want to support this for now in the web user interface, other apps/clients may be able to take advantage of a larger number of responses in order to speed up this type of analysis. 

A new system setting has been added `keyDataQualityMaxLimit` which should allow system admins to increase the limit of records available via an API request. 

For the validation rule analysis, the current web client does not specify a number of max results. In this case, we stick with the previous default of 500. 

For the outlier analysis, the current web client does send a maximum number of records, but currently 500 is the maximum available via the user interface.


jason-p-pickering:
Converting this to a draft until we can test on an actual "largish" implementation. Will provide feedback.

jbee (approved):
Looks good. 

Just noticed that in your example in the PR description you give the value as `"5000"` and not `5000`. While this works a number should be given as a JSON number, not a string.

jason-p-pickering:
We tested this on a copy of an actual country database. The patch was applied to 2.41.3. A combination of paramaters were chosen such that more than 500 validation rule results were obtained. 

image


However, when looking directly at the API with Postman, we got back 1005 results
image

This request itself took about 30 seconds, which is deemed  to be acceptable for this type of analysis.

jason-p-pickering:
We tested this using a custom Python client on a copy of a country HMIS database. We ran a national level analysis for the past 12 months for about 15 validation rule groups of various sizes. Using 4 concurrent worker threads, the server seemed to hold up fine and the process completed after about 10 minutes. This is substantially faster than our previous approach, where we needed to query individual organisation units with many more requests. So in summary, it looks like this works!

david-mackessy (approved):

jason-p-pickering:
This was tested on the same database using the outlier analysis. 

We were able to retrieve more than 500 results, once the system setting had been modified. The query took 2.67 seconds.

```
{
"metadata": {
"algorithm": "Z_SCORE",
"threshold": 3,
"orderBy": "MEAN_ABS_DEV",
"maxResults": 5000,
"count": 568
}

```

david-mackessy (approved):

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/DataAnalysisController.java
```diff
@@ -171,6 +175,17 @@ public class DataAnalysisController {
       throw new WebMessageException(badRequest("No organisation unit defined"));
     }
 
+    int maxResults =
```
larshelge, line 178: You can use this for a more readable and compact syntax:

`ObjectUtils.firstNonNull( validationRulesAnalysisParams.getMaxResults(), ValidationService.MAX_INTERACTIVE_ALERTS )`

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ObjectUtils.html

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/DataAnalysisController.java
```diff
@@ -171,6 +175,17 @@ public class DataAnalysisController {
...
+        validationRulesAnalysisParams.getMaxResults() != null
+            ? validationRulesAnalysisParams.getMaxResults()
+            : ValidationService.MAX_INTERACTIVE_ALERTS;
+
+    final int MAX_ALLOWED_RESULTS = settingsProvider.getCurrentSettings().getDataQualityMaxLimit();
+
+    if (maxResults <= 0 || maxResults > MAX_ALLOWED_RESULTS) {
+      throw new BadRequestException("maxResults must be between 1 and " + MAX_ALLOWED_RESULTS);
```
larshelge, line 186: Use "Max results" instead of "maxResults" in error message which is what is typically used.

PR #20556, opened 2025-04-07
chore: Clean up controller.event package in dhis-api [DHIS2-17403]

enricocolasante:
Move `PagingAndSortingCriteriaAdapter` close to `DimensionCriteria` which is the only class using it and remove all unused methods.


teleivo (approved):
lets maybe let @dhis2/analytics-backend give the second ✅

maikelarabori (approved):

PR #20569, opened 2025-04-08
feat: Add database indices to improve data summary

jason-p-pickering:
The data summary API `api/dataSummary` . Using a profiler, the actual query used by DHIS2 to obtain the latest count of events is something like this

```
events=# EXPLAIN ANALYZE select count(distinct event0_.eventid) as col_0_0_ from event event0_ where event0_.lastUpdated>= (SELECT now() - '30 days'::interval);
                                                            QUERY PLAN                                                            
----------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1443924.36..1443924.37 rows=1 width=8) (actual time=2748.414..2748.416 rows=1 loops=1)
   InitPlan 1 (returns $0)
     ->  Result  (cost=0.00..0.01 rows=1 width=8) (actual time=0.006..0.006 rows=1 loops=1)
   ->  Seq Scan on event event0_  (cost=0.00..1434693.64 rows=3692284 width=8) (actual time=273.921..2739.198 rows=60316 loops=1)
         Filter: (lastupdated >= $0)
         Rows Removed by Filter: 11016535
 Planning Time: 0.085 ms
 Execution Time: 2748.762 ms
(12 rows)

```

This particular database has around 11 million events and 20k enrollments The query takes around 3 seconds per query to execute, resulting in a sequential scan from the query planner.

With the index in place, the query is sped up by quite a bit. 

```
events=# EXPLAIN ANALYZE select count(distinct event0_.eventid) as col_0_0_ from event event0_ where event0_.lastUpdated>= (SELECT now() - '30 days'::interval);
                                                                         QUERY PLAN                                                                         
------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=604275.71..604275.72 rows=1 width=8) (actual time=127.377..127.378 rows=1 loops=1)
   InitPlan 1 (returns $0)
     ->  Result  (cost=0.00..0.01 rows=1 width=8) (actual time=0.007..0.008 rows=1 loops=1)
   ->  Index Scan using in_event_lastupdated on event event0_  (cost=0.43..595044.99 rows=3692284 width=8) (actual time=30.313..118.915 rows=60320 loops=1)
         Index Cond: (lastupdated >= $0)
 Planning Time: 0.095 ms
 Execution Time: 127.734 ms
(11 rows)
```

Indices can be added to  `event` and `enrollment` to speed these queries up. Of course there is going to be some additional disk space required in order to support the index. 

```
events=#  \di+ in_event_lastupdated;
                                               List of relations
 Schema |         Name         | Type  |  Owner   | Table | Persistence | Access method |  Size  | Description 
--------+----------------------+-------+----------+-------+-------------+---------------+--------+-------------
 public | in_event_lastupdated | index | postgres | event | permanent   | btree         | 237 MB | 
(1 row)

events=#  \di+ in_enrollment_lastupdated;
                                                    List of relations
 Schema |           Name            | Type  |  Owner   |   Table    | Persistence | Access method |  Size  | Description 
--------+---------------------------+-------+----------+------------+-------------+---------------+--------+-------------
 public | in_enrollment_lastupdated | index | postgres | enrollment | permanent   | btree         | 424 kB | 
(1 row)
```

From the profiler, we see that the data statistics operations can be singificantly sped up with these indices: 

Before (~ 4 seconds)

image

After ( ~ 0.3 seconds)
image


ameenhere (approved):

jason-p-pickering:
> Looking at previous migrations, the convention seems to be to mention the table name(s) in the filename, making it very clear where the new indices are being added (without having to go in and read the code). Makes it much clearer when browsing through these files.

Thanks @david-mackessy . Will fix.

david-mackessy (approved):

PR #20570, opened 2025-04-08
task: Introduce tracker program service  [DHIS2-19437]

muilpp:
### Part I
We are manually filtering the program list to retrieve the programs accessible to the logged-in user. This is currently done in multiple places. I want to centralize this logic, so I'm creating a new service called `TrackerProgramService`.
This should help reduce complexity when validating whether a user has access to a particular TE/program pair.

### Next
We still need to move the ACL validation, which is currently performed on tracked entities after the database query.
Validating after retrieving TEs from the database breaks pagination.
Using the new service mentioned above will help facilitate this next step.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/TrackerProgramService.java
```diff
@@ -0,0 +1,42 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.acl;
+
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.hisp.dhis.program.Program;
+import org.hisp.dhis.trackedentity.TrackedEntityType;
+
+public interface TrackerProgramService {
```
teleivo, line 37: Add javadocs.

* Why does it exist? when should we use it? Especially regarding: when reading `TrackerProgramService` a part of me is wondering if this is dealing with tracker programs or if the tracker prefix is just to distinguish it from the `ProgramService` 😅 
* mention to whom these programs are accessible
* do we assume the user has access to the TET? or do we check that as well? document

Add `@Nonnull`/`@CheckForNull` annotations also to return values.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/acl/TrackerProgramServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+    TrackedEntityType trackedEntityType = manager.get(TrackedEntityType.class, "ja8NY4PW7Xm");
+    assertNotNull(trackedEntityType);
+    assertEquals(
+        accessiblePrograms,
+        getUids(trackerProgramService.getAccessibleTrackerPrograms(trackedEntityType)));
+  }
+
+  @Test
+  void shouldReturnEmptyListWhenNoProgramAccessible() {
+    makeProgramInaccessible("TsngICFQjvP");
```
teleivo, line 106: I would expect us to have two tests making the same call, using the same input. The result of one test should then include a program and the other should not.

It looks like `shouldReturnAllAccessibleProgramsByTrackedEntityType` and this test `shouldReturnEmptyListWhenNoProgramAccessible` are those tests but the use a different TET as input.

teleivo (commented):
You mention

> Using the new service mentioned above will help facilitate this next step.

How does it do that? 🤔  

Because by encapsulating this logic, we reduce the complexity of the validations, so it should be easier to move part of it to the operations mapper, and the other part in the query.

muilpp:
> You mention
> 
> > Using the new service mentioned above will help facilitate this next step.
> 
> How does it do that? 🤔

Because by encapsulating this logic, we reduce the complexity of the validations, so it should be easier to move part of it to the operations mapper, and the other part in the query.

enricocolasante (approved):

ameenhere (approved):

PR #20574, opened 2025-04-08
chore: update Syrian flag

kabaros:
The Syrian flag was formally changed by [article 6 of the interim constitution](https://en.wikipedia.org/wiki/Flag_of_Syria#In_Constitution) on 13 March 2025.

- requested by HISP MENA


kabaros:
@Philip-Larsen-Donnelly = this was requested by Adulrahman from HISP MENA (they have demos soon in Syria for DHIS2). I am not sure if we have a formal process for updates like this, let me know if we do (I discussed it with Max and he said it is sensible to update the flag). 

(If we update it, I think we should also backport it to all supported versions?)

jason-p-pickering (approved):

tonsV2 (approved):

PR #20576, opened 2025-04-08
fix: (2.39) notification trigger cause server halt [DHIS2-19452]

zubaira:
While testing [this](https://dhis2.atlassian.net/browse/DHIS2-18517) issue, we observed that the server occasionally halts or restarts when a user attempts to send scheduled notifications.

This PR will now be deployed to the IM environment for further testing. If everything works as expected there, we can proceed with merging it.

https://dhis2.atlassian.net/browse/DHIS2-19452

**Filtering at the Database Level**
Previously, all ProgramNotificationInstance records were loaded from the database and filtered in memory based on the scheduledAt date.
This PR moves that filtering to the database query itself, ensuring that only relevant notifications scheduled for the current day are fetched.

**Reducing JSONB Payload Size**
The ProgramNotificationTemplateSnapshot column was previously storing the full serialized template object, including redundant and unnecessary data.
This caused heavy query delays due to bloated row sizes.
The updated logic stores only the essential fields in the snapshot, significantly improving query performance and database efficiency.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/HibernateProgramNotificationInstanceStore.java
```diff
@@ -118,9 +124,30 @@ private List<Function<Root<ProgramNotificationInstance>, Predicate>> getPredicat
...
+          root ->
+              builder.and(
+                  builder.greaterThanOrEqualTo(root.get("scheduledAt"), scheduledAt.getLeft()),
+                  builder.lessThan(root.get("scheduledAt"), scheduledAt.getRight())));
     }
 
     return predicates;
   }
+
+  public Pair<Date, Date> getStartAndEndOfDay(Date scheduledAt) {
```
zubaira, line 139: Only for this branch. For higher branches, I will have a param mapper mapping `scheduledAt` to `scheduledFrom` and `scheduledTo`

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationServiceTest.java
```diff
@@ -658,13 +661,13 @@ void testScheduledNotifications() {
               return new BatchResponseStatus(Collections.emptyList());
```
teleivo, line 661: Can we delete this mock test which is what we generally try to do and rely on dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationServiceIntegrationTest.java instead

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationServiceIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+        4, ruleEffects.size(), "Expected 3 rule effects for scheduled messages " + ruleEffects);
+  }
+
+  @Test
+  @Timeout(value = 30, unit = TimeUnit.SECONDS)
+  void testScheduledNotificationShouldBeTriggeredBeforeTimeout() {
+    /* No assertions are made here because this test is intended to verify that
+    scheduled notifications are fetched and processed within the specified time limit
+    (performance check only) */
+    programNotificationService.sendScheduledNotifications(NoopJobProgress.INSTANCE);
```
teleivo, line 121: This test is not really needed is it? The `notificationInstanceService.getAll()` was the issue right? So if you add a test to the `ProgramNotificationInstanceServiceTest` for `getAll` and the method you are using now showing they do not timeout that would be enough. This would also allow you to assert on their results.

This test here could be useful but IMHO you would need to assert that the notifications you expected to be sent were actually sent.

dhis-2/dhis-services/dhis-service-core/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationServiceTest.java
zubaira: yes. I will do that after creating integration tests for `ProgramNotificationService`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceService.java
```diff
@@ -47,4 +47,6 @@ List<ProgramNotificationInstance> getProgramNotificationInstances(
   Long countProgramNotificationInstances(ProgramNotificationInstanceParam params);
 
   void validateQueryParameters(ProgramNotificationInstanceParam param);
+
+  List<ProgramNotificationInstance> getAll();
```
enricocolasante, line 51: I agree with Ivo, this was the major culprit of the performance issue and it is only used in tests.
Can we remove it?

zubaira: The main performance issue was caused by `ProgramNotificationInstanceStore.getAll()`.
To test this properly, I added a corresponding method in `ProgramNotificationInstanceService`, which is now used in `ProgramNotificationInstanceServiceTest`.

teleivo: > The main performance issue was caused by ProgramNotificationInstanceStore.getAll().

which you then fixed by removing the orgUnit snapshot and using 
 `notificationInstanceService.getProgramNotificationInstances(param)` to get only the instances scheduled for today. This means that you should test `notificationInstanceService.getProgramNotificationInstances(param)` returns the instances you would expect (does so in a reasonable amount of time). If `ProgramNotificationInstanceStore.getAll()` is not used anymore you should remove it.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/HibernateProgramNotificationInstanceStore.java
teleivo: What do you mean by

> Only for this branch. For higher branches

zubaira: Normally, we handle this in the parameter mapper class. However, for version 2.39, I implemented it directly in the store. For higher versions, I plan to follow the standard approach using the parameter validation mapper class.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceService.java
zubaira: the performance fix involved two distinct changes:

- Filtering notifications in the database instead of in-memory (getProgramNotificationInstances(param)).
- Reducing the JSONB payload (by excluding unnecessary data like organisationUnit from the snapshot).

Both changes should be tested isn't? If not so, then we can remove `.getAll()`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/programrule/engine/ProgramNotificationInstanceServiceTest.java
```diff
@@ -136,6 +169,36 @@ void testGetProgramNotificationInstance() {
...
+    // Only one instance is scheduled for today's date
+    assertEquals(1, instancesForToday.size());
+    ProgramNotificationInstance instanceForToday = instancesForToday.get(0);
+    assertEquals(instanceForToday.getScheduledAt(), todayDate);
+  }
+
+  @Test
+  @Timeout(value = 20, unit = TimeUnit.SECONDS)
+  void testShouldGetAndSendScheduledNotificationInstanceWithoutTimeout() {
+    List<ProgramNotificationInstance> instances = programNotificationInstanceService.getAll();
```
enricocolasante, line 195: I don't fully get this test.
We do not need to have `getAll` implemented, to assert the same you could check for today and tomorrow notifications and see if they are there.

On the timeout, it is really a shot in the dark, this test with this timeout was probably passing even before your changes, so I am not sure it makes sense to have it at all.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceService.java
ameenhere: > Filtering notifications in the database instead of in-memory (getProgramNotificationInstances(param)).
For the above change, as long as you test/assert that the method is filtering notifications in the database and not in-memory, that should suffice. You don't necessarily need to test the "performance" side of this. 

> Reducing the JSONB payload (by excluding unnecessary data like organisationUnit from the snapshot).
For the above change, maybe simply asserting organisationUnit is not part of the snapshot json may also suffice? Again no need to create integration tests that focus on performance for this I feel. 

Thoughts @zubaira  @enricocolasante @teleivo ?

teleivo: I discussed this with @zubaira. We settled on writing an integration test for the service that gets the notification instances of today. That previously would have timed out. So now having a test for it that finishes is proof enough. We now have a global Junit timeout on tests of 1m so there is not even a need to add a `@Timeout` annotation on this test.

So even though the description mentions performance we test that it works without any target numbers.

ameenhere (approved):

PR #20581, opened 2025-04-08
task: Remove TET validation on tracked entities service [DHIS2-19437]

muilpp:
In `/trackedEntities`, we run a few validations after retrieving TEs from the database. One of these validations checks the TET access for each resulting TE. However, we can skip this validation at this point based on the following reasoning.

In a request to `/trackedEntities`, the user must specify one of the following parameters:
1. Program –> The operations mapper validates data access to the supplied program and its associated TET. Since a TE can only be enrolled in a program that matches its TET, this validation implicitly ensures access to the TE TET as well.
2. Tracked entity type –> The operations mapper validates data access to the specified TET.
3. Tracked entity UID(s) –> The operations mapper retrieves all TETs accessible to the user, which are then used in the query to ensure that the resulting TEs match the TETs the user is allowed to access.

For this reason, I'm replacing the call to `canRead` with the ownership check. Eventually, this check will need to be moved into the query itself, as it still breaks pagination. However, before doing that, we need to have a discussion with product about when a TE should be visible to the user.


teleivo (commented):
Are you saying we can remove the validation because its redundant?

I am wondering as you say

> However, before doing that, we need to have a discussion with product about when a TE should be visible to the user.

which makes me think, should we not discuss this first before making further changes to ACL 😅 

Does this PR make any changes to the service you introduced in https://github.com/dhis2/dhis2-core/pull/20570 ? It looks like this PR depends on the other one.

muilpp:
> Are you saying we can remove the validation because its redundant?
> 
> I am wondering as you say
> 
> > However, before doing that, we need to have a discussion with product about when a TE should be visible to the user.
> 
> which makes me think, should we not discuss this first before making further changes to ACL 😅
> 
> Does this PR make any changes to the service you introduced in #20570 ? It looks like this PR depends on the other one.

IMO this validation is redundant, as we can now achieve the same before hitting the database.
What I meant by discussing with product is related to the ownership validation. At the moment a user can see a TE if the program owner is accessible or could potentially enroll that TE in another program. This is what needs to be revisited.
This PR however, only removes the TET check after we get the results from the database, because that's already validated before. This TET check we need and it's independent from the ownership check.

Anyway, this PR should have been just a draft, it's still on progress, let's just focus on https://github.com/dhis2/dhis2-core/pull/20570 for now.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -288,7 +290,7 @@ public Set<String> getOrderableFields() {
   }
 
   private Predicate<TrackedEntity> filterAccessibleTrackedEntities(
-      UserDetails user, TrackedEntityQueryParams queryParams) {
+      UserDetails user, TrackedEntityQueryParams queryParams, List<Program> programs) {
     boolean skipOwnershipCheck = queryParams.getOrgUnitMode() == ALL;
 
     if (queryParams.hasEnrolledInTrackerProgram()) {
```
enricocolasante, line 296: Not part of this PR but this logic could be simplified with something like:
`return te -> (skipOwnershipCheck ||
ownershipAccessManager.hasAccess(user, te, queryParams.getEnrolledInTrackerProgram());`

enricocolasante (approved):

teleivo (approved):

PR #20586, opened 2025-04-09
fix: /tracker/events?filter/filterAttributes in, sw, ew, ... DHIS2-19347

teleivo:
[Detailed context is in](https://docs.google.com/document/d/1n5q4Z01aqpLabDnhs0tzsH_o6PZho-s0UJIdOocp-GU/edit?tab=t.0).

https://dhis2.atlassian.net/browse/DHIS2-13676 cast number/integer value types to numeric in `/events?filter/filterAttributes`. This broke all DHIS2 operators backed by SQL operators which do not have a numeric implementation like text based ones `like, sw, ew, ...`.

This change only casts attribute and data values to use DHIS2 value type semantics in comparison. So `170 == 170.0` when using `eq, in, ...` on number value type while `"170"` would not be considered equal to `"170.0"` using text. As well as `2 < 10` while `"2" > "10"`.

Operators such as `like, sw, ew, ...` will now work again as we do not cast values in that case but rely on the fact that all values are stored as strings.

The same will also be applied to `/trackedEntities?filter` in the next PR. Extracted the common code into `JdbcPredicate` as it will be used 3 times in total. This also makes it easier to understand and test.

## Tech Details

* Transform and encapsulate user input in a JDBC template parameter
https://docs.spring.io/spring-framework/reference/data-access/jdbc/parameter-handling.html#jdbc-in-clause. This shields against sql injection. It also handles multiple values like in `in (1, 2, 3)` without us having to build the SQL.
* [Single quote and backslash](https://jdbc.postgresql.org/documentation/escapes) are escaped for us by the JDBC driver
  * https://github.com/pgjdbc/pgjdbc/blob/156d724e1d95052b41a19fb568b2f81919ae2197/pgjdbc/src/main/java/org/postgresql/core/v3/SimpleParameterList.java#L213
  * https://github.com/pgjdbc/pgjdbc/blob/156d724e1d95052b41a19fb568b2f81919ae2197/pgjdbc/src/main/java/org/postgresql/core/Utils.java#L47
* like wildcards present in the user input need to be escaped by us `"%rainy_" -> '\%rainy\_'`

## Sample requests

```http
### valid: range filter
  program=VBqh0ynB2wv&
  fields=dataValues[dataElement,value]&
  filter=qrur9Dvnyt5:ge:10:le:20

### works now as it will not cast values if the operator is not implemented for non-text
  program=VBqh0ynB2wv&
  fields=dataValues[dataElement,value]&
  filter=qrur9Dvnyt5:sw:2

### 200: ew is implemented using SQL like
# can find value with like wildcards and quote
# like wildcard is escaped: %' \%case
  program=q04UBOqq3rp&
  fields=dataValues[dataElement,value]&
  filter=Da03a96443r:ew:' %case
```

## Next PR(s)

* fix `/trackedEntities?filter`
* In a separate issue: fix order for `/trackedEntities` which [does not take the TEA](https://github.com/dhis2/dhis2-core/blob/7d329fec7e2624b11568a4ab28e4801fa59aa62a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java#L998) value type into account. That means "2" appears after "10" due to `"2" > "10"` for numeric/integer value types. I think the same is true for `/events` (see [here](https://github.com/dhis2/dhis2-core/blob/7d329fec7e2624b11568a4ab28e4801fa59aa62a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java#L1478-L1481)).


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
```diff
@@ -0,0 +1,230 @@
...
+ * driver</a>!
+ *
+ * <p>On {@link SqlParameterValue}: Spring handles {@code List.of()} values for us no matter if we
+ * pass it to an <a
+ * href="https://docs.spring.io/spring-framework/reference/data-access/jdbc/parameter-handling.html#jdbc-in-clause">in</a>
+ * operator allowing multiple values or an operator only expecting one like {@code eq}.
+ * JdbcPredicate ensures we never pass a collection with multiple elements to such operators as this
+ * leads to a {@code BadSqlGrammarException}.
+ */
+record JdbcPredicate(String sql, Optional<Parameter> parameter) {
```
enricocolasante, line 84: Is there a way to make the constructor of the record private?
I understand that we should create `JdbcPredicate`s only with the static constructors, so it would be good to enforce that at code level.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
```diff
@@ -0,0 +1,230 @@
...
+
+  public static JdbcPredicate of(
+      @Nonnull DataElement de, @Nonnull QueryFilter filter, @Nonnull String tableName) {
+    Parameter parameter = parseFilterValue(de, filter);
+    String sql = generateSql(de, filter, parameter, tableName);
+    return new JdbcPredicate(sql, Optional.ofNullable(parameter));
+  }
+
+  @SuppressWarnings("unchecked")
+  private static Parameter parseFilterValue(
```
enricocolasante, line 101: I am not really fond of `Optional` fields in object but I can see the point here.
Can we at least make this method return a `Optional<Parameter>` so we avoid to call `Optional.ofNullable` in the caller?

enricocolasante (commented):
Nice job :clap: 
Just a few minor comments

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventQueryParams.java
```diff
@@ -474,44 +471,18 @@ public EventQueryParams filterBy(@Nonnull TrackedEntityAttribute tea) {
   }
 
   public EventQueryParams filterBy(@Nonnull DataElement de, @Nonnull QueryFilter filter) {
-    validateNumericFilterValue(de, filter);
     this.dataElements.putIfAbsent(de, new ArrayList<>());
-    this.dataElements.get(de).add(filter);
+    this.dataElements.get(de).add(JdbcPredicate.of(de, filter, "ev"));
```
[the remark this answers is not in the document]
teleivo: We are going to use this in TE store. That's why I did pass it in so its independent of a particular store.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  @Nonnull
+  private static String leftOperandSql(DataElement de, QueryOperator operator, String table) {
+    String column = table + ".eventdatavalues #>> '{" + de.getUid() + ", value}'";
+
+    String leftOperand;
+    if (operator.isUnary()) {
+      // `ev.eventdatavalues->'vANAXwtLwcT'` is not null is different from `lower(ev.eventdatavalues
+      // #>> '{deabcdefghB, value}') is not null`
```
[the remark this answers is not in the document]
teleivo: the uids should be the same yes. the point is both sql conditions (using the same uid) mean different things 😅 so we have to use `ev.eventdatavalues->'vANAXwtLwcTL' is not null`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
teleivo: We could only do that by making it a class with a private constructor. Should we?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
teleivo: If I do that I get linting complaints 😂 I don't mind as its there to avoid null, providing a different API. I don't really see why it would be odd. Either you deal with null or with Optional 🤷🏻 (or you make subclasses or so for Unary, Binary which seems too much)

enricocolasante: I would prefer the linting complaint over the use of `Optional.ofNullable`.
Just saying that because we already know from the `parseFilterValue` if the parameter is there or not.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcPredicate.java
enricocolasante: Maybe.
After all your work if someone arrives tomorrow and creates a `new JdbcPredicate` then all the consistency is gone.

enricocolasante (approved):

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/SqlUtils.java
```diff
@@ -133,13 +151,28 @@ public static Set<String> getArrayAsSet(ResultSet rs, String columnLabel) throws
...
+
   /**
    * Cast the given value to numeric (cast(X as numeric).
    *
    * @param value the value.
    * @return a string with the numeric cast statement.
    */
-  public static String castToNumber(String value) {
+  public static String castToNumeric(String value) {
     return "cast (" + value + " as numeric)";
```
ameenhere, line 176: Nitpicky, but couldn't this reuse the cast(String expression, String type) method by passing on numeric into it?

ameenhere (approved):
LGTM! 
Good job and a nice refactor. We need to make sure we test this together.

dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/SqlUtils.java
teleivo: I'll do that in the next PR

PR #20606, opened 2025-04-11
chore: Clean up controller.event package in dhis-api [DHIS2-17403]

enricocolasante:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/dxf2/webmessage/responses/TrackerJobWebMessageResponse.java
```diff
@@ -27,7 +27,7 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package org.hisp.dhis.webapi.controller.tracker.imports;
+package org.hisp.dhis.dxf2.webmessage.responses;
```
muilpp, line 30: isn't this one used in `TrackerImportController`?
If so, why do we want to move it to dxf2?
should it not be part of tracker?

muilpp (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/dxf2/webmessage/responses/TrackerJobWebMessageResponse.java
enricocolasante: I just moved it here because all the `WebMessageReponse` objects are in this package

teleivo: mhm, I also feel like this should be living with our code. or if possible not have it at all. this response looks so generic. is there another response with these fields we could reuse?

teleivo (approved):

PR #20637, opened 2025-04-14
fix: Move param validations into transfer service [DHIS2-19472]

muilpp:
I'm doing a bit of a refactor in the `TrackerOwnershipController`, where I:

- Move all validations into the service layer. So far, we had some validations in the controller and some in the service. The goal is to have all of them in the service layer. This way, we ensure that service API users won't experience unexpected behavior due to missing validations. There's a temporary exception when transferring a tracked entity, since we can't use the `trackedEntityService`, as explained [here](https://github.com/dhis2/dhis2-core/pull/20637/files#diff-e770e251f69cf4452ad2bdb19338b709c016ceffdc470b72b8b9a7ddf4ba4432R118).
- Use the `TrackerProgramService` to automatically perform ACL validations.
- Return an exception if no owner is found for the given tracked entity/program. Currently, when this happens, no changes are made but no exception is thrown, so the controller returns `ok("Ownership transferred")`, which is misleading.
- Introduce a new method in `TrackerProgramService` to get a tracker program by UID.
- Remove the deprecated `ou` param in the transfer endpoint.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -104,26 +113,28 @@ public DefaultTrackerOwnershipManager(
...
-  // -------------------------------------------------------------------------
-
   @Override
   @Transactional
-  public void transferOwnership(
-      TrackedEntity trackedEntity, Program program, OrganisationUnit orgUnit)
-      throws ForbiddenException {
-    if (trackedEntity == null || program == null || orgUnit == null) {
-      return;
+  // TODO(tracker) This method should accept a tracked entity UID instead. The problem is, we can't
```
enricocolasante, line 118: We are solving the same problem in 2 quite different ways.
In this method we are passing the trackedEntity from the service and we know that it is already completely validated, then we need to get the managed entity to be able to save a `ProgramOwnershipHistory`. I agree that we should get only the UID and do everything in the method.

In `grantTemporaryOwnership` we are passing the UID, and we are manually validating the tracked entity (tracked entity type, existence, etc), it would be better to use `TrackedEntityService` here as well in the controller and then just retrieve the managed entity from the `manager`, so we know all the proper validations are run.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -178,51 +199,29 @@ public void grantTemporaryOwnership(
...
-      TrackedEntity trackedEntity, Program program, UserDetails user) throws ForbiddenException {
-    if (program == null) {
-      throw new ForbiddenException(
-          "Temporary ownership not created. Program supplied does not exist.");
-    }
-
+  private void validateTrackedEntity(TrackedEntity trackedEntity, UserDetails user)
+      throws ForbiddenException {
     if (trackedEntity == null) {
       throw new ForbiddenException(
```
enricocolasante, line 205: If we use the `TrackedEntityService` we do not need to worry about this but this one should be more a `NOT_FOUND` exception

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/TrackerProgramService.java
```diff
@@ -52,9 +58,39 @@ public class TrackerProgramService {
...
   @Nonnull private final AclService aclService;
 
+  /**
+   * Returns the tracker program associated with the provided UID if it exists and is accessible to
+   * the current user.
+   */
+  @Transactional(readOnly = true)
+  public @Nonnull Program getTrackerProgram(@CheckForNull UID programUid)
+      throws BadRequestException, ForbiddenException {
+    if (programUid == null) {
```
enricocolasante, line 68: I would personally give this responsibility to the caller and annotate `programUid` as `@Nonnull`

enricocolasante (commented):
We need to use the same strategy in `transferOwnership` and `grantTemporaryOwnership` methods.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -150,16 +161,26 @@ public void transferOwnership(
...
     }
   }
 
   @Override
   @Transactional
-  public void grantTemporaryOwnership(
-      @Nonnull TrackedEntity trackedEntity, Program program, UserDetails user, String reason)
-      throws ForbiddenException {
-    validateGrantTemporaryOwnershipInputs(trackedEntity, program, user);
+  public void grantTemporaryOwnership(UID trackedEntityUid, UID programUid, String reason)
```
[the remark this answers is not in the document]
muilpp: this is the service API, so it could be null if supplied by the API user
but we can do the same as in `getTrackerProgram` and give this responsibility to whoever calls the method

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
```diff
@@ -104,26 +113,28 @@ public DefaultTrackerOwnershipManager(
...
+
+    if (orgUnit == null) {
+      throw new ForbiddenException(
+          "Tracked entity not transferred. Org unit supplied does not exist.");
     }
 
-    UserDetails currentUser = CurrentUserUtil.getCurrentUserDetails();
-
-    if (hasAccess(currentUser, trackedEntity, program)) {
+    if (hasAccess(getCurrentUserDetails(), trackedEntity, program)) {
```
ameenhere, line 133: Perhaps we can flip some of these if conditions to have easy to read flow (avoids chained or inner ifs)

Like

if (orgUnit == null)
throw

if(!hasAccess....)
throw

if (!programService.hasOrgUnit(program, orgUnit))
throw

if (any other validation failure)
throw

....normal success flow...

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
muilpp: as explained [here](https://github.com/dhis2/dhis2-core/pull/20637/files#r2046343406), I can't use the `TrackedEntityService`, but I agree with throwing a `NOT_FOUND`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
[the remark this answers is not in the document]
enricocolasante: API user can never supplies a null value, the request parameter is required.

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerOwnershipManager.java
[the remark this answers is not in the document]
muilpp: I meant if the service API is used directly, in that case it could be null, but that's now annotated with `@Nonnull` anyway

larshelge (approved):

PR #20652, opened 2025-04-15
backport: fix: Filter non-numeric data from outlier and min/max queries

jason-p-pickering:
https://github.com/dhis2/dhis2-core/pull/20599


jbee (approved):
I think you should update the regex with a length limit as discussed yesterday on slack.

PR #20673, opened 2025-04-16
fix: Add performance test for scheduled notifications[DHIS2-19452]

zubaira:
**PR Description**
**Problem:**

The scheduled notifications were facing performance issues during the nightly job run. The root causes were:

- Large JSON payloads are being stored in the database, leading to bloated rows and slow queries.
- Fetching all notification instances from the database and filtering them in memory, causing high memory usage.

**Solution:**

- Optimized the notification fetching by filtering notifications based on the scheduled date at the database level.
- Reduced redundant and cyclic data in the `ProgramNotificationTemplateSnapshot` to prevent bloating and improve query performance. So `OrganisationUnitSnapshot` was removed from `ProgramNotificationTemplateSnapshot` as we only need `UserGroup` and its members.

**Test:**

- 
- Initially created a test that highlighted the performance issue (it was designed to fail).
- After applying the fix, the test now passes, confirming the performance issues have been addressed.


teleivo (commented):
Can we please focus on getting https://github.com/dhis2/dhis2-core/pull/20576 merged and then this one. Reviewing two PRs with almost the same code at the same time is just wasting time 😬

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    getNotifications().forEach(n -> notificationSender.send(n, enrollment));
+  }
+
+  @Test
+  @DisplayName("Should fetch scheduled notifications within timeout")
+  void testScheduledNotificationShouldBeTriggeredBeforeTimeout() {
+    ProgramNotificationInstanceParam param =
+        ProgramNotificationInstanceParam.builder().scheduledAt(NOW).build();
+    List<ProgramNotificationInstance> instances =
```
teleivo, line 100: Separate your setup, method you test and your assertions by an empty line. This makes it easy to see what you are testing.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+    manager.save(trackedEntity);
+    Enrollment enrollment = createEnrollment(program, trackedEntity, organisationUnit);
+    manager.save(enrollment);
+
+    getNotifications().forEach(n -> notificationSender.send(n, enrollment));
+  }
+
+  @Test
+  @DisplayName("Should fetch scheduled notifications within timeout")
+  void testScheduledNotificationShouldBeTriggeredBeforeTimeout() {
```
teleivo, line 97: Name it `shouldReturnProgramNotificationInstancesForGivenDate` as that is what you are testing. The fact that this would have timed out in the past is secondary. There is no need to mention that in our code. One can get that context from the issue/PR.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+    ProgramNotificationInstanceParam param =
+        ProgramNotificationInstanceParam.builder().scheduledAt(NOW).build();
+    List<ProgramNotificationInstance> instances =
+        programNotificationInstanceService.getProgramNotificationInstances(param);
+    assertEquals(
+        EXPECTED_NOTIFICATIONS,
```
teleivo, line 103: Assert that the data you get back matches your expectations in terms of the setup. Here I would guess it has something to do with a date being relative to your setup date `NOW`. Assertions like this can hide bugs and are annoying when changing code/setup. I also have no clue why its 20. If I change some code/setup and this fails with actual being 18, how would I know if the returned notifications are still correct?

Make sure that your test is not flaky when comparing dates! This is a classic when using `now()`.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+    pnt.setNotificationRecipient(ProgramNotificationRecipient.USER_GROUP);
+    pnt.setNotificationTrigger(NotificationTrigger.PROGRAM_RULE);
+    pnt.setAutoFields();
+    manager.save(pnt);
+    return pnt.getUid();
+  }
+
+  private void createAndAddUsersToUserGroupAndOrgUnit() {
+    Set<User> users = new HashSet<>();
+    for (int i = 1; i <= TEST_USER_COUNT; i++) {
```
teleivo, line 124: How did you come up with the 60 users? Why is that loop necessary?

teleivo:
Will this change leave unused data in the users DB? If so is that data duplicated? If so we should remove it. If not the users should decide.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
zubaira: The number is set to 20 because fetching a single notification instance wouldn’t reveal any noticeable performance issues, even without the changes introduced in this PR. A slightly higher number ensures that potential performance bottlenecks are more likely to surface.
Moreover, I will add assertions on date as well.

zubaira:
> Will this change leave unused data in the users DB? If so is that data duplicated? If so we should remove it. If not the users should decide. We had a case like this where we left data that only the users would know if its still valuable to them. That would need a note in the release notes.

Yes, it might (though not necessarily) leave some data in the database, but it won't result in any duplicated entries.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
teleivo: You are saying is unlikely. Did you test your implementation against the code still having the orgUnit snapshot? As that should give you the minimum amount of data needed to show we had a problem. Then you take that data/test and fix the issue.

teleivo:
> > Will this change leave unused data in the users DB? If so is that data duplicated? If so we should remove it. If not the users should decide. We had a case like this where we left data that only the users would know if its still valuable to them. That would need a note in the release notes.
> 
> Yes, it might (though not necessarily) leave some data in the database, but it won't result in any duplicated entries.

I mean is that data a denormalized version of data in other tables? If so you should delete it as that wastes spaces on users servers.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
zubaira: Yes, I created this test before implementing the fix — that’s how I arrived at these numbers.

zubaira:
> > > Will this change leave unused data in the users DB? If so is that data duplicated? If so we should remove it. If not the users should decide. We had a case like this where we left data that only the users would know if its still valuable to them. That would need a note in the release notes.
> > 
> > 
> > Yes, it might (though not necessarily) leave some data in the database, but it won't result in any duplicated entries.
> 
> I mean is that data a denormalized version of data in other tables? If so you should delete it as that wastes spaces on users servers.

No,  the data resides only in one table. It is not a denormalized copy from other tables.

teleivo (commented):
Please change your assertion see https://github.com/dhis2/dhis2-core/pull/20673/files#r2068501215

zubaira:
> Please change your assertion see https://github.com/dhis2/dhis2-core/pull/20673/files#r2068501215

fixed

ameenhere (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    List<ProgramNotificationInstance> instances =
+        programNotificationInstanceService.getProgramNotificationInstances(param);
+
+    assertEquals(
+        expectedNotifications,
+        instances.size(),
+        "Expected " + expectedNotifications + " notification instances, but got " + instances);
+
+    instances.forEach(instance -> assertEquals(DATE, instance.getScheduledAt()));
```
teleivo, line 112: Try failing your test by changing the implementation. Would the assertion output help you quickly fix the implementation? Imagine this happens only on ci.

teleivo (approved):

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -107,9 +107,24 @@ void shouldReturnProgramNotificationInstancesForGivenDate() {
     assertEquals(
         expectedNotifications,
         instances.size(),
-        "Expected " + expectedNotifications + " notification instances, but got " + instances);
+        () ->
+            String.format(
+                "Expected %d notification instances, but got %d.\nNo instances were found for scheduledAt: %s."
```
teleivo, line 112: `No instances were found for scheduledAt: %s`

Is that true if you expect 20 and get 19 or so?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/program/notification/ProgramNotificationInstanceServiceTest.java
```diff
@@ -107,9 +107,24 @@ void shouldReturnProgramNotificationInstancesForGivenDate() {
...
 
-    instances.forEach(instance -> assertEquals(DATE, instance.getScheduledAt()));
+    instances.forEach(
+        instance ->
+            assertEquals(
+                DATE,
+                instance.getScheduledAt(),
+                () ->
+                    String.format(
+                        "Expected scheduledAt=%s but got %s in " + instance,
```
teleivo, line 125: either use string concatenation or `String.format`

teleivo (approved):

PR #20688, opened 2025-04-17
task: Get TEAs by program in tracker aggregate [DHIS2-19482]

muilpp:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
```diff
@@ -173,10 +172,9 @@ List<TrackedEntityAttribute> getTrackedEntityAttributesByDisplayOnVisitSchedule(
...
   /**
-   * Get all {@link TrackedEntityAttribute} grouped by {@link Program}
+   * Fetches all {@link TrackedEntityAttribute} UIDs of the given {@link Program}
    *
-   * @return a Map, where the key is the {@link Program} and the values is a Set of {@link
-   *     TrackedEntityAttribute} associated to the {@link Program} in the key
+   * @return a Set of {@link TrackedEntityAttribute} UIDs
    */
-  Map<Program, Set<TrackedEntityAttribute>> getTrackedEntityAttributesByProgram();
+  Set<String> getTrackedEntityAttributesInProgram(Program program);
```
teleivo, line 179: Please add `@Nonnull/@CheckForNull`

teleivo (commented):
I think this should also go into 2.41.1 as this is an easy change that removes wasted compute.

In what other places do we have the logic of returning teta + ptea in the context of a program or only teta without a program? As one could argue that the change you are making could move entirely into the attribute store in something like `getTrackedEntityAttributes(@CheckForNull program)`.

teleivo (approved):

muilpp:
> I think this should also go into 2.41.1 as this is an easy change that removes wasted compute.
> 
> In what other places do we have the logic of returning teta + ptea in the context of a program or only teta without a program? As one could argue that the change you are making could move entirely into the attribute store in something like `getTrackedEntityAttributes(@CheckForNull program)`.

Into 2.42.1? if so, I agree.

Do you mean in [this](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/attribute/AttributeStore.java) store? I refactored the one in `TrackedEntityAttributeService`, as there are other methods getting TEAs, so it made sense to me to have it there.

PR #20690, opened 2025-04-17
fix: Limit Postgres double pattern to 1e307 (#20690)

jason-p-pickering:
Adjust Postgres double precision pattern based on dev team feedback

Limit accepted numbers to a maximum of approximately 1e307, which is below the theoretical PostgreSQL double precision limit (~1.7976931348623158e308), providing a practical safety margin.

Allow numeric strings with:

- A decimal but no fractional part (e.g., "42.")
- A fractional part without a leading zero (e.g., ".42")
- A leading + sign for positive numbers
- Large decimal portions are tolerated since PostgreSQL will truncate to 15–16 significant digits automatically.

Helps avoid SQL runtime casting errors while not limiting significantly numeric strings which can be cast to doubles, at least for realistic DHIS2 use cases.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
@@ -31,6 +31,7 @@ package org.hisp.dhis.common;
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toMap;
+import static java.util.stream.Collectors.toSet;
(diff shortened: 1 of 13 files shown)


jason-p-pickering:
@jbee I have reworked this now a bit based on feedback from your review. Please let me know if this captures it.

jason-p-pickering:
Results of SQL profiling
1) This PR https://explain.depesz.com/s/nFwe#html (2.8 seconds)
2) Current master https://explain.depesz.com/s/LU9c (5.7 seconds)

jbee (approved):

david-mackessy (approved):

PR #20691, opened 2025-04-17
chore: use named placeholders to compose SQL

jbee:


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/gist/GistBuilder.java
```diff
@@ -880,85 +912,95 @@ && isPersistentReferenceField(path.get(0))
...
-        fieldTemplate = "lower(%s)";
+        template = "lower(${property})";
       }
     }
-    str.append(String.format(fieldTemplate, field));
-    str.append(" ").append(createOperatorLeftSideHQL(operator));
-    if (!operator.isUnary()) {
-      str.append(" :f_").append(index).append(createOperatorRightSideHQL(operator));
-    }
-    return str.toString();
```
jbee, line 924: FYI: This was super hard to understand because the HQL for the operator and their arguments was so split apart. That is why I moved the SQL/HQL of just the operator itself into the `Operator` enum but everything beyond is now in the template and the variables.

david-mackessy (approved):

vietnguyen (approved):

PR #20692, opened 2025-04-17
fix: respect order of PI mapping [DHIS2-19483]

jbee:


vietnguyen24 (approved):

david-mackessy (commented):
A test would be nice. If it's important, at least we might be notified if it breaks.

jbee:
> A test would be nice. If it's important, at least we might be notified if it breaks.

I was just helping out, don't really know much about it where I could write a meaningful test.

vietnguyen (approved):

jimgrace (changes requested):
Hi @jbee, thank you for the thought.

However, nothing is gained by making this a List instead of a Set. The Program defines a bunch of disaggregation mappings that are available to any of its Program Indicators. These IDs in the program indicator tell which of these mappings the PI will use. If a program indicator has a non-default Category Combo, there needs to be an ID in this list for each Category in the Category Combo to say which mapping to use. Likewise, if a PI has a non-defualt Attribute Combo, there needs to be an ID in this list for each Category in the Attribute Combo to say which mapping to use.

**Frontend:** In the frontend, the UI allows the user to select the Category Combo and Attribute Combo used for the PI. If the user selects a non-default combo for either, this list must contain mappings for the categories in the combo. The frontend has to go through this collection to find the mapping IDs, regardless of the order.

**Backend:** In the backend, each mapping ID tells how a certain category value will be found from the data (using the category option filters defined in the program for this mapping ID). The mapping IDs are only referenced to find out how to map each category. Again, the order doesn't matter.

**Performance:** The number of IDs referenced by a program indicator is expected to be small; it's the sum of the number of categories in the Category Combo (if not default) plus the number of categories in the Attribute Combo (if not default). Each combo is most likely to have 1 or 2 categories, but there could be more, the likelihood decreasing rapidly as the number of categories grows. As such, I don't expect any noticeable storage overhead or performance difference between using a Set or List. Using a Set just means that the jsonb structure doesn't have to preserve the order, so I think that is more appropriate.

But thanks anyway!

PR #20699, opened 2025-04-18
chore: Remove unused methods from TrackedEntityAttributeValueService [DHIS2-17679]

enricocolasante:
Remove `getTrackedEntityAttributeValue` methods only used in tests


muilpp (approved):

teleivo (approved):
How does this fit into the bug fix? to understand if there is more that is wrong other than us not validating certain value types 😅

enricocolasante:
> How does this fit into the bug fix? to understand if there is more that is wrong other than us not validating certain value types 😅

The goal here was to simplify `DefaultTrackedEntityAttributeValueService` and then manage to remove `addTrackedEntityAttributeValue` that is only used in tests and it is doing more harm than good because we have a lot of tests that are testing not production code.
In next PR I will arrive to the point and then we can decide how deep to go on the fix.

PR #20700, opened 2025-04-18
fix: Move TE ownership validation to store [DHIS2-19485]

muilpp:
I'm moving the TE filter we run after hitting the database into the store, so we no longer break pagination on `/trackedEntities`


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,59 +593,97 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
     StringBuilder orgUnits = new StringBuilder();
 
-    handleOrganisationUnits(params);
+    UserDetails userDetails = getCurrentUserDetails();
+    Set<OrganisationUnit> effectiveSearchOrgUnits =
+        getOrgUnitsFromUids(userDetails.getUserEffectiveSearchOrgUnitIds());
+    Set<OrganisationUnit> captureScopeOrgUnits =
+        getOrgUnitsFromUids(userDetails.getUserOrgUnitIds());
+
+    handleOrganisationUnits(params, effectiveSearchOrgUnits, captureScopeOrgUnits);
```
teleivo, line 627: inline this as this indirection only makes the logic harder to follow

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,59 +593,97 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
+      params.setOrgUnitMode(DESCENDANTS);
     }
   }
 
+  private Set<OrganisationUnit> getOrgUnitsFromUids(Set<String> uids) {
+    return new HashSet<>(organisationUnitStore.getByUid(uids));
+  }
+
+  private String[] getOrgUnitsPathArray(Set<OrganisationUnit> orgUnits) {
+    return orgUnits.stream().map(ou -> ou.getStoredPath() + "%").toArray(String[]::new);
```
teleivo, line 684: Can like wildcard characters `%` and `_` and its default escape character `\` be part of the orgUnit path? If so we need to escape them https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/SqlUtils.java#L124 (I need to escape the `\` in this function).

https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,59 +593,97 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
-      orgUnits.append(getSelectedQuery(params));
+    if (params.isOrganisationUnitMode(ALL) || getCurrentUserDetails().isSuper()) {
+      return orgUnits.toString();
     }
 
+    sqlParams.addValue("effectiveSearchScopePaths", getOrgUnitsPathArray(effectiveSearchOrgUnits));
+    sqlParams.addValue("captureScopePaths", getOrgUnitsPathArray(captureScopeOrgUnits));
+
+    orgUnits.append(
+        "and ((P.accesslevel in ('OPEN', 'AUDITED') and (OU.path like any (:effectiveSearchScopePaths))) ");
```
teleivo, line 652: is `like null` a possibility

> If the array expression yields a null array, the result of ANY will be null. If the left-hand expression yields null, the result of ANY is ordinarily null (though a non-strict comparison operator could possibly yield a different result). Also, if the right-hand array contains any null elements and no true comparison result is obtained, the result of ANY will be null, not false (again, assuming a strict comparison operator)
https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-ANY-SOME

and would that be handled correctly if it is?

Is there an index on the org unit path? Looks like we are doing a prefix search on the orgUnit path. Do we know any ballpark numbers on users scope paths?

teleivo (commented):
I would say this is actually a bug and not a task 😅 
which also has no test so you should create one.

muilpp:
> I would say this is actually a bug and not a task 😅 which also has no test so you should create one.

yes, I realized that later, I'm writing the test, but it's a bit painful to write ACL tests with the current setup, I'll bring that up on Slack.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
muilpp: `like null` can't happen here, because it's mandatory for the user to have at least one org unit in their capture scope. So, `effectiveSearchScopePaths` should always return at least one org unit.

There is an index on the org unit path. However, in my tests, the planner is not using it. I'm not sure if that's because there aren't enough results to make it worthwhile, or because of the combination of the `LIKE` and `OR` operators.

Regarding the number of scope paths a user can have, they are assigned in the Maintenance app. There's no "select all" option or anything similar. If a user needs access to the entire country, you just select the root org unit, in our case, that would be Sierra Leone. With that, the user has access to all org units in the system but only has one org unit path. The same goes for districts or any lower level org unit.
My point is, even though it’s theoretically possible to manually select and assign all available org units to a user, it doesn’t really make sense, since having access to a parent automatically grants access to all its children.

One more thing I want to mention, this query was already being used in the tracker aggregate. I just removed it from there and am now reusing part of it in the hibernate store.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,57 +593,81 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
+   * <p>If a program is specified, the join is based on program ownership (PO). If no program is
+   * specified, the join is based either on program ownership or the tracked entity's registering
+   * unit.
+   *
+   * <p>The specific JOIN conditions depend on the {@code ouMode}:
+   *
+   * <ul>
+   *   <li>{@code DESCENDANTS} – matches organisation units using the org unit's PATH
+   *   <li>{@code CHILDREN} – matches the org unit's PATH or any of its immediate children
+   *   <li>{@code SELECTED} – matches the specified org unit ID directly
```
enricocolasante, line 607:
```suggestion
   *   <li>{@code SELECTED} – matches the specified org unit UID directly
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
enricocolasante, line 1: We should somehow remove the deprecated call 
`return ids.withFilteredItems(trackedEntities);` in `findTrackedEntities` method with `withMappedItems`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,57 +593,81 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
+      }
     }
 
-    if (params.isOrganisationUnitMode(OrganisationUnitSelectionMode.DESCENDANTS)) {
-      orgUnits.append(getDescendantsQuery(params));
-    } else if (params.isOrganisationUnitMode(OrganisationUnitSelectionMode.CHILDREN)) {
-      orgUnits.append(getChildrenQuery(params));
-    } else if (params.isOrganisationUnitMode(OrganisationUnitSelectionMode.SELECTED)) {
-      orgUnits.append(getSelectedQuery(params));
+    if (params.isOrganisationUnitMode(ALL) || getCurrentUserDetails().isSuper()) {
```
muilpp, line 650: We had this validation in the old code to check if the user has super privileges, which I believe is incorrect.
As far as I know, we only skip sharing validations when the user is a superuser, but we never skip ownership validations, unless the org unit mode is explicitly set to `ALL`, so I should remove it from here.
@ameenhere?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
muilpp: isn't this something to do in a different ticket? this deprecated method is also used in relationship and enrollments

enricocolasante: It depends.
If this ticket is fixing pagination for tracked entities endpoint, then it would make sense to get rid of that call.
We need to remove not only because it is deprecated, but because that call means that the results in the page can be filtered, hence break pagination.
If we are planning to work on https://dhis2.atlassian.net/browse/DHIS2-16493 soon, maybe we can fix it there.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -588,57 +593,81 @@ private String getFromSubQueryJoinProgramOwnerConditions(TrackedEntityQueryParam
...
     StringBuilder orgUnits = new StringBuilder();
 
-    handleOrganisationUnits(params);
+    UserDetails userDetails = getCurrentUserDetails();
+    Set<OrganisationUnit> effectiveSearchOrgUnits =
+        getOrgUnitsFromUids(userDetails.getUserEffectiveSearchOrgUnitIds());
+    Set<OrganisationUnit> captureScopeOrgUnits =
+        getOrgUnitsFromUids(userDetails.getUserOrgUnitIds());
+
+    if (params.isOrganisationUnitMode(ACCESSIBLE)) {
```
muilpp, line 627: This two checks are no longer necessary, they will be implicitly filtered when validating the ownership in the new clause at the end of this method.
However, to remove it, I need first the answer to [this question](https://github.com/dhis2/dhis2-core/pull/20700/files#r2054194412), I'll

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -240,10 +236,11 @@ public Page<TrackedEntity> findTrackedEntities(
...
         trackedEntityStore.getTrackedEntityIds(queryParams, pageParams);
 
-    List<TrackedEntity> trackedEntities =
-        findTrackedEntities(ids.getItems(), operationParams, queryParams, user);
+    Map<String, TrackedEntity> trackedEntities =
+        findTrackedEntities(ids.getItems(), operationParams, queryParams, user).stream()
+            .collect(Collectors.toMap(TrackedEntity::getUid, Function.identity()));
 
-    return ids.withFilteredItems(trackedEntities);
+    return ids.withMappedItems(id -> trackedEntities.get(id.uid()));
```
enricocolasante, line 243: Now that we are mapping items, can we remove the ordering from the aggregate query?
In my head it makes sense that pagination and order is defined in the ids query and then the aggregate is just getting more data and do not need to worry on order.

So it would be removing the call to `applySortOrder` [here](https://github.com/dhis2/dhis2-core/blob/ba958355873f0eadac98a1824a02371a688792e4/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/aggregates/TrackedEntityStore.java#L86)

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
muilpp: There's a bit of confusion when it comes to this, since an admin user is not a super user.
A superuser is the one that has the authority `ALL`, and that's usually the system user, the admin user has other privileges, but not this one. 
That means, in our current logic, we would skip the ownership validations for the system user, but not for admin users.

Is that correct?
If it is, then we are saying, since you are a superuser, we assume you have capture scope access to all org units in the system, so we skip the ownership validations. Same goes for sharing validations. Is that it?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
enricocolasante: We could still map from ids to tracked entity and make findTrackedEntities unordered.
It should be faster to map than order in the DB.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/DefaultTrackedEntityService.java
```diff
@@ -271,34 +268,11 @@ private List<TrackedEntity> findTrackedEntities(
...
-    if (queryParams.hasEnrolledInTrackerProgram()) {
-      return te ->
-          skipOwnershipCheck
-              || ownershipAccessManager.hasAccess(
-                  user, te, queryParams.getEnrolledInTrackerProgram());
-    }
-
-    return te ->
-        queryParams.getAccessibleTrackerPrograms().stream()
-            .filter(
```
ameenhere, line 298: Do we have an existing test to reaches this filtering based on "any" program ownership?

ameenhere (approved):
Nice work @muilpp ! 
Looks good to me. Let us keep a close eye on things even after merge. 👍

PR #20701, opened 2025-04-18
fix: Properly validate attribute value type [DHIS2-17679]

enricocolasante:
Clean up and extract to `ValidationUtils` value type validation in attribute and data value validators.
- Removed code `E1085` that was overlapping with already existing `E1007` for attributes and `E1302` for data values. Errors with code `E1085` were always happening together with other errors and they were not providing any useful information to the user.
- Remove code `E1084` that was treating attribute and data values of type file resources in a different way from any other value type.
- Fix the error code for org unit value type for data values that was using the attribute error code `E1007`. Now it is using the correct `E1032` code.
-


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -245,18 +250,82 @@ public static <T extends ValueTypedDimensionalItemObject> void validateOptionSet
...
+    if (valueType.isFile()) {
+      validateFileNotAlreadyAssigned(reporter, bundle, dto, value);
+    }
+  }
+
+  private static String validateImage(TrackerBundle bundle, String value) {
+    FileResource fileResource = bundle.getPreheat().get(FileResource.class, value);
+
+    return Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
+        ? "File resource with uid '" + value + "' is not a valid image"
```
teleivo, line 303: We could be more specific saying what format this file resource has and what the valid ones are. That helps the user fix their issue 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -245,18 +250,82 @@ public static <T extends ValueTypedDimensionalItemObject> void validateOptionSet
...
+      String value) {
+
+    FileResource fileResource = bundle.getPreheat().get(FileResource.class, value);
+
+    if (bundle.getStrategy(trackerDto).isCreate()) {
+      reporter.addErrorIf(
+          () -> fileResource != null && fileResource.isAssigned(), trackerDto, E1009, value);
+    }
+
+    if (bundle.getStrategy(trackerDto).isUpdate()) {
```
teleivo, line 320: can the strategy be create and update? if not use an else if or return

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -245,18 +250,82 @@ public static <T extends ValueTypedDimensionalItemObject> void validateOptionSet
...
+      reporter.addErrorIf(
+          () ->
+              fileResource != null
+                  && fileResource.getFileResourceOwner() != null
+                  && !fileResource.getFileResourceOwner().equals(trackerDto.getUid().getValue()),
```
teleivo, line 325: Do we do this kind of comparison often? Not sure if its worth adding a `boolean UID.isEqualTo(@CheckForNull UidObject uid)`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/AttributeValidator.java
```diff
@@ -103,7 +101,7 @@ public void validate(Reporter reporter, TrackerBundle bundle, Enrollment enrollm
 
         attributeValueMap.put(attribute.getAttribute(), attribute.getValue());
         validateAttributeValue(reporter, enrollment, teAttribute, attribute.getValue());
```
teleivo, line 103: Can we get rid of inheritance and move that validation from `validateAttributeValue` into `validateValueType` which actually also validates the value

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/service/attribute/TrackedAttributeValidationService.java
```diff
@@ -1,164 +0,0 @@
...
-                ValueType.NUMBER,
-                v -> !MathUtils.isNumeric(v),
-                " '%s' is not a valid numeric type for attribute %s "),
-            new ValueTypeValidationFunction(
-                ValueType.BOOLEAN,
-                v -> !MathUtils.isBool(v),
-                " '%s' is not a valid boolean type for attribute %s "),
-            new ValueTypeValidationFunction(
-                ValueType.DATE,
-                v -> DateUtils.parseDate(v) == null,
```
teleivo, line 88: where did these validations go?

teleivo (approved):
Did you check the github org for usages of the error codes you removed? Not sure if this is a breaking change or not 🤷🏻 You will also need to remove them from https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#webapi_tracker_error_codes

Would be good to know if we consider these codes as part of our API or not.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
enricocolasante: I think it would make sense but in this specific case we are not comparing on the uid of the object but on another field.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/AttributeValidator.java
enricocolasante: I can remove the inheritance moving `validateAttributeValue` into `ValidationUtils` but I cannot move it into `validateValueType` as it is shared with data values validation as well and `validateAttributeValue` is specific for attributes

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/service/attribute/TrackedAttributeValidationService.java
enricocolasante: These vaidations were completely overlapping the validation of `ValueType` done [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/ValidationUtils.java#L515-L539)

enricocolasante:
> Did you check the github org for usages of the error codes you removed? Not sure if this is a breaking change or not 🤷🏻 You will also need to remove them from https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/tracker.html#webapi_tracker_error_codes
> 
> Would be good to know if we consider these codes as part of our API or not.

On our organization it seems that only Android has some knowledge of the error codes.

Here we have 3 different cases:
1. `E1085` was completely useless as it is contained in `E1007`
2. `E1084` was treated as a special case of `E1007` but it was removing some context, so the user wouldn't know that the error was happening on an attribute.
3. Error on data value of type `OrgUnit` was thrown as an error on an attribute.

In my head, 3 is a bug.
1 and 2 can be considered breaking changes but I am not sure if we have a way of going through the process of deprecation like we should do.
I would say that codes are part of the API and any change should be properly highlighted in the release notes.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/AttributeValidator.java
teleivo: sounds good 😄

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/service/attribute/TrackedAttributeValidationService.java
teleivo: but where do we call these?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -300,7 +301,13 @@ private static String validateImage(TrackerBundle bundle, String value) {
     FileResource fileResource = bundle.getPreheat().get(FileResource.class, value);
 
     return Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
-        ? "File resource with uid '" + value + "' is not a valid image"
+        ? "File resource with uid '"
+            + value
+            + "' is using invalid image format "
+            + fileResource.getFormat()
+            + ". Valid formats are: ("
+            + StringUtils.join(Constant.VALID_IMAGE_FORMATS)
```
teleivo, line 309: AFAIK without passing a separator you'll turn `"a","b"` into `"ab"` which is probably not what we want.

Remove the `()` from the sentence.

teleivo (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/service/attribute/TrackedAttributeValidationService.java
enricocolasante: [Here](https://github.com/dhis2/dhis2-core/blob/7ecb3576230c66886c48d23042b92fb409891061/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java#L262)
We call the common value type validation that is the same for the whole system and it is covering the validations that were deleted.

ameenhere (approved):

PR #20709, opened 2025-04-22
feat: Add support for identifying invalid numeric outliers

jason-p-pickering:
**Summary**

This PR introduces a new outlier detection algorithm type: INVALID_NUMERIC.

It enables the system to detect and report invalid numeric strings that cannot be safely cast to PostgreSQL double precision values, which otherwise could cause runtime SQL exceptions.

This fills a gap in the original outlier detection logic, which assumed all value fields were castable.

**Key changes:** 

- Extend OutlierDetectionAlgorithm enum to support INVALID_NUMERIC.
- Add NumericPatternSqlStatementProcessor to generate SQL filtering out non-standard numeric values
- Add NumericPatternOutlierDetectionManager for mapping and returning rawValue in API responses
- Update DefaultOutlierDetectionService to dispatch based on algorithm type
- Update API to allow clients to call:GET /api/outlierDetection?algorithm=INVALID_NUMERIC with the normal parameters
- Preserve all existing outlier detection behavior (Z_SCORE, MIN_MAX, MOD_Z_SCORE) with no changes to ensure backwards compatibility. 

Example returned JSON for http://localhost:9090/api/43/outlierDetection?ds=lyLU2wR22tC&ou=ImspTQPwCqd&startDate=2024-01-01&endDate=2025-04-22&algorithm=INVALID_NUMERIC&maxResults=500

```
{
  "de": "tZLjsE0VXrL",
  "deName": "New on AZT 300mg / 3TC 150mg + EFV 600mg",
  "pe": "202401",
  "ou": "djMCTPYvltl",
  "ouName": "Govt. Hosp. Kenema",
  "rawValue": "abc123",
  "followup": false
}
```

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseAnalyticalObject.java
@@ -31,6 +31,7 @@ package org.hisp.dhis.common;
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toMap;
+import static java.util.stream.Collectors.toSet;
 import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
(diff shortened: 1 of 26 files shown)


jbee (approved):
LGTM :+1:

david-mackessy (approved):

PR #20715, opened 2025-04-22
fix: Doris subquery on event table population (DHIS2-19489)

luciano-fiandesio:
## Summary

This PR rewrites the legend‑lookup inside `INSERT … SELECT` of the Analytics Event population process, so that the correlation with the outer query is done in `WHERE` instead of in the `JOIN`. 
The change keeps exactly the same business logic, runs unchanged on PostgreSQL, and removes the “access outer query’s column in join is not supported” error on Apache Doris.

This change is affecting the select statement that is generated when a Program has one or more Tracked Entity Attribute with a Legendset.

### What was wrong

- Doris supports scalar sub‑queries in the SELECT list only when the correlation is in a filter predicate; references in an ON clause trigger errCode 2. 
- The original query placed `av.trackedentityid = en.trackedentityid` inside the join between `maplegend` and `trackedentityattributevalue`, so Doris failed to compile while PostgreSQL succeeded.

### Fix

1. Moves the outer reference to the `WHERE` clause inside the scalar sub‑query
2. Adds `LIMIT 1` to guarantee the sub‑query remains scalar on both engines

```sql
-- BEFORE
SELECT
    l.uid
FROM pg_dhis.public.maplegend l
JOIN trackedentityattributevalue av
  ON l.startvalue <= CASE WHEN value REGEXP '...' THEN CAST(value AS bigint) END
  AND l.endvalue  > CASE WHEN value REGEXP '...' THEN CAST(value AS bigint) END
  AND l.maplegendsetid = 41785
  AND av.trackedentityid = en.trackedentityid   -- 💀  outer ref in ON
  AND av.trackedentityattributeid = 41942
  AND value REGEXP '...';

-- AFTER
SELECT  l.uid
FROM    pg_dhis.public.maplegend           l
JOIN    trackedentityattributevalue        av
       ON av.trackedentityattributeid = 41942
      AND av.value REGEXP '^-?[0-9]+(\\.[0-9]+)?$'
      AND l.maplegendsetid = 41785
      AND l.startvalue <= CAST(av.value AS bigint)
      AND l.endvalue   > CAST(av.value AS bigint)
WHERE   av.trackedentityid = en.trackedentityid  -- 🐱  correlation moved here
LIMIT   1;
```

### DIFF
--- a/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
+++ b/dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
@@ -631,9 +631,7 @@ public class JdbcEventAnalyticsTableManager extends AbstractEventJdbcTableManage
       return List.of();
     }
 
-    String columnExpression =
-        sqlBuilder.cast(
-            "av.value", NUMERIC); // getColumnExpression(attribute.getValueType(), "value");
+    String columnExpression = sqlBuilder.cast("av.value", NUMERIC);
     String numericClause = getNumericClause("value");
 
     String query =


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
```diff
@@ -632,14 +632,19 @@ private List<AnalyticsTableColumn> getColumnForAttributeWithLegendSet(
...
+
     String query =
         """
-        \s(select l.uid from ${maplegend} l \
-        inner join trackedentityattributevalue av on l.startvalue <= ${selectClause} \
-        and l.endvalue > ${selectClause} \
-        and l.maplegendsetid=${legendSetId} \
-        and av.trackedentityid=en.trackedentityid \
-        and av.trackedentityattributeid=${attributeId} ${numericClause}) as ${column}""";
+            \s(select l.uid \
```
maikelarabori, line 638: Shouldn't we keep `${selectClause}` in place?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/table/JdbcEventAnalyticsTableManager.java
```diff
@@ -632,14 +632,19 @@ private List<AnalyticsTableColumn> getColumnForAttributeWithLegendSet(
...
+              FROM   ${maplegend} l \
```
maikelarabori, line 639: Please, use lowercase for SQL.
I remember someone mentioning using uppercase for SQL, but I think we should still stick to lowercase (even though I prefer uppercase too)

maikelarabori (approved):

vietnguyen (approved):

PR #20718, opened 2025-04-22
chore: simplify fields propagation in tracker exporter DHIS2-19137

teleivo:
## Background

Users can specify what `fields` should be included in the JSON. The `FieldFilterService` then removes what should not be included

https://github.com/dhis2/dhis2-core/blob/a7c9affd49bbb02db3b48c2597e377f098bfb779/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/RequestHandler.java#L125

We use that knowledge to not fetch potentially expensive collections like `attributes`, `relationships`, `events`, ... The `FieldFilterService` has

https://github.com/dhis2/dhis2-core/blob/a7c9affd49bbb02db3b48c2597e377f098bfb779/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/FieldFilterServiceTest.java#L66

which we use to ask: is this field included in the JSON?

We captured that knowledge in a cumbersome hierarchy of ~ 5 classes + 4 helper classes 😅 which are now 3. 

## Changes

Simplify the hierarchy by following the existing hierarchy of our entities `TE -> EN -> EV`:

`TrackedEntityFields -> EnrollmentFields -> EventFields`

and using composition.

These fields classes can be constructed using builders for our internal Java API use. They can also be constructed using a predicate that given a field name says if the field is included or not. The `FieldFilterService` is wrapped in a predicate in the controllers. The fields classes are thus simple value classes.

The composition works by prefixing the field name with the parents path. Each class is only concerned with its fields and delegates to its children.

Take the example `/trackedEntities?fields=enrollments[events[relationships]]`:
* Asking the `FieldFilterService` using the dot path `enrollments`, `enrollments.events` or `enrollments.events.relationships` returns true
* `TrackedEntityFields` will thus include `enrollments`
* `TrackedEntityFields` will set its `EnrollmentFields` passing in its predicate with a new one prefixing paths with `enrollments.`. So internally `EnrollmentFields` asks is `events` included? which is actually `enrollments.events`
* The same happens for `EventFields` inside of `EnrollmentFields`

## API

Right now you have to specify the fields of a nested entity which itself has a fields class:

```java
TrackedEntityFields.builder()
  .includeRelationships()
  .includeEnrollments(EnrollmentFields.all())
  .build()
```

We could default to all fields with `.includeEnrollments()` but I'd rather we be specific so the user is aware of the cost.

## Next PR

Implement `RelationshipFields` so we only fetch what's necessary. Think of `fields=events[relationships[relationship]`.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityFields.java
```diff
@@ -0,0 +1,139 @@
...
+    return new TrackedEntityFields(includesFields, pathSeparator);
+  }
+
+  /** Use this if you do not want fields to be exported. */
+  public static TrackedEntityFields none() {
+    // the path separator does not matter as the predicate returns false regardless of the path
+    return new TrackedEntityFields(f -> false, "x");
+  }
+
+  /** Use this if you do want fields to be exported. This is potentially expensive! */
```
enricocolasante, line 94:
```suggestion
  /** Use this if you do want all fields to be exported. This is potentially expensive! */
```

ameenhere (approved):

PR #20738, opened 2025-04-23
chore: Query Order decoupled from Schema

jbee:
### Summary
`Order` should be a value object that describes the ordering of a query but not be involved in the processing and execution of the query. The burden of providing a `Property` to order by instead of the property's name should not be put on the caller of the query. Not just is that inconvenient but it also pulls logic and implementation details into the web layer where it should not be.
Therefore the property to order by is now simply stated by its property name as a `String`.
This allowed simplification in many regards but required to move the `Property` lookup into the lower layers of query execution.
I also moved some utilities that were used in user queries to the user store since that is the only place that code was used. 

### Validation
I went light on trying to make validation be consistent (again? not sure it was). For this PR I throw `IllegalArgumentException` when a name of a non-existing property is used for ordering. I want to do a followup PR that does validation properly as a step in the processing and uses `BadRequestException`. I didn't go that far in this PR to keep the change small.

### Automatic Testing
I adjusted most integration tests that used order in some form.
A couple of unit tests got removed since the code is inaccessible now as it requires the `Schema` lookup and `Property` resolution to take place. I think this is indirectly covered by integration tests that use ordering.


david-mackessy (approved):
👍

PR #20739, opened 2025-04-23
chore: Remove API version code [DHIS2-19486]

david-mackessy:
## Summary
See [DHIS2-18098](https://dhis2.atlassian.net/browse/DHIS2-18098) & [DHIS2-19486](https://dhis2.atlassian.net/browse/DHIS2-19486) for details about why this code is being removed.
Change includes:
- Removing all code related to API version
- All controllers using `@ApiVersion` have been updated by removing the annotation
- Some endpoints had the API version as an optional request param, this param has been removed
- Keeping behaviour of allowing to call the API with a version number e.g. `/api/42/dataSets`
- supports versions 28-43

I've tagged the Tracker & Analytics teams for review, to share the changes & in case any potential issues were missed that we're not aware of. Please tag anyone else you think may need to know about this change.

## Changes
- Add filter for requests that match a regex (API endpoint with a version)
  - examples:
    - `/api/42/icons` will match
    - `/dashboard` will not
    - `/42/dashboard` will not
  - filter will re-route any versioned request match to its non-versioned endpoint e.g. `/api/42/icons` -> `/api/icons`
- removes custom request mappings for versioned endpoints
  - e.g. no more `/api/42/icons`
  - only `/api/icons` will be mapped (taken from the Controller request mapping)
  - this change ([along with this one](https://github.com/dhis2/dhis2-core/pull/20523)) removes ~50k endpoint mappings
- kept minimal `CustomRequestMappingHandlerMapping`, as when removing and trying to use default Spring mapping, low-level issues were encountered with headers & possibly other unknowns

## Testing
- All existing tests pass, including ones with versions
- Issues were encountered with the E2E tests on GH, which helped find the deploy to Tomcat under custom context issue
- Locally tested using embedded server & standalone Tomcat

## Changes in memory use between 2.41 v 2.43dev
- ~10MB lighter on endpoint mappings

Screenshot 2025-04-18 at 13 41 43
Screenshot 2025-04-18 at 13 45 24

[DHIS2-18098]: https://dhis2.atlassian.net/browse/DHIS2-18098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-19486]: https://dhis2.atlassian.net/browse/DHIS2-19486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/filter/ApiVersionFilter.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Nonnull
+  private static HttpServletRequest getReroutedWrappedRequest(
+      @Nonnull HttpServletRequest req, String api, String endpoint) {
+    String versionlessEndpoint = api + endpoint;
+
+    // this will keep all properties of the original request, except for the overridden method
+    return new HttpServletRequestWrapper(req) {
+      @Override
+      public String getRequestURI() {
```
david-mackessy, line 113: this is required in order to work when deployed in a standalone Tomcat server with a custom context

jbee (approved):
Nice! That is a lot of code gone.

Good to see that this actually does make a difference in memory usage. 10MB might not be much in the grand scheme of things but this is just on `String` which I think are a good indicator for the overall costs of the mappings as they will have lots of other small objects that we now don't have to keep. Also the lookup doesn't need to go via large maps :)

netroms (approved):

larshelge:
Just checking that one point is not lost:

> Some endpoints had the API version as an optional request param, this param has been removed

The API version at the controller param level was used as part of an internal solution where the selected API version was automatically injected into the controller. I.e. not meant to be passed as a query parameter by the client. This was then used at the service level to adjust the response according to the version.

Just making sure this is understood before we remove it @david-mackessy .

david-mackessy:
> Just checking that one point is not lost:
> 
> > Some endpoints had the API version as an optional request param, this param has been removed
> 
> The API version at the controller param level was used as part of an internal solution where the selected API version was automatically injected into the controller. I.e. not meant to be passed as a query parameter by the client. This was then used at the service level to adjust the response according to the version.
> 
> Just making sure this is understood before we remove it @david-mackessy .

Hi @larshelge. Yes, it was understood that the API version was indeed injected (that code has also been removed) and not passed as a param.
- I checked with @maikelarabori about analytics code & he confirmed it is not used anywhere
- No tests fail, which suggests that tests which expect certain responses (with a version or without) still work
- If there is something specific that you think might need checking, please advise & it can be looked into

PR #20748, opened 2025-04-24
feat: Enrollment Program Indicator query refactor

luciano-fiandesio:
## Summary

This PR addresses the issue with Program Indicator expressions generating sub-queries that are not compatible with Apache Doris.

For a complete description of the problem and how this PR attempts at fixing it, please refer to this design document: https://github.com/dhis2/wow-backend/blob/master/docs/program_indicator_placeholders.md


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/programindicator/ctefactory/placeholder/PlaceholderParser.java
```diff
@@ -0,0 +1,218 @@
...
+   */
+  public static Optional<D2FuncFields> parseD2Func(String placeholder) {
+    Matcher m = D2_FUNC_PATTERN.matcher(placeholder);
+    if (!m.matches()) {
+      return Optional.empty();
+    }
+    try {
+      return Optional.of(
+          new D2FuncFields(
+              m.group(0), // raw
```
maikelarabori, line 165: Could we create constants for each integer? So we do not depend on comments.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CteDefinition.java
```diff
@@ -29,54 +29,260 @@
...
+        CteType.PROGRAM_INDICATOR); // Set type
+  }
 
-  /** Whether the CTE is a exists, used for checking if the enrollment exists * */
-  private boolean isExists = false;
+  /** Creates a CTE definition for filter CTEs (replacing filter subqueries). */
+  public static CteDefinition forFilter(
+      String queryItemId, String programStageUid, String cteDefinition) {
+    // Calls private constructor, passing CteType.FILTER
+    return new CteDefinition(
```
larshelge, line 200: Not saying it necessarily is better, though perhaps this is a sign that we should use a builder class instead. We can get one from Lombok. A high number of null and false parameters could lead to issues and maintenance problelsm later.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/CteDefinition.java
```diff
@@ -29,54 +29,260 @@
...
+  }
 
-  /** Whether the CTE is a filter * */
-  @Getter private boolean filter = false;
+  /** Creates a CTE definition for program indicators. */
+  public static CteDefinition forProgramIndicator(
+      String programIndicatorUid, String cteDefinition, boolean requiresCoalesce) {
+    // Calls private constructor, passing CteType.PROGRAM_INDICATOR
+    return new CteDefinition(
+        programIndicatorUid, // itemId
```
larshelge, line 182: Not saying it necessarily is better, though perhaps this is a sign that we should use a builder class instead. We can get one from Lombok. A high number of null and false parameters could lead to issues and maintenance problelsm later.

larshelge (approved):

larshelge:
Well done @luciano-fiandesio . Code looks great. Nice work on testing as usual.

maikelarabori (approved):

PR #20763, opened 2025-04-24
Update minio image version

david-mackessy:
Update the docker image for MinIO used in the integration tests.
We don't use the latest tag, to avoid potentially unstable behaviour.


teleivo (approved):
Do they have semver releases?

PR #20808, opened 2025-04-28
chore: only map/fetch fields the user wanted DHIS2-19137

teleivo:
Add `RelationshipFields` and `RelationshipItemFields` to avoid fetching relationship related data that is not going to be returned to the user as per their `fields` request parameter. This will affect

* `/tracker/trackedEntities`
* `/tracker/enrollments`
* `/tracker/events`
* `/tracker/relationships` - did not have its `fields` request parameter wired through into the service/store

This means that a request like `/trackedEntity?fields=relationships[relationship]` will not need to get the TE relationships `from` and `to` items with its nested entities.

There is one caveat which is the `RelationshipItem.trackedEntity/enrollment/event` which we currently always fetch. The `RelationshipService.findRelationshipItems` returns all items for a given entity. That item then has a relationship with a `from` and `to` `RelationshipItem` which is what we only fetch if needed. I can walk you through the 🤯 data model if you want to.

The difference to the other endpoints/entities is that relationships are fetched from the DB using hibernate. The reason is that hibernate resolves the cycle in our data model (due to relationships/relationship items) for us. The way we prevent unnecessary calls to the DB is to not map what is not needed in the `RelationshipItemMapper` in our service. Otherwise, the hibernate proxy of the `RelationshipItem` will call the DB.

## Change to query

Moved the Java `filter` into the query to not make hibernate proxy DB calls in a loop and allow the DB to optimize the query.

## Next

* try [fetching TEA in /trackedEntities only](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/aggregates/TrackedEntityAggregate.java#L154-L155) when the user wants them
* pass `Fields` class into event store instead of mapping it to new Java fields

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
@@ -142,7 +142,6 @@ class HibernateRelationshipStore extends SoftDeleteHibernateObjectStore<Relation
   }
 
   public List<Relationship> getRelationships(@Nonnull RelationshipQueryParams queryParams) {
-
     return relationshipsList(queryParams, null);
   }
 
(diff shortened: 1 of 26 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/HibernateRelationshipStore.java
```diff
@@ -175,11 +174,16 @@ public List<RelationshipItem> getRelationshipItemsByTrackedEntity(
     @Language("hql")
     String hql =
         """
-                from RelationshipItem ri
-                where ri.trackedEntity.uid = :trackedEntity
-                """;
+           select ri
+           from RelationshipItem ri
+           join ri.relationship r
+           join r.relationshipType rt
```
ameenhere, line 180: Is this join to rt necessary? Quick look makes me feel rt is not used?

enricocolasante (approved):

ameenhere (approved):

PR #20809, opened 2025-04-28
fix: make the hasCurrentUser more strict

netroms:
## Summary
The hasCurrentUser() in CurrentUserUtil should do the same checks as the getCurrentUser() method does, to avoid potential situations where the current user principal is set to a string.

A test for icon lookup/redirection was failing sometimes, since it is not the purpose of the LoginTest to deal with icons, that has its own life cycle, I removed it.

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/login/LoginTest.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/login/LoginTest.java
@@ -264,12 +264,6 @@ public class LoginTest extends BaseE2ETest {
   }
 
   @Test
-  void testRedirectIcon() {
-    testRedirectAsUser(
-        "/api/icons/medicines_positive/icon.svg", "api/icons/medicines_positive/icon");
-  }
-
-  @Test
   void testJsonResponseForFailedLogin() {
     try {
       getWithWrongAuth("/me", Map.of());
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/login/LoginTest.java
```diff
@@ -263,12 +263,6 @@ void testRedirectMissingEndingSlash() {
     testRedirectWhenLoggedIn("dhis-web-dashboard", "dhis-web-dashboard/");
   }
 
-  @Test
```
david-mackessy, line 266: Is this purposely being deleted? No mention of it in the PR details.

netroms: Yes, forgot to mention this. It was causing issues, in some cases. The lifecycle is not the purpose to test here, so it was better to eliminate it.

teleivo:
> ## Summary
> The hasCurrentUser() in CurrentUserUtil should do the same checks as the getCurrentUser() method does, to avoid potential situations where the current user principal is set to a string.

could you use the same code? that would make sure they use the same logic i.e. `hasCurrentUser() { return getCurrentUser() != null; }` or so

david-mackessy (approved):

vietnguyen (approved):

PR #20810, opened 2025-04-28
fix: Analytics e2e tests require Order

maikelarabori:
A few e2e Analytics tests need “Order”. So, we need to add the configuration back. See:

NoAnalyticsTablesErrorsScenariosTest
DataItemsAnalyticsTest

This setting has been removed as part of this PR:
https://github.com/dhis2/dhis2-core/pull/20759 

It should be reverted, as it causes our Analytics tests to fail.

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/resources/junit-platform.properties
+++ b/dhis-2/dhis-test-e2e/src/test/resources/junit-platform.properties
@@ -1,5 +1,5 @@
 junit.jupiter.extensions.autodetection.enabled=true
 # Uncomment and use @Order if you suspect test failures are due to test isolation/order issues
 # see https://junit.org/junit5/docs/current/user-guide/index.html#writing-tests-test-execution-order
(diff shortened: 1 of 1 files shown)


maikelarabori:
> > A few e2e Analytics tests need `@Order`. So, we need to add the configuration back. See:
> > NoAnalyticsTablesErrorsScenariosTest DataItemsAnalyticsTest
> > This setting has been removed as part of this PR: #20759
> > It should be reverted, as it causes our Analytics tests to fail.
> 
> The fix is to adapt your tests not to rely on order not the other way around 😬 its ok as a temporary solution.

Not really...
We need some tests to run before all others.
These tests are special because they have to run without any data in the DB, to handle the missing data accordingly.

After that, we can finally populate our analytics database, allowing the rest of the suite of tests to run on the actual data.
There is no issue with that, as we are using the `@Order` safely. There is no way that those specific ones will become flaky tests in the future.

The `@Order` exists for a reason, and it should be used as needed (assuming it's safe and the use case makes sense).

PR #20818, opened 2025-04-29
refactor: annotation mapping

vietnguyen:
## Summary
- This PR attempt to migrate Object Model mapping from xml to annotation for a few classes. The entities updated are: `CategoryCombo`, `DataSetElement`, `OptionSet`.
- The annotations used are from `jakarta.persistence` package together a few specific from `org.hibernate.annotations`.
- Removed the extend of `BaseIdentifiableObject` for affected classes. We will need to create another base object which comply with annotation mapping rules.

## Test
- Manually test the basic functions such as: crud operations, translations, attributes, sharing.


david-mackessy:
The option set hbm file is being removed, is there a reason why the other 2 hbm files are not?
Can you mention what kind of testing was done in the description to help reviewers understand how this change has been tested?

jbee (approved):
:tada: Great work!

jbee:
> The option set hbm file is being removed, is there a reason why the other 2 hbm files are not?

I was confused by the same thing first, I checked at least for the `CategoryOptionCombo`. That got updated because the `CategoryCombo` was converted (not the `CategoryOptionCombo`)

david-mackessy (approved):

PR #20836, opened 2025-04-30
fix: Weekly period name generation [DHIS2-19038]

abyot:
- on the fly adding+caching of days in a year provides unstable result. Since days in a year are known from a mapping table, better to pre-add them.
- Week names are calculated from period start/end dates (which are iso). For non iso calendars, start/end dates should first be converted to corresponing calendar dates before week names are generated.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitPeriodTypeParser.java
```diff
@@ -117,7 +117,7 @@ private DateInterval parseInternal(
               week,
               calendar,
               PeriodType.MAP_WEEK_TYPE.get(periodType.getName()),
-              new DateTimeUnit(year, 1, 1));
+              periodType.adjustToStartOfWeek(new DateTimeUnit(year, 1, 4), calendar));
```
jbee, line 120: Unclear: Why are we adjusting, why use `4` as parameter?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
```diff
@@ -271,45 +255,42 @@ public String formatPeriod(Period period, boolean shortVersion) {
...
-            endDate.getMonth().getValue(),
-            endDate.getDayOfMonth());
+      } else {
+        start = calendar.fromIso(period.getStartDate());
+        end = calendar.fromIso(period.getEndDate());
+        week = String.valueOf(calendar.week(start));
       }
 
-      year += dateTime.dayOfWeek().getAsShortText() + " " + year;
+      return String.format(
```
jbee, line 282: Unclear: It looks as if we now always use this formatting, not just for weekly period types. Why is that? Is that a change in observable behavior?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
```diff
@@ -271,45 +255,42 @@ public String formatPeriod(Period period, boolean shortVersion) {
...
-            startDate.getYear(),
-            startDate.getMonth().getValue(),
-            startDate.getDayOfMonth(),
-            endDate.getYear(),
-            endDate.getMonth().getValue(),
-            endDate.getDayOfMonth());
+      } else {
+        start = calendar.fromIso(period.getStartDate());
+        end = calendar.fromIso(period.getEndDate());
+        week = String.valueOf(calendar.week(start));
```
jbee, line 279: Unclear: I would read the change to the initialization of the 3 fields as it being important that the `calendar` is used for some reason. What is that reason? Why does it not apply to the other branch of the if?

jbee (commented):
As far as I can tell the changes look reasonable. But the details are not motivated or explained. I am not aware of what problem this solves. I am not sure what team of group of people is responsible for maintaining calendar logic but as of now my understanding is that platform is part of it. So changes to calendar logic that will affect all calendars need to be discussed so everyone is on the same page with this.

As I understand it the changes to table lookup in the Nepali calendar is a pure improvement without changing observable behavior so that to me is not problematic. 
I leave a few comments on changes where I would like to know more about the background.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitPeriodTypeParser.java
abyot: In ISO week, the first week of a year should contain Jan 4.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
abyot: the `calendar` is used when dealing with non `iso` calendar. We wanted to know which `calendar` so that we generate the date based on this calendar. The other branch of the `if` uses just period start and end dates which are based on `iso` calendar.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
abyot: for other period types we are not using start and end dates in the name ... we just say 012025 for example Monthly period january 2025. For weeks saying W12025 ... leaves users into lots of guess for the exact week. Hence the need for start and end dates based on user calendar

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitPeriodTypeParser.java
maikelarabori, line 120: @abyot, correct if I'm wrong. But I tried the following unit test, in the test class `NepaliCalendarTest`.

```
    dateTimeUnit = new DateTimeUnit(2025, 1, 4, true);
    period = periodType.createPeriod(dateTimeUnit, calendar);
    startDate = new DateTime(2024, 12, 30, 0, 0);
    endDate = new DateTime(2025, 1, 5, 0, 0);

    assertEquals("2025W1", period.getIsoDate());
    assertEquals(startDate.toDate(), period.getStartDate());
    assertEquals(endDate.toDate(), period.getEndDate());
```
And I get assertion error:

```
Expected :2025W1
Actual   :2024W52
```

These websites say it should be 2025W1, assuming my test is right:
`https://www.epochconverter.com/weeks/2025`
`https://isoweeks.com/`

Is my test correct? What do you think?

Thanks!

jbee: Sorry for late response, I wasn't very clear with my comment. I didn't want an explanation of the line of code, I wanted the code to be updated in a way so it becomes clear reading it what we are doing and why. In the simplest form this could be a comment on the branch or line. In this case it should be clear why you want to check the 4. of January of all days. There must be a reason.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
jbee: Same here. I wanted the code to make it clear why it uses different approaches, something like "Note, this needs to use ISO because..."

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitPeriodTypeParser.java
abyot: Thanks @maikelarabori. 

Yes, it helps. However, our calendar implementations assume the dateTimeUnit to be on the same calendar as the param calendar. In your case we have dateTimeUnit in iso but calendar in nepali, this is unlikely to happen in implementation usecase.

I have made fix for the calendar implementation to distiguish between iso and non-iso dateTimeUnit.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
abyot: sure, have put a comment documenting the reason behond.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/i18n/I18nFormat.java
abyot: yes, that is how it is currently. the changed code for name formatting sits under 
`if (periodType instanceof WeeklyAbstractPeriodType) `

dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitPeriodTypeParser.java
abyot: thanks. I have put a comment mentioning the reason.

jbee:
What is the state of this PR, are more changes expected or is this waiting for reviews?

abyot:
> What is the state of this PR, are more changes expected or is this waiting for reviews?

waiting for approval.

PR #20839, opened 2025-04-30
refactor: Get job once in scheduler service tryExecuteNow

david-mackessy:
The job config is retrieved **_at least once_** in the `tryExecuteNow` method.
The code is repeated 3 times however.
Refactor to only have 1 line which satisfies all code paths.
It means a DB call to get the job is done first, which doesn't look like it would make much difference.


david-mackessy:
close off as it looks like the first check in the method is required to perform some action on the job config

PR #20842, opened 2025-04-30
chore: only fetch TEA if returned to user on /trackedEntities DHIS2-19137

teleivo:
* and remove some unused fields from `EventQueryParams`: they were never written to so we remove them and the SQL snippets they would add to the event query

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
@@ -565,15 +565,7 @@ class JdbcEventStore {
     }
 
     sqlBuilder.append(") as event left join (");
-
-    if (queryParams.isIncludeAttributes()) {
-      sqlBuilder.append(getAttributeValueQuery());
-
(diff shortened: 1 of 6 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
```diff
@@ -1285,7 +1261,7 @@ inner join (select coc.uid, coc.code, coc.name, coc.attributevalues, coc.categor
  coc.categoryoptioncomboid \
 """;
 
-    if (!isSuper(user)) {
+    if (isNotSuperUser(user)) {
```
muilpp, line 1264: Not sure if this is something we want to address in this PR, but at some point we should drop this user and start using `UserDetails` instead.

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/JdbcEventStore.java
teleivo: feel free to do it anytime 👌🏻

enricocolasante (approved):

PR #20850, opened 2025-05-01
fix: Aggregate Data Exchange job creator as job executor [DHIS2-18324]

david-mackessy:
# Issue
The Aggregate Data Exchange job is broken & throws an error when executing - no real info about the root cause.
```text
13:57:20.089 ERROR [pool-4-thread-1] o.h.d.s.n.DefaultNotifier                : Process failed: Transaction silently rolled back because it has been marked as rollback-only
13:57:20.089 ERROR [pool-5-thread-2] o.h.d.s.RecordingJobProgress             : [AGGREGATE_DATA_EXCHANGE HYCkfYFbMot] Process failed after 0.039s: Transaction silently rolled back because it has been marked as rollback-only
org.springframework.transaction.UnexpectedRollbackException: Transaction silently rolled back because it has been marked as rollback-only
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:804)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:758)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:698)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:416)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:119)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:727)
	at org.hisp.dhis.dataexchange.aggregate.AggregateDataExchangeService$$SpringCGLIB$$0.exchangeData(<generated>)
	at org.hisp.dhis.dataexchange.aggregate.AggregateDataExchangeJob.execute(AggregateDataExchangeJob.java:86)
```

# Cause
There is no valid `User` associated with the job when executing. A `NullPointerException` is thrown in `DefaultAnalyticsSecurityManager#decideAccessDataViewOrganisationUnits` when accessing a `null` `User`.
The exception marks the transaction for rollback which explains the error above.

It looks like this issue may have been introduced in #18362 which removed the old anti-pattern of treating a `null` `User` as a super user.

# Fix
Make the job creator the job executor by default, by marking it as a job type with this behaviour (no client changes required, when `executedBy` param not supplied)
Also, clients can optionally supply an `executedBy` value in the job config payload to set this value e.g.
```json
{
    "name": "test-dx-job-2",
    "jobType": "AGGREGATE_DATA_EXCHANGE",
    "cronExpression": "3 2 * ? * *",
    "jobParameters": {
        "dataExchangeIds": [
            "R9Urc25BSio"
        ]
    },
    "executedBy": "UserUid0001"
}
```

# Info
Improved the job error logs for this job type. Instead of returning a generic error like:
```text
Aggregate data exchange completed with errors
```
return an error with more info if available like:
```text
Aggregate data exchange completed with errors: User has no data write access for AggregateDataExchange: ANC test
```

# Testing
## Automated
Added a job config integration test to check that the job creator is saved in the job config.
Added E2E test with a Aggregate Data Exchange job.

## Manual
Tested locally & in IM, creating jobs and also running the exchange normally through the exchange endpoint.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
```diff
@@ -190,7 +190,7 @@ static Defaults dailyRandomBetween3and5(String uid, String name) {
    *     creation unless it was set explicitly
    */
   public boolean isDefaultExecutedByCreator() {
-    return this == HTML_PUSH_ANALYTICS;
+    return this == HTML_PUSH_ANALYTICS || this == AGGREGATE_DATA_EXCHANGE;
```
david-mackessy, line 193: Adding the job type here means that when a job of this type is created, whoever creates the job is set as the `executedBy` value in the job config. That user will be used to execute the job even if triggered by someone else.

dhis-2/dhis-services/dhis-service-data-exchange/src/main/java/org/hisp/dhis/dataexchange/aggregate/AggregateDataExchangeJob.java
```diff
@@ -89,7 +90,11 @@ public void execute(JobConfiguration config, JobProgress progress) {
     notifier.addJobSummary(config, NotificationLevel.INFO, allSummaries, ImportSummaries.class);
     ImportStatus status = allSummaries.getStatus();
     if (status == ImportStatus.ERROR) {
-      progress.failedProcess("Aggregate data exchange completed with errors");
+      String errors =
+          allSummaries.getImportSummaries().stream()
+              .map(ImportSummary::getDescription)
+              .collect(Collectors.joining(","));
+      progress.failedProcess("Aggregate data exchange completed with errors: {}", errors);
```
david-mackessy, line 97: Adds some meaningful info when the job fails instead of just a generic message of:
`Aggregate data exchange completed with errors`

dhis-2/dhis-test-e2e/src/test/resources/metadata/adex-metadata.json
```diff
@@ -0,0 +1,111 @@
+{
```
david-mackessy, line 1: this data was taken from the bug report & trimmed down

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
jbee: As I understand it you mad this change to fix 2.42+ but in 2.41- you intend to use the added `executedBy` in the job's parameters? Having that job specific parameter but then it is not used is also a bit odd. Wouldn't it be more consistent to use it in all versions?

jbee (approved):

PR #20876, opened 2025-05-05
refactor: use JDBC for min-max batch requests

jbee:
### Summary
Implements bulk upsert and delete via JDBC. All single create/update and delete operations are implemented as a special case of single bulk operation. This also means the error responses are now aligned across all methods.

### API changes
Since the same (bulk) code is used for all single and bulk operations the error responses for some of the existing single object operations has now changed slightly. This means instead of `CONFLICT` they now throw `BAD_REQEST` (since it is a user/client mistake they indicate). Existing error codes got reused but the messages were partly adjusted slightly. 

### CSV => POJO
For the CSV to POJP mapping a new utility `CSV` was created. Jackson's CSV support was tested but it doesn't really work for CSV => POJO direction as soon as a bit of flexibility is wanted. I wanted to allow presence of not required columns, aliases for column names to allow lower case, random column order (order based on header), validation of required columns. This not only wasn't possible after trying many suggestions (AI) but also is dangerous as some configuration combinations would silently map values to the wrong properties which might not be recognized later if these in the POJO are of a compatible type (both strings or both numbers and alike). So I decided we should not use Jackson CSV for CSV => POJO. POJO => CSV is less problematic. 

### Automatic Testing
The API has been covered with both success and error scenarios. 

The bulk API does not repeat all error scenarios that single (which goes via bulk) has already covered. Their behaviour is almost the same. Just that bulk does not throw an error if not all (or even any) of the values was not successful (because the key-combination does not exist) whereas single operation does throw an error in that case.

I converted e2e tests to controller tests. They were originally created as e2e because it would not work as controller tests.

### Manual Testing
...


netroms (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/csv/CSV.java
```diff
@@ -0,0 +1,274 @@
...
+          Object[] args = new Object[types.length];
+          for (int i = 0; i < values.size(); i++) {
+            int compIdx = compIdxByColIdx[i];
+            if (compIdx >= 0) {
+              Column<?> column = columns.get(compIdx);
+              String value = values.get(i);
+              if (column.required && value == null)
+                throw new IllegalArgumentException(
+                    "Column %s is required and cannot be empty".formatted(column.name));
+              args[compIdx] = column.deserializer.apply(value);
```
david-mackessy, line 146: Are there potentials for exceptions here when the column is not marked as required and the value is null?
e.g. csv row data of `fKiYlhodhB1,elD9B1HiTJO,HllvX50cXC0,0,,`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/minmax/MinMaxValue.java
```diff
@@ -29,44 +29,43 @@
...
-public class MinMaxDataElementUtils {
-
-  private MinMaxDataElementUtils() {}
+public record MinMaxValue(
+    @Nonnull @OpenApi.Property({UID.class, DataElement.class}) UID dataElement,
+    @Nonnull @OpenApi.Property({UID.class, OrganisationUnit.class}) UID orgUnit,
+    @Nonnull @OpenApi.Property({UID.class, CategoryOptionCombo.class}) UID optionCombo,
+    Integer minValue,
+    Integer maxValue,
+    Boolean generated)
```
david-mackessy, line 50: It looks like the constraints are spread out in different places for this record.
`dataElement`, `orgUnit`, `optionCombo` are marked nonnull & get checked when importing through CSV.
`minValue` and `maxValue` are not primitive and not marked nonnull but it looks like they do get checked in the service as being required.
Could we just mark them as nonnull here as well for clarity?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MinMaxValueImportControllerTest.java
```diff
@@ -0,0 +1,262 @@
...
+      }""";
+
+  @Language("csv")
+  private final String csv =
+      """
+    dataElement,orgUnit,optionCombo,minValue,maxValue
+    %1$s,%3$s,%2$s,0,10
+    %1$s,%4$s,%2$s,0,10
+    %1$s,%5$s,%2$s,0,10
+    %1$s,%6$s,%2$s,0,10
```
david-mackessy, line 116: Never knew about this formatting option, nice 👍

david-mackessy (approved):
Nice work. Aligning behaviour between single & bulk is nice. Left a few small comments.
I'm sure rolling with our own CSV parser has pros and cons. I'm a bit surprised there wasn't a readily-available library that suited our requirements.

jason-p-pickering (commented):
1. I think the use of the `optionCombo`is not really necessary, since in the previous implementation I used `categoryOptionCombo` which is the same terminology as with data values. This change also has the effect of breaking min/max entry in the current data entry client. It just feels like an unnecessary change to me, unless it is part of something bigger. 
2. The original implementation used multi-row inserts and multi-row deletes, chunked in batches of 1000 or so. Is there a reason you decided to use single row operations? I would think that would result in a performance drop?

jbee:
> The original implementation used multi-row inserts and multi-row deletes, chunked in batches of 1000 or so. Is there a reason you decided to use single row operations? I would think that would result in a performance drop?

@jason-p-pickering , @david-mackessy I think you did read this wrong. There is a loop that uses JDBC batching, within the loop you see a "single" upsert in the batch. The chunking is now implicit and depends on JDBC level settings for batch processing, the default is 5000 (I think if I remember correctly). I AI can be trusted this should be the best for large scale batching while also being very readable and straight forward to use.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/csv/CSV.java
jbee: I will check and add a test :+1:

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MinMaxValueImportControllerTest.java
jbee: :guide_dog:

jbee:
@david-mackessy 

> I'm sure rolling with our own CSV parser has pros and cons. I'm a bit surprised there wasn't a readily-available library that suited our requirements.

I mentioned that in the description, I tried to make to make it work with Jackson's CSV module but it actually cannot do it the way I think CSV should be handled, I will explain more in our next Tuesday meeting.

jason-p-pickering:
@jbee no, I think I did not read this wrong at all. 

 Here are the results from the profiler. 
image

With the current implementation, we get n-single row insert statements. That they are batched internally, is totally different issue. 

With the original implementation, there were far fewer (but **mutli-row**) insert statements. 

image

These themselves might be batched internally, but it was the difference between many single row inserts versus fewer mutli-row inserts and why you did not chose to do it this way. So while I certainly do trust your expertise, I would think that it might be worth looking into.

jason-p-pickering:
Its not a huge difference but its slower with the same payload. :-D

With multi-row inserts

Selection_935


With single row inserts
Selection_936

jason-p-pickering:
> @david-mackessy 
> 
> > I'm sure rolling with our own CSV parser has pros and cons. I'm a bit surprised there wasn't a readily-available library that suited our requirements.
> 
> I mentioned that in the description, I tried to make to make it work with Jackson's CSV module but it actually cannot do it the way I think CSV should be handled, I will explain more in our next Tuesday meeting. 

Did you consider the Apache Commons CSV library? I agree that there are pros and cons of building our own , but you only mentioned Jackson here and the problems with it.  Was this the only alternative you considered or which was feasible?

jbee:
Open comments were addressed in #21003

PR #20884, opened 2025-05-06
fix: route connection leaks and allowed URLs - DHIS2-19556 and DHIS2-19551 blocking v42 release

cjmamo:
* Temporarily buffer response body like in v41 to prevent databse connection leaks
* Validate only the hostname in the route URL since the `route.remote_servers_allowed` setting does not accept paths which causes the validation to fail when creating a route with a subpath or even a normal path

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/RouteControllerTest.java
@@ -103,9 +103,7 @@ class RouteControllerTest extends PostgresControllerIntegrationTestBase {
     @Bean
     public DhisConfigurationProvider dhisConfigurationProvider() {
       Properties override = new Properties();
-      override.put(
-          ConfigurationKey.ROUTE_REMOTE_SERVERS_ALLOWED.getKey(),
-          "http://*,https://stub");
+      override.put(ConfigurationKey.ROUTE_REMOTE_SERVERS_ALLOWED.getKey(), "http://*,https://stub");
 
       PostgresDhisConfigurationProvider postgresDhisConfigurationProvider =
           new PostgresDhisConfigurationProvider(null);


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -376,6 +386,23 @@ protected WebClient.RequestHeadersSpec<?> buildRequestSpec(
...
+              .doOnNext(DataBufferUtils.releaseConsumer());
+      dataBufferFlux.blockLast(Duration.ofSeconds(route.getResponseTimeoutSeconds()));
+
+      return outputStream.toByteArray();
+    } else {
+      return null;
+    }
+  }
+
+  // FIXME: unsafe since it leaks database connections (see
```
Birkbjo, line 404: Is this comment correct? I would think it leaks http connections?

david-mackessy (approved):
If this solves the issue that was found, great 👍
There is one big question that pops up from looking at the routes feature, the use of WebFlux (Reactor). This goes against what the rest of the system uses (MVC) and quick checks on the internet say that they don't play nicely together & may not even work as expected (possibly related to this issue?). Potentially something that should be looked into (and proved to be an issue or not).

Caveats of mixing (from AI)
| Concern | Explanation |
|-- | --|
|Different runtimes | Spring MVC uses Servlet API, runs on traditional containers (like Tomcat); WebFlux uses non-blocking runtimes (like Netty). They are fundamentally different. |
| Threading model clash | Mixing blocking MVC code with non-blocking WebFlux can lead to thread starvation or performance bottlenecks.
| Complexity | It increases your application's complexity, configuration burden, and mental overhead. |
| Incompatibility | You can't use @Controller with @RestController in MVC and expect full interop with WebFlux features like Mono/Flux unless you're fully on the reactive stack. |

We use MVC in the Controller & then WebFlux in the service??😵‍💫

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
david-mackessy: I'd delete this whole method, especially if this PR is the fix. Why keep it?

Birkbjo (approved):
Tested this locally and I haven't experienced any crashes with this fix. So LGTM!

PR #20904, opened 2025-05-08
feat: Enable HikariCP database pool leak detection

jason-p-pickering:
Implements https://dhis2.atlassian.net/browse/DHIS2-19576

### DIFF
--- a/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
+++ b/dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
@@ -264,6 +264,12 @@ public enum ConfigurationKey {
   CONNECTION_POOL_TIMEOUT("connection.pool.timeout", String.valueOf(SECONDS.toMillis(30)), false),
 
   /**
+   * Hikari DB pool feature. Connection leak detection threshold: Set the maximum number of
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -263,6 +263,12 @@ public enum ConfigurationKey {
    */
   CONNECTION_POOL_TIMEOUT("connection.pool.timeout", String.valueOf(SECONDS.toMillis(30)), false),
 
+  /**
+   * Hikari DB pool feature. Connection leak detection threshold: Set the maximum number of
+   * milliseconds that a connection can be out of the pool before a message is logged. (default: 0 -
+   * no leak detection)
+   */
+  CONNECTION_POOL_LEAK_THRESHOLD("connection.pool.leak.threshold", "0", false),
```
jbee, line 271: I'd prefer a name that without reading comments or docs tells more about what sort of value this is. Maybe something like `connection.pool.warn_max_age` ? When it comes to use of the `.` I think this should be understood as a grouping. So words in the name of the last level should not use `.` but it seems we are using `_` so isolate words. For example, `system.sql_view_write_enabled` and `system.sql_view_table_protection` use `system` as a group, then have two configurations in that group with a name that has multiple words.

PR #20911, opened 2025-05-09
Port DHIS2-19577

cjmamo:
* re-enable streaming of response body

* include OpenEntityManagerInViewFilter in the async request filter chain so that it closes the DB connection when a response to the async request is returned

Refs: https://dhis2.atlassian.net/browse/DHIS2-19577


dhis-2/dhis-web-api/src/test/resources/dhis.conf
jbee, line 1: Should this file be deleted? It is empty, isn't it?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/servlet/DhisWebApiWebAppInitializer.java
```diff
@@ -127,8 +127,10 @@ public static void setupServlets(
         context.addFilter("openSessionInViewFilter", OpenEntityManagerInViewFilter.class);
     openSessionInViewFilter.setInitParameter(
         "entityManagerFactoryBeanName", "entityManagerFactory");
-    openSessionInViewFilter.addMappingForUrlPatterns(null, false, "/*");
-    openSessionInViewFilter.addMappingForServletNames(null, false, "dispatcher");
+    openSessionInViewFilter.addMappingForUrlPatterns(
+        EnumSet.of(DispatcherType.REQUEST, DispatcherType.ASYNC), false, "/*");
```
jbee, line 131: Does this mean now any URL will be handled as if it is async? Might that have unintended consequences?

jbee (approved):

david-mackessy (approved):

PR #20927, opened 2025-05-12
chore: Format code blocks

larshelge:


netroms (approved):

vietnguyen (approved):

jbee (commented):
I hugely in favor of using code blocks. 
But do we really need to maintain not having newlines by using the `\` at the end? 
This breaks the easy integration of copying the SQL from and to between Java and a database console. Therefore I would prefer not to use the `\` and simply have the SQL be multiple lines. Yes, that might increase the lines in logs but even in logs when you are actually interested in the SQL breaking it down helps reading it.

PR #20957, opened 2025-05-14
fix: Fix pagination for TE enrolled in more programs [DHIS2-19595] (2.41)

enricocolasante:
Backport of https://github.com/dhis2/dhis2-core/pull/20943


zubaira (approved):
LGTM. 
Shouldn't we also assert that the returned event UID matches the one we requested?

PR #20973, opened 2025-05-15
task: use jdbc in enrollment store [DHIS2-19543]

muilpp:
This PR adds a new enrollment store, which will be our default, using JDBC instead of Hibernate.

I've tested it on an environment with 73k enrollments, here are the results:

**1. Get enrollment by TE and Program**
     This query is frequently triggered in the Capture app. [Query plan](https://explain.dalibo.com/plan/043ef0c88bc7e1c8)
        - Uses indexes effectively.
        - Returns a single row.
        - Exec time: ~0.19 ms.

**2. Get all enrollments for a given TE** [Query plan](https://explain.dalibo.com/plan/63953g4b4a8512ah#raw)
       - Similar efficiency as above.
       - Uses indexes well.
       - Returns a single row.
       - Exec time: ~0.23 ms.

**3. Worst case: Get all enrollments and all fields (no filters, no pagination)** [Query plan](https://explain.dalibo.com/plan/2fc7c1dd0d3271h9#raw)
       - Joins with notes and trackedentity tables for every enrollment.
       - Most enrollments don't have notes, so the join adds unnecessary overhead.
       - Exec time: ~467 ms.

We are not running N+1 queries when fetching notes or attributes, which is great.
However, fetching all enrollments without filters results in thousands of joins for notes, even when most entries don’t have any.

Since Hibernate splits the logic into multiple queries (enrollments, tracked entities, notes, attributes, etc.), it’s hard to do a direct query plan comparison.
To benchmark, I ran the same API request 20 times locally on two instances: one using JDBC, the other using Hibernate.

I tested the Endpoint which I think it's the most common: `/api/tracker/enrollments?enrollmentId=enrollment&orgUnitId=orgUnit&programId=program&teiId=trackedEntity`.
- JDBC average response: 379 ms
- Hibernate average response: 392 ms

These results should be taken with a grain of salt, but while the difference isn’t massive, JDBC shows slightly better performance and provides more control and transparency over the queries.

### Possible Improvements
- If soft-deleted enrollments are common, we might benefit from a partial index that only includes non-deleted records.
-  Notes are joined by default, even though most enrollments don’t have any. We could consider an opt-in approach—similar to how we handle events, relationships, or attributes—where notes are only fetched when explicitly requested.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentQueryParams.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentQueryParams.java
@@ -95,6 +95,9 @@ class EnrollmentQueryParams {
   /** Indicates whether to include soft-deleted enrollments */
   private boolean includeDeleted;
 
+  /** Indicates whether to include tracked entity attribute data in the enrollment */
+  private boolean includeAttributes;
+
   private List<Order> order;
 
   // -------------------------------------------------------------------------
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
@@ -29,61 +29,20 @@
  */
 package org.hisp.dhis.tracker.export.enrollment;
 
-import static org.hisp.dhis.common.IdentifiableObjectUtils.getUids;
-import static org.hisp.dhis.commons.util.TextUtils.getQuotedCommaDelimitedString;
-import static org.hisp.dhis.util.DateUtils.nowMinusDuration;
-import static org.hisp.dhis.util.DateUtils.toLongDateWithMillis;
-import static org.hisp.dhis.util.DateUtils.toLongGmtDate;
-
 import jakarta.persistence.EntityManager;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
-import java.util.StringJoiner;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.annotation.Nonnull;
-import lombok.Builder;
-import lombok.Getter;
-import org.apache.commons.lang3.StringUtils;
-import org.hibernate.query.Query;
-import org.hisp.dhis.common.OrganisationUnitSelectionMode;
-import org.hisp.dhis.common.UID;
 import org.hisp.dhis.common.hibernate.SoftDeleteHibernateObjectStore;
-import org.hisp.dhis.commons.util.SqlHelper;
-import org.hisp.dhis.organisationunit.OrganisationUnit;
 import org.hisp.dhis.program.Enrollment;
 import org.hisp.dhis.program.EnrollmentStatus;
-import org.hisp.dhis.program.ProgramType;
 import org.hisp.dhis.security.acl.AclService;
-import org.hisp.dhis.tracker.Page;
-import org.hisp.dhis.tracker.PageParams;
-import org.hisp.dhis.tracker.export.Order;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Component;
 
 // This class is annotated with @Component instead of @Repository because @Repository creates a
 // proxy that can't be used to inject the class.
-@Component("org.hisp.dhis.tracker.export.enrollment.EnrollmentStore")
+@Component("org.hisp.dhis.tracker.export.enrollment.HibernateEnrollmentStore")
 class HibernateEnrollmentStore extends SoftDeleteHibernateObjectStore<Enrollment> {
-
-  private static final String DEFAULT_ORDER = "id desc";
-
-  private static final String STATUS = "status";
-
-  /**
-   * Enrollments can be ordered by given fields which correspond to fields on {@link
-   * org.hisp.dhis.program.Enrollment}.
-   */
-  private static final Set<String> ORDERABLE_FIELDS =
-      Set.of(
-          "completedDate",
-          "created",
-          "createdAtClient",
-          "enrollmentDate",
-          "lastUpdated",
-          "lastUpdatedAtClient");
 
   public HibernateEnrollmentStore(
       EntityManager entityManager,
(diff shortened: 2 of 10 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/HibernateEnrollmentStore.java
```diff
@@ -29,62 +29,21 @@
...
-import org.hisp.dhis.tracker.PageParams;
-import org.hisp.dhis.tracker.export.Order;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Component;
 
 // This class is annotated with @Component instead of @Repository because @Repository creates a
 // proxy that can't be used to inject the class.
-@Component("org.hisp.dhis.tracker.export.enrollment.EnrollmentStore")
+@Component("org.hisp.dhis.tracker.export.enrollment.HibernateEnrollmentStore")
```
muilpp, line 44: @enricocolasante We still need the Hibernate store to persist enrollments.
I've seen we've got the `HibernateEventStore` in [service-core](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/hibernate/HibernateEventStore.java), should I move this one there too?

enricocolasante: No, at some point we should go the other way around and move `HibernateEventStore` into tracker module.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    private Date formatDate(Timestamp timestamp) {
+      return timestamp == null ? null : toDate(timestamp.toInstant().atZone(ZONE_ID));
+    }
+
+    private Date toDate(ZonedDateTime zonedDateTime) {
+      return Date.from(zonedDateTime.toInstant());
+    }
+  }
```
enricocolasante, line 599: Did you check in https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/util/DateUtils.java if we don't have something already implemented here?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    private Set<TrackedEntityAttributeValue> mapTrackedEntityAttributeValues(
+        String jsonAttributes, TrackedEntity trackedEntity) {
+      List<JdbcAttribute> attributes;
+      ObjectMapper mapper = new ObjectMapper();
+      try {
+        attributes = mapper.readValue(jsonAttributes, new TypeReference<>() {});
+      } catch (JsonProcessingException e) {
+        log.error("Error mapping enrollment attributes: {}", jsonAttributes);
```
enricocolasante, line 559: When is this exception being thrown?
Can you reproduce it?
Swallowing this exception for attributes that the user explicitly asked for it is scary

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+  }
+
+  private void addTrackedEntityConditions(
+      StringBuilder sql,
+      EnrollmentQueryParams params,
+      MapSqlParameterSource sqlParams,
+      SqlHelper hlp) {
+    if (params.hasTrackedEntity()) {
+      sql.append(hlp.whereAnd()).append("te.uid = :trackedEntityUid");
```
enricocolasante, line 290: When we are mapping to query params we are doing ACL on the tracked entity, this is why we need to get it from the DB at the moment of the mapping. Is it necessary or are we able to have all our ACL check in the query here?

Probably nothing to fix now, I am just commenting so we can discuss it

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
```diff
@@ -0,0 +1,641 @@
...
+  private void addProgramConditions(
+      StringBuilder sql,
+      EnrollmentQueryParams params,
+      MapSqlParameterSource sqlParams,
+      SqlHelper hlp) {
+    sql.append(hlp.whereAnd()).append("p.type = :programType");
+    sqlParams.addValue("programType", ProgramType.WITH_REGISTRATION.name());
+
+    if (params.hasProgram()) {
+      sql.append(hlp.whereAnd()).append("p.uid = :programUid");
```
enricocolasante, line 244: When we will push all the ACL to this store, can we have only the program UID in query params? So we avoid to retrieve the program when mapping from operation params to query params

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
```diff
@@ -0,0 +1,641 @@
...
+import org.locationtech.jts.io.WKBReader;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component("org.hisp.dhis.tracker.export.enrollment.EnrollmentStore")
+@RequiredArgsConstructor
+public class JdbcEnrollmentStore {
```
enricocolasante, line 90: Can we remove `public` access modifier?

enricocolasante (commented):
Is one of next step to move ACL check inside the store?

muilpp:
> Is one of next step to move ACL check inside the store?

yes, but I think it would make sense to first decide if we'll make a request parameter mandatory (TE or program), as that will probably simplify the ACL validation.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
muilpp: This exception would be thrown if the json is malformed and can't be deserialized to `JdbcAttribute`.
However, I think it should not happen, because we validate the payload before we persist attributes into the database, so the data we fetch should always be in a correct state, that's why I'm swallowing this exception.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
muilpp: The TE is an optional field in the request.
The ideal scenario for us is when it's specified, because then we can run the ACL before the query.
If it's not, then we have the ACL check after we get the DB results. This is currently breaking pagination, so we need to move it into the query.

As mentioned somewhere else in this PR, it'd be great to decide if we want to make the TE param mandatory in this endpoint, as that would make the ACL check easier to implement.

As a side note, we'd still need to get the TE UID, as that's expected in the API response. What I think we won't need to return is the TET UID and TET sharing object.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
muilpp: Yes, that would make sense.
I'd just mention the same as above, which is that it'd be good to know if we want to make the program a mandatory parameter in the request or not before we start with the ACL validation.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentQueryParams.java
```diff
@@ -95,6 +95,9 @@ class EnrollmentQueryParams {
   /** Indicates whether to include soft-deleted enrollments */
   private boolean includeDeleted;
 
+  /** Indicates whether to include tracked entity attribute data in the enrollment */
+  private boolean includeAttributes;
```
ameenhere, line 99: Is there a reasoning behind this? Will this be a breaking change?

muilpp: The reason of this flag is because I want to get the enrollment attributes from the database only when the user requests them. 
It's not a breaking change, we currently support this on the API level if the user includes `fields=attributes` in the request. With Hibernate we didn't use it, because the attributes were loaded lazily, but we do need it with JDBC.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/JdbcEnrollmentStore.java
enricocolasante: I am still not convinced that we should swallow this exception.
If user is asking for enrollment with attributes and he gets back enrollment with no attributes, he would think enrollment has no attributes unless he goes and look into the logs.
I would prefer the call to fail with a 500 explaining what is happening and explaining that if the user still needs the enrollment he can ask for it without the attributes.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentQueryParams.java
ameenhere: I mean. What was the previous behaviour in the API? if attributes was not explicitly requested as part of fields queryParam, we never sent it? Was the default that we did not send the attributes?

ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/EnrollmentQueryParams.java
muilpp: Yes, this behavior didn't change. 
Previously, we only returned attributes if explicitly requested, as in `fields=attributes`. By default we didn't return them.

PR #20974, opened 2025-05-15
feat: add BaseMetadataObject

vietnguyen:
This pull request refactors metadata-related classes in the DHIS2 codebase to improve code reuse and simplify translation handling. The changes include introducing a new base class for metadata objects, embedding translation properties, and refactoring existing classes to extend the new base class.

### Refactoring Metadata Classes

* **Introduction of `BaseMetadataObject`**: Created a new base class, `BaseMetadataObject`, to centralize common metadata properties such as `uid`, `created`, `lastUpdated`, `createdBy`, and `lastUpdatedBy`. Classes like `CategoryCombo` and `OptionSet` now extend this base class, reducing code duplication. 
* BaseMetadataObject will have `@MappedSuperClass` meaning all properties must be mapped to a database column, no exception. Hence, all classes extending this new base class must also map those properties to columns in corresponding database table.

### Translation Handling Improvements

* **Embedded `TranslationProperty`**: Replaced the direct handling of translations in `CategoryCombo` with a new embedded `TranslationProperty` class. This encapsulates translation-related logic, such as caching and locale-based retrieval, and maps translations to a JSONB column in the database.

### Class-Specific Changes

* **Refactoring `CategoryCombo`**: Updated `CategoryCombo` to extend `BaseMetadataObject`, removed redundant properties, and integrated `TranslationProperty` for handling translations. 
* **Refactoring `OptionSet`**: Updated `OptionSet` to extend `BaseMetadataObject`, removing duplicate metadata properties like `uid`, `created`, and `lastUpdated`.

### Test
- Add tests in `SchemaServiceTest` and `TranslationServiceTest` to make sure the  `translations` property work as expected.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
```diff
@@ -0,0 +1,70 @@
...
+  @Column(name = "lastUpdated", nullable = false)
+  @Temporal(TemporalType.TIMESTAMP)
+  protected Date lastUpdated;
+
+  @ManyToOne
+  @JoinColumn(name = "lastupdatedby", foreignKey = @ForeignKey(name = "fk_lastupdateby_userid"))
+  protected User lastUpdatedBy;
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "userid", foreignKey = @ForeignKey(name = "fk_categorycombo_userid"))
```
david-mackessy: foreign key is `fk_categorycombo_userid`, which might work for category combo. Will every other metadata object need a key here as well? Or will a generic foreign key work for all?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
```diff
@@ -113,31 +111,17 @@
 @Setter
 @Entity
 @Table(name = "optionset")
-public class OptionSet implements IdentifiableObject, VersionedObject, MetadataObject {
+public class OptionSet extends BaseMetadataObject
+    implements IdentifiableObject, VersionedObject, MetadataObject {
```
david-mackessy: If BaseMetadataObject implements MetadataObject, do we still need to add the implements MetadataObject here as well?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
david-mackessy: The BaseMetadataObject probably shouldn't know about any other child class extending it (category combo in this case).

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
```diff
@@ -573,30 +537,27 @@ public void setOwner(String ownerId) {
...
-    return translations;
+    return translations.getTranslations();
   }
 
-  /** Clears out cache when setting translations. */
   @Override
-  public void setTranslations(Set<Translation> translations) {
-    this.translationCache.clear();
-    this.translations = translations;
+  public int compareTo(@Nonnull IdentifiableObject o) {
```
jbee: This stripes me as odd, comparing `CategoryCombo` objects should not just be based on their `displayName`. If such comparison is used somewhere that code belongs to that context but not here as the default order. I also find it odd that it seems to implement `Comparable<IdentifiableObject>` and not `Comparable<CategoryCombo>`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
jbee: Yes, good points, we either have to unify the naming to be independent of the table or this has to be duplicated in all objects because the FK name is individual.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
jbee: We should not, the IDE does have a warning for it as well.

jbee (approved):
Some small things to sort out with the compare and FK naming but generally good to go.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
david-mackessy: Good spot, there is already an `equals` method in this class for comparisons ([backed by tests](#19803)).
I wonder is this a weird legacy hack being kept alive? Could we get rid of it?
Where does this method get called from?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
vietnguyen: The foreign key is optional here. We can remove it. If not specify then hibernate will auto generate a name for the foreign key. It is used by hibernate for generating DDL, which in our project only used in h2 tests.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
vietnguyen: Fixed

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
vietnguyen: I move this compareTo method to a default method in the interface IdentifiableObject. will need to investigate and remove it in another PR.

david-mackessy (approved):

PR #20988, opened 2025-05-18
feat: new app bundling

netroms:


dhis-2/dhis-web-server/src/main/java/org/hisp/dhis/web/appbundler/AppBundler.java
```diff
@@ -0,0 +1,438 @@
...
+
+    // Write bundle info file
+    writeBundleFile(targetArtifactDir);
+
+    log.info("App bundling process completed successfully");
+  }
+
+  private void downloadApps(
+      List<AppGithubRepo> appRepoInfos, Path artifactDirPath, Path checksumDirPath) {
+    ExecutorService executor = Executors.newFixedThreadPool(DOWNLOAD_POOL_SIZE);
```
jbee, line 152: You could also consider using `parallelStream` to make the download parallel without adding much code level ceremony.

jbee (commented):
I wish we could avoid adding the `AppInfo` into the `App` as temporary field but then we can always take care of such code level details in another PR. Main thing is that the mechanism is a big improvement.

dhis-2/dhis-web-server/src/main/java/org/hisp/dhis/web/appbundler/AppBundler.java
netroms: Done!

netroms:
> I wish we could avoid adding the `AppInfo` into the `App` as temporary field but then we can always take care of such code level details in another PR. Main thing is that the mechanism is a big improvement.

Agree, I've refactored it out and made the App class untouched.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -268,16 +281,37 @@ public App installApp(File file, String filename, Cache<App> appCache) {
...
+      //      // make sure any other version of same app is removed
+      //      List<App> otherVersions = new ArrayList<>();
+      //      String key = app.getKey();
+      //      String version = app.getVersion();
+      //      discoverInstalledApps(
+      //          other -> {
+      //            if (key.equals(other.getKey()) && !version.equals(other.getVersion()))
+      //              otherVersions.add(other);
+      //          });
+      //      otherVersions.forEach(this::deleteAppAsync);
```
david-mackessy: Is the old version not being deleted anymore?
If I remember correctly, @jbee fixed a bug about loading old (incorrect) app versions a year or so ago. If there's no tests for that, this could be a regression. Unless it's handled somewhere else now?

netroms: I believe it will always get overwritten anyway, there are no effective versioning as far as I can see. Need to confirm this, hence why I've left the comment.

Birkbjo: @david-mackessy are you maybe referring to https://github.com/dhis2/dhis2-core/pull/16488 ? This caused the backend to serve another version in some cases due to a bug with authorities.

david-mackessy: That's different to the one I was thinking of but still good to call out 👍. This is the bug I was thinking of #14977 .
@netroms thanks for the explanation.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppBundleInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /** Class representing information about a single bundled app. */
+  @Data
+  public static class BundledAppInfo {
+    @JsonProperty private String name;
+    @JsonProperty private String url;
+    @JsonProperty private String branch;
+    @JsonProperty private String etag;
+    @JsonProperty private String downloadDate;
```
david-mackessy, line 62: why use a String here? could LocalDate/LocalDateTime be used instead?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
```diff
@@ -0,0 +1,120 @@
...
+      Resource[] resources = resolver.getResources(ZIPPED_APPS_PATH);
+      return Arrays.stream(resources)
+          .collect(Collectors.toMap(Resource::getFilename, Function.identity()));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static AppBundleInfo getAppBundleInfo() {
+    InputStream appBundleInfoInputStream = getAppBundleInfoInputStream();
```
david-mackessy: Is the stream being handled/closed correctly?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
```diff
@@ -0,0 +1,120 @@
...
+      PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+      Resource[] resources = resolver.getResources(APPS_BUNDLE_INFO_PATH);
+
+      // Ignore if not exists, this is the case when running tests
+      if (resources.length == 0 || resources[0] == null || !resources[0].exists()) {
+        log.warn(String.format("Bundled apps info file not found at: '%s'", APPS_BUNDLE_INFO_PATH));
+        return null;
+      }
+
+      return resources[0].getInputStream();
```
david-mackessy, line 182: are the streams being handled/closed correctly?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -341,34 +409,46 @@ public List<App> filterApps(List<String> filters, String contextPath) {
...
+    return app;
+  }
 
-    if (app.getAppState().ok()) {
-      installApp(app);
+  @Nonnull
+  public App installBundledAppResource(
+      Resource resource, String fileName, BundledAppInfo bundledAppInfo) {
+    try {
+      Path tempFile = Files.createTempFile("tmp-bundled-app-", fileName);
```
david-mackessy: what happens to the temp file? is it cleaned up/deleted?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/JCloudsAppStorageService.java
```diff
@@ -115,28 +134,21 @@ private void discoverInstalledApps(Consumer<App> handler) {
         app.setAppStorageSource(AppStorageSource.JCLOUDS);
         app.setFolderName(resource.getName());
 
-        handler.accept(app);
+        Blob bundledAppInfo = jCloudsStore.getBlob(resource.getName() + BUNDLED_APP_INFO_FILENAME);
+        if (bundledAppInfo != null) {
+          InputStream bundledAppInfoStream = bundledAppInfo.getPayload().openStream();
```
david-mackessy: is this stream being closed?

david-mackessy (commented):
I like the overall idea 👍, much better than downloading/cloning the repo.
There's a lot of work with streams that aren't being used in try with resource blocks (comments left).
Was it tested using MinIO also? Any app changes should always be tested using both types of file storage (local file system / object storage). There's [a guide here](https://github.com/dhis2/wow-backend/blob/master/guides/minio_storage.md) for using MinIO.

dhis-2/dhis-web-server/src/test/java/org/hisp/dhis/web/appbundler/AppBundlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+    // Settings app should be downloaded
+    Path importExportAppZip = downloadArtifactDir.resolve("import-export-app.zip");
+    assertTrue(Files.exists(importExportAppZip), "Import export app ZIP should exist");
+
+    // Settings app should be downloaded
+    Path settingsAppZip = downloadArtifactDir.resolve("data-quality-app.zip");
+    assertTrue(Files.exists(settingsAppZip), "Settings app ZIP should exist");
+
+    // Dashboard app should be downloaded
+    Path dashboardAppZip = downloadArtifactDir.resolve("aggregate-data-entry-app.zip");
```
Philip-Larsen-Donnelly: The comments and the names of the zip files here are confusing me.

dhis-2/dhis-web-server/src/main/java/org/hisp/dhis/web/appbundler/AppBundler.java
```diff
@@ -0,0 +1,447 @@
...
+      String appListPath = System.getProperty("APP_LIST");
+      String defaultBranch = System.getProperty("DEFAULT_BRANCH");
+      info("DOWNLOAD_DIR: {}", downloadDir);
+      info("BUILD_DIR: {}", buildDir);
+      info("ARTIFACT_ID: {}", artifactId);
+      info("APP_LIST: {}", appListPath);
+      info("DEFAULT_BRANCH: {}", defaultBranch);
+
+      if (downloadDir == null || buildDir == null || artifactId == null || appListPath == null) {
+        error("System properties DOWNLOAD_DIR, BUILD_DIR, ARTIFACT_ID, and APPS must be set");
```
Philip-Larsen-Donnelly:
```suggestion
        error("System properties DOWNLOAD_DIR, BUILD_DIR, ARTIFACT_ID, and APP_LIST must be set");
```

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -117,49 +112,105 @@ public class DefaultAppManager implements AppManager {
...
+          String appKey = app.getKey();
+          installedApps.computeIfAbsent(
+              appKey,
+              x ->
+                  Pair.of(
+                      installBundledAppResource(zipFileResource, bundledAppInfo), bundledAppInfo));
+
+          // If the bundled app is already installed and the Etag is different, overwrite the
+          // existing.
+          if (installedApps.containsKey(appKey)) {
```
kabaros, line 178: > If the bundled app is already installed and the Etag is different, overwrite the

when would this be the case?  wouldn't it just be an overridden app in that case and follows the JClouds path in the code.. I don't understand how a bundled app would be already installed, but needs to be overwritten as a bundled app still

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -117,49 +112,105 @@ public class DefaultAppManager implements AppManager {
...
+  private void cacheApp(@Nonnull App app) {
+    if (app.getAppState() == AppStatus.OK) {
+      appCache.put(app.getKey(), app);
+      registerDatastoreProtection(app);
+    }
+  }
 
   private Stream<App> getAppsStream() {
-    return appCache.getAll().filter(app -> app.getAppState() != AppStatus.DELETION_IN_PROGRESS);
+    return appCache.getAll();
```
kabaros, line 204: I am not too familiar with this bit, but from what I remember this check for DELETION_IN_PROGRESS was necessary to avoid some race condition on deletion, is that not an issue anymore? https://github.com/dhis2/dhis2-core/pull/20171/commits/f38ff55f0215f995425b0f197c0991af1067c4dc

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * <p>It detects .zip files in the bundled apps directory and installs them using the AppManager.
+ *
+ * <p>It also caches the the AppBundleInfo file and the list of apps names it contains for fast
+ * lookup.
+ */
+@Slf4j
+@RequiredArgsConstructor
+@Component
+public class BundledAppManager {
```
kabaros, line 71: I find the name for this class  a bit confusing - In general,  it would have been nice if it was another implementation of the `AppManager` interface and we'd have a BundledManager and a DefaultManager (or JCloudManager), or this whole new logic lives in a `bundledAppStorageService` so that it is inline with `jClouds` implementation, but it's somwhere in the middle now .. Both of these options might be too much, but in that case, I would still say that the name implies that it implements the AppManager interface while it's really mostly a service used internally by the `DefaultAppManager`.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @CheckForNull
+  public static InputStream getAppBundleInfoInputStream() {
+    try {
+      PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+      Resource[] resources = resolver.getResources(APPS_BUNDLE_INFO_PATH);
+      // Ignore if not exists, this is the case when running tests
+      if (resources.length == 0 || resources[0] == null || !resources[0].exists()) {
+        log.error("Bundled apps info file not found at: {}", APPS_BUNDLE_INFO_PATH);
```
kabaros, line 179: I guess the absence of bundled apps is handled gracefully all the way to the top - that was also the case before this PR. I am just wondering that - now that the process is much lighter with the ditching of node and cloning - whether things should blow up if no bundled apps are found?

Of course, this is from my point of view where I am almost always interested in the apps, but I guess others might not be - so feel free to ignore this comment.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -352,36 +406,55 @@ public List<App> filterApps(List<String> filters, String contextPath) {
...
-            "Installed App with ID %s (status: %s)", app.getAppHubId(), app.getAppState()));
+            "Installed App with AppHub ID %s (status: %s)", app.getAppHubId(), app.getAppState()));
+    cacheApp(app);
+    return app;
+  }
 
-    if (app.getAppState().ok()) {
-      installApp(app);
+  @Nonnull
+  public App installBundledAppResource(
```
kabaros, line 437: just to be consistent with other methods, I would think this logic to be part of a `BundledAppStorageService` similar to `JCloudsAppStorageService`?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -566,26 +562,37 @@ public ResourceResult getAppResource(App app, String pageName, String contextPat
 
     if (resource instanceof ResourceResult.ResourceFound resourceFound) {
       if (pageName.equals("/manifest.webapp")) {
-        // If request was for manifest.webapp, check for * and replace with host
-        if (app.getActivities() != null
-            && app.getActivities().getDhis() != null
-            && "*".equals(app.getActivities().getDhis().getHref())) {
-          app.getActivities().getDhis().setHref(contextPath);
-        }
+        app.getActivities().getDhis().setHref(contextPath);
```
kabaros, line 565: is `app.getActivities().getDhis()` never null? (that would be good to know for me for fixing another bug)

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -566,26 +562,37 @@ public ResourceResult getAppResource(App app, String pageName, String contextPat
...
-
         ByteArrayOutputStream bout = new ByteArrayOutputStream();
         template.apply(resourceFound.resource().getInputStream(), bout);
-
         ByteArrayResource byteArrayResource =
             toByteArrayResource(bout.toByteArray(), resourceFound.resource());
         return new ResourceResult.ResourceFound(byteArrayResource, "text/html;charset=UTF-8");
+      } else if (pageName.endsWith(".js")
+          || (resourceFound.resource().getFilename() != null
+              && resourceFound.resource().getFilename().endsWith(".js"))) {
```
kabaros, line 582: nit: I guess all these double checks for `null` then `endsWith` (or other string operations) can be replaced by null-safe `StringUtils.endsWith` - there are plenty of those though around these App services, so maybe something for later

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -566,26 +562,37 @@ public ResourceResult getAppResource(App app, String pageName, String contextPat
...
+              && resourceFound.resource().getFilename().endsWith(".html"))) {
         AppHtmlTemplate template = new AppHtmlTemplate(contextPath, app);
-
         ByteArrayOutputStream bout = new ByteArrayOutputStream();
         template.apply(resourceFound.resource().getInputStream(), bout);
-
         ByteArrayResource byteArrayResource =
             toByteArrayResource(bout.toByteArray(), resourceFound.resource());
         return new ResourceResult.ResourceFound(byteArrayResource, "text/html;charset=UTF-8");
+      } else if (pageName.endsWith(".js")
```
kabaros, line 580: just curious why we need special handling for `.js` now?

dhis-2/dhis-web-server/src/main/java/org/hisp/dhis/web/appbundler/AppBundler.java
```diff
@@ -0,0 +1,446 @@
...
+      throws IOException {
+    BundledAppInfo appBundleResultInfo = new BundledAppInfo();
+    appBundleResultInfo.setName(repoInfo.repo());
+    appBundleResultInfo.setUrl(repoInfo.originalUrl());
+    appBundleResultInfo.setBranch(repoInfo.ref());
+
+    try {
+      String zipName = repoInfo.repo() + ".zip";
+      Path zipFilePath = targetDir.resolve(zipName);
+      Path etagFile = etagsDir.resolve(repoInfo.repo() + ".etag");
```
kabaros, line 316: I was a bit worried that the zip url generated (i.e. `https://codeload.github.com/d2-ci/reports-app/zip/refs/heads/master`) doesn't work with all repos, but it seems it is generic and it does. This flow is useful for us when working on things like manifest changes or translations or global shell to point the `apps-to-bundle` to a special repo with a special version of a bundled app for testing.

In general, I just wanted to highlight the usecase from a developer point of view who's working mostly with apps serving - so the move to `.zip` files makes it harder to test changes in the manifest as before when we could just change the file .. it's not a big deal if we have to unzip then zip, but an extra step (I haven't tested yet, but I hope the checksum is not enforced beyond the initial download?)

kabaros (commented):
some general comments - I will do more testing with different scenarios (bundled, installed, overridden etc...) and confirm it all works as expected

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: This happens when there is a new version, or a change in the say, for example: https://github.com/d2-ci/capture-app, repo.
It works as follows:
1. During the Maven build we check the repo for new changes, by comparing the last saved Etag against the current one of the repo. If there is a new Etag aka. a change, we download the new .zip file of the app and save the Etag.
2. When we start the server, we compare the existing installed app against the new Etag, if the Etags don't match we install the new one.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: This is not necessary because all apps now are installing in a unique named folder, and there can not be any more collisions.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
netroms: Yes, it's more of a service used by the DefaultAppManager, but not just that. It is also used from a couple of other places, like the app override filter. That's why I wanted it as s separate "thing", "managing" the bundled apps.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: I don't feel the same way, because I see the "StorageService" concept as where you store the apps, the bundled apps .zip files are just build into the .war file. You install the apps into the JCloud, or before the LocalStorage, the previous bundled app service made sense because you were using the special "storage" for runtime actions, like reading the actual files directly from the BundledAppStorage.
This is not the case anymore, all apps are now stored in the JCloudStorage.
The bundled apps are just .zip files, that we read during the installation phase.

netroms: But I understand this method name and placeing is a bit confusing, maybe we can try to move it into the BundledAppManager

dhis-2/dhis-web-server/src/main/java/org/hisp/dhis/web/appbundler/AppBundler.java
netroms: Well, when you develop locally, you can now just modify the installed apps directly in the DHIS2_HOME apps folder, that would be even possible during runtime.

netroms: Any GitHub repo supports .zip download, you can still change the "apps-to-bundle", you can even add new apps there now without having to change the source code.

netroms: If you point the existing bundled app definition in the apps-to-bundle to a test repo, as long as the test repo has the same name aka. key as the installed bundled app (in the manifest), it will then be downloaded and overwrite the installed one, because the Etags will be different.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
netroms: if no apps are found, the server will still build and start, but you won't have any login app and can't really do much for a UI perspective.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: I see that @Nonnull was added to the getter recently by @amcgee , so I assume at least that is the intention.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: Yes, agree. I'm trying to keep this PR small, so I would prefer to do further non-critical refactorings in a separate PR dedicated to refactorings.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
netroms: This is to fix a corner case with the Approval app, I needed to force/overwrite it to change the basepath it uses for the API calls.

netroms: I'm not happy with this hack, and I hope someone from the front-end might be able to have a look at app source code and apply a solution there instead.

PR #21003, opened 2025-05-19
perf: do multi-values upserts for min-max values

jbee:


jason-p-pickering:
Test Setup:
Environment: SL database
Profiling Tool: JProfiler
Payload: ~14,000 values
Requests: 10 sequential API calls per endpoint (/upsert and /delete)
Baseline (Master): Many single-row operations
Feature Branch: Fewer multi-row operations

Master (single row inserts)
Selection_989
Selection_988
 Average times: 
 Upserts: 796 ms
 Deletes: 512 ms
 
 Feature branch: 
 
Selection_987
Selection_986
Average times: 
Upserts: 416 ms
Deletes: 346 ms



Operation | Master Avg Time | Feature Branch Avg Time | Improvement
-- | -- | -- | --
Upsert | 796 ms | 416 ms | 48% faster
Delete | 512 ms | 346 ms | 32% faster

Backend DB Load
Master:
137,160 INSERT/DELETE operations executed
High JDBC time (e.g., INSERT INTO minmaxdataelement: 5708 ms total)
Many small, fast operations → high overhead

Feature Branch:
~270 INSERT/DELETE operations executed
Total time dropped significantly (INSERT total: 2,307 ms)
Each query processes multiple rows, dramatically reducing total query count and overhead

In spite of the additional code complexity, if we are going to raw performance, then this multi-row approach may offer significant performance advantages.

jason-p-pickering (approved):
Looks good to me. I have profiled and the performance looks like a win. Tested via the front-end data entry app, and that seems to work fine as well.

david-mackessy (approved):

PR #21008, opened 2025-05-20
fix: Check compulsory data when completing a DataSet [DHIS2-18295]

david-mackessy:
## Issues 
1. The API `POST` `/api/dataEntry/dataSetCompletion` allows completing a `DataSet` when required compulsory data element operands are missing data.

2. ** Bonus bug **  While testing (same endpoint), another bug was found (after above fix was in place) whereby a `DataSet` with compulsory data elements, with data, could not be completed.

## Causes
1. There is no check for missing compulsory data element operands when completing a `DataSet`.

2. The `Period` being used in the SQL query to find data values had an ID of 0, resulting in no data being found when there is in fact data.

## Fix
1. Use an existing check for missing compulsory data element operands when completing a `DataSet`.

2. Load the `Period` so the correct `Period` ID is used in the SQL where clause.

## Testing
Automated controller tests added for:
- `DataSet` without any compulsory elements
- `DataSet` with compulsory elements & are required (no data filled)
- `DataSet` with compulsory elements & are required (with data filled)
- `DataSet` with compulsory elements & are not required

## Background Info
It seems the endpoint `POST` `/api/dataEntry/dataSetCompletion` is actually a [new-ish endpoint](#11673) for single `DataSet` completion.
The old (bulk) endpoint is `POST` `/api/completeDataSetRegistrations`, which does have the check for missing compulsory data element operands.


jbee (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/dataset/DefaultCompleteDataSetRegistrationService.java
```diff
@@ -113,9 +120,41 @@ public void saveCompleteDataSetRegistration(CompleteDataSetRegistration registra
...
+              registration.getPeriod(),
+              registration.getSource(),
+              registration.getAttributeOptionCombo());
+      if (!missingDataElementOperands.isEmpty()) {
+        String deos =
+            missingDataElementOperands.stream()
+                .map(DataElementOperand::getDisplayName)
+                .collect(Collectors.joining(","));
+        throw new IllegalStateException(
+            "All compulsory data element operands need to be filled: [%s]".formatted(deos));
```
larshelge, line 148: Should we instead throw an exception with an `ErrorCode` here to make it consistent with other validation errors, and allow programmatic client handling? This state is not a "programming error" and could happen e.g. due to user error in an external client.

jbee: I was about to also leave a comment on error handling but I didn't because I think most if the code for this endpoint/operation does not really follow our more recent goals and I think we better make a PR that moves all the error handling in one PR instead of making it a mixed solution by doing this one in line while other error handling is not.

david-mackessy: - this error message is consistent with how the mature/bulk API for completing data sets works
  - the bulk API returns a report summary with conflicts (no error code)
  - the single/newish API was implemented to not return any response body/summary
    - it doesn't look like we tried (or thought about) keeping similar behaviour (validation, errors, response etc.) between these APIs for whatever reason
  - if we add an error code in this fix, then behaviour (what little there is) would be inconsistent with the bulk API
- we could change both to return error codes for this particular issue
  - probably best doing that in a separate PR
  - we could improve the single API to return a report also to try to keep things more consistent
  - best not to change behaviour of something else during a bug fix

netroms (approved):

PR #21022, opened 2025-05-22
feat: backport of bulk min-max imports

jason-p-pickering:


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/MinMaxDataElementControllerTest.java
jbee, line 1: I don't think this should be deleted, or was it renamed or moved?

jbee (commented):
Looks good apart from the delete test :thinking:

PR #21040, opened 2025-05-24
fix: PI Disaggregation scaling bugs [DHIS2-19663]

jimgrace:
See [DHIS2-19663](https://dhis2.atlassian.net/browse/DHIS2-19663). This fix addresses two problems discovered during PI Disaggregation scaling testing:
1. Importing metadata fails when the metadata includes both program category mappings and new category options those mappings reference. (`ProgramObjectBundleHook` was calling `IdentifiableObjectManager` with the assumption that the category options already existed.) The fix is to remove the checks for existing categories and options during import, to check for valid (and non-duplicating) category option UIDs, and to continue the full check at analytics runtime.
2. Evaluating a disaggregated program indicator failed when there were more than 40 category options because the Postgres `concat()` function (like all other Postgres functions) accepts a maximum of 40 arguments. The new code uses the PostgreSQL string concatenation operator `||` instead of the `contact()` function.

This PR also adds the e2e `AnalyticsPiDisaggregationTest` that creates a program indicator able to disaggregate a category with 43 options (to demonstrate that more than 40 options can be disaggregated). To allow this, the default value for `METADATA_CATEGORIES_MAX_OPTIONS` is increased from 31 to 50. This is a new maximum for v42 (see [DHIS2-18521](https://dhis2.atlassian.net/issues/DHIS2-18521)). When the PR is merged, the documentation will also be updated (see [dhis2-docs/pull/1464](https://github.com/dhis2/dhis2-docs/pull/1464/files)).

Also, `transient CategoryOption option` is removed from `ProgramCategoryOptionMapping`. (It was declared but ultimately never used.)

[DHIS2-19663]: https://dhis2.atlassian.net/browse/DHIS2-19663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-18521]: https://dhis2.atlassian.net/browse/DHIS2-18521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhis-2/olor=auto
```diff
@@ -0,0 +1,5 @@
+# Netscape HTTP Cookie File
```
maikelarabori, line 1: I think this file is a mistake.

maikelarabori (approved):
Looks good @jimgrace.
At the end of the PR, there is file that should be removed...

dhis-2/olor=auto
jimgrace: Thanks, I usually catch that but missed it this time. Not sure whey it keeps appearing in my source tree! :)

PR #21049, opened 2025-05-26
fix: Events analytics query now running in Doris

luciano-fiandesio:
### Summary

Aligned the Events analytics sql query generation with the approach used for Enrollment analytics (use CTEs instead of Sub-query, when needed) in order to have compatibility with Apache Doris and Clickhouse.

### Details

Events Analytics sql queries are generally more compatible with Apache Doris and Clickhouse, because they do not employ nested-subqueries. The only time that invalid subqueries are generated, is when an Event query references a Program Indicator of type Enrollment.

#### Using `CROSS JOIN`

The events analytics query was refactored to compute a metric in a CTE that was still `GROUP BY` enrollment and later joined on enrollment.
Result: the metric became per-enrollment, so one event stream showed many different numbers instead of the single constant we expect.

#### Fix
Drop the `GROUP BY` so the CTE returns exactly one row, then CROSS JOIN it to the event "stream". A cross-join to a one-row set just “broadcasts” that constant to every event row, restoring the original (scalar-subquery) behaviour without adding cost.

#### Other changes

- fix Apache Doris `log` function

### DIFF
--- a/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/ValidationHelper.java
+++ b/dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/analytics/ValidationHelper.java
@@ -35,7 +35,6 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
-import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.fail;
@@ -58,8 +57,6 @@ import org.hisp.dhis.test.e2e.dto.ApiResponse;
  */
 @UtilityClass
 public class ValidationHelper {
-
-  // --- Core Utility Methods ---
 
   /**
    * Finds a header map by its 'name' property within the extracted list of headers.
@@ -98,8 +95,6 @@ public class ValidationHelper {
             new AssertionError(
                 "Header with name '" + headerName + "' not found, cannot determine index."));
   }
-
-  // --- Response Structure Validation ---
 
   /**
    * Validates the overall structure (header count, width) based on whether PostGIS columns are
@@ -141,8 +136,6 @@ public class ValidationHelper {
         .body("width", equalTo(expectedSize))
         .body("headerWidth", equalTo(expectedSize));
   }
-
-  // --- Header Validation Methods ---
 
   /**
    * Asserts whether a header with the given name exists or does not exist in the response.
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/common/AnalyticsQueryType.java
```diff
@@ -0,0 +1,35 @@
...
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.analytics.common;
+
+public enum AnalyticsQueryType {
```
maikelarabori, line 32: I think we could `EndpointItem` enum, instead.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -1703,4 +1737,710 @@ private boolean hasNoAggregationType(EventQueryParams params) {
...
+    }
+    return operator;
+  }
+
+  private boolean isEqualityOperator(QueryOperator operator) {
+    return operator == QueryOperator.EQ || operator == QueryOperator.IEQ;
+  }
+
+  private String getSqlFilterValue(QueryFilter filter, QueryItem item) {
+    if ("NV".equals(filter.getFilter())) {
```
maikelarabori, line 2238: Is this constant reachable at this point?
`QueryKey.NV`?

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -1703,4 +1737,710 @@ private boolean hasNoAggregationType(EventQueryParams params) {
...
+        addFilterJoin(builder, itemUid, cteDef);
+      }
+    }
+  }
+
+  private void addProgramStageJoins(SelectBuilder builder, String itemUid, CteDefinition cteDef) {
+    for (Integer offset : cteDef.getOffsets()) {
+      String alias = cteDef.getAlias(offset);
+      String joinCondition =
+          alias + ".enrollment = ax.enrollment AND " + alias + ".rn = " + (offset + 1);
```
maikelarabori, line 1952:
```suggestion
          alias + ".enrollment = ax.enrollment and " + alias + ".rn = " + (offset + 1);
```

maikelarabori (approved):
Looks good 👍 
Just minor comments, thx

vietnguyen (approved):

PR #21051, opened 2025-05-26
fix: use server.base.url for OAuth2 issuer URI

netroms:
**Problem:**
1. When DHIS2 is deployed behind a reverse proxy that terminates SSL, the OAuth2 authorization server (enabled via `oauth2.server.enabled`) generates JWTs with an `iss` (issuer) claim starting with `http://` instead of `https://`. This is because the issuer URI is inferred from the internal request to Tomcat, which is HTTP.
2. The `AuthorizationServerConfig.java` was not conditionally loaded based on the `OAUTH2_SERVER_ENABLED` setting, meaning its beans might be initialized even if the feature was disabled.

**Impact:**
1. The incorrect issuer URI can cause issues with token validation for clients and resource servers that expect the issuer URI to match the public-facing HTTPS URL.
2. Unnecessary initialization of OAuth2 server components if the feature is disabled.

**Solution:**
1. The `AuthorizationServerSettings` in `AuthorizationServerConfig.java` has been updated to explicitly set the issuer URI using the `server.base.url` (from `ConfigurationKey.SERVER_BASE_URL`) configuration property. This ensures the correct public-facing HTTPS URL is always used for the issuer claim in generated JWTs.
2. `AuthorizationServerConfig.java` is now annotated with `@Conditional(AuthorizationServerEnabledCondition.class)`, ensuring its beans are only loaded if `ConfigurationKey.OAUTH2_SERVER_ENABLED` is true.

**File Modified:**
* `dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java`

**Verification Steps:**
1. Confirm `server.base.url` in `dhis.conf` is set to the correct HTTPS base URL.
2. Confirm `oauth2.server.enabled` in `dhis.conf` is set to `on`.
3. Obtain an access/refresh token using the DHIS2 OAuth2 server.
4. Decode the JWT and verify the `iss` claim uses the `https://` scheme and matches the `server.base.url`.
5. Set `oauth2.server.enabled` to `off` and verify that the OAuth2 server endpoints (e.g., `/oauth2/authorize`) are not available and its beans are not loaded.

### DIFF
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java
@@ -48,7 +48,6 @@ import java.util.Set;
 import java.util.UUID;
 import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
-import org.hisp.dhis.condition.RedisEnabledCondition;
 import org.hisp.dhis.external.conf.ConfigurationKey;
 import org.hisp.dhis.external.conf.DhisConfigurationProvider;
 import org.hisp.dhis.security.oidc.KeyStoreUtil;
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java
@@ -48,6 +48,7 @@ import java.util.Set;
 import java.util.UUID;
 import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/security/config/AuthorizationServerConfig.java
```diff
@@ -217,6 +219,8 @@ public JwtDecoder jwtDecoder(JWKSource<SecurityContext> jwkSource) {
 
   @Bean
   public AuthorizationServerSettings authorizationServerSettings() {
-    return AuthorizationServerSettings.builder().build();
+    return AuthorizationServerSettings.builder()
+        .issuer(config.getProperty(ConfigurationKey.SERVER_BASE_URL))
```
david-mackessy, line 223: We reviewed the usage of `server.base.url` recently.
Check out [this PR](#20227) to see if it's relevant and want to keep consistent behaviour in this scenario also.

PR #21056, opened 2025-05-27
chore: Reuse org unit predicate in stores [DHIS2-19645]

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OrgUnitJdbcPredicate.java
```diff
@@ -0,0 +1,168 @@
...
+import lombok.RequiredArgsConstructor;
+import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.commons.util.SqlHelper;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public class OrgUnitJdbcPredicate {
+
+  public static void buildOrgUnitModeClause(
```
enricocolasante, line 48: This and the next method uses side-effects to build the SQL and populate the SQL parameters.
I think it is very important that `StringBuilder sql` and `MapSqlParameterSource sqlParameters` are always the first 2 parameters in this kind of methods.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OrgUnitJdbcPredicate.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.Set;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+import org.hisp.dhis.common.OrganisationUnitSelectionMode;
+import org.hisp.dhis.commons.util.SqlHelper;
+import org.hisp.dhis.organisationunit.OrganisationUnit;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public class OrgUnitJdbcPredicate {
```
enricocolasante, line 46: We should have extensive JavaDoc here to make it clear when and how use this.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OrgUnitJdbcPredicate.java
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+  }
+
+  public static void buildOwnershipClause(
+      StringBuilder sql,
+      OrganisationUnitSelectionMode orgUnitMode,
+      MapSqlParameterSource sqlParameters,
+      Set<OrganisationUnit> effectiveSearchOrgUnits,
+      Set<OrganisationUnit> captureScopeOrgUnits,
```
enricocolasante, line 73: These parameters are not coming from the user input, could we calculate them inside the method?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/HibernateTrackedEntityStore.java
```diff
@@ -340,6 +338,8 @@ private void addTrackedEntityFromItem(
       TrackedEntityQueryParams params,
       PageParams pageParams,
       boolean isCountQuery) {
+    SqlHelper sqlHelper = new SqlHelper(true);
```
enricocolasante, line 341: This variable definition could remain closer to its first use to help readability.

enricocolasante (commented):
`FilterJdbcPredicate` idea was to move the sql generation when mapping to the query params and then only add or not such SQL in the query in order to make the SQL itself unit testable and to improve readability and composition.
Maybe it would make sense to try to have a static constructor for `OrgUnitJdbcPredicate` that will create an object holding the SQL and the possible parameters passed by the user.
Then, it should be easy to create a unit test and assert against the SQL to check all possible generated queries.

enricocolasante (approved):

PR #21089, opened 2025-05-29
chore: Remove unused methods in event service [DHIS2-19578]

enricocolasante:
***Big picture***
The final goal of this ticket is to have separate services and stores for single events and tracker events.
To achieve this we will need an intermediate phase where some endpoints will use the correct services and others that will still use the current event service that is serving both types of events.
As a final step, we will remove the generic event service and store and we will only have the ones for single and tracker events separated.

***Implemented in this PR***
- Refactor event exporter tests to use `TrackerEventService` or `SingleEventService`:
  - Add test cases in `FilterExporterTest`
  - Add test cases in `OrderAndPaginationExporterTest`
  - Remove `EventExporterTest`. Specific `single` and `tracker` versions were created in previous PR.
  - Remove `AclTrackerEventExporterTest` and create specific `single` and `tracker` versions.
- Remove all unused methods in `EventService`.

***Next steps***
- Use the right services in `getEventChangeLogsByUid` in `EventExportController`.
- Remove `export.event` package from `dhis-service-tracker` module.


dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/base_data.json
```diff
@@ -982,12 +994,29 @@
...
+        },
+        {
+          "dataElement": {
+            "idScheme": "UID",
+            "identifier": "DATAEL00005"
+          },
+          "value": "option1"
+        },
+        {
+          "created": "2022-04-22T06:00:30.559",
```
teleivo: Remove the server generated fields

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/base_metadata.json
```diff
@@ -2462,6 +2519,45 @@
...
+        {
+          "id": "DiszpKrYNg8"
+        }
+      ],
+      "userRoles": [
+        {
+          "id": "UbhT3bXWUyb"
+        }
+      ],
+      "username": "singleEventUser"
```
teleivo, line 2559: why do we have a separate user for single events?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/event/OrderAndFilterEventChangeLogTest.java
```diff
@@ -168,15 +168,17 @@ void shouldSortChangeLogsWhenOrderingByDataElementAsc() throws NotFoundException
             .getEventChangeLog(UID.of("kWjSezkXHVp"), params, defaultPageParams)
             .getItems();
 
-    assertNumberOfChanges(7, changeLogs);
+    assertNumberOfChanges(9, changeLogs);
```
teleivo, line 171: This to me suggests that changelogs should get their own meta-/data @muilpp to be independent. Every time we make a change we have to update changelog tests 😅

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/singleevent/SingleEventsExportControllerH2Test.java
```diff
@@ -67,15 +66,6 @@
 class SingleEventsExportControllerH2Test extends H2ControllerIntegrationTestBase {
```
teleivo: Why do need this test? 🤔

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/base_metadata.json
enricocolasante: It is easier to separate the different ACL to programs without registrations and orgUnits

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/singleevent/SingleEventsExportControllerH2Test.java
enricocolasante: We actually don't need this test to be duplicated.
I moved it to the common package and just use one of the services.

teleivo: Why do we need the test at all?

enricocolasante: To test that the `content-type` is correct in the header? :man_shrugging:

PR #21107, opened 2025-05-30
fix: Aggregate Data Exchange job creator as job executor [DHIS2-18324](2.40)

david-mackessy:
port of #20850 

This port needed extra work due to the very big differences between the job scheduling in older versions.
The issue is having no user associated with an async job (on a new thread). There is no easy way to automatically set the job creator as the user to be used when executing the job (A valid user is required for this job).

This fix uses the `lastUpdatedBy` field to set & retrieve the creator of the job. It's not nice but without making even more larger changes (DB migration for job config table) there is no other way to store and retrieve a user for jobs, without also requiring changes in clients.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
@@ -174,7 +174,10 @@ public class HibernateIdentifiableObjectStore<T extends BaseIdentifiableObject>
     object.setAutoFields();
 
     object.setAutoFields();
-    object.setLastUpdatedBy(user);
+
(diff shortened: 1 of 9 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/common/hibernate/HibernateIdentifiableObjectStore.java
```diff
@@ -174,7 +174,10 @@ public void update(@Nonnull T object, @CheckForNull User user) {
     object.setAutoFields();
 
     object.setAutoFields();
-    object.setLastUpdatedBy(user);
+
+    if (user != null) {
```
david-mackessy, line 178: Without this new check, the scheduler keeps overriding a valid user value with null, as at this point in the code, the job scheduler is executing on a new thread without a logged in user.

vietnguyen (approved):

PR #21128, opened 2025-06-03
fix: (2.40.8) )ensure program rule notifications correctly resolve dataElement [DHIS2-19685]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-19685

### DIFF
--- a/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEngineService.java
+++ b/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEngineService.java
@@ -85,7 +85,7 @@ public class DefaultProgramRuleEngineService implements ProgramRuleEngineService
     }
 
     List<ProgramRule> programRules =
-        programRuleEngine.getProgramRules(programInstance.getProgram());
+        programRuleEngine.getProgramRulesForEnrollment(programInstance.getProgram());
 
     if (programRules.isEmpty()) {
       return List.of();
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleService.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleService.java
@@ -99,6 +99,17 @@ public interface ProgramRuleService {
   List<ProgramRule> getProgramRulesByActionTypes(Program program, Set<ProgramRuleActionType> types);
 
   /**
+   * Returns the list of program rules associated with the given program that are not linked to any
+   * specific program stage. Only includes rules with action types supported by the backend rule
+   * engine.
+   *
+   * @param program the program to retrieve rules for
+   * @param types the set of supported {@link ProgramRuleActionType}s
+   * @return a list of applicable {@link ProgramRule}s for enrollment
+   */
+  List<ProgramRule> getProgramRulesForEnrollment(Program program, Set<ProgramRuleActionType> types);
+
+  /**
    * Get validation by {@link Program}
    *
    * @param program Program
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleStore.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/programrule/ProgramRuleStore.java
@@ -68,6 +68,8 @@ public interface ProgramRuleStore extends IdentifiableObjectStore<ProgramRule> {
 
   List<ProgramRule> getProgramRulesByActionTypes(Program program, Set<ProgramRuleActionType> types);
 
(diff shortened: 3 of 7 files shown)


dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEngineService.java
```diff
@@ -85,7 +85,7 @@ public List<RuleEffect> evaluateEnrollmentAndRunEffects(long enrollment) {
     }
 
     List<ProgramRule> programRules =
-        programRuleEngine.getProgramRules(programInstance.getProgram());
+        programRuleEngine.getProgramRulesForEnrollment(programInstance.getProgram());
```
enricocolasante, line 88: Can we simplify this fix and just filter out any program rule that has `programStage != null`?
So we do not need to add all this calls and create a new method in the store.

enricocolasante (commented):
I am confused about how this change can fix the issue.
We are just not evaluating some rules for the enrollment, but the issue is that a data value was not correctly evaluated.
First, I cannot see why the data element would not be correctly evaluated in the enrollment evaluation, we are passing all the right data values in the context to rule-engine, so it should work even if the rule shouldn't be run.
Secondly, if we remove the rule, we still need it to run for the right event, so when an event is completed or a data element is changed, and if that was working should have worked even before this change, either we were sending duplicated email or this change will just stop the email to be sent.

We should carefully test this fix to understand if everything is working fine.

zubaira:
> I am confused about how this change can fix the issue. We are just not evaluating some rules for the enrollment, but the issue is that a data value was not correctly evaluated. First, I cannot see why the data element would not be correctly evaluated in the enrollment evaluation, we are passing all the right data values in the context to rule-engine, so it should work even if the rule shouldn't be run. Secondly, if we remove the rule, we still need it to run for the right event, so when an event is completed or a data element is changed, and if that was working should have worked even before this change, either we were sending duplicated email or this change will just stop the email to be sent.
> 
> We should carefully test this fix to understand if everything is working fine.

Well i do not think so we are passing events/dataValues while evaluating enrollment. here you can see https://github.com/dhis2/dhis2-core/blob/5add8ac5ec863eb21bb53a7939c271327636fff7/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/ProgramRuleEngine.java#L73. 

So incase rule condition is `true` and has a `programStage` associated to it. then rule will execute while evaluating enrollment where as it should evaluate only when event is being processed.

Secondly duplicate emails will not happen because 
1. Rules associated with events will only be executed while processing events.
2. we have `externallogentry` to prevent duplicate emails.

dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/DefaultProgramRuleEngineService.java
zubaira: Filtering by programStage != null could work, but adding a new method keeps the logic isolated and safer for patch releases (which is i think today). It avoids risking regressions in shared code and makes testing more focused.

enricocolasante:
> > I am confused about how this change can fix the issue. We are just not evaluating some rules for the enrollment, but the issue is that a data value was not correctly evaluated. First, I cannot see why the data element would not be correctly evaluated in the enrollment evaluation, we are passing all the right data values in the context to rule-engine, so it should work even if the rule shouldn't be run. Secondly, if we remove the rule, we still need it to run for the right event, so when an event is completed or a data element is changed, and if that was working should have worked even before this change, either we were sending duplicated email or this change will just stop the email to be sent.
> > We should carefully test this fix to understand if everything is working fine.
> 
> Well i do not think so we are passing events/dataValues while evaluating enrollment. here you can see
> 
> https://github.com/dhis2/dhis2-core/blob/5add8ac5ec863eb21bb53a7939c271327636fff7/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/ProgramRuleEngine.java#L73
> 

The null is the event that will be used as target.
Here you can see that the events are passed in the context and the data values will be extracted from there.
https://github.com/dhis2/dhis2-core/blob/5add8ac5ec863eb21bb53a7939c271327636fff7/dhis-2/dhis-services/dhis-service-program-rule/src/main/java/org/hisp/dhis/programrule/engine/ProgramRuleEngine.java#L76.
This is the same logic when we evaluate an event, so if there is an issue here, we have for the enrollment we will have the same issue for events.

> .
> So incase rule condition is `true` and has a `programStage` associated to it. then rule will execute while evaluating enrollment where as it should evaluate only when event is being processed.
> 
> Secondly duplicate emails will not happen because
> 
> 1. Rules associated with events will only be executed while processing events.
> 2. we have `externallogentry` to prevent duplicate emails.

PR #20997, opened 2025-05-19
chore: Use program event services for get events [DHIS2-19578]

enricocolasante:
***Big picture***
The final goal of this ticket is to have separate services and stores for single events and tracker events.
To achieve this we will need an intermediate phase where some endpoints will use the correct services and others that will still use the current event service that is serving both types of events.
As a final step, we will remove the generic event service and store and we will only have the ones for single and tracker events separated.

***Implemented in this PR***
- Rename all occurrences of `programEvent` to `singleEvent`.
- Create new package `exporter.singleevent` in `dhis-service-tracker` module, duplicating service and store from `exporter.event` package.
- Use `singleEventService` to get events from `EventExportController` when the passed program is without registration. (`program` field is now mandatory).

***Next steps***
- Create new package `exporter.trackerevent` in `dhis-service-tracker` module, duplicating service and store from `exporter.event` package.
- Use `trackerEventService` to get events from `EventExportController` when the passed program is with registration. (`program` field is now mandatory).
- Use the right services in `getEventsAsJsonGzip` in `EventExportController`.  
- Use the right services in `getEventsAsJsonZip` in `EventExportController`.
- Use the right services in `getEventsAsCsv` in `EventExportController`.
- Use the right services in `getEventsAsCsvGZip` in `EventExportController`.
- Use the right services in ` getEventsAsCsvZip` in `EventExportController`.
- Use the right services in `getEventByUid` in `EventExportController`.
- Use the right services in `getEventDataValueFile` in `EventExportController`.
- Use the right services in `getEventDataValueImage` in `EventExportController`.
- Use the right services in `getEventChangeLogsByUid` in `EventExportController`.
- Remove `export.event` package from `dhis-service-tracker` module.

PR #21129, opened 2025-06-03
chore: Use event services for event changelog [DHIS2-19578]

enricocolasante:
***Big picture***
The final goal of this ticket is to have separate services and stores for single events and tracker events.
To achieve this we will need an intermediate phase where some endpoints will use the correct services and others that will still use the current event service that is serving both types of events.
As a final step, we will remove the generic event service and store and we will only have the ones for single and tracker events separated.

***Implemented in this PR***
- Use the right services in `getEventChangeLogsByUid` in `EventExportController`.

***Next steps***
- Remove `export.event` package from `dhis-service-tracker` module.


teleivo (commented):
Why are you removing all these classes?

enricocolasante:
> Why are you removing all these classes?

Because when I created the `singleevent`([here](https://github.com/dhis2/dhis2-core/pull/20997)) and `trackerevent` ([here](https://github.com/dhis2/dhis2-core/pull/21038)) packages I copied all the classes from `event` package, but then we are not going to need this duplication for changelogs.

teleivo (approved):

muilpp (approved):

PR #21142, opened 2025-06-04
DHIS2-19700: raise no. of allowed HTTP connections in Route API

cjmamo:
Raise no. of allowed HTTP connections in Route API to mitigate connection timeouts.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
@@ -71,8 +71,8 @@ import org.springframework.web.util.UriComponentsBuilder;
 @Slf4j
 @RequiredArgsConstructor
 public class RouteService {
-  public static final int MAX_TOTAL_HTTP_CONNECTIONS = 500;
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -70,6 +71,9 @@
 @Slf4j
 @RequiredArgsConstructor
 public class RouteService {
+  public static final int MAX_TOTAL_HTTP_CONNECTIONS = 500;
+  public static final int DEFAULT_MAX_HTTP_CONNECTION_PER_ROUTE = 50;
```
jbee: I know the "extract constant" refactoring makes the `public` by default but I think that is a bad thing. By default such constants should be `private` or at least package local. Otherwise people need to check usage later on when they want to refactor the code for some reason. Worst case they are actually used elsewhere hindering the refactor.

cjmamo: Will set them to protected since they're referenced from the test case.

PR #21151, opened 2025-06-05
feat: Add support for GZIPPED JSON/CSV min/max payloads

jason-p-pickering:
While testing this feature, it became fairly obvious that these CSV/JSON payloads of bulk min/max values can fairly easily exceed the limit imposed by the reverse proxy on body size. Adding support for GZIPPED payloads should speed things up a bit for larger payload sizes, without having to chunk things on the client side. 

@jbee I need your help/guidance with tests. It looks like the current HttpClientAdapter does not support binary bodies very well. I think you preferred the integration tests over the E2E tests where making a test for this might be a bit easier. Otherwise,I need some help there when you get a chance.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
@@ -163,6 +163,7 @@ public enum ErrorCode {
   E2045("Case insensitive operators can only be used with constant values"),
   E2046("Error parsing CSV file: {0}"),
   E2047("Data value key combination does not exist: {0}"),
+  E2048("Failure to read gzipped min max payload: {0}"),
 
   /* Outlier detection */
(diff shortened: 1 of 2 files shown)


jbee:
@jason-p-pickering here https://github.com/dhis2/dhis2-core/pull/21151/commits/3eb82ab94540fec0979cbcbd7c14c5363a3a094b you go

jason-p-pickering:
Thanks @jbee . Elegant as always. I think this one is good to go unless we want to also add support for gzipped delete payloads while we are at it?  Also, please have a look at the endpoint names to be sure they are consistent.

jbee:
yeah, we can do delete. I was also wondering about consistency with regards to using `MultipartFile` vs body. Now we have a mix. While it is easier to build a low tech HTML form with `file` input I think we have plenty of JS involved in our apps, using body seems no big deal

david-mackessy (approved):

PR #21152, opened 2025-06-05
fix: Avoid potential stack overflow

jason-p-pickering:
This PR was generated by Codex (ChatGPT)


jbee (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/java/org/hisp/dhis/db/migration/helper/JdbcSqlFileExecutor.java
```diff
@@ -54,11 +54,15 @@ public class JdbcSqlFileExecutor {
   private static final String DEFAULT_DELIMITER = ";";
 
   /**
-   * regex to detect delimiter. ignores spaces, allows delimiter in comment, allows an equals-sign
+   * Regex to detect delimiter in SQL scripts. Ignores spaces, allows delimiter in comments and
+   * supports an optional equals sign.
+   *
+   * <p>Previous implementation used nested repetition, which might trigger catastrophic
```
larshelge: I think we can remove the "previous implementation.." part from the Javadoc, as it is not very relevant here.

PR #21154, opened 2025-06-05
fix: use logger name in default log configuration

jbee:
### Summary
Using `%F` means _"the class name of the class that called the logger method"_.  I propose that the better and less confusing name to use is `%c{1}` which means the name of the logger (`%c`) as simple name (`{1}`), meaning if there is a package name in the logger name cut it out.

### Confusion
This can be very confusing if meta-programming is used to resolve different loggers. 
With `%F` the log line would state the name of the class containing the meta-logging code.
If someone would want to configure the log level based they are likely to mistakenly assume that the class name in the logs would be the target of the level change. But that is not correct. The target still needs to be selected based on the logger name.
Also using `%F` defies the purpose or usefulness of resolving different loggers because it does no longer matter which logger a log is written to, the name in the log entry will always be the one containing the code that calls the "logger.log" method.
That is why it is more useful and less confusing to use `%c` or `%c{1}`.

### Performance
Using `%F` is also discouraged since it has a larger performance overhead. This is because in order to find which class called the logger method the framework needs to throw a fake exception, catch it and look at the stack trace to find this out. This overhead does not exist with `%c`.


jbee:
@stian-sandvold , @Philip-Larsen-Donnelly please have a look and make up your mind about if we want to move forward with this change. Thanks.

PR #21161, opened 2025-06-07
feat(route): add support for additional response headers in RouteService

eirikhaugstulen:
Summary:
- Sometimes it would be really useful to allow preconfigured response headers that should be forwarded to the client. This is typically important when you do not control the upstream service, but rely on headers for functionality.
- I know this is potentially quite dangerous, so would love some extra focus on sec

- This PR adds a new field in dhis.conf that enables the system admin to define allowed response headers through the route API.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
@@ -43,6 +43,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/route/RouteService.java
```diff
@@ -511,4 +555,33 @@ private HttpHeaders filterHeaders(
...
+      throw new IllegalArgumentException("Header name cannot be null or empty");
+    }
+
+    String normalizedHeader = headerName.trim().toLowerCase();
+
+    if (!VALID_HEADER_NAME_PATTERN.matcher(headerName.trim()).matches()) {
+      throw new IllegalArgumentException("Header name contains invalid characters: " + headerName);
+    }
+
+    if (DANGEROUS_RESPONSE_HEADERS.contains(normalizedHeader)) {
```
larshelge, line 576: I guess "unsafe" is a bit more of an established term than "dangerous" :-)

PR #21176, opened 2025-06-09
bug: Add ACL validations in enrollment store [DHIS2-19644]

muilpp:
A few things in this PR:
- Validate ownership in the enrollment query, as done for events and TEs. This way we don't break pagination anymore.
- Validate TET when getting accessible programs.
- Don't validate user given TE, as that's data.
- With the `orgUnitQueryBuilder` helper in place, there's no need to merge org unit modes before the query anymore.
- Get enrollments by program owner, after the last refactor we were using the enrollment org unit. That's not the one used in ACL.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
@@ -146,7 +146,7 @@ class DefaultEnrollmentService implements EnrollmentService {
     EnrollmentQueryParams queryParams = paramsMapper.map(params, getCurrentUserDetails());
 
     List<Enrollment> enrollments =
-        findEnrollments(
+        mapEnrollment(
             new ArrayList<>(enrollmentStore.getEnrollments(queryParams)),
             params.getFields(),
             params.isIncludeDeleted());
@@ -165,7 +165,7 @@ class DefaultEnrollmentService implements EnrollmentService {
 
     Page<Enrollment> enrollmentsPage = enrollmentStore.getEnrollments(queryParams, pageParams);
     List<Enrollment> enrollments =
-        findEnrollments(enrollmentsPage.getItems(), params.getFields(), params.isIncludeDeleted());
+        mapEnrollment(enrollmentsPage.getItems(), params.getFields(), params.isIncludeDeleted());
 
     addTrackedEntityAudit(queryParams.getTrackedEntity(), enrollments);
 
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -257,19 +281,11 @@ private Set<TrackedEntityAttributeValue> getTrackedEntityAttributeValues(Enrollm
...
     List<Enrollment> enrollmentList = new ArrayList<>();
-    UserDetails currentUser = getCurrentUserDetails();
 
     for (Enrollment enrollment : enrollments) {
-      if (enrollment != null
-          && (orgUnitMode == ALL
-              || trackerOwnershipManager.hasAccess(
-                  currentUser, enrollment.getTrackedEntity(), enrollment.getProgram()))
-          && trackerAccessManager.canRead(currentUser, enrollment, orgUnitMode == ALL).isEmpty()) {
+      if (enrollment != null) {
```
enricocolasante: How can the enrollment be `null` here?
If it is possible, then we are still breaking pagination.

muilpp: That's a remnant of the old ACL validation and is no longer needed.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -277,21 +272,11 @@ private Set<TrackedEntityAttributeValue> getTrackedEntityAttributeValues(Enrollm
...
 
     for (Enrollment enrollment : enrollments) {
-      if (enrollment != null
-          && (orgUnitMode == ALL
-              || trackerOwnershipManager.hasAccess(
-                  currentUser, enrollment.getTrackedEntity(), enrollment.getProgram()))
-          && trackerAccessManager.canRead(currentUser, enrollment, orgUnitMode == ALL).isEmpty()) {
-        enrollmentList.add(getEnrollment(enrollment, fields, includeDeleted));
-      }
+      enrollmentList.add(getEnrollment(enrollment, fields, includeDeleted));
```
enricocolasante: Now we could just simplify this and use stream
`return enrollments.stream().map(e -> getEnrollment(enrollment, fields, includeDeleted)).toList()`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/enrollment/DefaultEnrollmentService.java
```diff
@@ -277,21 +272,11 @@ private Set<TrackedEntityAttributeValue> getTrackedEntityAttributeValues(Enrollm
   }
 
   private List<Enrollment> findEnrollments(
```
enricocolasante: Should we rename this method now as it is only a mapping and no more filtering is happening?

enricocolasante (approved):

PR #21177, opened 2025-06-09
fix: Keep executedBy value for isValidUserRequiredForJob jobs

david-mackessy:
## Issue
Issue in Job Scheduling found while testing existing issue in Agg Data Exchange job.  Agg Data Exchange job is failing after it is updated.

## Cause
Default executed by creator jobs (as defined in `JobType#isDefaultExecutedByCreator`) lose their `executedBy` value when updated. This defeats the purpose of automatically setting this value and using it for subsequent runs.

## Fix
Add a check when updating a job config to keep the `executedBy` value if:
- it already exists and
- no value has been supplied in the update
- is a job type included in the `JobType#isValidUserRequiredForJob` check

Rename JobType method from `isDefaultExecutedByCreator` to `isValidUserRequiredForJob` to express the requirement that a valid user is required for these jobs.

## Testing
New integration test added updating an agg data exchange job, ensuring that the original `executedBy` value is kept after an update.

## Info
This still allows supplying a new `executedBy` value if desired.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/scheduling/JobType.java
@@ -187,9 +187,10 @@ public enum JobType {
 
   /**
    * @return when true, the {@link JobConfiguration#getExecutedBy()} is set to the job creator on
-   *     creation unless it was set explicitly
+   *     creation unless it was set explicitly. This means that a valid User is required for a Job
+   *     (System User will not work).
    */
-  public boolean isDefaultExecutedByCreator() {
+  public boolean isValidUserRequiredForJob() {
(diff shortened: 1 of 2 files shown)


jbee (changes requested):
I don't think we can do it in the hook. It is valid to update this value to `null`. So we must be able to distinguish between this being set to `null` intentionally and unintentionally. Lets discuss tomorrow.

david-mackessy:
> I don't think we can do it in the hook. It is valid to update this value to `null`. So we must be able to distinguish between this being set to `null` intentionally and unintentionally. Lets discuss tomorrow.

We discussed this and some of the constraints about these jobs, one of which is that a valid `User` is required for these jobs, so setting to null should never be allowed for these jobs. Will rename a method and add the job type check for scenarios where we want to keep the `executeBy` value (so only those jobs will be impacted)

PR #21180, opened 2025-06-10
fix: wrong manifest returned by app controller

vietnguyen:
https://dhis2.atlassian.net/browse/DHIS2-19708

### Issue
- The check condition for `manifest.webapp` return false because the input argument was `/manifest.webapp`. 

### New functionality:
* Changed the condition so it will return true even if the input contains `/` in the beginning.
* Move the related code to service layer so I can add test for it.

### Refactoring:
* Refactored the `AppController` to delegate the handling of `manifest.webapp` to the `AppManager`'s `handlingManifest` method, removing duplicate logic from the controller.

### Testing:
* Added a new unit test, `handlingManifestTest`, to verify the behavior of the `handlingManifest` method, ensuring that the `href` is correctly updated when it is set to `*`.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
@@ -417,6 +417,23 @@ public class DefaultAppManager implements AppManager {
     }
   }
 
+  @Override
+  public boolean handlingManifest(String resource, App application, String contextPath) {
+    if (!resource.contains("manifest.webapp")) {
+      return false;
+    }
+    // If request was for manifest.webapp, check for * and replace with
+    // host
+    if (application.getActivities() != null
+        && application.getActivities().getDhis() != null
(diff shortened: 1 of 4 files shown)


larshelge (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -417,6 +417,23 @@ public int getUriContentLength(@Nonnull Resource resource) {
     }
   }
 
+  @Override
+  public boolean handlingManifest(String resource, App application, String contextPath) {
+    if (!resource.contains("manifest.webapp")) {
```
amcgee, line 422: This seems incorrect, we don't want to do a contains but rather an "endsWith".  Probably not going to have too many negative results in this case but it could lead to a bit unexpected behavior.

amcgee:
@jason-p-pickering @larshelge if possible please request that the extensibility team review changes to this part of the codebase before merging, thanks

PR #21181, opened 2025-06-10
fix: inner join to program owner in tracker event store [DHIS2-19735]

muilpp:


enricocolasante (commented):
Would it makes sense to use `TrackerImportService` in the tests that need to create tracker data so we do not need to manually add the ownership?

muilpp:
> Would it makes sense to use `TrackerImportService` in the tests that need to create tracker data so we do not need to manually add the ownership?

I think so, this way we would have the same setup as we'll encounter in production. I'm not sure if that would increase the overall test duration and if so, by how much, I'd need to check it.
Anyway, if you don't mind, I'd rather do that in a separate ticket, as there are many places where we create enrollments manually and the setup might not be the same for all of them.

PR #21182, opened 2025-06-10
chore: Create single and tracker event class in importer [DHIS2-19701]

enricocolasante:
***Big picture***
The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated in the preprocess phase.

***Implemented in this PR***
- `Event` is now an interface and `TrackerEvent` is its implementation.
- We already created `SingleEvent` class that will be used later.
- We are still using `Event` interface in all phases of the importer.

***Next steps***
- Use the concrete classes in the preprocess phase.
- Separate validation flow based on the event type.
- Separate rule-engine flow based on event type.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
@@ -53,8 +53,10 @@ import org.hisp.dhis.tracker.imports.ValidationMode;
 import org.hisp.dhis.tracker.imports.domain.Enrollment;
 import org.hisp.dhis.tracker.imports.domain.Event;
 import org.hisp.dhis.tracker.imports.domain.Relationship;
+import org.hisp.dhis.tracker.imports.domain.SingleEvent;
 import org.hisp.dhis.tracker.imports.domain.TrackedEntity;
 import org.hisp.dhis.tracker.imports.domain.TrackerDto;
+import org.hisp.dhis.tracker.imports.domain.TrackerEvent;
 import org.hisp.dhis.tracker.imports.preheat.TrackerPreheat;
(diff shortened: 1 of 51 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
```diff
@@ -165,6 +167,24 @@ public Set<UID> getUpdatedTrackedEntities() {
...
+        .filter(event -> preheat.getProgram(event.getProgram()).isWithoutRegistration())
+        .map(event -> new SingleEvent())
+        .toList();
+  }
+
+  public List<TrackerEvent> getTrackerEvents() {
+    return this.events.stream()
+        .filter(
+            event ->
+                preheat.getProgram(event.getProgram()) == null
```
muilpp, line 182: why do we assume this is a tracker event if the program is not found?

muilpp:
> We are still using Event interface in all phases of the importer.

Is this temporary and they will be separated in the importer by the end of this epic?
Or is it too messy to use `TrackerEvent` and `SingleEvent` in every phase?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerBundle.java
enricocolasante: At this point I am still not sure when `getTrackerEvents` and `getSingleEvents` will be used and I want to be sure no event gets lost, so I needed to put somewhere the invalid events that have no program.

enricocolasante:
> > We are still using Event interface in all phases of the importer.
> 
> Is this temporary and they will be separated in the importer by the end of this epic? Or is it too messy to use `TrackerEvent` and `SingleEvent` in every phase?

Good question, I am still not sure.
I think for validation phase it would be good to use the interface and avoid a lot of duplication, other phases could benefit from the split though.

PR #21205, opened 2025-06-12
Fix: Min max deserialization (43) (#21202)

jason-p-pickering:
* Fix controller: Deserialzation was not working properly in the special `dataEntry` endpoint


jason-p-pickering:
For the record @jbee , I again think that having to add the organisation units to the super user in order to get this test to work is a bug.

jbee (approved):
:pray:

PR #21212, opened 2025-06-13
chore: Separate event types in TrackerBundle [DHIS2-19701]

enricocolasante:
***Big picture***
The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated in the preprocess phase.

***Implemented in this PR***
- Remove `events` field from `TrackerBundle` and add `trackerEvents` and `singleEvents` fields.
- `trackerEvents` and `singleEvents` fields are created after preheat phase and any ambiguous event (without valid program and programStage) is mapped to a tracker event.
- We are maintaining (maybe only temporary) getter and setter for events.
- For now, `Event` interface contains fields that only belong to tracker events and `SingleEvent` class is implementing them. After we tackle validation phase we can review it and have a proper interface.

***Next steps***
- Use the concrete classes in the preprocess phase.
- Separate validation flow based on the event type.
- Separate rule-engine flow based on event type.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
@@ -87,7 +87,6 @@ public class DefaultTrackerImportService implements TrackerImportService {
       @Nonnull TrackerObjects trackerObjects,
       @Nonnull JobProgress jobProgress) {
     UserDetails currentUser = CurrentUserUtil.getCurrentUserDetails();
-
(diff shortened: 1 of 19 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/DefaultTrackerImportService.java
```diff
@@ -87,6 +87,7 @@ public ImportReport importTracker(
       @Nonnull TrackerObjects trackerObjects,
       @Nonnull JobProgress jobProgress) {
     UserDetails currentUser = CurrentUserUtil.getCurrentUserDetails();
+
```
enricocolasante:
```suggestion
```

muilpp (approved):

PR #21224, opened 2025-06-16
chore: Centralize error message

larshelge:
Centralize file resource error messages which were replicated across many classes.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/document/Document.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/document/Document.java
@@ -119,7 +119,7 @@ public class Document extends BaseIdentifiableObject implements MetadataObject {
     this.attachment = attachment;
   }
 
-  // Do not expose in the API
+  // Do not expose in API
   @JsonIgnore
   public FileResource getFileResource() {
     return fileResource;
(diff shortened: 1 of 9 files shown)


netroms (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/document/Document.java
```diff
@@ -77,6 +79,11 @@ public Document(String name, String url, boolean external, String contentType) {
     this.contentType = contentType;
   }
 
+  @JsonIgnore
+  public boolean isAttachmentDefaultFalse() {
```
david-mackessy, line 83: The name is a little confusing and ties itself to the default value which could change.
Would `isDownloadableAttachment` work?
It avoids using `isAttachment` which may interfere with getter/setter issues from some libraries.

larshelge: Right. Yes `isAttachment` made Hibernate complain. I think `isDownloadableAttachment` is also somewhat confusing as this refers to the content disposition HTTP header, which instructs the browser to download the resource as an attachment.

PR #21236, opened 2025-06-17
fix: Wrong manifest returned by app controller [DHIS2-19708](2.40)

david-mackessy:
backport of #21180

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
@@ -416,6 +416,23 @@ public class DefaultAppManager implements AppManager {
     }
   }
 
+  @Override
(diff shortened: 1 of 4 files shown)


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/appmanager/DefaultAppManager.java
```diff
@@ -416,6 +416,23 @@ public int getUriContentLength(@Nonnull Resource resource) {
     }
   }
 
+  @Override
+  public boolean handlingManifest(String resource, App application, String contextPath) {
+    if (!resource.contains("manifest.webapp")) {
```
amcgee: @david-mackessy I commented on the previous PR (after it was merged) but `contains` should probably be `endsWith` here - not likely to cause huge issues but it's more precise

david-mackessy: thanks, I'll update to use `endsWith`

PR #21242, opened 2025-06-17
chore: Separate event types in security validation phase [DHIS2-19701]

enricocolasante:
***Big picture***
The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated just after the preheat phase.

***Implemented in this PR***
- Create separate security validators for single and tracker events.
- We need to always check if orgUnit is in capture scope for tracker events (except for schedule status). We were skipping the check if we were updating the tracker event.

***Next steps***
- Review dispatcher that creates single and tracker event in case of delete strategy.
- Continuing separate validation flow based on event type.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityTrackerEventValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityTrackerEventValidator.java
@@ -97,6 +97,11 @@ class SecurityTrackerEventValidator
             ? event.isCreatableInSearchScope()
             : preheatEvent.isCreatableInSearchScope();
 
+    // TODO: Discuss with product how this should be fixed.
+    // At the moment we are checking capture scope for event org unit
+    // only when we are creating or deleting an event.
+    // When updating, ownership is enough.
+    // We need to understand what to do when updating the org unit.
     if (strategy.isCreate() || strategy.isDelete()) {
       checkEventOrgUnitWriteAccess(
           reporter, event, organisationUnit, isCreatableInSearchScope, bundle.getUser());
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/enrollment/SecurityOwnershipValidator.java
@@ -156,6 +156,12 @@ class SecurityOwnershipValidator implements Validator<Enrollment> {
       enrollmentOrgUnit = bundle.getPreheat().getOrganisationUnit(enrollment.getOrgUnit());
     }
 
+    // TODO: Discuss with product how this should be fixed.
+    // At the moment we are checking capture scope for enrollment org unit
+    // only when we are creating or deleting an enrollment.
+    // When updating, ownership is enough.
+    // We need to understand what to do when updating the org unit.
+
     // If enrollment is newly created, or going to be deleted, capture scope
     // has to be checked
     if (strategy.isCreate() || strategy.isDelete()) {


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityTrackerEventValidator.java
```diff
@@ -62,141 +62,69 @@
...
-    } else {
-      organisationUnit = bundle.getPreheat().getOrganisationUnit(event.getOrgUnit());
-    }
-
-    if (program.isWithoutRegistration() || strategy.isCreate() || strategy.isDelete()) {
-      checkEventOrgUnitWriteAccess(
-          reporter,
-          event,
-          organisationUnit,
+    if (event instanceof TrackerEvent) {
```
teleivo: invert and return early

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/OwnershipTests.java
```diff
@@ -126,7 +125,7 @@ public void shouldUpdateTrackedEntitiesInSearchScopeWhenGlassIsBroken() throws E
 
     JsonObject updatePayload =
         trackerImportExportActions
-            .getTrackedEntity(teId + "?fields=*")
+            .getTrackedEntity(teId + "?fields=*,!enrollments")
```
teleivo: why do we not need enrollments anymore here? and why did the error codes change?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecuritySingleEventValidator.java
```diff
@@ -0,0 +1,134 @@
...
+@RequiredArgsConstructor
+class SecuritySingleEventValidator
+    implements Validator<org.hisp.dhis.tracker.imports.domain.Event> {
+
+  @Nonnull private final AclService aclService;
+
+  @Override
+  public void validate(
+      Reporter reporter, TrackerBundle bundle, org.hisp.dhis.tracker.imports.domain.Event event) {
+    if (event instanceof SingleEvent singleEvent) {
```
teleivo: invert and return early

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityTrackerEventValidator.java
```diff
@@ -62,141 +62,69 @@
...
-
-    checkEventWriteAccess(
-        reporter,
-        bundle,
-        event,
-        programStage,
-        organisationUnit,
-        ownerOrgUnit,
-        categoryOptionCombo,
-        // TODO: Calculate correct `isCreateableInSearchScope` value
```
teleivo, line 160: Is this todo fixed or was it not relevant anymore?

enricocolasante: Not relevant anymore, it was fixed at some point, as now the variable is correctly calculated, but the TODO was not removed.

dhis-2/dhis-test-e2e/src/test/java/org/hisp/dhis/tracker/imports/OwnershipTests.java
enricocolasante: We were getting tracked entities and enrollments and events and update them but user hasn't the event org unit in capture scope and then it is not allowed to update. This specific test is for ownership, so just retrieving and updating tracked entities makes sense.

In the other test, the error code changes because the user doesn't have org unit in capture scope for events and this is now validated also for updates.

@karolinelien and @ameenhere confirmed that this new behavior is correct, as soon as @muilpp have a look I am going to create a bug and see if we need to backport it.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/EventValidator.java
```diff
@@ -62,15 +63,17 @@ public EventValidator(
                     new UpdatableFieldsValidator(),
                     new DataRelationsValidator(),
                     new CategoryOptionComboValidator(),
-                    securityOwnershipValidator,
+                    seq(
```
teleivo: I don't think you need to group them in a `seq` as they are already in one so will run sequentially and fail on the first one 🤔

enricocolasante: Maybe we can talk about this one.
I had to go this way because the bigger `seq` constructor take 8 params, I didn't want to create a bigger one.
Not sure what is the best way to go here

teleivo: ah, yes 😅 I think I remember. The same issue exists for `List.of` and `Map.of`. They define them up to 10 I think. You cannot use varargs as the array does not gel with the generics 😬

teleivo: I think we should add another factory 👌🏻

muilpp:
> We need to always check if orgUnit is in capture scope for tracker events (except for schedule status)

So it's possible to create scheduled events even if the user doesn't have the org unit in the capture scope? Why do we need to let the user create the event in that case?

enricocolasante:
> > We need to always check if orgUnit is in capture scope for tracker events (except for schedule status)
> 
> So it's possible to create scheduled events even if the user doesn't have the org unit in the capture scope? Why do we need to let the user create the event in that case?

@karolinelien can explain it better than me but it is a requirement for the system as a user could schedule (mark an appointment) to an org unit that it is not in its capture scope, so someone else with the right permission can carry on the event.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/event/SecurityTrackerEventValidator.java
```diff
@@ -62,141 +62,72 @@
...
-          program,
-          event.isCreatableInSearchScope());
+    OrganisationUnit ownerOrgUnit =
+        getOwnerOrganisationUnit(bundle.getPreheat(), teUid, programStage.getProgram());
+    boolean isCreatableInSearchScope =
+        strategy.isCreate()
+            ? event.isCreatableInSearchScope()
+            : preheatEvent.isCreatableInSearchScope();
+
+    if (strategy.isCreate() || strategy.isDelete()) {
```
muilpp, line 105: Is this temporary until we bring product into the discussion?
even if it is, can we add a comment explaining why in case of `UPDATE` we don't run the validation?
this is one of those things I won't remember in a couple of weeks

enricocolasante: Will add a comment

muilpp (approved):

PR #21260, opened 2025-06-19
backport: fix: Improve output of integrity check for OUGS with multiple groups

jason-p-pickering:
https://github.com/dhis2/dhis2-core/pull/21211

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/orgunits/orgunits_excess_group_memberships.yaml
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/orgunits/orgunits_excess_group_memberships.yaml
@@ -53,30 +53,56 @@ summary_sql: >-
   from ougs_multiple_groups
   WHERE array_length(ougs_groups, 1) > 1;
 details_sql: >-
-  WITH ougs_multiple_groups as (
-  SELECT ou_uid,ou_name,ougs_name,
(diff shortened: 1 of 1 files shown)


david-mackessy (approved):

jbee (approved):

jason-p-pickering:
@jbee @david-mackessy I disabled one integration test  to get  the build to pass. The test was failing due to the fact that the response format has changed for this integrity check. This will be an issue for the legacy integrity report which has not been removed from this version of 41. However, considering that the new front end does not use this legacy report, I feel like maybe it's not worth fixing or making it backwards compatible. Let me know what you think.

david-mackessy:
@jason-p-pickering if it's a lot of work to get it working I think it's ok to leave disabled.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityReportControllerTest.java
```diff
@@ -141,7 +142,7 @@ void testOrganisationUnitsWithCyclicReferences() {
             .toList(JsonString::string));
   }
 
-  @Test
+  @Disabled("Changed format of the response, but getDataIntegrityReport() is effectively deprecated")
```
jbee: Just test that the old DI app or API does not stumble with the changed format ;)

PR #21271, opened 2025-06-23
feat: Add filtering support for multi-text data elements [DHIS2-15568]

zubaira:
**Support filtering multi-text values for data elements**

### Context

This PR enhances the dynamic SQL query generation logic to correctly handle `MULTI_TEXT` data elements.

---

### What’s Included

- Updated `leftOperandSql()` to detect `MULTI_TEXT` value types and generate appropriate `EXISTS (...)` clauses using `unnest(...)`.
- Supported SQL operations:
  - `IN`, `EQ`, `IEQ` 
  - `LIKE`, `ILIKE`, `SW`, `EW`
  - `NE`, `NEQ`
  - `NULL`, `NNULL`
- Introduced helper methods like `multiTextClause(...)`

---

### Example

For a value stored as: `"Penicillin,SULFA"`  
And a filter like: `LIKE '%sulf%'`

The generated SQL is:

```sql
EXISTS (
  SELECT 1
  FROM unnest(string_to_array(lower(ev.eventdatavalues #>> '{abc123, value}'), ',')) AS val
  WHERE trim(val) LIKE '%sulf%'
)

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
@@ -282,7 +282,7 @@ public class FilterJdbcPredicate {
           throw new BadRequestException(
               String.format(
                   "Invalid filter: Operator '%s' is not supported for multi-text data element '%s'.",
-                  operator.getValue(), de.getUid()));
+                  operator.name(), de.getUid()));
     };
   }
 

### CODE REVIEW

File: dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java [hunk 1]
```diff
@@ -240,6 +248,42 @@ private static String rightOperandSql(QueryOperator operator, Parameter paramete
...
+    String trimmed = "trim(val)";
+    String param = parameter != null ? parameter.name() : null;
+
+    return switch (operator) {
+      case IEQ, EQ, IN ->
+          String.format("exists (select 1 from %s where %s in (:%s))", unnestSql, trimmed, param);
+      case NE, NEQ ->
+          String.format(
+              "not exists (select 1 from %s where %s in (:%s))", unnestSql, trimmed, param);
+      case LIKE, SW, EW, ILIKE ->
```
ameenhere: I have not fully understood the query yet, but why have we clubbed LIKE, SW, EW, ILIKE in the same switch?

LIKE,ILIKE can be clubbed together with the search being %searchTerm% (which also seems missing here).
SW with searchTem% and EW %searchTerm. 

Maybe I have misunderstood all of this. But please explain.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -269,6 +287,68 @@ void getEventByIdWithDataValues() {
     assertHasMember(dataValue, "storedBy");
   }
 
+  @Test
+  void getEventsWithMultiTextDataValueUsingINOperator() {
```
ameenhere: Can we have one test each that falls into each of the switch clauses above atleast.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
zubaira: you have understood it correctly. Separated SW and EW operators

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -269,6 +296,76 @@ void getEventByIdWithDataValues() {
...
+        "in:red",
+        "sw:bl",
+        "ew:een",
+        "like:red",
+        "ilike:green",
+        "nilike:yellow",
+        "!null",
+        "ne:yellow"
+      })
+  void shouldReturnEventsForMultiTextDataValuesUsingSupportedOperators(String operator) {
```
enricocolasante:
```suggestion
  void shouldReturnEventsForMultiTextDataValuesUsingSupportedOperators(String filter) {
```
This is the filter: operator + value

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -269,6 +296,76 @@ void getEventByIdWithDataValues() {
     assertHasMember(dataValue, "storedBy");
   }
 
+  @ParameterizedTest
+  @ValueSource(
+      strings = {
+        "in:red",
```
enricocolasante:
```suggestion
        "in:(red,yellow)",
```
Could we trying some more cases here?
I am not sure myself what is the expected behavior if we use more values in the in clause and only some of them match some of the multitext data value

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
```diff
@@ -240,6 +248,50 @@ private static String rightOperandSql(QueryOperator operator, Parameter paramete
...
+              "not exists (select 1 from %s where %s like :%s)", unnestSql, trimmed, param);
+      case SW ->
+          String.format(
+              "exists (select 1 from %s where %s like CONCAT(:%s, '%%'))",
+              unnestSql, trimmed, param);
+      case EW ->
+          String.format(
+              "exists (select 1 from %s where %s like CONCAT('%%', :%s))",
+              unnestSql, trimmed, param);
+      case NULL ->
```
enricocolasante, line 273: `NULL` and `NNULL` case follow the same logic as a data value that is not multitext, right?
Can we make sure we use the same logic? If we have an empty string for a `TEXT` data value and we use the `NULL` operator, what is the outcome there?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
```diff
@@ -240,6 +248,50 @@ private static String rightOperandSql(QueryOperator operator, Parameter paramete
...
+          String.format(
+              "not exists (select 1 from %s where %s is not null and %s <> '')",
+              unnestSql, trimmed, trimmed);
+      case NNULL ->
+          String.format(
+              "exists (select 1 from %s where %s is not null and %s <> '')",
+              unnestSql, trimmed, trimmed);
+      default ->
```
enricocolasante, line 281: Should we remove the validation from dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackerevent/TrackerEventOperationParamsMapper.java and make this return `BadRequestException`?

zubaira: I’ve added the validation in `TrackerEventOperationParamsMapper` since we’re already performing similar validations there. It seemed like the most consistent and appropriate place for this logic. If we decide to keep validation in only one place, I’d suggest removing it from `FilterJdbcPredicate`.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
zubaira: resolved

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -269,6 +296,76 @@ void getEventByIdWithDataValues() {
...
+        "sw:bl",
+        "ew:een",
+        "like:red",
```
ameenhere:
```suggestion
        "like:ello",
```
To see if it finds yellow.

All the like and ilike examples are using full search text, please use something in the middle like "ree" for green or "ell" for yellow, to make sure the query works as expected

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
```diff
@@ -269,6 +296,76 @@ void getEventByIdWithDataValues() {
...
+        "ilike:green",
+        "nilike:yellow",
```
ameenhere: Same as above

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
```diff
@@ -240,6 +248,50 @@ private static String rightOperandSql(QueryOperator operator, Parameter paramete
...
+    String param = parameter != null ? parameter.name() : null;
+
+    return switch (operator) {
+      case IEQ, EQ, IN ->
+          String.format("exists (select 1 from %s where %s in (:%s))", unnestSql, trimmed, param);
+      case NE, NEQ ->
+          String.format(
+              "not exists (select 1 from %s where %s in (:%s))", unnestSql, trimmed, param);
+      case LIKE, ILIKE ->
+          String.format("exists (select 1 from %s where %s like :%s)", unnestSql, trimmed, param);
```
ameenhere, line 269: I am thinking why LIKE,ILIKE,NLIKE,NILIKE does not have the wildcards (%) concatenated before and after, (Similar to what you did for EW and SW). 

There is a test case down below that I added a comment on, if you modify the test case and if it fails, then probably it is because of this.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
```diff
@@ -240,6 +248,50 @@ private static String rightOperandSql(QueryOperator operator, Parameter paramete
...
+          String.format(
+              "exists (select 1 from %s where %s like CONCAT(:%s, '%%'))",
+              unnestSql, trimmed, param);
+      case EW ->
+          String.format(
+              "exists (select 1 from %s where %s like CONCAT('%%', :%s))",
+              unnestSql, trimmed, param);
+      case NULL ->
+          String.format(
+              "not exists (select 1 from %s where %s is not null and %s <> '')",
```
muilpp, line 275: minor thing, but wouldn't it be easier to understand if we do 
`"exists (select 1 from %s where %s is null and %s <> '')",`
?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
zubaira: we are adding wild card down the line. 
https://github.com/dhis2/dhis2-core/blob/7e7d0f58a06d0ab5ecbb59eb2bd0152f08ec99f1/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryFilter.java#L207

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
zubaira: added test

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
zubaira: a value which is null and not equal to `''`.  is that possible?
we need values which are not null but also not `''`

muilpp: you're right, I misread the code, nevermind my question

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
zubaira: The syntax `in:(red,yellow)` results in an error, whereas `in:red;yellow` works correctly.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
zubaira: If a data value is stored as an empty string `("")`, then using the filter `filter=de_uid:null` will retrieve this event. In contrast, the filter `filter=de_uid:!null` will retrieve all events except the one with the empty string.
This case is covered in one of the tests as well.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/FilterJdbcPredicate.java
enricocolasante: I think it would be better to have it in `FilterJdbcPredicate`.
Having it there make sure that the object `FilterJdbcPredicate` is always valid and there is no way to create an invalid object and even more we do not need to replicate this validation.
At the moment there is a missing validation in `SingleEventOperationParamsMapper` about this.

zubaira: fixed

PR #21280, opened 2025-06-24
feat: Program stages with no programs integrity check

jason-p-pickering:
- https://dhis2.atlassian.net/browse/DHIS2-19795
- Adds SQL based check for program stages which are not associated with a program
- Note to devs: The current unit test that creates a program stage with no program is done with the service layer. At the point in time that the change is made to the API which forces a program to be associated with a program stage, then this unit test is going to (lets say should) start to fail. At that point, I think we can disable the unit test going forward, but for purposes of backporting this to earlier versions we can keep it.


dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+---
+name: program_stages_no_programs
+description: Program stages with no associated program
+section: Programs
+section_order: 2
+summary_sql: >-
+  WITH rs as (
+  SELECT a.uid,a.name, b.name as comment
+  from programstage a
+    INNER JOIN program b ON a.programid = b.programid
```
ameenhere: I am confused about this INNER JOIN, and then filtering on a.programid NULL. (Inner joins dont give results if either of the on sides are null right?)

Do we even need this join? We could simply use the programstage table to understand which of the programstages have null programid?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+  WHERE a.programid IS NULL
+  )
+  select count(*) as value,
+  100.0 * count(*) / NULLIF( (select count(*) from programstage),0) as percent
+  from rs;
+details_sql: >-
+  SELECT a.uid,a.name, b.name as comment
+  from programstage a
+    INNER JOIN program b ON a.programid = b.programid
```
ameenhere: Same comment above. I feel the "comment" column name has a special meaning? Perhaps the program stage name itself could be the comment then?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+  WHERE a.programid IS NULL;
+severity: SEVERE
+introduction: >
+  All program stages should be associated with a program. This check identifies
+  program stages that are not linked to any program. Certain version of DHIS2
+  do now allow program stages to exist without being linked to a program and may
```
ameenhere:
```suggestion
  used to allow program stages to exist without being linked to a program and may
```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+  from rs;
+details_sql: >-
+  SELECT a.uid,a.name, b.name as comment
+  from programstage a
+    INNER JOIN program b ON a.programid = b.programid
+  WHERE a.programid IS NULL;
+severity: SEVERE
+introduction: >
+  All program stages should be associated with a program. This check identifies
+  program stages that are not linked to any program. Certain version of DHIS2
```
ameenhere:
```suggestion
  program stages that are not linked to any program. Certain old versions of DHIS2
```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+  Obviously, as is always the case with direct database manipulation, you should
+  ensure that you have a backup of your database before running this command, perform
+  the change in a test environment first, and ensure that the system functions as expected after the change.
+  
+  Alternatively, if the program stage is not needed, you can delete it using the following SQL command:
+  ```sql
+    DELETE FROM programstage WHERE uid = '<program_stage_uid>';
+  ```
+  
+  Note that any data or enrollments associated with the program stage will also be deleted.
```
enricocolasante: I think only events can be associated with program stages, enrollments are only associated with programs.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,74 @@
...
+  WHERE a.programid IS NULL;
+severity: SEVERE
+introduction: >
+  All program stages should be associated with a program. This check identifies
+  program stages that are not linked to any program. Certain version of DHIS2
+  do now allow program stages to exist without being linked to a program and may
+  result in the system not being able to start.
+details_id_type: programStages
+recommendation: >
+  You will need to either delete the program stage or link it to a program.
```
enricocolasante: Here in recommendation I would suggest to check if there is data attached to the faulty program stages (provide a simple query to check that) and if it is the case, suggest to remove the program stage.
If there is any data attached, then leave it to the user if they want to try to find a program or delete all the data and then the program stage.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
enricocolasante: The data will not be deleted, I think is more correct to say that the data must be already deleted in order to be able to delete the program stage.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Reworked the query.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Haha, yeah, not sure what I was thinking there. Totally pointless inner join.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Fixed typo

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Clarified this explicitly.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Fixed

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: I added some details in the "comments" section of the details query which indicates whether data exists or not. At the moment, it is not possible to see the "comments" in the user interface though.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,105 @@
...
+  perform the change in a test environment first, and ensure that the system functions as expected after the change.
+  
+  Alternatively, if the program stage is not needed, you will need to delete it.
+  
+  The integrity check details will indicate whether the program stage has any associated data. If there is
+    associated data, you will need to delete that data first before deleting the program stage. You can 
+  delete the associated program stage data using the following SQL command:
+  
+  ```sql
+    DELETE FROM event WHERE programstageid = '<program_stage_id>';
```
enricocolasante: A few other tables have events as foreign keys and we do not have on cascade active so this `DELETE` query will most likely not work.
There are 2 way to completely delete events:
- Mark them as `deleted`, so change `deleted` column to `true` and then go to `https://play.im.dhis2.org/dev/apps/data-administration#/maintenance` and remove soft-deleted events
- Run a query similar to the one here: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java#L110-L145

I am not sure what is the best way to go

jason-p-pickering: Sounds good. I can update the guidance based on your comments.

jason-p-pickering: Yeah, I am not sure really either after looking at the program stage table. 

```
Referenced by:
    TABLE "programnotificationtemplate" CONSTRAINT "fk4uq2bl31hdu2s4e07rltemk3d" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "eventvisualization" CONSTRAINT "fk_evisualization_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "mapview" CONSTRAINT "fk_mapview_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programstagesection" CONSTRAINT "fk_program_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programrule" CONSTRAINT "fk_programrule_programstage" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programruleaction" CONSTRAINT "fk_programruleaction_programstage" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programrulevariable" CONSTRAINT "fk_programrulevariable_programstage" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programstagedataelement" CONSTRAINT "fk_programstagedataelement_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "event" CONSTRAINT "fk_programstageinstance_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programstageworkinglist" CONSTRAINT "fk_programstageworkinglist_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid) ON DELETE CASCADE
    TABLE "relationshipconstraint" CONSTRAINT "fk_relationshipconstraint_programstage_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "smscommands" CONSTRAINT "fk_smscommand_programstage" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "trackedentitydataelementdimension" CONSTRAINT "fk_tedataelementdimension_programstageid" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "trackedentityinstancereminder" CONSTRAINT "fke625d6778704ee73" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "trackedentityinstancereminder" CONSTRAINT "fke86df4bqx6ikwo9wdvbp8or4w" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)
    TABLE "programstagesection" CONSTRAINT "fkfa727aab8704ee73" FOREIGN KEY (programstageid) REFERENCES programstage(programstageid)


```
So some of these appear to have a cascade, while others do not. It would seem maybe to be quite difficult and maybe hazardous to remove a program stage with SQL?

ameenhere: I feel option 1 of what Enrico said might be less riskier?

To simply mark the programstage data as soft deleted (update event set deleted=true where programstageid=<programstageidWithNullProgram), and then heading into the Data Administration-> Maintenance -> Remove soft deleted events. 

The maintenance section should handle most reference deletions.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
```diff
@@ -0,0 +1,105 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+---
+name: program_stages_no_programs
+description: Program stages with no associated program
+section: Programs
+section_order: 2
+summary_sql: >-
+  WITH rs as (
```
jbee, line 33: I think we should be consistent with the SQL keyword case. I recently used upper case as I found it helping with readability and also with tooling going back and forth you often get it formatted that way out of AI so that made sense to me.

jason-p-pickering: You mean uppercase keywords?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: Check my comments in slack. Maybe the best way to deal with this is to create a program and associate the orphaned program stage with it. After that implementers could deal with it via normal means?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/program_stages_no_programs.yaml
jason-p-pickering: I cleaned up the case for the keywords @jbee

enricocolasante (approved):

PR #21281, opened 2025-06-24
fix: Make programid column not null in programstage table [DHIS2-19785]

enricocolasante:


enricocolasante:
This will not be solved now.
Moved https://dhis2.atlassian.net/browse/DHIS2-19785 back to TODO

PR #21282, opened 2025-06-24
fix: OpenAPI include Setting names as keys not properties [DHIS2-19787]

jbee:
The `SystemSettings` object schema should use the actual key names instead of the Java property names. 

This https://play.im.dhis2.org/dev/api/openapi.html?scope=entity:SystemSetting#SystemSettings currently looking like this:

DHIS-API--06-24-2025_02_09_PM

should look like this

DHIS-API--06-24-2025_02_10_PM

Note that some names now have a `key` prefix (or one even `Key`). This is the name of the setting as opposed to the property name. The property mostly also works because when stating keys it is allowed to omit the `key` prefix but that still posed some issues for machine-machine API usage so this PR corrects the OpenAPI schema to state key names.

### Automatic Testing
Adds a unit test to check the mapping between key and property is captured correctly.

### Manual Testing
Check  https://play.im.dhis2.org/dev/api/openapi.html?scope=entity:SystemSetting#SystemSettings and look for `key` in the names

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/LazySettings.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/LazySettings.java
@@ -55,6 +55,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


david-mackessy (approved):

jbee:
Reminder to me: backport to 2.42

PR #21283, opened 2025-06-24
feat: Gist API - support efficient OU tree search and offline caching [DHIS2-19789]

jbee:
### Summary
Adds two new Gist API URL parameters that are meant to be used with `/api/organisationUnits/gist`. They work (mostly) by forcing specific values for other parameters and as such they could be used for other objects but there they might not have an effect or cause errors because the properties forced do not exist. 

* `orgUnitsTree`: for doing searches with paging responding in matched results that can always be rendered in a tree because ancestors of matches not matching the query will be added into the results list. Forces `order=path` to give the tree depth-first order and `references=false`. The adding in of ancestors and the `match` indicator column is only happening for `/api/organisationUnits/gist` with JSON.
* `orgUnitsOffline`: for fetching partial (sub) trees for offline caching. Forces `fields=path,displayName,children:isNotEmpty`, `references=false` and does not use paging.

### Performance
A URL with `orgUnitsOffline` like `/api/organisationUnits/gist?orgUnitsOffline=true&filter=level:lt:4` results in SQL like
```sql
select 
  ou.uid, 
  ou.path, 
  ou.name, 
  ou.translations, 
  exists (select c.organisationunitid from organisationunit c where ou.organisationunitid=c.parentid) 
from organisationunit ou 
where ou.hierarchylevel<4
```
Most notably the `children:isNotEmpty` results in a reasonable sub-query `exists (select c.organisationunitid from organisationunit c where ou.organisationunitid=c.parentid)`. Since we use `displayName` a couple more columns need to be resolved compared to what is in the result. Tickets exists to improve translation handling in Gist API in general which could move both fetching and filtering `displayName` fully into the DB.

For a URL with `orgUnitTree` like `/api/organisationUnits/gist?orgUnitsTree=true&filter=id:eq:a04CZxe0PSe&fields=path,id,name`
this results in 2 SQL queries, 1st the fetching of matches, 2nd the fetching of ancestors:
```sql
select ou.path, ou.uid, ou.name, ou.translations
from organisationunit ou
where ou.uid='a04CZxe0PSe' order by ou.path asc limit 50;

select ou.path, ou.uid, ou.name, ou.translations
from organisationunit ou 
where (ou.uid in ('C9uduqDZr9d' , 'ImspTQPwCqd' , 'at6UHUQatSo')) order by ou.organisationunitid asc
```
The set of IDs in the `in` clause of the 2nd query depends on what ancestors are missing from the page.
So the size should always be limited to a smallish number limited by the max `pageSize`.
Again the query would further benefit from doing general translation improvements. 
In case a `totalPages=true` is requested a 3rd query is made.

### _Is (not) empty_ Transform & Sharing
There is a easily missed change with these transformations. When these were based on size = 0 / size > 0 this did consider sharing access in the counting the collection table. I dropped including sharing as an unintentional aspect of the original implementation. I am not sure this ever was intentional or a requirement. 

Since _is (not) empty_ transforms are now allowed to be used with non-paged lists it makes sense to ensure these are actually reasonably cheap and do not require full table scans but will be using indexes. For this to be the case we cannot consider sharing.  Alternative we should add new transformation with the semantic to not consider sharing. 

### Examples
`/api/organisationUnits/gist?orgUnitsTree=true&filter=id:eq:a04CZxe0PSe&fields=path,id,name`
on SL demo results in
```json
{
  "pager": {
    "page": 1,
    "pageSize": 50
  },
  "organisationUnits": [
    {
      "match": false,
      "path": "/ImspTQPwCqd",
      "id": "ImspTQPwCqd",
      "name": "Sierra Leone"
    },
    {
      "match": false,
      "path": "/ImspTQPwCqd/at6UHUQatSo",
      "id": "at6UHUQatSo",
      "name": "Western Area"
    },
    {
      "match": false,
      "path": "/ImspTQPwCqd/at6UHUQatSo/C9uduqDZr9d",
      "id": "C9uduqDZr9d",
      "name": "Freetown"
    },
    {
      "match": true,
      "path": "/ImspTQPwCqd/at6UHUQatSo/C9uduqDZr9d/a04CZxe0PSe",
      "id": "a04CZxe0PSe",
      "name": "Murray Town CHC"
    }
  ]
}
```

### Automatic Testing
Added basic coverage with controller tests for the new parameters

### Manual Testing
See examples in description above


HendrikThePendric:
Thanks for providing such an elaborate PR description. The response payload in the example seems to be exactly what we would need. I think playing around with this a bit would be helpful, so I've added the `deploy` label.

I seem to remember you mentioned we can filter on name `filter=name:ilike:mysearchtext`, but not by displayName `filter=displayName:ilike:mysearchtext`. I'm assuming this is a performance related restriction, but wonder what the implications of this are reg. localisation. If this means that users ultimately cannot do a text search for an org-unit in their own language, then perhaps we should check with someone from the product side if this is an acceptable restriction before we proceed.

Just to clarify: I am focussing solely on the `orgUnitsTree=true` side of things. I'm hoping @tomzemp will evaluate the `orgUnitsOffline=true` part.

vietnguyen (approved):

PR #21297, opened 2025-06-25
fix: (2.39)ensure program rule notifications correctly resolve dataElement [DHIS2-19685]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-19685


enricocolasante (commented):
I think 2.39 is now out of support

linadhis2:
> I think 2.39 is now out of support

Yes! This one is not necessary as it did get in for the last patch 39.10 separately after it got frozen. @Philip-Larsen-Donnelly is there any reason to also get this to 39dev? maybe we'd still need it or likely not necessary?

zubaira:
end of support

PR #21298, opened 2025-06-25
feat: Introduce min characters on TEA to start search [DHIS2-19757]

muilpp:


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -200,6 +200,19 @@ private void mapAttributeFilters(
                 attributeFilter.getKey()));
       }
 
+      List<QueryFilter> binaryFilters =
+          attributeFilter.getValue().stream().filter(qf -> qf.getOperator().isBinary()).toList();
```
enricocolasante, line 204: I understand that it will work from a configuration point of view, but I cannot see the value of limiting the number of digits for operators like `GT` or `LT`.

muilpp: Do you mean you'd do it for certain operators as `EQ` or `LIKE` only?

enricocolasante: Yes, maybe having some logic around operators and value type.
For example, if the value is numeric and we are using GT as an operator, why would we have minimum characters?
I know that in the end it would be a configuration from the admin, but we should help them not having configuration that do not makes sense.

muilpp: In my mind, the min characters feature isn’t just useful for trigram indexes, but also for B-tree indexes.
Even with numeric values, providing more characters makes a more efficient use of the index than when fewer characters are used.
That’s why I thought it could still make sense to let the admin decide whether a minimum number of characters should be required before starting a search, regardless of the field's data type.

But as always, I'm not entirely sure how implementations use it or whether it'd make sense for them to have this configuration available on all TEAs.
@ameenhere any thoughts on this?

ameenhere: Good point. 
Imo, I just want the code to be maintainable for the first version. Putting too much logic (or magic) in trying to help admins will result in too messy code I fear. We could just shift focus to proper configuration guide instead? 

As a middle ground, I don't think we should restrict this configuration based on valueType atleast, maybe just operators? So for the operator LT,LE,GT,GE (lets call it rangeFilters), maybe we don't have to validate the minimumCharactersToSearch config? 

Thoughts?

muilpp: I'm just trying to wrap my head around this before we add more corner cases to the validation.
What’s the downside of letting admins decide whether they want to enforce a minimum number of characters, regardless of the operator used?

To me, there are two advantages:
- The index could still benefit from it, even if the field is numeric and a range filter operator is used.
- By default, it's set to 0, meaning no restriction. Admins only need to configure it if they actually want to enforce a minimum length.

If we drop validation based on the operator, then admins lose the ability to enforce this behavior even if they want to. For example, they might configure a TEA with both `EQ` and `GT` as allowed operators and set a minimum character length. But if validation is conditional on the operator used in the request, that logic becomes harder to explain.

Are we saying there’s no use case where a range filter operator could benefit from a minimum character length?

larshelge: I think I agree with @enricocolasante . For text values, it is often sensible to always require at least two characters for search.

Now for integer values, it may be outright prohibitive to enforce min characters. Example: If "age in years" is a data element, you can no longer search for any child (from 0 to 9 years).

The only vague use-case I can see for limiting characters is for long unique IDs. However, for those, one normally uses the text value type, as if a unique ID is numeric it is usually has no meaning from a numerical point of view.

So by enforcing min limit of characters on numeric values, implementations may just say this is not working for them.

Characters also does not have the same meaning for letters as for digits in numbers.

This is also another argument in favor of equality matching for unique IDs, as if the matching is unique, the users won't typically try to "scrape" the system with fewer characters.

So to conclude, even if we want to start simple, I would say we start simple by only introducing this constraint for text based value types (text, email, phone nr) @ameenhere @muilpp .

enricocolasante (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
ameenhere: I agree with the principle @larshelge . 

If you think about pure API support, we currently support all operators on all valueTypes. And some of the combinations does not make sense currently. A LIKE on a numeric value type is currently possible in the API. Does it have a use-case? Maybe/Maybe not. A GT on a text value type is currently possible in the API. Does it have a use-case? Maybe/Maybe not.

When we think in terms of specific context. "Why would anybody want to restrict search characters on an Age/Number data element?" "Why would anybody want to restrict search characters when doing a GT operator?"
All of these questions does not have 100% concrete answer for DHIS2. 

My thinking is to avoid any kind of magic in a generic platform. And keep it open for configuration/customizability. If we start adding conditional validation for minCharactersToSearch for operators and valueType combination, it will lose its customizability, it gets complicated to document and explain to the administrators, it gets harder to maintain in the code ( Think of all the if conditions ) and so on. 

From an API point of view, it is acceptable to keep all of this open and have a source of truth in the configuration. The default is always no limit. If an admin intentionally modifies something, he has already shown intent. And if we then conditionally apply what he intended, that might get more confusing. 

So to conclude, I would not say "introducing this constraint selectively" is starting simple from developer/administrator/documentation pov, that is starting complicated. 😄 


I can be convinced with applying the min characters constraint only on non-range filters though (irrespective of valueType). Atleast that is just one dimension to check for and that may be easier to explain/document. 

What do you think @enricocolasante ?

enricocolasante: I agree with @ameenhere and we could just make it work and then shape it down to check all the cases and corner cases.

What I am confused is 
`The index could still benefit from it, even if the field is numeric and a range filter operator is used.`
If you need to filter by some blood value that could be between -100.000 and 100.000, then limiting the number of digits just doesn't make sense to me and I cannot see how it will help the index.

To resume, I think we should let all the possibility opened and let the admin configure the system but it is good to know on our side which configurations do not make sense, and maybe somewhere, at some point have warnings in the frontend on those. (In my head, the only configuration that doesn't make sense and do not make the index work better is the range for numeric types).

PR #21310, opened 2025-06-26
fix: validate zip files before unpacking (#20572)

netroms:
Backport: https://github.com/dhis2/dhis2-core/pull/20595


Philip-Larsen-Donnelly:
Tested successfully by QA

PR #21314, opened 2025-06-27
feat: Program rules linked to an inconsistent program stage

jason-p-pickering:
- Program rules can be optionally linked to a program as well as a program stage. The model theoretically allows program rules to be linked to a program stage which does not exist in the associated program. 
- This check identifies program rules which are linked to both a program as well as a program stage, but the program stage is not part of the linked program.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks.yaml
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks.yaml
@@ -46,6 +46,7 @@ checks:
   - program_rules/program_rules_without_condition.yaml
   - program_rules/program_rules_message_no_template.yaml
   - program_rules/program_rules_no_priority.yaml
+  - program_rules/program_rules_inconsistent_program_program_stage.yaml
   - analytical_objects/visualizations_not_used_1year.yaml
   - analytical_objects/maps_not_used_1year.yaml
   - analytical_objects/dashboards_not_used_1year.yaml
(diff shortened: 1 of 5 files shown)


david-mackessy (approved):

jason-p-pickering:
Noted @zubaira . Looks like there could be several potential problems there and we should probably create separate tickets and do a bit of analysis. 

1) Program rule actions linked to a program rule and a program stage, having inconsistent program stages (via the program rule linked to a stage). 
2) Program rule actions linked to a program stage and a program section id, having an inconsistent stage (via the program section). 

Perhaps there are other variants?

zubaira:
> Noted @zubaira . Looks like there could be several potential problems there and we should probably create separate tickets and do a bit of analysis.
> 
> 1. Program rule actions linked to a program rule and a program stage, having inconsistent program stages (via the program rule linked to a stage).
> 2. Program rule actions linked to a program stage and a program section id, having an inconsistent stage (via the program section).
> 
> Perhaps there are other variants?

Yes, there are other cases as well, but they are likely to be less problematic. Since we’re including the SCHEDULEEVENT rule action, and we’re adding validation in the metadata importer to ensure that the programStage in this action belongs to the correct program, those scenarios should be well covered.

jason-p-pickering:
@zubaira were these additional checks added relatively recently? If so then I assume that some databases could be afflicted with this integrity issue?

ameenhere (approved):

zubaira:
> @zubaira were these additional checks added relatively recently? If so then I assume that some databases could be afflicted with this integrity issue?

No. We aiming that for next release.

PR #21331, opened 2025-06-30
fix: Sort columns/rows + limit flag [DHIS2-19028]

maikelarabori:
This fixes the sorting issue on columns and rows, and adds a new system settings flag that allows users to define the limit of combinations of columns and rows allowed in the `download` feature of Data Visualizer.
If the limit is exceeded, an HTML file describing the error is returned.

This flow is hit when we define `tableLayout=true` in the URL params.


dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/grid/GridUtils.java
```diff
@@ -253,6 +254,19 @@ private static void addPdfTimestamp(Locale locale, Document document, boolean pa
     addTableToDocument(document, table);
   }
 
+  /**
+   * Generates an Excel file that contains a descriptive error message.
```
luciano-fiandesio: Is this comment correct? Does this method generates an Excel file?

maikelarabori: Fixed 👍

vietnguyen (approved):

PR #21332, opened 2025-06-30
chore: Fix separation of event types for delete strategy [DHIS2-19701]

enricocolasante:
***Big picture***
The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated just after the preheat phase.

***Implemented in this PR***
- `ParamsConverter` was separating events wrongly when the import strategy was `DELETE`. We shouldn't check the program or program stage in the payload, we need to just check the event in the DB and look at the type of the program.
OBS. If the converter cannot decide the type of the event (for create, when there is no program or program stage, and for delete or update, when the event does not exist), by default the event is considered a tracker event.

***Next steps***


dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/ParamsConverterTest.java
```diff
@@ -0,0 +1,268 @@
...
+    programWithRegistration = createProgram('A');
+    programWithRegistration.setProgramType(ProgramType.WITH_REGISTRATION);
+    programWithoutRegistration = createProgram('B');
+    programWithoutRegistration.setProgramType(ProgramType.WITHOUT_REGISTRATION);
+    programStageWithRegistration = createProgramStage('C', programWithRegistration);
+    programStageWithoutRegistration = createProgramStage('D', programWithoutRegistration);
+  }
+
+  @Test
+  void shouldSuccessToConvert() {
```
muilpp: Would you add the 'when' part in the test name?
I'm not really sure where do the events com from here, are they empty?
What about relationships? Are they empty too?

muilpp:
> _**Big picture**_ The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated just after the preheat phase. This will allow us to have separate flows for validation, rule-engine and persistence (even though for now we will use the same DB table).
> 
> _**Implemented in this PR**_
> 
>     * `ParamsConverter` was separating events wrongly when the import strategy was `DELETE`. We shouldn't check the program or program stage in the payload, we need to just check the event in the DB and look at the type of the program.
>       OBS. If the converter cannot decide the type of the event (for create or update, when there is no program or program stage, and for delete, when the event does not exist), by default the event is considered a tracker event.
> 
> 
> _**Next steps**_
> 
>     * Separate persistence flow based on event type.

What does OBS stand for?

enricocolasante:
> > _**Big picture**_ The goal of this ticket is to have separate domain objects for single events and tracker events that will be generated just after the preheat phase. This will allow us to have separate flows for validation, rule-engine and persistence (even though for now we will use the same DB table).
> > _**Implemented in this PR**_
> > ```
> > * `ParamsConverter` was separating events wrongly when the import strategy was `DELETE`. We shouldn't check the program or program stage in the payload, we need to just check the event in the DB and look at the type of the program.
> >   OBS. If the converter cannot decide the type of the event (for create or update, when there is no program or program stage, and for delete, when the event does not exist), by default the event is considered a tracker event.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > _**Next steps**_
> > ```
> > * Separate persistence flow based on event type.
> > ```
> 
> What does OBS stand for?

Observation

enricocolasante:
> Why do we still need to filter by program and program stage when updating? I don't understand why is it different to deleting.
> 
When updating we could go either way. Then if the program or programStage changed, it will be caught by the validation phase.

> When it comes to deleting an event, if the user provides the program and event ids, but the event is not in that program, shouldn't we fail? If I understand this PR correctly, in this case we would delete the event anyway, right?

https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-242/tracker.html#delete
When deleting the payload should only have UIDs of the entity to delete, any other field will be ignored.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/ParamsConverterTest.java
enricocolasante: Done

enricocolasante:
> > Why do we still need to filter by program and program stage when updating? I don't understand why is it different to deleting.
> 
> When updating we could go either way. Then if the program or programStage changed, it will be caught by the validation phase.

I move update in the same flow of delete, you were right we do not need to check the program there.

muilpp (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/ParamsConverter.java
```diff
@@ -85,7 +103,16 @@ private static List<TrackerEvent> buildTrackerEvents(List<Event> events, Tracker
         .toList();
   }
 
-  private static List<SingleEvent> buildSingleEvents(List<Event> events, TrackerPreheat preheat) {
+  private static List<SingleEvent> buildSingleEvents(
+      List<Event> events, TrackerPreheat preheat, TrackerImportStrategy importStrategy) {
+    if (importStrategy == TrackerImportStrategy.DELETE) {
+      return events.stream()
+          .map(e -> preheat.getEvent(e.getUid()))
+          .filter(Objects::nonNull)
```
teleivo: can you structure the above filter like this or the other way round AFAIK the main difference should be the isRegistration right? that should be easy to see if thats the case.

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/imports/ParamsConverterTest.java
```diff
@@ -0,0 +1,268 @@
...
+    when(trackerPreheat.getProgramStage(MetadataIdentifier.ofUid(programStageWithoutRegistration)))
+        .thenReturn(programStageWithoutRegistration);
+
+    TrackerImportParams params = TrackerImportParams.builder().build();
+    TrackerObjects trackerObjects =
+        TrackerObjects.builder()
+            .events(List.of(trackerEventWithProgramStage(), singleEventWithProgramStage()))
+            .build();
+    TrackerBundle trackerBundle =
+        ParamsConverter.convert(params, trackerObjects, user, trackerPreheat);
```
teleivo, line 193: is it easy to not mock in this test as the preheat is passed in?

PR #21355, opened 2025-07-04
Dhis2 19338

david-mackessy:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/category/CategoryComboServiceTest.java
```diff
@@ -276,57 +277,16 @@ void testUpdateCategoryOptionCombosB() {
...
     assertOptionCombos(categoryComboA.getOptionCombos());
-    categoryService.addAndPruneOptionCombos(categoryComboA);
+    categoryOptionComboGenerateService.addAndPruneOptionCombos(categoryComboA);
     assertNotNull(categoryComboA.getOptionCombos());
     assertEquals(8, categoryComboA.getOptionCombos().size());
     assertOptionCombos(categoryComboA.getOptionCombos());
   }
 
-  @Test
-  void testUpdateCategoryOptionCombosC() {
```
david-mackessy, line 290: removing as it's an exact duplicate of `testUpdateCategoryOptionCombosA` on line 235

vietnguyen (approved):

jason-p-pickering:
Nice work. :+1:

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
```diff
@@ -271,32 +271,33 @@ public List<List<CategoryOption>> getCategoryOptionsAsLists() {
   }
 
   /**
-   * Generates a list of all possible combinations of category option combos for this category
-   * combo. This is done by generating all possible combinations of the category options in the
-   * categories of this category combo.
+   * Generates a set of all possible combinations of category option combos for this category combo.
+   * This is done by generating all possible combinations of the category options in the categories
+   * of this category combo. This used to return a list but that had the potential to return
+   * duplicates if categories shared similar category options.
```
jason-p-pickering: Just maybe one small nitpick here. What does "similar" actually mean? Does this mean the same category option which is present in two categories which are part of the same category combo? We do have an integrity check for this, but I think maybe we could be more precise as to what similar means?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/category/CategoryServiceTest.java
```diff
@@ -381,11 +387,79 @@ void addAndPruneCategoryComboWithSummaryDelete() {
...
+    Category cat3 = createCategory('3', co1, co4);
+    entityManager.persist(cat1);
+    entityManager.persist(cat2);
+    entityManager.persist(cat3);
+    entityManager.flush();
+
+    CategoryCombo cc = createCategoryCombo('1', cat1, cat2, cat3);
+    // combo consists of 3 categories with options: [co1,co2] [co3,co4] [co1,co4]
+    // all combinations (7) should be:
+    // [co1,co3] [co1,co4] [co1,co3,co4] [co2,co3,co1] [co2,co3,co4] [2,4,1] [2,4]
```
david-mackessy, line 457: before the changes in this PR, this test fails as the COC [1,4] is generated twice, resulting in 8 COCs

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
david-mackessy: yeah cool 👍, I'll update 'similar' to 'the same'.
The test `cocGenerationDuplicateTest` has an example

PR #21366, opened 2025-07-07
chore: Separate changelog logic in preparation to table separation [DHIS2-19848]

enricocolasante:
The idea of this task is to prepare the code for the separation of `event` table that will happen later.
We will need to create 2 separate tables for `eventchangelog` as well, so we are already creating different hibernate mappings, stores and services.

***Chnages in this PR***

- Created 2 mappings for the different changelogs, but they still point to the same table. When we will separate the event tables, we will only need to point to the correct table and link to the correct event type (single or tracker event table).
- Separate changelog hibernate object and DTO changelog. DTO does not need to link to the event, so it can be shared by the 2 different types of changelog, while we need 2 different hibernate objects that will represent the 2 different tables.
- Create `EventService` to be used in `EventChangeLogService`. `EventChangeLogService` is abstract and the correct event service will be injected, here we only need to check if an event exists (or it is accessible) in order to retrieve or not the changelogs.
- `HibernateEventChangeLogStore` is an abstract class, as all the logic is the same for the 2 types, only the name of the table is different.
- `EventChangeLogService` is abstract because the changelogs retrieved here are the DTOs, that are the same for the 2 types; the only difference is the hibernate object that we are persisting to the DB and this is injected by the implementation of the service.

***What we will need to do when separating the event table***
- Update the mapping files
- Update the code to link to the correct `Event` type


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
```diff
@@ -31,77 +31,25 @@
...
 import org.hisp.dhis.changelog.ChangeLogType;
 import org.hisp.dhis.dataelement.DataElement;
-import org.hisp.dhis.program.Event;
 import org.hisp.dhis.program.UserInfoSnapshot;
 
-@NoArgsConstructor
 @Getter
-@Setter
+@RequiredArgsConstructor
 public class EventChangeLog {
```
teleivo: Could this now be a record 🤔 or `@Value`. Maybe before it was not possible because of Hibernate 🤷🏻 @muilpp

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -29,57 +29,173 @@
...
+    } else if (isFieldUpdated(oldValue, newValue)) {
+      return UPDATE;
+    } else {
+      return DELETE;
+    }
+  }
 
-  void addFieldChangeLog(
-      @Nonnull Event currentEvent, @Nonnull Event event, @Nonnull String userName);
+  private boolean isFieldCreated(String originalValue, String payloadValue) {
```
teleivo: can you please make these static as they are pure functions

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
```diff
@@ -152,7 +155,6 @@ public Page<EventChangeLog> getEventChangeLogs(
         results.stream()
             .map(
                 row -> {
-                  Event e = (Event) row[0];
```
teleivo, line 155: why is this not needed anymore?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/singleevent/HibernateSingleEventChangeLogStore.java
```diff
@@ -0,0 +1,48 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.tracker.export.singleevent;
+
+import jakarta.persistence.EntityManager;
+import org.hisp.dhis.tracker.export.event.HibernateEventChangeLogStore;
+import org.springframework.stereotype.Repository;
+
+@Repository("org.hisp.dhis.tracker.export.singleevent.HibernateSingleEventChangeLogStore")
+public class HibernateSingleEventChangeLogStore
```
teleivo: can this be package private?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
teleivo: Why do we have this `EventChangeLog` and the 2 new event changelog classes?

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/singleevent/OrderAndFilterSingleEventChangeLogTest.java
```diff
@@ -0,0 +1,497 @@
...
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+@Transactional
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class OrderAndFilterSingleEventChangeLogTest extends PostgresIntegrationTestBase {
```
teleivo: Are you splitting these tests as you envision the impls of these two changelogs to drift? Otherwise we could have one `OrderAndFilterEventChangeLogTest` for both like we do for the exporters.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLog.java
enricocolasante: I can make it a record, I was just lazy to rename all the getters and I didn't want to pollute the PR.
I will do it after the reviews.

enricocolasante: `EventChangeLog` is the object that we retrieve from the DB and we then map to the view object.
This object does not need to link to an event because we are always retrieving changelogs in the context of one event and in the view object the link to the event is not needed.
On the other side, we need the hibernate objects to link to the event and they will be different events (singleevent and trackerevent) as soon as we split the event table.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
enricocolasante: We are creating an `EventChangeLog` that does not need to link to an event.
That event was never needed, as it is not mapped to the view in the web layer.

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/singleevent/OrderAndFilterSingleEventChangeLogTest.java
enricocolasante: Done

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/TrackerEventChangeLog.hbm.xml
```diff
@@ -0,0 +1,43 @@
...
+
+    <property name="previousValue" length="50000" />
+
+    <property name="currentValue" length="50000" />
+
+    <property name="changeLogType" column="changelogtype" length="100" not-null="true">
+      <type name="org.hibernate.type.EnumType">
+        <param name="enumClass">org.hisp.dhis.changelog.ChangeLogType</param>
+        <param name="useNamed">true</param>
+        <param name="type">12</param>
```
muilpp: I don't think we need to specify the type. 
Hibernate infers it because we use the name method of the `changeLogType` enum, which already returns a String.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/HibernateEventChangeLogStore.java
muilpp: In that case, can we remove the event from the select clause?

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/eventchangelog.hiberante/TrackerEventChangeLog.hbm.xml
enricocolasante: Done

enricocolasante: Nope, I cannot remove it, hibernate is complaining that the type is wrong and the application is not starting at all.

muilpp (approved):

PR #21374, opened 2025-07-08
DHIS2-13818: default to HikariCP and deprecate C3P0

cjmamo:
Default to HikariCP and deprecate C3P0

### DIFF
--- a/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
+++ b/dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
@@ -293,6 +293,10 @@ public final class DatabasePoolUtils {
   private static ComboPooledDataSource createC3p0DbPool(
       String username, String password, String driverClassName, String jdbcUrl, DbPoolConfig config)
       throws PropertyVetoException {
+
+    log.warn(
+        "Using deprecated `c3p0` in `db.pool.type` DHIS2 configuration setting. It is highly recommended to set `db.pool.type` to `hikari` in your `dhis.conf`. Future versions of DHIS2 will drop support for C3P0");
+
     ConfigKeyMapper mapper = config.getMapper();
     DhisConfigurationProvider dhisConfig = config.getDhisConfig();
 
@@ -365,12 +369,14 @@ public final class DatabasePoolUtils {
 
   private static void logC3p0UnusedSettingsWarning(
       DhisConfigurationProvider dhisConfig, DbPoolType dbPoolType) {
-    Consumer<ConfigurationKey> logConsumer =
-        configurationKey ->
-            log.warn(
-                "`{}` C3P0 setting is not used anymore because the default database pool type is set to `hikari` since DHIS2 v43. Remove this C3P0 setting from your `dhis.conf`. Future versions of DHIS2 may stop giving this warning.",
-                configurationKey.getKey());
+
     if (!dbPoolType.equals(DbPoolType.C3P0)) {
+      Consumer<ConfigurationKey> logConsumer =
+          configurationKey ->
+              log.warn(
+                  "`{}` C3P0 setting is not used anymore because the default database pool type is set to `hikari` since DHIS2 v43. Remove this C3P0 setting from your `dhis.conf`. Future versions of DHIS2 may stop giving this warning",
+                  configurationKey.getKey());
+
       applyIfPropertyExists(dhisConfig, CONNECTION_POOL_MIN_SIZE, logConsumer);
       applyIfPropertyExists(dhisConfig, ANALYTICS_CONNECTION_POOL_MIN_SIZE, logConsumer);
       applyIfPropertyExists(dhisConfig, CONNECTION_POOL_INITIAL_SIZE, logConsumer);


dhis-2/dhis-support/dhis-support-external/src/main/java/org/hisp/dhis/external/conf/ConfigurationKey.java
```diff
@@ -138,58 +138,68 @@ public enum ConfigurationKey {
   ANALYTICS_CONNECTION_POOL_MAX_SIZE("analytics.connection.pool.max_size", "80", false),
 
   /** Minimum number of Connections a pool will maintain at any given time (default: 5). */
+  @Deprecated(since = "v43", forRemoval = true)
```
ameenhere, line 141: I am wondering why several of these are marked as deprecated? Even HikariCP has functionally matching equivalents for some of these right? @cjmamo

cjmamo: For some yes like `connection.pool.max_size` but for others no unfortunately since they go by different config keys. For instance, the counterpart of `connection.pool.min_size` is `connection.pool.min_idle` in HikariCP. There are also C3P0 config keys like `connection.pool.initial_size` for which there isn't a one-to-one mapping with a HikariCP setting but still the same functionality can be achieved conceptually with `connection.pool.max_idle_time` together with `connection.pool.min_idle`.  In some cases, like for `connection.pool.num.helper.threads`, there's not even an equivalent because HikariCP works differently. 

If we assume that the admin has not explicitly set the DB pool type to `c3p0` in `dhis.conf`, which I reckon it would be most of the time, then any C3P0 settings configured by the admin (e.g., `connection.pool.min_size`) would be effectively ignored following the upgrade to v43. Perhaps we should give an error at start-up if any the C3P0 config keys are set AND the DB pool type is not explicitly set to `c3p0`.

ameenhere: Yes agreed and understood. But I am still wondering if we are still not doing a best effort matching? 
If we consider the dhis configuration keys as db pool agnostic and take the real meaning/intention, we should be able to match some more of them to their corresponding counterpart in Hikari. I feel we are already doing some of them, but perhaps just check if we have left some out? Like the DHIS2 config key `connection.pool.min_size` to HikariCP config `connection.pool.min_idle`   is a sensible one to translate to within the HikariDb pool creation logic. Maybe there are more?

If there are very specific c3p0 config keys , yes those will (and can) be ignored from v43 by default. I just wanted to ask if we are doing a best effort matching of HikariCP counterparts. 

@netroms I assume you already have gone through this exercise when you introduced Hikari the first time. Any thoughts?

cjmamo: >If we consider the dhis configuration keys as db pool agnostic and take the real meaning/intention, we should be able to match some more of them to their corresponding counterpart in Hikari

HikariCP works very differently from C3P0. Many concepts in C3P0 do not exist in HikariCP.

>But I am still wondering if we are still not doing a best effort matching?

Probably `connection.pool.min_size` would be the only setting that can be applied to HikariCP but I'll check if there are others

cjmamo: Also, IMO we should discourage the use of different config keys which mean the same thing since it could lead to undesired behaviour (e.g., the admin sets `connection.pool.min_size` and then sets `connection.pool.min_idle` to a different value when switching to HikariCP but forgets to remove `connection.pool.min_size`)

ameenhere: I agree. In my head, the dhis configuration keys are (or can be) connection pool library agnostic (atleast since we started supporting both CP). 

It is better to not have tight coupling with the connection pool library that will then force
 us to name our configuration keys matching theirs. `connection.pool.min_idle` currently is not a DHIS2 configuration key and need not be. 

In an ideal world, we would have "generic" configuration pool configuration keys in DHIS2 with some unit (seconds / milliseconds) or whatever. And during instantiation of the chosen CP, these config keys values will be translated to the libraries configuration keys. We have the possibility to do these translations right now  (without creating duplicate DHIS2 config keys) with the way we instantiate the CP. 

Maybe for this PR, we translate the` connection.pool.min_size` into the minIdle configuration of the HikariCP then? Since we know it is matching behaviour.

ameenhere: After the backend meeting today, it seems the direction we want to take is to not abstract away the CP config keys. In that case, this PR LGTM!

cjmamo:
@david-mackessy as agreed in the backend meeting, we'll be logging a warning now whenever a C3P0-specific setting is used but the DB pool type is not set `c3p0`.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -359,6 +363,49 @@ private static ComboPooledDataSource createC3p0DbPool(
...
   }
 
+  private static void logC3p0UnusedSettingsWarning(
+      DhisConfigurationProvider dhisConfig, DbPoolType dbPoolType) {
+    Consumer<ConfigurationKey> logConsumer =
+        configurationKey ->
+            log.warn(
+                "`{}` C3P0 setting is not used anymore because the default database pool type is set to `hikari` since DHIS2 v43. Remove this C3P0 setting from your `dhis.conf`. Future versions of DHIS2 may stop giving this warning.",
+                configurationKey.getKey());
+    if (!dbPoolType.equals(DbPoolType.C3P0)) {
```
david-mackessy, line 373: Should we do the same when Hikari-only config is used when pool type is c3p0?
We're essentially trying to clean up (by using log warnings) something that we made confusing to users by not making it clear which config is used with each pool type.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
```diff
@@ -359,6 +363,49 @@ private static ComboPooledDataSource createC3p0DbPool(
...
+          dhisConfig, ANALYTICS_CONNECTION_POOL_MAX_IDLE_TIME_EXCESS_CON, logConsumer);
+      applyIfPropertyExists(dhisConfig, CONNECTION_POOL_IDLE_CON_TEST_PERIOD, logConsumer);
+      applyIfPropertyExists(
+          dhisConfig, ANALYTICS_CONNECTION_POOL_IDLE_CON_TEST_PERIOD, logConsumer);
+      applyIfPropertyExists(dhisConfig, CONNECTION_POOL_TEST_ON_CHECKOUT, logConsumer);
+      applyIfPropertyExists(dhisConfig, ANALYTICS_CONNECTION_POOL_TEST_ON_CHECKOUT, logConsumer);
+      applyIfPropertyExists(dhisConfig, CONNECTION_POOL_TEST_ON_CHECKIN, logConsumer);
+      applyIfPropertyExists(dhisConfig, ANALYTICS_CONNECTION_POOL_TEST_ON_CHECKIN, logConsumer);
+      applyIfPropertyExists(dhisConfig, CONNECTION_POOL_NUM_THREADS, logConsumer);
+      applyIfPropertyExists(dhisConfig, ANALYTICS_CONNECTION_POOL_NUM_THREADS, logConsumer);
```
david-mackessy, line 402: I'm not for or against logging each config individually vs logging once with an array of configs. Just wondering was it considered and why this approach was used?

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
cjmamo: I don't think it's worth the effort since C3P0 is going away. We can log a warning though saying that C3P0 is deprecated.

dhis-2/dhis-support/dhis-support-hibernate/src/main/java/org/hisp/dhis/datasource/DatabasePoolUtils.java
cjmamo: Didn't give it too much thought. I can tweak the logging if you think it will be too verbose but my guess is not many C3P0-specific settings are used practice.

david-mackessy: No it's fine 👍. Just curious.

PR #21376, opened 2025-07-09
feat: Introduce preferred search operators on TEA [DHIS2-12183]

muilpp:
This PR introduces the preferred search operator field on a TEA.
It's an optional field that's added to the TEA to define the recommended operator for search or comparison. 
This is merely a suggestion for the end user and will not be enforced.

Ticket: https://dhis2.atlassian.net/browse/DHIS2-12183

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -163,7 +163,7 @@ public enum QueryOperator {
       case IEQ -> EQ;
       case NIEQ -> NEQ;
       case ILIKE -> LIKE;
-      case NLIKE -> NLIKE;
+      case NILIKE -> NLIKE;
       default -> this;
     };
   }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -151,7 +151,7 @@ public enum QueryOperator {
 
   /**
    * Case‑insensitive operators are analytics specific and should not be used in tracker, because
-   * the values there are already converted to lowercase before comparison. For now, we are not
+   * the values there, are already converted to lowercase before comparison. For now, we are not
    * enforcing this rule in the API, so those operators can still be used. Adding such validation
    * would be a breaking change, and we are not ready for that yet.
    *
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -85,9 +85,6 @@ public enum QueryOperator {
 
   private static final Set<QueryOperator> UNARY_OPERATORS = EnumSet.of(NULL, NNULL);
 
-  private static final Set<QueryOperator> CASE_INSENSITIVE_OPERATORS =
-      EnumSet.of(IEQ, NIEQ, ILIKE, NILIKE);
-
   private final String value;
 
   private final boolean nullAllowed;
@@ -150,10 +147,6 @@ public enum QueryOperator {
 
   public boolean isUnary() {
     return UNARY_OPERATORS.contains(this);
-  }
-
-  public boolean isCaseInsensitive() {
-    return CASE_INSENSITIVE_OPERATORS.contains(this);
   }
 
   /**
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -33,6 +33,7 @@ import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.replaceOnce;
 
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Set;
 import lombok.Getter;
@@ -149,6 +150,9 @@ public enum QueryOperator {
     return UNARY_OPERATORS.contains(this);
   }
 
+  private static final Set<QueryOperator> TRACKER_OPERATORS =
+      EnumSet.of(EQ, GT, GE, LT, LE, LIKE, IN, SW, EW, NULL, NNULL);
+
   /**
    * Case‑insensitive operators are analytics specific and should not be used in tracker, because
    * the values there, are already converted to lowercase before comparison. For now, we are not
@@ -166,5 +170,9 @@ public enum QueryOperator {
       case NILIKE -> NLIKE;
       default -> this;
     };
+  }
+
+  public static Set<QueryOperator> getTrackerOperators() {
+    return Collections.unmodifiableSet(TRACKER_OPERATORS);
   }
 }
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -153,25 +153,6 @@ public enum QueryOperator {
   private static final Set<QueryOperator> TRACKER_OPERATORS =
       EnumSet.of(EQ, GT, GE, LT, LE, LIKE, IN, SW, EW, NULL, NNULL);
 
-  /**
-   * Case‑insensitive operators are analytics specific and should not be used in tracker, because
-   * the values there, are already converted to lowercase before comparison. For now, we are not
-   * enforcing this rule in the API, so those operators can still be used. Adding such validation
-   * would be a breaking change, and we are not ready for that yet.
-   *
-   * <p>This method should therefore be used to map case‑insensitive operators, which are analytics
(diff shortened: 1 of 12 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -148,4 +148,22 @@ public boolean isBinary() {
...
+   * would be a breaking change, and we are not ready for that yet.
+   *
+   * <p>This method should therefore be used to map case‑insensitive operators to their
+   * case‑sensitive equivalents.
+   */
+  public QueryOperator mapToCaseSensitiveOperator() {
+    return switch (this) {
+      case IEQ -> EQ;
+      case NIEQ -> NEQ;
+      case ILIKE, NLIKE -> LIKE;
```
ameenhere: I don't think NLIKE is the same as LIKE?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -148,4 +148,22 @@ public boolean isBinary() {
...
+  /**
+   * Case‑insensitive operators are analytics specific and should not be used in tracker, because
+   * the values there, are already converted to lowercase before comparison. For now, we are not
+   * enforcing this rule in the API, so those operators can still be used. Adding such validation
+   * would be a breaking change, and we are not ready for that yet.
+   *
+   * <p>This method should therefore be used to map case‑insensitive operators to their
+   * case‑sensitive equivalents.
+   */
+  public QueryOperator mapToCaseSensitiveOperator() {
```
ameenhere: I kind of want to rename this function to something else. mapToCaseSensitiveOperator sounds like we are making it case sensitive. 
something like harmonizeToTrackerQueryOperator or mapToTrackerQueryOperator might sound less intrusive. feel free to add your suggestions.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/TrackedEntityAttributeObjectBundleHook.java
```diff
@@ -71,6 +71,11 @@ public void validate(
                 "Not a valid TextPattern 'TEXT' segment."));
       }
     }
+
+    if (attr.getPreferredSearchOperator() != null) {
```
ameenhere: I dont think we should support "accepting" the I variant as the preferredSearchOperator. Let us validate and block that.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_11__add_tea_preferred_operator.sql
```diff
@@ -0,0 +1,3 @@
+-- Migration script to add the field preferred operator in the trackedentityattribute table
+alter table trackedentityattribute
```
ameenhere, line 2: Can you migrate from a 2.41 instance directly into this branch build and see if any issue show up? 
The reason I am asking is, if we have separate migration files dealing with the same alter table <tablename>, there might be issues with keeping all of it in a single transaction. So just try it out. The test should cover running both this migration, and the minsearchcharacters migration in a single migration attempt.

dhis-2/dhis-test-integration/src/test/resources/tracker/te_with_tea_metadata.json
```diff
@@ -90,6 +91,7 @@
       "translations": [],
       "attributeValues": [],
       "legendSets": [],
+      "preferredSearchOperator": "IEQ",
```
ameenhere: The I variant operators should not be allowed to be preferredSearchOperator. Since this is a new feature, we will not be breaking anything. So let us block this in this PR.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/TrackedEntityAttributeObjectBundleHook.java
muilpp: On the other hand, I variants will be accepted as allowed operators, mostly to avoid breaking the API, right?
As I understand it, I variant operators should not be used in Tracker, but we don’t (or can’t) enforce this rule.
Wouldn’t it be more consistent to accept these operators in both the preferred and allowed fields until we find a better solution?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -148,4 +148,23 @@ public boolean isBinary() {
...
+   *
+   * <p>This method should therefore be used to map case‑insensitive operators, which are analytics
+   * specific, to their case‑sensitive equivalents.
+   */
+  public QueryOperator mapToTrackerQueryOperator() {
+    return switch (this) {
+      case IEQ -> EQ;
+      case NIEQ -> NEQ;
+      case ILIKE -> LIKE;
+      case NLIKE -> NLIKE;
```
enricocolasante:
```suggestion
      case NILIKE -> NLIKE;
```

enricocolasante (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_11__add_tea_preferred_operator.sql
muilpp: During development I tried a few times and it worked.
I imported a 2.41 database (no `minsearchcharacters` no `preferredsearchoperator` present), and after the migration, both fields are present in the database.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/TrackedEntityAttributeObjectBundleHook.java
ameenhere: We should not accept I-variants to be accepted in preferred or allowed/blockedOperators. 

Let us block it as part of TEA metadata saving validation. Either we block it, or we "map it" to the Tracker Operator variant before persistence.  I am in favour of blocking it. 

In the UI, only the Tracker Operators (non-i variants) will be used anyway. 

In the actual Export API, we do the "Mapping of I-variant to Tracker Variant" before the preferred/allowed operator validation kicks in.

ameenhere (changes requested):
Overall LGTM. Only comment is to block preferredSearchOperator having any I-variant operator. (And for the other issue for allowedOperators as wel the same blocking of I-variant should happen)

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -331,6 +331,7 @@ public enum ErrorCode {
   E4079(
       "Program `{0}` category mapping `{1}` has multiple option mappings for Category Option `{2}`"),
   E4080("Program `{0}` category mapping `{1}` has an invalid option mapping `{1}`"),
+  E4081("The provided preferred TEA operator `{0}` is not part of the tracker operators `{1}`"),
```
ameenhere, line 334:
```suggestion
  E4081("The provided preferred TEA operator `{0}` is not part of supported tracker operators `{1}`"),
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
```diff
@@ -208,8 +208,8 @@ private void mapAttributeFilters(
                 || queryFilter.getFilter().length() < tea.getMinCharactersToSearch())) {
           throw new IllegalQueryException(
               String.format(
-                  "At least %d character(s) should be present in the filter to start a search, but the filter for operator %s doesn't contain enough.",
-                  tea.getMinCharactersToSearch(), queryFilter.getOperator()));
+                  "At least %d character(s) should be present in the filter to start a search, but the filter for the TEA %s doesn't contain enough.",
```
ameenhere, line 211: Nitpicky: Do we actually use TEA and TET in error messages? Or do we want to use the full "Tracked Entity Attribute" ?

ameenhere (approved):
Minnor nitpicky comments. Feel free to decide if you want to include it or not.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapper.java
muilpp: I'll change the message in the coming PR

PR #21387, opened 2025-07-10
fix(DHIS-19708): ensure manifest activities field is always populated

kabaros:
backport of https://github.com/dhis2/dhis2-core/pull/21377


KaiVandivier (approved):
Looks good 👍

PR #21400, opened 2025-07-14
chore(manifest-translations): change log level to avoid spamming the logs

kabaros:
eventually all of the bundled apps should have manifest translations, and this might become a warning then, but for now, it's spamming the logs


KaiVandivier (approved):
👍

david-mackessy (approved):

PR #21401, opened 2025-07-14
chore: revert log4j upgrade causing test errors

netroms:
Upgrade is causing Jenkins CI pipeline to fail on tests: 
[ERROR]   RelativePeriodsTest.assertPeriodEquals:826 expected: org.hisp.dhis.period.Period@153ce630<LAST_MONTH> but was: org.hisp.dhis.period.Period@1b3aa246<LAST_MONTH>

### DIFF
--- a/dhis-2/pom.xml
+++ b/dhis-2/pom.xml
@@ -216,7 +216,7 @@
     <prometheus-metrics-model.version>1.3.1</prometheus-metrics-model.version>
 
     <!-- Logging -->
-    <log4j.version>2.25.1</log4j.version>
+    <log4j.version>2.24.3</log4j.version>
     <slf4j.version>1.7.36</slf4j.version>
 
     <!-- Test -->


ameenhere (approved):

david-mackessy (commented):
It's hard to see how this could cause the Period equals test to fail.
Looking at the test `assertPeriodEquals`, it creates 2 periods with a `RelativePeriodEnum`, the constructor `public Period(RelativePeriodEnum isoRelativePeriod)` creates new new start and end dates using `new Date()`.
I'd guess this is just a race condition and one is having a slightly different value sometimes.
Whoever is the Period expert or knows what constitutes equality for this type should look into it I think. If equals impl is ok, then the test setup probably needs tweaking so the same date is used for each instance.
Also, I think the test was failing before this library upgrade wasn't it?

jbee:
> Whoever is the Period expert or knows what constitutes equality for this type should look into it I think. 

@david-mackessy 
I would not claim to be an expert but I learned a fair bit the last weeks and as far as I can tell the implementation is quite wrong. 
IMHO equals should be the ISO value. Ignoring the pesky `dateField` that IMO should not be there at all, type + start + end is wrong because given a type you only need **a date** to create a unique ISO period. You can argue that start is or could be that date but really it can be any date in the time span. If end would be properly stored exclusive start of day that isn't included then end would not work but ATM end it start of day of last day in the period so type + end gives the same ISO than type + start. So really is should not be any combination of type + date because these are **not unique** for a unique ISO value. Hence, ISO should be a DB column with a unique index. The combination of type + date should not have a unique constraint because that is semantically wrong.  Now, since I am in the process of fixing this in the PR for data value import I can tell you that this isn't an easy fix at all even if it does appear that way. So Morten and I agreed to revert this change for the time being as it appears to somehow increase the chance of the race condition that you correctly pointed out.

PR #21408, opened 2025-07-14
chore: Separate changelog logic in preparation to table separation

enricocolasante:
Before separating the event table, we need to separate the hibernate objects that are mapping the table.
In this PR we are creating a new hibernate object for `SingleEvent` and make it map the same table as `Event` object.

***In This PR***
- Create `SingleEvent` hibernate object and `SingleEvent.hbm.xml` mapping file that for now it is still pointing to `event` table.
- Use `SingleEvent` in all relevant places in the code.
- Create `ChangelogableEvent` interface in order to have a generic `EventChangeLogService`.
- Make the `TrackerIdentifierCollector` load all event uids as `Event` and `SingleEvent` types. Then, the strategies (`EventStrategy` and `SingleEventStrategy`) in the preheat will look into the DB and decide the correct type of the event (checking the program type). So, after preheat we already have the correct type for events that are already present in the DB, in `ParamConverter` then we need to decide the type of events that are going to be created. (By convention we are creating invalid events as tracker events). When the strategy is update or delete, we just look into the preheat to decide the type (if not found, we will create a tracker event).
- Update `StrategyPreProcessor` to correctly handle the 2 types of events.
- Make tracker importer send notification using the correct event type.

***Next steps***
- Rename `Event` hibernate object to `TrackerEvent`. It wasn't done in this PR to avoid polluting the logic.
- Update `RelationshiItem` hibernate mapping to link to the correct event.
- Update `ProgramMessage` hibernate mapping to link to the correct event.
- Update `ProgramNotificationInstance` hibernate mapping to link to the correct event.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -584,6 +615,24 @@ public List<String> canRead(
     return errors;
   }
 
+  @Override
+  public List<String> canRead(
+      @Nonnull UserDetails user, SingleEvent event, DataElement dataElement) {
```
muilpp, line 619: Probably not part of this PR, but I thought we weren’t able to validate ACLs on data elements nor attributes, since we can't support that level of granularity.
I see this method is called only when the user requests a file. Is this the only accepted exception of this use case?
@ameenhere

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventChangeLogService.java
```diff
@@ -48,13 +48,14 @@
...
 import org.hisp.dhis.feedback.NotFoundException;
+import org.hisp.dhis.program.ChangeLogableEvent;
 import org.hisp.dhis.program.Event;
 import org.hisp.dhis.tracker.Page;
 import org.hisp.dhis.tracker.PageParams;
 import org.locationtech.jts.geom.Geometry;
 import org.springframework.transaction.annotation.Transactional;
 
-public abstract class EventChangeLogService<T> {
+public abstract class EventChangeLogService<T, R extends ChangeLogableEvent> {
```
muilpp: Minor thing, but by convention `R` should be used as the return type. I don't see it being returned from any method in this class, so I'm thinking if we should use `S` to indicate the second type, as it's recommended.

muilpp (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ChangeLogableEvent.java
```diff
@@ -0,0 +1,41 @@
...
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.program;
+
+import java.util.Date;
+import org.locationtech.jts.geom.Geometry;
+
+public interface ChangeLogableEvent {
```
teleivo: add javadoc to explain its purpose

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Event.java
```diff
@@ -59,7 +59,7 @@
  * @author Abyot Asalefew
  */
 @Auditable(scope = AuditScope.TRACKER)
-public class Event extends SoftDeletableObject {
+public class Event extends SoftDeletableObject implements ChangeLogableEvent {
```
teleivo: Is this going to become TrackerEvent?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  @Override
+  @JsonIgnore
+  public Date getScheduledDate() {
+    return null;
+  }
+
+  @Override
+  @JsonProperty("eventDate")
```
teleivo: why do we still have these these different names 😬 🤔

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @JsonProperty
+  public String getCompletedBy() {
+    return completedBy;
+  }
+
+  @Override
+  @JsonIgnore
+  public Date getScheduledDate() {
+    return null;
```
teleivo: why is this needed?

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/SingleEvent.hbm.xml
```diff
@@ -0,0 +1,82 @@
...
+
+    <id name="id" column="eventid">
+      <generator class="sequence">
+        <param name="sequence_name">programstageinstance_sequence</param>
+      </generator>
+    </id>
+
+    <property name="uid" column="uid" unique="true" length="11" />
+
+    <property name="code" column="code" not-null="false" unique="true" length="50" />
```
teleivo, line 17: are these identifiable fields not usually embedded/imported from some other hibernate file?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -327,6 +328,36 @@ public List<String> canRead(@Nonnull UserDetails user, Event event, boolean skip
     return errors;
   }
 
+  @Override
+  public List<String> canRead(@Nonnull UserDetails user, SingleEvent event) {
+    // always allow if user == null (internal process) or user is superuser
```
teleivo: ` always allow if user == null` how can that be if we add `@Nonnull` on the user param?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -327,6 +328,36 @@ public List<String> canRead(@Nonnull UserDetails user, Event event, boolean skip
...
+      return List.of();
+    }
+
+    Program program = programStage.getProgram();
+    List<String> errors = new ArrayList<>();
+    if (!aclService.canDataRead(user, program)) {
+      errors.add("User has no data read access to program: " + program.getUid());
+    }
+
+    OrganisationUnit ou = event.getOrganisationUnit();
```
teleivo, line 349:
```suggestion
    OrganisationUnit orgUnit = event.getOrganisationUnit();
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/TrackerAccessManager.java
```diff
@@ -88,6 +91,16 @@ public interface TrackerAccessManager {
   List<String> canRead(
       UserDetails user, Event event, DataElement dataElement, boolean skipOwnershipCheck);
 
+  /**
+   * Checks the sharing read access to EventDataValue
```
teleivo, line 95: you say EventDataValue but whats actually checked is DataElement, that is confusing 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/event/EventService.java
```diff
@@ -42,5 +41,5 @@ public interface EventService {
    * @return an {@link Optional} containing the event if found, or an empty {@link Optional} if not
    */
   @Nonnull
```
teleivo: primitive return does not need `@Nonnull`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipOperationParamsMapper.java
```diff
@@ -105,17 +107,27 @@ private Enrollment getEnrollment(UID enrollmentUid, boolean includeDeleted)
...
-    Event event =
-        relationshipStore
-            .findEvent(eventUid, includeDeleted)
-            .orElseThrow(() -> new NotFoundException(Event.class, eventUid));
-    if (!trackerAccessManager
-        .canRead(CurrentUserUtil.getCurrentUserDetails(), event, false)
-        .isEmpty()) {
-      throw new ForbiddenException(Event.class, eventUid);
+    Optional<SingleEvent> singleEvent = relationshipStore.findSingleEvent(eventUid, includeDeleted);
+    Optional<Event> event = relationshipStore.findEvent(eventUid, includeDeleted);
```
teleivo, line 112: It looks like the logic does not require us loading both from the DB. So only load one, check if its present and only if you must get the other one do that.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/job/NotificationSender.java
```diff
@@ -113,6 +115,26 @@ public void send(Notification notification, Event event) {
...
+
+    Event event = TrackerObjectsMapper.map(singleEvent);
+
+    if (notification.scheduledAt() != null) {
+      ProgramNotificationInstance notificationInstance =
+          notificationTemplateService.createNotificationInstance(
+              template, notification.scheduledAt());
+
+      notificationInstance.setEvent(event);
+      notificationInstance.setEnrollment(null);
```
teleivo, line 129: is setting to null needed? isnt it bull by default

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/strategy/EventStrategy.java
```diff
@@ -57,22 +57,32 @@ public EventStrategy(
...
     List<List<String>> uidsPartitions = Lists.partition(uids, 20000);
 
     for (List<String> uidsPartition : uidsPartitions) {
       if (!uidsPartition.isEmpty()) {
         events.addAll(
             getSession()
-                .createQuery("from Event as ev where ev.uid in (:uids)", Event.class)
+                .createQuery(
+                    """
+                        select distinct ev
```
teleivo: why do we need distinct?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/Event.java
enricocolasante: Yes, PR is already in the line

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
enricocolasante: Obsolete comment, it cannot be null. The code is fine.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
enricocolasante: To implement the `ChangeLogableEvent` interface.
Both `SingleEvent` and `TrackerEvent` can be passed to changeLog and there are some properties that can be logged. `ScheduledDate` is one of them even though a single event does not have such property.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/TrackerAccessManager.java
enricocolasante: It is confusing but here we are checking access to the event (data) + the data element (metadata), so it is indeed checking the access to the EventDataValue.
But I accept any suggestion (@muilpp) to make the docs more understandable.

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/SingleEvent.hbm.xml
enricocolasante: I think we cannot use that file here because of the different mapping on `lastUpdatedBy` https://github.com/dhis2/dhis2-core/blob/3180693f6364583f82dc99a39c37a02c3e81ed26/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/common/identifiableProperties.hbm#L5

We are using `lastUpdatedByUserInfo` in `event` table.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
teleivo: Is this somethings we will change at a later stage? Otherwise its a mistake waiting to happen, you think it exists because of its type but its just an NPE 😅

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/SingleEvent.hbm.xml
teleivo: A bit confusing though as Event is a BaseIdentifiableObject/IdentifiableObject so should it not have the identifiable properties 🤷🏻

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
```diff
@@ -162,7 +162,7 @@ public Date getScheduledDate() {
   }
 
   @Override
-  @JsonProperty("eventDate")
```
teleivo: mhm, so if this does not fail any test I wonder why we even have the Json annotations on these classes? we have our view models which we use to serialize to json. Can we remove all of these annotations here? or are these classes still put into json somewhere like a job/queue?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ChangeLogableEvent.java
```diff
@@ -32,6 +32,10 @@
 import java.util.Date;
 import org.locationtech.jts.geom.Geometry;
 
+/**
+ * Represents an event that can be logged in the change log system. Implementations of this
+ * interface should provide information about the properties that can be logged.
```
teleivo: > Implementations of this interface should provide information about the properties that can be logged.

How would implementations provide this information? I don't get what I would need to do.

dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/program/hibernate/SingleEvent.hbm.xml
enricocolasante: We can and we should review the whole `event` table when I am actually creating the `singleevent` table

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/SingleEvent.java
enricocolasante: I think that we can actually remove the annotations.
I will try it in the next PR when I rename `Event` to `TrackerEvent`

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/ChangeLogableEvent.java
enricocolasante: I removed the interface.
It was not clear because it was just an attempt at reducing duplicated code but in the wrong way.
Now every entity is passing to changeLog all the properties that makes sense for that entity and there is no NPE risk anymore.

PR #21411, opened 2025-07-15
fix: Fix negative data statistics

jason-p-pickering:
https://dhis2.atlassian.net/browse/DHIS2-19866

image


david-mackessy (approved):
We usually try to accompany a fix with a test

jbee (approved):

PR #21417, opened 2025-07-15
feat: Add optional specification of script in database locales

jason-p-pickering:
Originally discussed in https://github.com/dhis2/dhis2-core/pull/19783

- Uzbek can be written in either Cyrillic (uz_UZ_Cyrl) or Latin (uz_UZ_Latn) characters. We did not fully support this as a UI or DB language. Note that this is the legacy Java notation, which we continue to support. Moving to BCP 47 notation is not possible due to dependencies on front end apps on the legacy notation. 
- Previously, the third position of the locale supported a "variant". It is not expected that this was used in practice, since it is not exposed via the API and was hard-coded to `null` in the service layer. The semantics of this parameter have been changed from "variant" to "script". 
- Add the optional "script" parameter to the POST `api/dbLocales` endpoint, to allow creation of locales with a 
- Added tests to be sure the new format works.  
- Added new centralized API methods to parse/format locales. 
- Removed uses of deprecated `new Locale` and `Locate.toString`, which did not properly support the use of the script parameter.
- Added BCP-47 style language tag to GET `api/locales/*` endpoint.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/option/OptionSet.java
@@ -12,7 +12,7 @@
  * this list of conditions and the following disclaimer in the documentation
  * and/or other materials provided with the distribution.
  *
- * 3. Neither the name of the copyright holder nor the names of its contributors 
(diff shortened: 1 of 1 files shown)


jason-p-pickering:
Closing this in favor of https://github.com/dhis2/dhis2-core/pull/22819

PR #21422, opened 2025-07-16
chore: Remove FK constraints from unused audit tables [DHIS2-19865]

enricocolasante:
Remove foreign keys from unused `trackedentitydatavalueaudit` and `trackedentityattributevalueaudit` tables.
This change is done as part of this epic because we are going to separate `event` table and all the tables with a foreign key to this table will need to be changed.


muilpp (approved):

teleivo:
> Remove foreign keys from unused `trackedentitydatavalueaudit` and `trackedentityattributevalueaudit` tables. This change is done as part of this epic because we are going to separate `event` table and all the tables with a foreign key to this table will need to be changed.

Are these tables orphaned? Did we move the data that could be moved to changelog and these tables are the remnants? Just to make this clear and document this on the PR itself as well if that is the case.

teleivo (approved):
Did we control the name of the fks?

ameenhere:
> Did we control the name of the fks?

Yes. We created the dhis2-base schema with the fks name, we use the same name in the hbm.xml files.

PR #21430, opened 2025-07-16
feat: Introduce TEA blocked operators [DHIS2-19835]

muilpp:
From now on, administrators can configure which search operators are blocked to be used for a TEA. The main reason is we want to exclude operators known for not being performant. 
If an operator is used and it's blocked for the given TEA, we'll throw an exception.

By default no operators are blocked, so we don't break the API after migration.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
@@ -173,7 +173,7 @@ public enum QueryOperator {
    * <p>This method should therefore be used to map case‑insensitive operators, which are analytics
    * specific, to their case‑sensitive equivalents.
    */
-  public QueryOperator mapToTrackerQueryOperator() {
+  public QueryOperator stripCaseVariant() {
     return switch (this) {
       case IEQ -> EQ;
       case NIEQ -> NEQ;
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
@@ -34,6 +34,7 @@ import static org.hisp.dhis.security.Authorities.F_TRACKED_ENTITY_INSTANCE_SEARC
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -280,6 +281,29 @@ public class OperationsParamsValidator {
           String.format(
               "Operators %s are blocked for attribute '%s'.",
               blockedUsedOperators, attributeFilter.getKey()));
+    }
+  }
+
+  /**
+   * Validates the specified filter contains at least the minimum number of characters to start the
+   * search
+   *
+   * @throws BadRequestException if the filter doesn't contain enough characters
+   */
+  public static void validateMinimumCharactersToSearch(
+      Map.Entry<UID, List<QueryFilter>> attributeFilter, TrackedEntityAttribute tea)
+      throws BadRequestException {
+    List<QueryFilter> binaryFilters =
+        attributeFilter.getValue().stream().filter(qf -> qf.getOperator().isBinary()).toList();
+    for (QueryFilter queryFilter : binaryFilters) {
+      if (tea.getMinCharactersToSearch() > 0
+          && (queryFilter.getFilter() == null
+              || queryFilter.getFilter().length() < tea.getMinCharactersToSearch())) {
+        throw new BadRequestException(
+            String.format(
+                "At least %d character(s) should be present in the filter to start a search, but the filter for the tracked entity attribute %s doesn't contain enough.",
+                tea.getMinCharactersToSearch(), tea.getUid()));
+      }
     }
   }
 }
--- a/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapperTest.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapperTest.java
@@ -562,9 +562,9 @@ class TrackedEntityOperationParamsMapperTest {
 
     Exception exception =
         Assertions.assertThrows(
-            IllegalQueryException.class, () -> mapper.map(trackedEntityOperationParams, user));
+            BadRequestException.class, () -> mapper.map(trackedEntityOperationParams, user));
     assertContains(
-        "At least 2 character(s) should be present in the filter to start a search, but the filter for the TEA "
+        "At least 2 character(s) should be present in the filter to start a search, but the filter for the tracked entity attribute "
             + TEA_1_UID,
         exception.getMessage());
   }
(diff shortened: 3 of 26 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
```diff
@@ -157,10 +160,26 @@ public boolean isUnary() {
...
+  /**
+   * Case‑insensitive operators are analytics specific and should not be used in tracker, because
+   * the values there, are already converted to lowercase before comparison. For now, we are not
+   * enforcing this rule in the API, so those operators can still be used. Adding such validation
+   * would be a breaking change, and we are not ready for that yet.
+   *
+   * <p>This method should therefore be used to map case‑insensitive operators, which are analytics
+   * specific, to their case‑sensitive equivalents.
+   */
+  public QueryOperator mapToTrackerQueryOperator() {
```
ameenhere: Wasn't this already done as part of the other PR?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -331,7 +331,10 @@ public enum ErrorCode {
   E4079(
       "Program `{0}` category mapping `{1}` has multiple option mappings for Category Option `{2}`"),
   E4080("Program `{0}` category mapping `{1}` has an invalid option mapping `{1}`"),
-  E4081("The provided preferred TEA operator `{0}` is not part of the tracker operators `{1}`"),
+  E4081(
+      "The TrackedEntityAttribute preferred operator `{0}` is not part of the supported tracker operators `{1}`"),
```
ameenhere:
```suggestion
      "The preferred search operator `{0}` provided for the tracked entity attribute `{1}` is not part of the supported tracker operators `{2}`"),
```

ameenhere: If tracked entity attribute uid is not available, then 
The preferred search operator `{0}` provided for the tracked entity attribute is not part of the supported tracker operators `{1}`

Is also fine

dhis-2/dhis-api/src/main/java/org/hisp/dhis/feedback/ErrorCode.java
```diff
@@ -331,7 +331,10 @@ public enum ErrorCode {
...
+  E4082(
+      "The TrackedEntityAttribute preferred operator `{0}` is blocked for the selected TrackedEntityAttribute `{1}`"),
```
ameenhere:
```suggestion
      "The preferred search operator `{0}` is blocked for the selected tracked entity attribute `{1}`"),
```

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/OperationsParamsValidator.java
```diff
@@ -254,4 +260,27 @@ public Set<OrganisationUnit> validateOrgUnits(Set<UID> orgUnitIds, UserDetails u
...
+   *
+   * @throws BadRequestException if the operator(s) are blocked
+   */
+  public static void validateAttributeOperators(
+      Entry<UID, List<QueryFilter>> attributeFilter, TrackedEntityAttribute tea)
+      throws BadRequestException {
+    Set<QueryOperator> blockedUsedOperators =
+        attributeFilter.getValue().stream()
+            .map(QueryFilter::getOperator)
+            .map(QueryOperator::mapToTrackerQueryOperator)
```
ameenhere: Can this mapping be done way before? Like when `List<QueryFilter>> attributeFilter` is being built the first time around?

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
```diff
@@ -299,6 +300,18 @@ void shouldContainPreferredSearchOperatorWhenSet() {
     assertAttributePreferredOperator(json, Set.of("LIKE"));
   }
 
+  @Test
+  void shouldContainBlockedSearchOperatorsWhenSet() {
+    teaA.setBlockedSearchOperators(Set.of(QueryOperator.LIKE, QueryOperator.NNULL));
+    manager.update(teaA);
+
+    JsonObject json =
+        GET("/trackedEntityAttributes?indexableOnly=true&filter=name:in:[AttributeA]&fields=*")
```
ameenhere: indexableOnly?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackerevent/TrackerEventOperationParamsMapperTest.java
```diff
@@ -523,6 +527,49 @@ void shouldMapOrgUnitAndModeWhenModeAllAndUserIsAuthorized(String userName)
     assertEquals(ALL, params.getOrgUnitMode());
   }
 
+  @Test
```
ameenhere, line 530: Do we have a test to make sure the "mapToTrackerOperator" works as expected?

dhis-2/dhis-services/dhis-service-tracker/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityOperationParamsMapperTest.java
```diff
@@ -638,6 +639,55 @@ void shouldMapTeaWhenTeaMinCharactersSetAndReached()
...
+                    new QueryFilter(ILIKE, "0")))
+            .build();
+
+    TrackedEntityQueryParams queryParams = mapper.map(trackedEntityOperationParams, user);
+    assertContainsOnly(
+        List.of(TEA_2_UID.getValue()), UID.toUidValueSet(queryParams.getFilters().keySet()));
+  }
+
+  @Test
+  void shouldNotMapAttributeFiltersWhenOperatorsAreBlocked()
```
enricocolasante:
```suggestion
  void shouldFailToMapAttributeFiltersWhenOperatorsAreBlocked()
```
so it is clear that an exception is thrown.

enricocolasante (approved):
Should we have someone from platform team to at least have a look when we make changes in metadata importer?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/QueryOperator.java
muilpp: Originally yes, but I removed it after we decided we wouldn't accept I-variant operators as the preferred one.

david-mackessy: Would a different name keep it non-tracker specific while still being useful for Tracker?
It just maps to its case sensitive counterpart if found. It's nothing to do with Tracker, it just happens to be useful in Tracker.
maybe `mapToCaseSensitiveOperator` ?

david-mackessy: Otherwise tracker logic is bleeding into other areas of the system which we try to keep separated.

ameenhere: The only reason we moved away from the name `mapToCaseSensitiveOperator` is because in Tracker all searches are case insensitive. I was thinking about code readability when using `mapToCaseSensitiveOperator` in the Tracker service might lead to a false assumption of the search case sensitivity 🤔 

What do you think @muilpp ? I am fine with either way as long as we ensure code readability is not lost

muilpp: I agree with you both, what about `stripCaseVariant`?

IMO, even if we used `mapToTrackerQueryOperator`, we still need to remember Tracker is case insensitive when it comes to searching.

david-mackessy: Sounds ok to me 👍

PR #21431, opened 2025-07-16
chore: Rename Event hibernate object to TrackerEvent [DHIS2-19865]

enricocolasante:
Follow up of https://github.com/dhis2/dhis2-core/pull/21408 work.
Rename `Event` hibernate object to `TrackerEvent`.
At this moment we have `SingleEvent` and `TrackerEvent` objects with respective mapping that are both pointing to `event` table.
The PR is huge because `Event` object is used in a lot of places but there is not so much logic in the PR.

Luckily, OpenAPI is still working the same because `view/Event` object was annotated with `@OpenApi.Shared(name = "TrackerEvent")` and now with the renaming everything is just working the same way.

***Next steps***
- Use `SingleEvent` and `TrackerEvent` in `RelationshipItem` and separate columns in `relationshipitem` table.
- Split `eventchangelog` table to `singleeventchangelog` and `trackereventchangelog` tables
- Split `event_notes` table to `singleevent_notes` and `trackerevent_notes` tables


dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Objects.java
```diff
@@ -90,19 +90,21 @@ public enum Objects {
   PROGRAM("program", Program.class),
   TRACKEDENTITY("trackedEntity", TrackedEntity.class),
   ENROLLMENT("enrollment", Enrollment.class),
-  EVENT("event", Event.class),
+  EVENT("event", TrackerEvent.class), // Event inclues tracker and single events
```
muilpp:
```suggestion
  EVENT("event", TrackerEvent.class), // Event includes tracker and single events
```

muilpp: why are both types of event included if only the `TrackerEvent` class is used?

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/category/optioncombo/DataCategoryOptionComboMergeHandler.java
```diff
@@ -182,7 +182,7 @@ public void handleEvents(
           "delete from event where attributeoptioncomboid in (%s)".formatted(aocIds));
     } else {
       log.info("Merging source events as dataMergeStrategy is LAST_UPDATED");
-
+      // TODO(DHIS2-19702): Should we consider single events?
```
ameenhere, line 185: Yes. At the moment it is possible to have AOC in both Single Event and Tracker Event. 
So I think we should handle it.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/Objects.java
enricocolasante: Class field is only used in tests and do not represent the type that is going to be used. I deprecated the field `clazz` for this class.
The string `event` is what it is used and now it includes all entities in table `event`, later it will include all entities from tables `trackerevent` and `singleevent`.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/handler/TrackerDataElementMergeHandler.java
```diff
@@ -203,24 +203,25 @@ public void handleProgramRuleAction(List<DataElement> sources, DataElement targe
...
+   * @param target {@link DataElement} which will be set as the {@link DataElement} in {@link
+   *     TrackerEvent} eventDataValues
    * @param request merge request
    */
   public void handleEventDataValues(
       List<DataElement> sources, DataElement target, MergeRequest request) {
     Set<UID> sourceDeUids = UID.of(sources.toArray(new DataElement[0]));
     DataMergeStrategy mergeStrategy = request.getDataMergeStrategy();
 
+    // TODO(DHIS2-19702): Should we consider single events
```
ameenhere, line 224: I think we should handle single events here as well. Since data elements could be either for Tracker or for Single Events.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/category/optioncombo/DataCategoryOptionComboMergeHandler.java
enricocolasante: I have a few comments like this in this PR.
Right now things are working because we have all the events in one table, but when we separate the table, we need to take care of those cases.
The question is quite rhetorical.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/split/orgunit/handler/DataOrgUnitSplitHandler.java
```diff
@@ -62,8 +62,9 @@ public void splitData(OrgUnitSplitRequest request) {
     migrate(request, "Interpretation", PARAM_ORG_UNIT);
 
     migrate(request, "ProgramMessage", "recipients." + PARAM_ORG_UNIT);
-    migrate(request, "Event", PARAM_ORG_UNIT);
+    migrate(request, "TrackerEvent", PARAM_ORG_UNIT);
     migrate(request, "Enrollment", PARAM_ORG_UNIT);
+    // TODO(DHIS2-19702): should we consider single events?
```
ameenhere, line 67: Yes.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datastatistics/DefaultDataStatisticsService.java
```diff
@@ -237,12 +237,16 @@ public DataSummary getSystemStatisticsSummary() {
     dataValueCount.put(30, dataValueService.getDataValueCount(30));
     statistics.setDataValueCount(dataValueCount);
 
+    // TODO(DHIS2-19702): Should we consider single events?
```
ameenhere, line 240: Yes. I would also suggest we can expand the data statistics response to include trackerEvent count and singleEvent count separately (in addition to setting the existing eventCount as the sum for backward compatibility).

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/merge/dataelement/handler/TrackerDataElementMergeHandler.java
enricocolasante: Will do in DHIS2-19702.
Right now things are working because there is only one event table.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/split/orgunit/handler/DataOrgUnitSplitHandler.java
enricocolasante: Will do in DHIS2-19702.
Right now things are working because there is only one event table.

dhis-2/dhis-services/dhis-service-reporting/src/main/java/org/hisp/dhis/datastatistics/DefaultDataStatisticsService.java
enricocolasante: Will do in DHIS2-19702.
Right now things are working because there is only one event table.
We can have separate and aggregated statistics. Will do when we are separating tables.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackerAccessManager.java
```diff
@@ -401,7 +402,7 @@ public List<String> canCreate(
 
   @Override
   public List<String> canUpdate(
-      @Nonnull UserDetails user, Event event, boolean skipOwnershipCheck) {
+      @Nonnull UserDetails user, TrackerEvent event, boolean skipOwnershipCheck) {
```
ameenhere, line 405: SingleEvent access check already handled separately? Then this method could be further simplified maybe

enricocolasante: Yes, there is already a `canRead` for both types.
`canUpdate` is only used when checking relationships, it will come in next PR where I am fixing `relationshipItem` modeling.

enricocolasante:
> One thing that was not clear to me, but maybe already handled/planned to handle is ProgramMessage and ProgramNotification for Single Events. Maybe you already have it in your mind.

Yes, it is present in next step section in the description.

PR #21433, opened 2025-07-17
feat: Prevent duplicate event creation from SCHEDULEEVENT rule action [DHIS2-19849]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-19849

This PR continues the work from [#19868](https://github.com/dhis2/dhis2-core/pull/19868). It introduces a validation step in the rule-engine–based scheduling logic to avoid creating duplicate events.

If the program stage is non-repeatable, the system will skip scheduling a new event if an existing one is already present for the same enrollment.
If the program stage is repeatable, the scheduling logic will proceed and allow the new event to be created.


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/programrule/config/ServiceConfig.java
```diff
@@ -106,8 +107,8 @@ public Map<ProgramRuleActionType, ProgramRuleActionValidator> programRuleActionV
             ProgramRuleActionType.HIDEFIELD,
             getProgramRuleActionValidatorByClass(BaseProgramRuleActionValidator.class))
         .put(
-            ProgramRuleActionType.CREATEEVENT,
```
enricocolasante, line 109: Do we need to keep the base validator for CREATEEVENT?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EventProgramStageMapSupplier.java
```diff
@@ -81,9 +81,10 @@ public void preheatAdd(TrackerObjects trackerObjects, TrackerPreheat preheat) {
 
     List<String> notRepeatableProgramStageUids =
         trackerObjects.getEvents().stream()
-            .map(Event::getProgramStage)
+            .map(Event::getProgram)
```
enricocolasante: Why do you need to change this code?
We are potentially introducing a lot of unnecessary data into the Preheat.

dhis-2/dhis-support/dhis-support-test/src/main/resources/tracker/tracker_multi_text_attribute_data.json
enricocolasante, line 1: Why is this change here? Do you need to rebase?

ameenhere:
Just a thought, would be a good first version if we only allow one scheduled event to exist even if it is a repeatable program stage? So if there already is a scheduled event for that program stage, then no new scheduled event is created. @enricocolasante  @zubaira thoughts?

zubaira:
> Just a thought, would be a good first version if we only allow one scheduled event to exist even if it is a repeatable program stage? So if there already is a scheduled event for that program stage, then no new scheduled event is created. @enricocolasante @zubaira thoughts?

@ameenhere that's what we agreed upon.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/ScheduledEventProgramStageMapSupplier.java
```diff
@@ -0,0 +1,109 @@
...
+import org.hisp.dhis.tracker.imports.preheat.TrackerPreheat;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.RowCallbackHandler;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Zubair Asghar
+ */
+@Component
```
enricocolasante: You either use something different than `programStageWithEvents` in Preheat or you should work with just one supplier.
I would go with one supplier that will put into the `programStageWithEvents` all the program stages by enrollment that have at least one not skipped event (so we do not care if the program stage is repeatable or not).
So we could remove this supplier.
At the moment is very confusing, 2 different suppliers are feeding the same map with some overlapping data.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
```diff
@@ -0,0 +1,114 @@
...
+      MetadataIdentifier attributeOptionCombo,
+      MetadataIdentifier program,
+      MetadataIdentifier orgUnit) {
+
+    final String scheduledAt = validationEffect.data();
+    final UID programRule = validationEffect.rule();
+    final UID programStage = validationEffect.field();
+
+    // check for existing event
+    if (bundle
```
enricocolasante, line 72: Did we took into consideration payloads that could trigger the same rule more than once and create multiple scheduled events in the same run?
For example importing an enrollment and an event could trigger the rule twice and we could end up with 2 created scheduled events.

We need to think about it and I would start writing a test to validate this case.

zubaira:
> LGTM. Can we create a test to check that 2 entities triggering the same rule in a payload will only create one scheduled event?

Added

enricocolasante:
> > LGTM. Can we create a test to check that 2 entities triggering the same rule in a payload will only create one scheduled event?
> 
> Added

I cannot see any added test

zubaira:
> > > LGTM. Can we create a test to check that 2 entities triggering the same rule in a payload will only create one scheduled event?
> > 
> > 
> > Added
> 
> I cannot see any added test

look for `shouldSuccessfullyScheduleOnlyOneEvent` in [this](https://github.com/dhis2/dhis2-core/pull/19868) PR.

enricocolasante:
> > > > LGTM. Can we create a test to check that 2 entities triggering the same rule in a payload will only create one scheduled event?
> > > 
> > > 
> > > Added
> > 
> > 
> > I cannot see any added test
> 
> look for `shouldSuccessfullyScheduleOnlyOneEvent` in [this](https://github.com/dhis2/dhis2-core/pull/19868) PR.

I am really really confused about these 2 PRs.
How can you have a test about something that is implemented in this PR in another PR?
It seems to me that if the implementation is not present and the test is passing then there is something wrong.
Anyway, can you create one single PR? We are going to merge altogether anyway, I cannot see the point of the 2 separated PRs

zubaira:
> > > > > LGTM. Can we create a test to check that 2 entities triggering the same rule in a payload will only create one scheduled event?
> > > > 
> > > > 
> > > > Added
> > > 
> > > 
> > > I cannot see any added test
> > 
> > 
> > look for `shouldSuccessfullyScheduleOnlyOneEvent` in [this](https://github.com/dhis2/dhis2-core/pull/19868) PR.
> 
> I am really really confused about these 2 PRs. How can you have a test about something that is implemented in this PR in another PR? It seems to me that if the implementation is not present and the test is passing then there is something wrong. Anyway, can you create one single PR? We are going to merge altogether anyway, I cannot see the point of the 2 separated PRs

This PR specifically addresses the prevention of duplicate events — ensuring that an already-created event is not re-created. The test in question, however, covers a different scenario, where a single payload produces two identical events. regarding merging of the PRs, i think https://github.com/dhis2/dhis2-core/pull/21433 this one already has changes from the other PR.

ameenhere (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/imports/TrackerScheduleEventControllerTest.java
```diff
@@ -0,0 +1,371 @@
...
+                        orgUnit.getUid(),
+                        "2025-11-11",
+                        enrollment.getOrganisationUnit().getUid(),
+                        enrollment.getProgram().getUid(),
+                        programStageA.getUid(),
+                        enrollment.getTrackedEntity().getUid()))
+            .content(HttpStatus.OK)
+            .as(JsonImportReport.class);
+
+    assertEquals(
```
enricocolasante, line 178: We have an issue in the report, even though the event is only created once, the warning for the creation of the event is appearing twice.
If you try:
`assertEquals(
        "E1320",
        importReport
            .getObject("validationReport.warningReports[1]")
            .getString("warningCode")
            .string());`
You can see the other warning there.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/programrule/executor/ScheduleEventHelper.java
```diff
@@ -0,0 +1,109 @@
...
+      TrackerBundle bundle,
+      UID enrollment,
+      MetadataIdentifier attributeOptionCombo,
+      MetadataIdentifier program,
+      MetadataIdentifier orgUnit) {
+
+    final String scheduledAt = validationEffect.data();
+    final UID programRule = validationEffect.rule();
+    final UID programStage = validationEffect.field();
+
```
enricocolasante, line 66: To solve the issue that I commented in the test, we can add a guard clause here to return as soon as we discover that an event for the same program stage is already present in the scheduled event map.

`if (bundle.getProgramRuleScheduledEvents().get(programStage) != null) {
        return Optional.empty();
    }`
This way we are avoiding the duplication of the event and we are not raising duplicated warnings.

zubaira: fixed

enricocolasante (approved):

PR #21440, opened 2025-07-19
chore: use userdetails in preheat

netroms:
### Summary
Further refactors use of User object to use UserDetails instead. This PR focuses on the importer preheat.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserDetails.java
```diff
@@ -286,6 +288,15 @@ default boolean canIssueUserRole(UserRole role, boolean canGrantOwnUserRole) {
     return getAllAuthorities().containsAll(role.getAuthorities());
   }
 
+  default boolean canManage(UserGroup userGroup) {
```
david-mackessy, line 291: tests for this new method would be nice

david-mackessy (approved):

PR #21441, opened 2025-07-19
chore: convert user to userdetails refactoring

netroms:
### Summary:
Refactoring the use of User to UserDetails, this PR focuses to eliminate the use of UserDetails.fromUser() method, which usually means UserDetails can be used of properly refactored from the top, typically the controller.


dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
```diff
@@ -550,6 +550,9 @@ Map<String, Optional<Locale>> findNotifiableUsersWithPasswordLastUpdatedBetween(
    */
   UserDetails createUserDetails(String userUid) throws NotFoundException;
 
+  @CheckForNull
+  UserDetails createUserDetailsSafe(@Nonnull String userUid);
```
david-mackessy, line 554: It would be nice to use the `UID` type, especially for new methods, if possible.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
```diff
@@ -495,6 +502,8 @@ public WebMessage replicateUser(@PathVariable String uid, HttpServletRequest req
       return conflict(result.getErrorMessage());
     }
 
+    Session session = entityManager.unwrap(Session.class);
```
david-mackessy, line 505: Using the entity manager and session at the controller layer feels a bit wrong. It kind of hints that this should be done at the service level.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/user/UserService.java
netroms: Yeah, would be great to do a big refactor just for that. Feels right to do that after UserDetails are mostly done.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/user/UserController.java
netroms: yes, totally. We should refactor this one.

vietnguyen (approved):

PR #21478, opened 2025-07-22
Dhis2 19583 2.40

david-mackessy:


dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/config/ServiceConfig.java
```diff
@@ -186,13 +184,9 @@ public NamedParameterJdbcTemplate namedParameterJdbcTemplate(JdbcTemplate jdbcTe
...
   public RetryTemplate retryTemplate() {
     ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
-
     backOffPolicy.setInitialInterval(Long.parseLong((String) initialInterval.getObject()));
-
-    Map<Class<? extends Throwable>, Boolean> exceptionMap = new HashMap<>();
-    exceptionMap.put(MetadataSyncServiceException.class, true);
     SimpleRetryPolicy simpleRetryPolicy =
-        new SimpleRetryPolicy(Integer.parseInt((String) maxAttempts.getObject()), exceptionMap);
+        new SimpleRetryPolicy(Integer.parseInt((String) maxAttempts.getObject()));
```
david-mackessy, line 189: default behaviour is to use `Exception` when no exception map supplied.

PR #21485, opened 2025-07-23
chore: Separate eventid column in relationshipitem table [DHIS2-19889]

enricocolasante:
This is the first of a series of PRs to prepare the tables that have a foreign key to `event` table for the split into `trackerevent` and `singleevent` tables.

***In this PR***
- Update `RelationshipItem` hibernate object: split `event` field into `singleEvent` and `trackerEvent` fields and update XML mapping file.
- Split `eventid` column in `relationshipitem` table with a flyway script and populate those columns with the correct data (put into `singleevent` column only single events and same for `trackerevent` column).
- Update `RelationshipItemMapper` in web layer to map either a single event or a tracker event to a view event.

***Migration plan***

***Next steps***
- Split `eventchangelog` table to `singleeventchangelog` and `trackereventchangelog` tables
- Split `event_notes` table to `singleevent_notes` and `trackerevent_notes` tables

### DIFF
--- a/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/relationship/hibernate/RelationshipItem.hbm.xml
+++ b/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/relationship/hibernate/RelationshipItem.hbm.xml
@@ -27,7 +27,7 @@
 
         <many-to-one name="trackerEvent" class="org.hisp.dhis.program.TrackerEvent"
                      column="trackereventid"
-                     foreign-key="fk_relationshipitem_programstageinstanceid"/>
+                     foreign-key="fk_relationshipitem_trackereventid"/>
 
         <many-to-one name="singleEvent" class="org.hisp.dhis.program.SingleEvent"
                      column="singleeventid"


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_16__update_relationshipitem_table.sql
```diff
@@ -0,0 +1,24 @@
+alter table if exists relationshipitem add column if not exists trackereventid bigint;
```
ameenhere: I think we can group several of these alter table commands together. Which should reduce migration time. 
At the very least all add column can happen in a single DDL sql. And drop constraint in a single one and add constraints in a single one. 

But I am also wondering about the migration as a whole. This leaves us in an intermediate state right?  The overall migration for this initiative is going to be big. So I think we should avoid migrating to intermediate stages. We need to plan this properly.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_16__update_relationshipitem_table.sql
```diff
@@ -0,0 +1,24 @@
...
+alter table if exists relationshipitem add column if not exists singleeventid bigint;
+alter table if exists relationshipitem drop constraint if exists fk_relationshipitem_trackereventid;
+alter table if exists relationshipitem drop constraint if exists fk_relationshipitem_singleeventid;
+alter table if exists relationshipitem
+    add constraint fk_relationshipitem_trackereventid FOREIGN KEY (trackereventid) REFERENCES event(eventid);
+alter table if exists relationshipitem
+    add constraint fk_relationshipitem_singleeventid FOREIGN KEY (singleeventid) REFERENCES event(eventid);
```
ameenhere, line 10: The reason I said intermediate stage is because of this. This ideally should be the single event table eventually right? 

Let us discuss how we should plan our migration

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_16__update_relationshipitem_table.sql
enricocolasante: Done

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_16__update_relationshipitem_table.sql
enricocolasante: I added a section in the description.
We thought a solution to not have any intermediate state in the flyway process in the final version but we will have intermediate states while the development is in progress.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipItem.java
```diff
@@ -51,59 +53,7 @@ public class RelationshipItem implements EmbeddedObject {
...
-    return id;
-  }
-
-  public void setId(int id) {
-    this.id = id;
-  }
-
-  @JsonProperty
-  @JacksonXmlProperty(namespace = DxfNamespaces.DXF_2_0)
-  @JsonSerialize(as = BaseIdentifiableObject.class)
```
teleivo, line 68: Are the `@Json` annotations just remnants of the days when these classes were also used to return the JSON in our API? (asking as I would not know if removing `@JsonSerialize(as = BaseIdentifiableObject.class)` and replacing it with `@Getter` would have the same effect.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipItemMapper.java
```diff
@@ -257,6 +260,30 @@ Set<TrackedEntityAttributeValue> mapTrackedEntityAttributeValues(
   @Mapping(target = "geometry")
   TrackerEvent map(TrackerEvent event);
 
+  @BeanMapping(ignoreByDefault = true)
+  @Mapping(target = "uid")
+  @Mapping(target = "enrollment", qualifiedByName = "mapEnrollmentForEvent")
```
teleivo, line 265: will we remove this mapping later?

dhis-2/dhis-api/src/main/java/org/hisp/dhis/relationship/RelationshipItem.java
enricocolasante: Yes, we are not serializing this class anymore.
We are not using it at any moment in the API and we are not serializing this class to create the import job, tracker model is used for that.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/export/relationship/RelationshipItemMapper.java
enricocolasante: Do you mean the mapping to the enrollment?
We need to be careful here because this is defining what we show in the API.
We need to provide the program and right now this is happening through the enrollment, we could start deprecate enrollment and have program field mapped.

PR #21495, opened 2025-07-25
chore: annotation mapping for DashboardItem

vietnguyen:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateTimeUnit.java
```diff
@@ -49,13 +48,13 @@
  */
 public class DateTimeUnit {
   /** Year of date. Required. */
-  @Nonnull private int year;
+  private int year;
```
vietnguyen, line 51: Got an error with this `@Nonnull` while running test on local. The annotation doesn't work with primitive type.

jbee: A primitive field/parameter/variable should also not have it, it is correct to remove it

david-mackessy (approved):

PR #21500, opened 2025-07-28
chore: Split event_notes table [DHIS2-19914]

enricocolasante:
Split `event_notes` table into `trackerevent_notes` and `singleevent_notes`.

***In this PR***
- Update `SingleEvent` hibernate mapping: point to the correct table and update foreign key constraint name.
- Rename `Event` hibernate mapping to `TrackerEvent`.
- Update `TrackerEvent` hibernate object: point to the correct table and update foreign key constraint name.
- Update maintenance store.

***Migration plan***

***Next steps***

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
@@ -133,9 +133,11 @@ public class JdbcMaintenanceStore implements MaintenanceStore {
               union all select noteid from singleevent_notes
               union all select noteid from enrollment_notes)""",
           // delete other objects related to events
-          "delete from relationshipitem where eventid in " + eventSelect,
+          "delete from relationshipitem where trackereventid in " + eventSelect,
+          "delete from relationshipitem where singleeventid in " + eventSelect,
           "delete from trackedentitydatavalueaudit where eventid in " + eventSelect,
-          "delete from eventchangelog where eventid in " + eventSelect,
+          "delete from trackereventchangelog where eventid in " + eventSelect,
+          "delete from singleeventchangelog where eventid in " + eventSelect,
           "delete from programmessage where eventid in " + eventSelect,
           "delete from programnotificationinstance where eventid in " + eventSelect,
           // finally delete the events
@@ -219,9 +221,11 @@ public class JdbcMaintenanceStore implements MaintenanceStore {
               union all select noteid from singleevent_notes
               union all select noteid from enrollment_notes)""",
           // delete other entries linked to events
-          "delete from relationshipitem where eventid in " + eventSelect,
+          "delete from relationshipitem where trackereventid in " + eventSelect,
+          "delete from relationshipitem where singleeventid in " + eventSelect,
           "delete from trackedentitydatavalueaudit where eventid in " + eventSelect,
-          "delete from eventchangelog where eventid in " + eventSelect,
+          "delete from trackereventchangelog where eventid in " + eventSelect,
+          "delete from singleeventchangelog where eventid in " + eventSelect,
           "delete from programmessage where eventid in " + eventSelect,
           "delete from programnotificationinstance where eventid in " + eventSelect,
           // delete other entries linked to enrollments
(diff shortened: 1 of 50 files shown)


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/JdbcNoteStore.java
```diff
@@ -83,6 +84,25 @@ INSERT INTO event_notes(eventid, noteid, sort_order)
     jdbcTemplate.update(sql, Map.of("event", event.getValue(), "noteId", noteId));
   }
 
+  public void saveSingleEventNote(
+      @Nonnull UID event, @Nonnull Note note, @Nonnull UserDetails user) {
```
teleivo, line 88: in dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/DefaultNoteService.java we get the event to check if exists, would be better if we pass it in to get rid of two subselects here

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/JdbcNoteStore.java
```diff
@@ -83,6 +84,25 @@ INSERT INTO event_notes(eventid, noteid, sort_order)
...
+    long noteId = saveNote(note, user);
+    String sql =
+        """
+                INSERT INTO singleevent_notes(eventid, noteid, sort_order)
+                VALUES ((select eventid from event where uid = :event),
+                        :noteId,
+                        coalesce(
+                                    (select max(sort_order) + 1
```
teleivo, line 96: What is the sort_order here?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_19__split_event_notes_table.sql
```diff
@@ -0,0 +1,34 @@
+alter table if exists event_notes
+    drop constraint if exists fk_programstageinstancecomments_programstageinstanceid;
+
+CREATE TABLE trackerevent_notes (LIKE event_notes INCLUDING ALL);
```
teleivo, line 4: did you double-check the tables defaults/contraints/indices? just to make sure we are not missing anything like we have in other tables

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
```diff
@@ -125,8 +125,13 @@ public int hardDeleteEvents(List<String> eventsToDelete, String eventSelect, Str
           "delete from programmessage_phonenumbers where programmessagephonenumberid in "
               + pmSelect,
           // delete related events notes
-          "delete from event_notes where eventid in " + eventSelect,
-          "delete from note where noteid not in (select noteid from event_notes union all select noteid from enrollment_notes)",
+          "delete from trackerevent_notes where eventid in " + eventSelect,
+          "delete from singleevent_notes where eventid in " + eventSelect,
+          """
+            delete from note where noteid not in
```
teleivo, line 131: What is the `note` table?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_19__split_event_notes_table.sql
enricocolasante: There are just 3 columns here:
 - `noteid` is primary key
 - (`eventid`, `sort_order`) is a key
 So all columns are not-nullable.
 We do not have any of the issues that we found in other tables.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/JdbcNoteStore.java
enricocolasante: It is used to have a defined order for notes when more than one note is imported.

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/maintenance/jdbc/JdbcMaintenanceStore.java
enricocolasante: All the notes are stored in the `note` table, then `event_notes`, `enrollment_notes` tables are used to implement the many to many constraint.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/note/JdbcNoteStore.java
enricocolasante: We are postponing this improvement for now because we will need to change the mapping in service layer to include `id` field.

teleivo (approved):

ameenhere (approved):

PR #21502, opened 2025-07-28
chore: Split eventid column in notification tables [DHIS2-19888]

enricocolasante:
Split `eventid` column in `programmessage` and `programnotificationinstance` tables.
I didn't manage to split one table at a time because the logic is quite coupled.
The goal here was to not change too much what we have and copy the logic that worked for events to work both for tracker and single events.

***In this PR***
- Update `ProgramMessage` hibernate mapping: split `event` into `trackerEvent` and `singleEvent`.
- Update `ProgramNotificationInstance` hibernate object: split `event` into `trackerEvent` and `singleEvent`.
- `ProgramMessage` and `ProgramNotificationInstance` are hibernate objects that are also exposed in the API. To make it work and maintain compatibility with previous version, `singleEvent` and `trackerEvent` fields were added with `@JsonIgnore` annotation.
- `ProgramNotificationInstanceController` only has `GET` endpoints, so we are mapping either `singleEvent` or `trackerEvent` to `event` field in the result.
- We are doing the same thing for `ProgramMessageController` `GET` endpoints.
- In `ProgramMessageController` `POST` endpoint, the mapping from `event` to `trackerEvent` or `singleEvent` is done in the service.
- Update deletion handler and maintenance store.

***Migration plan***

***Next steps***

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
@@ -400,7 +400,7 @@ public class TrackerObjectsMapper {
     return dbNote;
   }
 
-  // To remove when refactoring ProgramNotificationInstanceController
+  // TODO(tracker): To remove when refactoring ProgramNotificationInstanceController
   @Deprecated
   public static TrackerEvent map(SingleEvent singleEvent) {
     if (singleEvent == null) {


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/DefaultProgramMessageService.java
```diff
@@ -200,8 +201,10 @@ private boolean hasDataWriteAccess(ProgramMessage message) {
 
     if (message.hasEnrollment()) {
       object = message.getEnrollment().getProgram();
-    } else if (message.hasEvent()) {
-      object = message.getEvent().getProgramStage();
+    } else if (message.hasTrackerEvent()) {
+      object = message.getTrackerEvent().getProgramStage();
+    } else if (message.hasSingleEvent()) {
+      object = message.getSingleEvent().getProgramStage();
```
ameenhere, line 207: Possibly something for future, we will not be having ProgramStage anymore in SingleEvents right?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultProgramNotificationService.java
```diff
@@ -379,11 +419,12 @@ protected TrackerEvent postProcessObject(TrackerEvent event) {
 
   private MessageBatch createScheduledMessageBatchForDay(
       ProgramNotificationTemplate template, Date day) {
-    List<TrackerEvent> events = getWithScheduledNotifications(template, day);
 
+    // Single events cannot be scheduled.
+    List<TrackerEvent> events = getWithScheduledNotifications(template, day);
```
ameenhere, line 424: Can you double check this?  This may be about scheduled notifications. And I think scheduled notifications can also happen based on "Completion" of event, which is also possible for Single Events.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/ProgramMessageOperationParamMapper.java
```diff
@@ -59,14 +61,34 @@ public class ProgramMessageOperationParamMapper {
...
   public ProgramMessageQueryParams map(ProgramMessageOperationParams operationParams)
       throws NotFoundException {
+    // TODO(DHIS2-19702): Refactor this method
+    Optional<TrackerEvent> trackerEvent = Optional.empty();
+    Optional<SingleEvent> singleEvent = Optional.empty();
     Enrollment enrollment = getEntity(operationParams.getEnrollment(), Enrollment.class);
-    TrackerEvent event = getEntity(operationParams.getEvent(), TrackerEvent.class);
+    if (operationParams.getEvent() != null) {
+      trackerEvent = findEntity(operationParams.getEvent(), TrackerEvent.class);
+      if (trackerEvent.isEmpty()) {
```
teleivo, line 70: why do we throw if we cannot find a trackerEvent, could it not still eb a singleEvent?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
```diff
@@ -400,7 +400,8 @@ private TrackerObjectsMapper() {
     return dbNote;
   }
 
-  // TODO: should we remove this method when we will separate the event tables?
+  // To remove when refactoring ProgramNotificationInstanceController
+  @Deprecated
```
teleivo, line 404: set forRemoval true if this is going to be removed soon

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/job/NotificationTrigger.java
```diff
@@ -35,6 +35,7 @@
 public enum NotificationTrigger {
   ENROLLMENT,
   ENROLLMENT_COMPLETION,
-  EVENT_COMPLETION,
+  TRACKER_EVENT_COMPLETION,
```
teleivo, line 38: is this ever stored in the DB?

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_20__split_event_columns_in_notification_tables.sql
```diff
@@ -0,0 +1,52 @@
...
+             JOIN programstage ps ON ev.programstageid = ps.programstageid
+             JOIN program p ON ps.programid = p.programid
+    WHERE p.type = 'WITHOUT_REGISTRATION'
+);
+update programmessage
+set trackereventid = eventid
+where singleeventid is null;
+
+alter table if exists programmessage drop column if exists eventid;
+ ------
```
teleivo, line 26: odd comment

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageController.java
```diff
@@ -77,7 +80,9 @@ public List<ProgramMessage> getProgramMessages(ProgramMessageRequestParams reque
       throws BadRequestException, ConflictException, NotFoundException {
     ProgramMessageOperationParams params = requestParamMapper.map(requestParams);
 
-    return programMessageService.getProgramMessages(params);
+    List<ProgramMessage> programMessages = programMessageService.getProgramMessages(params);
+    programMessages.forEach(this::setEvent);
```
teleivo, line 84: are these hibernate objects and if so is this an issue that we set the fields?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/notification/ProgramNotificationInstanceController.java
```diff
@@ -88,17 +90,25 @@ public class ProgramNotificationInstanceController {
...
 
-    TrackerEvent storedEvent = null;
+    TrackerEvent storedTrackerEvent = null;
+    SingleEvent storedSingleEvent = null;
     if (requestParams.getEvent() != null) {
       // TODO(tracker) jdbc-hibernate: check the impact on performance
-      storedEvent = manager.get(TrackerEvent.class, requestParams.getEvent());
-      if (storedEvent != null) {
-        if (storedEvent.getProgramStage().getProgram().isRegistration()) {
+      // TODO(DHIS2-19702): Simplify this code
```
teleivo, line 99: keep the jdbc-hibernate comment or make sure its in the jira issue

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/message/ProgramMessage.java
```diff
@@ -65,6 +67,10 @@ public class ProgramMessage extends BaseIdentifiableObject implements Serializab
   @JsonSerialize(as = BaseIdentifiableObject.class)
   private Enrollment enrollment;
 
+  @JsonIgnore private TrackerEvent trackerEvent;
```
teleivo, line 70: Some comments would be good as this will stay for a while AFAIK. `event` is now not used for persistence but only for the view, right? The other two fields are only used for persistence but not for the view, right? That's what I would document as that is not straightforward when looking at the class. A `@JsonProperty getEvent()` instead of an `event` field might also make this clearer.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/DefaultProgramMessageService.java
enricocolasante: Yes, at some point single events will link directly to the program.
It is going to happen is a later PR.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultProgramNotificationService.java
enricocolasante: @zubaira Can you help here?
What I understood from the code is that here we are sending notifications for scheduled events and single event schedule date is always null because a single event cannot be in `SCHEDULED` state.

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/message/ProgramMessageOperationParamMapper.java
enricocolasante: Yes, but no.
This is the most confusing PR, from the next one things should start to make sense.
At this moment we have `TrackerEvent` and `SingleEvent` hibernate objects that point to the same table `event`, so if we try to get a `TrackerEvent` with the manager and it is not there, we know that there is no event with such a uid in table `event`.
Then based on the type of the program we decide to retrieve it as a single event or not.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
enricocolasante: This refactor is not part of this epic.
I would like to remove it soon but I am not sure when we will do so.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/job/NotificationTrigger.java
enricocolasante: No.
This is only used to decide the notifications to be sent.
It only lives in `TrackerNotificationThread`.

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_20__split_event_columns_in_notification_tables.sql
enricocolasante: Removed

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/message/ProgramMessageController.java
enricocolasante: They are hibernate objects but the `event` field is not mapped, so it is safe to set it. It is only used as a view.
I will add the comments as suggested later.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/notification/ProgramNotificationInstanceController.java
enricocolasante: I am keeping the jdbc-hibernate comment

dhis-2/dhis-api/src/main/java/org/hisp/dhis/program/message/ProgramMessage.java
enricocolasante: Added the comments to `ProgramMessage` and `ProgramNotificationInstance`

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/program/notification/DefaultProgramNotificationService.java
ameenhere: Maybe you are right. 🤔

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/bundle/TrackerObjectsMapper.java
teleivo: mhm, keep it as `TODO(tracker)` so we can easily find all of our todos in the entire codebase?

PR #21505, opened 2025-07-29
Dhis2 19583

david-mackessy:


dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
```diff
@@ -137,7 +136,7 @@ public class CategoryCombo extends BaseMetadataObject
   @Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
   private List<Category> categories = new ArrayList<>();
 
-  @OneToMany(cascade = CascadeType.ALL, fetch = FetchType.LAZY, mappedBy = "categoryCombo")
+  @OneToMany(fetch = FetchType.LAZY, mappedBy = "categoryCombo")
```
david-mackessy, line 139: Let me know if there was a reason the cascade all was added @vietnguyen

netroms (approved):

vietnguyen (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryCombo.java
vietnguyen: Thanks David. Not really remember reason for this. I only had problem with the categories collection why working on the annotation migration.

PR #21513, opened 2025-07-30
feat: add new settings for Azure maps API key

netroms:
# Summary
Adds a new settings key for Azure maps API keys. Bing maps API keys are being deprecated, but needs to be kept during the transition phase.

### DIFF
--- a/dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
+++ b/dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
@@ -514,6 +514,10 @@ public non-sealed interface SystemSettings extends Settings {
     return asString("keyBingMapsApiKey", "");
   }
 
+  default String getAzureMapsApiKey() {
+    return asString("keyAzureMapsApiKey", "");
+  }
+
   default Date getLastMetaDataSyncSuccess() {
     return asDate("keyLastMetaDataSyncSuccess", new Date(0));
   }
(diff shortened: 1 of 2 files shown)


larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
```diff
@@ -514,6 +514,10 @@ default String getBingMapsApiKey() {
     return asString("keyBingMapsApiKey", "");
   }
 
+  default String getAzureMapsApiKey() {
```
david-mackessy, line 517: Should `@Confidential` be used for this key?
It's not mentioned anywhere so I'm not sure if it has been considered.

netroms: I don't think so, this key needs to be exposed in the web layer, so that the web components can use it to call the Azure maps API from the web page. None of the other API keys are confidential. Marking them confidential will ignore them when calling the settings endpoint.

larshelge: I think @netroms is right. We have/had the same issue with the Maps app with Google Maps/Bing Maps. The base maps etc are retrieved by Javascript APIs and are handled client side.

larshelge (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/setting/SystemSettings.java
david-mackessy: cool 👍

PR #21519, opened 2025-07-31
fix: Add integrity check to detect geometry columns with incorrect SRID (not 4326) [DHIS2-12547]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-12547

### DIFF
--- a/.github/workflows/analyse-pr.yml
+++ b/.github/workflows/analyse-pr.yml
@@ -23,7 +23,7 @@ jobs:
     if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -9,7 +9,7 @@ jobs:
   check-formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,7 +34,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           # We must fetch at least the immediate parents so that if this is
           # a pull request then we can checkout the head.
--- a/.github/workflows/deploy-instance.yml
+++ b/.github/workflows/deploy-instance.yml
@@ -39,7 +39,7 @@ jobs:
           LATEST_IM_TAG=$(git ls-remote --tags --sort='v:refname' https://github.com/dhis2-sre/im-manager | tail -1 | awk '{print $2}')
           echo "LATEST_IM_TAG=$LATEST_IM_TAG" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           repository: dhis2-sre/im-manager
           ref: ${{ env.LATEST_IM_TAG }}
--- a/.github/workflows/destroy-instance.yml
+++ b/.github/workflows/destroy-instance.yml
@@ -40,7 +40,7 @@ jobs:
           LATEST_IM_TAG=$(git ls-remote --tags --sort='v:refname' https://github.com/dhis2-sre/im-manager | tail -1 | awk '{print $2}')
           echo "LATEST_IM_TAG=$LATEST_IM_TAG" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           repository: dhis2-sre/im-manager
           ref: ${{ env.LATEST_IM_TAG }}
--- a/.github/workflows/generate-merge-boms.yml
+++ b/.github/workflows/generate-merge-boms.yml
@@ -16,7 +16,7 @@ jobs:
         working-directory: dhis-2/
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
--- a/.github/workflows/run-api-analytics-tests.yml
+++ b/.github/workflows/run-api-analytics-tests.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     if: "contains(github.event.pull_request.labels.*.name, 'run-api-analytics-tests')"
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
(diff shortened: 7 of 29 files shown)


dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/gis/tracker_geometry_invalid_srid.yaml
```diff
@@ -0,0 +1,88 @@
...
+       )
+  SELECT COUNT(*) AS value,
+         100.0 * COUNT(*) / NULLIF((SELECT COUNT(*) FROM all_geom), 0) AS percent
+  FROM invalid_geom;
+
+details_sql: >-
+  SELECT uid, table_name as name,
+         'Expected SRID 4326 but found ' || srid AS comment
+  FROM (
+    SELECT 'trackedentity' AS table_name, uid, ST_SRID(geometry) AS srid
```
jason-p-pickering: I think since we are touching a potentially large table, this check should be marked as `is_slow: TRUE`, which will have the effect of excluding it from the default checks. It would need to be triggered manually in this case by the user.

jason-p-pickering:
https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity

See here for many examples of the integration tests we usually write for the checks

zubaira:
> Thanks for this new check. Do you think you can add an unit test?

@jason-p-pickering done.

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityTrackerEntitiesInvalidSRIDControllerTest.java
jason-p-pickering, line 1: It would be good (as with the other integration tests) to test 
1) `COUNT(*) / NULLIF((SELECT COUNT(*) FROM all_geom), 0)`...this is basically the case when we have no data. 
2) The case when we have only valid results . 

I think the current test is generally fine, but maybe adding the case when there are NO TEIs would be good.

zubaira: added

jason-p-pickering (approved):

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
```diff
@@ -510,6 +510,7 @@ Set<String> addSQLChecksToFlattedReport() {
     checks.add("organisation_units_violating_exclusive_group_sets");
     checks.add("orgunits_compulsory_group_count");
     checks.add("indicators_violating_exclusive_group_sets");
+    checks.add("tracker_geometry_invalid_srid");
```
jbee, line 513: I think we use a naming pattern where the check itself also has the main object as prefix, so here it should start with "programs" I think. But then in the check it seems this is not about programs but tracked entities so then maybe it is wrongly under `programs/` ? Something does not add up here naming wise.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/gis/tracker_geometry_invalid_srid.yaml
```diff
@@ -0,0 +1,89 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+---
+name: tracker_geometry_invalid_srid
+description: Tracker entities with geometry not using SRID 4326
+section: Tracker
+section_order: 4
+severity: WARNING
+details_id_type: tracker
```
jbee, line 33: There is no object of type `tracker`. It seems the object type here is mixed and can be of several tables. In that case I think we should not state this property ATM.

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/gis/tracker_geometry_invalid_srid.yaml
```diff
@@ -0,0 +1,89 @@
...
+  ),
+       invalid_geom AS (
+           SELECT * FROM all_geom WHERE ST_SRID(geometry) != 4326
+       )
+  SELECT COUNT(*) AS value,
+         100.0 * COUNT(*) / NULLIF((SELECT COUNT(*) FROM all_geom), 0) AS percent
+  FROM invalid_geom;
+
+details_sql: >-
+  SELECT uid, table_name as name,
```
jbee: I don't think this conforms to the pattern we use for detail results.
While it has the minimal form the `uid` and `name` do not belong to the same object but are a tuple of a different type of information. This is not how this is supposed to work. Have a look at `DataIntegrityIssue`, there is a `refs` `List<String>` which is where further information is supposed to go. 
Your result should then have 4 columns, `uid`, `name` (of the object with the `uid`), `comment` (you can hard code to `null` or use this for your table hint) and an array for the `refs`. So either make it 3 columns and use the comment for the table hint or put it in an array with a single value of the table name at column index 4. Since this also is of mixed type I'd think we might just use the comment here. Then it also makes sense that `details_id_type` is not defined.

jason-p-pickering:
So what happens when the client has sent a geometry with an SRID other than 0?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/programs/gis/tracker_geometry_invalid_srid.yaml
jason-p-pickering: This check is a bit special, since it refers to data and not metadata objects (as do many of the other checks). Looking at this from the implementation side of things, I might want to check some of this data. While we do make a "best effort" to link to the objects in the maintenance app, this check cannot really follow that pattern. I would need to know though 1) Is it an event, an enrollment or a TEI? 2) What is the UID? in order to investigate it more thoroughly.  I think the current SQL does provide that information...it is just that it is not really going to work in the current UI. Personally, I think that is OK, as this is a "slow" check and indicates that something is maybe special about it.

jbee: It can still follow the `uid` + `name` of the same object + additional info in `comment` or `refs` pattern. then you get the info you need and at some point we can improve functionality with the pattern still followed on a fundamental level.

zubaira: changed it to return `uid`, `name` ( of object with `uid` ) and `comment` for additional information.

ameenhere (approved):

PR #21522, opened 2025-08-01
feat: Org unit DE filter in enrollment aggregate [DHIS2-19937]

maikelarabori:
The analytics enrollment aggregate endpoint needs to support filtering by data elements of type organization unit. This should be done for the `filter` and `dimension` params.

`filter=PFDfvmGpsR3.n1rtSHYf6O6:IN:ImspTQPwCqd`
`dimension=PFDfvmGpsR3.n1rtSHYf6O6:IN:ImspTQPwCqd`

These changes require a few checks in different places, which are all part of the same requirement.
To make it more readable and maintainable, I created the `EnrollmentOrgUnitFilterHandler` class, so all checks are concentrated in a single place.

Example of queries:

```
/api/analytics/enrollments/aggregate/WSGAb5XwJ3Y.json?dimension=ou:ImspTQPwCqd&dimension=pe:LAST_5_YEARS&filter=PFDfvmGpsR3.n1rtSHYf6O6:IN:ImspTQPwCqd,&displayProperty=NAME&totalPages=false&relativePeriodDate=2025-07-01

/api/analytics/enrollments/aggregate/eBAyeGv0exc.json?dimension=Zj7UnCAulEk.S33cRBsnXPo:IN:ImspTQPwCqd;USER_ORGUNIT;USER_ORGUNIT_GRANDCHILDREN;lc3eMKXaEfw;bL4ooGhyHRQ;LEVEL-H1KlN4QIauv;USER_ORGUNIT_CHILDREN;Vth0fbpFcsO;OU_GROUP-CXw2yu5fodb&filter=ou:USER_ORGUNIT,pe:LAST_10_YEARS&totalPages=false&displayProperty=NAME&pageSize=100&page=1&includeMetadataDetails=true&relativePeriodDate=2025-07-01
```

### DIFF
--- a/.github/workflows/analyse-pr.yml
+++ b/.github/workflows/analyse-pr.yml
@@ -23,7 +23,7 @@ jobs:
     if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -9,7 +9,7 @@ jobs:
   check-formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,7 +34,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           # We must fetch at least the immediate parents so that if this is
           # a pull request then we can checkout the head.
--- a/.github/workflows/deploy-instance.yml
+++ b/.github/workflows/deploy-instance.yml
@@ -39,7 +39,7 @@ jobs:
           LATEST_IM_TAG=$(git ls-remote --tags --sort='v:refname' https://github.com/dhis2-sre/im-manager | tail -1 | awk '{print $2}')
           echo "LATEST_IM_TAG=$LATEST_IM_TAG" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           repository: dhis2-sre/im-manager
           ref: ${{ env.LATEST_IM_TAG }}
--- a/.github/workflows/destroy-instance.yml
+++ b/.github/workflows/destroy-instance.yml
@@ -40,7 +40,7 @@ jobs:
           LATEST_IM_TAG=$(git ls-remote --tags --sort='v:refname' https://github.com/dhis2-sre/im-manager | tail -1 | awk '{print $2}')
           echo "LATEST_IM_TAG=$LATEST_IM_TAG" >> $GITHUB_ENV
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           repository: dhis2-sre/im-manager
           ref: ${{ env.LATEST_IM_TAG }}
--- a/.github/workflows/generate-merge-boms.yml
+++ b/.github/workflows/generate-merge-boms.yml
@@ -16,7 +16,7 @@ jobs:
         working-directory: dhis-2/
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
--- a/.github/workflows/run-api-analytics-tests.yml
+++ b/.github/workflows/run-api-analytics-tests.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     if: "contains(github.event.pull_request.labels.*.name, 'run-api-analytics-tests')"
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Free disk space on runner (Ubuntu)
         shell: bash
--- a/.github/workflows/run-api-tests.yml
+++ b/.github/workflows/run-api-tests.yml
@@ -20,7 +20,7 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,7 +14,7 @@ jobs:
   unit-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
@@ -60,7 +60,7 @@ jobs:
   integration-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
@@ -111,7 +111,7 @@ jobs:
   integration-h2-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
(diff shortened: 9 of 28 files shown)


dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/EventQueryParams.java
```diff
@@ -1006,6 +1009,21 @@ public boolean hasFilterPeriods() {
     return isNotEmpty(getFilterPeriods());
   }
 
+  /**
+   * Verifies whether there is an org. unit filter associated with a query item.
+   *
+   * @return true if the org. unit filter is present, false otherwise.
+   */
+  public boolean hasOrgUnitFilterInItem() {
```
luciano-fiandesio, line 1017: Not critical: you may consider using a simple boolean at EventQueryParams build time, to avoid streaming over two lists on every call.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
```diff
@@ -1061,26 +1068,42 @@ protected String getAggregatedEnrollmentsSql(List<GridHeader> headers, EventQuer
...
+
+    List<String> list = Arrays.asList(orgUnitColumns, periodColumns, headerColumns);
     String columns =
-        (!isBlank(orgColumns) ? orgColumns : "," + ORGUNIT_DIM_ID)
-            + (!isBlank(periodColumns) ? "," + periodColumns : EMPTY)
-            + (!isBlank(headerColumns) ? "," + headerColumns : EMPTY);
+        list.stream().filter(StringUtils::isNotBlank).collect(Collectors.joining(", "));
+
+    String join = EMPTY;
+    if (handleEnrollmentOrgUnitFilter(params)) {
```
luciano-fiandesio: critical: I think that there is a bug here. When `handleEnrollmentOrgUnitFilter` is true, the new branch drops `filterWhereClause`.

The code should be something like:

```
join = " join analytics_event_" + … + whereClause + filterWhereClause
       + " and ev.eventstatus != 'SCHEDULE'";
```

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/JdbcEnrollmentAnalyticsManager.java
```diff
@@ -545,9 +546,32 @@ protected String getSelectClause(EventQueryParams params) {
...
   }
 
+  /**
+   * This method switches or add a new prefix to the given column if needed. It's applied for
+   * specific cases driven by the business rules of the invoker.
+   *
+   * @param column to be prefixed.
+   * @return the prefixed column (if required).
+   */
+  private String addEventPrefix(String column) {
```
luciano-fiandesio: possible bug: this implementation may be a bit fragile for certain cases.
For instance:

before  : `ST_Y(ax.geometry)`
after   : `ST_Y(ax.geometry)`

--

before  : count(*) as value
after   : ev.count(*) as value

---

before  : `ax.value as "A03MvHHogjR"`
after   : `ev.ax.value as "A03MvHHogjR"` 

---

I suggest you add some tests to make sure that the implementation is solid.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/EnrollmentQueryHelper.java
```diff
@@ -136,7 +136,7 @@ public static Set<String> getOrgUnitLevelColumns(EventQueryParams params) {
     if (!params.isOrganisationUnitMode(SELECTED) && !params.isOrganisationUnitMode(CHILDREN)) {
       Set<String> levels = new LinkedHashSet<>();
 
-      for (DimensionalItemObject itemObject : params.getDimensionOrFilterItems(ORGUNIT_DIM_ID)) {
+      for (DimensionalItemObject itemObject : params.getDimensionOptions(ORGUNIT_DIM_ID)) {
```
luciano-fiandesio: possible bug: 

The method used to call `getDimensionOrFilterItems(ORGUNIT_DIM_ID)`, which returns **both**
* organisation units passed in the `dimension=` parameter and  
* organisation units passed in the `filter=` parameter.

It now calls `getDimensionOptions(ORGUNIT_DIM_ID)`, which **ignores filter values**.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/AbstractJdbcEventAnalyticsManager.java
maikelarabori: The "filterWhereClause" is, actually, introducing a bug. It's duplicating some filtering from "filterWhereClause".
I have to review it, just adding it back will not work.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/EventQueryParams.java
maikelarabori: We have other methods doing the same. It does not worth making this one different from others. The lists are tiny.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/EnrollmentQueryHelper.java
maikelarabori: It, actually, fixes a bug. We do not want "filter" showing up as columns.

larshelge: Looks good. I guess ideally, we want to do bug fixes in separate PRs whenever it is feasible. This way we can test and backport bug fixes in isolation without impact from other changes.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/EnrollmentOrgUnitFilterHandler.java
```diff
@@ -0,0 +1,94 @@
...
+public class EnrollmentOrgUnitFilterHandler {
+
+  /**
+   * Checks if org. unit item filter handling is needed for enrollments/aggregate.
+   *
+   * @param params the {@link EventQueryParams}.
+   * @param item the {@link QueryItem}.
+   * @return true if handling is required, false otherwise.
+   */
+  public static boolean handleEnrollmentOrgUnitFilter(EventQueryParams params, QueryItem item) {
```
larshelge: Could this method be called `hasEnrollmentOrgUnitFilter`? I think that fits a bit better where the method is called.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/EnrollmentOrgUnitFilterHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        && item.getValueType() == ORGANISATION_UNIT;
+  }
+
+  /**
+   * Checks if org. unit item filter handling is needed for enrollments/aggregate.
+   *
+   * @param params the {@link EventQueryParams}.
+   * @return true if handling is required, false otherwise.
+   */
+  public static boolean handleEnrollmentOrgUnitFilter(EventQueryParams params) {
```
larshelge: Could this method be called hasEnrollmentOrgUnitFilter? I think that fits a bit better where the method is called.

dhis-2/dhis-services/dhis-service-analytics/src/main/java/org/hisp/dhis/analytics/event/data/JdbcEnrollmentAnalyticsManager.java
```diff
@@ -545,9 +546,32 @@ protected String getSelectClause(EventQueryParams params) {
...
+    if (column.startsWith("ax.")) {
+      column = column.replace("ax.", "ev.");
+    } else if (!column.contains("(")) {
```
larshelge: This one feels a bit fragile/prone to bugs later and hard to understand. Any chance we can make it a bit more readable?

larshelge (commented):
The Jira issue is a bit thin, trying to understand the context.

https://dhis2.atlassian.net/browse/DHIS2-19937

PR #21548, opened 2025-08-05
chore: annotation mapping for CategoryOption

vietnguyen:
### Summary
- This PR migrate Hibernate xml mapping to annotation for `CategoryOption` class.

### Add `FavoritableObject` interface
- This is an attempt to break down the `IdentifiableObject` interface. As the `favorites` property is only used in the `AbstractCrudController` for  and/remove favorites, this turned out to be a simple task.
- The similar approach will be used for other more common properties such as `sharing`, `attributeValues`.

### Other changes
- Transient properties `href` and `access` are moved to `BaseMetadataObject` because those properties are always processed in the API FieldFilterService for all metadata objects. 

### Test
- Add few controller tests to verify `CategoryOption` import


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntitiesExportControllerTest.java
```diff
@@ -142,12 +142,15 @@ void setUp() throws IOException {
...
+    setUpToBeMigrated();
 
-    trackerObjects = testSetup.importTrackerData();
+    manager.flush();
+    manager.clear();
     testSetup.importTrackerData("tracker/tracker_multi_text_attribute_data.json");
 
     manager.flush();
     manager.clear();
+    trackerObjects = testSetup.importTrackerData();
```
vietnguyen, line 153: All of the above code movement is caused by a mix between Hibernate queries and JDBC queries, which prevents data from being flushed to the database, so the delete execution cannot be performed.

dhis-2/dhis-api/src/test/java/org/hisp/dhis/category/CategoryOptionTest.java
```diff
@@ -152,20 +150,12 @@ private static Stream<Arguments> categoryOptionEqualsParams() {
...
-            "shortName",
-            "description",
-            QueryModifiers.builder().build(),
-            false));
+        Arguments.of("name", "uid", "code", "shortName", null, isEqual),
+        Arguments.of("name", "uid", "code", "shortName diff", null, isNotEqual),
+        Arguments.of("name", "uid", "code diff", "shortName", null, isNotEqual),
+        Arguments.of("name", "uid diff", "code", "shortName", null, isNotEqual),
+        Arguments.of("name diff", "uid", "code", "shortName", null, isNotEqual),
+        Arguments.of("name", "uid", "code", "shortName", QueryModifiers.builder().build(), false));
```
vietnguyen, line 158: I removed the `description` from the `equals` and `hashcode` method in `CategoryOption` class, therefore this test needs to be updated.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
```diff
@@ -29,53 +29,163 @@
...
+  @Column(name = "name", nullable = false, unique = true, length = 230)
+  private String name;
+
+  @Column(name = "shortname", nullable = false, unique = true, length = 50)
+  private String shortName;
+
+  @Column(name = "description", columnDefinition = "text")
+  private String description;
+
+  @Column(name = "formName", length = 230)
```
jbee: does this column actually use the camel case name?

jbee (approved):

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
```diff
@@ -29,53 +29,163 @@
...
+    implements DimensionalItemObject, SystemDefaultMetadataObject, Serializable {
+
   public static final String DEFAULT_NAME = "default";
 
+  @Id
+  @GeneratedValue(strategy = GenerationType.SEQUENCE)
+  @Column(name = "categoryoptionid")
+  private long id;
+
+  private String code;
```
david-mackessy, line 131: was this field missed with `@Column` or is there another reason?

vietnguyen: The `@Column` annotation is optional. All properties declared in the class will be mapped as long as we use the `@Entity` at the class level.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
```diff
@@ -48,14 +52,14 @@
 @MappedSuperclass
 public class BaseMetadataObject implements MetadataObject {
 
-  @Column(name = "uid", unique = true, nullable = false, length = 11)
```
david-mackessy: why was null prevention removed? Is it guaranteed somewhere else? Do we rely on the DB constraint existing for this?

david-mackessy (approved):
Looks good, just a few small questions left

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
vietnguyen: I think it will be auto converted to all lower case even if we use camel. But will update this to make it consistent.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/BaseMetadataObject.java
vietnguyen: Thanks David, nice catch! This was for debugging, I added the null check back now.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/category/CategoryOption.java
david-mackessy: Maybe we should have a chat about how we'd like this to look. I'd prefer to have all annotated so it's very quick and easy to know it's a persisted field. Mixing approaches can leave questions for the reader.

PR #21577, opened 2025-08-08
chore: Remove FK to enrollment from singleevent table [DHIS2-19959]

enricocolasante:
Clean up `singleevent` table. In particular, remove `enrollmentid` column from it, hence FK to `enrollment` table and completely remove the dummy enrollment logic from the system.

***In this PR***
- Remove program event enrollment service and store and remove the dummy enrollment logic from `CopyService` and from `ProgramObjectBundleHook`. @david-mackessy can you please review this part as it is part of the metadata importer?

- Remove `enrollment` field from tracker domain object `Event`. The consequence of this removal is that we cannot anymore use `Event` in `TrackerObjects` because some fields are missing if we will later map it to `TrackerEvent`. So we need to change `events` field in `TrackerObjects` to a list of `TrackerEvent` and later in `ParamsConverter` we will correctly transform the to the right event type based on the program type.
- Where relevant, get the program for single events through the program stage instead of the enrollment.

***Development plan***
This PR is changing a fundamental table in DHIS2 that is also used by analytics. Therefore, at this moment, analytics table generation is broken.
This PR should be reviewed and approved but it will not be merged until all the tests (in particular analytics tests, that are the only failing ones at the moment) will pass. Following PRs will be merged into this branch.
I already spoke with @maikelarabori and when the `trackerevent` and `singleevent` tables are cleaned up and they will not be furtherly modified, they will update the analytics code in order to work properly with the changes in the DB.

***Migration plan***

***Next steps***
- Merge all migration skip for the epic in one file.
- Document all breaking changes and new features.


dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/ParamsConverter.java
```diff
@@ -95,7 +94,7 @@ private static List<TrackerEvent> buildTrackerEvents(
   }
 
   private static List<SingleEvent> buildSingleEvents(
-      List<Event> events, TrackerPreheat preheat, TrackerImportStrategy importStrategy) {
+      List<TrackerEvent> events, TrackerPreheat preheat, TrackerImportStrategy importStrategy) {
```
muilpp, line 93: why don't we use `List<SingleEvent> events` in this case?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Event.java
muilpp, line 1: this class can't be removed yet?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EventProgramStageMapSupplier.java
```diff
@@ -85,15 +86,14 @@ public void preheatAdd(TrackerObjects trackerObjects, TrackerPreheat preheat) {
...
-            .filter(ps -> ps.getProgram().isRegistration())
             .filter(ps -> !ps.getRepeatable())
             .map(ProgramStage::getUid)
             .distinct()
             .toList();
 
     List<UID> enrollmentUids =
         trackerObjects.getEvents().stream()
-            .map(Event::getEnrollment)
+            .map(TrackerEvent::getEnrollment)
```
muilpp, line 96: why do we need to get the enrollment in the `EventProgramSupplier`?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/domain/Event.java
enricocolasante: No, it is needed to avoid duplication in validation phase and program rule phase.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EventProgramStageMapSupplier.java
enricocolasante: This is the `EventProgramStageMapSupplier` and it is providing a map of program stages and enrollments that have at least one event.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/ParamsConverter.java
enricocolasante: This is explained in the second bullet point in the description.
`TrackerObjects` field `events` now is a list of `TrackerEvent`s and we need to build the proper type (single or tracker) checking the program type. So we are always transforming a `TrackerEvent` into either a `SingleEvent` or a `TrackerEvent`.
This is going to be this confusing until we will not change the API and accept explicitly tracker or single events.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/preheat/supplier/EventProgramStageMapSupplier.java
muilpp: It confuses me a bit because it starts with `EventProgram`.
Since it deals with enrollments, I assume it can only use tracker events, right?
Would it make sense to rename it `TrackerEventProgramStageMapSupplier`?

enricocolasante: Done

muilpp (approved):

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/copy/CopyService.java
```diff
@@ -210,17 +201,6 @@ private Map<String, Program.ProgramStageTuple> copyStages(Program original, Prog
...
-  /**
-   * This method copies enrollments of event programs only, as these kinds of enrollments act as
-   * placeholders and are considered metadata. It won't copy enrollments of tracker programs because
-   * those are actual data.
-   */
-  private void copyEnrollments(Program original, Program copy) {
-    List<Enrollment> enrollments =
-        copyList(copy, eventProgramEnrollmentService.getEnrollments(original), Enrollment.copyOf);
-    enrollments.forEach(identifiableObjectManager::save);
-  }
```
david-mackessy, line 222: Is there a similar replacement for an Event type that is now considered metadata? This seems like more of a product question I think (how the feature should work). It might be fine to just delete this code, with no replacement of a similar type (if there even is one), or it might require replacing with a similar type (if there even is one). Any thoughts @stian-sandvold ?

david-mackessy (commented):
Hi @enricocolasante:
- I've left a comment on the `CopyService` change. It looks like more of a Product question to me.
- I can't really comment on the `ProgramObjectBundleHook`. It's not clear to me why that code is there & what happens if it's removed. Maybe @vietnguyen or @jbee might be able to comment on it?

dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/copy/CopyService.java
enricocolasante: There is no replacement for what is deleted here but everything needed by the data is in Program and Program Stage metadata. The dummy enrollment was there just as a placeholder to link to all the events but the same result can be achieved checking for all the events linking to the Program Stage.

enricocolasante:
> Hi @enricocolasante:
> 
> * I've left a comment on the `CopyService` change. It looks like more of a Product question to me.
> * I can't really comment on the `ProgramObjectBundleHook`. It's not clear to me why that code is there & what happens if it's removed. Maybe @vietnguyen or @jbee might be able to comment on it?

Thanks @david-mackessy.
Lets wait for some input from the others but in the meantime I am going to merge this PR in the branch because the change in the DB schema is needed by analytics to start their work and we are not merging to master until analytics team manages to make it work on their side.
So we can continue the discussion here and if there is something to change/fix we can do it on the epic branch.

stian-sandvold:
> > Hi @enricocolasante:
> > 
> > * I've left a comment on the `CopyService` change. It looks like more of a Product question to me.
> > * I can't really comment on the `ProgramObjectBundleHook`. It's not clear to me why that code is there & what happens if it's removed. Maybe @vietnguyen or @jbee might be able to comment on it?
> 
> Thanks @david-mackessy. Lets wait for some input from the others but in the meantime I am going to merge this PR in the branch because the change in the DB schema is needed by analytics to start their work and we are not merging to master until analytics team manages to make it work on their side. So we can continue the discussion here and if there is something to change/fix we can do it on the epic branch.

Looked through the PR: Both the CopyService and ProgramObjectBundleHook looks good to me.

PR #21610, opened 2025-08-12
fix: Add flyway to enforce valid srid for geometry column [DHIS2-19965]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-19965


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_22__Add_valid_srid_constraint.sql
```diff
@@ -0,0 +1,36 @@
+-- https://dhis2.atlassian.net/browse/DHIS2-19965
+
+UPDATE event
```
ameenhere: I think it is better we wait for this PR. Since we are almost done with separation of event table into trackerevent and singleevent table. We should apply this constraint there instead.

zubaira: sounds good.

jason-p-pickering: I suggest to proceed and so that we can backport this one. Once the tables are updated with that separation PR, we can make a very small change to this constraint, but maintain the backported versions as is (if you are planning to backport it)

ameenhere: The current plan is not to backport this flyway migration. The integrity checks should be used for detecting/resolving in older versions and they are already backported.

ameenhere (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityTrackerEntitiesInvalidSRIDControllerTest.java
enricocolasante, line 1: I am not sure I understand why we are removing this test.
It is still possible to add a geometry with an invalid SRID, right?
If not, we should just completely remove the integrity check

zubaira: The initial plan was to add only an integrity check, but later we decided to enforce this with a database-level constraint. So yes, that check is no longer needed

enricocolasante (approved):

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityTrackerEntitiesInvalidSRIDControllerTest.java
jason-p-pickering: Keep the check. You never know what will happen. Oftentimes, constraints are modified or cannot be put into place. If we cannot test it because we cannot create an object with an invalid SRID, that is I guess another issue and seems to be OK to remove that test, since it should in theory be impossible to recreate since its enforced at the DB level. But, we can keep the integrity check as is.

PR #21630, opened 2025-08-14
fix: (2.40) Add integrity check to detect geometry columns with incorrect SRID (not 4326) [DHIS2-12547]

zubaira:
backport https://dhis2.atlassian.net/browse/DHIS2-12547


jason-p-pickering:
Looks good, but the test might not work on this version.  Seems to fail in this version but works in higher versions. Sometimes we just leave out the integration tests if they are troublesome to backport.

zubaira:
> Looks good, but the test might not work on this version.  Seems to fail in this version but works in higher versions. Sometimes we just leave out the integration tests if they are troublesome to backport. 

Somehow i made it work :)

PR #21631, opened 2025-08-14
feat: Add trigram index integrity check [DHIS2-19852]

muilpp:
I'm adding an integrity check to identify tracked entity attributes that could benefit from a partial trigram index on the `trackedentityattributevalue` table, but because of an invalid configuration in the TEA, that index won't be created.
The reasons are:
- The TEA is marked as indexable, but blocks the operators `LIKE` and `EW`.
- The TEA doesn't block at least one of the `LIKE` or `EW` operators, but the indexable flag is false.
- The TEA is marked as indexable and allows both operators, but the `minCharactersToSearch` attribute is lower than 3, so the trigram index will never be used.

Since this configuration is likely incorrect, we want to make sure users are aware of what's missing in order to properly create the trigram index.

https://dhis2.atlassian.net/browse/DHIS2-19852

### DIFF
--- a/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/tracked_entity_attributes_invalid_trigram_search_configuration.yaml
+++ b/dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/tracked_entity_attributes_invalid_trigram_search_configuration.yaml
@@ -1,6 +1,6 @@
 ---
 name: tracked_entity_attributes_invalid_trigram_search_configuration
-description: Tracked entity attributes eligible for trigram indexing but misconfigured so the partial trigram index is not created.
+description: Tracked entity attributes having a suboptimal search configuration that potentially results in slower search performance when searching tracked entities using those attributes.
 section: Tracked entity attributes
 section_order: 1
 summary_sql: >-


dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
```diff
@@ -511,6 +511,7 @@ Set<String> addSQLChecksToFlattedReport() {
     checks.add("orgunits_compulsory_group_count");
     checks.add("indicators_violating_exclusive_group_sets");
     checks.add("tracker_geometry_invalid_srid");
+    checks.add("indexable_tracked_entity_attributes_block_trigram_operators");
```
muilpp: Is this necessary? 
I don't see this method being used anywhere

ameenhere: I  think the naming convention of the check itself is to start with the metadata object name. In this case the name could be something like `tracked_entity_attributes_indexable_but_block_trigram_search_operators`

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/indexable_tracked_entity_attributes_block_trigram_operators.yaml
```diff
@@ -0,0 +1,31 @@
+---
+name: indexable_tracked_entity_attributes_block_trigram_operators
+description: Tracked entity attributes that are trigram indexable but restrict the operators required for trigram-based searches.
```
ameenhere: ``
```suggestion
description: Tracked entity attributes that are marked as trigram indexable but block the operators required for trigram-based searches.
```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/indexable_tracked_entity_attributes_block_trigram_operators.yaml
```diff
@@ -0,0 +1,31 @@
...
+    from trackedentityattribute
+    where trigramindexable = true
+    AND (
+    	blockedsearchoperators @> CAST('["LIKE"]' AS jsonb)
+        AND blockedsearchoperators @> CAST('["EW"]' AS jsonb)
+    );
+details_id_type: trackedEntityAttributes
+severity: WARNING
+introduction: >
+    A trigram index can be created on a Tracked Entity Attribute (TEA) if the TEA has the trigram indexable flag set to true and it supports at least one of the operators: LIKE or EW.
```
ameenhere:
```suggestion
    A partial trigram index can be created on the `trackedentityattributevalue` table for a Tracked Entity Attribute (TEA) if the TEA has the trigram indexable flag set to true and it supports at least one of the operators: LIKE or EW.
```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/indexable_tracked_entity_attributes_block_trigram_operators.yaml
```diff
@@ -0,0 +1,31 @@
...
+    That means a TEA with the flag set to true but blocking both operators, will never have a trigram index.
+    The TRACKER_TRIGRAM_INDEX_MAINTENANCE job in the scheduler app is responsible for creating these indexes.
+recommendation:
+    Consider allowing at least one of the LIKE or EW operators on the TEAs below.
```
ameenhere:
```suggestion
    Depending on your needs, we recommend one of the following two actions for the Tracked Entity Attributes listed
    
- Unset the trigramIndexable flag on the TEA. Since trigram based search operators are blocked anyway.

- Unblock either the LIKE or EW operators on the TEA.

```

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/indexable_tracked_entity_attributes_block_trigram_operators.yaml
```diff
@@ -0,0 +1,31 @@
...
+    This will enable the creation of a trigram index on the tracked entity value, improving search performance on tracked entities.
```
ameenhere:
```suggestion
    This will make sure the trigram indexable flag is enabled only for those Tracked Entity Attributes where it is useful to have the partial trigram index.
```

dhis-2/dhis-services/dhis-service-core/src/main/resources/i18n_global.properties
```diff
@@ -2124,6 +2124,7 @@ data_integrity.push_analysis_no_recipients.name=Push analyses without recipients
 data_integrity.file_resources_no_icon.name=File resources that are missing an icon
 data_integrity.users_with_all_authority.name=Users with the ALL authority
 data_integrity.category_options_default_incorrect_sharing.name=Default category option should be publicly shared
+data_integrity.indexable_tracked_entity_attributes_block_trigram_operators.name=Tracked entity attributes that are trigram indexable but restrict the operators required for trigram-based searches.
```
ameenhere:
```suggestion
data_integrity.tracked_entity_attributes_indexable_but_block_trigram_search_operators.name=Tracked entity attributes that are marked as trigram indexable but block trigram based search operators.
```

dhis-2/dhis-services/dhis-service-administration/src/main/java/org/hisp/dhis/dataintegrity/DefaultDataIntegrityService.java
muilpp: since we want to add the three cases in this check, I'll make the name a bit more generic, `tracked_entity_attributes_wrong_trigram_index_configuration`

ameenhere: Maybe `tracked_entity_attributes_invalid_trigram_search_configuration` ?

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/tracked_entity_attributes_invalid_trigram_search_configuration.yaml
```diff
@@ -0,0 +1,71 @@
+---
+name: tracked_entity_attributes_invalid_trigram_search_configuration
+description: Tracked entity attributes eligible for trigram indexing but misconfigured so the partial trigram index is not created.
```
ameenhere:
```suggestion
description: Tracked entity attributes having a suboptimal search configuration that potentially results in slower search performance when searching tracked entities using those attributes.
```

PR #21666, opened 2025-08-19
fix: allow image uploads with valid formats [DHIS2-19733]

zubaira:
https://dhis2.atlassian.net/browse/DHIS2-19733

This PR fixes an issue where image uploads in the Capture app were failing due to overly strict format validation. Valid image formats (e.g. jpeg, png, gif, etc.) were being incorrectly rejected, preventing users from uploading image resources.

**Fix implemented**

- Updated validation logic to check against a single source of truth: Constant.VALID_IMAGE_FORMATS.
- Case-insensitive matching ensures both upper/lowercase variants of valid formats are accepted.
- Invalid formats (exe, dat, pdf, etc.) are still properly rejected.

### DIFF
--- a/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
+++ b/dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
@@ -304,7 +304,7 @@ public class ValidationUtils {
   private static String validateImage(TrackerBundle bundle, String value) {
     FileResource fileResource = bundle.getPreheat().get(FileResource.class, value);
 
-    return Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
+    return !Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
         ? "File resource with uid '"
             + value
             + "' is using invalid image format "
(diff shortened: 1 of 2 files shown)


ameenhere (approved):

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/imports/validation/validator/ValidationUtils.java
```diff
@@ -304,7 +304,7 @@ public static void validateValueType(
   private static String validateImage(TrackerBundle bundle, String value) {
     FileResource fileResource = bundle.getPreheat().get(FileResource.class, value);
 
-    return Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
+    return !Constant.VALID_IMAGE_FORMATS.contains(fileResource.getFormat())
```
enricocolasante, line 307: Doesn't this mean that at the moment in production an `.exe` file would be accepted?

ameenhere: I think this bug is not released yet. Only in master I believe. But @zubaira  can you confirm?

zubaira: @enricocolasante yes thats correct. 
@ameenhere no it does not exist in previous/released versions

PR #21690, opened 2025-08-19
feat: Remove trigram controller and indexableOnly param [DHIS2-19760]

muilpp:
We no longer need the `indexableOnly` parameter in the `TrackedEntityAttributeController`. It was intended for the old trigram index job, which never got to work. After completing [this epic](https://dhis2.atlassian.net/issues/?jql=ORDER%20BY%20created%20DESC&selectedIssue=DHIS2-12243), the job functions without this parameter, so we can safely remove it.

The same applies to the `TrigramSummaryController`, it has never been used, and its information is limited. With the work done on the [integrity checks](https://dhis2.atlassian.net/browse/DHIS2-19852), users can refer to them for more comprehensive insights into the trigram indexes and the TEA configuration. Therefore, it can also be safely removed.


enricocolasante (approved):
Is there any documentation about the parameter and controller that was removed that need to be updated?

muilpp:
> Is there any documentation about the parameter and controller that was removed that need to be updated?

yes, that's done [here](https://github.com/dhis2/dhis2-docs/pull/1635)

ameenhere (approved):

PR #21692, opened 2025-08-19
chore: Merge all separate event migrations into one [DHIS2-19996]

enricocolasante:


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
```diff
@@ -40,67 +45,139 @@ alter table if exists trackerevent
...
+    add constraint fk_singleevent_attributeoptioncomboid foreign key (attributeoptioncomboid) references categoryoptioncombo (categoryoptioncomboid),
+    add constraint fk_singleevent_organisationunitid foreign key (organisationunitid) references organisationunit (organisationunitid),
+    add constraint fk_singleevent_programstageid foreign key (programstageid) references programstage (programstageid);
 
+create sequence if not exists trackerevent_sequence;
+create sequence if not exists singleevent_sequence;
+select setval('trackerevent_sequence', max(eventid)) from trackerevent;
+select setval('singleevent_sequence', max(eventid)) from singleevent;
+
+alter table trackerevent drop column code;
```
ameenhere: Can we drop this (and corresponding in next line) in the original event table before creating a trackerevent and singleevent as copy (before line9)?

I see there is a unique constraint on the code as well with hibernate generated naming, so good to get all those out of the way before the migration of data starts.

ameenhere (commented):
Would it also be possible to find out how much time this single migration file took for a known event table size (5 million? 10 million?) with (for example 3:2 ratio of tracker events to single events) ?

We can then start playing around with variations of this script (for example using EXISTS instead of IN () and so on) to verify which of the variants is better for performance

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
enricocolasante: Done

enricocolasante:
> Would it also be possible to find out how much time this single migration file will take for a known event table size (5 million? 10 million?) with (for example 3:2 ratio of tracker events to single events) ?
> 
> We can then start playing around with variations of this script (for example using EXISTS instead of IN () and so on) to verify which of the variants is better for performance

Changes between `exists` or `in` weren't really relevant in terms of performance.
But we changed the way we create the tables and put the data in as commented in the description

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
```diff
@@ -1,64 +1,125 @@
...
-    where p.type = 'WITH_REGISTRATION'
-    );
+    where p.type = 'WITHOUT_REGISTRATION'
+);
 
-insert into singleevent
-select *
-from event
+-- Delete not used columns
+alter table singleevent drop column scheduleddate;
```
ameenhere, line 16: Are we sure scheduleddate does not exist for singleevent? I know it is impossible in the UI. But from API/backend perspective it might have been possible? 
Do we lose data in that case?
Should we be safe and include it as part of the migration? Interested on hearing your thoughts as well @enricocolasante

ameenhere (approved):
Approving. But something to discuss on the scheduled date of single events which we can finalize before the main PR is merge

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
enricocolasante: We could lose data but we can argue if that was meaningful data at all.
We agreed on the kickoff that single events cannot be scheduled and it was half true in previous versions because the API was not allowing to create a single event with a null `occurredAt`.
So, it was possible through the API to import a single event, even with the `SCHEDULE` status and with a `scheduledAt` value but the `occurredAt` had to be set.
I don't think we should make `scheduleddate` part of the migration, maybe we can think on an integrity check or something similar but we need to balance between the need to not loose data and not overwhelming the user with a lot of decisions over data that are not important for the system.

PR #21732, opened 2025-08-20
fix: make overridden bundled apps appear as bundled

netroms:
## Summary
Overridden bundled apps was not showing in the bundled app listing endpoint, because the app's bundled flag was not set to true.

### Problem
When apps that override the bundled apps, with for example a newer version, they need to be recognized as a bundled one during install and when retrieving them.

### Fix
Check for each app if it's key exists in the list of bundled apps, during retrieval and listing.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AppControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/AppControllerTest.java
@@ -31,9 +31,7 @@ package org.hisp.dhis.webapi.controller;
 
 import static java.nio.file.Files.createTempDirectory;
 import static org.hisp.dhis.util.ZipFileUtils.MAX_ENTRIES;
-import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/BundledAppManager.java
```diff
@@ -184,4 +184,18 @@ public static InputStream getAppBundleInfoInputStream() {
       throw new RuntimeException(e);
     }
   }
+
+  public boolean isBundledApp(@Nonnull App app) {
```
david-mackessy, line 188: Is this different to `App#isBundled`?
If so, does it mean we have different criteria of what bundled means in 2 different places?

netroms: App#isBundled is a property set on the app instances, but it's a transient property, i.e. it's never persisted on disk together with the manifest and all the other files. This means that every time we read the apps from disk into cache, we need to look up the app's key and check if it's in the static (built into the .war) list of bundled apps. This isBundledApp() does that, and set the app's bundled property, before it's added to the cache.

PR #21742, opened 2025-08-22
feat: add a description field to the usergroup

netroms:


david-mackessy (approved):
I'm not sure if we have any conventions for the names of these migrations, but a lot of migrations for adding columns to a table look like `Add_colName_to_tableName`

vietnguyen (approved):

larshelge (approved):

PR #21747, opened 2025-08-22
feat: Add trigramIndexed field on TEA [DHIS2-20014]

muilpp:
I'm adding a new field `trigramIndexed` to the `TrackedEntityAttribute` class to indicate whether a partial trigram index exists on the `trackedentityattributevalue` table for that TEA.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/TrackedEntityAttributeControllerTest.java
@@ -150,10 +150,23 @@ class TrackedEntityAttributeControllerTest extends PostgresControllerIntegration
   }
 
   @Test
-  void shouldReturnCollectionOfAttributesWithExpectedTrigramIndexField() {
+  void shouldReturnCollectionOfAttributesWithExpectedTrigramIndexFieldWhenFetchingAllFields() {
     createTrigramIndexes(Set.of(teaA, teaB));
 
     JsonObject json = GET("/trackedEntityAttributes?fields=*").content(HttpStatus.OK);
+
+    assertAttributeTrigramIndexedField(
+        json,
+        Map.of(
+            teaA.getUid(), true, teaB.getUid(), true, teaC.getUid(), false, teaD.getUid(), false));
+  }
+
+  @Test
+  void shouldReturnCollectionOfAttributesWithExpectedTrigramIndexFieldWhenFetchingTrigramField() {
+    createTrigramIndexes(Set.of(teaA, teaB));
+
+    JsonObject json =
+        GET("/trackedEntityAttributes?fields=id,trigramIndexed").content(HttpStatus.OK);
 
     assertAttributeTrigramIndexedField(
         json,
@@ -168,7 +181,8 @@ class TrackedEntityAttributeControllerTest extends PostgresControllerIntegration
     JsonObject jsonA =
         GET("/trackedEntityAttributes/" + teaA.getUid() + "?fields=*").content(HttpStatus.OK);
     JsonObject jsonB =
-        GET("/trackedEntityAttributes/" + teaB.getUid() + "?fields=*").content(HttpStatus.OK);
+        GET("/trackedEntityAttributes/" + teaB.getUid() + "?fields=id,trigramIndexed")
+            .content(HttpStatus.OK);
 
     assertEquals(teaA.getUid(), jsonA.getString("id").string());
     assertEquals(teaA.getTrigramIndexed(), jsonA.getBoolean("trigramIndexed").booleanValue());


dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
```diff
@@ -143,6 +143,15 @@ Set<TrackedEntityAttribute> getAllUserReadableTrackedEntityAttributes(
    */
   Set<TrackedEntityAttribute> getAllTrigramIndexableTrackedEntityAttributes();
 
+  /**
+   * Returns all {@link TrackedEntityAttribute} UIDs that have a partial trigram index on the
+   * trackedentityattributevalue table.
+   *
+   * @return a set of TrackedEntityAttribute UIDs, or an empty List if there are no
```
enricocolasante:
```suggestion
   * @return a set of TrackedEntityAttribute UIDs, or an empty set if there are no
```

dhis-2/dhis-api/src/main/java/org/hisp/dhis/trackedentity/TrackedEntityAttributeService.java
```diff
@@ -143,6 +143,15 @@ Set<TrackedEntityAttribute> getAllUserReadableTrackedEntityAttributes(
...
+   *     TrackedEntityAttributes that are indexed
+   */
+  Set<String> getAllTrigramIndexedTrackedEntityAttributes();
```
enricocolasante: Would it be possible to return a set of `UID`s?
It would make it clear that those string are uids without the need to check that in the java doc

dhis-2/dhis-services/dhis-service-administration/src/main/resources/data-integrity-checks/tracked_entity_attributes/tracked_entity_attributes_trigram_index_overview.yaml
```diff
@@ -32,10 +32,10 @@ details_sql: >-
...
-    The `TRACKER_TRIGRAM_INDEX_MAINTENANCE` job in the scheduler app is responsible for creating these indexes.
-    This check reports the tracked entity attributes whose related `trackedentityattributevalue` table entry has a partial trigram index.
+    A partial trigram index can be created on the "trackedentityattributevalue" table for a Tracked Entity Attribute (TEA) if the TEA has the trigram indexable flag set to true and supports at least one of the operators: "LIKE" or "EW".
+    The "TRACKER_TRIGRAM_INDEX_MAINTENANCE" job in the scheduler app is responsible for creating these indexes.
+    This check reports the tracked entity attributes whose related "trackedentityattributevalue" table entry has a partial trigram index.
 recommendation:
     This check is informational only, no action is required.
-    Below is the lists of tracked entity attributes that currently have a partial trigram index on the `trackedentityattributevalue` table.
+    Below is the lists of tracked entity attributes that currently have a partial trigram index on the "trackedentityattributevalue" table.
```
ameenhere:
```suggestion
    Below is the list of tracked entity attributes that currently have a partial trigram index on the "trackedentityattributevalue" table.
```

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -186,4 +197,51 @@ private TrackedEntityAttribute getTrackedEntityAttribute(String id) throws WebMe
 
     return trackedEntityAttribute;
   }
+
+  @Override
+  public ResponseEntity<StreamingJsonRoot<TrackedEntityAttribute>> getObjectList(
```
ameenhere: Instead of overriding `getObectList` , we can use the postProcessResponseEntities hook from the `AbstractFullReadOnlyController` instead I feel. Can you check it out?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -186,4 +197,51 @@ private TrackedEntityAttribute getTrackedEntityAttribute(String id) throws WebMe
...
+          .getObjects()
+          .forEach(
+              tea -> tea.setTrigramIndexed(indexedAttributeUids.contains(UID.of(tea.getUid()))));
+    }
+
+    return response;
+  }
+
+  @Override
+  public ResponseEntity<?> getObject(
```
ameenhere: Instead of overriding `getObect` , we can use the `postProcessResponseEntity` hook from the `AbstractFullReadOnlyController` instead I feel. Can you check it out?

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/event/TrackedEntityAttributeController.java
```diff
@@ -186,4 +197,51 @@ private TrackedEntityAttribute getTrackedEntityAttribute(String id) throws WebMe
...
+
+  @Override
+  public ResponseEntity<StreamingJsonRoot<TrackedEntityAttribute>> getObjectList(
+      GetObjectListParams params, HttpServletResponse res, UserDetails currentUser)
+      throws ForbiddenException, BadRequestException, ConflictException {
+
+    ResponseEntity<StreamingJsonRoot<TrackedEntityAttribute>> response =
+        super.getObjectList(params, res, currentUser);
+    if (response.getBody() != null) {
+      Set<UID> indexedAttributeUids =
```
david-mackessy, line 197: A standard call to `GET /api/trackedEntityAttributes` without any filters or fields query params will result in a response that looks something like this:
```json
{"trackedEntityAttributes": [
        {
            "displayName": "Address",
            "id": "VqEFza8wbwA"
        },
        {
            "displayName": "Age",
            "id": "spFvx9FndA4"
        },
        ....
     ]
}
```
Only the default response fields are present. So for the get all endpoint, unless `trigramIndexed` has been requested to be included in the response, this extra processing to set the value for this field may not be required.
I think you can just check in the params object to see what fields have been requested. I'm not sure if there are any convenience methods to use (taking into consideration requesting `*` or `all` fields. @jbee might be able to shed some light on this.
Same goes for the single endpoint, except all fields are returned by default, so checking that the field hasn't been excluded before extra processing sounds like a valid thing to do.

ameenhere (approved):

PR #21758, opened 2025-08-25
feat: Add validation when importing CategoryOptionCombos [DHIS2-19944]

david-mackessy:
# Background
We currently allow importing `CategoryOptionCombo`s through the `/api/metadata` endpoint with no validation. It is possible to create a `CategoryOptionCombo` with any `CategoryCombo` and any `CategoryOption`. This is bad and completely contradicts how the system is setup to behave. This new validation will prevent this.

It is understood that the majority of implementations allow the system to auto generate `CategoryOptionCombos` (never importing `CategoryOptionCombos`). In those cases these validations will never apply.

`CategoryOptionCombo` state is deterministic and they are always derived from the `CategoryCombo` -> `Categories` -> `CategoryOptions` model.

# Changes
## Metadata endpoint
When importing `CategoryOptionCombo`s through `/api/metadata` the following validations now occur:
1.  Does the number of imported `CategoryOptionCombo`s match the expected number of `CategoryOptionCombo`s
2.  Do the imported `CategoryOptionCombo`s match the expected state for those `CategoryOptionCombo`s:
    - expected `CategoryCombo`
    - expected `CategoryOption`s
 3.  Are the imported `CategoryOptionCombo`s seen as duplicates

If any of the validations fail, the appropriate new ErrorCodes are returned in the import summary:
1. `E1130`: `Importing {0} CategoryOptionCombos does not match the expected amount of {1} for CategoryCombo {2}`
2. `E1131`: `Unexpected CategoryOptionCombo provided with CategoryOptions: {0} for CategoryCombo: {1}. Missing expected CategoryOptionCombos with CategoryOption sets: {2}`
3. `E1132`: `Provided CategoryOptionCombo {0} cannot be processed (potential duplicate). An existing CategoryOptionCombo {1} has the same CategoryCombo {2} and same CategoryOptions {3}`

## CategoryOptionCombo endpoint
When posting a new `CategoryOptionCombo` through `/api/categoryOptionCombos` the request is rejected with the error message:
- `E1129`:`Creating a single CategoryOptionCombo is forbidden through this endpoint. CategoryOptionCombos should be auto generated or imported through the metadata import`

# Testing
Automated integration tests added the following scenarios:
- Creates
  - Importing expected CategoryOptionCombos should succeed
  - Importing expected number, but with duplicate CategoryOptionCombo, should fail
  - Importing fewer CategoryOptionCombos than expected should fail
  - Importing more CategoryOptionCombos than expected should fail
- Updates
  - Partial import (COCs only) with expected CategoryOptionCombos should succeed
  - Importing (full cat model) expected CategoryOptionCombos should succeed
  - Importing expected number, but invalid CategoryOptionCombos, should fail
  - Importing existing CategoryOptionCombos with new UIDs (duplicates) should fail
  - Importing fewer CategoryOptionCombos than expected should fail
  - partial importing fewer CategoryOptionCombos than expected should fail
  - Importing more CategoryOptionCombos than expected should fail

Had to clean up/fix tracker metadata test data so that metadata imports conform to new validation rules.

# Accepted risks
Now that `CategoryOptionCombo`s can only be imported as a full expected set, this might cause issues with metadata sync if a version snapshot diff contains a subset of `CategoryOptionCombo`s, which then cannot be imported.
This scenario is seen as an edge case and is likely not going to occur in the majority of use cases.
This scenario has been manually tested, the sync still succeeds but the new `CategoryOptionCombo`s are not imported. Manual steps using the API can get to the expected state:
- get full `CategoryOptionCombo` set for `CategoryCombo` from central instance `GET /api/categoryOptionCombos?fields=*&filter=categoryCombo.id:in:[{id}]`
- import that payload to the remote instance `POST /api/metadata`

### DIFF
--- a/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
+++ b/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
@@ -70,7 +70,7 @@ public class CategoryOptionComboObjectBundleHook
 
     checkNonStandardDefaultCatOptionCombo(combo, addReports);
     checkIsValid(combo, addReports);
-    checkIsExpectedState(combo, bundle, addReports);
+    checkFullCatOptComboSetProvided(combo, bundle, addReports);
   }
 
   /**
@@ -95,7 +95,7 @@ public class CategoryOptionComboObjectBundleHook
    * @param bundle bundle
    * @param addReports reports to add errors to
    */
-  private void checkIsExpectedState(
+  private void checkFullCatOptComboSetProvided(
       CategoryOptionCombo optionCombo, ObjectBundle bundle, Consumer<ErrorReport> addReports) {
     // is it a create or update
     boolean cocIsPersisted = bundle.isPersisted(optionCombo);


dhis-2/dhis-services/dhis-service-core/src/main/java/org/hisp/dhis/category/hibernate/HibernateCategoryOptionComboStore.java
```diff
@@ -163,4 +165,29 @@ public List<CategoryOptionCombo> getCategoryOptionCombosByCategoryOption(
...
+                """
+                select coc.uid from categoryoptioncombo coc
+                join categorycombos_optioncombos cc_coc on cc_coc.categoryoptioncomboid = coc.categoryoptioncomboid
+                join categoryoptioncombos_categoryoptions coc_co on coc_co.categoryoptioncomboid = coc.categoryoptioncomboid
+                join categorycombo cc on cc.categorycomboid = cc_coc.categorycomboid
+                join categoryoption co on co.categoryoptionid = coc_co.categoryoptionid
+                where cc.uid = :ccUid
+                and co.uid in (:coUids)
+                group by coc.categoryoptioncomboid
+                having count(co.uid) = :size
```
david-mackessy, line 184: As a very rough comparison, the cost and execution time for this query was checked against the SL DB and it is very similar with the similar query used in data value import
- this query `(cost=20.81..20.83 rows=1 width=20) (actual time=0.149..0.153 rows=1 loops=1)`  
- DV squery `(cost=26.73..26.75 rows=1 width=8) (actual time=0.141..0.144 rows=1 loops=1)`

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EventImportValidationTest.java
```diff
@@ -193,11 +193,7 @@ void testCantWriteAccessCatCombo() throws IOException {
     importReport = trackerImportService.importTracker(params, trackerObjects);
 
     assertHasOnlyErrors(
-        importReport,
-        ValidationCode.E1096,
-        ValidationCode.E1099,
```
david-mackessy, line 198: This error is not present anymore due to the cleanup in tracker test metadata. The cleanup removed and fixed invalid category model imports.

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_basic_metadata.json
```diff
@@ -2835,26 +2776,6 @@
...
-      "created": "2019-03-25T13:40:24.722",
-      "translations": [],
-      "sharing": {
-        "public": "rwrw----",
-        "external": false,
-        "owner": null,
-        "users": {},
-        "userGroups": {}
-      }
-    },
```
david-mackessy, line 2857: removed 4 CategoryOptions that were being used incorrectly with regards to a correct category model being generated (CategoryCombo -> Categories -> CategoryOptions -> CategoryOptionCombos)

dhis-2/dhis-test-integration/src/test/resources/tracker/tracker_basic_metadata.json
```diff
@@ -3119,50 +3021,6 @@
...
-      ],
-      "categoryCombo": {
-        "id": "DYUZwonFF9u"
-      },
-      "translations": [],
-      "startDate": "2019-12-01T00:00:00.000",
-      "lastUpdated": "2019-03-25T13:40:24.779",
-      "ignoreApproval": false,
-      "attributeValues": [],
-      "created": "2019-03-25T13:40:24.775"
```
david-mackessy, line 3165: removed 3 CategoryOptionCombos that were being used incorrectly with regards to a correct category model being generated (CategoryCombo -> Categories -> CategoryOptions -> CategoryOptionCombos)

david-mackessy:
Added @dhis2/tracker-backend as 1 tracker test was updated to reflect a valid category model.
see files:
- dhis-test-integration/src/test/java/org/hisp/dhis/tracker/imports/validation/EventImportValidationTest.java
- dhis-test-integration/src/test/resources/tracker/tracker_basic_metadata.json
- dhis-test-integration/src/test/resources/tracker/validations/events-aoc-and-co-dont-match.json

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/collection/CollectionUtils.java
```diff
@@ -425,4 +426,29 @@ public static <K, V> Map<K, V> merge(@Nonnull Map<K, V> m1, @Nonnull Map<K, V> m
...
+  /**
+   * Combines 2 Lists into a memory-efficient unmodifiable view of both. Safe to perform read/get
+   * operations on. Updates on original Lists are reflected in the returned view. Can't be modified.
+   *
+   * @param l1 list 1
+   * @param l2 list 2
+   * @return combined view of both lists
+   * @param <T> type
+   */
+  public static <T> List<T> combinedUnmodifiableView(List<T> l1, List<T> l2) {
```
jbee: This seems not worth the trouble. There is so much more somebody might do with the result that is not correct yet. For exmaple, `contains`, `iterator`, `isEmpty`, ...
I know it is frustrating that the JDK does not have a concat view already. I tend to use `Stream.concat(a.stream(), b.stream()).toList()`. Sure that copies to a new list but for short lists this isn't a big deal. 
Alternatively be very clear about the limitations of this view in the javadoc.

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
```diff
@@ -29,30 +29,258 @@
...
+   *       CategoryOption} set).
+   * </ul>
+   *
+   * <p>The validation uses {@link UID} matching.
+   *
+   * @param optionCombo option combo being validated
+   * @param bundle bundle
+   * @param addReports reports to add errors to
+   */
+  private void checkIsExpectedState(
```
jbee: Maybe the naming can reflect more what that expected state is?

jbee (approved):
Overall this looks good so I approve already even if minor tweaks might be a good idea.

dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/collection/CollectionUtils.java
david-mackessy: yep, I'll remove it 👍

dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/metadata/objectbundle/hooks/CategoryOptionComboObjectBundleHook.java
david-mackessy: `checkIsExpectedGeneratedState` ?
`checkHasCorrectCatComboCatOptionSet` ?
`checkFullCatOptComboSetProvided` ?

vietnguyen: isModifyingCOC
isMutatingCOC

david-mackessy: > isModifyingCOC
isMutatingCOC  

it's not about checking if a COC is being modified. The javadoc says what's being checked. It checks multiple things. I'll go with `checkFullCatOptComboSetProvided`

PR #21761, opened 2025-08-26
perf: improve tracker field filtering performance DHIS2-19910

teleivo:
## Background

[field filtering](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_metadata_field_filter) allows reducing the JSON response via the query parameter `fields`. The JSON response can also be changed using these [transformations](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_field_transformers). Field filtering is used by almost every endpoint in DHIS2. Tracker uses the same implementation as platform which was [re-implemented in 2.38](https://dhis2.atlassian.net/browse/TECH-650) using Jackson. It looks to me as if analytics still depends on the [previous field filter implementation](https://github.com/dhis2/dhis2-core/tree/master/dhis-2/dhis-services/dhis-service-node/src/main/java/org/hisp/dhis/fieldfilter) at least in some endpoint(s).

I ran a [Gatling](https://github.com/dhis2/performance-tests-gatling) performance tests locally with a single user making 100 requests against `/api/organisationUnits?fields=:all,!name,!id,!favorites,!translations,!children,!sharing` sequentially. I collected the following profile

[profile.html](https://github.com/user-attachments/files/22094934/profile.html)

If you select a thread processing an HTTP request and search (Ctrl+f) for `fieldfiltering` you'll find that 90% of CPU samples are dealing with it. This gets worse quickly if you increase the `pageSize` and thus the elements to be filtered. Field filtering outways time spent in the DB even though hibernate is making lots of requests to the DB (at least one per entity)!

To get a feel for the cost of field filtering I ran the above test against

1. dhis2/core:42.0 with no code changes
2. dhis2/core-dev:42.0 where I removed the field filtering feature, so no matter `fields` we serialize the entire object using Jackson

For our default `pageSize=50` we already pay ~20ms on field filtering even though 2. is at a disadvantage as it has to serialize more than 1. as there is no filtering.

analysis

The main reasons why the current implementation is slow are

1. The implementation is built around the full path of a field:

[FieldFilterService](https://github.com/dhis2/dhis2-core/blob/2151f37440c411b844ee59aa6c6a1420012f6dcb/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java#L250) turns `fields=:all,!name,!id,!favorites,!translations,!children,!sharing` into

```
paths = {ArrayList@31490}  size = 71
 0 = {FieldPath@31270} "FieldPath(name=parent, path=[], exclude=false, preset=false, transformers=[], fullPath=parent)"
 1 = {FieldPath@31415} "FieldPath(name=yearToDate, path=[queryMods], exclude=false, preset=false, transformers=[], fullPath=queryMods.yearToDate)"
 2 = {FieldPath@31272} "FieldPath(name=type, path=[], exclude=false, preset=false, transformers=[], fullPath=type)"
 3 = {FieldPath@31483} "FieldPath(name=minDate, path=[queryMods], exclude=false, preset=false, transformers=[], fullPath=queryMods.minDate)"
...
```

Our Jackson [FieldFilterSimpleBeanPropertyFilter](https://github.com/dhis2/dhis2-core/blob/a7c9affd49bbb02db3b48c2597e377f098bfb779/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterSimpleBeanPropertyFilter.java#L118-L150) is then called for every field on an object to decide whether it should be serialized. [FieldFilterSimpleBeanPropertyFilter](https://github.com/dhis2/dhis2-core/blob/a7c9affd49bbb02db3b48c2597e377f098bfb779/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterSimpleBeanPropertyFilter.java#L118-L150) builds the full path again and again for every field from Jacksons generator state. This involves lots of String splits, joins, StringBuilder operations like `insert(0` 😨.

To answer the question `should the field be include it or not?` we then iterate over all field paths (which can really be all or most of them with `fields=*`) and [String compare paths](https://github.com/dhis2/dhis2-core/blob/a7c9affd49bbb02db3b48c2597e377f098bfb779/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterSimpleBeanPropertyFilter.java#L109-L110).

2. We are using a post-processing approach:

We serialize the [Jackson filtered object to a Jackson JsonNode](https://github.com/dhis2/dhis2-core/blob/4b34f59a62e1ebb2c28d1f4e9fee3497587d093c/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java#L271) so we can then apply [transformations](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_field_transformers)

https://github.com/FasterXML/jackson-databind/blob/d4220ed27198847e13ce75d9bee805e028a100e0/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L3639
> Functionally similar to serializing value into token stream and parsing that stream back as tree model node, but more efficient as {@link TokenBuffer} is used to contain the intermediate representation instead of fully serialized contents.

The price of field transformations is also paid when users don't actually transform any fields. Remember [valueToTree](https://github.com/FasterXML/jackson-databind/blob/d4220ed27198847e13ce75d9bee805e028a100e0/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L3636-L3642) described above.

This is the flow of data until it reaches the HTTP body

`DHIS2 entity => Jackson property filter & serialize to a Jackson ObjectNode => serialized again in controllers/message converter to HTTP body`

So even though you find the words `stream` in some of the involved code we do not actually stream the entities to the HTTP body directly. We will create JSON in-memory representations first like in [StreamingJsonRoot.pager](https://github.com/dhis2/dhis2-core/blob/4b34f59a62e1ebb2c28d1f4e9fee3497587d093c/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/webdomain/StreamingJsonRoot.java#L47-L48) which is then serialized in [StreamingJsonRootMessageConverter](https://github.com/dhis2/dhis2-core/blob/4b34f59a62e1ebb2c28d1f4e9fee3497587d093c/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/mvc/messageconverter/StreamingJsonRootMessageConverter.java#L62) when it comes to metadata. Endpoints returning a Jackson `ObjectNode` or any other type rely on Spring serializing to the HTTP body using the `ObjectMapper`.

3. (maybe minor) Not following [Jacksons advice](https://github.com/fasterxml/jackson-docs/wiki/presentation:-jackson-performance#basics-things-you-should-do-anyway) as we copy the heavy thread safe [ObjectMapper on every request](https://github.com/dhis2/dhis2-core/blob/4b34f59a62e1ebb2c28d1f4e9fee3497587d093c/dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/FieldFilterService.java#L258).

It is important we optimize this
* as it affects almost every endpoint
* even when users don't send us a `fields` parameter a default will be in place like this one in `/tracker/trackedEntities` [fields=*,!relationships,!enrollments,!events,!programOwners](https://github.com/dhis2/dhis2-core/blob/4b34f59a62e1ebb2c28d1f4e9fee3497587d093c/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/trackedentity/TrackedEntityRequestParams.java#L68)
* you could argue users rarely use a `pageSize` larger than the default of `50`. It still matters as we have deeply nested entities like tracked entity that can have lots of collections of nested entities which are not affected by the `pageSize`
* our apps like Capture make many API requests even in a single page. Try clicking your way to an event dashboard by selecting an orgUnit and program and you'll find ~80 API requests (the exact number of API requests is lower due to some being for static content or not field filtered)

## Improved implementation

This PR shows how we can improve what was [re-implemented in 2.38](https://dhis2.atlassian.net/browse/TECH-650) using Jackson. The essence is that for every field Jackson needs to know `should I include it or not?`. This is essentially a predicate `boolean test(String field)` which must be answered quickly.

How?

1. We build up a structure of predicates (`Fields`) that reflect the structure of the data being filtered. [in `FieldsParser`]
2. We use the fact that Jackson will traverse the object tree in depth-first order. [in `FieldsPropertyFilter`]
3. We store the current objects predicate tree in Jacksons context and reset it once Jackson is done with serializing that object. [in `FieldsPropertyFilter`]
6. We do still use the current approach of building up the intermediate `Jackson ObjectNode` but only for fields that are actually transformed. [in `FieldsPropertyFilter`]

Transformations (4.) could be further improved. Some transformations like `rename` can be done without creating intermediate `Jackson ObjectNode`. Others like `isEmpty` are more challenging as they need JSON type information which we do not have in the property filter unless we actually serialize. I did not go further with optimizing 4. as I am not sure many people even use transformations.

Benefits of new implementation

* faster response times
* less garbage as without transformation we truly stream our Java entities to the HTTP body

## Benchmark

Benchmark `FieldFilterSerializationBenchmarkTest` compares the throughput of the current and the proposed implementation. It also compares the presence of field filtering when no filtering occurs i.e. `fields=*` with pure Jackson serialization. The **operation** being benchmarked is the parsing of `fields` and the serialization of `List<Event>` to a JSON `String` which includes filtering and transforming events.

JMH advice is to fork the JVM to prevent interference between the runs. I have set `fork(0)` because of Spring/authentication. It took me a while to get Spring and JMH to work. So don't put too much faith into the exact numbers (I also ran this locally). Since our workload is static, same POJOs are serialized, I think that should be alright. The difference in magnitude should also be enough indication. I also ran a performance test locally against the application so we can more clearly see the impact on the response times.

### `fields=*` (overhead analysis)

We can calculate an overhead of the `field filter` feature/implementation by comparing the field filter implementations throughput with `fields=*` to pure Jackson serialization.

benchmark_graph_star

These are the throughput values and the overhead i.e. `Overhead = (noFieldFiltering - xxxFieldFiltering) / noFieldFiltering`. The event count is the number of events serialized in the benchmarked operation.

|   Event Count | No Filtering (events/s)   | Better Filtering (events/s)   | Current Filtering (events/s)   | Better Overhead (%)   | Current Overhead (%)   | Better vs Current (x faster)   |
|--------------:|:--------------------------|:------------------------------|:-------------------------------|:----------------------|:-----------------------|:-------------------------------|
|            25 | 104,891                   | 63,939                        | 1,446                          | 39.0%                 | 98.6%                  | 44.2x                          |
|            50 | 84,331                    | 57,366                        | 1,504                          | 32.0%                 | 98.2%                  | 38.1x                          |
|           100 | 88,863                    | 53,817                        | 1,597                          | 39.4%                 | 98.2%                  | 33.7x                          |
|           200 | 91,592                    | 51,846                        | 1,637                          | 43.4%                 | 98.2%                  | 31.7x                          |
|           400 | 88,885                    | 55,407                        | 1,599                          | 37.7%                 | 98.2%                  | 34.6x                          |
|           800 | 91,857                    | 53,514                        | 1,694                          | 41.7%                 | 98.2%                  | 31.6x                          |

#### Allocations and GC

I ran the `fields=*` and `eventCount=50` benchmark from before with the async-profiler profiling allocations. In this run the proposed implementation processed 31x more events/s than the current implementation.

|   Event Count | No Filtering (events/s)   | Better Filtering (events/s)   | Current Filtering (events/s)   | Better Overhead (%)   | Current Overhead (%)   | Better vs Current (x faster)   |
|--------------:|:--------------------------|:------------------------------|:-------------------------------|:----------------------|:-----------------------|:-------------------------------|
|            50 | 71,569                    | 40,885                        | 1,333                          | 42.9%                 | 98.1%                  | 30.7x                          |

This was the allocation behavior:

  | Event Type                      | No Filtering   | Better         | Current       |
  |---------------------------------|----------------|----------------|---------------|
  | jdk.ObjectAllocationInNewTLAB   | 29,468 (592KB) | 16,756 (333KB) | 6,309 (120KB) |
  | jdk.ObjectAllocationOutsideTLAB | 0              | 0              | 0             |
  | jdk.GCHeapSummary               | 24 events      | 14 events      | 6 events      |

if we now compare the allocations per event

| Implementation | Total Allocs | Events/s | Duration | Total Events | Allocs/Event |
  |----------------|--------------|----------|----------|--------------|--------------|
  | Current        | 6,309        | 1,333    | 25s      | 33,325       | 0.189        |
  | Better         | 16,756       | 40,885   | 25s      | 1,022,125    | 0.016        |
  | No Filtering   | 29,468       | 71,569   | 25s      | 1,789,225    | 0.016        |

- Better implementation: 12x more allocation-efficient per event (0.016 vs 0.189)
- Better optimization: Achieves baseline allocation efficiency (same as no filtering) while maintaining filtering functionality
- The current implementation has less GC events which I would explain by it processing considerably less events

Allocation flamegraphs created using `jfrconv --alloc --total jfr-alloc.jfr alloc-flamegraph.html`:

- no field filtering [alloc-flamegraph.html](https://github.com/user-attachments/files/22166981/alloc-flamegraph.html)
- better field filtering [alloc-flamegraph.html](https://github.com/user-attachments/files/22166995/alloc-flamegraph.html)
- current field filtering [alloc-flamegraph.html](https://github.com/user-attachments/files/22167000/alloc-flamegraph.html)

You can see the current field filtering does lots of `String` manipulations. The better implementation shows allocations due to Jackson and date serialization which we can optimize further later on.

### `fields=*,!relationships`

|   Event Count | Better Filtering (events/s)   | Current Filtering (events/s)   | Performance Improvement   |
|--------------:|:------------------------------|:-------------------------------|:--------------------------|
|            25 | 188,110                       | 14,115                         | 13.3x                     |
|            50 | 185,000                       | 21,462                         | 8.6x                      |
|           100 | 172,609                       | 26,726                         | 6.5x                      |
|           200 | 165,711                       | 31,444                         | 5.3x                      |
|           400 | 171,590                       | 34,400                         | 5.0x                      |
|           800 | 165,264                       | 35,154                         | 4.7x                      |

### `fields=*,event~rename(foo)` (transformation)

From the results of serializing all fields with one renamed we can see

* The current implementation has similar throughput to `fields=*` as it always creates the intermediate `Jackson ObjectNode` in memory.
* You can see the transformation does affect the proposed implementation as it uses the same approach and creates an intermediate `Jackson ObjectNode` in memory. It still outperforms the current implementation as it will only create an `ObjectNode` for the node it actually transforms.

|   Event Count | Better Filtering (events/s)   | Current Filtering (events/s)   | Performance Improvement   |
|--------------:|:------------------------------|:-------------------------------|:--------------------------|
|            25 | 56,172                        | 1,461                          | 38.4x                     |
|            50 | 50,329                        | 1,521                          | 33.1x                     |
|           100 | 53,232                        | 1,657                          | 32.1x                     |
|           200 | 50,400                        | 1,584                          | 31.8x                     |
|           400 | 53,804                        | 1,588                          | 33.9x                     |
|           800 | 52,718                        | 1,659                          | 31.8x                     |

### `fields=event`

Filtering out all but a single string field with `fields=event` leads to

|   Event Count | Better Filtering (events/s)   | Current Filtering (events/s)   | Performance Improvement   |
|--------------:|:------------------------------|:-------------------------------|:--------------------------|
|            25 | 2,146,797                     | 95,280                         | 22.5x                     |
|            50 | 2,220,312                     | 162,887                        | 13.6x                     |
|           100 | 2,356,845                     | 248,318                        | 9.5x                      |
|           200 | 2,447,940                     | 343,439                        | 7.1x                      |
|           400 | 2,274,336                     | 416,872                        | 5.5x                      |
|           800 | 2,479,009                     | 445,075                        | 5.6x                      |

## HTTP API performance

I ran a [Gatling](https://github.com/dhis2/performance-tests-gatling) performance tests locally with a single user making 100 sequential HTTP requests against

1. dhis2/core-dev:42-47f0d4eae9 with no code changes
2. this PRs changes with the proposed implementation

I picked a slow request the Capture app makes as the target. Select a [Child Programme and orgUnit Ngelehun CHC](https://play.im.dhis2.org/stable-2-42-1/apps/capture#/?orgUnitId=DiszpKrYNg8&programId=IpHINAT79UW&selectedTemplateId=IpHINAT79UW-default). Capture app then makes request `https://play.im.dhis2.org/stable-2-42-1/api/42/tracker/trackedEntities?order=createdAt:desc&page=1&pageSize=15&orgUnits=DiszpKrYNg8&orgUnitMode=SELECTED&program=IpHINAT79UW&fields=:all,!relationships,programOwners[orgUnit,program]`.

The default `pageSize` on the Capture app page above is `15` with options `10, 15, 25, 50, 100`. This graph shows the difference in response times of `current - proposed`

analysis

You can see from the CPU profile that field filtering does not play a big role anymore. Once you find (Ctrl+f) the thread handling `TrackedEntitiesExportController.getTrackedEntities`, you'll see that

1. [profile.html before](https://github.com/user-attachments/files/22119668/profile.html):
 ~42% of stacktrace samples are related to `fieldfiltering`
2. [profile.html after](https://github.com/user-attachments/files/22119739/profile.html): ~0.4% of stacktrace samples are related to `fieldfiltering`

### Caveat

The drastic improvement we see with the tracker API above will unlikely be seen with metadata. The reason is that tracker was calling the current field filter service for ever item individually

```java
  public <T> ResponseEntity<Page<ObjectNode>> serve(
      HttpServletRequest request,
      String key,
      org.hisp.dhis.tracker.Page<T> page,
      FieldsRequestParam fieldParams) {
    return ResponseEntity.ok()
        .contentType(MediaType.APPLICATION_JSON)
        .body(
            Page.withPager(
                key,
                page.withMappedItems(
                    i -> fieldFilterService.toObjectNode(i, fieldParams.getFields())),
                getRequestURL(request)));
```

Tracker now field filters the entire list of items at once in the `FilteredPageHttpMessageConverter`.

## Tests

* The proposed parser is unit tested using the same `fields` input strings used for the current implementation while asserting both `fields` outputs match.
* The JSON output of proposed fields filtering vs current implementation is also tested in `FieldFilterSerializationTest`. This tests the parsing, Jackson config and actual field filtering using the Jackson property filter.
* Tracker also has e2e tests and API tests (to which I added transformations). This ensures Spring/Jackson is properly configured.

## Misc changes

* Tracker controller's must be annotated with `@OpenApi.EntityType` and the actual type that is returned/filtered. This type is used to fetch the schema needed to provide preset `fields=:simple`. Double-checked with Jan that this is ok from the OpenAPI point of view 👌🏻 
* Added Spring converter `FilteredPageHttpMessageConverter` (similar to `StreamingJsonRootMessageConverter`) which does the filtering and serialization by streaming to the HTTP body. This is a dedicated converter for Trackers `Page`.

## Jackson and Thread Safety

The proposed field filtering implementation is **thread-safe**. `FieldsConfig` creates a singleton `ObjectMapper` bean. The `FilteredPageHttpMessageConverter` creates a new `ObjectWriter` with `withAttribute()` for each HTTP request, and Jackson's architecture ensures isolation between concurrent serializations.

**Why it's thread-safe:**
  * Singleton [ObjectMapper](https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/latest/com/fasterxml/jackson/databind/ObjectMapper.html) (`jsonFilterMapper` bean) is thread-safe and [supposed to be reused](https://github.com/fasterxml/jackson-docs/wiki/presentation:-jackson-performance#basics-things-you-should-do-anyway).
  * Each request gets its own `ObjectWriter` instance via `filterMapper.writer().withAttribute()`
  * Jackson creates isolated `SerializerProvider` instances per serialization call
  * The `FieldsPropertyFilter.serializeAsField()` attribute modifications operate on per-call instances
  * No shared mutable state exists between concurrent requests

## Bugs/Behavior changes

1. unknown field

```http
  program=bMcwwoVnbSR&
  occurredAfter=2024-01-01&
  occurredBefore=2024-12-31&
  fields=dataValues[unknown]
```

returned all fields which suggests the `fields` param is correct. For metadata it will only return the `dataValues.id`. That is inline with tracker vs metadata behavior of objects like `fields=dataValues`.

We now return an empty object like we do for

```http
### events with unknown parent
  program=bMcwwoVnbSR&
  occurredAfter=2024-01-01&
  occurredBefore=2024-12-31&
  fields=unknown
```

or

```http
### events with unknown child
  program=bMcwwoVnbSR&
  occurredAfter=2024-01-01&
  occurredBefore=2024-12-31&
  fields=dataValues[value,unknown]
```

which only returns `value`.

2. The following fail with an HTTP 500 which are now rejected with a 400

  | fields                   |
  |--------------------------|
  | ]                        |
  | )                        |
  | group[name]]             |                                                              | group[group[name)]),code |
  | group[name))             |
  | group[name],id]          |
  | group[name],id)          |

3. `fields=[value]` is ignored by the current field filter parser. This is now rejected so users get an HTTP 400

4.

```http
### excluding a field and its children
GET {{PROTOCOL}}://{{AUTH}}@{{HOST}}/api/organisationUnits?
  pageSize=1&
  fields=:all,!parent

### excluding a field and its children
GET {{PROTOCOL}}://{{AUTH}}@{{HOST}}/api/organisationUnits?
  pageSize=1&
  fields=:all,!parent[]

### NPE in current impl when given a child, which is now treated like the above
GET {{PROTOCOL}}://{{AUTH}}@{{HOST}}/api/organisationUnits?
  pageSize=1&
  fields=:all,!parent[name]
```

5. https://dhis2.atlassian.net/browse/DHIS2-19950 - this PR adds validations so requests described in the issue are rejected with a 400

### To Discuss For v43

* `fields=name  code` leads to field `namecode`. Do we want to/have to strip inner whitespace?
* `fields=!:all` will include all. The exclusion is ignored. I think we should reject excluding presets.

## Differences in tracker vs metadata

These field filter features do not work in tracker

* https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/metadata.html#webapi_field_transformers
  * `skipSharing`: because we don't even expose this parameter
  * `defaults`: because we don't even expose this parameter
  * put `attributeValue` onto root object: because tracker entities do not have them, they have their own `attributes` which have nothing to do with `attributeValues`
* only [presets](https://github.com/dhis2/dhis2-core/blob/4459605590f853d9f53870894f24a080271c8c84/dhis-2/dhis-api/src/main/java/org/hisp/dhis/fieldfiltering/FieldPreset.java#L38) `:all` and `:simple`

This behaves differently

* `dataSets.id` is shown for `fields=dataSets` for reference/complex objects like `dataSets` in metatada. `fields=relationships` means `fields=relationships[*]` for tracker. The reason is AFAIK that tracker does not use `Schemas`

## How to apply this to metadata

This PR only applies to tracker endpoints. How could we also use this for metadata? What steps could we take? Tracker does not use all of the field filter features.

### DIFF
--- a/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
+++ b/dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
@@ -215,13 +215,13 @@ class EventsExportControllerTest extends PostgresControllerIntegrationTestBase {
     switchContextToUser(user);
 
     JsonEvent jsonEvent =
-        GET("/tracker/events/{id}?fields=orgUnit,status", event.getUid())
+        GET("/tracker/events/{id}?fields=orgUnit,status::rename(state)", event.getUid())
             .content(HttpStatus.OK)
             .as(JsonEvent.class);
 
-    assertHasOnlyMembers(jsonEvent, "orgUnit", "status");
+    assertHasOnlyMembers(jsonEvent, "orgUnit", "state");
     assertEquals(event.getOrganisationUnit().getUid(), jsonEvent.getOrgUnit());
-    assertEquals(event.getStatus().toString(), jsonEvent.getStatus());
+    assertEquals(event.getStatus().toString(), jsonEvent.getString("state").string());
   }
 
   @Test
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
@@ -31,10 +31,10 @@ package org.hisp.dhis.webapi.controller.tracker.export.enrollment;
 
 import static org.hisp.dhis.webapi.controller.tracker.ControllerSupport.assertUserOrderableFieldsAreSupported;
 import static org.hisp.dhis.webapi.controller.tracker.RequestParamsValidator.validatePaginationParameters;
+import static org.hisp.dhis.webapi.controller.tracker.export.FieldFilterRequestHandler.getRequestURL;
 import static org.hisp.dhis.webapi.controller.tracker.export.enrollment.EnrollmentRequestParams.DEFAULT_FIELDS_PARAM;
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 
-import com.fasterxml.jackson.databind.node.ObjectNode;
 import jakarta.servlet.http.HttpServletRequest;
 import java.util.List;
 import org.hisp.dhis.common.DhisApiVersion;
@@ -44,27 +44,27 @@ import org.hisp.dhis.common.UID;
 import org.hisp.dhis.feedback.BadRequestException;
 import org.hisp.dhis.feedback.ForbiddenException;
 import org.hisp.dhis.feedback.NotFoundException;
-import org.hisp.dhis.fieldfiltering.FieldFilterService;
 import org.hisp.dhis.fieldfiltering.FieldPath;
 import org.hisp.dhis.tracker.PageParams;
 import org.hisp.dhis.tracker.export.enrollment.EnrollmentFields;
 import org.hisp.dhis.tracker.export.enrollment.EnrollmentOperationParams;
 import org.hisp.dhis.tracker.export.enrollment.EnrollmentService;
-import org.hisp.dhis.webapi.controller.tracker.RequestHandler;
+import org.hisp.dhis.tracker.export.fieldfiltering.Fields;
 import org.hisp.dhis.webapi.controller.tracker.view.Enrollment;
+import org.hisp.dhis.webapi.controller.tracker.view.FilteredEntity;
+import org.hisp.dhis.webapi.controller.tracker.view.FilteredPage;
 import org.hisp.dhis.webapi.controller.tracker.view.Page;
 import org.hisp.dhis.webapi.mvc.annotation.ApiVersion;
 import org.mapstruct.factory.Mappers;
-import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
-@OpenApi.EntityType(Enrollment.class)
+@OpenApi.EntityType(org.hisp.dhis.webapi.controller.tracker.view.Enrollment.class)
 @OpenApi.Document(
-    entity = org.hisp.dhis.program.Enrollment.class,
+    entity = org.hisp.dhis.webapi.controller.tracker.view.Enrollment.class,
     classifiers = {"team:tracker", "purpose:data"})
 @RestController
 @RequestMapping("/api/tracker/enrollments")
@@ -77,21 +77,8 @@ class EnrollmentsExportController {
 
   private final EnrollmentService enrollmentService;
 
-  private final EnrollmentRequestParamsMapper paramsMapper;
-
-  private final RequestHandler requestHandler;
-
-  private final FieldFilterService fieldFilterService;
-
-  public EnrollmentsExportController(
-      EnrollmentService enrollmentService,
-      EnrollmentRequestParamsMapper paramsMapper,
-      RequestHandler requestHandler,
-      FieldFilterService fieldFilterService) {
+  public EnrollmentsExportController(EnrollmentService enrollmentService) {
     this.enrollmentService = enrollmentService;
-    this.paramsMapper = paramsMapper;
-    this.requestHandler = requestHandler;
-    this.fieldFilterService = fieldFilterService;
 
     assertUserOrderableFieldsAreSupported(
         "enrollment", EnrollmentMapper.ORDERABLE_FIELDS, enrollmentService.getOrderableFields());
@@ -104,11 +91,11 @@ class EnrollmentsExportController {
       // use the text/html Accept header to default to a Json response when a generic request comes
       // from a browser
       )
-  ResponseEntity<Page<ObjectNode>> getEnrollments(
+  FilteredPage<Enrollment> getEnrollments(
       EnrollmentRequestParams requestParams, HttpServletRequest request)
       throws BadRequestException, ForbiddenException {
     validatePaginationParameters(requestParams);
-    EnrollmentOperationParams operationParams = paramsMapper.map(requestParams);
+    EnrollmentOperationParams operationParams = EnrollmentRequestParamsMapper.map(requestParams);
 
     if (requestParams.isPaging()) {
       PageParams pageParams =
@@ -120,7 +107,8 @@ class EnrollmentsExportController {
       org.hisp.dhis.tracker.Page<Enrollment> page =
           enrollmentsPage.withMappedItems(ENROLLMENT_MAPPER::map);
 
-      return requestHandler.serve(request, ENROLLMENTS, page, requestParams);
+      return new FilteredPage<>(
+          Page.withPager(ENROLLMENTS, page, getRequestURL(request)), requestParams.getFields());
     }
 
     List<Enrollment> enrollments =
@@ -128,24 +116,23 @@ class EnrollmentsExportController {
             .map(ENROLLMENT_MAPPER::map)
             .toList();
 
-    return requestHandler.serve(ENROLLMENTS, enrollments, requestParams);
+    return new FilteredPage<>(
+        Page.withoutPager(ENROLLMENTS, enrollments), requestParams.getFields());
   }
 
   @OpenApi.Response(OpenApi.EntityType.class)
   @GetMapping(value = "/{uid}")
-  public ResponseEntity<ObjectNode> getEnrollmentByUid(
+  public FilteredEntity<Enrollment> getEnrollmentByUid(
       @OpenApi.Param({UID.class, org.hisp.dhis.program.Enrollment.class}) @PathVariable UID uid,
       @OpenApi.Param(value = String[].class) @RequestParam(defaultValue = DEFAULT_FIELDS_PARAM)
-          List<FieldPath> fields)
+          Fields fields)
       throws NotFoundException {
     EnrollmentFields enrollmentFields =
-        EnrollmentFields.of(
-            f -> fieldFilterService.filterIncludes(Enrollment.class, fields, f),
-            FieldPath.FIELD_PATH_SEPARATOR);
+        EnrollmentFields.of(fields::includes, FieldPath.FIELD_PATH_SEPARATOR);
 
     Enrollment enrollment =
         ENROLLMENT_MAPPER.map(enrollmentService.getEnrollment(uid, enrollmentFields));
 
-    return requestHandler.serve(enrollment, fields);
+    return new FilteredEntity<>(enrollment, fields);
   }
 }
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
@@ -32,7 +32,7 @@ package org.hisp.dhis.webapi.controller.tracker.view;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import java.util.LinkedHashMap;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
@@ -51,29 +51,20 @@ import org.springframework.web.util.UriComponentsBuilder;
 @ToString
 @EqualsAndHashCode
 public class Page<T> {
-  @OpenApi.Property(value = OpenApi.EntityType[].class)
+
+  @JsonIgnore private final String key;
+
+  /** Is serialized under the dynamic {@link #key} using {@link #getDynamicItems}. */
   @JsonIgnore
-  @JsonAnyGetter
-  private final Map<String, List<T>> items = new LinkedHashMap<>();
+  @OpenApi.Property(value = OpenApi.EntityType[].class)
+  private final List<T> items;
 
   @JsonProperty private final Pager pager;
 
-  private Page(String key, List<T> values) {
-    this.items.put(key, values);
-    this.pager = null;
-  }
-
-  private Page(String key, org.hisp.dhis.tracker.Page<T> pager, String requestURL) {
-    this.items.put(key, pager.getItems());
-    Integer pageCount = null;
-    if (pager.getTotal() != null) {
-      pageCount = (int) Math.ceil(pager.getTotal() / (double) pager.getPageSize());
-    }
-    String prevPage = getPageLink(requestURL, pager.getPrevPage());
-    String nextPage = getPageLink(requestURL, pager.getNextPage());
-    this.pager =
-        new Pager(
-            pager.getPage(), pager.getPageSize(), pager.getTotal(), pageCount, prevPage, nextPage);
+  private Page(String key, List<T> items, Pager pager) {
+    this.key = key;
+    this.items = items;
+    this.pager = pager;
   }
 
   /**
@@ -86,7 +77,19 @@ public class Page<T> {
       @Nonnull String key,
       @Nonnull org.hisp.dhis.tracker.Page<T> pager,
       @Nonnull String requestURL) {
-    return new Page<>(key, pager, requestURL);
+
+    Integer pageCount = null;
+    if (pager.getTotal() != null) {
+      pageCount = (int) Math.ceil(pager.getTotal() / (double) pager.getPageSize());
+    }
+    String prevPage = getPageLink(requestURL, pager.getPrevPage());
+    String nextPage = getPageLink(requestURL, pager.getNextPage());
+
+    Pager pagerObj =
+        new Pager(
+            pager.getPage(), pager.getPageSize(), pager.getTotal(), pageCount, prevPage, nextPage);
+
+    return new Page<>(key, pager.getItems(), pagerObj);
   }
 
   /**
@@ -94,7 +97,13 @@ public class Page<T> {
    * All other fields will be omitted from the JSON.
    */
   public static <T> Page<T> withoutPager(String key, List<T> items) {
-    return new Page<>(key, items);
+    return new Page<>(key, items, null);
+  }
+
+  /** Items are serialized under dynamic {@link #key}. */
+  @JsonAnyGetter
+  public Map<String, List<T>> getDynamicItems() {
+    return Collections.singletonMap(key, items);
   }
 
   @OpenApi.Shared(name = "TrackerPager")
--- a/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/fields/FieldsConverter.java
+++ b/dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/fields/FieldsConverter.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2004-2025, University of Oslo
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors 
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package org.hisp.dhis.webapi.fields;
+
+import java.util.Map;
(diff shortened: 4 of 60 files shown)


dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/better/FieldsParser.java
```diff
@@ -0,0 +1,514 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import org.hisp.dhis.common.collection.CollectionUtils;
+import org.hisp.dhis.schema.Schema;
+
+public class FieldsParser {
```
jbee: This seems a bit fat for what it is doing :smiling_face_with_tear:

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/better/FieldsPropertyFilter.java
```diff
@@ -0,0 +1,133 @@
...
+   * </ol>
+   *
+   * This means that objects are streamed to JSON with no intermediate in-memory representation if
+   * no transformations are needed. Transformations, however, come with the penalty of double
+   * (de)serialization. This approach is an improvement over the field filtering added in 2.38 where
+   * we always paid the cost of double (de)serialization. Some of the transformations need access to
+   * JSON type information. We could likely serialize/filter/transform the JSON stream if that would
+   * not be necessary.
+   */
+  private void serialize(
```
jbee: This is now short, readable and reasonable :100:

dhis-2/dhis-services/dhis-service-field-filtering/src/main/java/org/hisp/dhis/fieldfiltering/better/package-info.java
```diff
@@ -0,0 +1,44 @@
+/**
+ * DHIS2 field filtering system for JSON API responses.
+ *
+ * <p>User input {@code fields=id,name,group[code]} flows through this pipeline:
+ *
+ * <pre>
```
jbee: Nice!

dhis-2/dhis-services/dhis-service-field-filtering/src/test/java/org/hisp/dhis/fieldfiltering/better/FieldsParserTest.java
```diff
@@ -0,0 +1,739 @@
...
+                new ExpectField(true, "group"),
+                new ExpectField(true, "group.id"),
+                new ExpectField(true, "group.name"),
+                new ExpectField(true, "code"))),
+
+        // testIgnoreWhitespace
+        Arguments.of(
+            " id,name  , group[ id , name  ], code  ",
+            List.of(
+                new ExpectField(true, "id"),
```
jbee: Random line in this test: I feel like `ExpectField` should not be a thing. The parsing should result in a data model that you here could create problematically like this using the actual model classes. If that is not possible it indicates that you mix data and behaviour which usually makes things harder. In my mind the parser outputs a data model (record structure) of the parsed text that has no processing capabilities itself. it only tells you how to process. from that model you might derive another state model to manage your serialisation state if needed. but that model would be "internal" for most places.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
```diff
@@ -44,27 +44,27 @@
...
 import org.mapstruct.factory.Mappers;
-import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
-@OpenApi.EntityType(Enrollment.class)
+@OpenApi.EntityType(org.hisp.dhis.webapi.controller.tracker.view.Enrollment.class)
```
jbee, line 65: The `EntityType` is not necessarily meant to be the actual type used but the persisted or conceptual type it represents. But in the end it is all about making a better reflection of how the API works. But I still would want to know the persisted type. Maybe we have to think about extending the model there. In any case we should not use an OpenAPI annotation for field filtering as those are independent features or concerns that need their freedom to change independently. So I suggest we invent another annotation explicitly for the filtering.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/export/enrollment/EnrollmentsExportController.java
```diff
@@ -120,32 +107,32 @@ ResponseEntity<Page<ObjectNode>> getEnrollments(
...
 
-    return requestHandler.serve(ENROLLMENTS, enrollments, requestParams);
+    return new FilteredPage<>(
+        Page.withoutPager(ENROLLMENTS, enrollments), requestParams.getFields());
   }
 
   @OpenApi.Response(OpenApi.EntityType.class)
   @GetMapping(value = "/{uid}")
-  public ResponseEntity<ObjectNode> getEnrollmentByUid(
+  public FilteredEntity<Enrollment> getEnrollmentByUid(
```
jbee, line 125: I like this new way. We will have to patch OpenAPI to understand the `FilteredEntity` properly (in case you have not done it - just doing a quick scroll though).

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
```diff
@@ -51,29 +51,20 @@
...
 public class Page<T> {
-  @OpenApi.Property(value = OpenApi.EntityType[].class)
+
+  @JsonIgnore private final String key;
+
+  /** Is serialized under the dynamic {@link #key} using {@link #getDynamicItems}. */
   @JsonIgnore
-  @JsonAnyGetter
-  private final Map<String, List<T>> items = new LinkedHashMap<>();
+  @OpenApi.Property(value = OpenApi.EntityType[].class)
```
jbee, line 59: If you wanted to make this work with the annotation on the classes just be aware that (I think) you can also do this per method. The class level is just the default for the methods in the controller. Sometimes defining what the actual `EntityType` is makes more sense per method as not most methods in a controller would be about that type.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/fields/FieldsConverter.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Spring converter that converts field filter strings to {@link Fields}. Works for both
+ * {@code @RequestParam} and properties within {@code @ModelAttribute} objects. Uses request context
+ * to determine the controller's {@link @OpenApi.EntityType} annotation to fetch the {@link Schema}.
+ * The {@link Schema} is needed to expand presets into fields.
+ */
+@Component
+@RequiredArgsConstructor
+public class FieldsConverter implements ConditionalGenericConverter {
```
jbee, line 59: :+1:

jbee (commented):
First of all in the result these are some really good improvements in performance but also in clarity.

The filtered still being slower than the unfiltered does sting a bit but if most time is spend doing other parts of the request that is all we need.

Do we have an idea how memory and GC usage compares? I am sure it is much better now as well but I am curious about some numbers, not that it is very important since we have throughput numbers that do indicate a clear trend.

Code organisation wise I would have liked to see the field filtering core data structures be defined more "together". It is split into many files each doing a little peace of the entire thing. That is always harder to understand even if each little bit is easy.

I leave some suggestion in case you want to make further improvements. We should talk about the use of `EntityType`.

teleivo:
> First of all in the result these are some really good improvements in performance but also in clarity.
> 
> The filtered still being slower than the unfiltered does sting a bit but if most time is spend doing other parts of the request that is all we need.
> 
> Do we have an idea how memory and GC usage compares? I am sure it is much better now as well but I am curious about some numbers, not that it is very important since we have throughput numbers that do indicate a clear trend.

I added section `Allocations and GC` with allocation flamegraphs which show the current implementation does a lot of `String` manipulation as its based on the full dot path.

dhis-2/dhis-web-api/src/main/java/org/hisp/dhis/webapi/controller/tracker/view/Page.java
teleivo: I changed the `FieldsConverter` logic to first look on the method, then the class

dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/tracker/export/event/EventsExportControllerTest.java
enricocolasante, line 1: Should we add some tests in `EventsExportChangeLogsControllerTest`?

muilpp (approved):

PR #21774, opened 2025-08-27
fix: Fix integrity check tests

enricocolasante:
Fix integrity check tests that were failing because `event` table is not there anymore.
Set `occurredDate` when calling `createSingleEvent` in tests now that `occurreddate` in table `singleevent` is not null.

Now the only failing tests are the one in analytics module.


dhis-2/dhis-test-web-api/src/test/java/org/hisp/dhis/webapi/controller/dataintegrity/DataIntegrityProgramStagesNoPrograms.java
```diff
@@ -79,6 +81,28 @@ void testProgramStagesNoPrograms() {
         true);
   }
 
+  @Test
+  void shoulFailIntegrityCheckAndFindData() {
```
muilpp: does it fail because the PS doesn't have a program?
even if that's not the reason, can we add it to the test name?

it's not clear to me why does this test fail even when I read the code.

teleivo (approved):

PR #21775, opened 2025-08-27
 fix: Migrate singleevent status

enricocolasante:
A single event only allows 2 statuses, `ACTIVE` and `COMPLETED`.
- Migrate status in `singleevent` table:
  - Events with not allowed status and without value in `completeddate` column are set to `ACTIVE`
  - Events with not allowed status and with a value in `completeddate` column are set to `COMPLETED`
- Fail migration if there are null values in `occurreddate` column in `singleevent` table and provide info to the user on how to solve the issue.
- Create `inconsistentevent` table only if needed.

***We will create integrity checks in previous versions to look for single events without occurred data***

### DIFF
--- a/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
+++ b/dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
@@ -16,13 +16,21 @@ where eventid in (
 alter table singleevent drop column scheduleddate;
 alter table singleevent drop column enrollmentid;
 
+-- Update status to allowed values for single event
+update singleevent set status = 'ACTIVE'
+where status not in ('ACTIVE','COMPLETED')
+  and completeddate is null;
+
+update singleevent set status = 'COMPLETED'
+where status not in ('ACTIVE','COMPLETED')
+  and completeddate is not null;
+
 -- Create primary key
 alter table singleevent add primary key (eventid);
 -- Make uid column unique
 alter table singleevent add constraint unique_singleevent_uid unique (uid);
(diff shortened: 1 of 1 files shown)


dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
```diff
@@ -34,6 +42,16 @@ alter table singleevent alter column status set not null;
 alter table singleevent alter column eventdatavalues set default '{}'::jsonb;
 alter table singleevent alter column eventdatavalues set not null;
 
+-- Make occurreddate column not nullable. Raise an error if null values are present.
+do $$
+    begin
+        alter table singleevent alter column occurreddate set not null;
+    EXCEPTION
+        WHEN not_null_violation THEN
+            RAISE EXCEPTION 'There is inconsistent data in your DB. Please check https://github.com/dhis2/dhis2-releases/blob/master/releases/2.43/migration-notes.md#null-occurred-date-for-single-event to have more information on the issue and to find ways to fix it. Detailed error message: %', SQLERRM;
```
muilpp, line 51: this link and the one below don't exist yet, will you create them later on?

muilpp (approved):

dhis-2/dhis-support/dhis-support-db-migration/src/main/resources/org/hisp/dhis/db/migration/2.43/V2_43_21__split_event_table.sql
enricocolasante: Yes, we will take care of documentation here https://dhis2.atlassian.net/browse/DHIS2-19614

ameenhere (approved):

PR #21790, opened 2025-08-29
fix: escape routes once

KaiVandivier:
[DHIS2-20010](https://dhis2.atlassian.net/browse/DHIS2-20010)

Adds a solution suggested in the issue above -- before/after:

route-fix-before

route-fix-after


david-mackessy (approved):
It would be nice to have a unit test. It would help cover this scenario and could potentially catch future issues within `createTargetUri`

PR #21799, opened 2025-09-01
fix: Invalidate ownership cache after enrollment persisted [DHIS2-18338]

muilpp:


dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityCacheTest.java
```diff
@@ -129,6 +146,31 @@ void shouldFindTrackedEntityWhenAnotherOwnerAlreadyInCache()
...
+    User user = manager.get(User.class, "lPaILkLkgOM");
+
+    injectSecurityContextUser(user);
+
+    assertFalse(
+        trackerOwnershipManager.hasAccess(
+            UserDetails.fromUser(regularUser), trackedEntity, programA));
+
+    Enrollment enrollment =
+        createEnrollment(trackedEntity.getUid(), programA.getUid(), "h4w96yEMlzO");
```
enricocolasante: Would it make sense to extract `h4w96yEMlzO` to a variable and give it a meaningful name?

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackedEntityProgramOwnerService.java
```diff
@@ -117,4 +132,8 @@ private void updateTrackedEntityProgramOwner(
   public TrackedEntityProgramOwner getTrackedEntityProgramOwner(TrackedEntity te, Program program) {
     return trackedEntityProgramOwnerStore.getTrackedEntityProgramOwner(te, program);
   }
+
+  private String getOwnershipCacheKey(TrackedEntity trackedEntity, Program program) {
```
enricocolasante: It would be good to have this extracted somewhere and use the same code to create the key here and in `DefaultTrackerOwnershipManager`

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackedEntityProgramOwnerService.java
```diff
@@ -43,10 +46,19 @@
...
-@RequiredArgsConstructor
 @Service("org.hisp.dhis.tracker.acl.TrackedEntityProgramOwnerService")
 public class DefaultTrackedEntityProgramOwnerService implements TrackedEntityProgramOwnerService {
   private final TrackedEntityProgramOwnerStore trackedEntityProgramOwnerStore;
+  private final Cache<OrganisationUnit> ownerCache;
+
+  public DefaultTrackedEntityProgramOwnerService(
+      TrackedEntityProgramOwnerStore trackedEntityProgramOwnerStore, CacheProvider cacheProvider) {
+    checkNotNull(trackedEntityProgramOwnerStore);
+    checkNotNull(cacheProvider);
```
enricocolasante:
```suggestion
```
[`spring injected contstructors do not need nullability annotations`](https://github.com/dhis2/wow-backend/blob/master/guides/code-guide.md#constructors)

dhis-2/dhis-test-integration/src/test/java/org/hisp/dhis/tracker/export/trackedentity/TrackedEntityCacheTest.java
```diff
@@ -129,6 +146,31 @@ void shouldFindTrackedEntityWhenAnotherOwnerAlreadyInCache()
...
 
+  @Test
+  void shouldHaveAccessAfterEnrollmentPersistedAndOwnershipCreatedAndCacheInvalidated() {
+    TrackedEntity trackedEntity = manager.get(TrackedEntity.class, "H8732208127");
+    assertNotNull(trackedEntity);
+    User user = manager.get(User.class, "lPaILkLkgOM");
+
+    injectSecurityContextUser(user);
+
+    assertFalse(
```
enricocolasante, line 162: Maybe I would add some comments to describe what is happening with the cache, when it is populated and when it is invalidated.
Cache population/invalidation is a side effect so it is not easy to understand looking at the calls here.

dhis-2/dhis-services/dhis-service-tracker/src/main/java/org/hisp/dhis/tracker/acl/DefaultTrackedEntityProgramOwnerService.java
```diff
@@ -43,10 +46,16 @@
 /**
  * @author Ameen Mohamed
  */
-@RequiredArgsConstructor
 @Service("org.hisp.dhis.tracker.acl.TrackedEntityProgramOwnerService")
 public class DefaultTrackedEntityProgramOwnerService implements TrackedEntityProgramOwnerService {
   private final TrackedEntityProgramOwnerStore trackedEntityProgramOwnerStore;
+  private final Cache<OrganisationUnit> ownerCache;
```
ameenhere, line 52: Since this is "not" an inMemory cache, your approach of using the same ownerCache will work. 

But as a convention in our core, it will be safer to use the `ApplicationEventPublisher` to publish `CacheInvalidationEvent`  with the key and region you want to invalidate. 

I don't have strong opinions on whether we have to change this specific case though. Perhaps it is alright the way it is?  Your thoughts @muilpp @enricocolasante ?

muilpp: If it's a convention in the core, then I'll refactor this to use the event publisher.
Just curious, why wouldn't it work if the cache was in memory? I thought the `LocalCache` we use in tests is in memory, and there it works.

ameenhere: Ah. Ignore my previous comment. I mistook that ApplicationEventPublisher supports publishing events to all instances in a clustered setup. Basically the difference would have been, is there a way to invalidate cache in all nodes of a clustered dhis2 setup if the cache used was inMemory (Since then each instance will hold its own cache in memory). 

But currently there is no way to invalidate inMemory cache in "all" nodes effectively. And the understanding is that, they are in memory because stale data read is perhaps fine. 

So there is effectively no difference. Maybe your implementation is more consistently testable through our tests. Since using the event publisher there is some asynchronicity which means the changes might not happen real-time. 

Lets just go with your current implementation then.

PR #21800, opened 2025-09-02
fix: Use program owner instead of enrollment org unit [DHIS2-18338][2.42]

muilpp:
When we fetched enrollments, we were checking user access against the enrollment’s org unit. That was a mistake, since that field just shows where the enrollment came from and isn’t meant for ACL checks.

I’ve fixed it so we now use the org unit owner to validate access instead.


muilpp:
will be done [here](https://github.com/dhis2/dhis2-core/pull/21802)

